PR #29, opened 2016-07-27
Get the amount of clients in current conversation

bennyn:


app/script/util/DebugUtil.coffee
```diff
@@ -32,6 +32,17 @@ class z.util.DebugUtil
     return new Promise (resolve) =>
       @user_repository.get_user_by_id user_id, (user_et) -> resolve user_et
 
+  get_amount_of_clients_in_conversation: ->
```
gregor: Please name `get_number_of_....` like similar ones in the project

PR #41, opened 2016-07-29
Make screen sharing possible even if user has no camera (WEBAPP-3050)

mythsunwind:


gregor:
@mythsunwind Awesome catch. Thanks for the fix.

PR #66, opened 2016-08-06
Ignoring events with outdated schema

bennyn:
- [x] Don't support old events from WebSocket
- [x] Don't support old events from notification stream
- [x] Check if there are still old events (especially `VOICE_CHANNEL_ACTIVATE` & Co.)
- [x] https://wearezeta.atlassian.net/browse/WEBAPP-3076
- [x] https://wearezeta.atlassian.net/browse/WEBAPP-3077


app/script/event/EventRepository.coffee
```diff
@@ -329,15 +330,17 @@ class z.event.EventRepository
...
   ###
   _handle_event: (event, source) ->
     return new Promise (resolve, reject) =>
       sending_client = event.data?.sender
       if sending_client
         log_message = "Received encrypted event '#{event.type}' from client '#{sending_client}' of user '#{event.from}'"
       else if event.from
-        log_message = "Received unencrypted event '#{event.type}' from user '#{event.from}'"
+        log_message = "Received plain event '#{event.id}' of type '#{event.type}' from client '#{sending_client}' of user '#{event.from}'"
+        if event.type is z.event.Backend.CONVERSATION.MESSAGE_ADD
```
gregor: Aren't there more event types that should throw the OUTDATED_SCHEMA error?

app/script/storage/StorageRepository.coffee
```diff
@@ -215,45 +215,44 @@ class z.storage.StorageRepository extends cryptobox.CryptoboxStore
...
-        raw: event
+        raw: otr_message_event
         meta:
-          timestamp: new Date(event.time).getTime()
+          timestamp: new Date(otr_message_event.time).getTime()
           version: 1
+        mapped: mapped_json
+
+      # We don't need to keep ciphertext once it has been successfully decrypted
+      event_object.raw.data = undefined
```
gregor, line 235: Should we change this to some flag instead of undefined so that we know that we successfully decrypted the event?

app/script/event/EventRepository.coffee
herrmannplatz: could be also z.event.Backend.CONVERSATION.KNOCK and z.event.Backend.CONVERSATION.ASSET_ADD

app/script/storage/StorageRepository.coffee
bennyn: It has been always like this? So I would keep it for now.

bennycode, line 235: I think we went for this solution (storing `undefined`) to store less data in the database. So it's ok. We will change properties later when we introduce proper record classes.

PR #70, opened 2016-08-08
Fixed bug with big link previews, resolves WEBAPP-2921

bennyn:


app/script/cryptography/CryptographyRepository.coffee
```diff
@@ -404,7 +404,7 @@ class z.cryptography.CryptographyRepository
       else
         payload = @_construct_payload @current_client().id
         use_local_sessions = true
-        @logger.log @logger.levels.INFO, 'Encrypt message using local sessions'
+        @logger.log @logger.levels.INFO, "Encrypt '#{generic_message.content}' message using local sessions"
       return @get_sessions user_client_map, use_local_sessions
```
herrmannplatz, line 408: we should not log the content

bennycode, line 408: In this case `generic_message.content` is the type and not the textual content of the message. :smiley:

PR #74, opened 2016-08-09
Audio can only be muted if stream is existing

gregor:


bennycode:
Looks good to me but we definitely need to start writing unit tests & integrations tests for our calling functionality. Even if it's just about small things, we need to start somewhere.

gregor:
Definitely. Will add them for this one.

PR #80, opened 2016-08-10
Fixes for error reports

gregor:
- WebSocket fails to reconnect due to broken log message
- logging cleanup
- add 413 backend error to ignored list
- prevent exception when active_conversation is not set
- execute on_initial_rendering() only if set
- add safeguard for attempted calls from wrapper menus
- close archive after unarchiving last conversation
- execute menu actions on conversations only

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -340,6 +340,8 @@ class z.conversation.ConversationRepository
   @return [Boolean] Is the message marked as read
   ###
   is_message_read: (conversation_id, message_id) =>
+    return false if not conversation_id or not message_id
+
     conversation_et = @get_conversation_by_id conversation_id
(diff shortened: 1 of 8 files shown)


app/script/conversation/ConversationRepository.coffee
```diff
@@ -340,6 +340,8 @@ class z.conversation.ConversationRepository
   @return [Boolean] Is the message marked as read
   ###
   is_message_read: (conversation_id, message_id) =>
+    return false if not conversation_id or not message_id
```
bennycode, line 343: What is calling this method without parameters??

gregor, line 343: I hope nothing. Just adding a safeguard in order to verify my hypothesis that `conversation_et = @get_conversation_by_id conversation_id` in L345 sometimes returns undefined and we should continue to move in the Promise direction here as well with proper error handling.

PR #84, opened 2016-08-11
Log Localytics attributes as Strings

bennyn:


app/script/tracking/EventTrackingRepository.coffee
```diff
@@ -145,7 +145,9 @@ class z.tracking.EventTrackingRepository
     return @user_properties.settings.privacy.improve_wire
 
   _log_event: (event_name, attributes) =>
-    @logger.log "Localytics event '#{event_name}'", attributes
+    if attributes
+      @logger.log "Localytics event '#{event_name}' with attributes: #{JSON.stringify(attributes)}"
+    @logger.log "Localytics event '#{event_name}' without attributes"
```
gregor: In this case you need an else or a return or we will get two log lines for an event with attributes.

PR #86, opened 2016-08-11
Fix: Leave message is shown twice when leaving a conversation

bennyn:
**Screenshot of Bug**

unbenannt

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -554,7 +554,6 @@ class z.conversation.ConversationRepository
   leave_conversation: (conversation_et, next_conversation_et, callback) =>
     @conversation_service.delete_members conversation_et.id, @user_repository.self().id
     .then (response) =>
-      amplify.publish z.event.WebApp.EVENT.INJECT, response
       @member_leave conversation_et, response
     .then =>
       if callback?


app/script/conversation/ConversationRepository.coffee
```diff
@@ -554,7 +554,6 @@ class z.conversation.ConversationRepository
   leave_conversation: (conversation_et, next_conversation_et, callback) =>
     @conversation_service.delete_members conversation_et.id, @user_repository.self().id
     .then (response) =>
-      amplify.publish z.event.WebApp.EVENT.INJECT, response
       @member_leave conversation_et, response
```
gregor, line 557: This way it is not saved in the database though, is it? Shouldn't we remove line 558 instead? My guess it that with the current code change the message will be gone after page reload.

bennycode, line 557: You are right. Will check if line 558 brings the expected behaviour.

bennycode, line 557: Ok, did that... Now I just get one leave message but the conversation is not archived :'(

gregor, line 557: I guess it is archived but due to the async call with the INJECT it is immediately unarchived again when the leave event is properly processed... :(

bennycode, line 557: We need something like this:

``` coffeescript
  leave_conversation: (conversation_et, next_conversation_et, callback) =>
    @conversation_service.delete_members conversation_et.id, @user_repository.self().id
    .then (response) =>
      return event_repo.store
    .then (event)
      return @member_leave conversation_et, response
    .then =>
      ...
```

PR #90, opened 2016-08-15
delete everywhere receiving support (WEBAPP-3081)

herrmannplatz:


test/unit_tests/conversation/ConversationRepositorySpec.coffee
```diff
@@ -288,6 +288,48 @@ describe 'z.conversation.ConversationRepository', ->
...
+        data:
+          message_id: message_et.id
+        from: user_repository.self().id
+        time: new Date().toISOString()
+
+      conversation_repository.message_deleted event
+      expect(conversation_repository._delete_message).toHaveBeenCalled()
+      expect(conversation_repository._add_delete_message).not.toHaveBeenCalled()
+
+    it 'should delete message and add delete message if user is not self', ->
```
bennycode: The `it 'should ...` pattern is overweight. Instead of `it should delete message` you can simply write `it deletes messages`. Because that's what your test proves. :)

app/script/event/Backend.coffee
```diff
@@ -46,6 +46,7 @@ z.event.Backend =
     MEMBER_LEAVE: 'conversation.member-leave'
     MEMBER_UPDATE: 'conversation.member-update'
     MESSAGE_ADD: 'conversation.message-add'
+    MESSAGE_HIDDEN: 'conversation.message-hidden'
```
bennycode, line 49: Is this really a type from the backend or a custom type which we invented?

PR #91, opened 2016-08-15
Catch "message is too big" bugs

bennyn:


bennycode:
Needs to pass tests by @Yserz & @mythsunwind ...

PR #92, opened 2016-08-15
message context menu (WEBAPP-2861)

herrmannplatz:
This adds the new message context menu to the UI. 

The basic idea is that the context menu emits a custom event once its clicked and the data is updated. This way you can listen to the custom event and stop the propagation to e.g. display a different element with the data.

### DIFF
--- a/app/script/components/contextMenu.coffee
+++ b/app/script/components/contextMenu.coffee
@@ -1,0 +1,75 @@
+#
+# Wire
+# Copyright (C) 2016 Wire Swiss GmbH
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see http://www.gnu.org/licenses/.
+#
+
+window.z ?= {}
+z.components ?= {}
+
+class z.components.ContextMenuEntries
+
+  constructor: ->
+    @entries = []
+
+  push: (label, action) ->
+    @entries.push
+      label: label
+      action: action
+
+
+class z.components.ContextMenuViewModel
+
+  constructor: (params, component_info) ->
+    # parameter list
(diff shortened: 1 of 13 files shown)


app/script/components/contextMenu.coffee
```diff
@@ -0,0 +1,101 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see http://www.gnu.org/licenses/.
+#
+
+window.z ?= {}
+z.components ?= {}
+
+z.components.ContextMenuEvent =
+  CONTEXT_MENU: 'customContextMenu'
```
bennycode: Usually we declare custom values like this:

``` coffee
z.components.ContextMenuEvent =
  CONTEXT_MENU: 'z.components.ContextMenuEvent::CONTEXT_MENU'
```

With this we can copy the value `z.components.ContextMenuEvent::CONTEXT_MENU` and search through the code where this value is being used. Makes it easier to find it's reference.

app/script/components/contextMenu.coffee
```diff
@@ -0,0 +1,101 @@
...
+
+# Display context menu for the given entries. Entries should be provided as an Array of object containing
```
bennycode: `Array of objects` (s)

app/script/components/contextMenu.coffee
```diff
@@ -0,0 +1,101 @@
...
+#
+# entries example:
+# [
+#   {label: 'copy', action: 'action-id'},
+#   {label: 'delete', action: 'action-id'}
+# ]
+#
+# @param [Object] params
+# @option params [String] tag
+# @option params [Object] data additional data that is passed then an items in the context menu is clicked
```
bennycode: `additional data that is passed then an items in the context menu is clicked`
->
`additional data that is passed when items in the context menu are clicked`

app/style/components/context-menu.less
```diff
@@ -0,0 +1,30 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+context-menu {
```
lipis, line 20: `.context-menu` or I'm missing something?

herrmannplatz, line 20: nope its the custom element itself

lipis, line 20: just wanted to verify that I'm missing something :D 🏇

PR #99, opened 2016-08-18
Backwards compatibility and enabling unit tests for AudioRepository

gregor:
Turns out the bug in Karma was new in the latest release already filed and fixed:
https://github.com/karma-runner/karma/issues/2310

I also added backwards compatibility for Firefox for the non Promise based Audio.play()

### DIFF
--- a/app/script/audio/AudioRepository.coffee
+++ b/app/script/audio/AudioRepository.coffee
@@ -165,16 +165,24 @@ class z.audio.AudioRepository
   @return [Promise] Resolves with the HTMLAudioElement
   ###
   _play: (audio_id, audio_element, play_in_loop = false) ->
-    return new Promise (resolve, reject) ->
+    if not audio_id or not audio_element
+      return Promise.reject new z.audio.AudioError 'Missing AudioElement or ID', z.audio.AudioError::TYPE.NOT_FOUND
+
+    return new Promise (resolve, reject) =>
       if audio_element.paused
         audio_element.loop = play_in_loop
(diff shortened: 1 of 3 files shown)


app/script/audio/AudioRepository.coffee
```diff
@@ -165,16 +165,24 @@ class z.audio.AudioRepository
...
           resolve audio_element
-        .catch (error) ->
-          reject new z.audio.AudioError error.message, z.audio.AudioError::TYPE.FAILED_TO_PLAY
+
+        if play_promise
+          play_promise.then(_play_success).catch (error) ->
+            reject new z.audio.AudioError error.message, z.audio.AudioError::TYPE.FAILED_TO_PLAY
+        else
+          _play_success()
       else
```
herrmannplatz, line 186: Why not using a single promise chain like:

```
Promise.resolve()
.then ->
  if not audio_id or not audio_element
    throw new z.audio.AudioError 'Missing AudioElement or ID', z.audio.AudioError::TYPE.NOT_FOUND
.then ->
  if audio_element.paused
    audio_element.loop = play_in_loop
    audio_element.currentTime = 0 if audio_element.currentTime isnt 0
    return audio_element.play()
  else
    throw new z.audio.AudioError 'Sound is already playing', z.audio.AudioError::TYPE.ALREADY_PLAYING
.catch (error) ->
  throw new z.audio.AudioError error.message, z.audio.AudioError::TYPE.FAILED_TO_PLAY
.then ->
  @currently_looping[audio_id] = audio_id if play_in_loop
  return audio_element
```

PR #111, opened 2016-08-23
Conversation bot check does not fail for user without email

gregor:


lipis:
I'm starting to believe that you are always hunting the _cool_ issue numbers... :)

PR #121, opened 2016-08-26
Using release versions for "webrtc-adapter"

bennyn:


gregor:
Did you test this with a few calls? It used to not work with our minification (with open issue on the adapter repo) and they explicitly said that gh-pages is the way to go.

gregor:
Other than that definitely 👍Getting rid of yet another non standard dependency.

bennycode:
I just made a test call with @Yserz. Was working fine! :+1:

PR #140, opened 2016-08-30
Promisify 'z.conversation.ConversationRepository.get_conversation_by_id'

gregor:


gregor:
Will open a new PR once we have removed the wrapper dependency on it returning a value instead of a Promise

PR #141, opened 2016-08-31
Send confirmation status only to other user

gregor:
- implemented the new backend endpoint to only send the confirmation to a specific user in a conversation
- disabled the sending completely due to the side effects found in nightly run

### DIFF
--- a/app/script/conversation/ConversationService.coffee
+++ b/app/script/conversation/ConversationService.coffee
@@ -377,16 +377,20 @@ class z.conversation.ConversationService
     }
   }
 
-  @param conversation_id [String] ID of conversation to send message in
+  @note Options for the precondition check on missing clients are:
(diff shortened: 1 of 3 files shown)


app/script/conversation/ConversationService.coffee
```diff
@@ -381,12 +381,15 @@ class z.conversation.ConversationService
   @param payload [Object] Payload to be posted
   @option [OtrRecipients] recipients Map with per-recipient data
   @option [String] sender Client ID of the sender
-  @param force_sending [Boolean] Should the backend ignore missing clients
+  @param report_missing [Array<String>|Boolean] Option to handle missing clients
   @return [Promise] Promise that resolve when the message was sent
```
herrmannplatz, line 386: Mixing up these two values is a bit misleading. Since `report_missing=true` actually forces the message to be send instead of reporting missing clients.

PR #149, opened 2016-09-01
like ui

herrmannplatz:


app/script/view_model/MessageListViewModel.coffee
```diff
@@ -534,3 +534,7 @@ class z.ViewModel.MessageListViewModel
   click_on_cancel_request: (message_et) =>
     next_conversation_et = @conversation_repository.get_next_conversation @conversation_repository.active_conversation()
     @user_repository.cancel_connection_request message_et.other_user(), next_conversation_et
+
+  click_on_like: (message_et, like) =>
+    reaction = if like then z.message.ReactionType.LIKE else z.message.ReactionType.NONE
+    @conversation_repository.send_reaction(@conversation(), message_et, reaction)
```
gregor: No need for parentheses

app/script/conversation/ConversationRepository.coffee
```diff
@@ -1659,13 +1660,18 @@ class z.conversation.ConversationRepository
   _on_reaction: (conversation_et, event_json) ->
     @get_message_from_db conversation_et, event_json.data.message_id
     .then (message_et) =>
+      return @_update_message_reactions message_et, event_json if message_et
+    .then (message_et) =>
```
gregor, line 1674: Thanks

app/script/conversation/ConversationRepository.coffee
```diff
@@ -1659,13 +1660,18 @@ class z.conversation.ConversationRepository
...
       @logger.log "Updated reactions of message '#{message_et.id}' in database", message_et
       return @conversation_service.update_message_reactions_in_db message_et.primary_key, message_et.reactions()
     .then =>
       @logger.log @logger.levels.DEBUG, "Reaction to message '#{event_json.data.message_id}' in conversation '#{conversation_et.id}'", event_json
       return conversation_et.get_message_by_id event_json.data.message_id
     .then (message_et) =>
       return @_update_message_reactions message_et, event_json if message_et
+    .then (message_et) =>
+      @_update_user_ets message_et
+      return message_et
```
gregor: But I actually have my doubts about what we do here. :( Do we have to update the message reactions twice? And what happens if the message is no longer in the view since you are in another chat? `get_message_by_id` will fail in that case.

herrmannplatz: Indeed. The problem is that the one you get from the db is a different instance. Added it for the moment, but there should be a better way.

app/script/conversation/ConversationRepository.coffee
```diff
@@ -1815,10 +1821,21 @@ class z.conversation.ConversationRepository
...
     @user_repository.get_user_by_id message_et.from, (user_et) =>
       message_et.user user_et
+
       if message_et.is_member()
         @user_repository.get_users_by_id message_et.user_ids(), (user_ets) ->
           message_et.user_ets user_ets
-      else if message_et.has_asset_text()
+        return
+
+      if message_et.reactions?
```
gregor, line 1830: check for message type instead as all content message have reactions `if message_et.is_content()`

PR #150, opened 2016-09-01
Handle incoming delivery confirmation events

gregor:


app/script/conversation/EventMapper.coffee
```diff
@@ -95,6 +95,7 @@ class z.conversation.EventMapper
 
     if message_et.is_reactable()
       message_et.reactions event.reactions or {}
+      message_et.status event.status if event.status
```
herrmannplatz, line 99: we should set this for every message or?

gregor: Default is set in the message constructor.

PR #166, opened 2016-09-05
Show desktop notification for reactions to your message

gregor:


app/script/entity/message/Message.coffee
```diff
@@ -29,11 +29,11 @@ class z.entity.Message
     message_ets.sort (m1, m2) -> m1.timestamp > m2.timestamp
 
   # Construct a new base message entity.
-  constructor: ->
+  constructor: (id = '0', super_type = '') ->
```
bennycode, line 32: The default value surprises me a bit. It's a number which comes as a String and I think that Message IDs are UUIDs, are they? So this should be more obvious from the default value's perspective.

PR #168, opened 2016-09-05
Removing redundant 'else'

lipis:
In all those cases the `else` is redundant plus the extra indentation..

http://stackoverflow.com/a/3261855/8418

Didn't find anything for CoffeeLint.. http://eslint.org/docs/rules/no-else-return

Search query used to find them all :)

```
return (.+)\n( +)else\n
```

In some cases I added an extra empty line.. in some not :) What do you guys think?


bennycode:
I hope that there are tests to prove that the behaviour hasn't changed. ;)

lipis:
Tests were nice in this case..

bennycode:
I am afraid of merging this without tests. WDYT? I don't want to take the risk that it breaks something we don't know yet.

lipis:
There are conflicts at the moment.. will take another swing at it :)

bennycode:
@lipis What's the status on this one? Closing it?

lipis:
Looks like the swing took longer.. let's close it until further notice :D

PR #169, opened 2016-09-05
Imports on separate lines

lipis:
https://www.python.org/dev/peps/pep-0008/#imports

(Even though I also prefer the `from subprocess import Popen, PIPE` to be on separate lines)


bennycode:
Ok, please update the branch and merge it.

PR #194, opened 2016-09-08
Improvement/image compression 

herrmannplatz:
This PR moves image compression into a web worker. I used jimp for now since it can handle EXIF rotation, scaling as well as changing the quality. Payload is with 300kb quite high though, but we can get rid of some libs as well.


herrmannplatz:
Findings so far. Biggest bottleneck seems to be of initialization the worker. Which is understandable, since jimp is quite big. So i decided to spawn one worker and keep it around.

I would like to merge this and see how it feels. Next steps would be to do the same for profile assets and checking if there is an alternative to jimp.

app/script/assets/AssetService.coffee
```diff
@@ -143,28 +133,21 @@ class z.assets.AssetService
...
+    .then ([compressed_image, compressed_bytes]) ->
+      @_upload_asset compressed_bytes, options
+      .then (data) ->
+        [key_bytes, sha256, key, token] = data
+        image_meta_data = new z.proto.Asset.ImageMetaData compressed_image.width, compressed_image.height
+        asset = new z.proto.Asset()
+        asset.set 'original', new z.proto.Asset.Original file.type, compressed_bytes.length, null, image_meta_data
+        asset.set 'uploaded', new z.proto.Asset.RemoteData key_bytes, sha256, key, token
+        return asset
+      .catch (error) =>
```
gregor: catch does not catch enough

app/script/assets/AssetService.coffee
```diff
@@ -31,14 +31,9 @@ class z.assets.AssetService
     @rotator = new zeta.webapp.module.image.rotation.ImageFileRotator()
     @compressor = new zeta.webapp.module.image.ImageCompressor()
 
-    @BOUNDARY = 'frontier'
+    @worker = new z.util.Worker 'worker/image-worker.js'
```
gregor: maybe a little more specific on the naming? `compression_worker` `image_worker`

app/worker/image-worker.coffee
```diff
@@ -0,0 +1,19 @@
...
+self.addEventListener 'message', (e) ->
+  data = e.data
+
+  Jimp.read(data.message).then (image) ->
+
+    if image.bitmap.width > maxSize or image.bitmap.height > maxSize
+      image.scaleToFit maxSize, maxSize
+
+    if image.bitmap.data.length > 5 * 1024
+      image.quality 75
```
lipis: maybe putting those values in some sort of const..

app/worker/image-worker.coffee
```diff
@@ -0,0 +1,19 @@
+importScripts 'jimp.min.js'
+
+maxSize = 1448
```
lipis: `MAX_SIZE`?

app/worker/image-worker.coffee
```diff
@@ -0,0 +1,20 @@
...
+self.addEventListener 'message', (e) ->
+  data = e.data
+
+  Jimp.read(data.message).then (image) ->
+
+    if image.bitmap.width > MAX_SIZE or image.bitmap.height > MAX_SIZE
+      image.scaleToFit MAX_SIZE, MAX_SIZE
+
+    if image.bitmap.data.length > MAX_FILE_SIZE
+      image.quality 75
```
lipis: maybe we should enable this rule: http://eslint.org/docs/rules/no-magic-numbers or at least start with the `warning` :)

PR #238, opened 2016-09-22
Improvement/asset download error

herrmannplatz:


app/script/util/util.coffee
```diff
@@ -91,7 +91,11 @@ z.util.load_url_buffer = (url, xhr_accessor_function) ->
     xhr = new XMLHttpRequest()
     xhr.open 'GET', url, true
     xhr.responseType = 'arraybuffer'
-    xhr.onload = -> resolve [xhr.response, xhr.getResponseHeader 'content-type']
+    xhr.onload = ->
+      if xhr.status is 200
```
gregor, line 95: What happens if you get a not modified or similar? Don't we also have statics for these response codes?

herrmannplatz: 304 are not propagated and will result in 200

> For 304 Not Modified responses that are a result of a user agent generated conditional request the user agent must act as if the server gave a 200 OK response with the appropriate content.

PR #241, opened 2016-09-23
🏖 user avatar rewrite

herrmannplatz:
## Description
Improves user avatar hiding and timestamp rendering. Should fix issues when deleting messages and makes it more robust when user receives multiple messages in a short time frame.

## Type of change
Before the avatar and timestamp rendering was done in after_render and before_remove with jquery. This had several issues and there was also the need to know more about selectors then needed. Now its controlled via bindings, which checkes if the avatar should be visible.

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My Pull Request contains test coverage for my changes.


gregor:
We talked within the team. We will merge this now to avoid wrong/missing avatars as much as possible. Release will be pushed out a day or to in order to test and monitor any side effects that we have not found while implementing this change.

PR #244, opened 2016-09-26
Handle URL parameters on auth page

gregor:


app/script/view_model/AuthViewModel.coffee
```diff
@@ -205,6 +205,23 @@ class z.ViewModel.AuthViewModel
...
+    $(window)
+      .on 'dragover drop', -> false
+      .on 'hashchange', @_on_hash_change
+      .on 'keydown', @keydown_auth
+
+    $("[id^='wire-login'], [id^='wire-register'], [id^='wire-verify']").prevent_prefill()
+
+    # Select country based on location of user IP
+    @country_code (z.util.CountryCodes.get_country_code($('[name=geoip]').attr 'country') or 1).toString()
+    @changed_country_code()
```
bennycode, line 220: This functions expects `view_model, event`.

lipis: and `view_model` is not being used anywhere in that function.. so we might have to do something about it...

gregor: That's the parameters knockout will pass in on the bindings

PR #252, opened 2016-09-28
Unblock sending queue (WEBAPP-3222)

gregor:


app/script/conversation/ConversationRepository.coffee
```diff
@@ -133,11 +134,20 @@ class z.conversation.ConversationRepository
     queue_entry = @sending_queue()[0]
     if queue_entry
       @sending_blocked = true
+      @sending_interval = window.setInterval =>
+        return if @conversation_service.client.request_queue_blocked()
+        @logger.log @logger.levels.ERROR, 'Sending of message from queue failed, unblocking queue', {entry_function: @sending_queue()[0].function, queue: @sending_queue()}
+        @sending_blocked = false
+        @_execute_from_sending_queue()
+        window.clearInterval @sending_interval
+      , 60000
```
lipis: Magic! :) should be a constant..

gregor: Done

lipis: Somewhere in https://github.com/wireapp/wire-webapp/blob/dev/app/script/config.coffee would be better I would say :/

gregor: Done. I am not a big fan of the config file. We never remove unsused stuff from there. Just stumpled upon three obsolete entries that I removed

PR #253, opened 2016-09-28
Improved client handling (WEBAPP-2739, WEBAPP-2747)

gregor:


app/script/client/ClientRepository.coffee
```diff
@@ -534,45 +506,29 @@ class z.client.ClientRepository
...
-  @private
-  @param client_id [String] ID of client to be removed
-  ###
-  _client_removal: (client_id) ->
-    @remove_client client_id
-    .then =>
-      @logger.log "Removed client from database: #{client_id}"
-      amplify.publish z.event.WebApp.SIGN_OUT, 'client_removed', true if client_id is @current_client().id
+    if client_id is @current_client().id
+      amplify.publish z.event.WebApp.SIGN_OUT, 'client_removed', true
```
bennycode: String constant for logout reasons would be nice. Because there is now `'user_requested'`, `'session_expired'` and `'client_removed'`.

gregor: Good point. Added.

PR #257, opened 2016-09-28
Insert line break on alt+enter

lipis:
The best I could find is the this answer: http://stackoverflow.com/a/1064139


app/script/util/util.coffee
```diff
@@ -653,3 +654,35 @@ z.util.bucket_values = (value, bucket_limits) ->
 
   last_limit = bucket_limits[bucket_limits.length - 1]
   return "#{last_limit+1}-"
+
+
+# http://stackoverflow.com/a/1064139
+z.util.insert_at_caret = (txtarea, text) ->
```
bennycode: Do we have `KeyUtils` or anything similar? If not, please create one, so that we stop dumping everything in `util.coffee`.

lipis: Done.. 

By they way.. this new review thingy is not removing the outdated comments :(

bennycode:
I updated your branch. Please merge if Travis is fine with it.

PR #259, opened 2016-09-29
'media.completed_media_action' on text message (WEBAPP-3227, WEBAPP-3233)

gregor:
- remove duplicate session event counters
- unify `media.completed_media_action` event sending

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -878,47 +878,39 @@ class z.conversation.ConversationRepository
       @conversation_service.update_message_in_db message_et, {data: asset_data, status: z.message.StatusType.SENT}
 
   ###
-  Send an encrypted knock.
+  Send an knock to specified conversation.
   @param conversation_et [z.entity.Conversation] Conversation to send knock in
   @return [Promise] Promise that resolves after sending the knock
   ###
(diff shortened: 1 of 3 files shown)


app/script/conversation/ConversationRepository.coffee
```diff
@@ -1089,9 +1075,23 @@ class z.conversation.ConversationRepository
...
 
       return saved_event
 
+  _track_completed_media_action: (conversation_et, generic_message) ->
+    action_type = switch generic_message.content
+      when 'asset' then 'asset' if generic_message.asset.original?
+      when 'knock' then 'ping'
+      when 'text' then 'text' if not generic_message.text.link_preview.length
+
+    return if not action_type
```
bennycode: Use a default `case` in your `switch` construct instead of `return if not action_type`.

gregor: I can't as the `switch` assigns the value of `action_type`. I cannot return in `else` as the default case.

bennycode: It would be possible like this:

``` coffeescript
action_type = undefined

switch generic_message.content
  ...
```

But ok.

gregor: Yes, but that is more code that I would consider to be less explicit.

PR #263, opened 2016-10-04
add http to href attribute

herrmannplatz:


app/script/util/util.coffee
```diff
@@ -282,14 +282,23 @@ z.util.add_blank_targets = (text_with_anchors) ->
...
 
 ###
+Adds http to given url if protocol missing
+
+@param url [String] URL you want to open in a new browser tab
+###
+z.util.add_http = (url) ->
+  if not url.match /^http[s]?:\/\//i
+    url = "http://#{url}"
+  return url
```
lipis, line 292: Or just return instead of a variable

``` coffeescript
if not url.match /^http[s]?:\/\//i
  return "http://#{url}"
return url
```

test/unit_tests/util/UtilSpec.coffee
```diff
@@ -880,8 +880,15 @@ describe 'z.util.safe_window_open', ->
...
 
-    new_window = window.open url
-    expect(new_window.location).not.toBe 'http://wire.com/'
+  it 'does not add http if present', ->
+    url = 'http://wire.com/'
+    expect(z.util.add_http url).toBe 'http://wire.com/'
+
+  it 'does not add https if present', ->
+    url = 'https://wire.com/'
+    expect(z.util.add_http url).toBe 'https://wire.com/'
```
lipis, line 894: can we remove the trailing slash from the URLs?

gregor: I think you have to fight for that one with design.

lipis: what design has to do anything with our tests?

gregor: It was in the spec that they want the trailing slash.

PR #264, opened 2016-10-04
enable video asset rendering for FF >= 49

herrmannplatz:


app/script/entity/message/Asset.coffee
```diff
@@ -77,7 +77,9 @@ class z.entity.Asset
   @return [Boolean] Is asset of type video
   ###
   is_video: ->
-    if @type is z.assets.AssetType.FILE and @file_type?.startsWith('video') and not z.util.Environment.browser.firefox
+    is_video_asset = @type is z.assets.AssetType.FILE and @file_type?.startsWith 'video'
+    is_unsupported_firefox = z.util.Environment.browser.firefox and platform.version < 49
+    if is_video_asset and not is_unsupported_firefox
```
lipis: it will be clearer to have the variable named `is_supported_firefox` instead of having double negatives..

so it will become `if is_video_asset and is_supported_firefox`

herrmannplatz: should be `is_supported_browser` then cause we only want to exclude specific FF's

herrmannplatz: done

app/script/entity/message/Asset.coffee
```diff
@@ -77,7 +77,9 @@ class z.entity.Asset
   @return [Boolean] Is asset of type video
   ###
   is_video: ->
-    if @type is z.assets.AssetType.FILE and @file_type?.startsWith('video') and not z.util.Environment.browser.firefox
+    is_video_asset = @type is z.assets.AssetType.FILE and @file_type?.startsWith 'video'
+    is_supported_browser = not (z.util.Environment.browser.firefox and platform.version < 49)
```
lipis, line 81: I don't want to be an ass.. but I think the inequality change is _clearer_ than again using the not :)

```
platform.version > 48
```

or

```
platform.version >= 49
```

herrmannplatz: but we want FF >= 49 and all other browser

lipis: aaaaaaaaaaaaaa :D

PR #312, opened 2016-10-14
expose conversation_list under old namespace

herrmannplatz:


herrmannplatz:
New context menu should fix this issue

PR #319, opened 2016-10-17
Read keys from local directory

lipis:


trans.py
```diff
@@ -21,10 +21,9 @@
 import sys
 
 home_dir = os.path.expanduser('~')
-user_config = os.path.join(home_dir, '.crowdin.yaml')
 
-os.system('crowdin-cli-py --identity={} upload sources'.format(user_config))
-os.system('crowdin-cli-py --identity={} download'.format(user_config))
```
gregor, line 27: Any specific reason why we move away from the python implementation? Now I have to rely on some weird Java stuff on Windows

PR #362, opened 2016-10-26
remove new Promise calls where possible

herrmannplatz:


app/script/storage/StorageRepository.coffee
```diff
@@ -104,17 +104,11 @@ class z.storage.StorageRepository extends cryptobox.CryptoboxStore
...
-        else
-          resolve undefined
-      .catch (error) =>
-        @logger.log @logger.levels.ERROR, "Something failed: #{error?.message}", error
-        reject error
+    return @storage_service.load @storage_service.OBJECT_STORE_KEYS, 'local_identity'
+    .then (identity_key_pair) ->
+      if identity_key_pair
+        bytes = sodium.from_base64 identity_key_pair.serialised
+        return Proteus.keys.IdentityKeyPair.deserialise bytes.buffer
```
gregor, line 111: We need to return undefined otherwise.

herrmannplatz: implicit return

PR #368, opened 2016-10-27
Add support for receiving v3 images (WEBAPP-3297)

herrmannplatz:


app/script/cryptography/CryptographyMapper.coffee
```diff
@@ -82,7 +82,9 @@ class z.cryptography.CryptographyMapper
         throw new z.cryptography.CryptographyError z.cryptography.CryptographyError::TYPE.UNHANDLED_TYPE
 
   _map_asset: (asset, event_nonce, event_id) ->
-    if asset.uploaded?
+    if asset.uploaded? and asset.uploaded.asset_id and asset.original?.image?
```
gregor: Same if check twice. Move in?

gregor:
```
if asset.uploaded?
  if asset.uploaded.asset_id and asset.original?.image?
    return @_map_image_asset_v3 asset, event_nonce
  return @_map_asset_uploaded asset.uploaded, event_id
else if asset.not_uploaded?
```

app/script/conversation/EventMapper.coffee
```diff
@@ -420,7 +420,9 @@ class z.conversation.EventMapper
     asset_et.original_width = data.data.info.original_width
     asset_et.original_height = data.data.info.original_height
     asset_et.ratio = asset_et.original_height / asset_et.original_width
-    if asset_et.id
+    if data.data.key
+      asset_et.resource z.assets.AssetRemoteData.v3 data.data.key, data.data.otr_key, data.data.sha256, data.data.token
```
bennycode: So much `data.data` in this line... Maybe put this first into a variable like:

``` coffee
asset = data.data
asset_et.resource z.assets.AssetRemoteData.v3 asset.key, asset.otr_key, asset.sha256, asset.token
```

Or `key = data.data`. WDYT?

herrmannplatz: jipp. will rename param to event

PR #377, opened 2016-10-31
include speakingurl

herrmannplatz:
remove speakingurl from webapp-module-external and make it a dependency in webapp. Created a PR on the speakingurl repo with the Omega fix we used to have in our repo. I also upgraded from 4.0.0 to 10.0.0. 

see https://github.com/pid/speakingurl/pull/86

### DIFF
--- a/test/unit_tests/util/UtilSpec.coffee
+++ b/test/unit_tests/util/UtilSpec.coffee
@@ -927,8 +927,8 @@ describe 'z.util.compare_names', ->
   it 'ΠΑΝΑΓΙΩΤΑ equals PANAGIOTA', ->
     expect(z.util.compare_names('ΠΑΝΑΓΙΩΤΑ', 'PANAGIOTA')).toBeTruthy()
 
-  it 'Björn equals Bjorn', ->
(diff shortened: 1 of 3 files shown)


test/unit_tests/util/UtilSpec.coffee
```diff
@@ -916,3 +916,19 @@ describe 'z.util.add_http', ->
   it 'does not add https if present', ->
     url = 'https://wire.com/'
     expect(z.util.add_http url).toBe 'https://wire.com/'
+
+describe 'z.util.compare_names', ->
+  it 'René equals Rene', ->
+    expect(z.util.compare_names('René', 'Rene')).toBeTruthy()
+
+  it 'Παναγιὡτα equals Panagiota', ->
+    expect(z.util.compare_names('Παναγιὡτα', 'Panagiota')).toBeTruthy()
```
lipis: Change it to: `ώ` instead of `ὡ` (in the description as well)

herrmannplatz: done

bennycode:
Still failing because of `Παναγιώτα`?

herrmannplatz:
Yes, but it should work now. Forgot to bundle speakingurl.

PR #380, opened 2016-10-31
Upgraded libs

lipis:


herrmannplatz:
https://greenkeeper.io/

lipis:
Greenkeeper is fun.. but in our case is more complicated.. kind of regret already that I started it :)

gregor:
Let's definitely add greenkeeper

lipis:
#382

PR #400, opened 2016-11-03
AV preferences: Translations and visual tweaks

gregor:


app/style/content/preferences/av.less
```diff
@@ -67,5 +69,5 @@
 }
 
 .preferences-av-video-disabled {
-  background-color: fade(@graphite, 48%)!important;
+  background-color: fade(@graphite, 48%);
```
herrmannplatz, line 72: ❤️

PR #414, opened 2016-11-04
cached processed text

herrmannplatz:
## Description
Cache already processed text (links, media embeds, markdown). This should increase performance when switching conversations and reduce memory allocation.

Possible Risks:
1. Text in the cache gets overwritten by a different message. Since the key to cache is a V4 UUID the possibility of having conflicts is not an issue.
2. User edits the text. Since the edited message is stored as a new message with a new nonce, it would create a new entry in the cache. The old one would be remove when new message are stored in the cache

## Screenshots
bildschirmfoto 2016-11-03 um 19 07 12

## Type of change
I used an lru cache to store the processed text. 

- [x] Bug fix

## Checklist:

- [x] My Pull Request contains test coverage for my changes.

### DIFF
--- a/app/script/conversation/EventMapper.coffee
+++ b/app/script/conversation/EventMapper.coffee
@@ -333,7 +333,7 @@ class z.conversation.EventMapper
   ###
   _map_asset_text: (data) ->
     asset_et = new z.entity.Text data.id
-    asset_et.key = "text##{data.nonce}"
+    asset_et.nonce = data.nonce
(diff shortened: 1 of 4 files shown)


herrmannplatz:
The cache variable is not in the global scope. Everything you declare in the coffeescript file is not exposed to the global scope unless you do it. Its only accessible within the text asset.

bildschirmfoto 2016-11-04 um 08 51 33

PR #415, opened 2016-11-04
Disabled download option for ephemeral content

bennyn:


app/script/view_model/MessageListViewModel.coffee
```diff
@@ -383,7 +383,7 @@ class z.ViewModel.MessageListViewModel
 
     @_track_context_menu message_et
 
-    if message_et.has_asset() and not message_et.is_expired()
+    if message_et.has_asset() and message_et.ephemeral_status() is z.message.EphemeralStatusType.NONE
```
gregor: Now the download option would be shown in other cases where we do not want it. I think `if message_et.has_asset() and not message_et.is_ephemeral()` is what we are aiming for.

gregor: We should also change https://github.com/wireapp/wire-webapp/blob/e57bb2d9278359dc3f029bd4856bb6c6031edbcb/app/script/entity/message/Message.coffee#L185-L188 to return false for messages where `is_ephemeral()` is `true`

bennycode: @gregor Can you please give an example in which case the download option is now shown where we don't want it?

PR #421, opened 2016-11-04
Clarify device removal details

infotexture:
- [x] Copy tweak


app/script/localization/strings.coffee
```diff
@@ -415,7 +415,7 @@ z.string.preferences_devices_fingerprint_detail = 'Wire gives every device a uni
 z.string.preferences_devices_id = 'ID: '
 z.string.preferences_devices_remove = 'Remove'
 z.string.preferences_devices_remove_cancel = 'Cancel'
-z.string.preferences_devices_remove_detail = 'Remove this device if you have stopped using it. Your message history will be erased on this device and you will be logged out.'
+z.string.preferences_devices_remove_detail = 'Remove this device if you have stopped using it. You will be logged out immediately.'
```
gregor: @infotexture This does not exactly indicate that you will be logget out of the device you removed and not the one you are removing it from. I know that is the only reading that would make sense but the wording is somewhat confusing IMHO.

gregor: @infotexture This does not exactly indicate that you will be logget out of the device you removed and not the one you are removing it from. I know that is the only reading that would make sense but the wording is somewhat confusing IMHO.

infotexture: Force-pushed to revise second sentence as follows:

> You will be logged out **of this device** immediately.

PR #424, opened 2016-11-04
Image sent via assets v3 API will report as photo (WEBAPP-3335)

gregor:


app/script/conversation/ConversationRepository.coffee
```diff
@@ -2341,7 +2341,9 @@ class z.conversation.ConversationRepository
       is_ephemeral = false
 
     action_type = switch message_content_type
-      when 'asset' then 'file' if message.asset.original?
+      when 'asset'
+        if message.asset.original?
+         if message.asset.original.image? then 'photo' else 'file'
```
lipis:
``` coffeescript
when 'asset' if message.asset.original?.image? then 'photo' else 'file'
```

gregor: No, otherwise message with asset metadata will be counted as files.

lipis: Then just fix the indentation.. :)

gregor: Done :)

PR #427, opened 2016-11-04
remove colored border

herrmannplatz:


app/style/components/user-avatar.less
```diff
@@ -94,7 +94,7 @@ user-avatar {
     }
 
     .user-avatar-border {
-      border-width: @avatar-border-md;
```
gregor, line 97: Also delete the variables

herrmannplatz: done

PR #429, opened 2016-11-04
Unify ephemeral timer

gregor:


app/script/components/ephemeralTimer.coffee
```diff
@@ -21,54 +21,32 @@ z.components ?= {}
...
+    message_et.ephemeral_remaining.subscribe (remaining_time) =>
+      if Date.now() >= ephemeral_expires
+        @progress 1
       else
-        @remaining_time scheduled_time - Date.now()
-        elapsed_time = timer - @remaining_time()
-        @progress elapsed_time / timer
-    , update_interval
+        elapsed_time = ephemeral_duration - remaining_time
+        @progress elapsed_time / ephemeral_duration
```
herrmannplatz: you should dispose the subscription in destroy function to avoid memory leaks

app/script/entity/message/Message.coffee
```diff
@@ -226,3 +231,33 @@ class z.entity.Message
     else if @status() isnt updated_status
       return @status updated_status
     return false
+
+  start_ephemeral_timer: =>
+    return if @ephemeral_timer
+    @ephemeral_remaining new Date(@ephemeral_expires()) - Date.now()
```
herrmannplatz, line 237: i am not a fan of moving the localization inside the message, since this is what a component is for

app/script/entity/message/Message.coffee
```diff
@@ -49,6 +53,7 @@ class z.entity.Message
       else
         return z.message.EphemeralStatusType.NONE
 
+    @conversation_id = ''
     @from = ''
```
herrmannplatz, line 57: 👍

app/script/components/ephemeralTimer.coffee
```diff
@@ -21,54 +21,32 @@ z.components ?= {}
...
 class z.components.EphemeralTimer
   constructor: (params) ->
-    scheduled_time = new Date(params.expires()).getTime()
-    starting_time = Date.now()
-    timer = scheduled_time - starting_time
-    update_interval = 1000 / 10
+    message_et = params.message
+    ephemeral_expires = new Date(message_et.ephemeral_expires()).getTime()
+    ephemeral_started = new Date(message_et.ephemeral_started()).getTime()
+
```
herrmannplatz, line 25: with introducing this value there might be the possibility that we load a message from the db that does not have a started time. This would result in `new Date('').getTime() = NaN`

app/script/components/ephemeralTimer.coffee
gregor: Will do

app/script/entity/message/Message.coffee
gregor: Where should I move it instead if we need the caption at multiple spots in the UI?

app/script/components/ephemeralTimer.coffee
gregor: Will check

app/script/components/ephemeralTimer.coffee
gregor: How do I best dispose the interval for the remaining calculation as it is in then message and not the timer.

app/script/components/ephemeralTimer.coffee
gregor: Code changes should accomodate for this: "500" - "300" is still 200.

app/script/entity/message/Message.coffee
gregor: Moved the localization out of the message.

app/script/components/ephemeralTimer.coffee
gregor: Done

app/script/entity/message/Message.coffee
gregor: @herrmannplatz What do you think about the new approach? Feel free to 
merge if you think this is ready.

PR #474, opened 2016-11-21
Scalable link previews

lipis:


app/style/components/asset/link-preview-asset.less
```diff
@@ -46,7 +46,6 @@ link-preview-asset {
   align-items: center;
   display: flex;
   justify-content: center;
-  width: 480px;
   height: 160px;
```
herrmannplatz, line 48: i think we should keep a max-width of 480px

PR #480, opened 2016-11-23
Cleanup: Last notification retrieval on app start

gregor:


app/script/event/EventRepository.coffee
```diff
@@ -190,20 +187,34 @@ class z.event.EventRepository
...
+  ###
+  get_last_notification_id_from_db: (last_notification_id) =>
+    @notification_service.get_last_notification_id_from_db()
+    .then (last_notification_id) =>
+      @last_notification_id last_notification_id
+
+      @last_notification_id.subscribe (last_notification_id) =>
+        @logger.log @logger.levels.INFO, "Last notification ID updated to '#{last_notification_id}'"
+        @notification_service.save_last_notification_id_to_db last_notification_id if last_notification_id
+
```
herrmannplatz: I think there is an issue here with the subscription. Whenever `get_last_notification_id_from_db` is executed it will add another subscription. Means when you executed `get_last_notification_id_from_db ` a second time and change `last_notification_id`, the handler will fire twice.

herrmannplatz: We should also try to remove the number of sideeffects

gregor: True. Will add a safeguard to make sure we only have a single subscription at all times, despite the fact that his is only called once. In case we reuse the use.

app/script/event/EventRepository.coffee
```diff
@@ -224,25 +235,22 @@ class z.event.EventRepository
...
+      if last_notification_id
+        @logger.log @logger.levels.INFO, "ID of last notification fetched from stream is '#{last_notification_id}'"
+      return @notifications_total
+    .catch (error) =>
+      @notification_handling_state z.event.NotificationHandlingState.WEB_SOCKET
+      if error.type in [z.event.EventError::TYPE.NO_LAST_ID, z.event.EventError::TYPE.NO_NOTIFICATIONS]
+        @find_ongoing_calls()
+        @logger.log @logger.levels.INFO, 'No notifications found for this user', error
+        return 0
+      else
```
lipis: `return/ else` not needed

PR #481, opened 2016-11-23
Improving loading of conversation events

bennyn:


app/script/conversation/ConversationService.coffee
```diff
@@ -281,6 +281,33 @@ class z.conversation.ConversationService
       throw error
 
   ###
+  Load conversation events. Start and end are not included. Events are always sorted beginning with the newest timestamp.
+
+  TODO: This function can be removed once Microsoft Edge's IndexedDB supports compound indices:
```
gregor, line 286: Apparently you also need it when an end is not defined. Therefore it's not really deprecated and I think we should change the name.

app/script/conversation/ConversationService.coffee
```diff
@@ -289,25 +316,23 @@ class z.conversation.ConversationService
...
   ###
-  load_events_from_db: (conversation_id, start, end, limit) ->
+  load_events_from_db: (conversation_id, start, end, limit = z.config.MESSAGES_FETCH_LIMIT) ->
+    if z.util.Environment.browser.edge or not end
+      return @_load_events_from_db_deprecated conversation_id, start, end, limit
+
+    include_minimum = true
+    if start > end
+      include_minimum = false
+      [start, end] = [end, start]
```
lipis, line 333:
```coffeescript
include_minimum = start < end
if not include_minimum
  [start, end] = [end, start]
```

bennycode: I like! Good idea!

PR #491, opened 2016-11-24
Prevent infinite reaction update loop

gregor:
Does not fix the underlying database update race condition problem the sequential update check was intended to solve. But we will no longer end up in a loop. Investigating to improve the solution.

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -1847,21 +1847,22 @@ class z.conversation.ConversationRepository
     @get_message_in_conversation_by_id conversation_et, event_json.data.message_id
     .then (message_et) =>
       changes = message_et.update_reactions event_json
-      if changes
-        @_update_user_ets message_et
-        @_send_reaction_notification conversation_et, message_et, event_json
-        @logger.log @logger.levels.DEBUG, "Updated reactions to message '#{event_json.data.message_id}' in conversation '#{conversation_et.id}'", event_json
-        return @conversation_service.update_message_in_db message_et, changes, conversation_et.id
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationRepository.coffee
```diff
@@ -1847,21 +1847,22 @@ class z.conversation.ConversationRepository
...
+            @_on_reaction conversation_et, event_json, attempt + 1
           , 10 * attempt
-        else
-          @logger.log @logger.levels.INFO, "Failed to update reaction of message in conversation '#{conversation_et.id}'", error
-      else if error.type isnt z.conversation.ConversationError::TYPE.MESSAGE_NOT_FOUND
-        @logger.log @logger.levels.INFO, "Failed to handle reaction to message in conversation '#{conversation_et.id}'", error
+        @logger.log @logger.levels.ERROR, "Too many attempts to update reaction to message '#{message_et.id}' in conversation '#{conversation_et.id}'", error
+    .catch (error) =>
+      if error.type isnt z.conversation.ConversationError::TYPE.MESSAGE_NOT_FOUND
+        @logger.log @logger.levels.ERROR, "Failed to handle reaction to message in conversation '#{conversation_et.id}'", error
```
bennycode, line 1865: I think it's also nice to log the message ID in this log statement:

```coffeescript
 @logger.log @logger.levels.ERROR, "Failed to handle reaction to message '#{message_et.id}' in conversation '#{conversation_et.id}'", error
```

gregor: Message entity is undefined because of closure at this point. We should never end up here. If we do we throw the error to log to Rayun.

PR #509, opened 2016-11-28
Ensure recovery after re-connect (WEBAPP-3396)

gregor:


app/script/service/Client.coffee
```diff
@@ -115,17 +112,19 @@ class z.service.Client
 
   # Execute queued requests.
   execute_request_queue: =>
+    return if not @access_token or not @request_queue.length
```
bennycode, line 115: Can `@request_queue` be defined without `length`?

gregor: No, it cannot. Request queue is initialized as an empty array and reset to this at the end of `execute_request_queue` only.

PR #528, opened 2016-12-05
Decoupling media handling from calling

gregor:


gregor:
Passed the calling matrix tests and is good to be merged after review

PR #533, opened 2016-12-07
Add ability to drop files (WEBAPP-3145/WEBAPP-3394)

herrmannplatz:


app/script/view_model/ConversationInputViewModel.coffee
```diff
@@ -132,11 +132,48 @@ class z.ViewModel.ConversationInputViewModel
...
+  on_drop_files: (dropped_files) =>
+    images = []
+    files = []
+
+    if @_is_hitting_upload_limit dropped_files
+      return
+
+    for file in dropped_files
+      switch
+        when file.type in ['.jpg-large', 'image/jpeg', 'image/png', 'image/bmp', 'image/gif']
```
bennycode: In our `config.coffee` we have `SUPPORTED_IMAGE_TYPES`. What about adding `.jpg-large` there and reusing `SUPPORTED_IMAGE_TYPES`? See: https://github.com/wireapp/wire-webapp/blob/dev/app/script/config.coffee#L77

herrmannplatz: 👍🏻

Issue #218, opened 2016-09-16
wire-webapp UI should be able to properly handle different screen resolutions/form_factors


fdelapuente:
Hi,

This enhancement request is related to #217 and is a dependency to fix that one.

The wire-webapp should adapt properly to different screen resolutions and form factors.

As an example, in narrow screens (Like a Phone in portrait position) the buttons for Video/Audio call get easily out of the screen and the user must manually workaround the UI in order to get there.

Thanks!

gregor:
That's why we block those clients from accessing the web app. We do have native applications for Android and iOS and want to reduce the overhead created by maintaining a responsive design in our web application. We might reconsider this in the future but for now our UI is built and designed on the promise that we can always depend on a certain minimum width of the display.

fdelapuente:
For the record I added a new comment on #217 proposing a change in the condition to reject the connection from a non-supported browser.

The problem with blocking by user agent is that you end up blocking a family of browsers under the assumption that all the clients running this browser will not have enough screen space to properly use the wire-webapp, while being unaware of actually detecting the real problem which is the actual space available in the browser viewport.

I can mention the following undesired outcomes of this approach:
1. A Desktop browser connecting to the webapp with a small window size will not be filtered out and still will be facing the UI design issues.
2. A Mobile browser will be rejected from the webapp even if it has enough space in its browser viewport.

Best Regards.

lipis:
Web is a wild wild place indeed.. and we are officially supporting the [4 _major_ browsers](https://wire.com/download/) on Desktops..

max-baz:
> We might reconsider this in the future but for now our UI is built and designed on the promise that we can always depend on a certain minimum width of the display.

@gregor, @lipis: I want to add another reason to reconsider this. Given that desktop apps in reality just load this same webapp, you have to consider how they are used when taking such decisions.

I can see that right now the minimal supported width is around 760px, below which the layout starts to break. However people on desktop rarely spend their time _exclusively_ in the chatting apps, instead most people do something, and on a side they have a chat app open, often resized to take as less space as possible.

Have a look at 3 screenshots below that demonstrate the case, where I'm reading github and having a chat app on a side. The 3 pictures show:
1. How Telegram app is able to take advantage of a narrow window size (notice: I use most of the screen space for my primary activity, but can occasionally keep chatting on the right). 
2. How Wire app looks like given the same window size (notice: chat is not even visible).
[...]

PR #543, opened 2016-12-09
POC #218: Hide contacts on a narrow screen

maximbaz:
## Description
This is a proof of concept of a simple idea that originates from the discussion in #218.

We all know that today you don't officially support screen width less than 768px, but we also know real use-cases when having narrow screen is useful. And we know that today the Wire is entirely unusable on narrow screens.

What I want to achieve with this pull request:
- To make Wire _usable_ for chatting with one person/group at a time, instead of having it entirely unusable (switching between conversations is possible with hotkeys `alt + shift + ↑/↓`). 
- To show that the code doesn't have to be changed _a lot_ to support narrow screens, the layout is already pretty fluid.
- To gather the feedback on this, e.g. what could be done as a next incremental improvement.
- To see if you would theoretically accept to merge such proposals :wink:.

From what I can see, the layout with width >= 768px is not affected with changes made in this pull request. The experience with chatting with one person is much better (see screenshots below).

Obviously you shouldn't declare that Wire officially supports narrow screens right after merging this pull request, but you make a better experience for those who still want it even without an official support.

## Screenshots
Before:

After:

## Future ideas
A natural next improvement would be to add an ability to choose between conversations via UI in addition to keyboard shortcuts. What I think should be relatively easy to do:
- Add an icon `<` to the left of the currently opened contact name (inside the `.conversation-titlebar`).
- Clicking on it would show the left panel with the contact list (it should overlay the conversation pane).
- Clicking on any contact in the left panel would in addition hide the left panel again.

The changes should obviously only affect the narrow layout.
What do you think?


bennycode:
@helenakos Can you please show that to our design team? I am a big fan of this pull request. 👍

bennycode:
Hi @maximbaz,

Your Pull Request brings us tough times here at [Wire](https://wire.com/). 😁 I really like the idea of having more space for a conversation when chatting in a narrow window. It's useful! But it also puts some interaction questions on the table... You have already noticed, that it needs an element in the UI to go back to the conversation list because otherwise you can easily lose the overview. So the future ideas you brought up, become a requirement once we merge your PR. This in turn requires a responsiveness of our conversation list component and if we start there, then also our settings page needs to be responsive. 

As you can see, there is a domino effect, which requires solving individual tasks, which are currently not on our roadmap. That's why we will keep your PR **on hold** until we have the time and resources to think about it in an overall responsive way. 

Thanks for bringing this topic up. And 👍 for your proof of concept!

For all readers, here is a screenshot of the changes from this pull request (left) and our status quo (right):

screenshot

max-baz:
Hey @bennyn, thanks for the detailed response!

I'm really glad to hear that you guys consider user's feedback seriously.
I just have one follow-up question before letting this go :)

My observation is that on as of today the application is rather limited when communicating even with one person, below are the three screenshots:

**728px**: Ping functionality is not available.
image

**660px**: Attachment functionality is not available.
selection_218

**600px**: Image attachment functionality is not available, the conversation is not entirely visible.
selection_219

What I'm trying to say is that perhaps this pull request shouldn't be _binding_, you shouldn't feel obliged to keep improving right now. You should keep saying that officially the minimal supported width is 768px, with or without merging this pull request. The only thing to do is to consider what's better out of two "not ideal" solutions.

The choice to make is: given that some people still want to use not-officially-supported small screens, what's better: 
- having all functionality available but potentially missing an incoming message in another conversation (though notifications or the tray icon on desktop would indicate this)
- or having a good overview of all conversations while not being able to attach files or even fully see the text in the conversation.

Personally I'd prefer the first, but I'm just as open for the feedback :)

priiduzilmer:
Hi @maximbaz,

I am responsible for the design (team) at Wire. @bennyn asked me to provide a bit of background to this, so here goes…

First and foremost, thank you for taking the time to explain the issue. It really is awesome to have a piece of software out there and have smart people thinking along with suggestions about how to make it better.

The issue with List (it is what we call the listing of conversations) taking up too much space is a well known one for us at Wire Design. The List we are using today is a combination of many different variants that we have had for it over a long period of time. Back in the day we actually even had a Wire desktop app along the lines of what you suggest — a sliding List to replace the conversation area.

The problem with it that @bennyn points at though, is the resource that needs to go into this. The current solution is the most optimal we found back in the day to satisfy all the major use cases. You can rest assured that we are not super happy with it either… but we plan to fix it properly with a List overhaul soon. This means we do not want to do any intermediate solutions where it slides out etc. We just want to create a properly scalable single solution for different sizes.

Like @bennyn is saying – having multiple modes for the window is *extremely* expensive both in terms of design and development. So we are currently looking at a way to have a unified design language for mobile and desktop and implement it properly for both as soon as possible next year. All the concerns regarding window size and content visibility/cropping that you have brought up in this issue will be taken into account, that’s for sure.

I hope this explains a bit about where we are at and what we are aiming at. Please let us know if you have any further suggestions — we really appreciate all the feedback.

max-baz:
Thank you both @priiduzilmer and @bennyn, I really appreciate such detailed explanations! 

I did not realize that you already had several different variants of the List, and it makes sense that you don't want to spend time on merging and maintaining any intermediate solutions. I'm looking forward to see the redesign of the page once you guys get to it! :+1: 

I think if you want you can close this pull request, as it won't be merged. The discussion here is interesting and informative, and it will still be possible to link to this page and keep commenting if other suggestions appear, but it won't spam your list of open pull requests that are awaiting some action.

bennycode:
Thank you for your great cooperation, @maximbaz. We will definitely look at your changes as soon as we decide on a fully responsive design. 👍 

Because of your energetic support, I've had a hard time closing this pull request from you. But since a few minutes you are an official Wire contributor (https://github.com/wireapp/wire-webapp/pull/595). 🎉 

So I think we can close this PR here since it is not your only signature anymore. 😄 

Many thanks and happy holidays!

max-baz:
Thanks for the kind words @bennyn, and have a happy holidays too! :tada:

vegai:
I hope something will be done about this. The UX is a bit shameful on a desktop currently :-/

PR #570, opened 2016-12-14
🚀 feature copy image from clipboard

herrmannplatz:
## Type of change
Adds the ability to paste images from OS clipboard

- [x] test Chrome
- [x] test FF
- [ ] test Edge

## Screenshot
bildschirmfoto 2016-12-14 um 11 05 14


app/script/view_model/ConversationInputViewModel.coffee
```diff
@@ -27,15 +27,31 @@ class z.ViewModel.ConversationInputViewModel
     @conversation_et = @conversation_repository.active_conversation
     @conversation_et.subscribe =>
       @conversation_has_focus true
+      @pasted_file null
```
bennycode, line 30: Why do you prefer `null` over `undefined`?

herrmannplatz: not really in this case

app/style/fonts/zeta-neue.css
```diff
@@ -237,6 +237,9 @@
 .icon-video-off:before {
   content: "\e231";
 }
+.icon-Send:before {
```
gregor: can we not have this one be the only icon with a upper case name?

PR #572, opened 2016-12-14
🖼 cache image assets

herrmannplatz:
## Type of change
Cache assets remote data using service worker.


app/script/assets/AssetService.coffee
```diff
@@ -192,6 +192,7 @@ class z.assets.AssetService
   generate_asset_url: (asset_id, conversation_id) ->
```
gregor: You forgot to use the parameter

PR #588, opened 2016-12-18
Show user picture in notifications, fix Wire icon on desktops

maximbaz:
## Description
I'd like to suggest to show sender's picture in a notification instead of the Wire icon.

The following logic is preserved:
- Do not show sender's photo when `Notifications: hide message details` is set.
- Do not show any image on Electron for MacOS (though don't know why not).
- If an image URL cannot be generated for some reason, fallback to the Wire icon.

-----

I also discovered another issue: the wire icon was not showing in notifications on desktop (Linux only?), and a standard icon was used instead (light bulb in the screenshots below). 

The reason was this piece of code: `window.notification_icon or '/image/logo/notification.png'`. The variable `window.notification_icon` did have a value set to a correct local file path on my machine, but for whatever reasons it could not have been used. When I switched to always use `'/image/logo/notification.png'`, the icon started to appear on both desktop and web. 

------

/cc @gregor since I saw you worked on notifications earlier.

## Screenshots

Chrome, before:

image

image

Chrome, after:

image

image

Linux desktop, before:

image

image

Linux desktop, after:

image

image


someoneEsle:
This would be an awesome thing to have! The only thing I'd ask for (hoping the necessary changes arent too complicated) is to make the pictures round instead of square.
Edit: actually, never mind. It looks like they were square by design (I use Linux so I had no pictures at all 😂)

max-baz:
@someoneEsle I'm afraid we cannot style the images inside the notifications. If only _the image itself_ was round (with transparent background), that would have worked.... But I only found two ways to retrieve a picture of a user, both work by generating URLs that returns images in high and medium resolutions, and both are square.

That being said, I believe it was a deliberate decision to use square pictures for the notifications. There is a [folder with circle images](https://github.com/wireapp/wire-webapp/tree/dev/app/image/icon) for the Wire logo, but the app is not using them, but is using a different, square logo.

max-baz:
> I use Linux so I had no pictures at all :joy:

Exactly! But now we will :wink: 

I also tried a couple of other notification managers:

xfce4-notifyd, before:
image

xfce4-notifyd, after:
image

notify-osd, before:
image

notify-osd, after:
image

bennycode:
@mythsunwind, @Yserz, @lipis:  Do you guys have time to test this? 🕋

lipis:
I can do it tomorrow! It's a cool suggestion.. we'll have to run it with our designers, but looks very promising..

lipis:
The reason that we disabled the Electron/macOS icon is because it was showing two icons instead of one and we couldn't disable the main one.

lipis:
#### Testing

- [x] macOS 
  - [x] Desktop (as expected - no user's avatar to avoid double image)
  - [x] Chrome
  - [x] Firefox
- [x] Linux 
  - [x] Desktop
  - [x] Chrome
  - [x] Firefox
- [ ] Windows
  - [ ] Desktop
  - [x] Chrome
  - [x] Firefox
  - [x] Edge
- [x] Ephemeral messages
- [x] Disable contents

app/script/system_notification/SystemNotificationRepository.coffee
```diff
@@ -346,12 +346,22 @@ class z.SystemNotification.SystemNotificationRepository
 
     should_obfuscate = @_should_obfuscate_notification message_et
 
+    if z.util.Environment.electron and z.util.Environment.os.mac
+      icon = ''
+    else
+      try
+        icon = unless should_obfuscate then message_et.user().preview_picture_resource().generate_url()
```
max-baz: a reminder to myself: change to `should_obfuscate_sender` once #595 gets merged.

lipis: we might merge this one first :)

.gitignore
```diff
@@ -19,6 +19,7 @@
 /keys
 /nbproject
 /node_modules
+/tags
```
lipis, line 22: Can we remove that.. or you need it for something?

max-baz: It's a Ctags file, so that text editors like vim can understand syntax of the language and provide features like "go to definition".

I'd like to keep it if it doesn't bother you too much, otherwise I'll just remember to remove "/tags" file every time before pushing the code.

lipis: no no.. it's fine :) Thanks for clarifying it..

app/script/system_notification/SystemNotificationRepository.coffee
gregor: `icon = message_et.user().preview_picture_resource().generate_url() unless should_obfuscate`

gregor:
Same for this one. Will review with design tomorrow and try to get it merged before Christmas. Thank you so much for your contribution.

gregor:
Seeing some issues on Windows related to the ServiceWorker but will go ahead and merge it.

lipis:
👍

PR #589, opened 2016-12-19
Align download links

infotexture:
The ’get.wire.com’ variant is used in more places and offers platform-specific redirects that take users directly to the app stores on supported platforms.


app/script/localization/strings.coffee
```diff
@@ -302,7 +302,7 @@ z.string.invite_meta_key_pc = 'Ctrl'
 z.string.invite_hint_selected = 'Press %meta_key + C to copy'
 z.string.invite_hint_unselected = 'Select and Press %meta_key + C'
 z.string.invite_headline = 'Invite people to Wire'
-z.string.invite_message = 'I’m on Wire, search for %username or visit wire.com/download.'
+z.string.invite_message = 'I’m on Wire, search for %username or visit get.wire.com.'
```
gregor, line 305: @infotexture The string just one line below contains https://get.wire.com -> are we going with or without protocol?

infotexture: We prefer not to include protocol information in the link text, except in cases where it's necessary to ensure that the text is recognized as a hyperlink.

If [get.wire.com](https://get.wire.com) is hyperlinked without the protocol in this context, it should be removed.

infotexture: @gregor I'll push an extra commit to remove the protocol from the remaining instance.

PR #597, opened 2016-12-22
Bots (WEBAPP-2077)

gregor:
- URLs are backend environment depending (bot and announcements API, password reset link)
- added the ability to remove a bot from a conversation

To remove a bot you currently need a little workaround, we have no meaning of identifying a user as a bot in a conversation right now. If you want to remove a bot you have to set such flag manually. In a 1to1 conversation you can simply do so with the following command from the console while in that conversation:

`wire.app.repository.conversation.active_conversation().participating_user_ets()[0].is_bot(true)`

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -625,7 +625,7 @@ class z.conversation.ConversationRepository
   ###
   Remove bot from conversation.
   @param conversation_et [z.entity.Conversation] Conversation to remove member from
-  @param user_id [String] ID of bot to be removed from the the conversation
+  @param user_id [String] ID of bot to be removed from the conversation
   ###
   remove_bot: (conversation_et, user_id) =>
     @conversation_service.delete_bots conversation_et.id, user_id
@@ -637,7 +637,7 @@ class z.conversation.ConversationRepository
   ###
   Remove member from conversation.
   @param conversation_et [z.entity.Conversation] Conversation to remove member from
-  @param user_id [String] ID of member to be removed from the the conversation
+  @param user_id [String] ID of member to be removed from the conversation
   ###
   remove_member: (conversation_et, user_id) =>
     @conversation_service.delete_members conversation_et.id, user_id
@@ -645,6 +645,16 @@ class z.conversation.ConversationRepository
       if response
         amplify.publish z.event.WebApp.EVENT.INJECT, response
         return response
+
+  ###
+  Remove participant from conversation.
+  @param conversation_et [z.entity.Conversation] Conversation to remove participant from
+  @param user_et [z.entity.User] User to be removed from the conversation
+  ###
+  remove_participant: (conversation_et, user_et) =>
+    if user_et.is_bot()
+      return @conversation_repository.remove_bot @conversation(), user_et.id
+    return @conversation_repository.remove_member @conversation(), user_et
 
   ###
   Rename conversation.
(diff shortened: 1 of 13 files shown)


app/script/entity/User.coffee
```diff
@@ -49,6 +49,7 @@ class z.entity.User
   ###
   constructor: (@id = '') ->
     @is_me = false
+    @is_bot = ko.observable false
```
bennycode: Does this need to be an observable? This would mean that a user can turn into a bot or vice versa!?

app/script/conversation/ConversationRepository.coffee
```diff
@@ -621,23 +621,30 @@ class z.conversation.ConversationRepository
...
   ###
-  remove_member: (conversation_et, user_id, callback) =>
+  remove_member: (conversation_et, user_id) =>
     @conversation_service.delete_members conversation_et.id, user_id
     .then (response) ->
-      amplify.publish z.event.WebApp.EVENT.INJECT, response
-      callback?()
-    .catch (error) =>
-      @logger.error "Failed to remove member from conversation (#{conversation_et.id}): #{error}"
+      if response
```
bennycode, line 645: What happens if there is no response?

app/script/conversation/ConversationRepository.coffee
```diff
@@ -621,23 +621,30 @@ class z.conversation.ConversationRepository
...
   ###
-  Remove member from conversation.
+  Remove bot from conversation.
+  @param conversation_et [z.entity.Conversation] Conversation to remove member from
+  @param user_id [String] ID of bot to be removed from the the conversation
+  ###
+  remove_bot: (conversation_et, user_id) =>
+    @conversation_service.delete_bots conversation_et.id, user_id
+    .then (response) ->
+      if response
```
bennycode, line 633: What states do we have where there is no `response`?

app/script/conversation/ConversationRepository.coffee
gregor: Nothing. It means the user was not part of the conversation and we should not inject undefined. Tried that. Everything breaks.

app/script/conversation/ConversationService.coffee
```diff
@@ -88,6 +88,18 @@ class z.conversation.ConversationService
   ###############################################################################
 
   ###
+  Remove bot from conversation.
+
+  @param conversation_id [String] ID of conversation to remove bot from
+  @param user_id [String] ID of bot to be removed from the the conversation
+  @param callback [Function] Function to be called on server return
+  ###
+  delete_bots: (conversation_id, user_id) ->
```
bennycode, line 97: Is it really `delete_bots` (Plural)? Because `user_id` can be just a String and not an Array.

app/script/conversation/ConversationRepository.coffee
gregor: Currently there is no response ever from the backend. I requested to get the expected `conversation.member-leave` event in the 200 reponse just like we do on the delete member endpoint.

app/script/entity/User.coffee
gregor: Good point. Wanted it to be a computed first but haven't figured out a way to detect this.

app/script/conversation/ConversationService.coffee
gregor: Using the naming of the backend

gregor: It's a delete call on the bots endpoint. All our services are using this naming scheme.

PR #623, opened 2017-01-12
Properly filter search results based on usernames (WEBAPP-1694, WEBAPP-3462)

gregor:
## Type of change
Since the usernames launch our search results are not well filtered.

If a search query starts with an @, we will now return the following:
- connected users if their username starts with the search query
- groups if there is a member of that group whos username starts with the search query
- others if their username starts with that search query

Email matches are removed as the email address will no longer provided by the backend for any other user than yourself. Search resuls are also properly prioritized. Matches starting with the query come first, all other entries are alphabetically ordered as expected.

This also works in the filter to add people to a group

### Old search results
old_resultsold_results2

### New search results
new_resultsnew_results2

### DIFF
--- a/app/script/conversation/ConversationRepository.coffee
+++ b/app/script/conversation/ConversationRepository.coffee
@@ -324,14 +324,16 @@ class z.conversation.ConversationRepository
       .filter (conversation_et) ->
         return false if not conversation_et.is_group()
         if is_username
-          return true if z.util.compare_names conversation_et.display_name(), "@#{query}"
-          return true for user_et in conversation_et.participating_user_ets() when z.util.name_starts_with user_et.username(), query
+          return true if z.util.StringUtil.compare_transliteration conversation_et.display_name(), "@#{query}"
+          return true for user_et in conversation_et.participating_user_ets() when z.util.StringUtil.starts_with user_et.username(), query
         else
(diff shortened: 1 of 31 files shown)


app/script/conversation/ConversationRepository.coffee
```diff
@@ -315,24 +315,33 @@ class z.conversation.ConversationRepository
...
+          return true if z.util.StringUtil.compare_transliteration conversation_et.display_name(), "@#{query}"
+          return true for user_et in conversation_et.participating_user_ets() when z.util.StringUtil.starts_with user_et.username(), query
+        else
+          return true if z.util.StringUtil.compare_transliteration conversation_et.display_name(), query
+          return true for user_et in conversation_et.participating_user_ets() when z.util.StringUtil.compare_transliteration user_et.name(), query
+        return false
+      .sort (conversation_a, conversation_b) ->
+        if is_username
+          return z.util.StringUtil.sort_by_priority conversation_a.display_name(), conversation_b.display_name(), "#{@query}"
+        return z.util.StringUtil.sort_by_priority conversation_a.display_name(), conversation_b.display_name(), query
```
herrmannplatz, line 336: no need to repeat the whole line again. just change the param based on `is_username`

Issue #614, opened 2017-01-10
Password manager incompatibility & clipboard risk


fortran77:
The web app login page at https://app.wire.com/auth/#login appears to be incompatible with the Lastpass password manager used as a Google Chrome extension (even when I use the Lastpass feature that lets me manually initialize all possible password-related form fields). The email field gets filled in but the password field remains empty.

A minor consequence of this is the slight inconvenience of using manual copy-and-paste to fill in the password.

The more important risk here is that one's password ends up in the clipboard, where it might lurk unnoticed for a long time.

Lastpass is generally very good at identifying login and password fields, so if it doesn't work, it usually means there is something quite unusual about the web page.

It would be nice if the login page could be made more compatible with password managers. Perhaps an additional simple login page, with very few form fields and the least possible JavaScript, could be provided for use with password managers — if this can be done without compromising login security.

In the mean time, as a security precaution, the web app could simply clear the clipboard after every login.

ffflorian:
@fortran77 

> as a security precaution, the web app could simply clear the clipboard after every login

I am against this, because I don't want any web app to just clear my clipboard. The password manager is responsible for that.


> It would be nice if the login page could be made more compatible with password managers.

I just tried autofilling the login page with [KeePass2](http://keepass.info/) + [chromeIPass](https://chrome.google.com/webstore/detail/chromeipass/ompiailgknfdndiefoaoiligalphfdae) + Chromium. Works as intended.
Did you try the [trouble guide](https://lastpass.com/support.php?cmd=showfaq&id=455) in the LastPass FAQ? If it still doesn't work, we need to figure out what's the difference between the Wire WebApp login page and other login pages.

ffflorian:
Alright, after quickly creating a LastPass account, I figured it out. in [AuthViewModel.coffee#L1334](https://github.com/wireapp/wire-webapp/blob/prod/app/script/view_model/AuthViewModel.coffee#L1334) there is a fix to disable Chrome's auto form fill. This causes LastPass to ignore the password field.

As a temporary workaround you can right-click on the password field, select Auto-Fill and then click on the corresponding entry.

fortran77:
So it looks like this is fixable within the Wire web app? That's great.

The only useful part of the Lastpass trouble guide is the suggestion to use the Save All Entered Data feature, which I did try, hence my comment “even when I use the Lastpass feature that lets me manually initialize all possible password-related form fields”.

[...]

PR #631, opened 2017-01-16
Enabled WebKit prefill (WEBAPP-3210)

### LINKED ISSUE #198: Email autocompletion
Form autofilling doesn't work on the account creation form in Chrome.

ffflorian:
## Type of change
Fixes #198 and #614.


app/style/auth/account.less
```diff
@@ -131,4 +131,16 @@
       }
     }
   }
+
+  #wire-login-username,
```
gregor: What about the input fields on the register page? They have different IDs.

ffflorian: These fields are not auto completed - see [auth.html](https://github.com/wireapp/wire-webapp/blob/dev/app/page/auth.html#L136).
You are right, my PR actually only fixes #614, not #198.

PR #634, opened 2017-01-17
Revert "Enabled WebKit prefill (WEBAPP-3210)"

bennyn:
Reverts wireapp/wire-webapp#631

### DIFF
--- a/app/page/auth.html
+++ b/app/page/auth.html
@@ -133,17 +133,17 @@
 
           <form id="form-account-register" class="form" data-bind="submit: register">
             <div id="wire-register-name-container" class="form-control">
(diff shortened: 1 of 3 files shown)


gregor:
avs

lipis:
Maybe `color: #000;` instead of revert?

bennycode:
@lipis: But then the background would still be yellow. 🦅

gregor:
Background is the issue. White font color is correct. PR was intended to clear that.

lipis:
opa.. yes.. my bad!

bennycode:
There is another bug with... It prevents me from logging in:

capture

PR #648, opened 2017-01-19
Asynchronous address book

lipis:
- https://github.com/wireapp/node-addressbook/pull/3
- https://github.com/wireapp/wire-desktop/pull/332

### DIFF
--- a/app/script/connect/ConnectRepository.coffee
+++ b/app/script/connect/ConnectRepository.coffee
@@ -121,8 +121,8 @@ class z.connect.ConnectRepository
   ###
   _parse_macos_contacts: ->
     return new Promise (resolve, reject) =>
-      return resolve undefined if not window.zAddressBook
-      address_book = window.zAddressBook
+      return resolve undefined if not window.wAddressBook
+      address_book = window.wAddressBook
       phone_book = new z.connect.PhoneBook @properties_repository.self()
 
       me = address_book.getMe()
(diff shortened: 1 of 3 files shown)


app/script/connect/ConnectRepository.coffee
```diff
@@ -121,33 +120,33 @@ class z.connect.ConnectRepository
   @return [z.connect.PhoneBook] Encoded phone book data
   ###
   _parse_macos_contacts: ->
-    return if not window.zAddressBook
-
-    address_book = window.zAddressBook()
-    phone_book = new z.connect.PhoneBook @properties_repository.self()
+    return new Promise (resolve, reject) =>
+      return resolve undefined if not window.zAddressBook
+      address_book = window.zAddressBook
```
herrmannplatz: keep in mind that this breaks older clients

lipis: right.. shall we rename the `zAddressBook` to `wAddressBook` maybe? That should work.. and also maybe removing the option to import from AddressBook if that module is missing..

gregor: Do we want to shim the old one in webapp for a bit of time maybe? Version update on macOS takes a while.

herrmannplatz: @lipis old mac clients will still expose zAddressbook as a Function, so you should take that into account by checking for the new and the old syntax.

PR #650, opened 2017-01-19
🚧 collections #5

herrmannplatz:
## Type of change
separate items in collection by date

### DIFF
--- a/app/script/view_model/content/CollectionDetailsViewModel.coffee
+++ b/app/script/view_model/content/CollectionDetailsViewModel.coffee
@@ -53,20 +53,20 @@ class z.ViewModel.content.CollectionDetailsViewModel
       return true
 
     # we passed today
-    if not moment(message_et.timestamp).isSame(@last_message_timestamp, 'd') and moment(@last_message_timestamp).isSame new Date(), 'd'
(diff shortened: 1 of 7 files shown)


app/script/view_model/content/CollectionDetailsViewModel.coffee
```diff
@@ -35,30 +35,38 @@ class z.ViewModel.content.CollectionDetailsViewModel
...
+    # we passed the month
+    if not moment(message_et.timestamp).isSame(@last_message_timestamp, 'M')
+      @last_message_timestamp = message_et.timestamp
+      return true
 
-      if src.length is 0
-        window.clearInterval interval
+  get_title_for_header: (message_et) =>
+    message_date = moment(message_et.timestamp)
+    if message_date.isSame new Date(), 'd' # today
```
bennycode: I think a function for this would be nice and could avoid a comment:
```coffee
if _is_today message_date
  ...
```

Same applies to `# same year`. Could be also worth it to create a `DateUtil` for future usage.

bennycode:
Moment.js Plugin extension: http://momentjs.com/docs/#/plugins/

PR #652, opened 2017-01-20
Respecting ephemeral expiring status in data-uie-attribute

Yserz:
## Type of change
Respecting ephemeral expired status for data-uie-visible attribute within images

### DIFF
--- a/app/page/template/partials/template-message.htm
+++ b/app/page/template/partials/template-message.htm
@@ -64,7 +64,7 @@
       <!-- ko if: asset.is_image() -->
         <div class="message-asset-image">
           <div class="image image-loading" data-bind="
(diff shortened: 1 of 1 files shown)


app/page/template/partials/template-message.htm
```diff
@@ -64,7 +64,7 @@
       <!-- ko if: asset.is_image() -->
         <div class="message-asset-image">
           <div class="image image-loading" data-bind="
-            attr: {'data-uie-visible': $parent.visible},
+            attr: {'data-uie-visible': $parent.visible && !message.is_expired()},
```
gregor: My hunch is that is must be `attr: {'data-uie-visible': $parent.visible() && !message.is_expired()},` to get the correct state if you add the second check as both are observables

PR #653, opened 2017-01-20
Add accent color to sender name in message header (WEBAPP-3481)

gregor:
PR goes to branch `dev_next` as all features for the release are on `dev` locking it essentially.

### DIFF
--- a/app/style/content/conversation/message-list.less
+++ b/app/style/content/conversation/message-list.less
@@ -88,6 +88,18 @@
   &.accent-color-6 {color: @accent_colors-6}
   &.accent-color-7 {color: @accent_colors-7}
 }
+.message-header-label-sender {
+  font-weight: @font-weight-bold;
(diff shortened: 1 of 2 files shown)


app/style/content/conversation/message-list.less
```diff
@@ -88,6 +88,18 @@
   &.accent-color-6 {color: @accent_colors-6}
   &.accent-color-7 {color: @accent_colors-7}
 }
+.message-header-label-sender {
+  font-weight: @font-weight-bold;
+  text-transform: none;
```
herrmannplatz, line 93: think it should be capitalized or?

markesaaremets: Nope

gregor: Confirmed with @markesaaremets that is should not be uppercase and not capitalizing the first char either.

PR #674, opened 2017-01-24
© 2016

bennyn:

### DIFF
--- a/app/page/index.html
+++ b/app/page/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<!-- Wire, Copyright (C) Wire Swiss GmbH -->
+<!-- Wire, Copyright (C) 2016 Wire Swiss GmbH -->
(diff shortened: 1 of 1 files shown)


app/page/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<!-- Wire, Copyright (C) Wire Swiss GmbH -->
+<!-- Wire, Copyright (C) 2016 Wire Swiss GmbH -->
```
gregor, line 2: This is a different case though. File creation is the moment of serving it to the user. Thus it should be 2017. It is not our preamble in the git.

bennyn: @raphaelrobert Looks like we need clarification here. :D We also have (C) 2016 in all other HTML files, like https://github.com/wireapp/wire-webapp/blob/dev/aws/templates/aws/base_error.html

PR #680, opened 2017-01-24
new device message

herrmannplatz:


app/script/conversation/ConversationRepository.coffee
```diff
@@ -222,7 +221,7 @@ class z.conversation.ConversationRepository
   ###
   Get messages for given category. Category param acts as lower bound
   @param conversation_id [String]
-  @param catogory [z.message.MessageCategory.NONE]
+  @param category [z.message.MessageCategory.NONE]
   @return [Promise] Array of z.entity.Message entities
   ###
   get_events_for_category: (conversation_et, catogory = z.message.MessageCategory.NONE) =>
```
gregor, line 227: Same typo as in docs: catogory -> category

app/script/conversation/ConversationRepository.coffee
```diff
@@ -2243,6 +2236,8 @@ class z.conversation.ConversationRepository
         delete payload.recipients[user_id]
 
     return Promise.all @_map_user_client_map user_client_map, _remove_deleted_client, _remove_deleted_user
+    .then =>
+      @verification_state_handler.on_client_removed Object.keys(user_client_map)
     .then ->
```
gregor: Do we have to wait for the handler or can you just return the payload right away?

app/script/conversation/ConversationRepository.coffee
```diff
@@ -2271,6 +2266,8 @@ class z.conversation.ConversationRepository
         return @user_repository.add_client_to_user user_id, new z.client.Client {id: client_id}
 
       return Promise.all @_map_user_client_map user_client_map, _add_missing_client
+    .then =>
+      @verification_state_handler.on_client_add Object.keys(user_client_map)
     .then ->
```
gregor: Same as above - do we need another tick?

app/script/conversation/EventMapper.coffee
```diff
@@ -130,6 +147,7 @@ class z.conversation.EventMapper
   _map_all_verified: ->
     message_et = new z.entity.Message()
```
gregor: We could probably use fitting message entity for this as well and handle all e2ee related messages in a single entity type. z.entity.DegradedMessage -> z.entity.E2EMessage and degraded_message_type -> 'e2ee_message_type'

herrmannplatz: Yeah might be worth putting them together. Wanted to avoid to have all these cases in one message. But since verified is the only missing case we could totally do that. dont think e2e message is the best option since its quite meaningless. you could argue that decryption errors should also be an e2e message. i would prefer sth like verification message.

app/script/conversation/ConversationRepository.coffee
herrmannplatz: fixed in bennys branch :)

app/script/conversation/ConversationRepository.coffee
herrmannplatz: we only need to assure that the clients are mapped to the users

app/script/conversation/EventMapper.coffee
gregor: 👍

PR #681, opened 2017-01-24
Make the fingerprint selectable

lipis:
Fixes #320

### DIFF
--- a/app/style/components/device-card.less
+++ b/app/style/components/device-card.less
@@ -56,6 +56,7 @@ device-card {
 }
 
 .device-id-part {
+  .text-selection();
   display: inline-block;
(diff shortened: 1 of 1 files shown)


herrmannplatz:
for the future we could introduce a generic class that forces text to be selectable

herrmannplatz:
this way we could use it in several places

lipis:
That's what I wanted in the beginning, but the way this was setup.. I think that made more sense.

Maybe instead of mixin it could be a normal class `.text-selection` and we could use it wherever we want in HTML.

herrmannplatz:
Ah yeah sure. That might be the best option.

PR #684, opened 2017-01-25
Catching degraded cancellation errors

bennyn:


bennycode:
The purpose is to have no unhandled rejections when a user cancels sending a message. 

Currently `send_message` can only `resolve` or `reject`. If I catch errors inside `send_message`, then it resolves properly and messages will show up in black color in the UI. 

In this case we want to fail on message sending (so that messages will be rendered gray in the UI) but we don't want to show an unhandled rejection in the console (because it is intended behaviour). So @herrmannplatz suggested to catch the error within `unhandledrejection`.

bennycode:
@herrmannplatz & @gregor  ?!

PR #696, opened 2017-01-28
Handle source of update

gregor:


app/script/announce/AnnounceRepository.coffee
```diff
@@ -45,7 +45,7 @@ class z.announce.AnnounceRepository
   check_version: =>
     @announce_service.get_version()
     .then (server_version) ->
-      amplify.publish z.event.WebApp.LIFECYCLE.UPDATE if server_version > z.util.Environment.version false, true
+      amplify.publish z.event.WebApp.LIFECYCLE.UPDATE, 'webapp' if server_version > z.util.Environment.version false, true
```
bennycode: Which other values are possible for `webapp`? These values should be combined in a constant.

gregor: `wrapper` -> which namespace should I add them to?

PR #724, opened 2017-02-03
verificationstate (WEBAPP-3593)

herrmannplatz:
Updated clients via popover should also trigger verificationstatehandler. Shield is now also bound to the verification state.

### DIFF
--- a/test/unit_tests/conversation/ConversationVerificationStateHandlerSpec.coffee
+++ b/test/unit_tests/conversation/ConversationVerificationStateHandlerSpec.coffee
@@ -110,9 +110,9 @@ describe 'z.conversation.ConversationVerificationStateHandler', ->
       spyOn z.conversation.EventBuilder, 'build_degraded'
       spyOn z.conversation.EventBuilder, 'build_all_verified'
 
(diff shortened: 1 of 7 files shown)


test/unit_tests/conversation/ConversationVerificationStateHandlerSpec.coffee
```diff
@@ -127,6 +127,29 @@ describe 'z.conversation.ConversationVerificationStateHandler', ->
       expect(conversation_c.verification_state()).toBe z.conversation.ConversationVerificationState.VERIFIED
       expect(z.conversation.EventBuilder.build_all_verified.calls.count()).toEqual 3
 
+  describe 'on_clients_updated', ->
+
+    it 'should change state from DEGRADED to VERIFIED if last unverified client was removed by other user', ->
+      spyOn z.conversation.EventBuilder, 'build_degraded'
+      spyOn z.conversation.EventBuilder, 'build_all_verified'
+
+      new_client_b = new z.client.Client()
```
bennycode: Why is this called `new_client_b`? I can see a `new_client_a` or `new_client`?

PR #736, opened 2017-02-07
session id helper

herrmannplatz:


app/page/template/_dist/app.htm
```diff
@@ -178,6 +178,7 @@
 <script src="/script/cryptography/CryptographyMapper.js"></script>
 <script src="/script/cryptography/CryptographyRepository.js"></script>
 <script src="/script/cryptography/CryptographyService.js"></script>
+<script src="/script/cryptography/SessionID.js"></script>
```
bennycode, line 181: `SessionID.js` sounds like it is an entity. I would prefer a `SessionHelper` or `SessionUtil`.

PR #750, opened 2017-02-10
Some layout and font edits

antonshulgin:
## Some layout tweaks

Removed several media-query rules and max-widths to utilize available horizontal space more efficiently.

Also, smaller link/attachment previews.

### Before

image

### After

image

----

## Smaller font size on smaller viewports and other font-related edits

- replaced hardcoded default font weight with a variable;

- added a media query rule that scales down messages font size on smaller
viewports (@screen-lg-min and down);

- removed some rogue tabs.

### Before

image

### After

image

### DIFF
--- a/app/style/base.less
+++ b/app/style/base.less
@@ -27,7 +27,7 @@ body {
   color: @graphite-dark;
   font-size: @font-size-sm;
   font-family: @font-family-base;
-  font-weight: 400;
+  font-weight: @font-weight-regular;
   overflow-x: hidden;
   overflow-y: hidden!important;
   text-rendering: optimizeLegibility;
--- a/app/style/common/variables.less
+++ b/app/style/common/variables.less
@@ -25,18 +25,18 @@
 @font-family-base:      BlinkMacSystemFont, -apple-system, "Helvetica Neue", Arial, sans-serif;
 @font-family-ephemeral: 'Redacted Script';
 
-@font-size-xxs:       8px;
-@font-size-xs:        11px;
+@font-size-xxs:       11px;
+@font-size-xs:        13px;
 @font-size-sm:        16px;
 @font-size-md:        20px;
 @font-size-lg:        24px;
 @font-size-xl:        32px;
 @font-size-xxl:       56px;
 
-@line-height-xs:      11px;
-@line-height-sm:      16px;
-@line-height-md:      20px;
-@line-height-lg:      24px;
+@line-height-xs:      13px;
+@line-height-sm:      19px;
+@line-height-md:      24px;
+@line-height-lg:      28px;
 @line-height-xl:      32px;
 @line-height-xxl:     64px;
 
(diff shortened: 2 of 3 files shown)


bennycode:
@antonshulgin Thanks for your contribution!!
@markesaaremets Can you please give feedback?

max-baz:
Thanks for the contribution! I'd like to give my personal feedback on this pull request.

First of all, did the fonts became... bolder? This is visible both in the content and in the input box. I like the thin fonts very much, I wish to keep them that way :) 

As for decreasing the font size, I'm not sure yet, on the one hand I see more content, on the other hand it feels like it's getting harder to actually read it. Don't have a definite opinion yet.

www gifcreator me_fiuc66

Removing max-width on the other hand is a clear no, I'm afraid I can't share the full experience via a gif, but believe me, it is impossible to read such a wide text.

www gifcreator me_mhc5vd


All in all, I'd rather consider creating a separate "condensed" theme by using some bits of this PR, but not a replacement for the current theme. Smaller text, smaller images and videos, a general goal to fit more content - some people would like it.

antonshulgin:
Hi there.

> did the fonts became... bolder?

Yes, they're are (were 300, now 400), and many people with poor vision will thank you for that.

> As for decreasing the font size, I'm not sure yet, on the one hand I see more content, on the other hand it feels like it's getting harder to actually read it.

It only gets smaller if the viewport is less than a thousand-something pixels wide.

## The reason behind this change is that on small screens (talking 1280x768 ± 200) the current text size is unnecessarily huge.

> Removing max-width on the other hand is a clear no, I'm afraid I can't share the full experience via a gif, but believe me, it is impossible to read such a wide text.

That's an artificial issue. It would become a real issue if we were talking about a book or an article layout, yeah, but a typical chat log doesn't look like anything on your (and mine) screenshots.

Slack doesn't have a problem with that:

pek

Neither does Discord:

pek

Nor Telegram:

pek

Nor anybody else:

pek

And you shouldn't. Nobody normally talks in paragraphs.

> All in all, I'd rather consider creating a separate "condensed" theme by using some bits of this PR, but not a replacement for the current theme. Smaller text, smaller images and videos, a general goal to fit more content - some people would like it.

I personally will be happy with just a working cmd+plus/minus shortcut in the desktop app, to be honest. :o)

Cheers, have a good one.

markesaaremets:
Hi @antonshulgin. Thank you for your contribution!

At this point we are not looking to increase the line length for chat content as it worsens the readability. However we are working on some layout changes, that will make use of vertical space more efficient. Changing the design for Link Previews was the first step on that, for example.

bennycode:
@antonshulgin Thanks for your Pull Request and your precise description of the improvement.

You've already noticed that changing the line length is a hot topic. There are many contrasting opinions which we have to take into consideration if we want to make the best decision for all of us. That said, we will put your PR on hold until we officially start building a more responsive Wire application.

Our design team (involving @markesaaremets) will address the topic of "application responsiveness" later this year (see https://github.com/wireapp/wire-webapp/pull/543#issuecomment-268046541). We will then check your PR again and use it for inspiration.

Thank you so much!

And all the best from Berlin.

PR #763, opened 2017-02-14
Fixed duplication of own clients

bennyn:

### DIFF
--- a/app/page/template/modals.htm
+++ b/app/page/template/modals.htm
@@ -267,7 +267,7 @@
     </div>
   </div>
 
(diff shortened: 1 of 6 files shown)


app/page/template/modals.htm
```diff
@@ -267,7 +267,7 @@
     </div>
   </div>
 
-  <div class="modal modal-connected-device" data-uie-name="modal-conntected-device">
+  <div class="modal modal-connected-device" data-uie-name="modal-connected-device">
```
bennycode, line 270: @Yserz & @mythsunwind Can you please update our tests to use this new `data-uie-name`?

PR #820, opened 2017-03-01
Use privacy/how/ url for people_tabs_device_detail_how_to link

janraasch:
## Improvement

Send the User directly to `https://wire.com/privacy/how/` instead of opening `https://wire.com/privacy/why/` even though the link is labeled with the text `How do I do that?`

I found the current behavior a little weird, since I had already clicked on `Why verify conversation?` before and when I then clicked on `How do I do that?` I was a little annoyed that I was redirected to the `Why`-support-page again.

## Screenshot

This is not a design change, but just to be clear which link I am talking about:
screen shot 2017-03-01 at 09 23 14


gregor:
@Yorgos-V Can you please check and comment?

gregor:
@janraasch Thank you for your contribution. The link is indeed pointing to the wrong URL and the implementation differs from our mobile platforms. Thanks for pointing this out and contributing with a proper fix. Your implementation is exactly the intended behavior.

janraasch:
@gregor sure, glad I could help. :)

PR #879, opened 2017-03-13
remove ignore suggestions (WEBAPP-3744)

herrmannplatz:


gregor:
Do we have a ticket for this change?

herrmannplatz:
WEBAPP-3744

PR #883, opened 2017-03-14
Support JavaScript for writing source files

lipis:
So simply instead of creating `.coffee` files inside the `app/script` directory we can now create `.js` files and the rest remain the same.

### DIFF
--- a/grunt/config/copy.coffee
+++ b/grunt/config/copy.coffee
@@ -34,8 +34,15 @@ module.exports =
       'image/**/*'
       'font/**/*'
       'style/*.css'
+      'script/*/*.js'
       'worker/*'
     ]
+
+  dist_js:
+    cwd: '<%= dir.app_ %>'
+    dest: '<%= dir.dist %>'
+    expand: true
+    src: 'script/*/*.js'
 
   dist_audio:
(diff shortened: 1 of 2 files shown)


grunt/config/copy.coffee
```diff
@@ -34,9 +34,16 @@ module.exports =
...
       'style/*.css'
+      'script/*/*.js'
       'worker/*'
     ]
 
+  dist_js:
+    cwd: '<%= dir.app_ %>'
+    dest: '<%= dir.dist %>'
+    expand: true
+    src: 'script/*/*.js'
```
ffflorian: Shouldn't it be `script/**/*.js`?

bennycode: Shouldn't `script/*/*.js` become `script/**/*.js`?

grunt/config/copy.coffee
```diff
@@ -64,6 +71,7 @@ module.exports =
       'image/**/*'
       'font/**/*'
       'style/*.css'
+      'script/*/*.js'
```
ffflorian: Same here?

ffflorian: Shouldn't it be `script/**/*.js`?

herrmannplatz:
browsers!

PR #884, opened 2017-03-14
use backend timestamp if newer

herrmannplatz:


app/script/conversation/ConversationMapper.coffee
```diff
@@ -164,12 +164,20 @@ class z.conversation.ConversationMapper
         .map (other) -> other.id
 
       if not local_conversation.last_event_timestamp
+        # we can remove this once BE removed last_event_time
```
lipis: `# TODO: we can...`

PR #885, opened 2017-03-14
Gruntfile.coffee → Gruntfile.js

bennyn:


Gruntfile.js
```diff
@@ -0,0 +1,139 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+'use strict';
+
+module.exports = grunt => {
```
ffflorian: Let's stick to consistency:
```js
module.exports = (grunt) => {
```
instead of
```js
module.exports = grunt => {
```

Like in [line 119](https://github.com/wireapp/wire-webapp/pull/885/files#diff-35b4a816e0441e6a375cd925af50752cR119).

Gruntfile.js
```diff
@@ -0,0 +1,139 @@
...
+      return grunt.task.run("set_version:staging", 'init', "prepare_staging", 'aws_prepare');
+    } else {
+      return grunt.fail.warn('Invalid target specified. Valid targets are: "prod" & "staging"');
+    }
+  });
+
+  // Test Related
+  grunt.registerTask('test', () => grunt.task.run(['clean:docs_coverage', 'scripts', 'test_init', 'test_prepare', 'karma:test']));
+
+  grunt.registerTask('test_prepare', function(test_name) {
```
ffflorian: Why not `(test_name) =>` ?

Gruntfile.js
```diff
@@ -0,0 +1,139 @@
...
+    const vendor_files = prepare_file_names(scripts.vendor);
+    const test_files = test_name ? ["../test/js/#{test_name}Spec.js"] : ['../test/**/*Spec.js'];
+
+    const files = [].concat(helper_files, vendor_files, component_files, app_files, test_files);
+    grunt.config('karma.options.files', files);
+  });
+
+  grunt.registerTask('test_init', ['prepare_dist', 'prepare_test']);
+
+  grunt.registerTask('test_run', test_name => {
```
ffflorian: Here also
```js
(test_name) => {
```

Gruntfile.js
```diff
@@ -0,0 +1,139 @@
...
+  // Deploy to different environments
+  grunt.registerTask('app_deploy', ['gitinfo', 'aws_deploy']);
+  grunt.registerTask('app_deploy_staging', ['gitinfo', 'set_version:staging', 'aws_deploy']);
+  grunt.registerTask('app_deploy_prod', ['gitinfo', 'set_version:prod', 'aws_deploy']);
+
+  grunt.registerTask('app_deploy_travis', (target) => {
+    if (target === 'prod' || target === 'staging') {
+      return grunt.task.run(`set_version:${target}`, 'init', `prepare_${target}`, 'aws_prepare');
+    } else if (target === 'dev') {
+      return grunt.task.run("set_version:staging", 'init', "prepare_staging", 'aws_prepare');
```
ffflorian: `'` instead of `"`

Gruntfile.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const prepare_file_names = (file_name_array) => {
+      return file_name_array.map(file_name => file_name.replace('deploy/', ''));
+    };
+
+    const helper_files = grunt.config.get('karma.options.files');
+    const app_files = prepare_file_names(scripts.app);
+    const component_files = prepare_file_names(scripts.component);
+    const vendor_files = prepare_file_names(scripts.vendor);
+    const test_files = test_name ? ["../test/js/#{test_name}Spec.js"] : ['../test/**/*Spec.js'];
```
ffflorian: `'` instead of `"`

lipis:
First of all I'm 100% with the team to move away from CoffeeScript, even though I used to be a big supporter of it and the main influencer for this choice in this project. But times have changed and CoffeeScript is keeping us behind, as we can't use the latest and greatest features of JavaScript.

Having said that, I'm just a little bit against the timing of it. We should probably update the build scripts as a last step, instead of first.

### Here is why:

- The majority of our users/contributors don't care about the build scripts, but the source code
- We should spend our time on getting rid of CoffeeScript in the source code
- Once everything is in pure JS we can drop CoffeeScript by updating the build scripts as a final step, because:
  - These files are changed very rarely
  - Almost 100% of these files are consisting of variables, JSON objects and comments
  - CSON is much easier to read/manipulate than JSON and there is no need to declare variables in CoffeeScript
  - Readability for better maintenance is actually the main reasons of CS to exist at all
  - It already works, so no risk involved by updating a very important piece of the puzzle
  - No need to spend valuable resources at the moment
  - If refactoring we should maybe even drop Grunt and switch to Gulp (modern, better, faster)?!


### What should we do instead

Since #883 is already merged my vote is to do the following in that order:

1. All the new files should be written in pure JS
2. Create a new issue where we point out to all folders in script directory as checkboxes to track the progress
3. Update Grunt files to JavaScript
4. Remove CoffeeScript from the dependencies of this project

bennycode:
@lipis, dropping CoffeeScript means dropping CoffeeScript.

It doesn't matter with which files we start our "CoffeeScript to JavaScript" movement because by the end of this movement, there will be no more CoffeeScript. 

To start I just picked a file with little logic (which is `Gruntfile.coffee`). It was very easy to transpile `Gruntfile.coffee` because it doesn't use much CoffeeScript sugar. The main parts were:

**1.**

```coffee
prepare_file_names = (file_name_array) =>
  return (file_name.replace 'deploy/', '' for file_name in file_name_array)
```
→
```js
const prepare_file_names = (file_name_array) => {
  return file_name_array.map(file_name => file_name.replace('deploy/', ''));
};
```

**2.**

```coffee
test_files = if test_name then ["../test/js/#{test_name}Spec.js"] else ['../test/**/*Spec.js']
```
→
```js
const test_files = test_name ? ['../test/js/#{test_name}Spec.js'] : ['../test/**/*Spec.js'];
```

Everything else I just throwed into [js2.coffee](http://js2.coffee/) and from there into [Lebab](http://lebab.io/try-it). Afterwards I executed `eslint Gruntfile.js --fix`. Done!

P.S. I know that it hurts letting things go. But in the end it's the birth of something new. 🌻

lipis:
Sure and I don't mind letting it go.. It's done, dead, kaput, null. All I'm saying is that we just have to attack it from the other side :)

ffflorian:
:tada:

PR #890, opened 2017-03-14
Improvement/muted timestamp sync

herrmannplatz:


app/script/conversation/ConversationMapper.coffee
```diff
@@ -181,7 +181,12 @@ class z.conversation.ConversationMapper
         local_conversation.archived_state = remote_conversation.members.self.otr_archived
         local_conversation.archived_timestamp = new Date(remote_conversation.members.self.otr_archived_ref).getTime()
 
-      if not local_conversation.muted_state?
+      remote_muted_timestamp = remote_conversation.members.self.otr_muted_ref
```
gregor: Maybe clearer if you Immediately do the new Date.getTime() calc and then reuse that in L187 and L191?

Issue #891, opened 2017-03-14
Drop CoffeeScript


lipis:
We decided to slowly start moving away from CoffeeScript to JavaScript! As this step is impossible to do it in one go, here is the list of things needs to be done before we sign it off.

### Tools that could help

- [js2.coffee](http://js2.coffee/)
- [lebab.io](http://lebab.io/try-it)
- [decaffeinate](https://github.com/decaffeinate/decaffeinate)

### Things to do

- [x] Support JavaScript, along with CoffeeScript in the build scripts (#883)
- [x] Add ESLint to build pipeline (#889) 
- [x] `app/`
  - [x] `script/`
    - [x] `config.coffee` (#888)
    - [x] `announce/` (#899)
    - [x] `assets/` (#912)
    - [x] `audio/` (#925)
    - [x] `auth/` (#947)
    - [x] `bot/`(#941)
    - [x] `cache/` (#946)
    - [x] `calling/` (#1125)
      - [x] Remove deprecated calling v2 code (#1340)
    - [x] `client/` (#976)
    - [x] `components/` (#1207)
    - [x] `connect/` (#1030)
    - [x] `conversation/` (#1172)
    - [x] `cryptography/` (#1048)
    - [x] `entity/` (#1195)
    - [x] `entity/message` (#1097)
    - [x] `ephemeral/` (#950)
    - [x] `event/` (#1078)
    - [x] `extension/` (#1081)
    - [x] `links/` (#963)
    - [x] `localization/` (#1036)
    - [x] `location/` (#951)
    - [x] `main/` (#1084)
    - [x] `media/` (#1065)
    - [x] `message/` (#1086)
    - [x] `properties/` (#1079)
    - [x] `search/` (#1082)
    - [x] `service/` (#1080)
    - [x] `storage/` (#1083)
    - [x] `system_notification/` (#1052)
    - [x] `telemetry/` (#1089)
    - [x] `tracking/` (#1085)
    - [x] `ui/` (#1120)
    - [x] `user/` (#1138)
    - [x] `util/` (#1114)
    - [x] `view_model/` (#1236)
      - [x] `view_model/bindings/` (#1222)
      - [x] `view_model/content/` (#1235)
      - [x] `view_model/list/` (#1233)
  - [x] `worker` (#1374)
- [x] `test/api`
- [x] `test/unit_tests/` (#1201)
  - [x] `announce/` (#1201)
  - [x] `assets/` (#1201)
  - [x] `audio/` (#1201)
[...]

PR #895, opened 2017-03-15
MomentSpec.js

lipis:
What do we do with those errors?

screen shot 2017-03-15 at 2 18 42 pm

Related #891


bennycode:
@lipis:

> Use the global form of 'use-strict'. (strict)

Just put `use-strict` on top of the file

> 'describe' is not defined. (no-undef)

You need to define the function as a JS comment on top of the file, like: `/*global describe*/` (http://eslint.org/docs/2.0.0/rules/no-undef)

PR #899, opened 2017-03-15
ES6: Migrated "announce"

bennyn:


app/script/announce/AnnounceRepository.js
```diff
@@ -0,0 +1,138 @@
...
+    CHECK_INTERVAL: 3 * 60 * 60 * 1000,
+    CHECK_TIMEOUT: 5 * 60 * 1000,
+    UPDATE_INTERVAL: 6 * 60 * 60 * 1000,
+  };
+
+  window.z.announce.AnnounceRepository = class AnnounceRepository {
+    constructor(announce_service) {
+      this.logger = new z.util.Logger('z.announce.AnnounceRepository', z.config.LOGGER.OPTIONS);
+      this.announce_service = announce_service;
+      return this;
```
herrmannplatz: think we dont need to return this or is it chained?

gregor: It is needed in TestAPI I believe

herrmannplatz: not sure

app/script/announce/AnnounceService.js
```diff
@@ -0,0 +1,58 @@
...
+      if (z.util.Environment.frontend.is_production()) {
+        this.url += '&production=true';
+      }
+      return this;
+    }
+
+    get_announcements() {
+      return new Promise((resolve, reject) => {
+        $.get(this.url).done((data) => {
+          resolve(data['result']);
```
herrmannplatz, line 40: do we want to use fetch for this???

`return fetch(this.url).then((response) => response.json()).then((json) => json.result)`

app/script/announce/AnnounceRepository.js
bennycode: I removed the `this`. It was there because otherwise CoffeeScript would return the last statement from the constructor (instead the instance of the class).

app/script/announce/AnnounceService.js
bennycode: One thing at a time :D

herrmannplatz: means touching it twice

bennycode: Tests fail if I use `return fetch(this.url).then((data) => data.result)`.

lipis: If there "easy" fixes we should do them in one go.. but of course if the tests fail and stuff then maybe have somewhere a list that we can track of the things that we spotted..

lipis: It's a good time to have a complete review of the codebase!

bennycode: Ok, I will see what I can do here...

bennycode: `window.fetch` returns a `ReadableStream` in `data.body`. So the code needs some adjustments.

bennycode: Looks like it needs a reader: `var reader = response.body.getReader();` (https://jakearchibald.com/2015/thats-so-fetch/)

Maybe we should use a wrapper which wraps the fetch API. Like: https://github.com/typicode/fetchival

bennycode: This seems to do the trick: 

```js
get_announcements() {
  return window.fetch(this.url).then((response) => response.json()).then((data) => data.result);
}

get_version() {
  return window.fetch('version/').then((response) => response.json()).then((data) => data.version);
}
```

test/unit_tests/announce/AnnounceServiceSpec.coffee
```diff
@@ -100,3 +100,28 @@ describe 'z.announce.AnnounceService', ->
...
+      server.restore()
+
+    it 'fetches the webapp release version', (done) ->
+      announce_service.get_version()
+      .then (version) ->
+        expect(version).toBe response.version
+        done()
+      .catch done.fail
+
+
```
lipis, line 127: tsk tsk.. double empty line :)

PR #902, opened 2017-03-16
Disable audio output selection on Firefox (WEBAPP-3466)

gregor:
## Type of change
Audio output device selection is not supported on Firefox

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId

### DIFF
--- a/app/script/util/Environment.coffee
+++ b/app/script/util/Environment.coffee
@@ -59,6 +59,10 @@ z.util.Environment = do ->
       return false if window.Notification.requestPermission is undefined
       return false if document.visibilityState is undefined
       return true
+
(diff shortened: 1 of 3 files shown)


app/script/util/Environment.coffee
```diff
@@ -59,6 +59,10 @@ z.util.Environment = do ->
       return false if window.Notification.requestPermission is undefined
       return false if document.visibilityState is undefined
       return true
+
+    supports_audio_output_selection: ->
+      return true if @is_chrome()
+      return false
```
herrmannplatz: return @is_chrome ?

PR #914, opened 2017-03-17
ES6: Migrated "audio"

bennyn:


app/script/audio/AudioRepository.js
```diff
@@ -0,0 +1,250 @@
...
+    }).catch((error) => {
+      this.logger.error(`Failed stopping sound '${audio_id}': ${error.message}`);
+      throw error;
+    });
+  }
+}
+
+window.z = window.z || {};
+window.z.audio = z.audio || {};
+window.z.audio.AudioRepository = AudioRepository;
```
herrmannplatz: AudioRepository is exposed under `AudioRepository` and `window.z.audio.AudioRepository` since the `AudioRepository` function is defined in the global scope.

bennycode: Good catch! Fixed. 🌞

app/script/audio/AudioRepository.js
```diff
@@ -0,0 +1,251 @@
...
+
+  window.z.audio.AudioRepository = class AudioRepository {
+    constructor() {
+      this.logger = new z.util.Logger('z.audio.AudioRepository', z.config.LOGGER.OPTIONS);
+      this.audio_elements = {};
+      this.currently_looping = {};
+      this.audio_preference = ko.observable(z.audio.AudioPreference.ALL);
+      this.audio_preference.subscribe((audio_preference) => {
+        if (audio_preference === z.audio.AudioPreference.NONE) {
+          return this._stop_all();
```
lipis: no need for return

PR #940, opened 2017-03-24
WEBAPP-3764

herrmannplatz:


app/script/entity/Conversation.coffee
```diff
@@ -273,9 +273,14 @@ class z.entity.Conversation
     if first_message? and message_et.timestamp() < first_message.timestamp()
       return
 
-    amplify.publish z.event.WebApp.CONVERSATION.MESSAGE.ADDED, message_et
     @_update_last_read_from_message message_et
-    @messages_unordered.push @_check_for_duplicate_nonce message_et, @get_last_message()
+
+    last_message = @get_last_message()
+    amplify.publish z.event.WebApp.CONVERSATION.MESSAGE.ADDED, message_et
```
gregor: I think we should always call this regardless of whether we add it to the view or not as it updates the collection. Otherwise edits to events older than messages in view are not handled within the collection. Thus this should be called beofre the check in line 273

herrmannplatz: Makes sense. updated it

PR #954, opened 2017-03-28
eslint no-else-return rule

lipis:


ffflorian:
What about the already transpiled code? Will you fix  the no-else-return rule there, too? :)

lipis:
Checking only in `app/script` and the linter didn't find anything else.. it's all green on Travis

lipis:
We'll include the `tests` in a bit

PR #959, opened 2017-03-28
eslint test/unit_tests folder as well

lipis:


.eslintrc.yaml
```diff
@@ -10,13 +11,19 @@ globals:
...
+  entities: true
   ko: true
   LRUCache: true
+  moment: true
+  notification_service: true
   pako: true
   Raygun: true
+  search_service: true
+  user_service: true
+  web_socket_service: true
```
lipis: if there are better ideas.. I'm all ears :)

herrmannplatz:
const > let

ffflorian:
Why do you add `const self = this` everywhere? You could / should use `this`.

lipis:
It was the automatic `eslint --fix` but since there were too many errors again further in travis.. will split this into two again.. only the `unit_tests` for this PR

ffflorian:
Okay. If no further changes will be added, I can approve.

PR #962, opened 2017-03-29
Fixed bug within AnnounceRepository

bennyn:

### DIFF
--- a/app/script/announce/AnnounceRepository.js
+++ b/app/script/announce/AnnounceRepository.js
@@ -43,7 +43,9 @@
     }
 
     check_announcements() {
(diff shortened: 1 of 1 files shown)


app/script/announce/AnnounceRepository.js
```diff
@@ -72,13 +70,12 @@
 
     process_announce_list(announcements_list) {
       if (announcements_list) {
-        for (let announcement in announcements_list) {
+        for (let announcement of Array.from(announcements_list)) {
           if (!z.util.Environment.frontend.is_localhost()) {
```
herrmannplatz: Think there is no need to use Array.from

PR #967, opened 2017-03-29
add moment locales for all supported languages (WEBAPP-3755)

herrmannplatz:


gregor:
We need to make sure that we add new locales to the moment support when adding additional languages. How do we make that part of the process?

PR #977, opened 2017-03-31
Mute sound and hide UI for calls from notification stream (WEBAPP-3776)

gregor:


app/page/template/list/conversations.htm
```diff
@@ -1,157 +1,159 @@
...
-        <!-- /ko -->
-
-        <div class="calls-controls-row calls-controls-row-conversation"
-             data-bind="click: $parent.click_on_conversation">
-          <div class="name"
-               data-bind="attr: {'data-uie-uid': id, 'data-uie-value': display_name()},
-                          css: {'text-theme': $parent.is_selected_conversation($data)},
-                          text: display_name()"
-               data-uie-name="item-call"></div>
+  <!-- ko if: show_calls()-->
```
gregor, line 3: Will create a UI component for this next. Then we can always show the joined call while hiding what we do on the stream without duplicating hundreds of lines of code.

PR #984, opened 2017-04-03
like bug (WEBAPP-3218)

herrmannplatz:

### DIFF
--- a/app/style/common/like.less
+++ b/app/style/common/like.less
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2016 Wire Swiss GmbH
+ * Copyright (C) 2017 Wire Swiss GmbH
  *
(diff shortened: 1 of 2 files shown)


app/style/common/like.less
```diff
@@ -1,25 +1,52 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
```
ffflorian: 2017?

app/style/common/like.less
```diff
@@ -1,25 +1,52 @@
...
   position: relative;
 
   > span {
-    .absolute-center(10px);
-    .font-size-xs;
+    .like-transform-mixin(.75);
     color: @w-red;
+    left: 50%;
+    position: absolute;
+    top: 50%;
```
lipis, line 34: maybe the top/left in the same group :)

PR #989, opened 2017-04-03
🚧  service worker test

herrmannplatz:
- create workaround for ignoreSearch issue
- es6
- test with sw-toolbox


app/sw.js
```diff
@@ -0,0 +1,18 @@
+(global => {
```
bennycode: `(global => {` -> `((global) => {`?

app/sw.js
```diff
@@ -0,0 +1,18 @@
...
+  'use strict';
+
+  importScripts('/ext/js/sw-toolbox.js');
+  importScripts('/worker/lru-cache-strategy.js');
+
+  global.toolbox.options.debug = true;
```
herrmannplatz: false

PR #998, opened 2017-04-04
Removed unused libsodium rule

lipis:


bower.json
```diff
@@ -75,9 +75,6 @@
     "knockout": {
       "js": "dist/knockout.debug.js"
     },
-    "libsodium.js": {
```
bennycode, line 78: It's a transitive dependency from `wire-webapp-cryptobox`, so it's in-use! But it's not declared at the top of the `bower.json` file because it comes as a transitive dependency.

lipis: right.. :) I recently started from scratch and was getting this error during bower install.. works now with another clean start.

PR #1011, opened 2017-04-05
Prevent "undefined" error code

bennyn:

### DIFF
--- a/app/script/event/EventRepository.coffee
+++ b/app/script/event/EventRepository.coffee
@@ -355,7 +355,7 @@ class z.event.EventRepository
         return @cryptography_repository.decrypt_event event
         .catch (decrypt_error) =>
           # Get error information
-          error_code = decrypt_error.code
+          error_code = decrypt_error.code or 999
           remote_client_id = event.data.sender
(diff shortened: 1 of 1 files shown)


app/script/event/EventRepository.coffee
```diff
@@ -355,7 +355,7 @@ class z.event.EventRepository
         return @cryptography_repository.decrypt_event event
         .catch (decrypt_error) =>
           # Get error information
-          error_code = decrypt_error.code
+          error_code = decrypt_error.code or 999
```
gregor, line 358: Can we make it 666 to be the devil?

lipis: as long as it's not a magic number.. I'm down

bennycode: `666` is the [number of the beast](https://en.wikipedia.org/wiki/Number_of_the_Beast) and I don't want to have Satan in our app. 😋

lipis: I wasn't kidding about the "magic" number.. 999 should be a const :|

PR #1023, opened 2017-04-06
Handling reset session request

bennyn:
Note: This needs to get released first > https://github.com/wireapp/proteus.js/pull/62

## Testing

Open a 1:1 conversation and execute the following code:

```js
var conv_id = wire.app.repository.conversation.active_conversation().id;
var our_client_id = wire.app.repository.client.current_client().id;
var senders_client_id = wire.app.repository.conversation.active_conversation().participating_user_ets()[0].devices()[0].id;
var senders_user_id = wire.app.repository.conversation.active_conversation().participating_user_ets()[0].id;
var time = new Date(Date.now()).toISOString();

var event = {
  "conversation": conv_id,
  "time": time,
  "data": {
    "text": "💣",
    "sender": senders_client_id,
    "recipient": our_client_id
  },
  "from": senders_user_id,
  "type": "conversation.otr-message-add"
};

wire.app.repository.event._handle_event(event);
```

## Expected Result

Decryption error message with "Reset Session" button:

capture


app/script/cryptography/CryptographyRepository.coffee
```diff
@@ -19,8 +19,13 @@
 window.z ?= {}
 z.cryptography ?= {}
 
+
 # Cryptography repository for all cryptography interactions with the cryptography service.
 class z.cryptography.CryptographyRepository
+  @::EVENT =
+    RESET_SESSION_REQUEST: '💣'
+
```
herrmannplatz: can we rename this to e.g 'remote_encryption_failure'

app/script/cryptography/CryptographyRepository.coffee
```diff
@@ -19,8 +19,13 @@
 window.z ?= {}
 z.cryptography ?= {}
 
+
 # Cryptography repository for all cryptography interactions with the cryptography service.
 class z.cryptography.CryptographyRepository
+  @::EVENT =
+    RESET_SESSION_REQUEST: '💣'
```
gregor: This is a REMOTE_ENCRYPTION_FAILURE. Reset session requests have their own protobuf message type.

app/page/template/_dist/vendor.htm
```diff
@@ -46,6 +46,7 @@
 <script src="/ext/js/cryptojs/sha256.js"></script>
 <!-- Encryption: Base libraries (don't change the order) -->
 <script src="/ext/js/libsodium.js/sodium.min.js"></script>
+<script src="/ext/js/wire-webapp-cbor/wire-webapp-cbor.js"></script>
 <script src="/ext/js/wire-webapp-proteus/proteus.js"></script>
```
herrmannplatz, line 50: why including cbor?

app/script/cryptography/CryptographyRepository.coffee
```diff
@@ -19,8 +19,13 @@
 window.z ?= {}
 z.cryptography ?= {}
 
+
 # Cryptography repository for all cryptography interactions with the cryptography service.
 class z.cryptography.CryptographyRepository
+  @::EVENT =
```
gregor: No need to make this available everywhere if only used in here.

app/script/event/EventRepository.coffee
```diff
@@ -355,7 +357,8 @@ class z.event.EventRepository
         return @cryptography_repository.decrypt_event event
         .catch (decrypt_error) =>
           # Get error information
-          error_code = decrypt_error.code or 999
+          decrypt_error.code = UNKNOWN_DECRYPT_ERROR_CODE if not decrypt_error.code
```
gregor: Why? Not setting the 999 on the decrypt_error.code but only on the message was on purpose. Otherwise we will mask all unknown error messages in report_decrypt_error from Raygun and Localytics.

``error_code = decrypt_error.code or UNKNOWN_DECRYPT_ERROR_CODE``

app/page/template/_dist/vendor.htm
bennycode: Because the latest Proteus doesn't anymore bundle CBOR.

app/script/event/EventRepository.coffee
lipis: because the `decrypt_error.code` could be used further down the line as well somewhere..

bennycode: Yes, Lipis is right. The function `_report_decrypt_error` uses the `decrypt_error.code` at:

```coffee
    custom_data =
      cryptobox_version: cryptobox.version
      client_local_class: @current_client().class
      client_local_type: @current_client().type
      error_code: decrypt_error.code
      event_type: event.type
      session_id: session_id
```

So I thought it might be good adding the `999` as `code` property to the `decrypt_error` object.

Why do you think that it's masking unknown errors? The `999` is only added in case it's an error without a code which means that it is unknown to us.

app/script/cryptography/CryptographyRepository.coffee
bennycode: Done! :)

app/script/cryptography/CryptographyRepository.coffee
bennycode: Good call!

Issue #541, opened 2016-12-08
Feature request: Emoji autocomplete


max-baz:
I'd like to suggest to implement emoji autocomplete, using [the following cheat sheet](http://www.webpagefx.com/tools/emoji-cheat-sheet/). 

![70370296-d9ff-11e4-80ac-34fc67e16af8](https://cloud.githubusercontent.com/assets/1177900/21028944/b4fb63b8-bd97-11e6-8262-ba7abb68d904.png)

If you ever used Slack, you would probably know how much this is useful. Implement with care, this has to be unobtrusive. For example, on Slack the popup will not appear unless a user enters `:` and a minimum of two characters after it (so any textual emojis like `:)` or `:D` do not cause pop-up to open), and the popup is closed immediately once a user hits Enter to select an emoji to post.

Please note: The emoji button #408 should not be considered as a replacement, they both can coexist nicely. Some people (like me) prefer to search, others prefer to click a button and visually select an emoji.

max-baz:
@herrmannplatz @gregor just wondering if you guys are aiming to implement your own control to show emojis, or you would consider integrating third-party libraries as well?

Just a quick search discovered a few options (without considering licensing or requirements/dependencies, but only looking if something like that exists at all):

https://github.com/missive/emoji-mart
https://github.com/mervick/emojionearea
https://github.com/tommoor/emojione-picker

If going with a third-party library, I would consider only those that could be used for both a button (for #408) and an autocomplete. But before investigating them closer I want to ask if (a) you would consider integrating a third-party library at all (if yes, what are the conditions & requirements), and (b) whether you are not working on this _right now_ (so that I don't interfere :wink:).

gregor:
To manifest my answer here as well. This is something that is high on our wish list. Generally we do consider third-party libraries as well. But mostly for under the hood stuff. As this is a very user centric and visible part of the UI my guess at this time is that we want to integrate a custom solution. I will talk to design when people are back at the office after the Christmas holiday early next year.

DragonzMaster:
+1 For https://github.com/tommoor/emojione-picker

gregor:
[...]

PR #1038, opened 2017-04-07
Emoji autocomplete 🎉 ⚡

maximbaz:
As agreed in #541, I've migrated the code from [Surfingkeys](https://github.com/brookhong/Surfingkeys) to add Emoji autocomplete. I also stripped out some code which is unnecessary for Wire.

I've made it to function properly, and although I'm planning to test more, I think now it's a good time to ask you guys to start giving some feedback on it. Give it a try 😉 

General notes:

- I placed the list of all emojis in the `/app/image` folder, but strictly speaking it is not an image. Any suggestions on where to better place it? My only need is to be able to download this file from javascript, so it must be accessible as a static resource.
- I placed the whole code in a separate file, because I wrote it in javascript due to #891 and due to Surfingkeys being written in javascript. The new ViewModel looks a bit artificial, let me know if you have some better ideas how to organize the code.

Unresolved questions:

- [x] When emoji list is open, press on "Settings" button and expect emoji list to hide - it does not.
- [x] How to better sort emojis in the list?

I'll be posting some more notes and questions inline.

/cc @bennyn, @gregor


app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,167 @@
...
+window.z = window.z || {};
+z.ViewModel = z.ViewModel || {};
+
+z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewModel {
+  constructor() {
+    this.emoji_list = $('<div class="conversation-input-emoji-list" />');
+    this.emoji_dict = undefined;
+    this.emoji_start_pos = -1;
+
+    $.get('/image/emoji.tsv', (text) => {
```
max-baz: Maybe you can suggest a better way to load emojis, not by downloading the file like that?
I was thinking of creating JS file with one multiline constant and just including it in `app.htm`, but I'm not sure if that is any better.

This happens once per page load anyway.

app/style/content/conversation/conversation-input.less
```diff
@@ -167,3 +167,27 @@
...
+  display: block;
+  box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.2);
+  background: white;
+
+  div {
+    border-bottom: 1px solid #eee;
+    padding: 4px 8px;
+
+    &.selected {
+      background: #2391d3;
```
max-baz: I wanted to use user's color theme here, but I can't find a LESS variable that would give me the color. Did I miss it?
Saw `z.entity.User.accent_color()`, but definitely do not want to do this via JS.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,175 @@
...
+    if (query.length < ConversationInputEmojiViewModel.MIN_QUERY_LENGTH || query[0] === ' ' || !this.emoji_dict) {
+      this.emoji_list.remove();
+    } else {
+      let emoji_matched = this.emoji_dict
+        .filter((emoji) => emoji.indexOf(query) !== -1)
+        .slice(0, ConversationInputEmojiViewModel.EMOJI_LIST_LENGTH)
+        .map((emoji) => {
+          let [code, name] = emoji.split('\t');
+          let parsed_unicode_emoji = String.fromCodePoint.apply(null, code.split(','));
+          return `<div><span>${parsed_unicode_emoji}</span>${name}</div>`;
```
max-baz: I know the project uses knockout.js and this whole creation of emoji list could have been done nicer. However I am quite hesitant to do such refactoring now, as this will add complexity to the whole feature. 

This code is tested by many Surfingkeys users, we only need to test how well it integrates with Wire. Refactoring will be more risky, all of us would have to invest more time to get it right. 

Considering this, do you think we can live with such code for emojis v1?

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  rotate_emoji_list(backward) {
+    let previous = this.emoji_list.find('>div.selected');
+    let new_selection = (previous.index() + (backward ? -1 : 1)) % this.emoji_list.find('>div').length;
+    previous.removeClass('selected');
+    this.emoji_list.find(`>div:nth(${new_selection})`).addClass('selected');
+  }
+
+  get_cursor_pixel_pos(input) {
```
max-baz: This is some dark magic... 😉 
but it works well, the emoji list is always positioned correctly from what I've tested so far.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    event.preventDefault();
+    return true;
+  }
+
+  on_input_key_up(data, event) {
+    if (!this.suppress_key_up) {
+      let input = event.target;
+      let text = input.value || '';
```
lipis: `const`

lipis:
Just awesome! 🎠

app/style/content/conversation/conversation-input.less
lipis:
```
background-color: @w-blue;
```

It's "hidden" a bit deeper: https://github.com/wireapp/wire-theme/blob/master/app/style/wire-theme.less#L29

app/style/content/conversation/conversation-input.less
```diff
@@ -167,3 +167,27 @@
...
+      color: white;
```
lipis: `#fff`

app/style/content/conversation/conversation-input.less
```diff
@@ -167,3 +167,27 @@
 .conversation-input-paste-cancel {
   margin-right: 32px;
 }
+
+.conversation-input-emoji-list {
+  z-index: 2147483000;
+  position: fixed;
+  display: block;
+  box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.2);
+  background: white;
```
lipis: `background-color: #fff`

app/style/content/conversation/conversation-input.less
```diff
@@ -167,3 +167,27 @@
 .conversation-input-paste-cancel {
   margin-right: 32px;
 }
+
+.conversation-input-emoji-list {
+  z-index: 2147483000;
```
lipis: Maybe [`@z-index-level-7`](https://github.com/wireapp/wire-webapp/blob/dev/app/style/common/variables.less#L155)

app/style/content/conversation/conversation-input.less
max-baz: cool :+1:  but is there a way which will give me currently selected color, not always blue?
I stumbled upon some mixin `.accent-background-color()`, but I can't get it to work 😞

app/script/view_model/ConversationInputEmojiViewModel.js
lipis: If going with this solution we should use `fetch` instead of `$.get`..

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,175 @@
...
+        break;
+      case z.util.KEYCODE.ARROW_UP:
+      case z.util.KEYCODE.ARROW_DOWN:
+        this.rotate_emoji_list(event.keyCode === z.util.KEYCODE.ARROW_UP);
+        this.suppress_key_up = true;
+        break;
+      case z.util.KEYCODE.ENTER: {
+        let input = event.target;
+        let emoji = this.emoji_list.find('>div.selected>span').html();
+        let val = input.value;
```
lipis: `const`

app/style/content/conversation/conversation-input.less
lipis: Right.. some magic will be needed with the themes here then! But it's a bit too late to figure out what exactly :)

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,175 @@
...
+          this.update_emoji_list(input);
+        }
+      }
+    }
+
+    this.suppress_key_up = false;
+    return true;
+  }
+
+  update_emoji_list(input) {
```
lipis: Shall we maybe start using camelCase in the newly created modules? `updateEmojiList`

lipis: My personal vote is for `camelCase`

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: Nice, I didn't even know about this :+1:

app/script/view_model/ConversationInputEmojiViewModel.js
maximbaz: Mine too! and it was in camelCase originally, but I checked out Wire's JS code and saw this, I thought it's your code style and made it compatible 😃

maximbaz: To confirm: everything besides CONST is in camelCase, right? variables, functions?

lipis: we used `snake_case` because of CoffeeScript (and other reasons long long time ago) and for sure we would like to change that since we are moving to ES6.. we'll have to confirm from the rest of the crew if they all (or majority) agree to use `camelCase` in the new modules..

maximbaz: okay, will hold off with changing case until I hear from you :+1:

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,177 @@
...
+        this.emoji_dict = text.split('\n');
+      });
+  }
+
+  static get EMOJI_LIST_LENGTH() {
+    return 5;
+  }
+
+  static get MIN_QUERY_LENGTH() {
+    return 2;
```
lipis: Magic numbers... not sure what would be the best approach.. maybe something like this (from [SO](http://stackoverflow.com/a/32647583/8418)):

```js
const EMOJI_LIST_LENGTH = 5;
const MIN_QUERY_LENGTH = 2;
...

  static get EMOJI_LIST_LENGTH() {
    return EMOJI_LIST_LENGTH;
  }
```

(even though smells like shadowing)

app/style/content/conversation/conversation-input.less
```diff
@@ -167,3 +167,27 @@
 .conversation-input-paste-cancel {
   margin-right: 32px;
 }
+
+.conversation-input-emoji-list {
+  z-index: @z-index-level-7;
+  position: fixed;
+  display: block;
+  box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.2);
```
lipis: even though the designers will have the final say here.. at least style wise we use this format:`rgba(0,0,0,.2)`

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: hm, I actually considered this `static get` to _be_ the place where we define constants 😄 It does feel like an overkill to define a constant, that will be used to return from a constant 😉

lipis: My dream is to include this rule pretty soon: http://eslint.org/docs/rules/no-magic-numbers and while it feels like overkill I'm sure it will throw a warning.. any other ideas?

(obviously excluding `0s` and `1s` I guess from the rule)

lipis:
Works pretty nicely.. but the suggestions are not clickable (nor changeable with arrow keys) is this by design?

screen shot 2017-04-08 at 02 10 32

(Chrome 59dev / macOS)

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: I enabled it locally, and yes, it does throw a warning - will fix it then.

max-baz:
Not clickable - yes, but should be changeable with arrows up and down. Up / down to change, Enter to select, Esc to cancel. Hmm..

max-baz:
Heh, arrows break after clicking 😃 nice catch.
The focus has to stay in text input for arrows to work, so just click in the text input.

Hm, need to think how to handle this properly, as the keydown/keyup listener is registered only on the input box, and when you click elsewhere, the input is not the one who will receive keyup/keydown anymore...

lipis:
And magically fix if you press `Esc`.. good luck :)

lipis:
Also you forgot to /cc our Wire Ambassador of Emojis: @herrmannplatz 😬💩

max-baz:
I think a lot of people would find it natural to click on a emoji in the list, so I added this functionality... it looks a bit hacky, but works 😉

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: Actually, we don't need to create `static get`ters, I created these constants not for exposing them outside of the module, but simply to make it clear what these numbers mean. I'll just leave the `const` definitions, and remove the rest.

app/script/view_model/ConversationInputEmojiViewModel.js
bennyn: Hi Maxim, we decided that we will keep going with `snake_case`. Once we are fully on ES6, we will then start another round of refactoring with `camelCase`. One thing at a time first.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,190 @@
...
+
+  rotate_emoji_list(backward) {
+    const previous = this.emoji_list.find('.emoji.selected');
+    const new_selection = (previous.index() + (backward ? -1 : 1)) % this.emoji_list.find('.emoji').length;
+    previous.removeClass('selected');
+    this.emoji_list.find(`.emoji:nth(${new_selection})`).addClass('selected');
+  }
+
+  enter_emoji(input, emoji_line) {
+    const emoji = emoji_line.find('.symbol').html();
```
bennycode: Can this be done with [.text()](http://api.jquery.com/text/)?

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,190 @@
...
+  get_cursor_pixel_pos(input) {
+    const css = getComputedStyle(input);
+    const ibr = input.getBoundingClientRect();
+    const mask = document.createElement('div');
+    const span = document.createElement('span');
+
+    mask.style.font = css.font;
+    mask.style.position = 'fixed';
+    mask.innerHTML = input.value;
+    mask.style.left = (input.clientLeft + ibr.left) + 'px';
```
bennycode: Please use a [Template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here.

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: Yes, good point :+1:

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,190 @@
...
+    const previous = this.emoji_list.find('.emoji.selected');
+    const new_selection = (previous.index() + (backward ? -1 : 1)) % this.emoji_list.find('.emoji').length;
+    previous.removeClass('selected');
+    this.emoji_list.find(`.emoji:nth(${new_selection})`).addClass('selected');
+  }
+
+  enter_emoji(input, emoji_line) {
+    const emoji = emoji_line.find('.symbol').text();
+    const val = input.value;
+    input.value = val.substr(0, this.emoji_start_pos - 1) + emoji + val.substr(input.selectionStart);
```
bennycode: Here I would do the following:
1. Assign `val.substr(0, this.emoji_start_pos - 1)` to a variable with a meaningful name
1. Assign `val.substr(input.selectionStart);` to a variable with a meaningful name
1. Use a template string for the whole expression

max-baz: Makes sense :+1:

markesaaremets:
Hi. This is starting to look good. Here are some interaction and visual design comments:
 
- Results in the list are currently sorted randomly. If you search `sun` then and `Sun` should be the first one in the list, not `Sunglasses`
- Separators between items can be removed
- Selected result background can be 16% graphite
- Text: 12px Regular
- Row height: 32px
- Emoji size 16px (with 12px left and right padding, so the emoji column would be 40px in total)
- border-radius: 4px
- box-shadow: 0 0 1px 0 rgba(0,0,0,.08), 0 8px 24px 0 rgba(0,0,0,.16)

emoji autocomplete

max-baz:
Hey, thanks @markesaaremets for the suggestions!

- [ ] Results in the list are currently sorted randomly. If you search sun then and Sun should be the first one in the list, not Sunglasses.

Very good point, there is no sorting at the moment, I'll have a look on how to make it better.

- [x] Separators between items can be removed
- [x] Selected result background can be 16% graphite
- [x] border-radius: 4px
- [x] Row height: 32px
- [x] box-shadow: 0 0 1px 0 rgba(0,0,0,.08), 0 8px 24px 0 rgba(0,0,0,.16)

These I like, things look nicer indeed 👍 

- [x] Text: 12px Regular

Judging by the `variables.less`, font size of `12px` is not a standard value, and also judging by your screenshot you have a light font, not a regular one. I added "12px regular" to the branch, could you double check if this is what you expected?

- [x] Emoji size 16px (with 12px left and right padding, so the emoji column would be 40px in total)

The emoji symbols feel too tiny, perhaps consider making them 18px?

emoji

I also noticed these things in your screenshot, which you didn't mention in the text:

- [x] Capitalize the first word of the emoji
    - [x] Also need to make the search case insensitive.

These I like.

- [x] There is a arrow below the emoji list, as if it is a bubble.

When the last item is selected, this looks weird, as if arrow does not really belong there. What do you think about it, should I keep it?

image

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -127,9 +127,10 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
         this.emoji_list.find('.emoji:nth(0)').addClass('selected');
 
         const pos = this.get_cursor_pixel_pos(input);
-        const top = pos.top - this.emoji_list.height();
+        const top = pos.top - this.emoji_list.height() - 10;
+        const left = pos.left - 20;
```
lipis: magic :)

`OFFSET_TOP/LEFT` (or something like that)

max-baz:
So I found an interesting edge case, when emoij list is open, user can click elsewhere and the emoji list should hide itself. It didn't happen before, I extended the code so that the list is removed when edit is cancelled... but this doesn't cover all cases. 

For example, enter `:sm` and while emoji list is open, click on Settings button. 
Suggestions on how to handle this?

markesaaremets:
@maximbaz 

Thanks for the quick response!

- 12px are the names in Conversation, for example. Also the system messages will have the same style (12px Normal and Semibold) very soon.
- 16px emojis are also in conversation (if in the text), so two different sizes for emojis in general should be good.
- Arrow we can keep for now, i'd say.

max-baz:
Got it, thanks, makes sense to be consistent. :+1: 

I'm looking at the sorting now, and I'm struggling a bit with how to make this better. The simple alphabetic sorting will make query like `:sm` compose a list of completely useless results, as 99% of time you are searching for some "smile" emojis when you are typing `:sm`, but instead with alphabetic sorting you are getting:

* Small orange diamond
* Small blue diamond
* Small red triangle
* Small red triangle down

The current code uses the order of emojis as they appear in `emoji.tsv`, so perhaps we could sort this file based on the "popularity", somehow. I did already put some very popular emojis in the top (the first 82 lines), but I didn't order the rest of the file.

Maybe you have some other ideas how to handle this nicely?

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -0,0 +1,197 @@
...
+'use strict';
+
+window.z = window.z || {};
+z.ViewModel = z.ViewModel || {};
+
+const EMOJI_LIST_LENGTH = 5;
+const QUERY_MIN_LENGTH = 2;
+
+const EMOJI_LIST_OFFSET_TOP = 10;
+const EMOJI_LIST_OFFSET_LEFT = 20;
```
lipis: sort them all by name.. no need to group them.

app/script/view_model/ConversationInputEmojiViewModel.js
lipis: I'll delete all those comments to reduce noise..

max-baz:
I tend to keep sorting as it is, based on the "popularity" defined by the ordering in the file. Subjectively I believe the "sunglasses" emoji is used more often than "sunny", that's why it comes first in the list.

image

When you see a specific emoji that should be above in the list, perhaps we could just bring it up in the file.

The cool engineering solution would be to keep a per-user statistics on what emojis this particular user is typing, and use this to further order the list. That way the fans of "Sunny" emoji would have it first in the list for `:sun`. Out of scope for this PR though 😜 

Let me know if you agree. 

I keep the list of unsolved questions in the first message here, besides the emoji ordering there's a bug with "Settings" button that I'm not sure yet how to solve, and apart from that I hope we are pretty close 🎉

lipis:
I personally like this approach to sorting. Makes sense. Good stuff..

and no.. we are not going to personalize the usage of emojis :D

max-baz:
I've just had another look, and I think I might have figured out a way to fix the last issue, to hide emoji list on Settings / Collection buttons click. The trick I used is to subscribe to the `CONTENT.SWITCH` and hide emoji list when it occurs. That helped for the Collection button, but the problem was that click on Settings button did not generate this event... so now it does 🙂 

I believe there are no unresolved issues anymore 🙂

markesaaremets:
alignment issue
- Text alignment seems to be dependent on emoji's width. It'd be a good to fix the width of left column (40px).
- I'd disable the arrow for now
- Name padding-right can be 16px

max-baz:
Nice catch with variable width, @makingthematrix! Fixed all the suggestions. 

Since we don't have the arrow anymore, which offset do you think is the most pleasant to the eye? 🙂 
With arrow we had 20px, but I think it's too much. As of now I changed it to 10px.

emoji

markesaaremets:
@maximbaz Offset 8px ;)

markesaaremets:
@bennyn, @lipis Design-wise, it's good to go.
Thank you, @maximbaz!

lipis:
Boom!! Thanks again @maximbaz 🍻 🍺

bennycode:
Hi @maximbaz, the whole @wireapp team is thankful beyond words.

Your pull request is a great addition to our app. It builds the foundation for emoji input support, one of the most requested features. Based on your contribution we can add an [emoji picker](https://github.com/wireapp/wire-webapp/issues/541#issuecomment-289290025) and further input improvements...

Thanks for that!

Your friends at Wire. 🌝

max-baz:
Thank you guys for the support, reviews and suggestions, I'm glad we made it in the product 🎉

PR #1045, opened 2017-04-10
Selectable name & username

bennyn:
## Type of change

untitled

### DIFF
--- a/app/style/common/label.less
+++ b/app/style/common/label.less
@@ -48,6 +48,9 @@
   font-weight: @font-weight-bold;
 }
 
-.label-username:not(:empty):before {
-  content: "@";
+.label-username {
+  .text-selection;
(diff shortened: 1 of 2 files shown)


app/style/common/label.less
```diff
@@ -48,6 +48,9 @@
   font-weight: @font-weight-bold;
 }
 
-.label-username:not(:empty):before {
-  content: "@";
+.label-username {
+  .text-selection;
+  &:not(:empty):before {
+    content: "@";
```
lipis, line 54: I noticed that we have a mix of quotes.. in our styles :) Maybe switch all to single quotes?

bennycode: In LESS files we always used `"`, so I don't want to make this change atm.

PR #1052, opened 2017-04-10
ES6: migrated "system_notification"

gregor:


lipis:
Maybe the tests as well in the same PR?

gregor:
@lipis: No, skipping that on purpose. If I change the tests at the same time I have not garantuee that the new code passes the old tests. I might make the same migration mistake in both test and code. Thuse lets keep this separate.

PR #1087, opened 2017-04-15
Add linting rules

gregor:

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -65,10 +65,12 @@ rules:
   quotes: [2, single, avoid-escape]
   semi: [2, always]
   semi-spacing: [2, {before: false, after: true}]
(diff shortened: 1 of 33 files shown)


.eslintrc.yaml
```diff
@@ -65,10 +66,13 @@ rules:
   quotes: [2, single, avoid-escape]
   semi: [2, always]
   semi-spacing: [2, {before: false, after: true}]
+  sort-keys: [2, asc, {caseSensitive: true, natural: true}]
```
lipis, line 69: let's do `caseSensitive: false` (even though I doubt that we will have a mix case in any list.. but it is still better I guess.

bennycode: Let's go with the default setting of `true` and revalidate if it bothers us.

gregor: I agree to Benny on this one. Let's stick to the default for now.

PR #1102, opened 2017-04-19
Sort emoji in autocomplete

gregor:
emoji

### DIFF
--- a/app/script/util/keycode.coffee
+++ b/app/script/util/keycode.coffee
@@ -28,6 +28,7 @@ z.util.KEYCODE =
   DELETE: 8
   ENTER: 13
   ESC: 27
+  TAB: 9
   V: 86
 
 z.util.KEYCODE.is_arrow_key = (keyCode) ->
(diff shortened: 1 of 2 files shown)


max-baz:
I don't think this was a good idea, I already attempted to sort emojis in #1038 and I still think we better leave it as it is.

For example, before the sorting, when the priority was defined by the position in the file, this is what you get by typing `:sm`:

image

After this change, this is what you get now:

image

These are literally useless emojis, now people have to type more characters to get to the smileys.

The file purposefully has most common emojis on top so that they are listed first, and people type less to reach them. I'd rather revert the sorting piece, I personally think it doesn't do any good. If you are missing some common emojis, just move them up in the file.

----------------

Another example that I personally use very often, I send 😉 by typing `:wi<Enter>`, before it was like that:

image

Now it is `:wi<Down><Down><Down><Enter>`:

image

Not cool 😉 And I have to type even more characters to get :stuck_out_tongue_winking_eye:, as it is not in the list now at all. 

I am confident many more people need "wink" than "wind blowing face" :wink:

-----------

Another one, for people in love 😍 Before:

image

After:

image

Headphones first? 😝

max-baz:
I added a few examples above, I hope they are convincing enough 😜 
@gregor @bennyn

PR #1104, opened 2017-04-19
Enforce valid JSDoc 3 types

bennyn:
In JSDoc 3 (and TypeScript definitions) the common practice is to use types with lowercase (like `number` and `string`). The only exception is `Object`. 

As a rule of thumb, use lowercase for [primitives](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) and uppercase for objects and object equivalents.

You can also follow along the issue on this topic here: https://github.com/jsdoc3/jsdoc3.github.com/issues/149

### DIFF
--- a/app/script/links/LinkPreviewHelpers.js
+++ b/app/script/links/LinkPreviewHelpers.js
@@ -38,7 +38,7 @@ z.links.LinkPreviewHelpers = {
   /**
    * Get first link and link offset for given text.
    * @param {string} text - Text to parse
-   * @returns {Array<string, number>} First link and it's offset
+   * @returns {Array<string, number>} First link and its offset
    */
   get_first_link_with_offset(text) {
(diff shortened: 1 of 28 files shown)


app/script/links/LinkPreviewHelpers.js
```diff
@@ -38,7 +38,7 @@ z.links.LinkPreviewHelpers = {
   /**
    * Get first link and link offset for given text.
    * @param {string} text - Text to parse
-   * @returns {Array<string, Number>} First link and its offset
+   * @returns {Array<string, number>} First link and it's offset
```
gregor: Please correct spelling `@returns {Array<string, number>} First link and its offset` - its is not it is in this case but a genitive thus it`s its

bennycode: Fixed! Thanks for explainig it to me. 😃

PR #1110, opened 2017-04-21
Fixed context for file download

bennyn:


herrmannplatz:
you could also bind this inside the class

PR #1122, opened 2017-04-24
Rearrange emoji ordering

maximbaz:
Based on my usage and the usage of my contacts, I believe the new ordering makes more sense. Also shows how valuable this in-file positioning really is, and why sorting is not the best choice. 

This pull request depends on the sorting #1102 being reverted.

image

image

image

image

image

Think of it also as a demo of how easy it is to adjust the ordering based on your or users' suggestions 😉 

/cc @gregor, @bennyn, @helenakos, @herrmannplatz

### DIFF
--- a/app/image/emoji.tsv
+++ b/app/image/emoji.tsv
@@ -1,6 +1,7 @@
 0x1f600	grinning
 0x1f601	grin
 0x1f602	joy
+0x1f642	simple smile
 0x1f603	smiley
 0x1f604	smile
 0x1f605	sweat smile
@@ -12,6 +13,7 @@
 0x1f60a	blush
 0x1f60b	yum
 0x1f60c	relieved
+0x2764	heart
 0x1f60d	heart eyes
 0x1f60e	sunglasses
 0x1f60f	smirk
@@ -22,6 +24,7 @@
 0x1f614	pensive
 0x1f615	confused
 0x1f616	confounded
+0x1f48b	kiss
 0x1f617	kissing
 0x1f618	kissing heart
 0x1f619	kissing smiling eyes
@@ -53,8 +56,13 @@
 0x1f633	flushed
 0x1f634	sleeping
 0x1f635	dizzy face
+0x1f644	face with rolling eyes
+0x1f641	slightly frowning face
+0x1f642	slightly smiling face
+0x1f643	upside down face
 0x1f636	no mouth
 0x1f637	mask
+0x1f491	couple with heart
 0x1f638	smile cat
 0x1f639	joy cat
 0x1f63a	smiley cat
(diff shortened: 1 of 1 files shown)


bennycode:
Hi @maximbaz, we just talked about your suggestion and we see your point but we also asked ourselves if it isn't confusing to see "Raised hand" first when typing ":ha"? Thinking of autocompletion I would expect some suggestions to start with "ha" when typing ":ha".

We checked with GitHub and they also show "Haircut" and "Hamburger" first.

**Screenshot:**

unbenannt

Taking this into consideration your PR does not cover all cases (which is also a tough problem to solve!). That's why I think that we should tacle your issue from a different angle. Instad of needing to write `:smiley:` to post a smiley we should just convert `:)` into an emoji. And for commonly used emojis we should introduce a proper emoji picker which gives you the ability to select frequently used icons.

max-baz:
I'm fine with moving "Raising hand" down in the list, but still I see no point in showing "small airplane" as a very first suggestion for "sm", or "wind blowing face" for "wi". I guess I just don't see the point in doing the sorting, except perhaps to make screenshots nice for blog posts?

But you are right that the PR certainly doesn't cover all cases (I still believe it's better than sorting 😛). Converting `:)` to 🙂 and `;)` to 😉 will cover very little subset of cases, but I'd like to see that anyway 🙂. 

As a keyboard user I'd still stay with autocomplete even if there is a picker, and I'd hate to see "wind blowing face" be the first suggestion every single time.

Github is not doing it optimally, I agree. We can do better 😉

max-baz:
By the way, when you say:

> And for commonly used emojis we should introduce a proper emoji picker which gives you the ability to select frequently used icon

Do you have a per-user ordering in mind? 

Honestly I think that would be the best if Wire can learn within a week what are *my* favorite emojis and move them up in the list. A simple count per emoji would do, where emojis with higher count go on top, and emojis with the same count are sorted alphabetically. 

Except, I'm sure there is a privacy concern 😉 Perhaps keep it locally?

max-baz:
The more I think about per-user count-based ordering, the more I believe it's the best choice.

As you pointed out, we can never "guess" which manual ordering will make everyone happy, there will always be someone who has different favorite emojis, and they would hate them being not on top of the list.

Simple alphabetic sorting will make everyone unhappy, just because nobody is sending "wind blowing face".

But a count-based alphabetic sorting will immediately solve the issue for everybody. The "wink" will not be on top in the beginning, but as soon as you choose it at least once, it will have bigger count than every other emoji, and already on the second try the "wink" will be on top.

For the same reason it's fine to keep the counts locally, it's OK to lose them after reinstallation because it's so cheap to recreate the counts, it just requires you sending a favorite emoji _once_ to put it again on top.

> Thinking of autocompletion I would expect some suggestions to start with "ha" when typing ":ha".

With this approach, assuming that I often send "stuck out tongue with winking eye 😜", I would find it very natural to see this emoji in the top when typing ":wink", because it's based on my own preference.

Think about it with the team @bennyn, if you like the idea, I could give a try on implementing it 😉 
Then we will cancel this PR, and I'll create a new one.

helenakos:
👍 for per-user count-based ordering. There's a reason messaging apps have a "Recent" section in the emoji selectors.

max-baz:
Implemented in #1126, thus closing this one.

PR #1126, opened 2017-04-25
Per-user count-based alphabetic emoji sorting

maximbaz:
### Sorting-wise :heart: 
Important: the count is stored per emoji _name_, not per emoji _code_. Because the same emoji can have multiple aliases (for example 🙂 is accessible via 'simple smile' and 'slightly smiling face'). If the emoji _code_ was used, both of these would jump in the top of the list and you will see duplicates.

#### 1. Filter matches only the beginning of the words. 

To reduce noise, "wi" now matches "Stuck out tongue winking eye", but not "Rewind" or "Arrows clockwise" like it is on prod (see the first screenshot below).

#### 2. Reduced minimal number of characters before emoji popup appears from 2 to 1.

First I noticed that there are some emojis whose name consists only of 1 letter: `a`, `b`, `m`, `o` and `x`. There was no way to enter them before.

But this also helps saving keystrokes, i.e. as soon as `wink` is remembered as my favorite, I can type it with just `:w<Enter>`.

Typing `:)` or `:D` doesn't show emoji popup, which was my primary concern. So I actually don't see any reason why not reduce this limit from 2 to 1. Can you think of a counter-example?

#### 3. ~Increased number of emojis in the popup from 5 to 10.~

~We chatted briefly about this with @helenakos. Current 5 feels too small, 10 is kind of a random number, I can adjust if you like. I think with the current vertical layout it becomes noisy to show more than 10 emojis at once, but 10 looks OK (screenshots below).~

#### 4. Sorted `emoji.tsv` by the "name" column.

It doesn't matter how this file is organized, but I thought it would be nice to have _some_ structure in it, especially for future when more emojis will be added. Sorting by name feels more natural.

---------------------

### Before:

image

### After:

image

/cc @bennyn, @gregor


bennycode:
@markesaaremets, what's your take on that?

markesaaremets:
@maximbaz, thanks, I think it's good, but I wouldn't increase the number of emojis we show. It starts to cover the content. I would also sort the top-used ones still alphabetically.

max-baz:
OK, thanks, I'll reduce the number of emojis back to 5.

The top-used are still sorted alphabetically if and only if they were used exactly identical number of times. But this is rare, often one emoji is always used more than another, so it will be on top.

To elaborate: we don't group emojis in two groups "top-used" and "rarely-used", but we use direct counts to choose the order. This gives more agility, depending on the usage pattern a given user might have more than 2 groups. And then, when any two emojis have exactly the same usage count (let's say 20 and 20, but for most of them 0 and 0), then they will be sorted alphabetically.

max-baz:
Hey @bennyn, I hacked around your idea for inline emoji replacement 😛. Do you like it?

emojis

It will require some code cleanup before it can be sent as a PR, but if you do, I can prepare and submit it after this PR is merged 🙂

bennycode:
@maximbaz, your GIFs always make me smile. 😃 Big thumbs up for the inline emoji replacement. 👍 

But as you said, first things first... So back to per-user count-based alphabetic emoji sorting... We can merge your PR once the top-used emojis will be sorted alphabetically (like @markesaaremets suggested). Do you see an option how this is technically possible?

By the way, we are still looking for a [Senior Web Developer](https://wire.softgarden.io/job/830651). 😏

markesaaremets:
@maximbaz, @bennyn Thanks! Regarding the sorting, we got a lot a of feedback for the initial version, which led us to #1102. I'm still afraid it could create some confusion, so if it's an easy tweak, I'd make the last effort to make it extra-nice. If it proves not to work, we can always roll back.

max-baz:
@bennyn @markesaaremets, I really don't think we should sort top-used emojis, and here is why:

The purpose of counting is to make user as efficient as possible, so when they used 😉 83 times and 😜 95 times, this really means that they use 😜 more, and by introducing sorting and placing 😉 above 😜 you introduce artificial delay for the user, you force them to click more buttons to achieve what they do most often.

And it's not an easy tweak, it's just impossible to come up with a "fair" approximation where you say "now these two emojis are used _almost_ equally often" that will satisfy everyone. Make the approximation range too low, and the sorting will never kick in. Make the range too high, and users will have to choose 😜 like hundreds of times before Wire becomes convinced that this is really a more often used emoji.

However, when the usage count is exactly the same, this means user doesn't really care about prioritizing one emoji over another, and this is the perfect place (the only place) to step up and sort alphabetically.

By the way, maybe this will be a convincing argument 😉 

Slack (the master of emojis 😛) is doing exactly what I implemented in this PR, it doesn't attempt to approximate anything and shows most used on top (sorted exactly by usage count), and only when any two emojis have exactly the same usage count, then it sorts alphabetically.

I personally think it's a very pleasant experience, I open search box for the first time and see alphabetically sorted list, I search for "wink" and enter it, I open the search box for the second time and voila, "wink" is already on top - how nice of Wire to remember my preference 😊

max-baz:
Perhaps the best option to avoid user confusion is to communicate something like this on Medium / Twitter?

> Wire now remembers which emojis you like most, and prioritizes them in the suggestion list. The more you use an emoji, the higher it will be in the list, allowing you to enter it faster next time!

max-baz:
@bennyn @markesaaremets just a thought, maybe we should have a quick call and discuss this, to reach consensus? 

Today I think sorting most-used emojis by name would only lead to more confusion, not less, but to be fair I haven't heard the feedback you got that lead to #1102, so maybe I'm just missing something. 

You decide if you want / need this call, and of course if you have time 🙂

max-baz:
@bennyn travis likes me again 🙂

helenakos:
@maximbaz @bennyn @markesaaremets Looping in here. Agree that most used emoji should always be on top. Talked briefly to @markesaaremets, he is in transit now, and agreed that most emoji used should be on top based on count and that we can merge it with that implementation. After it's tested we can see if alphabetical sorting would be more preferred by users.

PR #1133, opened 2017-04-26
Fix some recent bugs with emoji autocomplete

maximbaz:
1. `remove_emoji_list` is executed multiple times on click, even if emoji popup is closed.

This happens because `this.remove_emoji_list.bind(this)` creates a new function every time, and `window.removeEventListener('click', this.bound_remove_emoji_list)` is trying to unassign a different function and silently fails to do so. The fix is to create a bound function and use it to assign and unassign a click-listener, so that `addEventListener` and `removeEventListener` are operating on exactly the same function.

2. Unable to fix typo while typing emoji name

Previously if you entered a typo (like `winn`), emoji popup was closed, but as soon as you removed the second `n` it reappeared with suggestions like `wink`. Now it doesn't reappear, so you have to remove everything and start from scratch. This happenned because when nothing is matched, previously emoji popup was only closed, but then code was changed to actually _remove_ the popup.

/cc @gregor

### DIFF
--- a/app/script/view_model/ConversationInputEmojiViewModel.js
+++ b/app/script/view_model/ConversationInputEmojiViewModel.js
@@ -54,6 +54,7 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
         this.emoji_dict = text.split('\n');
       });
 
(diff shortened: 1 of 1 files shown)


gregor:
Thanks. Still getting used to the ES6 stuff

PR #1139, opened 2017-04-27
Use public emoji list from EmojiOne (284 new emojis!)

maximbaz:
The code is based on #1126 and #1133, **do not spend time reviewing the code until those are merged**. The diff will become much simpler once those 2 are merged.

-----

@helenakos you'll be interested in this one 🙂

Look what I found: https://github.com/Ranks/emojione/tree/master/extras/alpha-codes

> The purpose is to unify various alpha code lists into a single table developers could contribute to and share. With a few different tables floating around, we've done our best to organize and de-fragment the tables...
> Our goal here is to help everyone involved and we appreciate contributions to this list. Please know we don't recommend the primary EAC's change unless there's a major reason for it.

Two reasons for this PR:

#### 1. We are indeed missing some emojis right now

Our custom list of emoji contains 1238 unique emoji. The list hosted by EmojiOne contains 1522 unique emojis + 905 additional tones (2427 emojis in total). Even without counting additional emojis with tone we are missing 284 emojis. 

> 😁 My favorite addition that I discovered so far is :facepalm: 🤦

#### 2. We don't have to maintain the emojis list anymore

Whether a new emoji or a new name for an existing emoji gets into the community, we don't have to take care of our list anymore. Just download [the public list](https://github.com/Ranks/emojione/blob/master/extras/alpha-codes/eac.json) from time to time and place it in `/app/image/emoji.json`. No other modifications required 🙂

-----

For the moment I simply exclude all toned emojis, because they clutter the suggestions list too much (see below). Our list is simply not designed for the tone, it should come separately, where you choose a tone in some settings and all the emojis are used with that tone.

image


max-baz:
@bennyn @gregor FYI this one is ready for review now.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -49,10 +49,29 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
...
     });
 
-    fetch('/image/emoji.tsv')
-      .then((response) => response.text())
-      .then((text) => {
-        this.emoji_dict = text.split('\n').filter((event) => event.length > 0);
+    fetch('/image/emoji.json')
+      .then((response) => response.json())
+      .then((json) => {
+        this.emoji_dict = Object.entries(json)
```
gregor: `Object.entries()` is only supported on Chrome 54 and newer. Please replace.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -120,28 +139,21 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
       this.emoji_list.remove();
     } else {
       const query_words = query.split(' ');
-      const emoji_matched = this.emoji_dict
+      const emoji_matched = Object.values(this.emoji_dict)
```
gregor: Also only supported on Chrome 54 and upwards

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: Good catch @gregor. In fact, I think it was a wrong choice to use dictionary for storing emojis, as we always consume them as arrays. So I changed the code, should be simpler now and not use any experimental features 🙂

bennycode:
@maximbaz, we updated your branch. It will be merged within the next minutes. 🎈

max-baz:
Cool, thanks guys! 
And after it's merged I'll rebase the #1142, the final PR about emojis 😜

bennycode:
Let's go!!! 💃 🎉

PR #1142, opened 2017-04-27
Emoji inline replacement

maximbaz:
The code is based on #1139, do not spend time reviewing the code until that one is merged.

------------

@bennyn this is what you wanted 🎉 

I don't know if you want to keep this list short or to make it as extensive as possible (in which case we need to think about potential false positives, where users actually didn't want the replacement). 

So to start with something, I made a simple short list which covers my personal needs. 
Do you have any additions / adjustments?

- `:)` => 🙂
- `:(` => 🙁
- `:D` => 😄
- `;)` => 😉
- `:*` => 😘
- `:P` => 😛
- `;P` => 😜
- `:/` => 😕

-----

One thing to keep in mind. The replacement list must only contain capital letters after colon, otherwise it would prevent user from searching an emoji. Better shown with a gif:

emojis

### DIFF
--- a/app/script/view_model/ConversationInputEmojiViewModel.js
+++ b/app/script/view_model/ConversationInputEmojiViewModel.js
@@ -27,25 +27,94 @@ const EMOJI_LIST_OFFSET_LEFT = 8;
 const EMOJI_LIST_OFFSET_TOP = 8;
 const QUERY_MIN_LENGTH = 1;
 
+// DO NOT USE COLON WITH LOWERCASE LETTERS IN THE SHORTCUTS, or you will prevent searching emojis.
+// For example, while :D should be replaced with unicode symbol, :d should allow searching for :dancer:
+/* eslint-disable sort-keys, no-multi-spaces */
+const EMOJI_INLINE_REPLACEMENT = [
+  {'shortcut': ':)',    'name': 'slight smile'},
+  {'shortcut': ':-)',   'name': 'slight smile'},
+  {'shortcut': ':D',    'name': 'smile'},
+  {'shortcut': ':-D',   'name': 'smile'},
+  {'shortcut': ':-d',   'name': 'grinning'},
+  {'shortcut': 'B-)',   'name': 'sunglasses'},
+  {'shortcut': 'b-)',   'name': 'sunglasses'},
+  {'shortcut': '8-)',   'name': 'sunglasses'},
+  {'shortcut': ':(',    'name': 'disappointed'},
+  {'shortcut': ':-(',   'name': 'disappointed'},
+  {'shortcut': ';)',    'name': 'wink'},
+  {'shortcut': ';-)',   'name': 'wink'},
+  {'shortcut': ';-]',   'name': 'wink'},
+  {'shortcut': ';]',    'name': 'wink'},
+  {'shortcut': ':/',    'name': 'confused'},
+  {'shortcut': ':-/',   'name': 'confused'},
+  {'shortcut': ':P',    'name': 'stuck out tongue'},
+  {'shortcut': ':-P',   'name': 'stuck out tongue'},
+  {'shortcut': ':-p',   'name': 'stuck out tongue'},
+  {'shortcut': ';P',    'name': 'stuck out tongue winking eye'},
+  {'shortcut': ';-P',   'name': 'stuck out tongue winking eye'},
+  {'shortcut': ';-p',   'name': 'stuck out tongue winking eye'},
+  {'shortcut': ':O',    'name': 'open mouth'},
+  {'shortcut': ':-o',   'name': 'open mouth'},
+  {'shortcut': 'O:)',   'name': 'innocent'},
+  {'shortcut': 'O:-)',  'name': 'innocent'},
+  {'shortcut': 'o:)',   'name': 'innocent'},
+  {'shortcut': 'o:-)',  'name': 'innocent'},
+  {'shortcut': ';^)',   'name': 'smirk'},
+  {'shortcut': ':@',    'name': 'angry'},
+  {'shortcut': '>:(',   'name': 'rage'},
+  {'shortcut': '}:-)',  'name': 'smiling imp'},
+  {'shortcut': '}:)',   'name': 'smiling imp'},
+  {'shortcut': '3:-)',  'name': 'smiling imp'},
+  {'shortcut': '3:)',   'name': 'smiling imp'},
+  {'shortcut': ':\'-(', 'name': 'cry'},
+  {'shortcut': ':\'(',  'name': 'cry'},
+  {'shortcut': ';(',    'name': 'cry'},
+  {'shortcut': ':\'-)', 'name': 'joy'},
+  {'shortcut': ':\')',  'name': 'joy'},
+  {'shortcut': ':*',    'name': 'kissing heart'},
+  {'shortcut': ':^*',   'name': 'kissing heart'},
+  {'shortcut': ':-*',   'name': 'kissing heart'},
+  {'shortcut': ':-|',   'name': 'neutral face'},
+  {'shortcut': ':|',    'name': 'neutral face'},
+  {'shortcut': ':$',    'name': 'flushed'},
+  {'shortcut': ':-X',   'name': 'no mouth'},
+  {'shortcut': ':X',    'name': 'no mouth'},
+  {'shortcut': ':-#',   'name': 'no mouth'},
+  {'shortcut': ':#',    'name': 'no mouth'},
+  {'shortcut': '\\o/',  'name': 'raised hands'},
+  {'shortcut': '<3',    'name': 'heart'},
+  {'shortcut': '</3',   'name': 'broken heart'},
+].sort((first, second) => {
+  if (first.shortcut.length !== second.shortcut.length) {
+    return second.shortcut.length - first.shortcut.length;
+  }
+  return first.shortcut.localeCompare(second.shortcut);
+});
+/* eslint-enable sort-keys, no-multi-spaces */
+
+const EMOJI_INLINE_MAX_LENGTH = Math.max(...EMOJI_INLINE_REPLACEMENT.map((item) => item.shortcut.length));
+
 z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewModel {
   constructor() {
-    const emoji_list_class = 'conversation-input-emoji-list';
+    const emoji_div_class = 'conversation-input-emoji-list';
 
-    this.emojis = [];
-    this.emoji_list = $(`<div class='${emoji_list_class}' />`);
+    this.emoji_list = [];
+    this.emoji_dict = {};
+
+    this.emoji_div = $(`<div class='${emoji_div_class}' />`);
     this.emoji_start_pos = -1;
     this.emoji_usage_count = z.util.StorageUtil.get_value(z.storage.StorageKey.CONVERSATION.EMOJI_USAGE_COUNT) || {};
 
-    $(document).on('click', `.${emoji_list_class}`, (event) => {
+    $(document).on('click', `.${emoji_div_class}`, (event) => {
       const clicked = $(event.target);
-      const emoji = clicked.hasClass('emoji') ? clicked : clicked.closest('.emoji');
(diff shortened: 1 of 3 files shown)


mat1th:
@maximbaz Nice improvement! Maybe you should use the same list as used for the iOS auto correct. You can find it here [Wire-iOS/Sources/emoticons.min.json](https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/emoticons.min.json) It's just a suggestion the Wire team decides :) 

Edit: But there are some shortcuts that are not using a capital first character, may you should delete them.

max-baz:
I personally don't mind, they mostly intersect, the difference is only in two emojis:

`:)` => 🙂 (mine) vs 😊 (iOS)
`:(` => 🙁 (mine) vs 😞 (iOS)

I'd argue my choice for `:)` is better, but it is a personal preference 😄

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -117,25 +172,31 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
...
-          return z.util.StringUtil.sort_by_priority(emoji_name_a, emoji_name_b, query);
+          const usage_count_a = this.get_usage_count(emoji_a.name);
+          const usage_count_b = this.get_usage_count(emoji_b.name);
+          if (usage_count_a > usage_count_b) {
+            return -1;
+          } else if (usage_count_a < usage_count_b) {
+            return 1;
+          } else {
+            return z.util.StringUtil.sort_by_priority(emoji_a.name, emoji_b.name, query);
+          }
```
lipis:
```js
if (usage_count_a === usage_count_b) {
  return z.util.StringUtil.sort_by_priority(emoji_a.name, emoji_b.name, query);
}
return usage_count_b - usage_count_a;
```

max-baz: Good point, this code is introduced in #1126 so I fixed it there :+1: 
The diff in this PR will be much smaller once all the earlier PRs are merged 🙂

mat1th:
@maximbaz I agree with you, but it should have the same (emoji) auto correct on every platform. So :) should always be :blush:  or :slightly_smiling_face: on iOS, Android and web.

max-baz:
And I agree with you 🙂 Let's await to hear what Wire devs say, I created a big backlog of PRs already... #1126, #1133 and #1139 have to be merged first before this one anyway 😄

And looks like there are some problems with Travis, it doesn't process my commits...

bennycode:
> And looks like there are some problems with Travis, it doesn't process my commits..

@maximbaz, I wrote to the team at [Travis CI](https://travis-ci.org/) to see why your commits processed. They told me that their "abuse detection has kicked in". 😁 

> The reason for that seems to be that the user `maximbaz` who created the PR, is not known to us. This is generally a good indicator that a user sends malicious PRs around to trigger unnecessary builds. If `maximbaz` would enable the Travis CI integration on their account, we would let new requests through.

So it looks like you need to connect your GitHub account with https://travis-ci.org/. Can you please do that?

max-baz:
So too many PRs are coming from me 😀 

I created an account at Travis, and it was immediately blocked with the same explanation. 
Dropped a message to their support, hopefully they respond quickly 😉

max-baz:
@bennyn, @gregor and finally this one is ready for review 🎉

gregor:
This is shaping up really nice. @maximbaz Thank you so much for contributing and improving (or rather adding) our emoji game.

Guess we should see next how we can get @ffflorian emoji menu side project into shape that it is design ok so we can also be merged.

max-baz:
Looking forward to see this in production 🎉

Agree, with such menu not only it would be easier to enter emojis with a mouse, but also helpful to search for an emoji if you don't know how it is called. 

Ping me when that work will start, I could perhaps help out with testing and early feedback, and also code reviews - it will be important to align the two solutions, so that they complement each other, and not contradict in the behavior.

gregor:
@maximbaz The prototype is there (https://github.com/wireapp/wire-webapp/tree/emoji-picker) thanks to @ffflorian. We just need the time to polish it involve design and their requirements in the process, which we currently haven't been able to allocate.

bennycode:
Regarding the "inline emojis" we need to see how they behave if a smiley code is part of a word like "**RE:PUBLICA**".

max-baz:
This is a very good catch @bennyn, in the current implementation it is rather tricky to type this word 😄


emoji

I'll think about it, do you have some ideas? I'm also wondering if this is a reaaally rare case, or not too rare.

gregor:
@maximbaz Replacement should only happen if you type ` _SPACE_:P_ SPACE_`.  Check for a single whitespace before and after the match. That should catch all valid use cases.

Similar the popup should immediately close once you type a whitespace. ` :see` should show a menu - `:see_SPACE_ ` should no longer show a menu. That's the feedback we have gathered internally.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -27,6 +27,20 @@ const EMOJI_LIST_OFFSET_LEFT = 8;
 const EMOJI_LIST_OFFSET_TOP = 8;
 const QUERY_MIN_LENGTH = 1;
 
+// DO NOT USE COLON WITH LOWERCASE LETTERS IN THE KEYS, or you will prevent searching emojis.
+// For example, while :D should be replaced with unicode symbol, :d should allow searching for :dancer:
+const EMOJI_INLINE_REPLACEMENT = {
```
mikeger: Hey @maximbaz ! Thanks for the commitment, team loves the new emojis logic!
Small 5 cents: there is already a list of emoticons that iOS client is substituting in the text, it can be potentially used for web too: https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/emoticons.min.json

max-baz:
> Replacement should only happen if you type _SPACE_:P_ SPACE_. Check for a single whitespace before and after the match. That should catch all valid use cases.

There is a catch, with this logic a message `Hello :P` would not be converted to `Hello 😛`, you'd have to type additional space in the end `Hello :P<SPACE>` before it is converted. I think quite many messages end with an emoji, this would be frustrating to remember to always append every such message with a space.

I have an alternative idea, what if we keep replacing like now, immediately, but a "Backspace" hit immediately after would revert the replacement? Something like that (I'm typing here `RE:P<BS>UBLIC`):

emoji

My assumption here is that muuuch more ofter you want your emoji, and quite rarely you want an actual text that has `:P` or `;P` in it.

What do you think?

gregor:
Reverting is not the way to go. People will not learn how to use it. I guess the following space can be replaced by enter as well.

Valid:
`:P<SPACE>World`
`:P<RETURN>`
`Hello<SPACE>:P<SPACE>World`
`Hello<SPACE>:P<RETURN>`

Not valid:
`:PWorld`
`Helllo<SPACE>:PWorld`
`Hello:P<SPACE>World`
`Hello:P<RETURN>`

max-baz:
Yeah, that makes sense, and I can't come up with an counter-argument 🙂. I will adapt the code :+1:

max-baz:
> Similar the popup should immediately close once you type a whitespace. :see should show a menu - :see_SPACE_ should no longer show a menu. That's the feedback we have gathered internally.

This won't work, because emojis consist of several words and we are very much limiting the number of visible emojis to only 5. For example, try to enter `:arrow up` emoji on the current `dev` branch using only one word:

emoji

gregor:
@maximbaz Good point about the multi word descriptions. But it should collapse if you enter two spaces in a row, which it currently doesn't. Let's implement it like that and gather feedback from there.

bennycode:
@maximbaz A bit off topic... but... How do you make these cool GIFs? 😁

max-baz:
@gregor just to understand, is entering two consequent spaces needed specifically to close annoying popup? In other words, would you then remove the second space, after popup is closed? If yes, for the purpose of manually closing popup I would rather recommend hitting `Esc`, this is already implemented. But it is easy to add a check for two spaces if you really want it...


@bennyn 😁 
The tool is called `byzanz-record`, and I have [my own wrapper](https://github.com/maximbaz/dotfiles/blob/972476f1d271bdcfa3249c0d23a24e975c185743/bin/byzanz-record-region.sh) that prompts me to select a region to record.

app/script/view_model/ConversationInputEmojiViewModel.js
max-baz: @mikeger I was waiting for this comment, I will update :+1: 

I have one question... I just have to ask 😄: 

Is it set in stone that `:)` transforms into 😊 and not 🙂 ? The second one feels like a better match, so I wonder if it is possible to reconsider and also update iOS client to transform `:)` into 🙂?

The other replacements are fine, except that I will remove a few, those that have lowercase letter after colon (the reason is described in the code comment just above).

mikeger: @maximbaz the replacements in iOS client are there for quite some years now, so I cannot give you a precise information about why one emoji was chosen over other ones. I think it is acceptable if in the web the replacement would be different from iOS, since the difference is subtle.

mikeger: @maximbaz after discussion with @helenakos we would change the smiling face on iOS at some point to 🙂

gregor:
@maxim Yes, it replicates the `Esc` behavior but is faster to type. No need to even remove  the space since we will trim it if it is at the end of a message anyway.

One other issue we found internally: With the new emoji list, a number of emojis have multiple, similar names. Thus they are shown as duplicates in the list. We should filter that and ensure that each matching emoji is only shown once in the popup.

thumbs

max-baz:
> With the new emoji list, a number of emojis have multiple, similar names.

Good point. I think we shouldn't totally ignore the list of aliases, but we could be smart in choosing what to show in the emoji popup. When two names of the same emoji match user's input, we show only one of them as a match. Will do :+1:

max-baz:
All right, a lot has happened in this branch, I'd appreciate if someone would give it a test run. 

The functional changes (I think I covered all mentioned ideas?):

* The list of inline replacements grew to 53 replacements, it is now almost identical to the list in iOS.
* The character case is now strict, only lowercase letters will trigger emoji popup. That is, contrary to current behavior in `dev` branch, the `:D` will not open popup at all, only `:d` will do. So `:DaGGer` will not search in popup anymore, lower case is for popup, upper case if for inline emojis. This is done to prevent false positives, when typing `:D` you want an inline emoji, you don't want to be annoyed by the popup.
    * There is one special case, a minor annoyance. The list of inline replacements contains emojis with dashes (like `:-)`), but there is also an emoji in the popup which starts with dash: `-1`. This means, when typing `:-)` instead of `:)` you will still get inline replacement, but the popup will also open. 

        emoji

        Since all emojis can be typed both with dash `:-)` and without dash `:)`, I personally will just never use the ones with dash.
* Inline replacement will happen only when typing `Space`, `Enter` or `Tab`, and only when the shortcut (like `:)`) is the first word or has spaces before it.

    emoji2

 * Double spaces and Shift+Enter cancel emoji popup:
    emoji3

* No more duplicates in the popup. If user input matches multiple aliases of the same emoji, only one of them is shown in the popup (primary alias is considered first, if possible). Notice how I can find the same emoji 👍 by `:+1` or `:thumbsup`, but the secondary alias `:thumbup` is not shown when the primary alias `:thumbsup` also matches user input. The other example is with 🙂, while user input matches its primary alias `:slight smile`, only that one is shown, but as soon as user input doesn't match primary alias `:slight smile`, but still matches secondary alias `:slightly smiling face`, the secondary is shown.

    emoji4

-----

The code diff is growing here, I'd prefer to merge this and take further improvement ideas in separate reviews. 

But I don't want to discourage the testing and further feedback 😛, @gregor the suggestions you shared were very interesting and helpful - if the team finds more suggestions, could you maybe drop me a private message, or create a github issue and mention me there?

gregor:
@maximbaz Thank you so much for your work and detailed documentation. We will test, approve and merge your code. This has turned into a full fledged, pretty detailed feature. Let's test this with our colleagues internally and see if it flies.

PR #1148, opened 2017-05-01
Improve Windows fonts: use OpenSans before Arial if available

maximbaz:
I had a chance to see Wire on Windows, and... oh wow, I feel bad for Windows users and their `Arial` font.

Apparently `Open Sans` is used on my Linux, and IMHO it is making Wire look soooo much better than with `Arial` on Windows. It is thinner, and well, just nicer 🙂

@markesaaremets I guess this review is mostly for you. Click on gif to see it full-size.

Before (thick): `Arial`
After (thin): `Open Sans`

windows-fonts

### DIFF
--- a/app/style/common/variables.less
+++ b/app/style/common/variables.less
@@ -30,7 +30,7 @@
 // ----------------------------------------------------------------------------
 @ars-font-path:       '../font/';
 
-@font-family-base:      BlinkMacSystemFont, -apple-system, Helvetica Neue, emoji, Arial, sans-serif;
+@font-family-base:      BlinkMacSystemFont, -apple-system, Helvetica Neue, emoji, Open Sans, Arial, sans-serif;
 @font-family-ephemeral: Redacted Script;
 
 @font-size-xxs:       8px;


markesaaremets:
@maximbaz We're introducing new type styles with overhauled List and will tweak the entire app typography right after that. Also, we try to rely as much as possible on system fonts, meaning on Windows we use Arial (In the future Segoe UI), before third party stuff.

lipis:
Coming soon to your nearest platform 🍻 🏖 ⛰

max-baz:
Good point about using system fonts for text and emoji, this is actually what I like about Wire 🙂

It's nice that you will review the typography in the future, however as a short-term solution, @markesaaremets what would you say about me adding Segoe UI today, just before the Arial? It is a system font after all, and it looks __much better__ than Arial, especially on HiDPI screens. 

I can't share this experience with an image, but Wire with Arial looks like something from 20 years ago on 4k screens...

markesaaremets:
@maximbaz Segoe had legibility issues with our current setup — reason we switched to Arial

max-baz:
Interesting, all right then :+1:

PR #1150, opened 2017-05-01
current_client() can potentially return a NULL if the device limit ha…

segahm:
…s reached and new client is not set


gregor:
@segahm Thanks for your contribution.

Can you please elaborate a little more on the issue you are fixing. When is this safety net needed? You should never end up inside the actual app without the current_device being set. If the device limit page is shown and closed with out deleting a device you are still on the auth page. What are the steps to make this fail?

bennycode:
I will close this issue as we don't have any CoffeeScript code anymore in "app/script/view_model/content".

PR #1161, opened 2017-05-02
eslint rule: id-length

lipis:
http://eslint.org/docs/rules/id-length


app/script/assets/AssetCrypto.js
```diff
@@ -33,10 +33,10 @@ z.assets.AssetCrypto = (() => {
   function decrypt_aes_asset(ciphertext, key_bytes, reference_sha256) {
     return window.crypto.subtle.digest('SHA-256', ciphertext)
       .then(function(computed_sha256) {
-        const a = new Uint32Array(reference_sha256);
-        const b = new Uint32Array(computed_sha256);
+        const first = new Uint32Array(reference_sha256);
+        const second = new Uint32Array(computed_sha256);
```
lipis: I'm open for better names here :)

app/script/media/MediaEmbeds.js
```diff
@@ -100,7 +100,7 @@ z.media.MediaEmbeds = (function() {
         .substr(url.indexOf('?'), url.length)
         .replace(/^[?]/, '&')
         .replace(/[&]v=[a-zA-Z0-9_-]{11}/, '')
-        .replace(/[&#]t=([a-z0-9]+)/, (a, b) => `&start=${_convert_youtube_timestamp_to_seconds(b)}`)
+        .replace(/[&#]t=([a-z0-9]+)/, (first, second) => `&start=${_convert_youtube_timestamp_to_seconds(second)}`)
```
lipis: here as well :)

app/script/assets/AssetCrypto.js
bennycode: Here you are:

```javascript
const typed_reference_sha256 = new Uint32Array(reference_sha256);
const typed_computed_sha256 = new Uint32Array(computed_sha256);
```

app/script/media/MediaEmbeds.js
bennycode: Can you give values for `first` and `second`?

app/script/util/ArrayUtil.js
```diff
@@ -60,8 +60,8 @@ z.util.ArrayUtil = {
       return undefined;
     }
 
-    for (let i = next_index; i <= array.length; i++) {
-      const current_item = array[i];
+    for (let idx = next_index; idx <= array.length; idx++) {
```
gregor: Let's stick with proper names instead of abbreviations

bennycode: Turning `i` into `idx`. 🙈

lipis: It will shadow the `index` that is already used above :D

lipis: the other option is `index_`

bennycode: What about `counter`?  :D

app/script/media/MediaEmbeds.js
lipis: WDYM?

bennycode: To find a better name for `first` and `second` I need to know what they are (values). :D

lipis: I guess URL and timestamp.. will use those

PR #1162, opened 2017-05-03
Fix emoji flags detection

maximbaz:
When you send a message containing only emoji, it will have a larger font.

When you send a message containing only _flag_ emoji (like 🇪🇺), it will not be recognized as emoji-only and thus will have a regular smaller font.

It is just a matter of adding extra range defining which symbols compose flags emojis (and in fact they consist of [regional indicator symbol letters](https://github.com/Ranks/emojione/blob/master/extras/alpha-codes/eac.csv))


test/unit_tests/util/EmojiSpec.coffee
```diff
@@ -32,6 +32,7 @@ describe 'z.util.emoji', ->
       expect(z.util.emoji.includes_only_emojies('☕')).toBeTruthy()
       expect(z.util.emoji.includes_only_emojies('⛄')).toBeTruthy()
       expect(z.util.emoji.includes_only_emojies('⚽')).toBeTruthy()
+      expect(z.util.emoji.includes_only_emojies('🇩🇰')).toBeTruthy()
```
lipis, line 35: Nice touch 🇩🇰

PR #1167, opened 2017-05-03
Eslint won't test recursively on unix

lipis:
It looks like on macOS and Linux `yarn lint-script` won't test the files are deeper than one level, while running the command directly it works :(

With this patch it searches one more level..

screen shot 2017-05-03 at 11 23 57


gregor:
https://github.com/eslint/eslint/issues/1663#issuecomment-240066799

lipis:
Nice one.. I was searching in the different part of the internet :)

PR #1201, opened 2017-05-09
Migrate 'test/unit_tests' to ES6

gregor:


.eslintrc.yaml
```diff
@@ -15,19 +15,26 @@ globals:
   CryptoJS: true
   dcodeIO: true
   Dexie: true
+  entities: true
   hljs: true
+  jQuery: true
   keyboardJS: true
   ko: true
   libsodium: true
+  lorem_ipsum: true
```
bennycode, line 24: Where is `lorem_ipsum` being used?

gregor: It's part of the payloads.js. We might want to move this, entities and whatever is in there to a dedicated namespace in a separate PR.

PR #1204, opened 2017-05-10
conversation list ui translation

herrmannplatz:
This is a proposal to simplify our translations and make the api less verbose. It also introduces new unified placeholder identifiers using `{{\w+}}`.

Would like to get your feedback on this since we should update the current strings later. Right now i would only use it for newly added strings to not break the existing ones. `z.localization.Localizer.get_text` is still present but deprectated.

@lipis @bennyn @gregor


app/script/localization/Localizer.js
```diff
@@ -70,6 +78,53 @@ class Localizer {
...
+z.l10n = (() => {
+
+  function replaceWithString(string, substitute) {
+    return string.replace(/{{\w+}}/, substitute);
+  }
+
+  function replaceWithObject(string, substitute) {
+    for (let identifier in substitute) {
+      if (substitute.hasOwnProperty(identifier)) {
+        string = string.replace(`{{${identifier}}}`, substitute[identifier]);
```
lipis: This will only change the first occurrence, so it will fail on this one:

```js
replaceWithObject('{foo} {bar} {foo}', {'foo': 'FOO', 'bar': 'BAR'})
```

Maybe something like [this](http://stackoverflow.com/a/17606289/8418):

```js
string = string.replace(new RegExp(`{{${identifier}}}`, 'g'), substitute[identifier]);
```

test/unit_tests/localization/LocalizerSpec.js
```diff
@@ -66,3 +66,32 @@ describe('Localizer', function() {
...
+
+  it('can replace placeholders in localized strings using an object', function() {
+    const text = z.l10n.text('Hey {{name}}', {name: 'Tod'});
+    expect(text).toBe('Hey Tod');
+  });
+
+  it('can replace placeholders in localized strings using a more complex object', function() {
+    const text = z.l10n.text('{{greeting}} {{name}}', {name: 'Tod', greeting: 'Hey'});
+    expect(text).toBe('Hey Tod');
+  });
```
lipis: Maybe adding one more test for using the same key multiple times.

app/script/localization/Localizer.js
herrmannplatz: Yes, not sure if we will have same same key twice.

lipis: Future proof ^^

app/script/localization/Localizer.js
```diff
@@ -70,6 +78,53 @@ class Localizer {
...
+     * z.l10.text('Hey {{name}}', 'Tod') // returns 'Hey Tod'
+     *
+     * @example using an object as substitute
+     * z.l10.text('{{greeting}} {{name}}', {name: 'Tod', greeting: 'Hey') // returns 'Hey Tod'
+     *
+     * @param {Observable|string} value
+     * @param {string|Object} substitute
+     * @returns {string}
+     */
+    text(value, substitute) {
```
gregor: Shouldn't this still use the string identifier as input value instead of the string itself or do you want to wrap this once more?

gregor:
As discussed yesterday I like the approach and general idea of this. We should stick to passing in string identifiers instead of strings though.

app/script/localization/Localizer.js
lipis: The substitute is either string or object. For simple ones and where is only one substitute there is no need to use Objects.. Like the first example in the docs above.

gregor: @lipis Not talking about the substitute but the source string. Currently we pass in the string identifier to retrieve the localized string. This just takes a string so we have to get the localized variant of the string in another step.

lipis: I see what you mean. But it will be fine to pass just that: `z.string.collection_section_links` for `value` as it will contain the correct translation because of line 41 above. Am I missing something?

gregor: No, you are right. Thanks.

app/script/localization/Localizer.js
herrmannplatz: Sure i could do that was just thinking if we have situations like this. But sure i can do it :)

test/unit_tests/localization/LocalizerSpec.js
herrmannplatz: Changed it, but use it wisely!

lipis:
❤️

PR #1218, opened 2017-05-12
Use linkify to render email addresses

mythsunwind:
Replaces the regex to link an email address with the linkify way.

* Should make it faster, because the message text needs to be only parsed once now
* Fixes security related issue: https://github.com/wireapp/wire-desktop/issues/528
* Email addresses are not parsed in `<code>` and `<pre>` tags anymore


app/script/util/util.js
```diff
@@ -403,12 +398,21 @@ z.util.render_message = function(message) {
 
   // Parse links with linkifyjs library, ignore code tags
   const options = {
-    attributes: {
-      rel: 'nofollow noopener noreferrer',
+    attributes: function(href, type) {
+      if (type === 'url') return {rel: 'nofollow noopener noreferrer'};
+      if (type === 'email') return {onclick: 'z.util.safe_mailto_open(\'' + href.replace('mailto:', '') + '\')'};
```
lipis: can we skip the one line if statements here?

app/script/util/util.js
```diff
@@ -403,12 +398,21 @@ z.util.render_message = function(message) {
...
+      return {};
+    },
+    formatHref: function(href, type) {
+      if (type === 'email') return '#';
+      return href;
```
herrmannplatz: `return (type === 'email') ? '#' : href`

PR #1223, opened 2017-05-12
archive conversation when previously deleted (WEBAPP-3862)

herrmannplatz:
This ensure that when the conversation was deleted, archive can still be performed locally. I also removed the callback in `clear_conversation`

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -863,12 +863,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
    */
   clear_conversation(conversation_et, leave = false) {
     const next_conversation_et = this.get_next_conversation(conversation_et);
-    return Promise.resolve()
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationRepository.js
```diff
@@ -856,22 +859,21 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
-
-    return _clear_conversation();
+    return Promise.resolve()
+      .then(() => {
+        if (leave) {
+          return this.leave_conversation(conversation_et, next_conversation_et);
+        }
+      })
+      .then(() => {
+        this._update_cleared_timestamp(conversation_et);
```
gregor: No need for extra tick: 

```
const promise = leave ? this.leave_conversation(conversation_et, next_conversation_et) : Promise.resolve()
promise.then(() => {
````

PR #1226, opened 2017-05-14
Replace all inline emoji on Enter

maximbaz:
I'm pretty happy with the inline replacement in the internal build, however I found one annoying issue.

Today, for efficiency purposes the inline emojis are only replaced at the cursor position, that is a user types `:)` and after hitting <kbd>Space</kbd> we check only the last few characters for inline emoji, not the whole text.

However, this logic doesn't work as nicely for <kbd>Enter</kbd>:

emoji

Therefore, the change I'm proposing here is the following:
- When user types <kbd>Space</kbd> (or <kbd>Tab</kbd>), we keep checking only the last word near the cursor for potential inline emoji.
- However, when user types <kbd>Enter</kbd> (both for entering a new line or for sending the msg), we check the whole text and possibly fill up missing cases.

We could have just checked the whole text every time, but my assumption is that <kbd>Space</kbd> is hit a lot more often than <kbd>Enter</kbd> and it would be a waste of resources to do such checks so often.

/cc @gregor 

---------

HINT: Due to a small refactoring, this particular PR is easier to review with ignoring whitespace changes. Append `?w=1` to the URL.


lipis:
Didn't know about `w` argument.. thanks!

gregor:
@maximbaz Indeed it works really nicely. Some people have voiced their desire to be able to send plain ASCII emojis in some cases and we are evaluating that feedback. Other than that we are really happy and excited about the emoji functionality you added to the product. Thanks for proposing this improvement. I fully agree, checked the code and will go ahead and merge.

max-baz:
Cool, let me know what you decide about ASCII emojis :+1:

bennycode:
Thanks @maximbaz. I am a big fan of inline emojs. 🍻 🍻 Thanks for this little diamond! 💎

PR #1244, opened 2017-05-18
map proper conversation id (WEBAPP-3953)

herrmannplatz:


bennycode:
Ok, but EventMapperSpec fails.

herrmannplatz:
fixed

PR #1250, opened 2017-05-18
Use emoji font only after text fonts

maximbaz:
This is a follow-up on what we discussed in Linux Feedback Group. With the current order of fonts, when a user doesn't have either of `BlinkMacSystemFont, -apple-system, Helvetica Neue`, the `emoji` font is attempted to be used to render all text. In particular, this makes it render space character as a zero-width space, like so:

wire 2017-05-18 at 17 57 44

Defining the emoji font as the very last seems to solve the issue.

wire 2017-05-18 at 17 58 25

### DIFF
--- a/app/style/common/variables.less
+++ b/app/style/common/variables.less
@@ -30,7 +30,7 @@
 // ----------------------------------------------------------------------------
 @ars-font-path:       '../font/';
 
-@font-family-base:      BlinkMacSystemFont, -apple-system, Helvetica Neue, emoji, Arial, sans-serif;
+@font-family-base:      BlinkMacSystemFont, -apple-system, Helvetica Neue, Arial, sans-serif, emoji;
 @font-family-ephemeral: Redacted Script;
 
 @font-size-xxs:       8px;


bennycode:
@maximbaz We realized that using the `emoji` font on Windows makes bold emojis appear in black & white. If we don't use `font-weight: bold`, then they become colorful again. Do you see any problem in removing the `emoji` font family at all?

max-baz:
Hi 👋 Do you mean removing [this entire block](https://github.com/wireapp/wire-webapp/blob/ae3b1818d4d05daece7069c7c1d9d79ce3b060c8/app/style/common/variables.less#L23-L87)?

I've never seen emojis actually being bold, maybe just adding `font-weight: normal;` to that `@font-face` would be sufficient?

AndyLnd:
Hi @maximbaz 
unfortunately, just adding `font-weight: normal;` doesn't help.
I investigated further and it seems nothing can be done about this at this point.
We could use a webfont just for the emojis, to make sure that they look the same on every OS and browser. But the added cost of a couple MBs is not justified for such a relatively small issue.

max-baz:
FWIW I actually consider it a huge benefit that Wire respects system emojis and doesn't force some predefined emoji set upon me 😉 

It does sounds like a bug with the emoji font on Windows (or Windows itself), I wish I could tell you how to efficiently report this to Microsoft but I don't know.

PR #1279, opened 2017-05-24
Prevent undefined values from being saved

bennyn:
## Type of change
Possible fix for "WEBAPP-3974"

### DIFF
--- a/app/script/storage/StorageService.js
+++ b/app/script/storage/StorageService.js
@@ -399,6 +399,10 @@ z.storage.StorageService = class StorageService {
    * @returns {Promise} Resolves with the primary key of the persisted object
    */
   save(store_name, primary_key, entity) {
+    if (!entity) {
+      return Promise.reject(new z.storage.StorageError(z.storage.StorageError.TYPE.NO_DATA));
+    }
+
     return this.db[store_name].put(entity, primary_key)
       .catch((error) => {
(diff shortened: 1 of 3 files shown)


app/script/storage/StorageService.js
```diff
@@ -399,6 +399,10 @@ z.storage.StorageService = class StorageService {
    * @returns {Promise} Resolves with the primary key of the persisted object
    */
   save(store_name, primary_key, entity) {
+    if (!entity) {
+      throw new z.storage.StorageError('Value is undefined');
```
gregor: Storage error expects a known type. This will be mapped to 'Unknown storage error' instead.

Also this should return a Promise since dependent calls will otherwise fail.

`return Promise.reject(new z.storage.StorageError(XXX));`

gregor:
I also opened a ticket with Dexie to see what's going on: https://github.com/dfahlander/Dexie.js/issues/541

bennycode:
`z.system_notification.SystemNotificationRepository` test case fails and I don't know why. :'(

gregor:
@bennyn The test failed because I then block after a catch block is always executed. Thus you first passed the test with done() only to then fail it with done.fail(). Apparently that is possible. As the test execution was finished it looked like the next test had failed.

bennycode:
Cool, thanks for the explanation! 😃 👍

PR #1282, opened 2017-05-24
Make `includes_only_emojies` follow Unicode 5.0 standard

maximbaz:
I found more cases like #1162 when emojis are not being recognized as emojis, but instead of fixing this on a case-by-case basis I updated the check so that it follows Unicode 5.0 standard and really checks for all emoji symbols.

Together with #1277 should be really nice 🙂

----------------------

Nice test string: `🙂 😄 ❤ 🏌️‍♀️ 👩‍🚀 ☺️`


app/script/util/Emoji.js
```diff
@@ -22,15 +22,31 @@
...
+  .replace(/U\+/g, '')
+  .split(', ')
+  .reduce((list, codepoint) => {
+    if (codepoint.indexOf('-') === -1) {
+      list.push(String.fromCodePoint(`0x${codepoint}`));
+    } else {
+      const [start, end] = codepoint.split('-').map((code) => parseInt(code, 16));
+      for (let code = start; code <= end; code++) {
+        list.push(String.fromCodePoint(`0x${code.toString(16)}`));
```
lipis: Let's start applying the [no-magic-numbers](http://eslint.org/docs/rules/no-magic-numbers) to the new code :)

PR #1289, opened 2017-05-28
Fix cursor position after inserting complex emoji

maximbaz:
One more addition for complex emojis besides #1277 and #1282.
Code assumed that emoji always consists of one character, but as I now know, this is not correct.

Before:

wire-before

After:

wire-after

---------------

Still not perfect, you see extra wide spaces and generally Chrome treats emojis as separate characters. I know why this happens, but I need to speak to EmojiOne maintainer first, he might assist creating a generic solution, not specific only for Wire.

wire-fun

> 🏌️‍♀️ is really 🏌️ and ♀️
> :family_man_boy_boy:  is really 👨, 👦 and 👦


lipis:
These gifs are just awesome! :)

PR #1322, opened 2017-06-02
Trim new name of user before changing it (WEBAPP-3542)

gregor:


app/script/view_model/content/PreferencesAccountViewModel.js
```diff
@@ -65,8 +65,8 @@ z.ViewModel.content.PreferencesAccountViewModel = class PreferencesAccountViewMo
     this.user_repository.change_accent_color(id);
   }
 
-  change_name(name, event) {
-    const new_name = event.target.value;
+  change_name(view_model, event) {
```
lipis, line 68: is this `view_model` used anywhere?

gregor: No, but you cannot avoid it due to the knockout binding that passes it in.

PR #1347, opened 2017-06-06
Add Prettier

lipis:
[styled with prettier](https://github.com/prettier/prettier)

Related: https://github.com/wireapp/wire-desktop/pull/678


gregor:
Looks like a very convenient approach. I would vote for parens around parameter lists even if there is only one parameter and they are thus obsolete. We had this rule in the linting previously.

gregor:
And "then" on new line with the appropriate indentation.

lipis:
The thing is that [Prettier](https://github.com/prettier/prettier) has more strict rules and there is no such option. I know that there are some things that are done differently and they are explaining some of their decisions in the videos. I strongly recommend watching those.

If everything will work fine, we can even extend it and use git hooks to reformat the code before committing so it will always look exactly the same.

.eslintrc.yaml
```diff
@@ -45,59 +49,31 @@ globals:
...
   prefer-const: 2
   prefer-promise-reject-errors: 2
   prefer-spread: 2
   prefer-template: 2
-  quotes: [2, single, avoid-escape]
-  semi: [2, always]
-  semi-spacing: [2, {before: false, after: true}]
+  prettier/prettier: [2, {
+      singleQuote: true,
+      trailingComma: es5,
```
lipis: Damn... when I'm adding `all` the karma tests are failing :(

lipis:
It says:

```bash
Running "karma:test" (karma) task
Failed to parse file: /Users/lipis/work/wire/wire-webapp/dist/script/util/DebugUtil.js
08 06 2017 15:32:56.784:ERROR [preprocessor.coverage]: Line 31: Unexpected token )
  at /Users/lipis/work/wire/wire-webapp/dist/script/util/DebugUtil.js
...
```

Which is essentially this dangling comma: [DebugUtil.js#L30](https://github.com/wireapp/wire-webapp/blob/prettier/app/script/util/DebugUtil.js#L30)

Why is that?

PR #1376, opened 2017-06-12
ES6 migrate grunt

herrmannplatz:


grunt/config/postcss.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+'use strict';
+
+const processors = [
+  require('autoprefixer')({
+    browsers: [
+      'Chrome >= 52',
+      'Edge >= 14',
+      'Firefox >= 45',
```
gregor: Lets align the numbers with the ones that actually get to access the app in one go

grunt/tasks/misc.js
```diff
@@ -0,0 +1,140 @@
...
+    grunt.log.writeln(`Deployed: ${pkg.name} (v${pkg.version})`);
+  });
+
+  grunt.registerTask('set_version', (target) => {
+    if (target) {
+      grunt.option('target', target);
+      grunt.log.ok(`Version target set to ${grunt.option('target')}`);
+    } else {
+      target = grunt.config('gitinfo.local.branch.current.name');
+    }
```
gregor: I think this is actually error prone despite having the same bug in the coffeescript implementation. We should set target to a default value of grunt.config and then we also no longer need the if but just do it.

PR #1380, opened 2017-06-12
Drop CoffeeScript

lipis:
Final step for #891


grunt/config/todo.js
```diff
@@ -61,8 +61,6 @@ module.exports = {
   },
 
   src: [
-    '<%= dir.app_ %>/**/*.coffee',
```
gregor, line 64: We should add the js files instead. Otherwise the whole task is pointless.

lipis:
Boom!!

PR #1383, opened 2017-06-13
Update active team before setting last active conversation (WEBAPP-4068)

gregor:

### DIFF
--- a/app/script/view_model/content/ContentViewModel.js
+++ b/app/script/view_model/content/ContentViewModel.js
@@ -167,19 +167,8 @@ z.ViewModel.content.ContentViewModel = class ContentViewModel {
           return;
         }
 
+        this._update_active_team(conversation_et);
         this._release_content(this.content_state(), conversation_et);
-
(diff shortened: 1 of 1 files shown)


app/script/view_model/content/ContentViewModel.js
```diff
@@ -279,4 +268,18 @@ z.ViewModel.content.ContentViewModel = class ContentViewModel {
...
     return this._shift_content(this._get_element_of_content(new_content_state));
   }
+
+  _update_active_team(conversation_et) {
+    const {is_guest, team_id} = conversation_et;
+
+    if (this.team_repository.active_team().id !== team_id) {
+      let team_et;
+      if (team_id && !is_guest()) {
+        [team_et] = this.teams().filter((_team_et) => _team_et.id === team_id);
```
herrmannplatz, line 278: maybe `find` is a bit more appropriate here.

herrmannplatz: should we fall back to personal_space if team is undefined?

PR #1388, opened 2017-06-14
Removed GUI faking because of headless Chrome

bennyn:


gregor:
Does travis use the latest Chrome that includes headless support?

bennycode:
Yepp, Travis installs the latest Chrome. Otherwise the Karma run would fail completly with the new settings.

Chrome installation config for Travis is:

```yml
addons:
  apt:
    packages:
      - google-chrome-stable
```

And `google-chrome-stable` refers to v59.0.3071.86 which supports headless mode on Unix systems (Windows systems are supported with v60).

PR #1391, opened 2017-06-14
Add Polish/Dutch

lipis:


gregor:
We also need to add the localized moment versions to `bower.json` and `vendor.htm`

PR #1394, opened 2017-06-15
Check for presence of user in context menu

gregor:
Users reported that the context menu fails to open in some cases. This should fix one of them.


app/script/view_model/list/ListViewModel.js
```diff
@@ -229,7 +229,7 @@ z.ViewModel.list.ListViewModel = class ListViewModel {
     if (!conversation_et.is_group()) {
       const [user_et] = conversation_et.participating_user_ets();
 
-      if (user_et.is_connected() || user_et.is_request()) {
+      if (user_et && user_et.is_connected() || user_et.is_request()) {
```
bennycode: `user_et && (user_et.is_connected() || user_et.is_request())`

PR #1396, opened 2017-06-15
Check for content message on upload failed update

gregor:


herrmannplatz:
can we throw an Error with the super_type?

gregor:
Sure, sounds like a reasonable idea.

PR #1402, opened 2017-06-15
use correct property (WEBAPP-4092)

herrmannplatz:


herrmannplatz:
this fixes https://github.com/wireapp/wire-desktop/issues/675

PR #1405, opened 2017-06-15
Beautify description of inline emoji replacement

maximbaz:
Make it consistent with which emoji will actually be used as a replacement for `:-)`.
Choose a nicer arrow ;)

-------------

Before (current `dev`):

image

After:

image

----

@herrmannplatz 🙂


gregor:
Using our icon font emoji was an explicit requirement from our design team. It even caused us some headache making sure it aligns properly. Thank you for your contribution but we cannot accept this PR.

PR #1410, opened 2017-06-17
Prevent big messages from reaching the client (WEBAPP-4086)

pwnsdx:


pwnsdx:
(WIP)

lipis:
More descriptive title would be nice and you can include the internal issue number in brackets

app/script/event/Client.js
```diff
@@ -36,6 +36,7 @@ z.event.Client = {
     LOCATION: 'conversation.location',
     MESSAGE_DELETE: 'conversation.message-delete',
     MESSAGE_HIDDEN: 'conversation.message-hidden',
+    MESSAGE_TOO_BIG: 'conversation.message-too-big',
```
bennycode: We have different limits for incoming & outgoing messages so maybe better to call it `INCOMING_MESSAGE_TOO_BIG`.

app/script/conversation/EventMapper.js
```diff
@@ -365,10 +366,11 @@ z.conversation.EventMapper = class EventMapper {
    * Maps JSON data of local decrypt errors to message entity
    *
    * @private
-   * @param {Object} error_code - Error data received as JSON
+   * @param {Object} error - Error data received as JSON
    * @returns {DecryptErrorMessage} Decrypt error message entity
    */
-  _map_event_unable_to_decrypt({error_code}) {
+  _map_event_unable_to_decrypt(error) {
+    const {error_code: error_code} = error;
```
gregor: No need to rename a const with the name it already has when descructuring.

app/script/conversation/EventBuilder.js
```diff
@@ -115,6 +115,20 @@ z.conversation.EventBuilder = (function() {
...
+    const {conversation: conversation_id, data: event_data, from, time} = event;
+
+    return {
+      conversation: conversation_id,
+      error: `${message_error.message} (${event_data.sender})`,
+      error_code: `${error_code} (${event_data.sender})`,
+      from: from,
+      id: z.util.create_random_uuid(),
+      time: time,
+      type: z.event.Client.CONVERSATION.INCOMING_MESSAGE_TOO_BIG,
```
gregor, line 128: Nice adding a new message type but how is this getting rendered in the UI?

app/script/conversation/EventMapper.js
ghost: Right. I did it because originally I needed type as error_type as well but now it's not the case anymore

app/script/conversation/EventBuilder.js
ghost: Like the decryption error

PR #1425, opened 2017-06-22
Skip multiple tabs check in electron

gregor:


bennycode:
Should we maybe revert this check in general as there are also some false negatives on webapp?

gregor:
Would keep it for now to understand where the false detections stem from.

max-baz:
One use-case I heard: Wire tab is always pinned in browser, browser is closed (possibly without letting Wire tab remove the cookie) and then reopened (with all tabs being remembered).

PR #1452, opened 2017-06-27
Webview support on webapp (WEBAPP-3942)

pwnsdx:


app/script/media/MediaEmbeds.js
```diff
@@ -42,18 +43,25 @@ z.media.MediaEmbeds = (function() {
...
     }
 
-    if (z.util.Environment.desktop) {
-      options.allowfullscreen = '';
-    }
+    return z.util.StringUtil.format((() => {
+
+      // Build the iframe/webview
+      let iframe_container = '<div class="{0}">';
+      if(z.util.Environment.electron) {
```
gregor: Use `z.util.Environment.desktop` instead as electron will also return true in rambox for example that does not have your implementation. We also need to figure out a way to have this work with older versions of the desktop wrapper. This will break compat with all currently released clients.

pwnsdx: Right. Maybe we can expose a variable from the desktop wrapper to the webapp that says it is supporting webviews and use this variable as the check instead?

app/script/util/Environment.js
```diff
@@ -104,6 +104,10 @@ window.z.util = z.util || {};
       }
       return this.is_firefox();
     },
+    supports_webviews: function() {
+      // Only available in the wrapper since 2.15
+      return (_check.is_desktop() && window.electron_version.localeCompare('2.15') !== -1);
+    },
```
herrmannplatz: `window.electron_version` is deprecated an will be removed

app/script/media/MediaEmbeds.js
```diff
@@ -42,18 +43,24 @@ z.media.MediaEmbeds = (function() {
...
 
     options = _.extend(defaults, options);
-    const iframe_container = `<div class="{0}"><iframe class="${options.type}" width="{1}" height="{2}" src="{3}" frameborder="{4}"{5}></iframe></div>`;
 
-    if (!options.video) {
-      options.allowfullscreen = '';
-      options.class = 'iframe-container';
+    if (options.video) {
+      options.allowfullscreen = true;
+      options.class += ' iframe-container-video';
```
gregor: We use ````options.class = `${options.class} iframe-container-video` ````

app/script/util/Environment.js
```diff
@@ -135,13 +147,7 @@ window.z.util = z.util || {};
...
 
   z.util.Environment = {
-    _electron_version: function(user_agent) {
-      const result = /(Wire|WireInternal)\/(\S+)/.exec(user_agent);
-      if (result) {
-        return result[2]; // [match, app, version]
-      }
-      return undefined;
-    },
+    _electron_version: _check.get_electron_version,
```
gregor: No reason to expose this here.

app/script/util/Environment.js
```diff
@@ -45,6 +45,13 @@ window.z.util = z.util || {};
   };
 
   const _check = {
+    get_electron_version: function(user_agent) {
```
gregor: This is not the electron version but the Wire version. I would name it desktop_version

app/script/util/Environment.js
```diff
@@ -197,7 +204,7 @@ window.z.util = z.util || {};
         return app_version();
       }
 
-      const electron_version = this._electron_version(navigator.userAgent);
+      const electron_version = _check.get_electron_version(navigator.userAgent);
```
gregor: see above. We are not talking about the electron version but our Wire desktop version

pwnsdx: This has nothing to do with my PR, instead it is related to this PR which you approved: https://github.com/wireapp/wire-webapp/pull/1455/files

But I'll fix it.

app/script/util/Environment.js
pwnsdx: https://github.com/wireapp/wire-webapp/pull/1455/files#diff-0c0cbb5c61b8e61b497e31e1cb858b4cR24

PR #1462, opened 2017-06-29
Call setup improvement: 'add_or_update participant' (WEBAPP-4145)

gregor:

### DIFF
--- a/app/script/calling/entities/Call.js
+++ b/app/script/calling/entities/Call.js
@@ -399,7 +399,7 @@ z.calling.entities.Call = class Call {
    */
   _on_send_group_check_timeout(timeout_in_seconds) {
     if (this.participants().length) {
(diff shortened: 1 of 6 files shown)


app/script/calling/entities/Call.js
```diff
@@ -626,35 +646,20 @@ z.calling.entities.Call = class Call {
   /**
    * Add an participant to the call.
```
Yserz: Add or update

app/script/calling/entities/Call.js
```diff
@@ -626,35 +646,20 @@ z.calling.entities.Call = class Call {
...
    *
-   * @param {z.entities.User} user_et - User entity to be added to the call
-   * @param {CallMessage} call_message_et - Call message entity of type z.calling.enum.CALL_MESSAGE_TYPE.SETUP
-   * @param {boolean} [negotiate=true] - Should negotiation be started immediately
+   * @param {string} user_id - ID of user to be removed from the call
```
Yserz: Is removed correct here?

PR #1463, opened 2017-06-30
Use first link preview only (WEBAPP-4120)

gregor:

### DIFF
--- a/app/script/links/LinkPreviewRepository.js
+++ b/app/script/links/LinkPreviewRepository.js
@@ -143,19 +143,12 @@ z.links.LinkPreviewRepository = class LinkPreviewRepository {
         if (error) {
           resolve(undefined);
         }
-        const _first_item_of_array = (property) => {
-          const property_value = data[property];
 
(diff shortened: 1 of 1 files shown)


app/script/links/LinkPreviewRepository.js
```diff
@@ -139,7 +139,26 @@ z.links.LinkPreviewRepository = class LinkPreviewRepository {
...
+          }
+        };
+
+        _first_item_of_array('description');
+        _first_item_of_array('image');
+        _first_item_of_array('title');
+        _first_item_of_array('type');
+        _first_item_of_array('url');
+
+        resolve(data);
```
herrmannplatz: we could to that in a more generic way that cleans up every property

```
Object.keys(data).reduce((filteredData, key) => {
	const value = data[key]
	filteredData[key] = Array.isArray(value) ? value[0] : value
	return filteredData
}, {})
```

gregor: Thanks for the helpful review. Switched to your proposed approach.

PR #1478, opened 2017-07-04
Prevent ghost group calls (WEBAPP-4145)

gregor:
- add call message log
- ensure group check reset on call deactivation

### DIFF
--- a/app/script/main/app.js
+++ b/app/script/main/app.js
@@ -703,6 +703,14 @@ z.main.App = class App {
   }
 
   /**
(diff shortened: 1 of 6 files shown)


app/script/main/app.js
```diff
@@ -703,6 +703,14 @@ z.main.App = class App {
   }
 
   /**
+   * Print call log to console.
+   * @returns {undefined} No return value
+   */
+  print_call_log() {
```
bennycode: good place for debug utils

PR #1512, opened 2017-07-11
fix: Show devices for team members (WEBAPP-4184)

gregor:


gregor:
@Yorgos-V There is not technical reason not to show devices for unconnected users as well. As discussed we will add team members for now and change it to show for all users once discussed with all stakeholders.

PR #1535, opened 2017-07-17
refactor: Use original speakingurl

Yserz:
## Type of change
Switches from forked v10 speakingurl library to the original v13 one.


bennycode:
We need to check that `O Webapp` matches `ω Webapp`.

PR #1536, opened 2017-07-18
fix: asset merge (WEBAPP-3825)

herrmannplatz:


app/script/conversation/ConversationRepository.js
```diff
@@ -2371,6 +2371,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
             return this._on_member_update(conversation_et, event_json);
           case z.event.Backend.CONVERSATION.MESSAGE_ADD:
             return this._on_message_add(conversation_et, event_json);
+          case z.event.Backend.CONVERSATION.ASSET_ADD:
```
gregor: Can we remove the outdated asset related message types from this switch case?

app/script/cryptography/CryptographyMapper.js
```diff
@@ -127,22 +106,62 @@ z.cryptography.CryptographyMapper = class CryptographyMapper {
...
+
+      if (original.image) {
+        data.info.height = original.image.height;
+        data.info.width = original.image.width;
+      } else {
+        data.meta = this._map_asset_meta_data(original);
+      }
+    }
+
+    if (preview != null) {
```
gregor, line 131: No need for != null as preview is an object

herrmannplatz: Sure its an Object (`not_uploaded` is the only exception). But since i want to check for existence, i think its still valuable to check in this way.

app/script/conversation/ConversationRepository.js
herrmannplatz: 👍

herrmannplatz: done

PR #1542, opened 2017-07-20
fix: Disallow message overwrite from same user without edit flag (WEBAPP-4230)

gregor:


gregor:
Need to reconsider the approach this swallows valid messages as well.

gregor:
But it works as expected if we only filter remote messages.

PR #1559, opened 2017-07-26
feat: calling ui redesign (WEBAPP-3080)

herrmannplatz:
Calling UI redesign according to WEBAPP-3080.

Moved all the calling UI into a separate component. We might wanna simplify some of the conditions. But for now it should be good to go.


app/script/components/list/conversationListCallingCell.js
```diff
@@ -0,0 +1,201 @@
...
+    this.on_toggle_video = () => {
+      amplify.publish(z.event.WebApp.CALL.MEDIA.TOGGLE, this.conversation.id, z.media.MediaType.VIDEO);
+    };
+
+    this.users = ko.pureComputed(() => this.conversation.participating_user_ets());
+    this.call = ko.pureComputed(() => this.conversation.call());
+    this.call_participants = ko.pureComputed(() => {
+      return this.call().participants()
+        .map((participant_et) => participant_et.user)
+        .filter((user_et) => !user_et.is_me);
```
gregor: Is this necessary? The self user should never be among the call participants. If you saw that, it is a bug in my code.

app/script/components/list/conversationListCallingCell.js
```diff
@@ -68,6 +68,7 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
     this.call_participants_rest = ko.observable(0);
     this.call_participants.subscribe((user_ets) => {
       const displayed_user_ets = user_ets.slice(0, MAX_DISPLAYED_PARTICIPANTS);
+      console.log(displayed_user_ets, user_ets, MAX_DISPLAYED_PARTICIPANTS);
```
gregor: A console.log ended in the code

PR #1565, opened 2017-08-01
refactor: Media types in calls

gregor:


app/script/calling/CallingRepository.js
```diff
@@ -923,19 +935,21 @@ z.calling.CallingRepository = class CallingRepository {
   /**
    * User action to toggle the call state.
    *
-   * @param {boolean} [video_send=false] - Is this a video call
+   * @param {boolean} media_type - Media type of call
```
bennycode: `{boolean}` should be `{z.media.MediaType}`

PR #1566, opened 2017-08-02
fix: Skip sent status update for reactions (WEBAPP-4260)

gregor:

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -1821,10 +1821,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
           .then((payload) => {
             this._track_completed_media_action(conversation_et, generic_message);
 
-            if (z.event.EventTypeHandling.STORE.includes(message_et.type) || message_et.has_asset_image()) {
-              const backend_timestamp = sync_timestamp ? payload.time : undefined;
-              return this._update_message_as_sent(conversation_et, message_stored, backend_timestamp);
-            }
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationRepository.js
```diff
@@ -1837,22 +1837,22 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
-        const changes = {
-          status: z.message.StatusType.SENT,
-        };
-        message_et.status(z.message.StatusType.SENT);
-
-        if (event_time) {
-          message_et.timestamp(new Date(event_time).getTime());
-          changes.time = event_time;
-        }
+    if (z.event.EventTypeHandling.STORE.includes(message_et.type) || message_et.has_asset_image()) {
```
bennycode: `message_et` needs to become `message` in `if (z.event.EventTypeHandling.STORE.includes(message_et.type) || message_et.has_asset_image())`?!

app/script/conversation/ConversationRepository.js
```diff
@@ -1837,22 +1837,22 @@ z.conversation.ConversationRepository = class ConversationRepository {
    * @returns {Promise} Resolves when sent status was updated
    */
   _update_message_as_sent(conversation_et, message, event_time) {
```
bennycode, line 1839: Renaming `message` into `message_et` here would also do the trick to make Travis happy.

gregor: Only the other way around works. Otherwise the linter is unhappy.

PR #1568, opened 2017-08-02
fix: Remove duplicates from team users

gregor:


gregor:
Tested and verified by @bennyn

PR #1626, opened 2017-08-29
fix: Answering an audio call with video should send video (WEBAPP-4286)

gregor:


app/script/calling/entities/Flow.js
```diff
@@ -873,6 +879,7 @@ z.calling.entities.Flow = class Flow {
 
     const offer_options = {
       iceRestart: restart,
+      offerToReceiveVideo: this.offer_to_receive_video() ? true : undefined,
```
bennycode: I would add a comment there why it is important to use `undefined` instead of `false`.

PR #1652, opened 2017-09-01
fix: File transfers missing from conversation history and collection (WEBAPP-4302)

gregor:

### DIFF
--- a/app/script/storage/StorageService.js
+++ b/app/script/storage/StorageService.js
@@ -260,6 +260,16 @@ z.storage.StorageService = class StorageService {
               this.db[StorageService.OBJECT_STORE.EVENTS].bulkPut(items);
             });
         });
+      this.db.version(14).stores(version_12)
+        .upgrade((transaction) => {
+          this.logger.warn('Database upgrade to version 14', transaction);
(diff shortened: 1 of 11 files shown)


app/script/storage/StorageService.js
```diff
@@ -260,6 +260,16 @@ z.storage.StorageService = class StorageService {
               this.db[StorageService.OBJECT_STORE.EVENTS].bulkPut(items);
             });
         });
+      this.db.version(14).stores(version_12)
+        .upgrade((transaction) => {
+          this.logger.warn('Database upgrade to version 14', transaction);
+          transaction[StorageService.OBJECT_STORE.EVENTS].toCollection()
+            .modify((event) => {
+              if (event.type === 'conversation.asset-meta') {
```
bennycode, line 268: Do we have a constant for `'conversation.asset-meta'`?

gregor: No, this is outdated and deprecated. Choose the string on purpose as this will be the only place we are going to use it moving forward - and we won't break it by accidentally removing the constant elsewhere.

PR #1668, opened 2017-09-05
fix: Upload files with broken metadata

bennyn:


app/script/conversation/ConversationRepository.js
```diff
@@ -1338,7 +1343,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
         return this._send_and_inject_generic_message(conversation_et, generic_message);
       })
       .catch((error) => {
-        this.logger.warn(`Failed to upload metadata for asset in conversation ${conversation_et.id}: ${error.message}`, error);
+        this.logger.warn(`Failed to upload metadata for asset in conversation "${conversation_et.id}": ${error.message}`, error);
```
gregor: Can we stick to single quotes for this use case?

PR #1676, opened 2017-09-06
fix: Removed link from expired previews from self (WEBAPP-4313)

gregor:


bennycode:
I think it would be good to test that a message (which has set `message_et.is_expired()` to `true`) has no link content anymore.

PR #1712, opened 2017-09-18
fix: Set proper initial value for server timestamp (WEBAPP-4325)

gregor:


app/script/conversation/ConversationMapper.js
```diff
@@ -31,22 +31,26 @@ z.conversation.ConversationMapper = class ConversationMapper {
 
   /**
    * Convert a JSON conversation into a conversation entity.
+   *
    * @param {Object} json - Conversation data
+   * @param {number} clock_drift - Approximate time different to backend in milliseconds
```
lipis: `time_offset` maybe?

lipis: or `clock_offset`

PR #1716, opened 2017-09-19
feat: Replaced Localytics with Mixpanel (WEBAPP-4320)

bennyn:


bower.json
```diff
@@ -92,6 +93,12 @@
     "marked": {
       "js": "lib/marked.js"
     },
+    "mixpanel": {
+      "js": [
+        "mixpanel-jslib-snippet.js",
+        "mixpanel.min.js"
```
lipis, line 99: we could include the not minified version.. as we are minifying them anyway in the end..

lipis: or maybe not.. I can see that is used in a different way.. :|

bennycode: Yeah, it's required by Mixpanel that it can load the full-fledged script from it's snippet script: https://github.com/mixpanel/mixpanel-js#using-bower-to-load-the-entire-library

app/script/tracking/CustomDimension.js
```diff
@@ -27,11 +27,10 @@ window.z.tracking = z.tracking || {};
  *
  * @note Custom Dimensions for Localytics have to be created server side and are
```
gregor: Lets search for Localytics and remove it from comments as well.

app/script/tracking/EventTrackingRepository.js
```diff
@@ -80,181 +77,144 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
-    if (this.privacy_preference && !this.session_interval) {
-      if (!this.localytics) {
-        this._init_localytics();
-      }
+    // During the transition phase (Localytics -> Mixpanel), we only want to log certain events:
+    const allowed_events = [
+      z.tracking.EventName.MEDIA.COMPLETED_MEDIA_ACTION,
+      z.tracking.EventName.TRACKING.OPT_IN,
+      z.tracking.EventName.TRACKING.OPT_OUT,
+    ];
```
gregor: Let's define this in config up top like in other repositories instead of creating a variable each time we call `_track_event`

PR #1729, opened 2017-09-21
style: Prettier

lipis:
## Prettier
Read more about it here: https://github.com/prettier/prettier

### Key changes

- [`package.json`](https://github.com/wireapp/wire-webapp/blob/prettier/package.json#L43-L60) - new commands and pre-commit hooks should be much faster now
- [`.eslintrc.yaml`](https://github.com/wireapp/wire-webapp/blob/prettier/.eslintrc.yaml#L68-L74) - for `*.js` files
- [`.prettierrc.yaml`](https://github.com/wireapp/wire-webapp/blob/prettier/.prettierrc.yaml) - for `*.json`, `*.less`

- - - 

### New/modified commands:

- `yarn fix`
  Fixes all the possible file extensions that we support (js, json, less)

- `yarn fix:script`
  Fixes all the script files (js)

- `yarn fix:config`
  Fixes all the config files (json)

- `yarn fix:style`
  Fixes all the styke files (less)

- `yarn test`
  Executes all the tests

- `yarn test:script`
  Tests if the scripts are fine, but not modifying them


ffflorian:
Someone wants to become [contributor](https://github.com/wireapp/wire-webapp/graphs/contributors) nr. 1? :stuck_out_tongue:

lipis:
To become #1 I need to push 800 different commits.. 🗡 :D

lipis:
and I'm still going to be second :P

lipis:
Vote here

app/script/audio/AudioRepository.js
```diff
@@ -149,15 +147,13 @@ window.z.audio.AudioRepository = class AudioRepository {
...
-          audio_element.load();
-        }
-      });
+    Object.keys(this.audio_elements).forEach(audio_id => {
+      if (this.audio_elements.hasOwnProperty(audio_id)) {
+        const audio_element = this.audio_elements[audio_id];
+        audio_element.preload = 'auto';
+        audio_element.load();
+      }
+    });
```
ffflorian, line 156: I like Prettier, but things like this are really a disadvantage. It makes the code not really more readable.

lipis: I think it's quite minor compare to overall readability.. and these cases over time could be simplified anyway.

bennycode:
Thanks @lipis for taking all that effort! I will give my 2 cents when being back from vacation. 

Just on the small side... The PR should be called "style: Prettier": https://seesparkbox.com/foundry/semantic_commit_messages

lipis:
Changed the name.. :) Yes, let's put all the cents here.. so anyone from the community could contribute as well :)

lipis:
Worth reading... https://medium.freecodecamp.org/why-robots-should-format-our-code-159fd06d17f7

bennyn:
Some of my thoughts regarding all this linting:

- https://github.com/prettier/prettier/issues/2716#issuecomment-335799469
- https://github.com/prettier/prettier-eslint/issues/101#issuecomment-335787545
- https://github.com/twbs/bootstrap/pull/24323#issuecomment-335784583
- https://github.com/okonet/lint-staged/issues/150#issuecomment-335776170

bennycode:
## My thoughts regarding all this linting
- I don't like how Prettier [handles array items](https://github.com/prettier/prettier/issues/2716#issuecomment-335799469)
- I don't like Prettier's "[three or more](https://github.com/prettier/prettier/pull/2673)" rule for chained methods ([reason](https://github.com/prettier/prettier/issues/1565#issuecomment-300133602))
- I don't see a value of [Prettier over ESLint](https://github.com/prettier/prettier-eslint/issues/101#issuecomment-335787545)
- I don't like that "lint-staged" [doesn't work for modified files](https://github.com/okonet/lint-staged/issues/150#issuecomment-335776170)
- Because Prettier is opinionated [it doesn't solve every preference](https://github.com/twbs/bootstrap/pull/24323#issuecomment-335736213), whereas preferences can be specified in ESLint (or implemented using ESLint Plugins)
- I don't like that Prettier [can conflict with ESLint](https://github.com/prettier/eslint-config-prettier)

Taking everything from above into consideration, I think we are best off using ESLint (with our [own shareable config](https://eslint.org/docs/developer-guide/shareable-configs)) for everything related to JavaScript and JSX.

For styling CSS, Less and SCSS we can go with Prettier.

lipis:
> I don't like how Prettier handles array items

You win some, you lose some. But this is minor compare to [benefits](https://medium.freecodecamp.org/why-robots-should-format-our-code-159fd06d17f7).

> I don't like Prettier's "three or more" rule for chained methods (reason)

In real life it's not that bad.

> I don't see a value of Prettier over ESLint

Prettier is a code formatter, ESLint is to let you know more about the code, it doesn't fix the long lines for example. Let's not compare the two.

> I don't like that "lint-staged" doesn't work for modified files

That's how Git works in general.. it's called lint-staged for staged files, not modified.

> Because Prettier is opinionated it doesn't solve every preference, whereas preferences can be specified in ESLint (or implemented using ESLint Plugins)

We can combine the two as it's already how it's done in this PR.

> I don't like that Prettier can conflict with ESLint

There are no conflicts what's over because there is config for it: https://github.com/prettier/eslint-config-prettier.

> Taking everything from above into consideration, I think we are best off using ESLint (with our own shareable config) for everything related to JavaScript and JSX.

Let's not get to the conclusions as some of the points are not valid.

bennycode:
@lipis Please tell me which of my points are not valid? If "In real life it's not that bad", "You win some, you lose some" and "Let's not compare the two" are your only arguments, then I win.

The pure existence of a project like [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier) shows that both projects (Prettier & ESLint) are not compatible by default.

To me using Prettier feels like using the [Airbnb JavaScript Style Guide](https://github.com/airbnb/javascript). Only difference is that Prettier has been created by different people and that their code style is enforced by a tool.

Instead of using some other people's style guide we should define what suits us (and our team) best.

lipis:
Yes ESLint is a different thing than Prettier so that's why the one doesn't exclude the other.. and that's why we should not compare them.. but either use any or both of them. As for the rest.  with similar points that are taken from the article I can also conclude that we need it.. 

Also just saying that you don't like this rule is not an argument :)

With Prettier we don't have to ever think about formatting again..

bennycode:
@lipis: We can use [Prettier](https://prettier.io) for `.css`, `.less`, `.scss`, or `.json`.

We can also use it for `.js` and `.jsx` when it's possible to override it's settings with ESLint (in cases where it's needed). Unfortunately, it doesn't seem to be possible to override all of Prettier's styling: https://github.com/prettier/prettier-eslint-cli/issues/109

lipis:
But what's bothering you so much?! What about the rest of the team..@ffflorian @gregor @AndyTheGiant @Yserz (@herrmannplatz :P)? We already using it in the team settings :)

bennycode:
> But what's bothering you so much?!

What bothers me is the way Prettier [handles array items](https://github.com/prettier/prettier/issues/2716#issuecomment-335799469) and [chained methods](https://github.com/prettier/prettier/issues/1565#issuecomment-300133602). 
If that can be changed, I am good with Prettier. If not, ditch it in favor of `eslint --fix`.

gregor:
I am also not a fan of `prettier`. I see the point of code style. I see the point of not having to agree on a code style within the team but using a universal one - **BUT** `prettier` is too opinionated for me. I do not like at all how it handles line lenght. Just breaking stuff into multiple line lengths just does not do it for me. I am fine with `eslint` complaining about line length so that you actually improve code readibility by pulling things out into variable you compute before for example, but just forcing the `prettier` code style on top of code that has readibility issue from the beginning does not do it for me.

lipis:
that's the whole point if the long line needs to be splitted anyway in more parts we should still don't hat to increase readability.. convention over configuration..

gregor:
But with `prettier` these no longer get flagged but just pressed into some format that `prettier` opionionatedly calls desired.

lipis:
Kind of forced... of the result is bad we should refactor anyway . right?

gregor:
But it won't be flagged any longer. If you add a max line lenght in `eslint` you can set it to warn you or fail your build. A set line lenght in prettier will simply, silently reformat the code.

lipis:
We are talking for something that we have it disable actually: https://eslint.org/docs/rules/max-len

gregor:
Yes, it is disabled since it was also creating too much noise in the past. We are not discussing the current state but how we can move forward with the tools to server our needs.

lipis:
@bennyn updated

PR #1743, opened 2017-09-22
feat: Reset Mixpanel super properties on logout

bennyn:

### DIFF
--- a/app/script/tracking/EventTrackingRepository.js
+++ b/app/script/tracking/EventTrackingRepository.js
@@ -140,6 +140,19 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
         this._track_event(...args);
       }
     });
+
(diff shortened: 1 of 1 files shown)


app/script/tracking/EventTrackingRepository.js
```diff
@@ -140,6 +140,19 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
         this._track_event(...args);
       }
     });
+
+    amplify.subscribe(z.event.WebApp.LIFECYCLE.SIGN_OUT, this._reset_super_properties.bind(this));
```
gregor: Let's bind this to `z.event.WebApp.LIFECYCLE.SIGNED_OUT` instead, which is triggered when the user is actually signed out and we are about to redirect to the auth page.

bennycode: Updated. Please feel free to merge. :)

PR #1764, opened 2017-09-27
fix: Trigger emoji popup after word boundary only (WEBAPP-4327)

gregor:
## Type of change
This addresses two issues in the current design:

**Emoji popup would open if you type `this:1`**

Previous
Popup_previous

Now
Popup_now

**Word boundary is respected resulting earlier invalid match dismissal**

Previous
Boundary_previous

Now
Boundary_now

### DIFF
--- a/app/script/view_model/ConversationInputEmojiViewModel.js
+++ b/app/script/view_model/ConversationInputEmojiViewModel.js
@@ -290,14 +290,14 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
   }
 
   _update_emoji_popup(input) {
-    const text = input.value;
+    const {selectionStart: selection, value: text} = input;
     if (!text) {
       return;
     }
 
-    const query = text.substr(this.emoji_start_pos, input.selectionStart - this.emoji_start_pos);
+    const query = text.substr(this.emoji_start_pos, selection - this.emoji_start_pos);
     if (!query.length) {
-      return this.emoji_div.remove();
+      return this._close_emoji_popup();
     }
 
     if (!this.emoji_list.length || query.startsWith(' ') || /\s{2,}/.test(query)) {
--- a/app/script/view_model/ConversationInputEmojiViewModel.js
+++ b/app/script/view_model/ConversationInputEmojiViewModel.js
@@ -212,17 +212,19 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
     }
 
     const input = keyboard_event.target;
-    const {selectionStart: selection, value: text = ''} = input;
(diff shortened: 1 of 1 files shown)


gregor:
@maximbaz Wdyt?

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -286,56 +290,68 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
...
-        const pos = this._get_cursor_pixel_pos(input);
-        const top = pos.top - this.emoji_div.height() - EMOJI_LIST_OFFSET_TOP;
-        const left = pos.left - EMOJI_LIST_OFFSET_LEFT;
-
-        this.emoji_div.css('left', left);
-        this.emoji_div.css('top', top);
-      }
+    const query = text.substr(this.emoji_start_pos, input.selectionStart - this.emoji_start_pos);
+    if (!query.length) {
+      return this.emoji_div.remove();
```
max-baz: This is a tiny bug from before, but you could as well fix it now - here should be the call to `this._close_emoji_popup()`, not just removal of the `div`, so that the `click` listener is removed from the window. 

Steps to repro:
1. Put breakpoint into `remove_emoji_popup()`
2. Type `:a`, then <kbd>Backspace</kbd> to remove `a`.
3. There is no popup on the screen, now click with a mouse somewhere in the chat and see `remove_emoji_popup()` getting called.
4. Now click again on the input field, try typing `a` and see that popup will no longer appear.

app/script/view_model/ConversationInputEmojiViewModel.js
```diff
@@ -286,56 +290,68 @@ z.ViewModel.ConversationInputEmojiViewModel = class ConversationInputEmojiViewMo
...
-        this.emoji_div.find('.emoji:nth(0)').addClass('selected');
-
-        const pos = this._get_cursor_pixel_pos(input);
-        const top = pos.top - this.emoji_div.height() - EMOJI_LIST_OFFSET_TOP;
-        const left = pos.left - EMOJI_LIST_OFFSET_LEFT;
-
-        this.emoji_div.css('left', left);
-        this.emoji_div.css('top', top);
-      }
+    const query = text.substr(this.emoji_start_pos, input.selectionStart - this.emoji_start_pos);
```
max-baz: Since you need `input.selectionStart` here anyway, you could as well use this fancy way of writing on line 293 🙂
```
const {selectionStart: selection, value: text} = input;
```

gregor: Thanks for the hint. Added the fancy way in order to shorten some of the lines and remove redundancy in the writing. Tried to address another bug here but gave up on it eventually for now...

PR #1800, opened 2017-10-09
feat: Add option to manage teams (WEBAPP-4380)

gregor:
Also exposes the role to https://github.com/wireapp/wire-desktop

### DIFF
--- a/app/script/team/TeamPermission.js
+++ b/app/script/team/TeamPermission.js
@@ -1,0 +1,103 @@
+/*
+ * Wire
+ * Copyright (C) 2017 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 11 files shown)


app/script/team/TeamPermission.js
```diff
@@ -0,0 +1,103 @@
...
+  const _combine_permissions = (permissions) => {
+    let result = 0;
+    for (const permission of permissions) {
+      result = result | permission;
+    }
+    return result;
+  };
+
+  const _has_permission = (member_permissions, role) => {
+    const role_permissions = _permissions_for_role(role);
```
Yserz: Let's extract the role mapping into something like `_has_role_permission` for keeping `_has_permission` generic for tasks like single permission checks.

PR #1849, opened 2017-10-25
chore: Code cleanup

gregor:


app/script/assets/AssetCrypto.js
```diff
@@ -88,7 +88,7 @@ z.assets.AssetCrypto = (() => {
   function _generate_random_bytes(length) {
     const randomValues = new Uint32Array(length / 4).map(() => libsodium.getRandomValue());
     const ramdonBytes = new Uint8Array(randomValues.buffer);
```
lipis, line 90: I can see that we already using `camelCase` and I would recommend using them for new variables.. this change can't happen in one go.. so replacing them here and there shouldn't be a problem :)

app/script/calling/SDPMapper.js
```diff
@@ -84,43 +84,43 @@ z.calling.SDPMapper = {
...
         }
       } else if (sdp_line.startsWith('a=rtpmap')) {
-        if (
-          flow_et.negotiation_mode() === z.calling.enum.SDP_NEGOTIATION_MODE.ICE_RESTART ||
-          (sdp_source === z.calling.enum.SDP_SOURCE.LOCAL && flow_et.is_group)
-        ) {
+        const is_ice_restart = flow_et.negotiation_mode() === z.calling.enum.SDP_NEGOTIATION_MODE.ICE_RESTART;
+        const is_local_sdp = sdp_source === z.calling.enum.SDP_SOURCE.LOCAL;
+
+        if (is_ice_restart || (is_local_sdp && flow_et.is_group)) {
```
lipis, line 123: for example here.. we could maybe use the `camelCases`.. wdyt?

gregor: I have talked to @bennyn and we are going to start with this.

lipis: nice!

PR #1861, opened 2017-10-30
fix: Add type check for conversation ID

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -547,11 +552,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
       })
       .catch(error => {
         if (error.type !== z.conversation.ConversationError.TYPE.NOT_FOUND) {
-          this.logger.log(
-            this.logger.levels.ERROR,
-            `Failed to get conversation '${conversation_id}': ${error.message}`,
-            error
-          );
+          this.logger.error(`Failed to get conversation '${conversation_id}': ${error.message}`, error);
```
ffflorian, line 555: `this.logger.levels.ERROR` is not needed anymore?

gregor: No. This `this.logger.log(this.logger.levels,ERROR)` is the same as `this.logger.error()`.

lipis: Same for `info`, `warning`?

gregor: Si senor - I changed most of them a long time ago.

PR #1863, opened 2017-10-30
fix: Validate input in show conversation

gregor:


lipis:
maybe `camelCase` if you like :)

PR #1865, opened 2017-10-30
feat: New auth flow

Yserz:


app/style/auth/style.scss
```diff
@@ -0,0 +1,3 @@
+* {
+  color: blue;
+}
```
lipis: blue-man-group

app/style/auth/style.scss
```diff
@@ -0,0 +1 @@
+
```
gregor, line 1: Do we still need this file at all?

app/script/strings.js
```diff
@@ -0,0 +1,118 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {defineMessages} from 'react-intl';
+
+/* eslint-disable sort-keys */
```
bennycode, line 22: We can do better than this! :D

PR #1870, opened 2017-11-01
refactor: React + HMR + Redux

Yserz:
## Type of change
Adds React with HMR and Redux

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -5,6 +5,7 @@ env:
   node: true
 
 extends:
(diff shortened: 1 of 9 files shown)


.eslintrc.yaml
```diff
@@ -73,6 +81,9 @@ rules:
   prefer-spread: 2
   prefer-template: 2
   prettier/prettier: 2
+  react/jsx-tag-spacing: [2, {beforeSelfClosing: always}]
```
lipis: is that needed with Prettier?

PR #1875, opened 2017-11-02
fix: render markdown and linkify correctly

AndyTheGiant:


app/script/util/util.js
```diff
@@ -481,11 +471,50 @@ z.util.render_message = function(message) {
       },
     },
   };
+
   message = linkifyHtml(message, options);
 
-  // Remove this when this is merged: https://github.com/SoapBox/linkifyjs/pull/189
-  message = message.replace(/ class="linkified"/g, '');
+  console.log('linkify: ', message);
```
bennycode: unused code can be removed

app/script/util/marked.js
```diff
@@ -1,32 +1,1073 @@
...
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program. If not, see http://www.gnu.org/licenses/.
- *
- */
-
-'use strict';
+/**
```
gregor: Is this the plain marked.js code now? Then we should use it as a proper dependency.

app/script/util/util.js
```diff
@@ -454,22 +454,12 @@ z.util.markup_links = function(message) {
 
 // Note: We are using "Underscore.js" to escape HTML in the original message
 z.util.render_message = function(message) {
-  message = marked(message);
-
   // Parse links with linkifyjs library, ignore code tags
+  // Keep the type of the link for later
+  /*
```
gregor: If we no longer need the code, it should be gone.

bower.json
```diff
@@ -0,0 +1,191 @@
+{
+  "name": "wire-webapp",
```
gregor: Bower is history

app/script/util/marked.js
```diff
@@ -1,32 +1,1067 @@
...
+    this.renderer = this.options.renderer || new Renderer();
+    this.renderer.options = this.options;
+
+    if (!this.links) {
+      throw new Error('Tokens array requires a `links` property.');
+    }
+
+    if (this.options.gfm) {
+      if (this.options.breaks) {
+        this.rules = inline.breaks;
```
gregor: `this.rules = this.options.breaks ? inline.breaks : inline.gfm`

app/script/util/marked.js
AndyLnd: The changes are vast and this version now differs a lot from the marked main branch. It contains a lot of scenario specific tweaks and fixes. It wouldn't make much sense to maintain it a single file repo.

gregor: Makes sense. As discussed f2f we will go ahead and merge this after the upcoming release in order to get as much internal test coverage as we can before we release this to the public.

gregor:
@AndyTheGiant Is this goood to be merged now?

PR #1888, opened 2017-11-06
feat: Add support for stale bot

lipis:
WDYT?  

Read more on: https://github.com/probot/stale


lipis:
I'll remove comments if we're about to merge it..

gregor:
I like the idea. Discussed it f2f with @bennyn a couple of weeks ago. I think we should go for it but need to think about the configuration we want to set it to. Can we ensure only stale issues that have no been triaged as valid bugs get closed for example?

lipis:
Of course.. check the line 8.. all we need is to have some labels that are not OK to close..

.github/stale.yml
```diff
@@ -0,0 +1,23 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale Issue or Pull Request is closed
+daysUntilClose: 7
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - security
+  - "type: Bug"
```
lipis: @gregor here..

gregor: Thanks for pointing me to it

.github/stale.yml
```diff
@@ -0,0 +1,23 @@
...
+# Label to use when marking as stale
+staleLabel: wontfix
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
lipis: Also words.. :)

lipis:
If we about to merge it.. let's remove the comments and modify maybe a little bit the message and make sure that these are the labels that we want not to be staled..

lipis:
Will merge this one and we can configure it as we go.. for adjustments..

lipis:
Shall we try this little bot?

PR #1918, opened 2017-11-20
refactor: Authentication logic

Yserz:


.eslintrc.yaml
```diff
@@ -57,6 +57,7 @@ parserOptions:
   sourceType: module
   ecmaFeatures:
     jsx: true
+    experimentalObjectRestSpread: true
```
bennycode: sorting

app/script/auth/config.js
```diff
@@ -16,37 +16,8 @@
...
-
-export function getLocale() {
-  return navigator.languages && navigator.languages.length ? navigator.languages[0] : navigator.language;
-}
-
-export function currentLanguage() {
-  const LANGUAGE_SHORTHAND_LENGTH = 2;
-  return getLocale().substr(0, LANGUAGE_SHORTHAND_LENGTH) || DEFAULT_LANGUAGE;
-}
+export const APP_ENVIRONMENT = window.APP_ENVIRONMENT || Environment.PRODUCTION;
```
bennycode, line 21: No need to define `|| Environment.PRODUCTION` because it's already the default of `getEnvironment()` from `Environment.js`

app/script/auth/Environment.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import * as config from 'script/config';
+
+export const LOCAL = 'LOCAL';
+export const STAGING = 'STAGING';
+export const PRODUCTION = 'PRODUCTION';
+
+export function checkEnvironment() {
+  console.log(`Starting with environment ### ${getEnvironment()} ###`);
```
bennycode: Instead of calling `getEnvironment()` three times I would just reference the value:

```js
const environment = getEnvironment();
console.log(`Starting with environment ### ${environment} ###`);
if (![LOCAL, STAGING, PRODUCTION].includes(environment)) {
  throw new Error(`Invalid environment ${environment}`);
}
```

app/script/auth/localeConfig.js
```diff
@@ -0,0 +1,33 @@
...
+  'ru',
+  'sk',
+  'sl',
+  'tr',
+  'uk',
+];
+export const DEFAULT_LANGUAGE = 'en';
+
+export function getLocale() {
+  return navigator.languages && navigator.languages.length ? navigator.languages[0] : navigator.language;
```
bennycode, line 27: In my case:
`navigator.languages` -> `["de-DE", "en-XA", "de", "en-US", "en"]`
`navigator.language` -> `"de"`

Because of this `navigator.languages[0]` returns `de-DE` and `navigator.language` returns just `de`.

We need to make sure that we always get the 2 characters locale code (ISO 639-1).

app/script/auth/module/action/creator/AuthActionCreator.js
```diff
@@ -0,0 +1,121 @@
+/* @flow */
```
bennycode: What's the benefit of `/* @flow */` here?

app/script/auth/page/Root.js
```diff
@@ -29,8 +29,8 @@ import de from 'react-intl/locale-data/de';
 
 addLocaleData([...de]);
 
-const Root = ({locale}) => (
-  <IntlProvider locale={locale} messages={require(`../../../i18n/webapp-${locale}.json`)}>
+const Root = ({language}) => (
+  <IntlProvider locale={language} messages={require(`../../../i18n/webapp-${language}.json`)}>
```
bennycode, line 33: Let's keep the imports at the top of the file (to be consistent).

AndyLnd: I agree with you, but in this case it makes sense as the language (and therefor the name of the file to import) isn't known before.

app/script/auth/module/action/UserAction.js
```diff
@@ -0,0 +1,40 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export function doActivateAccount(code, key) {
+  const params = [...arguments];
```
AndyLnd, line 21: wouldn't it be better to be a bit more expressive here? `[code, key]`

app/script/auth/config.js
Yserz: I changed it inside `Environment `

app/script/auth/localeConfig.js
Yserz: Unfortunately that's something we have to deal with. There is a function below this one which handles just the language part though.

app/script/auth/module/action/UserAction.js
Yserz: It's fine in this case since we are just interested in the params in general not a particular one.

app/script/auth/config.js
bennycode: 👍

app/script/auth/localeConfig.js
bennycode: 👍

app/script/auth/main.js
```diff
@@ -24,8 +24,10 @@ import Root from './page/Root';
 import {AppContainer} from 'react-hot-loader';
 import {Provider} from 'react-redux';
 import configureStore from './configureStore';
+import {Account} from '@wireapp/core';
```
bennycode: Sorting 😈

app/script/auth/module/action/creator/AuthActionCreator.js
```diff
@@ -0,0 +1,120 @@
...
+  return {type: LOGOUT_SUCCESS};
+}
+
+export function failedLogout(error) {
+  return {payload: error, type: LOGOUT_FAILED};
+}
+
+export function resetError() {
+  return {type: AUTH_RESET_ERROR};
+}
```
ffflorian, line 120: Please sort variables and functions.

PR #1920, opened 2017-11-21
chore: Deployment flow

Yserz:


aws/application.py
```diff
@@ -48,9 +48,19 @@ def index():
 
 @application.route('/auth/')
 @main.latest_browser_required
+def auth_new():
```
lipis: leave that `auth`

aws/application.py
```diff
@@ -48,9 +48,19 @@ def index():
...
+  response = flask.make_response(flask.render_template(
+    'auth/index_new.html',
+    country=util.geoip_country(),
+  ))
+  return response
+
+
+@application.route('/auth/old/')
+@main.latest_browser_required
 def auth():
```
lipis: make it `auth_old`

grunt/tasks/misc.js
```diff
@@ -96,11 +97,11 @@ module.exports = grunt => {
     'copy:deploy_favicon',
     'includereplace:prod_index',
     'includereplace:prod_auth',
+    'includereplace:prod_auth_new',
     'clean:deploy_app',
     'uglify:deploy',
     'concat:deploy',
     'clean:deploy_script',
-    'raygun:prod',
```
lipis, line 103: If we actually don't need it anymore.. let's remove the whole task to clean up things

Yserz: Done

PR #1923, opened 2017-11-22
feat: added all register team pages

AndyLnd:


app/script/auth/page/CreateAccount.js
```diff
@@ -35,25 +44,24 @@ const CreateAccount = ({history}) => (
...
+                <Input placeholder={'Name'} autoFocus />
+                <Input placeholder={'you@yourcompany.com'} placeholderTextTransform="unset" />
+                <Input placeholder={'Password (min 8 characters)'} />
+              </InputBlock>
               <Checkbox>
                 <Small textTransform="uppercase">
                   {'I ACCEPT THE '}
-                  <Link href="#">{'TERMS AND CONDITIONS'}</Link>
+                  <Link href="#" bold fontSize="12px">
+                    {'TERMS AND CONDITIONS'}
```
gregor, line 56: Let's come up with a system that allows moving this link around. In some languages it is hard because you split a sentence into two and enforce the second part to be the ToUs. Let's put this into a single sentence instead where we replace with the link tag on our own. @lipis

gregor:
Approving for now but let's take a look at the translation issue in due time.

PR #1939, opened 2017-11-24
feat: hl parameter

AndyLnd:


app/script/auth/localeConfig.js
```diff
@@ -24,7 +24,10 @@ export const SUPPORTED_LANGUAGE = [
 export const DEFAULT_LANGUAGE = 'en';
 
 export function getLocale() {
-  return navigator.languages && navigator.languages.length ? navigator.languages[0] : navigator.language;
+  const languageValue = new URL(window.location).searchParams.get('hl');
```
Yserz: `getLanguage` would be a better place for this.

PR #1940, opened 2017-11-24
feat: Add Environment to auth flow

gregor:


app/script/util/Environment.js
```diff
@@ -114,8 +114,8 @@ window.z.util = z.util || {};
   };
 
   const formatted_app_version = () => {
-    const version = app_version().split('-');
-    return `${version[0]}.${version[1]}.${version[2]}.${version[3]}${version[4]}`;
+    const [year, month, day, hour, minute] = app_version().split('-');
+    return `${year}.${month}.${day}.${hour}${minute}`;
```
bennycode, line 118: very good improvement! 👍

lipis: Solving magic numbers like boss... :)

That's why is nice to have this rule..

bennycode: I took this as an example: https://vote4code.com/5715683958587392/array-destructuring 💃

lipis: use camelCase :P `formattedAppVersion`

PR #1960, opened 2017-11-27
fix: outbound links with hl param, copy changes, email placeholder

AndyLnd:


gregor:
Please point to edge branch

PR #1963, opened 2017-11-27
fix: Grant message before encryption in verified conversation (WEBAPP-4494)

gregor:


app/script/calling/CallingRepository.js
```diff
@@ -541,7 +541,7 @@ z.calling.CallingRepository = class CallingRepository {
 
       if (!event_from_stream) {
         promises.push(
-          this.conversation_repository.grant_message(conversation_id, z.ViewModel.MODAL_CONSENT_TYPE.INCOMING_CALL, [
+          this.conversation_repository.grantMessage(conversation_id, z.ViewModel.MODAL_CONSENT_TYPE.INCOMING_CALL, [
```
lipis, line 544: for the future maybe it's better to have the actual renames in a separate commit as they are hiding the actual change for this particular PR..

PR #1969, opened 2017-11-28
feat: Account form error handling

Yserz:
image


app/script/auth/module/action/ValidationError.js
```diff
@@ -0,0 +1,94 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export default class ValidationError extends Error {
+  constructor(params) {
+    super();
```
AndyLnd: `super(params)`

app/script/auth/module/action/ValidationError.js
```diff
@@ -0,0 +1,94 @@
...
+    this.name = this.constructor.name;
+    this.label = params.label;
+    this.message = params.label;
+  }
+
+  is = label => {
```
bennycode: No `static`, no `const`, what is `is`?

app/script/auth/module/action/ValidationError.js
```diff
@@ -0,0 +1,94 @@
...
+  static handleValidationState(fieldName, validationState) {
+    const field = ValidationError.getFieldByName(fieldName);
+    const validationStateKeys = ValidationError.getAllPropertyNames(validationState);
+    const errorKeys = [];
+    Object.keys(ValidationError.ERROR).forEach(errorName => errorKeys.push(ValidationError.ERROR[errorName]));
+    for (const key of validationStateKeys) {
+      if (errorKeys.includes(key) && validationState[key]) {
+        return new ValidationError({
+          label: field[ValidationError.getErrorByKey(key)],
+        });
```
gregor: This looks like you can just to a `map` on `Object.keys` or with the help of babel even on `Object.entries()` to full `errorKeys` immeditately without all these validations.

app/script/auth/module/action/ValidationError.js
Yserz: It's a convenience method to check if an instance of an error has a given label.
`if(someValidationError.is(ValidationError.NAME.PATTERN_MISMATCH)) ...`

PR #1978, opened 2017-11-28
chore: Notify about edge deployments

bennyn:

### DIFF
--- a/bin/deployment_notification.js
+++ b/bin/deployment_notification.js
@@ -42,16 +42,16 @@ const content = {
 
 switch (commit.branch) {
   case 'dev':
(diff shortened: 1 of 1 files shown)


bin/deployment_notification.js
```diff
@@ -44,6 +44,9 @@ switch (commit.branch) {
   case 'dev':
     build.url = 'https://wire-webapp-dev.zinfra.io/auth/?env=prod#login';
     break;
+  case 'edge':
+    build.url = 'https://wire-webapp-edge.zinfra.io/auth/?env=staging#login';
```
gregor: This should point to `https://wire-webapp-edge.zinfra.io/login/?env=staging#login` instead for convenience - with the others being adjusted.

PR #1985, opened 2017-11-29
fix: Add polyfill for SearchParams (WEBAPP-4505)

Yserz:


package.json
```diff
@@ -48,16 +48,17 @@
...
     "react-dom": "16.1.1",
-    "react-intl": "2.4.0",
     "react-intl-po": "2.1.1",
+    "react-intl": "2.4.0",
     "react-redux": "5.0.6",
-    "react-router": "4.2.0",
     "react-router-dom": "4.2.2",
-    "redux": "3.7.2",
+    "react-router": "4.2.0",
+    "react": "16.1.1",
```
gregor: I guess we don't really agree on the alphabet here.

lipis: Will post an issue on VS Code.. `-` < `"`

PR #1988, opened 2017-11-29
Remove ‘on’ from Edited/Deleted timestamps

infotexture:
Since timestamps are dynamic, they may appear as either date or time values.

‘On’ works for date values, but not for time values, where ‘at’ would be correct.

Rather than creating 2 separate strings to handle these grammar differences, this commit removes the preposition entirely, so the same text can be used for both cases.


lipis:
We need the translations as well

gregor:
For us this is a tool tip on hover that will always show a full date. But I agree that it is easier to maintain moving forward.

gregor:
I will go ahead an merge this for now. Translations will follow with next sync.

PR #1993, opened 2017-11-29
fix: Show general invite message on registration form (WEBAPP-4501)

gregor:
https://github.com/wireapp/wire-account/pull/28


gregor:
Current behavior is as designed. Thereby closing this one.

PR #2001, opened 2017-11-30
fix: changed font size on index page

AndyLnd:


app/script/auth/page/Index.js
```diff
@@ -96,7 +96,7 @@ class Index extends Component {
                 {_(indexStrings.createTeam)}
               </Bold>
               <br />
-              <Text light fontSize="24px" color={COLOR.LINK}>
+              <Text light fontSize="16px" color={COLOR.LINK} style={{lineHeight: '36px'}}>
```
lipis, line 90: We are back to the inline styling.. what's the policy in general! It's the same inline rule for both elements.. back in the good old days in 2017 we used to solve them with classes.. what's the deal here?

AndyLnd: The general policy is that smaller style adjustments (stuff like margins or paddings) may be done via inline styling, as it is easy to understand what is happening here (as opposed to classNames like "link-block__text--small" or sth.). If it starts to be used in more places than just these two or the inline styles become ridiculously unreadable, it will become a new component.

PR #2004, opened 2017-11-30
fix: added silent logout, updated error message strings

AndyLnd:


app/script/auth/module/action/creator/AuthActionCreator.js
```diff
@@ -41,6 +41,7 @@ export const REGISTER_JOIN_FAILED = 'REGISTER_JOIN_FAILED';
 export const LOGOUT_START = 'LOGOUT_START';
 export const LOGOUT_SUCCESS = 'LOGOUT_SUCCESS';
 export const LOGOUT_FAILED = 'LOGOUT_FAILED';
+export const SILENT_LOGOUT_SUCCESS = 'LOGOUT_FAILED';
```
Yserz: String should be `SILENT_LOGOUT_SUCCESS ` as well

Yserz:
The copy changes are from design.

PR #2015, opened 2017-12-01
refactor: Using platform module where possible

bennyn:
## Changes

1. In `app/script/util/Environment.js` we are using `window.platform` and `navigator.platform`. To prevent confusion between these two I prefixed them (either with `window` or with `navigator`).
2. I replaced some `navigator.platform` calls in favor of `window.platform` because `window.platform` is defined by [platform.js](https://github.com/bestiejs/platform.js) and platform.js is a superset of `navigator`, so there is no need to keep the pure `navigator`.
3. I made `RuntimeUtil.js` independant of `app/script/util/Environment.js` because on the long run `RuntimeUtil.js` will replace `app/script/util/Environment.js` (when moving fully to React). The short-term benefit is, that testing React components becomes much easier because the `RuntimeUtil.js` imports only [platform](https://www.npmjs.com/package/platform) and thus doesn't have any dependencies to `window` or `navigator`.

### DIFF
--- a/app/script/auth/util/RuntimeUtil.js
+++ b/app/script/auth/util/RuntimeUtil.js
@@ -50,7 +50,7 @@ export default class RuntimeUtil {
   }
 
   static getPlatform() {
-    if (isDesktop()) {
+    if (RuntimeUtil.isDesktop()) {
       if (platform.os.family.includes(RuntimeUtil.PLATFORM_NAME.WINDOWS)) {
         return RuntimeUtil.PLATFORM_TYPE.DESKTOP_WINDOWS;
       }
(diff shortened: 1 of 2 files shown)


app/script/auth/util/RuntimeUtil.js
```diff
@@ -1,19 +1,60 @@
...
+    EDGE: 'Microsoft Edge',
+    ELECTRON: 'Electron',
+    FIREFOX: 'Firefox',
+    OPERA: 'Opera',
+    WIRE: 'Wire',
+  };
+
+  static PLATFORM_NAME = {
+    MACINTOSH: 'Mac',
+    WINDOWS: 'Win',
```
lipis, line 34: `Windows`, `macOS`

bennycode: No, the code would fail if I do that, because macOS is not part of the User Agent string. I tried to be compatible to our current code.

lipis: makes sense then :)

PR #2024, opened 2017-12-04
chore: Jest testing setup

bennyn:


.babelrc
```diff
@@ -17,10 +17,39 @@
...
+    "test": {
+      "presets": [
+        "jest",
+        "react",
+        [
+          "env",
+          {
+            "debug": false,
+            "modules": "commonjs",
+            "targets": {
```
Yserz, line 41: Targets is duplicated now. Is it possible to extract that?

bennycode: It overrides the first target.

PR #2029, opened 2017-12-04
chore: Disable Prettier for auto-generated files

bennyn:

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+/package.json
+/yarn.lock
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,1 @@
-/package.json
 /yarn.lock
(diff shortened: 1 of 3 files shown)


.prettierignore
```diff
@@ -0,0 +1,2 @@
+/package.json
+/yarn.lock
```
lipis: yarn.lock is not being formatted anyway by Prettie.. but I don't agree that `package.json` is autogenerated..

lipis:
Why is that?!

bennycode:
@lipis I removed `/package.json` from `.prettierignore`. Fine now?

.prettierignore
```diff
@@ -0,0 +1 @@
...
```
lipis, line 1: this file will never be prettified anyway.. it's not supported.

bennycode: ok

lipis:
sorry man.. yeah.. it's fine now! :P

PR #2034, opened 2017-12-05
chore: Upgrade Prettier to 1.9

lipis:


.prettierrc.yaml
```diff
@@ -1,7 +1,8 @@
+arrowParens: avoid
```
lipis, line 1: Now there is an options if we the parenthesis back :)

gregor: I definitely want them

lipis: The we have to take a vote I guess :)

https://vote4code.com/5764748591235072/arrow-function-parentheses#red

PR #2042, opened 2017-12-07
feat: choose handle (WEBAPP-4490)

AndyLnd:


app/script/auth/module/action/SelfAction.js
```diff
@@ -32,6 +33,23 @@ export function fetchSelf() {
...
+        throw BackendError.handle(error);
+      });
+  };
+}
+
+export function setHandle(handle) {
+  return function(dispatch, getState, {apiClient}) {
+    dispatch(SelfActionCreator.startSetHandle());
+    return apiClient.self.api
+      .putHandle({handle: handle.toLowerCase()})
```
gregor: Is that all the normalization we need? What about trimming to remove whitespaces?

lipis: if more _complicated_ maybe have a normalisation function for the `handle`?

PR #2044, opened 2017-12-07
fix: better form behavior

AndyLnd:


app/script/auth/module/reducer/authReducer.js
```diff
@@ -122,6 +123,12 @@ export default function reducer(state = initialState, action) {
     case AuthActionCreator.ENTER_PERSONAL_CREATION_FLOW: {
       return {...state, isInTeamFlow: false};
     }
+    case UserActionCreator.USER_SEND_ACTIVATION_CODE_SUCCESS: {
```
Yserz: In order to set `fetching` to false in an error case there should be an `USER_SEND_ACTIVATION_CODE_FAILED `

app/script/auth/module/selector/InviteSelector.js
```diff
@@ -19,3 +19,4 @@
 
 export const getInvites = state => state.inviteState.invites;
 export const getError = state => state.inviteState.error;
+export const getFetching = state => state.inviteState.fetching;
```
Yserz: `isFetching`

PR #2051, opened 2017-12-11
Workaround for webdriver

mythsunwind:
## Type of change

window.location will be replaced with a workaround for selenium webdriver. Otherwise the chromedriver gets stuck with "Timed out receiving message from renderer", because the renderer process gets unattached.

### DIFF
--- a/app/script/auth/page/Verify.js
+++ b/app/script/auth/page/Verify.js
@@ -55,7 +55,7 @@ const Verify = ({account, authError, history, isInTeamFlow, intl: {formatMessage
         })
         .then(() => {
           const link = document.createElement('a');
-          link.href = pathWithParams('/login', 'reason=registration');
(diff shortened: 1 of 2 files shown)


app/script/auth/page/Verify.js
```diff
@@ -53,7 +53,11 @@ const Verify = ({account, authError, history, isInTeamFlow, intl: {formatMessage
           connected.trackEvent({name: TrackingAction.EVENT_NAME.PERSONAL.CREATED});
           connected.trackEvent({name: TrackingAction.EVENT_NAME.PERSONAL.VERIFIED});
         })
-        .then(() => (window.location = pathWithParams('/login', 'reason=registration')))
+        .then(() => {
+          const link = document.createElement('a');
+          link.href = pathWithParams('/login', 'reason=registration');
```
bennycode: Can `ROUTE.LOGIN` be used here?

PR #2056, opened 2017-12-12
fix: Make login link clickable in Firefox again

mythsunwind:
## Type of change

Adds another workaround on top of the workaround for the desktop application test framework. The former workaround broke the navigation in Firefox.


lipis:
giphy

PR #2071, opened 2017-12-13
fix: better - future proof - register flow track keeping

AndyLnd:


app/script/auth/page/Verify.js
```diff
@@ -28,38 +28,48 @@ import {parseError} from '../util/errorUtil';
...
+            connected.trackEvent({name: TrackingAction.EVENT_NAME.PERSONAL.VERIFIED});
+          })
+          .then(() => {
+            const link = document.createElement('a');
+            link.href = pathWithParams(ROUTE.LOGIN, 'reason=registration');
+            document.body.appendChild(link); // workaround for Firefox
+            link.click();
+          })
+          .catch(error => console.error('Failed to create personal account', error));
+      }
```
Yserz, line 72: Do we need a default here?

PR #2102, opened 2017-12-18
fix: Availability tweaks (WEBAPP-4556)

gregor:


app/script/user/UserRepository.js
```diff
@@ -568,18 +568,21 @@ z.user.UserRepository = class UserRepository {
     });
   }
 
-  changeAvailability(availability, method) {
+  setAvailability(availability, method) {
```
bennycode, line 571: The parameter name `method` needs to become a little more declarative (just saw it now 🙈 ).

gregor: Yeah, Any suggestions? Just used the name that was specified in the event we use this in.

PR #2124, opened 2017-12-19
chore: Migrate /team to camelCase

gregor:


app/script/user/UserRepository.js
```diff
@@ -1065,7 +1065,7 @@ z.user.UserRepository = class UserRepository {
   }
 
   map_guest_status(user_ets = this.users()) {
-    const team_members = this.team_members();
+    const team_members = this.teamMembers();
```
lipis, line 1068: maybe the `const` as well?

gregor: Will happen when I refactor /users

lipis: right..

lipis: just saw that it's in the `/user` makes sense..

PR #2130, opened 2017-12-20
fix: Translate goal failing to build pot file on Windows

gregor:


bennycode:
When dealing with longer strings that might have special characters, then we can also make use of the following pattern:
```json
"rip json2pot \"temp/i18n/**/*.json\" -o \"temp/i18n/webapp.pot\""
```

PR #2135, opened 2017-12-20
test: TeamName testcases

Yserz:


app/script/auth/page/TeamName.js
```diff
@@ -40,9 +40,9 @@ const wireTeamLink = `${ROUTE.WIRE_ROOT}/create-team/#features`;
 
 class TeamName extends Component {
   state = {
-    enteredTeamName: this.props.teamName,
+    enteredTeamName: this.props.teamName || '',
     error: null,
-    isValidTeamName: false,
+    isValidTeamName: !!this.props.teamName,
```
lipis, line 45: Even though I like the `!!`, maybe `Boolean(this.props.teamName)` is better

PR #2138, opened 2017-12-22
fix: user name cut off with emoji and state icon

AndyLnd:
Please check on windows as well, on mac it looks fine now.


app/style/list/conversation-list-cell.less
```diff
@@ -59,7 +59,7 @@
 /* elements */
 .conversation-list-cell-availability {
   .availability-state-icon {
-    margin-right: 8px;
+    margin-right: 1px 8px 1px 0;
```
mat1th: This is not possible it should be `margin: 1px 8px 1px 0;`

AndyLnd: of course. 
thanks!

PR #2189, opened 2018-01-10
chore: Query param constants

Yserz:


app/script/auth/localeConfig.js
```diff
@@ -5,15 +5,18 @@ export const SUPPORTED_LANGUAGE = require('./supportedLocales');
 export const DEFAULT_CURRENCY = 'eur';
 export const DEFAULT_LANGUAGE = 'en';
 
+const QUERY_LANGUAGE = 'hl';
+const QUERY_CURRENCY = 'currency';
```
gregor: Looks fine. I would probably do an object which contains both though.

PR #2215, opened 2018-01-15
feat: Add more service avatar sizes

AndyLnd:


app/style/components/service-avatar.less
```diff
@@ -46,10 +46,47 @@ service-avatar {
...
     border-radius: 15%;
+    overflow: hidden;
+    opacity: 0;
+
+    > img {
+      width: 100%;
+      height: 100%;
+    }
+
+    .user-avatar-image-loaded & {
```
gregor: Let's name this appropriately as well

gregor: Which I just did to merge this now.

PR #2226, opened 2018-01-16
feat: Service list compact and default mode

AndyLnd:


app/script/components/serviceList.js
```diff
@@ -22,26 +22,50 @@
 window.z = window.z || {};
 window.z.components = z.components || {};
 
+z.components.ServiceListMode = {
+  COMPACT: 'compact',
+  DEFAULT: 'default',
+  OTHERS: 'others',
```
gregor: No need for others at this time

PR #2230, opened 2018-01-17
feat: First time UX - No conversation hint (WEBAPP-4593)

Yserz:
image


app/page/template/list/conversations.htm
```diff
@@ -59,8 +59,10 @@
     <!-- ko if: conversations_unarchived().length === 0 && connect_requests().length === 0 && conversations_calls().length === 0 -->
       <!-- ko if: conversations_archived().length === 0 -->
         <div class="conversation-hint" data-uie-name="status-start-conversation-hint">
-          <div class="conversation-hint-header" data-bind="l10n_text: z.string.conversations_no_conversations"></div>
-          <div class="conversation-hint-detail" data-bind="l10n_text: z.string.conversations_no_conversations_detail"></div>
+          <div class="conversation-hint-text" data-bind="l10n_text: z.string.conversations_no_conversations"></div>
+          <svg style="opacity: 0.4;" width="16" height="53">
```
gregor: Let's remove the inline style

app/page/template/list/conversations.htm
```diff
@@ -59,8 +59,10 @@
     <!-- ko if: conversations_unarchived().length === 0 && connect_requests().length === 0 && conversations_calls().length === 0 -->
```
gregor: Can we turn this into an observable in the view model while we are at it?

app/script/localization/webapp.js
```diff
@@ -45,7 +45,8 @@ z.string.auth_account_sign_in_phone = 'Phone';
 
 // Authentication: BLOCKED section
 z.string.auth_blocked_cookies = 'Enable cookies to log in to Wire.';
-z.string.auth_blocked_database = 'Wire needs access to local storage to display your messages. Local storage is not available in private mode.';
+z.string.auth_blocked_database =
+  'Wire needs access to local storage to display your messages. Local storage is not available in private mode.';
```
gregor: Please no prettier changes to the strings file 🙈

lipis: https://github.com/wireapp/wire-webapp/blob/dev/.prettierrc.yaml maybe we can add another override for these files with huge printWidth

.prettierrc.yaml
```diff
@@ -14,3 +14,9 @@ overrides:
 - files: "*.json"
   options:
     printWidth: 200
+- files: "app/script/localization/webapp.js"
+  options:
+    printWidth: 200
+- files: "app/script/localization/translations/*.js"
```
lipis: if it's the same 200 we could combine them with regular expressions

.prettierrc.yaml
```diff
@@ -14,3 +14,9 @@ overrides:
 - files: "*.json"
   options:
     printWidth: 200
+- files: "app/script/localization/webapp.js"
```
gregor: Does files allow an array?

app/page/template/list/conversations.htm
```diff
@@ -56,11 +56,13 @@
...
     </ul>
     <!-- /ko -->
-    <!-- ko if: conversations_unarchived().length === 0 && connect_requests().length === 0 && conversations_calls().length === 0 -->
+    <!-- ko if: conversations_empty() -->
       <!-- ko if: conversations_archived().length === 0 -->
         <div class="conversation-hint" data-uie-name="status-start-conversation-hint">
-          <div class="conversation-hint-header" data-bind="l10n_text: z.string.conversations_no_conversations"></div>
-          <div class="conversation-hint-detail" data-bind="l10n_text: z.string.conversations_no_conversations_detail"></div>
+          <div class="conversation-hint-text" data-bind="l10n_text: z.string.conversations_no_conversations"></div>
+          <svg class="conversation-hint-arrow" width="16" height="53">
```
gregor: Width and height should also move to the CSS

.prettierrc.yaml
Yserz: Luckily it does 👍

PR #2231, opened 2018-01-18
feat: Add URL param to toggle integrations

gregor:


app/script/view_model/list/StartUIViewModel.js
```diff
@@ -140,23 +140,22 @@ z.ViewModel.list.StartUIViewModel = class StartUIViewModel {
...
 
-    this.enableIntegrations = ko.pureComputed(
-      () => true || (this.is_team() && !z.util.Environment.frontend.is_production())
-    );
+    this.enableIntegrations = this.integrationRepository.enableIntegrations;
 
-    this.show_content = ko.pureComputed(
-      () => this.show_contacts() || this.show_matches() || this.show_search_results()
-    );
+    this.show_content = ko.pureComputed(() => {
```
AndyLnd, line 145: why not the implicit return?

app/script/integration/IntegrationRepository.js
```diff
@@ -35,13 +35,25 @@ z.integration.IntegrationRepository = class IntegrationRepository {
...
 
     this.conversationRepository = conversationRepository;
+    this.teamRepository = teamRepository;
+
+    this.isTeam = this.teamRepository.isTeam;
     this.services = ko.observableArray([]);
+
+    this.supportIntegrations = ko.observable();
+    this.enableIntegrations = ko.pureComputed(() => {
+      if (Boolean(this.supportIntegrations())) {
```
AndyLnd: This would make it possible to enable integrations regardless of environment, even on production. Is this intended?

app/script/view_model/list/StartUIViewModel.js
gregor: Because prettier already turns it into a multiline thing. If code does not make the one line cutoff, I prefer this code style.

PR #2235, opened 2018-01-18
feat: participant avatar

AndyLnd:


app/script/components/participantAvatar.js
```diff
@@ -0,0 +1,152 @@
...
+ *
+ */
+
+'use strict';
+
+window.z = window.z || {};
+window.z.components = z.components || {};
+
+z.components.ParticipantAvatar = class ParticipantAvatar {
+  constructor(params, component_info) {
```
gregor: camelCase

app/script/components/participantAvatar.js
```diff
@@ -0,0 +1,152 @@
...
+    this.participant = ko.unwrap(params.participant);
+    this.isUser = this.participant instanceof z.entity.User && !this.participant.isBot;
+    this.isService = this.participant instanceof z.integration.ServiceEntity || this.participant.isBot;
+
+    this.delay = params.delay;
+    this.size = params.size;
+    this.element = $(component_info.element);
+
+    if (this.size) {
+      this.element.addClass(`${this.isService ? 'service' : 'user'}-avatar-${this.size}`);
```
gregor: how many things depend on this CSS class? Can we move from individual service and user classes to particpant classes for the sizing or do other things depend on it?

app/script/components/participantAvatar.js
```diff
@@ -0,0 +1,152 @@
...
+          case this.participant.is_team_member():
+            return '';
+          case this.participant.is_blocked():
+            return 'blocked';
+          case this.participant.is_request():
+            return 'pending';
+          case this.participant.is_ignored():
+            return 'ignored';
+          case this.participant.is_canceled() || this.participant.is_unknown():
+            return 'unknown';
```
gregor, line 72: If we are at it we can turn these into an enum

app/style/components/service-avatar.less
```diff
@@ -17,7 +17,8 @@
  *
  */
 
-service-avatar {
+service-avatar,
+participant-avatar {
```
gregor, line 21: Let's cleanup the CSS classes and remove the outdated service and user avatars in the next step

app/script/components/participantAvatar.js
AndyLnd: the less files will definitely be merged later on.

app/style/components/service-avatar.less
AndyLnd: Most definitely!

app/script/components/participantAvatar.js
AndyLnd: They are just used here. So there is no typo hazard.

PR #2250, opened 2018-01-19
fix: Do not log value before calculating (WEBAPP-4609)

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -2236,9 +2236,9 @@ z.conversation.ConversationRepository = class ConversationRepository {
       })
       .then(() => this.send_asset_remotedata(conversation_et, file, message_id))
       .then(() => {
+        const upload_duration = (Date.now() - upload_started) / 1000;
```
AndyLnd, line 2239: You could fix the snake_case here.

PR #2276, opened 2018-01-23
fix: better avatar behavior (WEBAPP-4621)

AndyLnd:


app/script/components/participantAvatar.js
```diff
@@ -23,13 +23,14 @@ window.z = window.z || {};
...
+  static get SIZE() {
     return {
-      LARGE: 'ParticipantAvatar.SIZE.LARGE',
-      MEDIUM: 'ParticipantAvatar.SIZE.MEDIUM',
-      SMALL: 'ParticipantAvatar.SIZE.SMALL',
-      X_SMALL: 'ParticipantAvatar.SIZE.X_SMALL',
-      XX_SMALL: 'ParticipantAvatar.SIZE.XX_SMALL',
+      LARGE: 'avatar-lg',
+      MEDIUM: 'avatar-md',
+      MEDIUMSMALL: 'avatar-sm',
```
gregor: Let's remove the obsolete size we do not use

PR #2286, opened 2018-01-23
fix: Trim spaces around user name in registration (WEBAPP-4604, WEBAPP-4618)

Yserz:


gregor:
Looks like we need to pull translations for this one.

PR #2296, opened 2018-01-24
fix: Add trailing slashes to ToU links

gregor:


lipis:
It's to have one redirect less :)

PR #2307, opened 2018-01-24
fix: Change capitalization of log in

gregor:


gregor:
As requested by @Yorgos-V

PR #2327, opened 2018-01-29
fix: remove button only shown when allowed, service bubble animations (WEBAPP-4634, WEBAPP-4636)

AndyLnd:


app/page/template/content/conversation/participants.htm
```diff
@@ -120,15 +119,19 @@
...
             <button data-uie-name="do-service-remove" class="participants-service-confirmation-button button-theme-red" data-bind="click: clickToRemoveService, l10n_text: z.string.people_service_removal_button"></button>
-          <!-- /ko -->
-        </div>
+          </div>
+        <!-- /ko -->
       </div>
-    <!-- /ko -->
+      </div>
+      <!-- /ko -->
+    </div>
```
gregor: Indentation is definitely off in this file

app/script/view_model/ParticipantsViewModel.js
```diff
@@ -189,6 +192,11 @@ z.ViewModel.ParticipantsViewModel = class ParticipantsViewModel {
       return this.stateParticipants() && this.selectedUser() && !this.selectedService();
     });
 
+    this.selectedIsInConversation = ko.observable(false);
```
gregor: This should be a computed. Otherwise the UI will not change when the group participants change.

PR #2337, opened 2018-01-30
fix: guests should not be able to remove bots (WEBAPP-4634)

AndyLnd:


app/script/view_model/ParticipantsViewModel.js
```diff
@@ -199,7 +199,7 @@ z.ViewModel.ParticipantsViewModel = class ParticipantsViewModel {
       return false;
     });
     this.showServiceRemove = ko.pureComputed(() => {
-      return this.stateServiceDetails() && !this.userRepository.self().is_guest() && this.selectedIsInConversation();
+      return this.stateServiceDetails() && !this.conversation().is_guest() && this.selectedIsInConversation();
```
gregor, line 202: Good point

PR #2339, opened 2018-01-30
fix: unicode response header probably broke the server

AndyLnd:


lipis:
and I thought about..! :)

PR #2351, opened 2018-01-31
feat: light background for start ui header

AndyLnd:


app/style/list/start-ui.less
```diff
@@ -31,6 +31,7 @@
 }
 
 .start-ui-header {
+  background-color: rgba(255, 255, 255, 0.08);
```
bennycode, line 34: Value is used in two places. Let's make a variable for that.

lipis, line 34: `fade(#fff, 8%)`

app/style/list/list.less
```diff
@@ -51,6 +51,9 @@
   height: @left-list-header-height;
   position: relative;
 }
+.left-list-header-start-ui {
+  background-color: rgba(255, 255, 255, 0.08);
```
lipis, line 55: `fade(#fff, 8%)`

bennycode: @Lipis: I prefer `background-color: rgba(255, 255, 255, 0.08)` over `fade(#fff, 8%)` because it's not less-specific and thus doesn't need to get rewritten if we switch to Sass or anything else.

lipis: If we ever going to switch to SASS we will have to use some sort of tool that does that for us anyway as we already using it in quite some places and since we are still using Less, we should embrace those..

PR #2357, opened 2018-02-01
chore: Update demo page

gregor:


app/demo/demo.html
```diff
@@ -130,47 +128,31 @@ <h1>Dots</h1>
...
           .addClass('theme-' + item.text())
-          .addClass('main-accent-color-' + (item.index('li') + 1))
+          .addClass('main-accent-color-' + (item.index('li') + 1));
       });
 
       let previousScroll = 0;
-      $(window).scroll(function(event){
-         let scroll = $(this).scrollTop();
+      $(window).scroll(() => {
+         const scroll = $(this).scrollTop();
```
AndyLnd: ```this``` will not have the right context ... or maybe it will. there is no way to know for sure.

PR #2358, opened 2018-02-01
fix: checkbox change shouldn't scroll to top (WEBAPP-4214)

AndyLnd:


gregor:
Let's revisit this: https://medium.com/claritydesignsystem/pure-css-accessible-checkboxes-and-radios-buttons-54063e759bb3

PR #2360, opened 2018-02-01
fix: Align error copy in case of 409 (WEBAPP-4647)

gregor:


gregor:
Let's see if this will actually be the final expected copy or if we can adjust it to be more descriptive and explanatory.

gregor:
After a longer discussion of everyone involved (and some others) we now have a copy and this is ready for review.

PR #2373, opened 2018-02-06
fix: Show member invite even if guests are shown (WEBAPP-4625)

Yserz:


app/script/view_model/list/StartUIViewModel.js
```diff
@@ -151,9 +151,7 @@ z.ViewModel.list.StartUIViewModel = class StartUIViewModel {
 
     this.showNoContacts = ko.pureComputed(() => !this.isTeam() && !this.showContent());
     this.showInviteMember = ko.pureComputed(() => {
-      return (
-        this.selfUser().isTeamOwner() && this.teamSize() === 1 && !this.showContacts() && !this.showSearchResults()
-      );
+      return this.selfUser().isTeamOwner() && this.teamSize() === 1 && !this.showSearchResults();
```
AndyLnd: Does this really need an explicit return?

gregor: Yes. IMHO it is much more readable this way if you have to break it up in multiple lines anyway.

app/script/view_model/list/StartUIViewModel.js
```diff
@@ -135,11 +135,10 @@ z.ViewModel.list.StartUIViewModel = class StartUIViewModel {
...
 
     // View states
-    this.hasSearchResults = ko.pureComputed(() => {
-      return (
+    this.hasSearchResults = ko.pureComputed(
+      () =>
         this.searchResults.groups().length || this.searchResults.contacts().length || this.searchResults.others().length
-      );
-    });
+    );
```
gregor, line 141: I am definitely not a fan of these. If we need multiple lines anyway we should go for explicit returns. Now I have to check if this actually returns somthing...

PR #2377, opened 2018-02-06
feat: Add group creation flow (WEBAPP-4650 - WEBAPP-4658, WEBAPP-4660)

gregor:


app/style/common/badge.less
```diff
@@ -52,7 +52,7 @@
 
 .cell-badge-dark {
   .cell-badge;
-  background-color: fade(#000, 16%);
+  background-color: #fff;
```
gregor: Causes badges in conversation list to be unreadable

PR #2402, opened 2018-02-08
fix: User profile in participants view fails to show (WEBAPP-4667)

gregor:


app/page/template/content/conversation/participants.htm
```diff
@@ -59,8 +59,8 @@
             pending: clickOnPending,
             profile: clickOnSelfProfile,
             remove: clickToRemoveMember,
-            unblock: clickToUnblock
-            user: selectedParticipant,
+            unblock: clickToUnblock,
+            user: selectedParticipant
```
gregor, line 63: The actual fix 🙈

PR #2408, opened 2018-02-09
fix: avatar sizes, "create group" height, 

AndyLnd:


app/style/components/search-list.less
```diff
@@ -55,6 +55,9 @@
     flex: 0 0 auto;
     height: @avatar-diameter-m;
     position: relative;
+    height: 40px;
```
gregor: remove the old hight just two lines above

PR #2433, opened 2018-02-12
fix: Prevent long wait on participant step

gregor:


app/script/view_model/content/GroupCreationViewModel.js
```diff
@@ -87,6 +89,12 @@ z.ViewModel.content.GroupCreationViewModel = class GroupCreationViewModel {
       }
       return $(document).off('keydown.groupCreation');
     });
+    this.stateIsParticipants.subscribe(stateIsParticipants => {
+      if (stateIsParticipants) {
+        return window.setTimeout(() => this.showContacts(true));
```
AndyLnd, line 95: `setTimeout` -> `requestAnimationFrame`
Might be a little bit more expressive.

gregor: I tried. Behavior is differently then and it does not address the problem. Will stick to this.

PR #2492, opened 2018-02-16
Migrate /conversation to camelCase

HemangiS:
Migrate /conversation to camelCase


gregor:
@HemangiS Thank you very much for your willingness to put in more effort to also migrate the `conversation` folder. We will have to make heavy changes in the code your are migrating in the coming days and weeks for the next features. Thus it will probably be really hard to keep up with all our changes. In order to not waste your time I would advise you to refrain from continueing at this point in time. We will probably change parts of the code to camelCase ourselves with the new code we need already.

PR #2559, opened 2018-02-26
chore: Code cleanup group creation

gregor:


app/script/components/userList.js
```diff
@@ -75,9 +75,9 @@ z.components.UserList = class UserList {
           const trimmedQuery = this.filter().trim();
           const isHandle = trimmedQuery.startsWith('@') && z.user.UserHandleGenerator.validate_handle(normalizedQuery);
           const excludedEmojis = Array.from(normalizedQuery).filter(char => EMOJI_UNICODE_RANGES.includes(char));
-          return this.userEntities().filter(userEntity =>
-            userEntity.matches(normalizedQuery, isHandle, excludedEmojis)
-          );
+          return this.userEntities().filter(userEntity => {
+            return userEntity.matches(normalizedQuery, isHandle, excludedEmojis);
```
AndyLnd, line 79: Why more overhead?

app/script/view_model/content/GroupCreationViewModel.js
```diff
@@ -54,11 +54,9 @@ z.ViewModel.content.GroupCreationViewModel = class GroupCreationViewModel {
 
     this.activateNext = ko.pureComputed(() => this.nameInput().length);
     this.contacts = ko.pureComputed(() => {
-      if (this.showContacts()) {
-        return this.teamRepository.isTeam() ? this.teamRepository.teamUsers() : this.userRepository.connected_users();
-      }
-
-      return [];
+      return this.showContacts()
+        ? this.teamRepository.isTeam() ? this.teamRepository.teamUsers() : this.userRepository.connected_users()
```
AndyLnd: IMO that is not really more readable.

app/script/components/userList.js
gregor: It's not overhhead. If we do not do an explicit return in this case you need to read two lines to see that something is returned due to it being split into multiple lines by prettier. With this you immediately see what's happening.

app/script/view_model/content/GroupCreationViewModel.js
lipis: I agree.. too many `?`, `:`

PR #2565, opened 2018-02-26
feat: Use Cryptobox from npm

ffflorian:
Closes #2676.


ffflorian:
Waiting for QA tests before merge.

app/script/user/UserRepository.js
```diff
@@ -705,7 +705,7 @@ z.user.UserRepository = class UserRepository {
         return this.save_user(user_et, true);
       })
       .catch(error => {
-        this.logger.error(`Unable to load self user: ${error}`);
+        this.logger.error(`Unable to load self user: ${error}`, error);
```
gregor: Seems like error was a string in this case. Adding it after the colon will cause the rest of the error message to disappear. If the error can be an error object or a string it would be better to do `this.logger.error(`Unable to load self user: ${error.message || error}`);`

grunt/tasks/npmCopy.js
```diff
@@ -0,0 +1,48 @@
...
+      moduleFiles.forEach(file => {
+        const from = path.join('node_modules', module, file);
+        const to = path.join(distPath, 'js', module, file);
+        grunt.file.copy(from, to);
+        grunt.log.writeln(chalk`Copied file from "{blue ${from}}" to "{blue ${to}}".`);
+      });
+    });
+
+    grunt.log.ok(chalk`Copied npm components from "{blue node_modules}" to "{blue ${distPath}}".`);
+  });
```
gregor: Why don't we use the grunt copy task we already have in place for this?

PR #2569, opened 2018-02-26
feat: Filter users on group creation

gregor:


app/script/view_model/panel/ParticipantsViewModel.js
```diff
@@ -163,13 +163,21 @@ z.viewModel.panel.ParticipantsViewModel = class ParticipantsViewModel {
...
-        .filter(userEntity => !this.participants().find(participant => userEntity.id === participant.id))
-        .sort((userA, userB) => z.util.StringUtil.sort_by_priority(userA.first_name(), userB.first_name()));
+      if (!this.isTeam()) {
+        userEntities = this.userRepository.connected_users();
+      }
+
+      if (this.isTeamOnly()) {
+        userEntities = this.teamMembers().sort((userA, userB) => {
+          return z.util.StringUtil.sort_by_priority(userA.first_name(), userB.first_name());
+        });
```
ffflorian, line 175: Not necessary, but would be consistent:

```js
userEntities = this.teamMembers().sort((userA, userB) =>
  z.util.StringUtil.sort_by_priority(userA.first_name(), userB.first_name())
);
```

gregor: No, I really dislike that. You need to read at least two lines in order to notice if something is being returned. This is more explicit and I only need to read a single line to know what is going on.

PR #2578, opened 2018-02-27
fix: Inline styles to prevent content flash

AndyLnd:
This is just a proposal to prevent the content flashing in Firefox and general issues related to style loading.
The downside would be that the style would no longer be cached (31k gzipped).
The upside would be that everything would be styled as soon as it enters the DOM and that the user will always have the latest version of the css (now accomplished by adding version numbers to the css href).

WDYT?


app/page/index.html
```diff
@@ -6,7 +6,9 @@
     #include('meta.htm')
     #include('graph.htm')
     <title>Wire · Secure messenger</title>
-    #include('#dest/style.htm')
+    <style>
+      #include('../../../dist/style/main.css')
```
gregor: We need to adjust the path

PR #2590, opened 2018-02-28
chore: Migrate /view_model/list to camelCase

gregor:


app/script/view_model/ModalsViewModel.js
```diff
@@ -138,7 +143,7 @@ z.viewModel.ModalsViewModel = class ModalsViewModel {
       this.logger.info(`Show modal of type '${type}'`);
     }
 
-    modal.autoclose = Boolean(options.preventClose);
+    modal.autoclose = !preventClose;
```
lipis, line 146: That's the opposite of what it was doing.. I guess you meant `!!`.. but I still prefer `Boolean` (even though I like the `!!`)

gregor: Nope. It is exactly what I meant. It was wrong before.

PR #2591, opened 2018-02-28
chore: Change Boolean to !!

lipis:
It's clearer..


gregor:
Can we agree on something before we open the PRs? And is there an eslint rule for that? I don't really care either way. The only thing I dislike is that `Boolean()` is longer and thus more like to cause prettier to break things into multiple lines, which is not easier to read.

lipis:
There is that: https://eslint.org/docs/rules/no-extra-boolean-cast but it's not exactly replacing anything.. just removing the redundant one.

PR #2598, opened 2018-03-01
feat: Add group participants selection for adding

gregor:


gregor:
This is still missing the service confirmation

PR #2599, opened 2018-03-01
chore: Link routes from react-webapp to classic-webapp in dev mode

Yserz:


bennycode:
With `yarn start:all` you can now run the React environment & Knockout.js environment. That's great! 🎉

PR #2615, opened 2018-03-02
feat: group-participant (w/o services)

AndyLnd:
Services still need to get a proper design.


gregor:
Some additional icons are still missing.

PR #2663, opened 2018-03-05
fix: Style conversation name input

Yserz:


app/style/panel/conversation-details.less
```diff
@@ -72,6 +72,15 @@
       justify-content: center;
     }
 
+    &--input {
+      width: 250px;
+      height: auto;
+      border: 0;
+      border-radius: 4px;
```
gregor, line 79: Seems obsolete. Let's remove on next iteration.

PR #2679, opened 2018-03-06
fix: Remove option missing from service participant

gregor:


app/script/view_model/PanelViewModel.js
```diff
@@ -212,6 +212,11 @@ z.viewModel.PanelViewModel = class PanelViewModel {
   }
 
   _hidePanel() {
+    const isStateGroupParticipant = this.state() === PanelViewModel.STATE.GROUP_PARTICIPANT;
```
lipis: 💯

PR #2708, opened 2018-03-07
fix: Conversations typo in title

ffflorian:


app/script/localization/translations/webapp-fr.js
```diff
@@ -393,7 +393,7 @@ z.string.fr.participantDevicesDetailVerify = 'Vérifié';
 z.string.fr.participantDevicesHeader = 'Appareils';
 z.string.fr.participantDevicesHeadline = 'Wire donne à chaque appareil une empreinte unique. Comparez-les avec {{user}} et vérifiez votre conversation.';
 z.string.fr.participantDevicesLearnMore = 'En savoir plus';
-z.string.fr.participantDevicesWhyVerify = 'Pourquoi vérifier une conversation ?';
+z.string.fr.participantDevicesWhyVerify = 'Pourquoi vérifier les conversations ?';
```
lipis, line 396: while we are here.. extra space before `?`

ffflorian: This is correct: https://unicode.org/udhr/n/notes_fra.html
Okay, actually it should be a non-breaking space, but we don't have that here..

app/script/localization/translations/webapp-de.js
```diff
@@ -74,7 +74,7 @@ z.string.de.authPlaceholderPhone = 'Telefonnummer';
 z.string.de.authErrorCode = 'Ungültiger Verifizierungs-Code';
 z.string.de.authErrorCountryCodeInvalid = 'Ungültige Landesvorwahl';
 z.string.de.authErrorEmailExists = 'E-Mail-Adresse bereits vergeben';
-z.string.de.authErrorEmailForbidden = 'Es tut uns leid. Diese E-Mail-Adresse ist verboten.';
+z.string.de.authErrorEmailForbidden = 'Diese Telefonnummer ist leider nicht nutzbar.';
```
gregor: Not good...

app/script/localization/translations/webapp-fr.js
lipis: TIL :)

lipis: We should maybe replace all the ` ?` during the post process in this file with a non-breaking space in order to avoid having the `?` ending up in a separate line if it can't fit..

(but in a separate PR)

PR #2735, opened 2018-03-08
feat: alternative info text for guest mode toggle

AndyLnd:


app/script/localization/webapp.js
```diff
@@ -360,6 +360,8 @@ z.string.groupCreationParticipantsPlaceholder = 'Search by name';
 z.string.guestRoomConversationName = 'Guest room';
 z.string.guestRoomToggleName = 'Allow guests';
 z.string.guestRoomToggleInfo = 'Open this conversation to people outside your team.';
+z.string.guestRoomToggleInfoLong = 'Open this conversation to people outside your team. You can always change it later.';
+
```
lipis: extra line :)

and maybe syncing with Crowdin?

gregor: Done and the extra line is good here.

PR #2753, opened 2018-03-09
fix: Login for Webapp Bot

bennyn:
**Note to myself:** Create a `travis-bot` npm package to avoid these kind of fixes. This could also make use of `'generate-changelog'`.


gregor:
Not sure how but it seems like this one actually caused a bunch of tests in smoke to fail.

PR #2755, opened 2018-03-10
chore: Migrate ConversationCellState to camelCase

gregor:


app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
-              activity_strings.push(z.l10n.text(z.string.conversationsSecondaryLineMissedCalls, count));
-            }
-            break;
-          default:
-        }
+    for (const messageEntity of unreadEvents) {
+      if (messageEntity.is_call() && messageEntity.was_missed()) {
+        activities[ACTIVITY_TYPE.CALL] = activities[ACTIVITY_TYPE.CALL] + 1;
+      } else if (messageEntity.is_ping()) {
+        activities[ACTIVITY_TYPE.PING] = [ACTIVITY_TYPE.PING] + 1;
```
AndyLnd: ```activities[ACTIVITY_TYPE.PING] = activities[ACTIVITY_TYPE.PING] + 1;```
Actually, I would prefer to write those as ```activities[ACTIVITY_TYPE.PING] += 1;```

app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
+        _getStateAlert,
+        _getStateGroupActivity,
+        _getStateUnreadMessage,
+        _getStateUserName,
+      ];
+      const iconState = states.find(state => state.match(conversationEntity));
+      const descriptionState = states.find(state => state.match(conversationEntity));
+
+      return {
+        description: (descriptionState || _getStateDefault).description(conversationEntity),
```
AndyLnd: Maybe move the default to line 295 (same for the line below).

app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
-      return '';
     },
-    icon() {
-      return z.conversation.ConversationStatusIcon.NONE;
-    },
-    match(conversation_et) {
-      return conversation_et.removed_from_conversation();
+    match: conversationEntity => {
+      const hasUnreadEvents = conversationEntity.unread_event_count() > 0;
+      return hasUnreadEvents && !!conversationEntity.unread_events().find(_isAlert);
```
AndyLnd: Better use ```.some()``` in this case, as it returns a boolean.

app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
-    const unread_events = conversation_et.unread_events();
-    const activities = {
-      call: 0,
-      message: 0,
-      ping: 0,
-    };
+  const _generateActivityString = activities => {
+    const activityStrings = [];
+
+    for (const activity in activities) {
```
AndyLnd: To avoid ```hasOwnProperty()``` use ```Object.keys(activities).map()```, which would make it more expressive and therefore more readable.

app/script/conversation/ConversationCellState.js
gregor: We decided against `+= 1` moniker a very long time ago. You should not find it on our code at all.

AndyLnd: I don't know your reasons (yet), but I don't think that is a very good decision for simple increments. It enables errors like this one.

lipis: Maybe we should revise it :)

app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
-        activities.ping = activities.ping + 1;
-      } else if (message_et.is_content()) {
-        activities.message = activities.message + 1;
+    for (const messageEntity of unreadEvents) {
+      if (messageEntity.is_call() && messageEntity.was_missed()) {
+        activities[ACTIVITY_TYPE.CALL] += 1;
+      } else if (messageEntity.is_ping()) {
+        activities[ACTIVITY_TYPE.PING] += 1;
+      } else if (messageEntity.is_content()) {
+        activities[ACTIVITY_TYPE.MESSAGE] = +1;
```
AndyLnd: +=

app/script/conversation/ConversationCellState.js
```diff
@@ -23,279 +23,282 @@ window.z = window.z || {};
...
-    icon() {
-      return z.conversation.ConversationStatusIcon.NONE;
-    },
+  const _generateActivityString = activities => {
+    return Object.keys(activities)
+      .map(activity => {
+        const activityCount = activities[activity];
+        const hasActivity = activityCount >= 1;
+
+        if (hasActivity) {
```
lipis: The following is still very clear.. since the `activityCount` is a number
```js
if (activityCount) {
```

gregor: Very true, Ende like this after multiple refactorings as we did the check for each case in the switch individually.

PR #2766, opened 2018-03-12
fix: keep scroll to bottom on panel toggle

AndyLnd:


app/script/util/util.js
```diff
@@ -772,6 +772,12 @@ z.util.format_time_remaining = function(time_remaining) {
   return title || '';
 };
 
+z.util.afterRender = callback => {
+  window.requestAnimationFrame(() => {
+    window.setTimeout(callback, 0);
+  });
```
gregor: I guess we are fine moving this into a single line.

PR #2782, opened 2018-03-13
fix: meaningful animation directions

AndyLnd:


app/script/view_model/PanelViewModel.js
```diff
@@ -267,10 +267,11 @@ z.viewModel.PanelViewModel = class PanelViewModel {
 
   _openPanel(newState) {
     if (!this.isAnimating()) {
+      const isVisible = this.isVisible();
```
gregor: Let's name this differently -> `previouslyVisible` for example to make clear what we are checking here.

PR #2797, opened 2018-03-14
feat: Call logs  (WEBAPP-4872)

pwnsdx:


app/script/util/util.js
```diff
@@ -779,6 +779,54 @@ z.util.afterRender = callback => {
...
+    super(message, options);
+  }
+
+  static get CONFIG() {
+    return {
+      MESSAGE_LOG_LENGTH: 10000,
+    };
+  }
+
+  _get_debug_type(number) {
```
lipis: Use camelCase for new stuff

app/script/util/util.js
```diff
@@ -779,6 +779,54 @@ z.util.afterRender = callback => {
...
+  _log_to_memory(args) {
+    while (z.util.CallingLog.length >= CallingLogger.CONFIG.MESSAGE_LOG_LENGTH) {
+      z.util.CallingLog.shift();
+    }
+
+    const logMessage = `[${new Date().toISOString()}] (${this._get_debug_type(args[0]())}) ${args[1]}`;
+    z.util.CallingLog.push(logMessage);
+  }
+
+  _print_log(args) {
```
lipis: everywhere

app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -71,6 +71,19 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
     amplify.publish(z.event.WebApp.CONNECT.IMPORT_CONTACTS, z.connect.ConnectSource.ICLOUD);
   }
 
+  save_call_logs() {
```
lipis: `camelCase`

app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -71,6 +71,19 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
...
+    if (z.util.CallingLog.length < 1) {
```
lipis:
```js
if (!z.util.CallingLog.length) {
```

ghost:
```
const b = Object.create([]);
b.length = -1;
if (!b.length) {
console.log('Empty');
} else {
console.log('Full of stuff');
}
```

Even if length should never be modified that way, not sure doing your way is the best.

lipis: True but modifying `b.length = 4` will also going to cause the same issues and that's why we should not modify `length` :)

app/script/util/util.js
ghost: `_print_log` is from the logging library and cannot be changed

app/script/calling/CallingRepository.js
```diff
@@ -1419,18 +1427,9 @@ z.calling.CallingRepository = class CallingRepository {
    * @returns {undefined} No return value
    */
   setDebugState(isDebuggingEnabled) {
-    const isStateChange = this.debugEnabled !== isDebuggingEnabled;
-    if (isStateChange) {
```
gregor: Please keep the isStateChange check to skip obsolete log messages

gregor:
We also need to review this with design before merging,

app/script/util/util.js
```diff
@@ -760,6 +760,18 @@ z.util.afterRender = callback => {
 
 /**
  * No operation
- * @returns {undefined}
+ * @returns {void}
  */
 z.util.noop = function() {};
+
+z.util.format_string = (strings, ...keys) => {
```
gregor: This should go to the StringUtil - where we actually have a function for that.

app/script/calling/CallingRepository.js
```diff
@@ -65,7 +64,7 @@ z.calling.CallingRepository = class CallingRepository {
     this.conversationRepository = conversationRepository;
     this.mediaRepository = mediaRepository;
     this.userRepository = userRepository;
-    this.logger = new z.util.Logger('z.calling.CallingRepository', z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger('z.calling.CallingRepository', z.config.LOGGER.OPTIONS);
```
gregor: Let's rename these to `callLogger` to make clear it has additional functionality whenever we use it.

app/script/calling/entities/CallEntity.js
```diff
@@ -53,7 +53,10 @@ z.calling.entities.CallEntity = class CallEntity {
     const {id: conversationId, is_group} = conversationEntity;
     const {mediaStreamHandler, mediaRepository, selfState, telemetry, userRepository} = this.callingRepository;
 
-    this.logger = new z.util.Logger(`z.calling.entities.CallEntity (${conversationId})`, z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger(
+      `z.calling.entities.CallEntity (${conversationId})`,
+      z.config.LOGGER.OPTIONS
+    );
```
gregor: Move the logger name to a variable to turn these 4 lines into 2

app/script/calling/entities/FlowAudioEntity.js
```diff
@@ -36,14 +36,24 @@ z.calling.entities.FlowAudioEntity = class FlowAudioEntity {
 
     this.flowEntity = flowEntity;
     this.mediaRepository = mediaRepository;
-    this.logger = new z.util.Logger(`z.calling.entities.FlowAudio (${this.flowEntity.id})`, z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger(
+      `z.calling.entities.FlowAudio (${this.flowEntity.id})`,
+      z.config.LOGGER.OPTIONS
+    );
```
gregor: Move the logger name to a variable to turn these 4 lines into 2

app/script/calling/entities/FlowEntity.js
```diff
@@ -51,7 +51,10 @@ z.calling.entities.FlowEntity = class FlowEntity {
     this.id = this.participantEntity.id;
     this.conversationId = this.callEntity.id;
 
-    this.logger = new z.util.Logger(`z.calling.entities.FlowEntity (${this.id})`, z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger(
+      `z.calling.entities.FlowEntity (${this.id})`,
+      z.config.LOGGER.OPTIONS
+    );
```
gregor: Move the logger name to a variable to turn these 4 lines into 2

app/script/calling/CallingRepository.js
```diff
@@ -1180,7 +1188,18 @@ z.calling.CallingRepository = class CallingRepository {
...
-      this.logger.info(`Outgoing '${mediaType}' call in conversation '${conversationName}'`, callEntity);
+      const conversationId = callEntity.conversationEntity.id;
+
+      this.logger.info(
+        {
+          data: {
+            default: [mediaType, conversationName],
+            obfuscated: [mediaType, conversationId],
+          },
+          message: z.util.format_string`Outgoing '${0}' call in conversation '${1}'`,
```
gregor: Why do you have z.util.format_string in each of these? Only the logger should call them internall, should't it?

app/script/calling/entities/ParticipantEntity.js
```diff
@@ -38,7 +38,10 @@ z.calling.entities.ParticipantEntity = class ParticipantEntity {
     this.id = this.user.id;
     this.sessionId = undefined;
 
-    this.logger = new z.util.Logger(`z.calling.entities.ParticipantEntity (${this.id})`, z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger(
+      `z.calling.entities.ParticipantEntity (${this.id})`,
+      z.config.LOGGER.OPTIONS
+    );
```
gregor: Move the logger name to a variable to turn these 4 lines into 2

app/script/telemetry/calling/FlowTelemetry.js
```diff
@@ -35,7 +35,10 @@ z.telemetry.calling.FlowTelemetry = class FlowTelemetry {
   constructor(id, remote_user_id, call_et, timings) {
     this.remote_user_id = remote_user_id;
     this.call_et = call_et;
-    this.logger = new z.util.Logger(`z.telemetry.calling.FlowTelemetry (${id})`, z.config.LOGGER.OPTIONS);
+    this.logger = new z.telemetry.calling.CallLogger(
+      `z.telemetry.calling.FlowTelemetry (${id})`,
+      z.config.LOGGER.OPTIONS
+    );
```
gregor: Move the logger name to a variable to turn these 4 lines into 2

app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -68,6 +68,19 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
     amplify.publish(z.event.WebApp.CONNECT.IMPORT_CONTACTS, z.connect.ConnectSource.ICLOUD);
   }
 
+  saveCallLogs() {
+    if (!z.telemetry.calling.CallLog.length) {
+      return;
+    }
+
+    const blob = new Blob([z.telemetry.calling.CallLog.join('\r\n')], {
```
gregor: Move the blob parts in to a variable so this fits into a single line.

app/script/calling/CallingRepository.js
ghost: @gregor What do you mean by obsolete log messages? setDebugState is always on true now since we need to activate WebRTC adapter logging: https://github.com/wireapp/wire-webapp/pull/2797/files/61a6dca5039c627341dfad0de075c228f06acd8a#diff-584962c1c38d95a8eccb8fe4948ed7b8R106

gregor: I see your point. But then you should rename the function and not use a parameter. Currently this is a feature to set the debug state. But now you always want to set it to true when initializing. So it turns into a private `_enableDebugging()` with no need to contain a log message at all.

app/script/calling/CallingRepository.js
```diff
@@ -1426,13 +1425,14 @@ z.calling.CallingRepository = class CallingRepository {
...
    * Set logging on adapter.js.
-   * @param {boolean} isDebuggingEnabled - Updated debug state
    * @returns {undefined} No return value
    */
-  setDebugState(isDebuggingEnabled) {
+  _enableDebugging() {
     if (window.adapter) {
-      this.callLogger.debug(`Set logging for WebRTC Adapter: ${isDebuggingEnabled}`);
-      window.adapter.disableLog = !isDebuggingEnabled;
+      this.logger.debug(`Set logging for WebRTC Adapter`);
```
gregor: Let's remove this. It's always going to happen.

app/script/calling/CallingRepository.js
```diff
@@ -1426,13 +1425,14 @@ z.calling.CallingRepository = class CallingRepository {
...
+      window.adapter.disableLog = false;
+    } else {
+      this.logger.debug(`WebRTC Adapter not found while trying to enable logging`);
```
gregor: We can keep this as a warning for the negative case. => `this.logger.warn`

app/script/telemetry/calling/CallLogger.js
```diff
@@ -0,0 +1,186 @@
...
+      IPV4: /(([0-1]?[0-9]{1,2}\.)|(2[0-4][0-9]\.)|(25[0-5]\.)){3}(([0-1]?[0-9]{1,2})|(2[0-4][0-9])|(25[0-5]))/,
+      IPV6: /(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))/,
+      MESSAGE_LOG_LENGTH: 10000,
+      OBFUSCATION_TRUNCATE_TO: 2,
+    };
+  }
+
+  obfuscate(string) {
+    //return CryptoJS.SHA256(`${string}${this.currentWeek}`)
+    //  .toString()
```
gregor: Remove if unused

app/script/telemetry/calling/CallLogger.js
```diff
@@ -0,0 +1,186 @@
...
+          return (v % 10).toString(10);
+        })
+        .join('')
+        .match(/.{1,3}/g)
+        .join('.');
+    } else if (type === 6) {
+      const fakeIp = CryptoJS.MD5(seed).toString();
+      return fakeIp.match(/.{1,4}/g).join(':');
+    }
+  }
```
gregor: remove if unused

app/script/telemetry/calling/CallLogger.js
```diff
@@ -0,0 +1,186 @@
...
+    const decodedSdpMessage = sdpTransform.parse(sdpMessage);
+
+    for (const index in decodedSdpMessage.media) {
+      // Remove fingerprint
+      if (
+        typeof decodedSdpMessage.media[index].fingerprint !== 'undefined' &&
+        typeof decodedSdpMessage.media[index].fingerprint.hash !== 'undefined'
+      ) {
+        decodedSdpMessage.media[index].fingerprint.hash =
+          'XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX:XX';
```
gregor: I suppose the fake data can be shortened

app/script/telemetry/calling/CallSetupTimings.js
```diff
@@ -28,7 +28,10 @@ z.telemetry.calling.CallSetupTimings = class CallSetupTimings {
     this.get = this.get.bind(this);
     this.log = this.log.bind(this);
     this.call_id = call_id;
-    this.logger = new z.util.Logger('z.telemetry.calling.CallSetupTimings', z.config.LOGGER.OPTIONS);
+    this.callLogger = new z.telemetry.calling.CallLogger(
+      'z.telemetry.calling.CallSetupTimings',
+      z.config.LOGGER.OPTIONS
```
gregor: Same as before. Move name to const to reduce number of lines

app/script/telemetry/calling/CallLogger.js
```diff
@@ -0,0 +1,186 @@
+z.telemetry.calling.CallLog = [];
```
gregor: This array should preferably be initiated within the CallTelemetry instance and passed as an argument into the constructor.

app/script/calling/CallingRepository.js
```diff
@@ -1136,7 +1133,17 @@ z.calling.CallingRepository = class CallingRepository {
...
-        this.logger.info(`Incoming '${mediaType}' call in conversation '${conversationName}'`, callEntity);
+
+        this.callLogger.info(
+          {
+            data: {
+              default: [mediaType, conversationName],
+              obfuscated: [mediaType, this.callLogger.obfuscate(conversationId)],
+            },
+            message: `Incoming '{0}' call in conversation '{1}'`,
+          },
```
gregor: I guess we should actually move all of these out into const objects for improved readability

app/script/localization/translations/webapp-fr.js
```diff
@@ -478,6 +478,9 @@ z.string.fr.preferencesOptionsNotificationsNone = 'Désactiver';
 z.string.fr.preferencesOptionsNotificationsObfuscate = 'Masquer les détails';
 z.string.fr.preferencesOptionsNotificationsObfuscateMessage = 'Afficher l’expéditeur';
 z.string.fr.preferencesOptionsNotificationsOn = 'Afficher l’expéditeur et le message';
+z.string.fr.preferencesOptionsCallLogs = 'Dépannage';
+z.string.fr.preferencesOptionsCallLogsGet = 'Sauvegarder le rapport de débogage des appels';
+z.string.fr.preferencesOptionsCallLogsDetail = 'Cette information aide l’assistance de Wire à diagnostiquer les problèmes d'appel.';
```
lipis: `d'appel` -> `d’appel`

There is the wrong apostrophe in the french translations.. and that's why the `;`  look red in the preview (lines 485-486)

app/script/telemetry/calling/CallLogger.js
```diff
@@ -0,0 +1,186 @@
...
+      case 400:
+      case 500:
+        return 'VERBOSE';
+      case 700:
+        return 'DEBUG';
+      case 800:
+        return 'INFO';
+      case 900:
+        return 'WARNING';
+      case 100:
```
lipis, line 164: Can we have constants for these numbers?! Maybe we can also have ranges instead of cases..

```js
if (number < DEBUG_CODE_ERROR) {
  return 'ERROR';
} else if (number < DEBUG_CODE_INFO) { 
  return 'INFO';
} else ...

```

app/script/localization/translations/webapp-fr.js
ghost: Yes, that was already fixed.

app/script/telemetry/calling/CallLogger.js
ghost: I don't think it's a good idea for readability to have ranges especially since you will never have a debug code 101.

app/script/telemetry/calling/CallLogger.js
ghost: Not sure this would work, it needs to be accessible and shared among all other classes where CallLogger is initialized.

app/script/telemetry/calling/CallLogger.js
ghost: This is the size of the fingerprint. We must keep the same size for readability. Using `new Array(26).fill('XX').join(':')` is possible but would make an additional function call so better keeping it like this.

app/script/telemetry/calling/CallLogger.js
ghost: It will be needed in the future. Should I uncomment it then?

app/script/telemetry/calling/CallLogger.js
ghost: It will be needed in the future. Should I uncomment it then?

app/script/telemetry/calling/CallLogger.js
gregor: Why do we need to keep it for readability? You can change it to "OBFUSCATED" and it will be obvious what you did as wel. Why should we print 26 couples of Xs here?

app/script/calling/CallingRepository.js
ghost: Not a good idea. Sometime there is 2 logging within the same function so we would need different const names and will therefore reduce readability instead of improving it.

app/script/telemetry/calling/CallLogger.js
ghost: So it looks like a fingerprint and it's easier to read as "X" are used for removed data. I moved those things to the config file so calls are only done once: https://github.com/wireapp/wire-webapp/pull/2797/files#diff-2d0c04a493e7ca82351e008e3e9ebd30R93 and used `Array.from({length: 32}).fill('XX').join(':')` for this specific line.

app/script/telemetry/calling/CallLogger.js
gregor: Where does 200 belong though?

PR #2802, opened 2018-03-14
fix: new core and dexie fix

AndyLnd:


gregor:
Not needed. We will do a reverse merge.

PR #2810, opened 2018-03-15
chore: Log type of event, which could not be decrypted

bennyn:
Added `eventType` to decryption error logs.

### DIFF
--- a/app/script/cryptography/CryptographyRepository.js
+++ b/app/script/cryptography/CryptographyRepository.js
@@ -441,7 +441,7 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
     // Get error information
     const errorCode = error.code || CryptographyRepository.CONFIG.UNKNOWN_DECRYPTION_ERROR_CODE;
 
-    const {data: eventData, from: remoteUserId} = event;
+    const {data: eventData, from: remoteUserId, type: eventType} = event;
 
     const isDuplicateMessage = error instanceof Proteus.errors.DecryptError.DuplicateMessage;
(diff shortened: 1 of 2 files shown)


bennycode:
I need to identify the event which could not be decrypted. 

Given the fact that they are all of type `'conversation.otr-message-add'`, which information would you add to distinguish between multiple OTR messages?

The common envelope is:

> [ "type"         .= <event type>                          -- event type
, "conversation" .= <uuid>                                -- conv id
, "from"         .= <uuid>                                -- user id
, "time"         .= <time utc>                            -- format: "2018-02-21T22:47:18.773Z"
, "data"         .= <data dependent on the event type>
]

gregor:
Timestamp is the only unique information you will find

app/script/cryptography/CryptographyRepository.js
```diff
@@ -470,7 +470,7 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
     }
 
     this.logger.warn(
-      `Failed to decrypt event from user '${remoteUserId}' (${remoteClientId}).\nError Code: '${errorCode}'\nError Message: ${
+      `Failed to decrypt event '${eventType}' from user '${remoteUserId}' with client '${remoteClientId}'.\nError Code: '${errorCode}'\nError Message: ${
```
gregor:
```
`Failed to decrypt event from client '${remoteClientId}' of user '${remoteUserId}' (${time}).\nError Code: '${errorCode}'\nError Message: ${
``

bennycode: updated

PR #2835, opened 2018-03-16
fix: copy changes (WEBAPP-4832)

AndyLnd:


app/script/localization/webapp.js
```diff
@@ -350,7 +350,7 @@ z.string.conversationTweetAuthor = ' on Twitter';
 z.string.groupCreationPreferencesAction = 'Next';
 z.string.groupCreationPreferencesErrorNameShort = 'At least 1 character';
 z.string.groupCreationPreferencesErrorNameLong = 'Too many characters';
-z.string.groupCreationPreferencesHeader = 'New group';
+z.string.groupCreationPreferencesHeader = 'Group name';
```
gregor: @Yorgos-V @eike96 I strongly disagree with this change. It does not solve the user problem and will only add to the confusion. The screen is not a group name screen but also hosts the guest options toggle for team users.

Just changing the title of the screen to relate to the group name also does not solve the user confusion when inputting text on the field of this screen. The label is not visually linked to the input box. The only identifier for the functionality of the input box is currently the placeholder text.

The UX problem will remain if we change the title as proposed. Not only that but we will also add another problem as the title no longer reflects the purpose of this page.

PR #2853, opened 2018-03-19
fix: Improved platform check

Yserz:


app/script/auth/Runtime.js
```diff
@@ -55,15 +57,25 @@ export default class Runtime {
   getBrowserName = () => this.getPlatform().name.toLowerCase();
   getBrowserVersion = () => this.getPlatform().version;
   getMajorBrowserVersion = () => parseInt(this.getPlatform().version.split('.')[0]);
+  getMinorBrowserVersion = () => parseInt(this.getPlatform().version.split('.')[1]);
```
gregor: Let's not do the version split twice but do a get browser version thingy that returns all parts of the version

app/script/auth/Runtime.js
```diff
@@ -55,15 +57,25 @@ export default class Runtime {
...
+    if (isFranz) {
+      return false;
+    }
+
     return Object.entries(SUPPORTED_BROWSERS).some(([browser, version]) => {
-      return this.getBrowserName() === browser && this.getMajorBrowserVersion() >= version;
+      const isSupportedBrowser = this.getBrowserName() === browser;
+      const isSupportedMajorVersion = this.getMajorBrowserVersion() >= version.major;
+      const isSupportedMinorVersion = this.getMinorBrowserVersion() >= version.minor;
+      return isSupportedBrowser && isSupportedMajorVersion && isSupportedMinorVersion;
```
lipis: this will fail if the major is bigger while minor is smaller:

Electron 1.6 vs Electron 2.1

app/script/auth/Runtime.js
```diff
@@ -53,17 +55,29 @@ export default class Runtime {
 
   getOSFamily = () => this.getOS().family.toLowerCase();
   getBrowserName = () => this.getPlatform().name.toLowerCase();
-  getBrowserVersion = () => this.getPlatform().version;
-  getMajorBrowserVersion = () => parseInt(this.getPlatform().version.split('.')[0]);
+  getBrowserVersion = () => {
+    const [majorVersion, minorVersion] = this.getPlatform().version.split('.');
+    return {major: parseInt(majorVersion, 10), major: parseInt(minorVersion, 10)};
```
gregor: major, minor

PR #2863, opened 2018-03-20
fix: Lesser expiration time option

Yserz:


app/script/auth/page/ConversationJoin.jsx
```diff
@@ -69,8 +70,11 @@ class ConversationJoin extends Component {
   readAndUpdateParamsFromUrl = (nextProps = this.props) => {
     const conversationCode = nextProps.match.params.conversationCode;
     const conversationKey = nextProps.match.params.conversationKey;
+    const expiresIn = parseInt(nextProps.match.params.expiresIn, 10) || undefined;
     const keyOrCodeChanged =
-      conversationCode !== this.state.conversationCode || conversationKey !== this.state.conversationKey;
+      conversationCode !== this.state.conversationCode ||
+      conversationKey !== this.state.conversationKey ||
+      expiresIn !== this.state.expiresIn;
```
gregor: Name no longer fits. Let's move these into separate consts and then combine it into asingle meaning for easier understanding.

PR #2874, opened 2018-03-21
feat: Handle team Id for users (WEBAPP-4866)

gregor:


app/script/assets/AssetService.js
```diff
@@ -174,9 +174,13 @@ z.assets.AssetService = class AssetService {
   }
 
   getAssetRetention(userEntity, conversationEntity) {
-    const isTeamMember = userEntity.is_team_member();
+    const isTeamMember = userEntity.inTeam();
```
AndyLnd, line 177: The naming should be changed soonish to reflect that the user can be a part of any team, not just yours (and it should be prefixed with a 'is' as it returns a boolean).

PR #2882, opened 2018-03-21
feat: wireless user features

AndyLnd:


app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -98,7 +98,11 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
 
     this.isNameEditable = ko.pureComputed(() => {
       if (this.hasConversation()) {
-        return this.conversationEntity().is_group() && !this.conversationEntity().removed_from_conversation();
+        return (
+          this.conversationEntity().is_group() &&
+          !this.conversationEntity().removed_from_conversation() &&
+          !this.mainViewModel.isTemporaryGuest()
```
gregor: Why guests can edit the name, why shouldn't temporary guests be able to do it as well?

PR #2896, opened 2018-03-22
feat: New login and temporary guest experience

gregor:


app/script/auth/page/ConversationJoin.jsx
```diff
@@ -91,15 +103,6 @@ class ConversationJoin extends Component {
 
   componentWillReceiveProps = nextProps => this.readAndUpdateParamsFromUrl(nextProps);
 
-  openWebapp = params => {
-    const link = document.createElement('a');
-    link.href = pathWithParams(EXTERNAL_ROUTE.LOGIN, params);
-    document.body.appendChild(link); // workaround for Firefox
```
gregor: Do we no longer need the workaround for Firefox? We did not remove it from https://github.com/wireapp/wire-webapp/pull/2896/files#diff-5faba9f0446a38fd86c2ec6241b9f713R67

app/script/auth/page/Index.js
```diff
@@ -59,10 +57,7 @@ class Index extends Component {
 
   onLoginClick = () => {
     this.props.trackEvent({name: TrackingAction.EVENT_NAME.START.OPENED_LOGIN});
-    const link = document.createElement('a');
-    link.href = pathWithParams(EXTERNAL_ROUTE.LOGIN, 'mode=login');
-    document.body.appendChild(link); // workaround for Firefox
-    link.click();
+    this.props.history.push(ROUTE.LOGIN);
```
gregor, line 60: See above. Why did we partially remove the workaround for Firefox but kept it in other places?

app/script/auth/page/Login.js
```diff
@@ -0,0 +1,282 @@
...
+        }
+      })
+      .then(() => {
+        const {email, password, persist} = this.state;
+        const login = {password, persist};
+        if (email.includes('@')) {
+          login.email = email;
+        } else {
+          login.handle = email.replace('@', '');
+        }
```
gregor, line 135: This is the spot to readd the secret phone login with password functionality we previously had.

app/script/auth/page/Root.js
```diff
@@ -61,8 +65,12 @@ class Root extends React.Component {
           <Router hashType="noslash">
             <Switch>
               <Route exact path={ROUTE.INDEX} component={Index} />
+              <Route path={ROUTE.CLIENTS} component={ClientManager} />
+              <Route path={`${ROUTE.LOGIN}/:conversationKey?/:conversationCode?`} component={Login} />
+              <Route path={ROUTE.CONVERSATION_JOIN_INVALID} component={ConversationJoinInvalid} />
+
```
gregor, line 71: Obsolete empty line

app/script/auth/route.js
```diff
@@ -19,13 +19,18 @@
...
+  CONVERSATION_JOIN_INVALID: '/join-conversation-invalid',
   CREATE_ACCOUNT: '/createaccount',
   CREATE_TEAM: '/createteam',
   CREATE_TEAM_ACCOUNT: '/createteamaccount',
+  HISTORY_INFO: '/historyinfo',
   INDEX: '/',
   INITIAL_INVITE: '/teaminvite',
   INVITE: '/invite',
+  LOGIN: '/login',
+  PASSWORD_RESET: '/forgot',
```
gregor, line 33: Is this really an internal route or rather an external path we set so that it does not really belong here?

app/script/entity/User.js
```diff
@@ -247,27 +249,30 @@ z.entity.User = class User {
...
+      this.expirationRemaining(timeValue * MILLISECONDS_IN_HOUR);
+      this.expirationRemainingText(`${timeValue}h`);
     } else {
       timeValue = Math.ceil(remainingMinutes / 60);
-      this.expirationRemaining(timeValue * timeToHours);
+      this.expirationRemaining(timeValue * MILLISECONDS_IN_HOUR);
+      this.expirationRemainingText(`${timeValue}h`);
     }
-
+    this.expirationIsUrgent(remainingMinutes < 120);
```
gregor: If this is only used for the self user in the temporary guest left list we should consider moving it to the viewmodel and thereby removing it from all the user entities.

app/script/localization/translations/webapp-de.js
```diff
@@ -105,10 +102,14 @@ z.string.de.modalAcknowledgeHeadline = 'Fehler aufgetreten';
 z.string.de.modalConfirmSecondary = 'Abbrechen';
 z.string.de.modalOptionSecondary = 'Abbrechen';
 
+z.string.de.modalAccountCreateAction = 'Verstanden';
+
```
gregor, line 106: Weird that the related headline and message are missing

app/script/auth/page/ConversationJoin.jsx
AndyLnd: The workaround is no longer necessary in this case, as we no longer redirect to a different page.

app/script/auth/page/Index.js
AndyLnd: See above.

PR #1615, opened 2017-08-22
fix: WEBAPP-4292

pwnsdx:

PR #2917, opened 2018-03-23
fix: WEBAPP-4292

pwnsdx:
Pako has been removed

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -29,7 +29,6 @@
     "marked": true,
     "moment": true,
     "OpenGraphMocks": true,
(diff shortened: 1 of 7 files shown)


gregor:
We definitely need to talk about this one. This was put in on request by the backend when we implemented address book uploads. Just removing it completely might not be what we want to do.

bennycode:
@pwnsdx Please also check my comments from https://github.com/wireapp/wire-webapp/pull/1615/files. 

I can see that you already integrated `JSON.stringify(login)` but we might need `JSON.stringify(new_user)` too.

PR #2949, opened 2018-03-26
chore: Update OG info (WEBAPP-4868)

lipis:


gregor:
@lipis This updates the OG info. I suppose this is not exposed on our react based pages though. The underying bug ticket asks for proper OG on the join link.

PR #2954, opened 2018-03-27
fix: Set the correct lang in <html>

lipis:
Specially for greek when you changing the text to uppercase with CSS it renders it wrong if the languages is not set.

Wrong:
screen shot 2018-03-27 at 10 42 54

Correct:
screen shot 2018-03-27 at 10 43 09

https://jsbin.com/zobisuziso/1/edit?html,css,output


Yserz:
LGTM

PR #2956, opened 2018-03-27
fix: Remove phoneFormat.js

gregor:


app/script/localization/webapp.js
```diff
@@ -326,7 +326,6 @@ z.string.conversationTeamLeave = ' was removed from the team';
 z.string.conversationPing = ' pinged';
 z.string.conversationPingYou = ' pinged';
 z.string.conversationToday = 'today';
-z.string.conversationVerified = 'Verified';
```
gregor, line 329: String is no longer used

lipis:
and let's not add it back.. phone formats are hard :)

gregor:
@lipis Well, we kind of need it for the phone login. No way around it.

PR #2964, opened 2018-03-27
fix: Set last event date to member join event for temporary guest (WEBAPP-4883)

gregor:


app/script/auth/module/action/NotificationAction.js
```diff
@@ -37,3 +37,9 @@ export function resetHistoryCheck() {
     return Promise.resolve().then(() => dispatch(NotificationActionCreator.resetHistoryCheck()));
   };
 }
+
+export function setLastEventDate(lastEventDate) {
+  return function(dispatch, getState, {core}) {
+    return Promise.resolve().then(() => core.service.notification.setLastEventDate(lastEventDate));
+  };
+}
```
ffflorian, line 45: Why do we return a function instead of the Promise here?

gregor: Because it is a thunk

PR #2965, opened 2018-03-27
fix: Init APIClient with clientType

Yserz:


app/script/auth/module/action/AuthAction.js
```diff
@@ -170,8 +171,14 @@ export function doRegisterWireless(registrationData) {
 export function doInit() {
   return function(dispatch, getState, {apiClient}) {
     dispatch(AuthActionCreator.startRefresh());
-    return apiClient
-      .init()
+    return Promise.resolve()
+      .then(() => dispatch(getLocalStorage(LocalStorageKey.AUTH.PERSIST)))
+      .then(persist => {
+        if (!persist) {
```
gregor: You want to check existence and not for a falsy value

PR #2976, opened 2018-03-28
feat: Add a no-response bot

lipis:
For all the issues that are labeled `status: Need more info` and there is no response for 2 weeks.. it will mark the issues as closed. 

https://probot.github.io/apps/no-response/


gregor:
Let's remove the sale bot at the same time then though

lipis:
yes.. that was my thought! This one is less aggressive..

gregor:
And let's wait with this change until after the long weekend just in case anything misbehaves and we need to deal with it.

infotexture:
@lipis I updated the default response per our earlier discussion.

PR #2978, opened 2018-03-28
refactor: Use event context for account events

gregor:


app/script/auth/module/action/TrackingAction.js
```diff
@@ -66,19 +70,18 @@ export function trackEvent(event) {
...
-                desktop_app: RuntimeUtil.getPlatform(),
-              },
-              event.attributes
-            );
-            mixpanel.track(event.name, attributes, successCode => resolve(successCode));
-          })
-      )
+      .then(() => {
+        return new Promise(resolve => {
+          const attributes = Object.assign(
```
Yserz: spread

app/script/auth/module/action/TrackingAction.js
```diff
@@ -51,9 +51,13 @@ export const EVENT_NAME = {
 };
 
 export const EVENT_CONTEXT = {
```
Yserz: Let's call it something like AUTHENTICATION_CONTEXT

PR #3026, opened 2018-04-03
fix: Get hasHistory from clientState

Yserz:


gregor:
Which ticket does this address?

PR #3063, opened 2018-04-05
fix: use base64 for avatar image urls

AndyLnd:
This should fix the issue of some avatars not being displayed under certain (not completely known) conditions.


app/script/assets/AssetRemoteData.js
```diff
@@ -97,9 +97,9 @@ z.assets.AssetRemoteData = class AssetRemoteData {
     const objectUrl = z.assets.AssetURLCache.getUrl(this.identifier);
     return objectUrl
       ? Promise.resolve(objectUrl)
-      : this.load().then(blob => {
+      : this.load().then(async blob => {
           if (blob) {
-            const url = window.URL.createObjectURL(blob);
+            const url = await z.util.loadDataUrl(blob);
```
gregor: Not supported by all browsers that we need to target.

PR #3072, opened 2018-04-05
fix: Improve personal invitation handling (WEBAPP-4954)

gregor:


app/script/auth/module/selector/AuthSelector.js
```diff
@@ -32,3 +32,8 @@ export const getCurrentFlow = state => state.authState.currentFlow;
 export const getAccount = state => state.authState.account || {};
 export const getAccountTeam = state => getAccount(state).team || {};
 export const getAccountTeamName = state => getAccountTeam(state).name;
+
```
Yserz: Do we still have to export `export const REGISTER_FLOW`?

PR #3096, opened 2018-04-09
fix: added synthetic links for firefox

AndyLnd:


gregor:
Maybe not needed after all since the tests fail for a different reason.

PR #3108, opened 2018-04-09
refactor: Handle immediate login

Yserz:


app/script/auth/page/Login.js
```diff
@@ -116,11 +116,23 @@ class Login extends React.PureComponent {
...
     this.readAndUpdateParamsFromUrl();
   };
 
   componentWillReceiveProps = nextProps => this.readAndUpdateParamsFromUrl(nextProps);
 
+  immediateLogin = () => {
+    return Promise.resolve()
+      .then(() => this.props.doInit({isImmediateLogin: true}))
+      .then(() => this.props.doInitializeClient(true, undefined))
+      .then(() => window.location.replace(URLUtil.pathWithParams(EXTERNAL_ROUTE.WEBAPP)))
```
gregor, line 132: @Yserz I suppose we should also check that the user has set a username in this case instead of immediately redirecting to the webapp.

PR #3156, opened 2018-04-13
fix: Move to promises for event mapping

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -2797,7 +2787,13 @@ z.conversation.ConversationRepository = class ConversationRepository {
             return this._onAddEvent(conversationEntity, eventJson);
         }
       })
-      .then((entityObject = {}) => this._handledConversationEvent(entityObject, eventSource, previouslyArchived));
+      .then((entityObject = {}) => this._handledConversationEvent(entityObject, eventSource, previouslyArchived))
+      .catch(error => {
+        const isMessageNotFound = error.type === z.conversation.ConversationError.TYPE.MESSAGE_NOT_FOUND;
+        if (!isMessageNotFound) {
```
lipis, line 2793: Would prefer the `isMessageFound` for a variable name.. instead of the double negative thingy.. but here is a bit trickier I guess, because the message type is `MESSAGE_NOT_FOUND`

lipis: Let's leave it as is.. used only here anyways.

app/script/conversation/EventMapper.js
```diff
@@ -34,25 +34,27 @@ z.conversation.EventMapper = class EventMapper {
   /**
    * Convert multiple JSON events into message entities.
    *
-   * @param {Object} events - Event data
+   * @param {Array} events - Event data
    * @param {Conversation} conversationEntity - Conversation entity the events belong to
    * @param {boolean} [createDummyImage] - Create a dummy image
-   * @returns {Array<Message>} Mapped message entities
+   * @returns {Promise} Resolves with an array of mapped message entities
```
atomrc: You could use the type `Promise<Array<Message>>`

app/script/conversation/ConversationRepository.js
gregor: Yeah. I see your point but I tried to stick with the error types in this case.

PR #3178, opened 2018-04-18
fix: Add blob: to media-src CSP headers (WEBAPP-4993)

lipis:


aws/headers.py
```diff
@@ -67,6 +67,7 @@
   'media-src': [
     "'self'",
     'data:',
+    'blob:',
```
gregor: Shouldn't this be above data though to show some respect to the ABC?

lipis: will sort them all..! Good point

aws/headers.py
```diff
@@ -29,43 +29,44 @@
...
     'https://*.wire.com',
-    'https://*.cloudfront.net',
     'https://*.zinfra.io',
+    'https://1-ps.googleusercontent.com',
+    'https://api.mixpanel.com',
     'https://csi.gstatic.com',
   ],
   'media-src': [
     "'self'",
+    'blob:',
```
lipis, line 69: that's the actual change..! the rest is sorting

PR #3184, opened 2018-04-19
chore: add data-uie-name for import file picker

AndyLnd:


atomrc:
I need to keep that in mind for my next features :)

PR #3186, opened 2018-04-19
fix: add jszip to grunt copy task

atomrc:


atomrc:
Turned out the `npmCopy` task is a not-used task anymore.

PR #3213, opened 2018-04-23
fix: show floored progress percentage (WEBAPP-5001)

AndyLnd:


app/script/view_model/content/HistoryImportViewModel.js
```diff
@@ -44,7 +44,9 @@ z.viewModel.content.HistoryImportViewModel = class HistoryImportViewModel {
 
     this.numberOfTables = ko.observable(0);
     this.numberOfProcessedTables = ko.observable(0);
-    this.loadingProgress = ko.pureComputed(() => this.numberOfProcessedTables() / this.numberOfTables() * 100);
+    this.loadingProgress = ko.pureComputed(() => {
+      return Math.floor(this.numberOfProcessedTables() / this.numberOfTables() * 100);
```
lipis, line 48: Maybe creating a function to transform a number to a percentage..? It's already used twice (and possible in more places) plus there is a magic number that could be stored in one place.

WDYT?

PR #3217, opened 2018-04-23
fix: improve import errors 

atomrc:
## Type of change

previously metadata check wouldn't actually stop the import and the UI would show up in a weird state

### DIFF
--- a/app/script/backup/Error.js
+++ b/app/script/backup/Error.js
@@ -63,10 +63,18 @@ window.z.backup = z.backup || {};
     }
   }
 
+  class IncompatiblePlatformError extends ImportError {
+    constructor(message = 'Backup created by incompatible platform') {
+      super(message);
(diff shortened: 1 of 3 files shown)


app/script/backup/Error.js
```diff
@@ -63,10 +63,18 @@ window.z.backup = z.backup || {};
     }
   }
 
+  class IncompatiblePlatformError extends ImportError {
+    constructor(message = 'Backup created by incompatible platform') {
```
lipis, line 67: We should make these strings translatable.

bennycode: These strings are just used internally. We don't show them to the user, so no translation needed.

AndyLnd: they are not used either way. what is shown, are localized versions. this string is just for logging (if used).

PR #3221, opened 2018-04-23
fix: Merge local and imported conversation state (WEBAPP-5002)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -341,6 +343,39 @@ z.conversation.ConversationRepository = class ConversationRepository {
       });
   }
 
+  updateConversations(conversationsData) {
+    this.changePersistConversationStateSubscription(false);
```
atomrc: To avoid this temporal coupling, I would argue that an optional parameter on the method that actually send the event would make thing way clearer.

if, for example, the `update_self_status` method send an event, it could have an optional `silent = false` parameter and in the implementation `if (!silent) { amplify.trigger(...) }`.

The 2 arguments are:
- explicitness (you can easily read that the method has been silenced) ;
- no temporal coupling (add/remove the listener)
- bonus: no hard to implement and backward compatible

atomrc: cc @gregor (I don't think you will be pinged since I created the PR)

PR #3227, opened 2018-04-24
fix: Images do not load after restore of history backup (WEBAPP-5004)

AndyLnd:


app/script/backup/BackupRepository.js
```diff
@@ -78,12 +78,6 @@ z.backup.BackupRepository = class BackupRepository {
           throw new z.backup.CancelError();
         }
         progressCallback(rows.length);
-
-        const isEventTable = tableName === z.storage.StorageSchemata.OBJECT_STORE.EVENTS;
```
atomrc: Wow, this is not normal at all. This is something we added this morning, it's not supposed to be removed

atomrc: There must have been a weird git operation I think

AndyLnd: weird ... i will add it back in

AndyLnd:
When the database content gets JSON.stringify-ed it cannot stringify Uint8Arrays and keep the types. So we have to cast them as such when importing them. Otherwise they won't work anymore.

atomrc:
Alright, thanks for the explanation @AndyLnd :+1:

app/script/backup/BackupRepository.js
```diff
@@ -138,6 +138,18 @@ z.backup.BackupRepository = class BackupRepository {
       );
     });
 
+    const mapEntityDataTypes = entity => {
+      const mappedFields = ['otr_key', 'sha256'];
+      if (entity.data) {
+        mappedFields.forEach(field => {
+          if (entity.data[field]) {
+            entity.data[field] = Uint8Array.from(Object.values(entity.data[field]));
```
gregor: Please check that Object.values is supported by our browser range. I recall that we had an issue with this and outdated wrappers in the wild at some point in the past.

app/script/backup/BackupRepository.js
```diff
@@ -138,6 +138,18 @@ z.backup.BackupRepository = class BackupRepository {
       );
     });
 
+    const mapEntityDataTypes = entity => {
+      const mappedFields = ['otr_key', 'sha256'];
```
gregor: I suppose we can move these out of here into a static config const.

app/script/backup/BackupRepository.js
AndyLnd: Good call. I will check.

PR #3229, opened 2018-04-24
fix: compress export archive file (WEBAPP-5008)

atomrc:
## Type of change

Use the `deflate` algorithm to compress the export archive file

### DIFF
--- a/app/script/view_model/content/HistoryExportViewModel.js
+++ b/app/script/view_model/content/HistoryExportViewModel.js
@@ -86,7 +86,7 @@ z.viewModel.content.HistoryExportViewModel = class HistoryExportViewModel {
       this.numberOfProcessedRecords(0);
       this.backupRepository
         .generateHistory(this.onProgress.bind(this))
-        .then(archive => archive.generateAsync({type: 'blob'}))
+        .then(archive => archive.generateAsync({compression: 'DEFLATE', type: 'blob'}))
         .then(archiveBlob => {
           const timestamp = new Date().toISOString().substring(0, 10);
           const filename = `Wire-${userName}-Backup_${timestamp}.${


app/script/view_model/content/HistoryExportViewModel.js
```diff
@@ -86,7 +86,7 @@ z.viewModel.content.HistoryExportViewModel = class HistoryExportViewModel {
       this.numberOfProcessedRecords(0);
       this.backupRepository
         .generateHistory(this.onProgress.bind(this))
-        .then(archive => archive.generateAsync({type: 'blob'}))
+        .then(archive => archive.generateAsync({compression: 'DEFLATE', type: 'blob'}))
```
bennycode, line 89: Which other compressions are there? Can we put them into a constant?

AndyLnd: Does this mean, that before this it wasn't compressed at all? The default value seems to be just 'STORE'.

atomrc: Only `STORE` (no compression) and `DEFLATE`
https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#compression-and-compressionoptions-options

atomrc: A constant would feel a bit overkill for the moment (since it's not used anywhere else). But it's open for debate :)

atomrc: @AndyLnd yes precisely :grimacing: @mythsunwind discovered this while comparing the backup he was generating VS. the backup generated by the app

bennycode: @atomrc Ok, I agree with you.

PR #3243, opened 2018-04-26
fix: database deletion issues (WEBAPP-5037, WEBAPP-5039) , false negative error on import

AndyLnd:


atomrc:
Could we also add some logger when the export/import fails with an error we haven't created?
This would ease future debugs a lot

gregor:
@atomrc I added logs to my PR

PR #3245, opened 2018-04-26
fix: Chain and wait for backup import progress (WEBAPP-5009)

gregor:


app/script/backup/BackupRepository.js
```diff
@@ -132,45 +138,67 @@ z.backup.BackupRepository = class BackupRepository {
...
+      return fileDescriptor.filename === BackupRepository.CONFIG.FILENAME.EVENTS;
+    });
 
-      const importPromises = fileDescriptors.map(fileDescriptor => {
-        if (this.isCanceled) {
-          throw new z.backup.CancelError();
-        }
-        const tableName = fileDescriptor.filename.replace('.json', '');
-        const isConversationsTable = tableName === z.storage.StorageSchemata.OBJECT_STORE.CONVERSATIONS;
+    return this._importHistoryConversations(conversationData, progressCallback).then(() => {
```
atomrc: maybe `conversationsFileDescriptor` would be semantically more correct than `conversationsData`

app/script/backup/BackupRepository.js
```diff
@@ -132,45 +138,67 @@ z.backup.BackupRepository = class BackupRepository {
...
-        return importPromise.then(progressCallback);
-      });
+  _importHistoryEvents(eventData, progressCallback) {
+    if (this.isCanceled) {
+      return Promise.reject(new z.backup.CancelError());
+    }
 
-      return Promise.all(importPromises);
+    const entities = JSON.parse(eventData.content);
+    return this.backupService.importEntities(this.EVENTS_STORE_NAME, entities).then(() => {
```
atomrc: I feel like this `importEntities` method is now somewhat useless since it cannot handle conversations properly.

So two options here:
- either `importEntities` is renamed `importEvents` and the semantic is now correct;
- either `importEntities` knows how to handle `conversations` and we can refactor a bit of code from the `backupRepository` (cause it won't have to branch out anymore)

atomrc: my point being, if another developer wants to use this method latter on, it won't work as expected

app/script/backup/BackupRepository.js
```diff
@@ -132,45 +138,67 @@ z.backup.BackupRepository = class BackupRepository {
...
-      return Promise.all(
-        Object.values(archive.files)
-          .filter(zippedFile => zippedFile.name !== BackupRepository.CONFIG.META_FILENAME)
-          .map(zippedFile => zippedFile.async('string').then(value => ({content: value, filename: zippedFile.name})))
-      );
+    return this.verifyMetadata(files)
+      .then(() => this._extractHistoryFiles(files))
+      .then(fileDescriptors => this._importHistoryData(fileDescriptors, initCallback, progressCallback))
+      .catch(error => {
+        this.logger.error(`Failed to import history: ${error.message}`, error);
```
atomrc: We are missing some generic loggers that catch all the errors.
We can easily add them on the `HistoryImportViewModel` and `HistoryExportViewModel` (in the `onError` callback)

gregor: Will add them there as well.

PR #3255, opened 2018-04-27
fix: fix unread state in DB (WEBAPP-5040)

atomrc:
## Type of change

Update timestamps of the conversations in DB


app/script/conversation/ConversationRepository.js
```diff
@@ -604,23 +601,23 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
    *
    * @note To reduce the number of backend calls we merge the user IDs of all conversations first.
-   * @private
-   * @param {Array<Conversation>} conversation_ets - Array of conversation entities to be updated
+   * @param {Array<Conversation>} conversationEntities - Array of conversation entities to be updated
    * @returns {undefined} No return value
    */
-  _update_conversations(conversation_ets) {
-    const user_ids = _.flatten(conversation_ets.map(conversation_et => conversation_et.participating_user_ids()));
+  updateConversations(conversationEntities) {
```
atomrc, line 614: @gregor Although I really don't like the `_` notation at least it gives a hint to the consumer that this function is not supposed to be used from outside of the class.
Is there a reason for removing this `_`?

gregor: Yes, we now use it from outside the class.

atomrc: :+1:

PR #3263, opened 2018-04-27
fix: Catch localStorage access for disabled cookie scenario (WEBAPP-4988)

gregor:


app/script/auth/configureStore.js
```diff
@@ -45,9 +45,20 @@ export const configureStore = (thunkArguments = {}) => {
 
 const createMiddleware = thunkArguments => {
   const middlewares = [thunk.withExtraArgument(thunkArguments)];
-  window.localStorage.removeItem('debug');
+
+  let localStorage;
+  try {
+    localStorage = window.localStorage;
+  } catch (error) {}
```
atomrc, line 52: One cool thing that can be done with localStorage is to fallback to a in-memory localStorage (like https://www.npmjs.com/package/localstorage-memory).
So in the `catch` fallback, you can actually do `localStorage = inMemoryLocalStorage` and you don't need those `if (localStorage)` afterward.

But that might not be the scope of this PR at all, I am not sure

gregor: To give some additional context to this PR:

Browsers throw an error if you access `window.localStorage` if cookies are disabled. All we intend to do is bypass these errors so that the react app can initialize to the point where we can show the warning page that Wire will not work as long as cookies are disabled. Currently these lines throw an error and a blank page is shown.

gregor:
https://github.com/dfahlander/Dexie.js/issues/617

PR #3272, opened 2018-04-30
Fix: (import) reactivate deleted conversation according to the backup state (WEBAPP-5043)

atomrc:


app/script/entity/Conversation.js
```diff
@@ -308,8 +309,9 @@ z.entity.Conversation = class Conversation {
         break;
     }
 
-    const updated_timestamp = this._increment_time_only(entity_timestamp(), timestamp);
-    if (updated_timestamp) {
+    const updated_timestamp = allowDecrement ? timestamp : this._increment_time_only(entity_timestamp(), timestamp);
```
lipis: camelCase for new variables.

lipis: or updated ones :P

app/script/entity/Conversation.js
```diff
@@ -308,8 +309,9 @@ z.entity.Conversation = class Conversation {
...
+
+    if (updated_timestamp !== false) {
```
lipis:
```js
if (updated_timestamp) {
```

We should stop comparing with `== false`, `== true`

lipis: /cc @atomrc

atomrc: The problem is that `0` is a falsy value, but in our case it's actually what we want (to set 0 there)

app/script/conversation/ConversationMapper.js
```diff
@@ -100,8 +100,8 @@ z.conversation.ConversationMapper = class ConversationMapper {
         conversation_et.archived_state(self_state.archived_state);
       }
 
-      if (cleared_timestamp) {
-        conversation_et.set_timestamp(cleared_timestamp, z.conversation.TIMESTAMP_TYPE.CLEARED);
+      if (cleared_timestamp !== undefined) {
```
lipis, line 103: can that be anything else than the actual timestamp? Can we avoid comparing with `!== undefined`?

app/script/entity/Conversation.js
atomrc: So we should limit the scope of our comparison matching against the value returned by `_increment_time_only`

app/script/conversation/ConversationMapper.js
atomrc: Same as before. It can be `0` which is valid in our case

atomrc:
@gregor I think a review from you can be valuable since it touches some low level logic

app/script/entity/Conversation.js
```diff
@@ -277,9 +277,10 @@ z.entity.Conversation = class Conversation {
    * @note This will only increment timestamps
    * @param {string|number} timestamp - Timestamp to be set
    * @param {z.conversation.TIMESTAMP_TYPE} type - Type of timestamp to be updated
+   * @param {boolean} allowDecrement - Allow setting a timestamp older than the current timestamp
    * @returns {boolean|number} Timestamp value which can be 'false' (boolean) if there is no timestamp
    */
-  set_timestamp(timestamp, type) {
+  set_timestamp(timestamp, type, allowDecrement = false) {
```
gregor: I am not fully sure on the naming. Wdyt about `forceUpdate` maybe?

atomrc: `forceUpdate` LGTM

PR #3273, opened 2018-04-30
fix: Hide canceled and blocked connections on import (WEBAPP-5058, WEBAPP-5059)

atomrc:
## Type of change

Imported conversations needed to be mapped with the current connections.


app/script/backup/BackupRepository.js
```diff
@@ -213,7 +213,11 @@ z.backup.BackupRepository = class BackupRepository {
 
     return this._importHistoryConversations(conversationEntities, progressCallback)
       .then(importedEntities => this._importHistoryEvents(eventEntities, progressCallback).then(() => importedEntities))
-      .then(importedEntities => this.conversationRepository.updateConversations(importedEntities));
+      .then(importedEntities => {
+        this.conversationRepository.updateConversations(importedEntities);
+        this.conversationRepository.map_connections(this.userRepository.connections());
+        return importedEntities;
+      });
```
gregor: Looking at the previous code we did not return the `importedEntities`. Adding that for a particular reason?

atomrc: We initially returned them in the promise of `updateConversation`. 
But now I realize that my code broke the previous behavior. I'll need to use a `Promise.all`, so nice catch!

PR #3277, opened 2018-05-02
feat: Refining export process UX (WEBAPP-5060)

atomrc:


app/script/view_model/content/HistoryExportViewModel.js
```diff
@@ -82,27 +90,35 @@ z.viewModel.content.HistoryExportViewModel = class HistoryExportViewModel {
   exportHistory() {
     this.state(HistoryExportViewModel.STATE.PREPARING);
     this.hasError(false);
-    this.backupRepository.getBackupInitData().then(({numberOfRecords, userName}) => {
+    this.backupRepository.getBackupInitData().then(({numberOfRecords}) => {
```
gregor: Let's completely remove the username if we no longer need it here.

app/script/view_model/content/HistoryExportViewModel.js
```diff
@@ -122,9 +138,10 @@ z.viewModel.content.HistoryExportViewModel = class HistoryExportViewModel {
     amplify.publish(z.event.WebApp.ANALYTICS.EVENT, z.tracking.EventName.HISTORY.BACKUP_FAILED);
   }
 
-  onSuccess(data) {
+  onSuccess(archiveBlob) {
     this.state(HistoryExportViewModel.STATE.DONE);
     this.hasError(false);
+    this.archiveBlob(archiveBlob);
     amplify.publish(z.event.WebApp.ANALYTICS.EVENT, z.tracking.EventName.HISTORY.BACKUP_SUCCEEDED);
```
gregor, line 145: @EvaZg Can you please check the slightly revised flow that requires a user click action to save the created backup. OK to trigger the 'BACKUP_SUCCEEDED' event before that when the file is ready for download but we are waiting for the user action? Even if we trigger it on the click the user can still decide to not save it. Thus a succeed does not ensure actual success.

PR #3278, opened 2018-05-02
fix: Use theme colors for export cancel buttons

atomrc:


atomrc:
@gregor nice catch. Done

PR #3281, opened 2018-05-02
fix: Wrap all accesses to indexedDB in try-catch blocks (WEBAPP-4988)

atomrc:
## Type of change

Adds missing check around `indexedDB` access

## Important note

This, alone, will not fix `WEPAPP-4988`. It's still missing a patch from `dexie`

### DIFF
--- a/app/script/auth/module/action/RuntimeAction.js
+++ b/app/script/auth/module/action/RuntimeAction.js
@@ -67,7 +67,13 @@ function hasCookieSupport() {
 }
 
 function hasIndexedDbSupport() {
-  if (!window.indexedDB) {
+  let supportIndexedDb;
+  try {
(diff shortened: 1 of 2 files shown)


app/script/auth/module/action/RuntimeAction.js
```diff
@@ -67,7 +67,13 @@ function hasCookieSupport() {
 }
 
 function hasIndexedDbSupport() {
-  if (!window.indexedDB) {
+  let supportIndexedDb;
+  try {
+    supportIndexedDb = !!window.indexedDb;
+  } catch (e) {
```
gregor: Let's not use one char names. Even for this one. Characters are free.

gregor:
https://github.com/dfahlander/Dexie.js/pull/619#pullrequestreview-116916013

PR #3282, opened 2018-05-02
demo: use patch-package to patch dexie module (WEPAPP-4988)

atomrc:


atomrc:
One problem to note:

I needed to remove the `"postinstall": "grunt init"` line in the `package.json` that would break the patch generation (because the gruntfile isn't copied to the tmp location where the patch is created)

gregor:
@atomrc Thanks for the demo. I will go ahead and close this now as we have a new dexie release that fixes the addressed bug.

PR #3290, opened 2018-05-03
fix: Empty call log warning (WEBAPP-4872)

gregor:


app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -72,7 +72,9 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
 
   saveCallLogs() {
     const messageLog = this.callingRepository.messageLog;
-    if (messageLog.length) {
+    // Very short logs will not contain useful information
+    const logContainsCall = messageLog.length > 10;
+    if (logContainsCall) {
```
lipis: maybe including count in the variable name.. what is this length actually represents?

app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -81,8 +83,10 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
     }
 
     amplify.publish(z.event.WebApp.WARNING.MODAL, z.viewModel.ModalsViewModel.TYPE.ACKNOWLEDGE, {
-      message: z.l10n.text(z.string.modalCallEmptyLogMessage),
-      title: z.l10n.text(z.string.modalCallEmptyLogHeadline),
+      text: {
```
atomrc, line 92: I am kind of surprised that the payload changed, but nothing changed on the consumer's side. 
Any clue on this?

gregor: I just made a mistake the first time around. This is the proper way to use our warning modals.

app/script/view_model/content/PreferencesOptionsViewModel.js
gregor: Adjusted it and got rid of the magic number.

PR #3312, opened 2018-05-07
test: markdown headers are ignored

atomrc:


test/unit_tests/util/UtilSpec.js
```diff
@@ -4270,6 +4270,12 @@ describe('Ignored Markdown syntax', () => {
     );
   });
 
+  it('does not render header', () => {
```
bennycode: `'does not render header'` -> `'does not render headers'`

bennyn:
This is superseeded by https://github.com/wireapp/wire-webapp/pull/3312, no?

PR #3324, opened 2018-05-08
feat: Add support for video in group calls

gregor:


app/script/calling/CallingRepository.js
```diff
@@ -1105,14 +1184,40 @@ z.calling.CallingRepository = class CallingRepository {
...
+
   //##############################################################################
   // call entity creation
   //##############################################################################
 
   /**
    * Constructs a call entity.
    *
-   * @private
+   * @privatemessage to conversation
```
ffflorian:
```js
* @private
```

gregor: Fixed

PR #3341, opened 2018-05-11
fix: Ensure notifications are always properly removed (WEBAPP-5075)

pwnsdx:
In some cases, some messages could stay in the notification center. This does not affect ephemeral messages.


app/script/notification/NotificationRepository.js
```diff
@@ -153,25 +152,6 @@ z.notification.NotificationRepository = class NotificationRepository {
...
-            const messageInfo = messageId
-              ? `message '${messageId}' of type '${messageType}'`
-              : `'${messageType}' message`;
-            this.logger.info(`Removed read notification for ${messageInfo} in '${conversationId}'.`);
-          }
-        });
-      }
-    });
-  }
-
```
gregor: Why do we want to remove this? This ensure a closure of notifcations when a conversation was marked as read on other devices etc.

PR #3362, opened 2018-05-16
feat: video calling container styling

atomrc:


app/style/video-calling.less
```diff
@@ -38,6 +38,16 @@
   border-radius: 4px;
 }
 
+.video-calling {
+  &--active {
+    position: absolute;
+    z-index: 101;
```
gregor, line 44: Let's stick to our variables for this. We already noticed that the guest label is erroneously shown above the video calling ui.

gregor: The `.full-screen` was also a good thing.

PR #3363, opened 2018-05-16
feat: adapt video size and ratio when in a group call

atomrc:


app/script/media/MediaConstraintsHandler.js
```diff
@@ -177,7 +177,7 @@ z.media.MediaConstraintsHandler = class MediaConstraintsHandler {
       }
 
       case z.media.VIDEO_QUALITY_MODE.GROUP: {
-        media_stream_constraints = MediaConstraintsHandler.CONFIG.VIDEO_CONSTRAINTS.FULL_HD;
+        media_stream_constraints = MediaConstraintsHandler.CONFIG.VIDEO_CONSTRAINTS.GROUP;
```
gregor, line 180: Good catch

atomrc: That was an easy win :laughing:

PR #3387, opened 2018-05-18
feat: minimize video call to call ui

AndyLnd:


app/script/components/list/conversationListCallingCell.js
```diff
@@ -57,6 +60,10 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
       amplify.publish(z.event.WebApp.CALL.MEDIA.TOGGLE, this.conversation.id, z.media.MediaType.VIDEO);
     };
 
+    this.onMaximizeVideoGrid = () => {
+      wire.app.view.content.multitasking.isMinimized(false);
```
atomrc: Do you think we can inject this service instead of relying on global stuffs ?

app/script/view_model/VideoCallingViewModel.js
```diff
@@ -53,14 +53,10 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
...
-      // FIXME the media repository should release the memory of inactive video streams
-      // Right now, only the status is changed but we keep the stream in memory
-      // until the next page reload.
-      let videoStreams = this.mediaRepository.stream_handler.remote_media_streams
-        .video()
-        .filter(stream => stream.active);
+      let videoStreams = streamHandler.remote_media_streams.activeVideo();
 
       if (videoStreams.length > 1) {
         videoStreams = videoStreams.concat(this.localVideoStream());
```
atomrc: Woups, this should be removed I believe. 
We are doing this twice (in the grid and in the callingviewmodel).

app/script/view_model/VideoCallingViewModel.js
```diff
@@ -184,12 +145,12 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
...
           this.visibleCallId = callEntity.id;
 
-          if (this.showLocalVideo() || this.showRemoteVideo()) {
+          if (this.showRemoteVideo()) {
             this.multitasking.isMinimized(false);
             return this.logger.info(`Maximizing video call '${callEntity.id}' to full-screen`, callEntity);
           }
 
-          this.multitasking.isMinimized(true);
+          //this.multitasking.isMinimized(true);
```
atomrc, line 155: @AndyLnd arf, I believe we have to handle that case. I couldn't really understand what this was about but that condition broke (it minimized the call every time).

Maybe we don't need this anymore. What do you think?

PR #3407, opened 2018-05-22
chore: Update translations

Yserz:


gregor:
Missing one last change before it is good to go.

PR #3411, opened 2018-05-22
feat: add group size info

AndyLnd:


app/script/localization/webapp.js
```diff
@@ -372,6 +372,7 @@ z.string.groupCreationParticipantsActionSkip = 'Skip';
 z.string.groupCreationParticipantsHeader = 'Add people';
 z.string.groupCreationParticipantsHeaderWithCounter = 'Add people ({{number}})';
 z.string.groupCreationParticipantsPlaceholder = 'Search by name';
+z.string.groupSizeInfo = 'Up to 128 people can join a group conversation. Video calls work in groups of 4 or less.';
```
gregor, line 375: @AndyLnd Let's rely on `z.conversation.ConversationRepository.GROUP.MAX_SIZE` and the corresponding value for group video calls to create this string so we will never miss to update it.

AndyLnd: https://github.com/wireapp/wire-webapp/pull/3414

PR #3415, opened 2018-05-22
chore: remove redundant search params after login (WEBAPP-4939)

AndyLnd:


app/script/auth/util/urlUtil.js
```diff
@@ -17,11 +17,16 @@
  *
  */
 
-export function pathWithParams(path, additionalParams) {
+import EXTERNAL_ROUTE from '../externalRoute';
+
+export function pathWithParams(path, additionalParams, excludeParams = []) {
```
gregor: Shouldn't we do a whitelist approach here? Only specific params will need to be forwarded. Otherwise we need to remember to add them to the blacklist each time.

gregor: https://github.com/wireapp/wire-webapp/blob/dev/app/script/view_model/AuthViewModel.js#L29-L35

PR #3434, opened 2018-05-23
feat: Integrated Dark Theme

Stainzl:
## Type of change

I integrated a darktheme which is enabled in the settings.

## Screenshot

kontakt
konto
settings


gregor:
@Stainzl First of all thank you very much for your amazing contribution. We have heard the user feedback regarding this feature request loud and clear. But after long internal discussions we have previously come to the conclusion that this improvement is not on our immediate roadmap.

This assessment has not changed. Adding a feature to the product is more complicated than just merging it into the codebase. We would need to allocate the resources needed to maintain this feauture in the long run. Our customer support needs to incorporate into all user communications,, we have to think about it with every design change moving forward, and maintain it with our codebase. While we greatlxy appreciate your effort by opening the PR we will not merge it and I am going to close your contribution instead.

This is in line with our statement in our contribution guidelines (https://github.com/wireapp/wire#contributing-to-the-code). Changes to the codebase like the one here are not accepted at this time.

Thank you very much again for your effort.

PR #3436, opened 2018-05-23
fix: Remove username from call log filename

pwnsdx:


app/script/view_model/content/PreferencesOptionsViewModel.js
```diff
@@ -79,9 +79,7 @@ z.viewModel.content.PreferencesOptionsViewModel = class PreferencesOptionsViewMo
     if (logExceedsMinimumLength) {
       const callLog = [messageLog.join('\r\n')];
       const blob = new Blob(callLog, {type: 'text/plain;charset=utf-8'});
-
-      const userName = this.userRepository.self().username();
-      const filename = `Wire-${userName}-Calling_${z.util.TimeUtil.getCurrentDate()}.log`;
+      const filename = `Wire-Calling_${z.util.TimeUtil.getCurrentDate()}.log`;
```
gregor: This was introduced on purpose. If we get logs from an internal group call, we need to be able to see, which log is from which user. If public users send in logs, the name is immediately visible and they can rename the file to their liking.

gregor:
Superseeded by #3488

PR #3443, opened 2018-05-24
Call ui animations

AndyLnd:


app/style/components/group-video-grid.less
```diff
@@ -27,6 +27,7 @@
     height: 168px;
     margin: 0 8px 16px;
     border-radius: 8px;
+    //animation: show-video-preview @animation-timing-fast ease-in-out 0s 1;
```
atomrc: Maybe we can remove if it's not used

PR #3447, opened 2018-05-25
feat: switches for input devices

AndyLnd:


app/page/template/video-calling.htm
```diff
@@ -51,8 +51,16 @@
...
-                <div class="video-controls__button__label" data-bind="l10n_text: z.string.videoCallOverlayVideo"></div>
-
+                <!-- ko if: showSwitchCamera() -->
+                  <device-toggle-button data-bind="click: clickedOnToggleCamera"
+                    params="index: currentDeviceIndex.video_input,
+                      devices: availableDevices.video_input,
+                      type: z.media.MediaDeviceType.VIDEO_INPUT">
+                  </device-toggle-button>
+                <!-- /ko -->
+                <!-- ko ifnot: showSwitchCamera -->
```
atomrc: Just for consistency, can we try to call the `showSwitchCamera` observable always the same way.
Either `showSwitchCamera()` or `showSwitchCamera`.

atomrc: Maybe `showSwitchCamera()` would be better, so that we don't, when reading the code, testing that the observable exists with testing the actual value of the observable.

app/script/view_model/VideoCallingViewModel.js
```diff
@@ -125,6 +125,17 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
       }
     });
 
+    this.showSwitchCamera = ko.pureComputed(() => {
+      const hasMultipleCameras = this.availableDevices.video_input().length > 1;
+      const isVisible = hasMultipleCameras && this.localVideoStream() && this.selfStreamState.videoSend();
+      return this.isCallOngoing() && isVisible;
+    });
+    this.showSwitchScreen = ko.pureComputed(() => {
+      const hasMultipleCameras = this.availableDevices.screen_input().length > 1;
```
atomrc: `hasMultipleCameras` => `hasMultipleScreens`

PR #3462, opened 2018-05-28
fix: improve video scaling

atomrc:


app/script/components/groupVideoGrid.js
```diff
@@ -58,31 +58,59 @@ z.components.GroupVideoGrid = (() => {
...
+
       this.minimized = params.minimized;
     }
 
-    scaleVideos(elements) {
+    scaleVideos(rootElement) {
+      const elements = Array.from(rootElement.querySelectorAll('.group-video-grid__element'));
       elements
         .filter(element => !!element.classList)
         .filter(element => element.classList.contains('group-video-grid__element'))
```
AndyLnd: No need to filter them by there class name as you already selected only those with that class.

app/script/components/groupVideoGrid.js
```diff
@@ -58,31 +58,59 @@ z.components.GroupVideoGrid = (() => {
...
+      this.scaleVideos = this.scaleVideos.bind(this, rootElement);
+      // scale videos when the grid is updated (on the next rendering cycle)
+      this.grid.subscribe(() => {
+        window.requestAnimationFrame(this.scaleVideos);
+      });
+
+      // scale the videos when the window is resized
+      window.addEventListener('resize', this.scaleVideos);
+
+      this.rootElement = rootElement;
```
AndyLnd: you already bound it to the function, do you still need it somewhere else?

atomrc: very good point. It was needed for the previous iteration and I forgot to remove it :D 
:+1:

app/script/components/groupVideoGrid.js
atomrc: Very true

app/script/components/groupVideoGrid.js
```diff
@@ -197,8 +220,8 @@ z.components.GroupVideoGrid = (() => {
 
   ko.components.register('group-video-grid', {
     template: `
-      <div class="group-video">
-        <div class="group-video-grid" data-bind="foreach: { data: grid, as: 'streamId', afterRender: scaleVideos}">
+      <div class="group-video" data-bind="template: { afterRender: scaleVideos }">
+        <div class="group-video-grid" data-bind="foreach: { data: grid, as: 'streamId' }">
```
gregor, line 224: Too much whitespace:

````
<div class="group-video" data-bind="template: {afterRender: scaleVideos}">
  <div class="group-video-grid" data-bind="foreach: {data: grid, as: 'streamId'}">
````

PR #3470, opened 2018-05-29
feat: add the mute overlay over self video

atomrc:


app/style/components/group-video-grid.less
```diff
@@ -121,24 +121,44 @@
...
       right: 8px;
       bottom: 8px;
       width: 80px;
-      border-radius: 4px;
       box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.16);
     }
+
+    &-video {
+      width: 100%;
+      border-radius: 4px;
```
AndyLnd: does thia mean that only the video will have the border-radius but not the overlay?

PR #3489, opened 2018-05-30
fix: unarchive conversations with call (WEBAPP-5105)

AndyLnd:


app/script/conversation/ConversationRepository.js
```diff
@@ -167,6 +167,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
 
       this.sorted_conversations().forEach(conversationEntity => {
         if (conversationEntity.has_active_call() && !conversationEntity.hasActiveDeclinedCall()) {
+          this.unarchive_conversation(conversationEntity, 'Call on conversation');
```
gregor: I suppose this should be call in conversation

PR #3494, opened 2018-05-31
fix: Avoid session creation race condition in encrytion/decryption (WEBAPP-5109)

gregor:


app/script/cryptography/CryptographyRepository.js
```diff
@@ -242,46 +242,23 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
...
-
-          payload.recipients[userId][clientId] = cipherText;
-        });
-
-        return this._encryptGenericMessageForNewSessions(recipientsWithMissingSessions, genericMessage);
+    return this._encryptGenericMessage(recipients, genericMessage, payload)
+      .then(({messagePayload, missingRecipients}) => {
+        return Object.keys(missingRecipients).length
+          ? this._encryptGenericMessageForMissingRecipients(missingRecipients, genericMessage, messagePayload)
+          : {messagePayload};
```
atomrc: I would suggest to add an empty object with the key `missingRecipients` so that it comply with the signature of the next function call

atomrc: Actually, we can just send the `missingRecipients` object as part of the return.

```
{messagePayload, missingRecipients}
```

gregor: Good point

app/script/cryptography/CryptographyRepository.js
```diff
@@ -355,36 +332,71 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
...
+          clientIds.forEach(clientId => {
+            const sessionId = this._constructSessionId(userId, clientId);
+            cipherPayloadPromises.push(this._encryptPayloadForSession(sessionId, genericMessage));
+          });
+        }
 
-          for (const userId in userPreKeyMap) {
-            const clientPreKeyMap = userPreKeyMap[userId];
+        const receivingUsers = Object.keys(messagePayload.recipients).length;
+        const logMessage = `Encrypting message of type '${genericMessage.content}' for '${receivingUsers}' users.`;
```
bennycode: We should have the same log messages also for `_encryptGenericMessageForMissingRecipients`.

PR #3504, opened 2018-06-01
fix: do not show the minimized grid if a call is already ongoing

atomrc:


app/script/components/list/conversationListCallingCell.js
```diff
@@ -143,7 +143,15 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
     this.showParticipants = ko.observable(false);
 
     this.showVideoPreview = ko.pureComputed(() => {
-      return this.isVideoCall() && (this.multitasking.isMinimized() || !this.callIsConnected());
+      const nbOfOtherOngoingCalls = this.calls()
```
gregor: We do not care about the number. You can just turn this into a boolean. Much easier on the naming as you can go with `hasOtherOngoingCall`

atomrc: Fair point :+1:

app/script/components/list/conversationListCallingCell.js
```diff
@@ -143,7 +143,14 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
...
     this.showVideoPreview = ko.pureComputed(() => {
-      return this.isVideoCall() && (this.multitasking.isMinimized() || !this.callIsConnected());
+      const hasOtherOngoingCalls =
+        this.calls()
+          .filter(call => call.id !== this.call().id)
+          .filter(call => call.state() === z.calling.enum.CALL_STATE.ONGOING).length > 0;
+
+      return (
+        !hasOtherOngoingCalls && this.isVideoCall() && (this.multitasking.isMinimized() || !this.callIsConnected())
+      );
```
gregor: May I ask you to pull the last `or` bit out into a separats variable and use that in the return. I suppose the return fits into a single line then.

atomrc: There you go :)

PR #3505, opened 2018-06-01
chore: have video grid in own repo

AndyLnd:


gregor:
As stated before. To me this is not business logic but a view state. Especially in our current knockout setup. This should not be handled in a new repo but be part of either the VideoCallingViewModel or a new VideoGridViewModel.

/cc @bennyn Wdyt architecturally?

app/script/calling/VideoGridRepository.js
```diff
@@ -0,0 +1,114 @@
...
+     * @param {CallingRepository} callingRepository - Repository for the calls
+     */
+    constructor(mediaRepository, callingRepository) {
+      const streamHandler = mediaRepository.stream_handler;
+      const streamsInfo = streamHandler.remoteMediaStreamInfoIndex.video;
+      const localMediaStream = streamHandler.localMediaStream;
+      const selfStreamState = callingRepository.selfStreamState;
+      const calls = callingRepository.calls;
+      this.grid = ko.observableArray([0, 0, 0, 0]);
+      this.selfId = ko.observable(null);
```
bennycode: To be consistent with the rest of the code base we should use `ko.observable(undefined)` here.

app/script/calling/VideoGridRepository.js
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @returns {Array<string|0>} the new grid
+     */
+    computeGrid(previousGrid, streams) {
+      const previousStreamIds = previousGrid.filter(streamId => streamId !== 0);
+      const currentStreamIds = streams.map(participant => participant.id);
+
+      const addedStreamIds = z.util.ArrayUtil.getDifference(previousStreamIds, currentStreamIds);
+
+      const filteredGrid = previousGrid.map(id => (currentStreamIds.includes(id) ? id : 0));
```
bennycode, line 83: `const filteredGrid = previousGrid.map(id => (currentStreamIds.includes(id) ? id : 0));` ->`const filteredGrid = previousGrid.map(id => currentStreamIds.includes(id) ? id : 0);`

AndyLnd: I fully agree, but prettier likes it the other way :/

app/script/calling/VideoGridRepository.js
AndyLnd: As discussed before, I am strongly of the opinion that ```undefined``` shouldn't be set explicitly if not absolutely necessary. 
I can change this occurrence now (to be consistent), but would really consider changing this behaviour throughout the codebase in the near future.

app/script/calling/VideoGridRepository.js
```diff
@@ -0,0 +1,114 @@
...
+window.z.calling = z.calling || {};
+
+z.calling.VideoGridRepository = (() => {
+  return class VideoGridRepository {
+    /**
+     * Construct an new VideoGridRepository.
+     * @param {MediaRepository} mediaRepository - Repository for the media streams
+     * @param {CallingRepository} callingRepository - Repository for the calls
+     */
+    constructor(mediaRepository, callingRepository) {
```
gregor: We usually do this in alphabetical ordeer

app/script/calling/VideoGridRepository.js
```diff
@@ -0,0 +1,114 @@
...
+      const streamsInfo = streamHandler.remoteMediaStreamInfoIndex.video;
+      const localMediaStream = streamHandler.localMediaStream;
+      const selfStreamState = callingRepository.selfStreamState;
+      const calls = callingRepository.calls;
+      this.grid = ko.observableArray([0, 0, 0, 0]);
+      this.selfId = ko.observable(null);
+      this.thumbnailStream = ko.observable(null);
+      this.mirrorSelf = ko.pureComputed(() => {
+        return !selfStreamState.screenSend();
+      });
```
gregor: ` this.mirrorSelf = ko.pureComputed(() => !selfStreamState.screenSend());`

app/script/components/groupVideoGrid.js
```diff
@@ -22,189 +22,131 @@
...
+ko.components.register('group-video-grid', {
+  template: `
       <div class="group-video" data-bind="template: {afterRender: scaleVideos}">
         <div class="group-video-grid" data-bind="foreach: {data: grid, as: 'streamId'}">
           <!-- ko if: streamId !== 0 -->
             <div class="group-video-grid__element" data-bind="css: $parent.getClassNameForVideo($index()), attr: {'data-uie-name': 'item-grid', 'data-uie-value': $parent.getUIEValueForVideo($index())}">
               <video autoplay class="group-video-grid__element-video" data-bind="sourceStream: $parent.getParticipantStream(streamId), muteMediaElement: $parent.getParticipantStream(streamId)">
               </video>
-              <!-- ko if: streamId === $parent.selfId() && !$parent.selfStreamState.audioSend() -->
+              <!-- ko if: streamId === $parent.selfId() && $parent.selfStreamMuted -->
```
gregor: Isn't `$parent.selfStreamMuted` an observable? I think in this case with the concatenated ko if you explicitly have to add (). At least it won't hurt to be more specific.

PR #3522, opened 2018-06-04
fix: request media on incoming call (WEBAPP-5113)

atomrc:
Turned out that to be able to mute a channel, the media channel needs to be requested first.
This allows the user to toggle any media channel off before picking up.

This changes the global behavior of the app. Now, when there is an incoming call, we request the userMedia (same behavior as video calls)

### DIFF
--- a/app/script/calling/CallingRepository.js
+++ b/app/script/calling/CallingRepository.js
@@ -1283,9 +1283,10 @@ z.calling.CallingRepository = class CallingRepository {
           this.injectActivateEvent(callMessageEntity, source);
 
           const eventFromWebSocket = source === z.event.EventRepository.SOURCE.WEB_SOCKET;
-          if (eventFromWebSocket && callEntity.isRemoteVideoSend()) {
-            const mediaStreamType = z.media.MediaType.AUDIO_VIDEO;
-            this.mediaStreamHandler.initiate_media_stream(callEntity.id, mediaStreamType, callEntity.isGroup);
+          if (eventFromWebSocket) {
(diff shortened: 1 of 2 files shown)


app/script/calling/CallingRepository.js
```diff
@@ -1283,9 +1283,10 @@ z.calling.CallingRepository = class CallingRepository {
           this.injectActivateEvent(callMessageEntity, source);
 
           const eventFromWebSocket = source === z.event.EventRepository.SOURCE.WEB_SOCKET;
-          if (eventFromWebSocket && callEntity.isRemoteVideoSend()) {
-            const mediaStreamType = z.media.MediaType.AUDIO_VIDEO;
-            this.mediaStreamHandler.initiate_media_stream(callEntity.id, mediaStreamType, callEntity.isGroup);
+          if (eventFromWebSocket) {
+            const fixedMediaType = mediaType === z.media.MediaType.VIDEO ? z.media.MediaType.AUDIO_VIDEO : mediaType;
+
+            this.mediaStreamHandler.initiate_media_stream(callEntity.id, fixedMediaType, callEntity.isGroup);
```
gregor: We do not want to automatically turn on the microphone on an incoming audio call. This is going to freak users out. If we need it to toggle off the mic before accepting a call, we need to make sure we save the state and properly set it after we request the stream.

atomrc: Ok, that's what I though, but then, it would also mean that have consistent behavior for audio and video calls.

I might need your help on this one then. Seems like the state is tied to the userMedia somehow, we might need some deep changes (not sure yet)

atomrc: So far, my investigations showed that the userMedia is the ground truth for the media state of the user

PR #3523, opened 2018-06-04
fix: keep video calling disabled in titlebar for non team users (WEBAPP-5115)

AndyLnd:


app/script/view_model/content/TitleBarViewModel.js
```diff
@@ -80,7 +80,9 @@ z.viewModel.content.TitleBarViewModel = class TitleBarViewModel {
     });
 
     this.supportsVideoCall = ko.pureComputed(() => {
-      return this.showCallControls() ? this.conversationEntity().supportsVideoCall(true) : false;
+      return this.showCallControls()
+        ? this.userRepository.self().inTeam() && this.conversationEntity().supportsVideoCall(true)
```
gregor: You can directly user `this.userRepository.isTeam()` instead of relying on the self user property.

app/script/view_model/content/TitleBarViewModel.js
```diff
@@ -80,7 +80,9 @@ z.viewModel.content.TitleBarViewModel = class TitleBarViewModel {
     });
 
     this.supportsVideoCall = ko.pureComputed(() => {
-      return this.showCallControls() ? this.conversationEntity().supportsVideoCall(true) : false;
+      return this.showCallControls()
+        ? this.userRepository.isTeam() && this.conversationEntity().supportsVideoCall(true)
+        : false;
```
gregor, line 85: Seems like this causes a regression. Probably it is best to move the `isTeam` check to `supportsVideoCall()` to check only in case of groups and base it on the self user there.

PR #3525, opened 2018-06-04
fix: Safeguard message encryption property existence

gregor:


app/script/cryptography/CryptographyRepository.js
```diff
@@ -345,13 +345,17 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
...
 
         for (const userId in recipients) {
-          const clientIds = recipients[userId];
-
-          messagePayload.recipients[userId] = messagePayload.recipients[userId] || {};
-          clientIds.forEach(clientId => {
-            const sessionId = this._constructSessionId(userId, clientId);
-            cipherPayloadPromises.push(this._encryptPayloadForSession(sessionId, genericMessage));
-          });
+          if (recipients.hasOwnProperty(userId)) {
```
atomrc: Probably using the `Object.keys` method to do this would be more efficient 
https://jsperf.com/object-keys-vs-hasownproperty/45

gregor: Valid point. We have moved to that in most cases. This is left over from the automated tools doing the heavy lifting during CoffeeScript to ES6 migration.

PR #3526, opened 2018-06-04
fix: don't show self user in freshly created conversations details (WEBAPP-5102)

AndyLnd:


app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -74,7 +74,9 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
             if (userEntity.isBot) {
               return this.serviceParticipants.push(userEntity);
             }
-            this.userParticipants.push(userEntity);
+            if (!userEntity.is_me) {
+              this.userParticipants.push(userEntity);
+            }
```
gregor: The self user should never be part of the `participating_user_ets` in the first place. This fixes the symptom but not the cause.

PR #3536, opened 2018-06-05
fix: Allow audio toggling on incoming audio call (WEBAPP-5113)

gregor:
Replaces #3522

### DIFF
--- a/app/script/media/MediaStreamHandler.js
+++ b/app/script/media/MediaStreamHandler.js
@@ -28,7 +28,7 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    * @param {MediaStream} mediaStream - MediaStream to detect the type off
    * @returns {MediaStream} MediaStream with new type information
    */
-  static detect_media_stream_type(mediaStream) {
+  static detectMediaStreamType(mediaStream) {
     const audioTracks = mediaStream.getAudioTracks();
     const videoTracks = mediaStream.getVideoTracks();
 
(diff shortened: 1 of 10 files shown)


app/script/media/MediaStreamHandler.js
```diff
@@ -545,53 +568,50 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
+    return this._toggleAudioSend();
   }
 
   // Toggle the screen.
-  toggle_screen_send() {
-    const hasActiveScreenStream = this.localMediaStream() && this.local_media_type() === z.media.MediaType.SCREEN;
-    return hasActiveScreenStream ? this._toggleScreenSend() : this.replace_input_source(z.media.MediaType.SCREEN);
+  toggleScreenSend() {
+    const hasActiveScreenStream = this.localMediaStream() && this.localMediaType() === z.media.MediaType.SCREEN;
+    return hasActiveScreenStream ? this._toggleScreenSend() : this.replaceInputSource(z.media.MediaType.SCREEN);
```
atomrc: I see this piece of code duplicated twice.
We could refactor this in another private function that takes a type and a toggle function parameter

```
toggleMediaState(type, toggleFn) {
  const hasActiveScreenStream = this.localMediaStream() && this.localMediaType() === type;
  return hasActiveScreenStream ? toggleFn() : this.replaceInputSource(type);
}
```

and then in each function

```
toggleScreenSend() {
  return this.toggleMediaState(SCREEN, this._toggleScreenSend.bind(this));
}
```

atomrc: Why exactly do we check the state of the mediaStream in the boolean toggling function?
Since this check is already (partially) done in the `_toggleStreamEnabled`, that should be enough, isn't it?

atomrc: If it's a matter of exposing a state that depends on both the internal state AND the real state of the mediaStream, then we should create a new observable that depends on both.
Something like

```
isAudioSent = ko.pureComputed(() => {
  return this.state.audioSend() && this.localMediaStream(AUDIO).enabled;
});
```

PR #3537, opened 2018-06-05
fix: allow video calls in personal one to one conversations (WEBAPP-5120)

AndyLnd:


app/script/view_model/content/TitleBarViewModel.js
```diff
@@ -80,9 +80,8 @@ z.viewModel.content.TitleBarViewModel = class TitleBarViewModel {
     });
 
     this.supportsVideoCall = ko.pureComputed(() => {
-      return this.showCallControls()
-        ? this.userRepository.isTeam() && this.conversationEntity().supportsVideoCall(true)
-        : false;
+      const userAllowedVideoCall = this.userRepository.isTeam() || this.conversationEntity().is_one2one();
+      return this.showCallControls() && this.conversationEntity().supportsVideoCall(true) && userAllowedVideoCall;
```
atomrc: This condition doesn't seem to be semantically correct `this.showCallControls`.
From a reader point of view, I don't really understand how the fact that supporting video call is related to the fact that we want to show the call controls.

gregor: Let's move this check into the `supportsVideoCall()` instead of unify it in a single place.

app/script/view_model/content/TitleBarViewModel.js
```diff
@@ -79,11 +79,7 @@ z.viewModel.content.TitleBarViewModel = class TitleBarViewModel {
       return !this.hasCall() && isSupportedConversation && isActiveConversation;
     });
 
-    this.supportsVideoCall = ko.pureComputed(() => {
-      return this.showCallControls()
-        ? this.userRepository.isTeam() && this.conversationEntity().supportsVideoCall(true)
-        : false;
-    });
+    this.supportsVideoCall = ko.pureComputed(() => this.conversationEntity().supportsVideoCall(true));
```
gregor: Misses safeguard in case `conversationEntity` is not defined

PR #3538, opened 2018-06-05
fix: undelete conversation on incoming call (WEBAPP-5110)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -2852,6 +2852,9 @@ z.conversation.ConversationRepository = class ConversationRepository {
         }
       }
 
+      if (conversationEntity.is_cleared()) {
+        conversationEntity.cleared_timestamp(0);
+      }
```
gregor: I think this should only happen if we did end up mapping a `messageEntity` onto the conversation.

PR #3540, opened 2018-06-05
fix: Pass email and password to CS changelog bot

bennyn:


bennycode:
@ffflorian Fixed! 

Would `-e "${WIRE_WEBAPP_BOT_EMAIL}" -p "${WIRE_WEBAPP_BOT_PASSWORD}"` also have been possible?

ffflorian:
Yes, would be even more explicit but not really needed in this case.

PR #3548, opened 2018-06-05
fix: release media streams when call is picked up from another device

atomrc:
Two small fixes here:

- do not show the `join` button when the group call is picked up from another device
- actually release the mediaStream when the call is picked up somewhere else

### DIFF
--- a/app/script/calling/CallingRepository.js
+++ b/app/script/calling/CallingRepository.js
@@ -379,6 +379,7 @@ z.calling.CallingRepository = class CallingRepository {
         if (isSelfUser && !callEntity.selfClientJoined()) {
           callEntity.selfUserJoined(true);
           callEntity.wasConnected = true;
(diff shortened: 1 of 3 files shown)


app/script/calling/CallingRepository.js
```diff
@@ -482,6 +482,7 @@ z.calling.CallingRepository = class CallingRepository {
         if (response && isSelfUser) {
           const conversationName = callEntity.conversationEntity.display_name();
           this.callLogger.info(`Incoming call in conversation '${conversationName}' accepted on other device`);
+          this.mediaStreamHandler.releaseMediaStream();
           return this.deleteCall(conversationId);
```
gregor: While this fixes the issue the root problem can be found in `needsMediaStream`.

Looking at above code you can see that `deleteCall()` is triggered which will in turn end up in `resetMediaStream()`. This only releases the MediaStream if it is not needed elsewhere.

PR #3555, opened 2018-06-06
fix: no outgoing video calls possible (WEBAPP-5125)

AndyLnd:


gregor:
Thx for adding this necessary change after the camelCasing...

PR #3564, opened 2018-06-06
fix: Don't serialize already serialized properties

bennyn:


app/script/util/DebugUtil.js
```diff
@@ -79,14 +79,18 @@ z.util.DebugUtil = class DebugUtil {
 
   getSerialisedSession(sessionId) {
     return wire.app.repository.storage.storageService.load('sessions', sessionId).then(record => {
-      record.serialised = z.util.arrayToBase64(record.serialised);
+      if (typeof record.serialised !== 'string') {
```
atomrc, line 82: Since the method is called `arrayToBase64`, shouldn't we test that the `record.serialized` is an array?

atomrc: What would happen for an object or a number for example?

bennycode: `z.util.arrayToBase64` accepts `ArrayBuffer|UInt8Array` so I would have to check for both then...

PR #3578, opened 2018-06-07
chore: Move to observables for call state checks

gregor:


app/script/calling/entities/CallEntity.js
```diff
@@ -159,7 +170,7 @@ z.calling.entities.CallEntity = class CallEntity {
       if (!isJoined) {
         this.isConnected(false);
 
-        if (z.calling.enum.CALL_STATE_GROUP.IS_ENDING.includes(this.state())) {
```
gregor, line 162: I changed the values in this group to be more semantically correct. `IS_ENDING` no longer exists as `ONGOING` should not be part of it.

app/script/calling/entities/CallEntity.js
```diff
@@ -184,11 +194,15 @@ z.calling.entities.CallEntity = class CallEntity {
...
       this._clearStateTimeout();
 
-      if (z.calling.enum.CALL_STATE_GROUP.STOP_RINGING.includes(state)) {
-        this._onStateStopRinging();
-      } else if (z.calling.enum.CALL_STATE_GROUP.IS_RINGING.includes(state)) {
-        const isIncomingCall = state === z.calling.enum.CALL_STATE.INCOMING;
-        this._onStateStartRinging(isIncomingCall);
+      const hasState = state !== z.calling.enum.CALL_STATE.UNKNOWN;
+      if (hasState) {
+        const isUnsnsweredState = z.calling.enum.CALL_STATE_GROUP.UNANSWERED.includes(state);
```
atomrc: isUn*a*nsweredState

app/script/components/list/conversationListCell.js
```diff
@@ -35,7 +35,7 @@ z.components.ConversationListCell = class ConversationListCell {
       .computed(() => this.cell_state(z.conversation.ConversationCellState.generate(this.conversation)))
       .extend({rateLimit: 500});
 
-    this.showJoinButton = ko.pureComputed(() => this.conversation.hasActiveDeclinedCall());
+    this.showJoinButton = ko.pureComputed(() => this.conversation.hasJoinableCall());
```
atomrc: you can even remove the `ko.pureComputed(() =>` since it is already an observable and you do not do any operation on it

atomrc:
Big :+1: for this PR.

Observables derived from the raw state improves code readability a lot IMO.

PR #3581, opened 2018-06-07
fix: Show video grid on top of conversation details sidebar (WEBAPP-5133)

atomrc:


app/style/video-calling.less
```diff
@@ -41,7 +41,7 @@
 .video-calling {
   &--active {
     position: absolute;
-    z-index: 101;
+    z-index: 1001;
```
gregor: We should not use these hard codes values here. We have z-index variables defined to avoid exactly this problem. Please use them.

gregor:
Much better but didn't we even define variables for the components based on those variables just a few lines below?

atomrc:
@gregor nice catch, this one is probably more semantically correct :)

PR #3585, opened 2018-06-07
feat: video pause support (WEBAPP-5123)

AndyLnd:


app/script/calling/VideoGridRepository.js
```diff
@@ -35,26 +35,42 @@ z.calling.VideoGridRepository = class VideoGridRepository {
     const selfStreamState = callingRepository.selfStreamState;
     const calls = callingRepository.calls;
     this.grid = ko.observableArray([0, 0, 0, 0]);
-    this.selfId = ko.observable();
     this.thumbnailStream = ko.observable();
-    this.mirrorSelf = ko.pureComputed(() => !selfStreamState.screenSend());
 
     const selfStream = ko.pureComputed(() => {
-      return selfStreamState.videoSend() || selfStreamState.screenSend() ? localMediaStream() : undefined;
+      const stream = selfStreamState.videoSend() || selfStreamState.screenSend() ? localMediaStream() : undefined;
```
atomrc: I probably wrote this in the first place but perhaps parentheses around the condition would make things easier to read. 

I personally don't know which has the priority between the `||` and `?` operators, so I guess not everybody knows, this can lead to unexpected results

app/script/calling/VideoGridRepository.js
```diff
@@ -35,26 +35,42 @@ z.calling.VideoGridRepository = class VideoGridRepository {
     const selfStreamState = callingRepository.selfStreamState;
     const calls = callingRepository.calls;
     this.grid = ko.observableArray([0, 0, 0, 0]);
-    this.selfId = ko.observable();
     this.thumbnailStream = ko.observable();
-    this.mirrorSelf = ko.pureComputed(() => !selfStreamState.screenSend());
 
     const selfStream = ko.pureComputed(() => {
```
atomrc, line 40: Since it's now a meta object (that contains a stream) I would argue that a new name for this variable would be good.
But I have to admit, I don't really have an idea for the name...

app/script/calling/VideoGridRepository.js
```diff
@@ -35,26 +35,42 @@ z.calling.VideoGridRepository = class VideoGridRepository {
...
+        isMuted: !selfStreamState.audioSend(),
+        isSelf: true,
+        stream: stream,
+      };
     });
-    this.selfStreamMuted = ko.pureComputed(() => !selfStreamState.audioSend());
 
-    this.streams = ko.pureComputed(() => {
-      const noVideoParticipantIds = calls()
+    const videoParticipants = ko.pureComputed(() => {
```
atomrc: Why creating a new observable for this? This is only used inside the `streams` observable, let's keep it this way and not publish it in the parent scope

app/script/calling/VideoGridRepository.js
atomrc: How about, `selfStreamDescriptor` and `streamDescriptors`?

app/script/calling/VideoGridRepository.js
AndyLnd: I agree, but prettier decided that no parentheses are needed.

app/script/calling/VideoGridRepository.js
AndyLnd: it's just there for historical reasons. will fix.

app/script/calling/VideoGridRepository.js
gregor: I definitely agree. Just pull it out into its own variable.

PR #3587, opened 2018-06-07
feat: update calling tracking (WEBAPP-4991)

atomrc:
@gregor this is a WIP, but most of the properties are done.

Here is the general state:
`calling.initiated_call` :heavy_check_mark: 
`calling.joined_call`  :heavy_check_mark: 
`calling.established_call` :heavy_check_mark: 
`calling.received_call` :heavy_check_mark: 
`calling.ended_call` :heavy_check_mark: 

~~for the `joined_call`, I added a property on the `callEntity` with the media type used when the call is created (so that it's immutable, and we are garantied we cannot mess with the value).~~
~~The side effect is that, if you wait for the `timeout` before picking up and then join (with audio), the call would have been created with the `video` property to `true` (I guess the call is not destroyed between the timeout and the join action).~~

~~For the `ended_call` we used to track `duration` and `duration_sec` I wonder if we really want to remove the logic for the old `duration` property. (the new spec says it should be a `number` while it's a string right now).~~

~~For the `ended_call`, there is the `AV_switch_toggle` left to do~~

### DIFF
--- a/app/script/calling/CallingRepository.js
+++ b/app/script/calling/CallingRepository.js
@@ -1228,10 +1228,17 @@ z.calling.CallingRepository = class CallingRepository {
    */
   _createCall(callMessageEntity, creatingUserEntity) {
     const {conversationId, sessionId} = callMessageEntity;
+    const mediaType = this._getMediaTypeFromProperties(callMessageEntity.properties);
 
     return this.getCallById(conversationId).catch(() => {
       return this.conversationRepository.get_conversation_by_id(conversationId).then(conversationEntity => {
-        const callEntity = new z.calling.entities.CallEntity(conversationEntity, creatingUserEntity, sessionId, this);
+        const callEntity = new z.calling.entities.CallEntity(
+          conversationEntity,
+          creatingUserEntity,
+          sessionId,
+          this,
+          mediaType
+        );
 
         this.calls.push(callEntity);
         return callEntity;
--- a/app/script/calling/CallingRepository.js
+++ b/app/script/calling/CallingRepository.js
@@ -1278,7 +1278,6 @@ z.calling.CallingRepository = class CallingRepository {
         callEntity.state(callState);
 
         return callEntity.addOrUpdateParticipant(userId, false, callMessageEntity).then(() => {
-          this.telemetry.initiateNewCall(mediaType);
           this.telemetry.track_event(z.tracking.EventName.CALLING.RECEIVED_CALL, callEntity);
           this.injectActivateEvent(callMessageEntity, source);
 
@@ -1331,7 +1330,6 @@ z.calling.CallingRepository = class CallingRepository {
       callEntity.direction = z.calling.enum.CALL_STATE.OUTGOING;
       callEntity.state(z.calling.enum.CALL_STATE.OUTGOING);
 
-      this.telemetry.initiateNewCall(mediaType);
       this.telemetry.track_event(z.tracking.EventName.CALLING.INITIATED_CALL, callEntity);
       return callEntity;
     });
--- a/app/script/calling/CallingRepository.js
+++ b/app/script/calling/CallingRepository.js
@@ -1228,17 +1228,10 @@ z.calling.CallingRepository = class CallingRepository {
    */
   _createCall(callMessageEntity, creatingUserEntity) {
     const {conversationId, sessionId} = callMessageEntity;
-    const mediaType = this._getMediaTypeFromProperties(callMessageEntity.properties);
 
     return this.getCallById(conversationId).catch(() => {
       return this.conversationRepository.get_conversation_by_id(conversationId).then(conversationEntity => {
-        const callEntity = new z.calling.entities.CallEntity(
-          conversationEntity,
-          creatingUserEntity,
-          sessionId,
-          this,
-          mediaType
-        );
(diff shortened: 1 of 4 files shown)


app/script/telemetry/calling/CallTelemetry.js
```diff
@@ -101,109 +109,68 @@ z.telemetry.calling.CallTelemetry = class CallTelemetry {
...
+    if (callEntity) {
+      const {conversationEntity, isGroup, direction} = callEntity;
+
+      const videoTypes = [z.media.MediaType.VIDEO, z.media.MediaType.AUDIO_VIDEO];
 
       attributes = Object.assign(
         {
           conversation_participants: conversationEntity.getNumberOfParticipants(),
-          conversation_participants_in_call: maxNumberOfParticipants ? maxNumberOfParticipants : undefined,
+          conversation_participants_in_call: this.maxNumberOfParticipants ? this.maxNumberOfParticipants : undefined,
```
gregor: This was renamed to contain a `max` in the name

app/script/calling/entities/CallEntity.js
```diff
@@ -135,7 +134,9 @@ z.calling.entities.CallEntity = class CallEntity {
       return false;
     });
 
-    this.participantsCount = ko.pureComputed(() => this.getNumberOfParticipants(this.selfUserJoined()));
+    ko.pureComputed(() => this.getNumberOfParticipants(this.selfUserJoined())).subscribe(usersInCall => {
+      this.telemetry.numberOfParticipantsChanged(usersInCall);
+    });
```
gregor: This is the only place where we use `getNumberOfParticipants`. Let turn it into an observable and use it.

app/script/telemetry/calling/CallTelemetry.js
```diff
@@ -101,109 +109,68 @@ z.telemetry.calling.CallTelemetry = class CallTelemetry {
...
-      }
-
-      let event_name = z.tracking.EventName.CALLING.ENDED_CALL;
-      if (this.media_type === z.media.MediaType.AUDIO_VIDEO) {
-        event_name = event_name.replace('_call', '_video_call');
-      }
-
-      amplify.publish(z.event.WebApp.ANALYTICS.EVENT, event_name, attributes);
+      const eventName = z.tracking.EventName.CALLING.ENDED_CALL;
+      this.track_event(eventName, callEntity, attributes);
```
gregor: I have the feeling, that this fits into a single line.

app/script/telemetry/calling/CallTelemetry.js
atomrc: nice catch

app/script/telemetry/calling/CallTelemetry.js
atomrc: it does indeed

app/script/calling/CallingRepository.js
```diff
@@ -1253,7 +1255,9 @@ z.calling.CallingRepository = class CallingRepository {
 
     return this.userRepository
       .get_user_by_id(userId)
-      .then(remoteUserEntity => this._createCall(callMessageEntity, remoteUserEntity))
+      .then(remoteUserEntity =>
+        this._createCall(callMessageEntity, remoteUserEntity, z.calling.enum.CALL_STATE.INCOMING)
+      )
```
gregor: Let's stick to using explicit returns if it doesn't fit into a single line:

````
.then(remoteUserEntity => {
  return this._createCall(callMessageEntity, remoteUserEntity, z.calling.enum.CALL_STATE.INCOMING);
})
````

File: app/script/calling/entities/CallEntity.js [hunk 5]
```diff
@@ -135,7 +133,11 @@ z.calling.entities.CallEntity = class CallEntity {
       return false;
     });
 
-    this.participantsCount = ko.pureComputed(() => this.getNumberOfParticipants(this.selfUserJoined()));
+    this.participants.subscribe(participants => {
+      const additionalCount = this.selfClientJoined() ? 1 : 0;
+      const numberOfParticipants = participants.length + additionalCount;
+      this.telemetry.numberOfParticipantsChanged(numberOfParticipants);
+    });
```
gregor: Doesn't work incase 4 people have joined the call and then you join as the last person. Max participants should be 5 but it will remain 4 as this does not trigger again.

app/script/calling/CallingRepository.js
atomrc: :+1:

app/script/calling/entities/CallEntity.js
```diff
@@ -135,7 +133,12 @@ z.calling.entities.CallEntity = class CallEntity {
       return false;
     });
 
-    this.participantsCount = ko.pureComputed(() => this.getNumberOfParticipants(this.selfUserJoined()));
+    ko.pureComputed(() => {
+      const additionalCount = this.selfClientJoined() ? 1 : 0;
+      return this.participants().length + additionalCount;
+    }).subscribe(numberOfParticipants => {
+      this.telemetry.numberOfParticipantsChanged(numberOfParticipants);
+    });
```
gregor, line 141: if we have the computed any way lets keep track of it with `this.participantsCount`

PR #3595, opened 2018-06-08
 chore: add data-uie to screenshare

AndyLnd:


app/page/template/video-calling.htm
```diff
@@ -58,8 +58,8 @@
             <!-- /ko -->
 
             <div class="video-controls__button"
-                 data-bind="click: clickedOnShareScreen, css: {'video-controls__button--active': selfStreamState.screenSend(), 'video-controls__button--disabled': disableToggleScreen()}"
-                 data-uie-name="do-call-controls-toggle-screen">
+                 data-bind="click: clickedOnShareScreen, css: {'video-controls__button--active': selfStreamState.screenSend(), 'video-controls__button--disabled': disableToggleScreen()}, attr: {'data-uie-value': selfStreamState.screenSend() ? 'active' : 'inactive', 'data-uie-enabled': disableToggleScreen() ? 'false' : true}"
```
gregor: Why is 'false' a string but true is not?

AndyLnd: oops

PR #3623, opened 2018-06-11
feat: Add Cryptobox export in debug mode

bennyn:
I have implemented two convenience functions:
1. `wire.app.util.debug.exportCryptobox` (will only work when running the app with debug utils)
2. `z.util.downloadText` (to download texts as files in the browser)


app/script/util/util.js
```diff
@@ -232,28 +232,40 @@ z.util.base64ToBlob = base64 => {
 /**
  * Downloads blob using a hidden link element.
  * @param {Blob} blob - Blob to store
- * @param {string} filename - Data will be saved under this name
```
gregor: Filename is a single word these days. No need to camelCase. https://en.wikipedia.org/wiki/Filename

app/script/util/util.js
```diff
@@ -232,28 +232,40 @@ z.util.base64ToBlob = base64 => {
...
-  link.download = filename;
-  link.style = 'display: none';
+  const anchor = document.createElement('a');
+  anchor.href = url;
   if (mimeType) {
-    link.type = mimeType;
+    anchor.type = mimeType;
   }
-  link.click();
+  return z.util.downloadFile(anchor, url, fileName);
```
gregor: I thin the download file method should take the href content, name and an optional mimetype and create the link element itself. Looks like this would simplify the code.

bennycode: Ok, I changed that.

PR #3628, opened 2018-06-12
fix: max time left for temporary users is 24h (WEBAPP-5151)

AndyLnd:


app/script/entity/User.js
```diff
@@ -276,7 +276,7 @@ z.entity.User = class User {
     const MILLISECONDS_IN_MINUTE = 60 * 1000;
     const MILLISECONDS_IN_HOUR = MILLISECONDS_IN_MINUTE * 60;
 
-    const remainingTime = Math.max(expirationTime - Date.now(), 0);
+    const remainingTime = Math.min(Math.max(expirationTime - Date.now(), 0), 24);
```
gregor: Let's get rid of the magic numbers for @lipis

PR #3629, opened 2018-06-12
fix: Show name of caller in incoming group call

AndyLnd:


app/script/components/list/conversationListCallingCell.js
```diff
@@ -157,7 +157,12 @@ ko.components.register('conversation-list-calling-cell', {
       <div class="conversation-list-cell-center" data-bind="css: {'conversation-list-cell-center-no-left': temporaryUserStyle}">
         <span class="conversation-list-cell-name" data-bind="text: conversation.display_name()"></span>
         <!-- ko if: isIncoming() -->
-          <span class="conversation-list-cell-description" data-bind="l10n_text: z.string.callStateIncoming" data-uie-name="call-label-incoming"></span>
+          <!-- ko if: call().isGroup -->
+            <span class="conversation-list-cell-description" data-bind="l10n_text: {id: z.string.conversationsSecondaryLineIncomingCall, substitute: call().creatingUser.first_name()}" data-uie-name="call-label-incoming"></span>
```
gregor: Even though it is the same string, we should not reuse the ID in case they diverge. The string IDs are essentially UI locator identifiers. Please create a `z,.string.callStateIncomingGroup` instead.

PR #3637, opened 2018-06-12
fix: Prevent audio leak on stream upgrade (WEBAPP-5117, WEBAPP-5140, WEBAPP-5141)

gregor:


app/script/calling/entities/FlowEntity.js
```diff
@@ -1222,25 +1226,77 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
 
-      const isExpectedError = expectedErrorTypes.includes(type);
-      if (isExpectedError) {
-        this.callLogger.debug(`Replacement of MediaStream and renegotiation necessary: ${message}`, error);
-        return this._replaceMediaStream(mediaStreamInfo);
-      }
-      throw error;
-    });
+  /**
+   * Replace the a MediaStreamTrack attached to the MediaStream of the PeerConnection.
```
Yserz: -a

app/script/calling/entities/FlowEntity.js
```diff
@@ -31,6 +31,10 @@ z.calling.entities.FlowEntity = class FlowEntity {
       NEGOTIATION_RESTART_TIMEOUT: 2500,
       SDP_SEND_TIMEOUT: 5 * 1000,
       SDP_SEND_TIMEOUT_RESET: 1000,
+      UPGRADE_TRACK_FAILED: [
```
atomrc: Since those constants are used in a single place in this file, I would rather move their definition in a narrower scope (next to the place they are consumed).

app/script/calling/entities/FlowEntity.js
```diff
@@ -1222,25 +1226,77 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+  /**
+   * Replace the MediaStreamTrack attached to the MediaStream of the PeerConnection.
+   *
+   * @private
+   * @param {z.media.MediaStreamInfo} mediaStreamInfo - Object containing the required MediaStream information
+   * @returns {Promise} Resolves when a MediaStreamTrack has been replaced
+   */
+  replaceMediaTrack(mediaStreamInfo) {
+    const {stream: mediaStream, type: mediaType} = mediaStreamInfo;
+    const [mediaStreamTrack] = mediaStream.getTracks();
```
atomrc, line 1254: Is it always safe to assume that we just need the first track? Are we absolutely sure we never need to handle multiple tracks?

gregor: Yes, in this case it is.

PR #3644, opened 2018-06-13
fix: fullscreen call ui flash for incoming calls (WEBAPP-5154)

AndyLnd:


app/script/view_model/VideoCallingViewModel.js
```diff
@@ -84,7 +84,9 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
     });
 
     this.isCallOngoing = ko.pureComputed(() => {
-      return this.joinedCall() ? this.videodCall() && this.joinedCall().isOngoing() : false;
+      const joinedCall = this.joinedCall();
+      const isSendingVideo = this.selfStreamState.screenSend() || this.selfStreamState.videoSend();
+      return joinedCall ? joinedCall.isOngoing() && (joinedCall.isRemoteVideoCall() || isSendingVideo) : false;
```
gregor: I think we can make this more readable.

gregor:
````
if (this.joinedCall()) {
  const isSendingVideo = this.selfStreamState.screenSend() || this.selfStreamState.videoSend();
  const isVideodCall = isSendingVideo || this.joinedCall().isRemoteVideoCall();
  return this.joinedCall().isOngoing() && isVideodCall;
}
````

PR #3663, opened 2018-06-13
chore: Simplify video state

gregor:


test/unit_tests/calling/videoGridRepositorySpec.js
```diff
@@ -138,16 +138,7 @@ describe('z.calling.VideoGridRepository', () => {
...
         generateCallingRepository({
-          calls: ko.observableArray([
-            {
-              participants: () => [generateParticipant('user-1', true)],
-            },
-          ]),
-          selfStreamState: {
-            audioSend: ko.observable(true),
-            screenSend: ko.observable(true),
-            videoSend: ko.observable(true),
```
gregor, line 149: This state should not exist. You cannot have video and screen enabled at the same time. They are mutually exclusive.

PR #3678, opened 2018-06-14
chore: Refactor MediaStream initialization

gregor:


app/script/media/MediaStreamHandler.js
```diff
@@ -273,20 +273,38 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
-    if (noAudioDevice) {
-      const mediaError = new z.media.MediaError(z.media.MediaError.TYPE.MEDIA_STREAM_DEVICE, z.media.MediaType.AUDIO);
-      return Promise.reject(mediaError);
-    }
+    return this._checkDeviceAvailability(mediaType).then(() => {
+      return this._requestMediaStream(mediaType, mediaStreamConstraints);
+    });
+  }
 
+  _checkDeviceAvailability(mediaType) {
```
atomrc, line 281: Can we try something new for private-single-used methods? I would suggest to create simple arrow function (so that they have access to `this`) bound to a local `const` instead of adding methods to the class.

which would give:

```javascript
requestMediaStream(mediaType, mediaStreamConstraints) {
  const checkDeviceAvailability = () => {/*...*/}
  const requestMediaStream = () => {/*...*/}

  return checkDeviceAvailability(mediaType).then(() => {
    return requestMediaStream(mediaType, mediaStreamConstraints);
  })
```

The arguments for this:
- private by design (not by convention) ;
- strong guarantees, for someone who discovers the code, that this function can be touched quite safely because not used anywhere else.

gregor: I totally understand your issue with the current paradigm. The strong guarantees is not something we have ever given. We preferred the readability of the code, single line Promise chains and shorts methods that are not littered with local functions. We need to see how we move forward but I will not set precedent and start change in this PR.

PR #3680, opened 2018-06-14
fix: track-wise local media stream tracks replacement

atomrc:
Fixes an issue (on Firefox) where the stream tracks are not released when switching video devices during the call and hanging up

### DIFF
--- a/app/script/media/MediaStreamHandler.js
+++ b/app/script/media/MediaStreamHandler.js
@@ -206,15 +206,15 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
 
     const replacePromise = this.joinedCall()
       ? this._updateJoinedCall(mediaStreamInfo)
-      : Promise.resolve(mediaStreamInfo);
+      : Promise.resolve({replace: false, streamInfo: mediaStreamInfo});
 
-    replacePromise.then(updatedMediaStreamInfo => this._handleReplacedMediaStream(updatedMediaStreamInfo));
+    replacePromise.then(this._handleReplacedMediaStream.bind(this));
   }
 
-  _handleReplacedMediaStream(mediaStreamInfo) {
+  _handleReplacedMediaStream({streamInfo: mediaStreamInfo, replace}) {
     const replaceMediaStreamLocally = newMediaStreamInfo => {
-      const newMediaStream = mediaStreamInfo.stream;
-      const newMediaStreamType = mediaStreamInfo.getType();
+      const newMediaStream = newMediaStreamInfo.stream;
+      const newMediaStreamType = newMediaStreamInfo.getType();
 
       this._releaseMediaStream(this.localMediaStream());
       this._setStreamState(newMediaStream, newMediaStreamType);
@@ -229,9 +229,7 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
       this._addTracksToStream(mediaStream, this.localMediaStream(), mediaType);
     };
 
-    return mediaStreamInfo.replaced
-      ? replaceMediaStreamLocally(mediaStreamInfo)
-      : replaceMediaTracksLocally(mediaStreamInfo);
+    return replace ? replaceMediaStreamLocally(mediaStreamInfo) : replaceMediaTracksLocally(mediaStreamInfo);
   }
 
   /**
@@ -594,8 +592,6 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
 
     const replaceMediaStreamInFlows = (streamInfo, flows) => {
       return this._updateMediaStream(streamInfo).then(newMediaStreamInfo => {
-        newMediaStreamInfo.replaced = true;
-
         const upgradePromises = flows.map(flowEntity => {
           return flowEntity.replaceMediaStream(newMediaStreamInfo, this.localMediaStream());
         });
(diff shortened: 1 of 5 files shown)


app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    * @returns {Promise} Resolves when the MediaStream has been replaced
    */
   replaceMediaStream(mediaStreamInfo) {
-    const {stream: mediaStream, type} = mediaStreamInfo;
+    const {stream: mediaStream} = mediaStreamInfo;
+
+    const replaceMediaStream = (localMediaStream, newMediaStreamInfo) => {
+      this._releaseMediaStream(this.localMediaStream());
```
AndyLnd: Use the parameter names please :)

app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
+      this._setStreamState(updateMediaStreamInfo);
```
AndyLnd: newMediaStreamInfo

app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
+      this.localMediaStream(updateMediaStreamInfo.stream);
```
AndyLnd: see above

app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
+    };
+
+    const replaceMediaTracks = (localMediaStream, newMediaStream) => {
+      localMediaStream.getTracks().forEach(localTrack => {
+        updateMediaStreamInfo.stream.getTracks().forEach(newTrack => {
```
AndyLnd: newMediaStream

app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
+          if (localTrack.kind === newTrack.kind) {
+            localTrack.stop();
+            localMediaStream.removeTrack(localTrack);
+            localMediaStream.addTrack(newTrack);
+          }
+        });
+      });
+    };
```
gregor: We already have this code. Let's not duplicate it.

atomrc: Where is it?

app/script/media/MediaStreamHandler.js
```diff
@@ -194,7 +194,25 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    * @returns {Promise} Resolves when the MediaStream has been replaced
    */
   replaceMediaStream(mediaStreamInfo) {
-    const {stream: mediaStream, type} = mediaStreamInfo;
+    const {stream: mediaStream} = mediaStreamInfo;
+
+    const replaceMediaStream = (localMediaStream, newMediaStreamInfo) => {
```
gregor: The naming is confusing. This is in the replaceMediaStream. We do a replace when we update the joined call. This is more like an update local media stream functionality.

app/script/media/MediaStreamHandler.js
```diff
@@ -209,11 +227,9 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
 
     return replacePromise.then(updateMediaStreamInfo => {
-      const mediaType = !updateMediaStreamInfo.replaced ? type : undefined;
-
-      this._setStreamState(updateMediaStreamInfo);
-      this._releaseMediaStream(this.localMediaStream(), mediaType);
-      this.localMediaStream(updateMediaStreamInfo.stream);
+      return updateMediaStreamInfo.replaced
+        ? replaceMediaStream(this.localMediaStream(), updateMediaStreamInfo)
+        : replaceMediaTracks(this.localMediaStream(), updateMediaStreamInfo.stream);
```
gregor: We can also move this to the methods within the `updateJoinedCall`. In `_replaceMediaStream` and `_replaceMediaTrack` we already know how we have to update the local media stream. Doing it out here is legacy and we can improve this.

app/script/media/MediaStreamHandler.js
atomrc: Quick suggestion: How about we rename the method `updateMediaStream` and depending on the case it can either `replaceMediaStream` or `replaceMediaTrack` so that we have the same naming for both subfunctions?
What do you think?

app/script/media/MediaStreamHandler.js
gregor: https://github.com/wireapp/wire-webapp/blob/dev/app/script/media/MediaStreamHandler.js#L601-L614

Not fully  the same but it does the remove and add of certain tracks.

atomrc: Yeah ok I see. I have to admit this `clone` part kind of scares me. I definitly don't want to clone the stream in that case and I wonder what could be the side effects of this...

app/script/media/MediaStreamHandler.js
atomrc: :+1:

atomrc: Oh wait, we also need to do that in the case where the call is not yet joined, isn't it? So it seems legit that it is in the `replaceMediaStream` method, isn't it?

PR #3696, opened 2018-06-18
feat: add video call disclaimer for temporary guests

AndyLnd:


app/script/localization/webapp.js
```diff
@@ -582,6 +582,7 @@ z.string.preferencesAVCamera = 'Camera';
 z.string.preferencesAVMicrophone = 'Microphone';
 z.string.preferencesAVPermissionDetail = 'Enable from your browser Preferences';
 z.string.preferencesAVSpeakers = 'Speakers';
+z.string.preferencesAVTemporaryDisclaimer = 'Guests can\'t start video conferences. Select the camera to use if you join one.';
```
gregor: We use the other `'` for this use case in order to avoid escaping all the instances.

PR #3708, opened 2018-06-18
feat: ephemeral messages timer's max and min bounds

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -2064,7 +2068,11 @@ z.conversation.ConversationRepository = class ConversationRepository {
    */
   _wrap_in_ephemeral_message(generic_message, millis) {
     const ephemeral = new z.proto.Ephemeral();
-    ephemeral.set('expire_after_millis', millis);
+    const fixedTimer = Math.min(
```
atomrc: @gregor ok, quick discussion about this. Do you think the conversation repo is the right place for this logic?

My argument for adding the logic here, is to have the dumbest entity classes (the anemic domain model pattern which is not loved by everyone, including Martin Fowler :D https://martinfowler.com/bliki/AnemicDomainModel.html but helps separating raw data from logic in my opinion.)

If you have any opinion on this, I am all ears

app/script/conversation/ConversationRepository.js
```diff
@@ -2064,7 +2068,12 @@ z.conversation.ConversationRepository = class ConversationRepository {
    */
   _wrap_in_ephemeral_message(generic_message, millis) {
     const ephemeral = new z.proto.Ephemeral();
-    ephemeral.set('expire_after_millis', millis);
+    const fixedTimer = z.util.NumberUtil.clamp(
```
atomrc, line 2067: @gregor ok, quick discussion about this. Do you think the conversation repo is the right place for this logic?

My argument for adding the logic here, is to have the dumbest entity classes (the anemic domain model pattern which is not loved by everyone, including Martin Fowler :D https://martinfowler.com/bliki/AnemicDomainModel.html but helps separating raw data from logic in my opinion.)

If you have any opinion on this, I am all ears

gregor: Definitely the place to put in IMHO as well. Otherwise people misunsing the code could circumvent it. We also need the reverse logic on the cryptographich mapper side that unpacks the proto message. It is where you just removed the match to closest value in the previous PR. Otherwise malicious clients can send anything in the proto. This is part of the protocol now. We have to enforce it on the client level - both receiving and sending end.

test/unit_tests/conversation/ConversationRepositorySpec.js
```diff
@@ -1031,6 +1031,43 @@ describe('ConversationRepository', () => {
     });
   });
 
+  describe('send_text_with_link_preview', () => {
```
gregor: ?

PR #3719, opened 2018-06-19
chore: Use z.util.NumberUtil() where applicable

gregor:


app/script/components/asset/controls/seekBar.js
```diff
@@ -104,11 +105,11 @@ z.components.SeekBarComponent = class SeekBarComponent {
 
   _update_seek_bar_style(progress) {
     // TODO check if we can find a css solution
-    if (this.dark_mode) {
-      return (this.seek_bar.style.backgroundImage = `linear-gradient(to right, currentColor ${progress}%, rgba(141,152,159,0.24) ${progress}%)`);
-    }
+    this.seek_bar.style.backgroundImage = this.dark_mode
+      ? `linear-gradient(to right, currentColor ${progress}%, rgba(141,152,159,0.24) ${progress}%)`
+      : `linear-gradient(to right, currentColor ${progress}%, rgba(255,255,255,0.4) ${progress}%)`;
```
atomrc: While you are at it, I think we can refactor this a little more and extract the `color` in a variable depending on the color mode.

```
const color = this.dark_mode ? "rgba(141,152,159,0.24)" : "rgba(255,255,255,0.4)";
```

And then use it in the gradient string

gregor: Good point

gregor: I also removed the return as this seems to be stemming from the tool assisted decaffeination.

PR #3723, opened 2018-06-19
feat: panel style, timed message disclaimer, fixed animation

AndyLnd:


app/page/template/panel/timed-messages.htm
```diff
@@ -6,20 +6,23 @@
...
     </div>
     <div class="panel__content">
-        <div class="panel__action-item" data-bind="click: clickOnMessageTimeOff" data-uie-name="item-timed-messages-option">
-            <div class="panel__action-item__text" data-bind="l10n_text: z.string.timedMessagesOff"></div>
-            <!-- ko if: currentMessageTimer() === 0 -->
+        <div class="panel__content__scroll" data-bind="antiscroll: shouldUpdateScrollbar">
+            <div class="panel__action-item" data-bind="click: clickOnMessageTimeOff" data-uie-name="item-timed-messages-option">
+                <div class="panel__action-item__text" data-bind="l10n_text: z.string.timedMessagesOff"></div>
+                <!-- ko if: currentMessageTimer() === 0 -->
                 <check-icon></check-icon>
```
atomrc: Shouldn't we indent this also?

app/page/template/panel/timed-messages.htm
```diff
@@ -6,20 +6,23 @@
...
-            <!-- /ko -->
-        </div>
-        <!-- ko foreach: messageTimes -->
+                <!-- /ko -->
+            </div>
+            <!-- ko foreach: {data: messageTimes, afterRender: afterRender} -->
```
atomrc: I would rather name this function by the job it does, not the moment when it's called (something like `setRendered`, since the function seems to set a boolean)

PR #3725, opened 2018-06-19
 fix: hide option from guests

AndyLnd:


app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -129,7 +129,7 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
     this.showActionClear = ko.pureComputed(() => {
       return !this.conversationEntity().is_request() && !this.conversationEntity().is_cleared();
     });
-    this.showActionGuestOptions = ko.pureComputed(() => this.conversationEntity().inTeam());
+    this.showTeamMemberActions = ko.pureComputed(() => this.conversationEntity().inTeam());
```
gregor: The timed message feature is also available for non team users. We need to handle these separately.

PR #3728, opened 2018-06-19
feat: message timer slow sync

atomrc:


test/unit_tests/conversation/ConversationMapperSpec.js
```diff
@@ -321,7 +346,7 @@ describe('Conversation Mapper', () => {
       expect(merged_conversation.last_event_timestamp).toBe(local_data.last_event_timestamp);
       expect(merged_conversation.last_read_timestamp).toBe(local_data.last_read_timestamp);
 
-      expect(merged_conversation.muted_timestamp).toBe(local_data.muted_timestamp);
+      expect(merged_conversation.muted_timestamp).not.toBe(local_data.muted_timestamp);
```
atomrc, line 349: Not a behavior change, that was a bug in the test implementation. 
Previously, we were mutating the local conversation in the `merge_conversation` method.
Which means we were actually comparing value of the merged_conversation with itself.

This is fixed thanks to the `Object.assign` in the `merge_conversation` method

PR #3744, opened 2018-06-20
 fix: support longer times in tooltip

AndyLnd:


app/page/template/partials/template-message.htm
```diff
@@ -86,9 +86,9 @@
     </div>
   <!-- /ko -->
       
-  <div class="message-body">
+  <div class="message-body" data-bind="attr: {'title': message.ephemeral_caption()}">
```
gregor: Did you discuss this change with design?

app/script/entity/message/Message.js
```diff
@@ -266,15 +266,18 @@ z.entity.Message = class Message {
...
-    this.ephemeral_remaining(this.ephemeral_expires() - Date.now());
+    const remainingTime = this.ephemeral_expires() - Date.now();
+    this.ephemeral_remaining(remainingTime);
 
     this.ephemeral_interval_id = window.setInterval(() => {
-      this.ephemeral_remaining(this.ephemeral_expires() - Date.now());
-      this.ephemeral_caption(z.util.formatTimeRemaining(this.ephemeral_remaining()));
+      const updatedRemainingTime = this.ephemeral_expires() - Date.now();
+      this.ephemeral_remaining(updatedRemainingTime);
+      this.ephemeral_caption(z.util.TimeUtil.formatDuration(this.ephemeral_remaining(), 3).text);
```
gregor: Let's pull this out into a variable for improved readability please.

````
const formatedDuration = z.util.TimeUtil.formatDuration(this.ephemeral_remaining(), 3);
this.ephemeral_caption(formatedDuration.text);
````

File: app/script/util/TimeUtil.js [hunk 3]
```diff
@@ -31,9 +31,10 @@ z.util.TimeUtil = {
   /**
    * Format milliseconds into 15s, 2m.
    * @param {number} duration - Duration to format in milliseconds
+   * @param {number} maximumUnits - Maximum number of units shown in the textual representation
    * @returns {Object} Unit, value and localized string
    */
-  formatDuration: duration => {
+  formatDuration: (duration, maximumUnits = 1) => {
```
gregor, line 37: Where else did we use this? Can we safely set this default value? I suppose it is ok, because it only affects the text which is set in one place only.

app/page/template/partials/template-message.htm
AndyLnd: This is the intended new behaviour.

app/script/util/TimeUtil.js
AndyLnd: I wrote this method to replace some scattered methods which did all pretty much the same. I know where and how this is used and can assure that the default value is save.

PR #3745, opened 2018-06-20
feat: timed messages system messages

atomrc:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -414,33 +414,12 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
       .catch(() => reset_progress());
   }
 
-  /**
-   * Gets CSS class that will be applied to the message div in order to style.
-   * @param {z.entity.Message} message - Message entity for generating css class
-   * @returns {string} CSS class that is applied to the element
-   */
-  get_css_class(message) {
```
atomrc, line 422: In case you are wondering, this is safe to remove because this css classe was overridden anyway by the next `css` binding here https://github.com/wireapp/wire-webapp/pull/3745/files#diff-b0ae56dc64faa60de5486cf73f73ff0cL8

app/script/entity/message/MessageTimerUpdateMessage.js
```diff
@@ -0,0 +1,42 @@
...
+    this.system_message_type = z.message.SystemMessageType.CONVERSATION_MESSAGE_TIMER_UPDATE;
+    this.message_timer = messageTimer;
+
+    this.caption = ko.pureComputed(() => {
+      if (this.message_timer === null) {
+        return z.l10n.text(z.string.conversationResetTimer);
+      }
+      const timeString = z.util.TimeUtil.formatDuration(this.message_timer).text;
+      const identifier = z.string.conversationUpdatedTimer;
+      return z.l10n.text(identifier, {time: timeString});
```
gregor: I think we can just put the identifier in here. Nothing gained from an extra variable.

app/script/entity/message/MessageTimerUpdateMessage.js
```diff
@@ -0,0 +1,42 @@
...
+z.entity.MessageTimerUpdateMessage = class MessageTimerUpdateMessage extends z.entity.SystemMessage {
+  constructor(messageTimer) {
+    super();
+
+    this.type = z.event.Backend.CONVERSATION.MESSAGE_TIMER_UPDATE;
+    this.system_message_type = z.message.SystemMessageType.CONVERSATION_MESSAGE_TIMER_UPDATE;
+    this.message_timer = messageTimer;
+
+    this.caption = ko.pureComputed(() => {
+      if (this.message_timer === null) {
```
gregor: Can we safely go with a null here or should we just check for falsy instead incase we have undefined or 0?

PR #3747, opened 2018-06-20
fix: Permission state handling for notifications and media

gregor:


app/script/calling/entities/CallEntity.js
```diff
@@ -63,13 +63,7 @@ z.calling.entities.CallEntity = class CallEntity {
...
     this.callLogger = new z.telemetry.calling.CallLogger(loggerName, this.id, z.config.LOGGER.OPTIONS, this.messageLog);
 
-    this.callLogger.info({
-      data: {
-        default: [this.id],
-        obfuscated: [this.callLogger.obfuscate(this.id)],
-      },
-      message: `Created new call entity in conversation {0}`,
-    });
+    this.callLogger.info(`Created new call entity in conversation ${this.id}`);
```
atomrc, line 66: Seems we have lost obfuscation here

app/script/notification/NotificationRepository.js
```diff
@@ -605,6 +603,15 @@ z.notification.NotificationRepository = class NotificationRepository {
     }
   }
 
+  /**
+   * Get the current browser permission state.
+   * @private
+   * @returns {Promise} Resolves with true if notifications are permitted
+   */
+  _updateBrowserPermissionState() {
```
atomrc: Do we really need to extract this in another function?

app/script/media/MediaStreamHandler.js
```diff
@@ -345,25 +345,31 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    *
    * @private
    * @param {z.media.MediaType} mediaType - Requested media type
-   * @returns {Promise} Resolves when the permissions are not denied
+   * @returns {Promise} Resolves true when permissions need to be prompted for
```
atomrc: The API of this method is confusing.
I would suggest to rename it to something like `hasPermissionToAccess(mediaType)` and reverse the returned boolean values

app/script/calling/entities/CallEntity.js
gregor: According to @pwnsdx obfuscation of IP addresses and UUID is handled by default. We only need to pass in the object in cases where we replace names with IDs. Not exactly obvious. We need to see how to improve the documentation and handling of the call logger.

PR #3753, opened 2018-06-21
feat: use a central handler for all timed messages

AndyLnd:


app/page/template/_dist/app.htm
```diff
@@ -162,6 +162,7 @@
 <script src="/script/conversation/ConversationStateHandler.js"></script>
 <script src="/script/conversation/ConversationStatus.js"></script>
 <script src="/script/conversation/ConversationStatusIcon.js"></script>
+<script src="/script/conversation/ConversationTimedHandler.js"></script>
```
gregor: Should we stick to the term ephemeral that  we have been using throughout the app.

app/script/conversation/ConversationRepository.js
```diff
@@ -3534,6 +3401,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
           conversationEntity.prepend_messages(messageEntities);
         } else {
           conversationEntity.add_messages(messageEntities);
+          messageEntities.forEach(this.timedMessageHandler.addTimedMessage);
```
gregor: You also need to do this in the prepend case. Take it out of the if block. Otherwise message you reveal when scrolling up are not restarted.

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,221 @@
...
+    return {
+      INTERVAL_TIME: 250,
+    };
+  }
+
+  constructor(conversationService, conversationRepository) {
+    this.updateTimedMessages = this.updateTimedMessages.bind(this);
+    this.addTimedMessage = this.addTimedMessage.bind(this);
+    this.removeTimedMessage = this.removeTimedMessage.bind(this);
+    this.checkMessageTimer = this.checkMessageTimer.bind(this);
```
gregor: A, C, R, U

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,221 @@
...
+    this.conversationRepository = conversationRepository;
+    this.logger = new z.util.Logger('z.conversation.ConversationTimedHandler', z.config.LOGGER.OPTIONS);
+
+    this.timedMessages = [];
+    window.setInterval(this.updateTimedMessages, ConversationTimedHandler.CONFIG.INTERVAL_TIME);
+  }
+
+  updateTimedMessages() {
+    const now = Date.now();
+    for (let index = this.timedMessages.length - 1; index >= 0; index--) {
```
gregor: It's es6. We can just use a forEach().

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+
+  addTimedMessage(message) {
+    if (message.ephemeral_status === z.message.EphemeralStatusType.NONE) {
+      return;
+    }
+    const isAlreadyAdded = this.timedMessages.some(
+      timedMessage => timedMessage.id === message.id && timedMessage.conversation_id === message.conversation_id
+    );
```
gregor: Please change to explicit return

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,221 @@
...
+    if (!isAlreadyAdded) {
+      this.timedMessages.push(message);
+      this.updateTimedMessages();
+    }
+  }
+
+  removeTimedMessage(message) {
+    const messageIndex = this.timedMessages.indexOf(message);
+    if (messageIndex > -1) {
```
gregor: Let's Filter based on conversation and message IDs.

app/script/entity/message/Message.js
```diff
@@ -261,7 +264,7 @@ z.entity.Message = class Message {
 
   // Start the ephemeral timer for the message.
   startMessageTimer() {
-    if (this.ephemeral_timeout_id) {
+    if (this.messageTimerStarted) {
```
gregor, line 267: Looks like we can remove the timeout ID in the constructor and in it the started check.

app/script/conversation/ConversationRepository.js
```diff
@@ -156,6 +156,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
     this._init_subscriptions();
 
     this.stateHandler = new z.conversation.ConversationStateHandler(this.conversation_service, this);
+    this.timedMessageHandler = new z.conversation.ConversationTimedHandler(this.conversation_service, this);
```
atomrc: This circular dependency (the fact that the `ConversationTimedHandler` depends on the `ConversationRepository` and vice versa) trigger a huge warning. 

It means the concerns are not clearly separated and that those two services should be siblings, not parent>child.

Since I am working on solving this issue right now, lets sit together and talk about this, WDYT?

app/script/conversation/ConversationTimedHandler.js
gregor: Ignore that one. Didn't see you where going backwards because of the splice....

app/page/template/_dist/app.htm
AndyLnd: I am trying to slowly move the terminology to 'timed messages' as this is the new name for this feature.

app/script/conversation/ConversationRepository.js
gregor: Once we have fully broken up the conversation repository into handlers we will end up with exactly that. We have many siblings (the handlers) that are combined using the parent (repo).

Or do you see any additional issues with this approach - which can be seen fully implemented in the `MediaRepository` and its handlers.

atomrc: Yes, absolutely, the handlers are all siblings. My problem is with the fact that we inject the `repository` in the `handler` (meaning that there is a bidirectional communication between the two).

We came up with a nice solution to that yesterday with @AndyLnd that completely get rid of the circular dependency.

app/script/conversation/ConversationTimedHandler.js
AndyLnd: Let's not in this case. I want a specific message object to be removed. There _should_ never be two with colliding conversation and message IDs, but if there by any freaky accident are, i still just want this specific message object out. I underscored the method though, to make it clear, that it should never be called from outside.

app/page/template/_dist/app.htm
gregor: It is not the new name. It is the marketing name. This naming could interfere with the other potential things from the roadmap.

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+    this.logger = new z.util.Logger('z.conversation.ConversationTimedHandler', z.config.LOGGER.OPTIONS);
+
+    this.timedMessages = [];
+    window.setInterval(this._updateTimedMessages.bind(this), ConversationTimedHandler.CONFIG.INTERVAL_TIME);
+  }
+
+  _updateTimedMessages() {
+    const now = Date.now();
+    for (let index = this.timedMessages.length - 1; index >= 0; index--) {
+      const message = this.timedMessages[index];
```
gregor: This should be `messageEntity`

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+        if (remainingTime > 0) {
+          message.ephemeral_remaining(remainingTime);
+        } else {
+          this._removeTimedMessage(message);
+        }
+      }
+    }
+  }
+
+  addTimedMessage(message) {
```
gregor: This should be `messageEntity`

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+    const isAlreadyAdded = this.timedMessages.some(timedMessage => {
+      return timedMessage.id === message.id && timedMessage.conversation_id === message.conversation_id;
+    });
+    if (!isAlreadyAdded) {
+      this.timedMessages.push(message);
+      this._updateTimedMessages();
+    }
+  }
+
+  _removeTimedMessage(message) {
```
gregor: This should be `messageEntity`

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+      this._obfuscateAssetMessage(messageEntity);
+    } else if (messageEntity.has_asset_image()) {
+      this.__obfuscateImageMessage(messageEntity);
+    } else {
+      this.logger.warn(`Ephemeral message of unsupported type: ${messageEntity.type}`);
+    }
+  }
+
+  _obfuscateAssetMessage(messageEntity) {
+    const asset = messageEntity.get_first_asset();
```
gregor: This should be `assetEntity`

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+        content_type: asset.file_type,
+        meta: {},
+      },
+      ephemeral_expires: true,
+    });
+    this.logger.info(`Obfuscated asset message '${messageEntity.id}'`);
+  }
+
+  _obfuscateImageMessage(messageEntity) {
+    const asset = messageEntity.get_first_asset();
```
gregor: This should be `assetEntity`

app/script/conversation/ConversationTimedHandler.js
```diff
@@ -0,0 +1,184 @@
...
+      },
+      ephemeral_expires: true,
+    });
+    this.logger.info(`Obfuscated image message '${messageEntity.id}'`);
+  }
+
+  _obfuscateTextMessage(messageEntity) {
+    const asset = messageEntity.get_first_asset();
+    const obfuscated_asset = new z.entity.Text(messageEntity.id);
+    const obfuscated_previews = asset.previews().map(link_preview => {
```
gregor: This should be `assetEntity` and let's properly camelCase in new files.

app/page/template/_dist/app.htm
AndyLnd: I think you are right. It's easier and less error prone to stick with 'ephemeral' on the code side.

app/script/conversation/ConversationEphemeralHandler.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  constructor(conversationService, onMessageTimeout) {
+    this.conversationService = conversationService;
+    this.onMessageTimeout = onMessageTimeout;
+    this.logger = new z.util.Logger('z.conversation.ConversationEphemeralHandler', z.config.LOGGER.OPTIONS);
+
+    this.timedMessages = [];
+    window.setInterval(this._updateTimedMessages.bind(this), ConversationEphemeralHandler.CONFIG.INTERVAL_TIME);
+  }
```
gregor, line 39: This looks good for now to be merged to edge. But let's improve this part and only `setInterval` if `!!timeMessages.length` and clear it when `!timedMessages`. We cann always do that if we turn timed messages into an `ko.obserableArray` and subscribe to it or manually handle it on add/remove/update.

gregor:
Another improvement we should be able to accomplish and look into now is addressing the concerns raised in https://github.com/wireapp/wire-desktop/issues/1120. Expired messages should never be added to the view in the first place. Thuse we should reword the `addTimedMessage` to a method that returns non ephemeral and non expired messages - so that the `add_event_to_conversation` methods that rely on it can only add those to the conversation.

PR #3757, opened 2018-06-21
Clarify GVC limitation

infotexture:
Since the number of participants shown in the conversation does not include the current user, we need to change this copy to prevent confusion when the conversation shows “4 people” and the Video Call icon is not available.

### DIFF
--- a/app/script/localization/webapp.js
+++ b/app/script/localization/webapp.js
@@ -379,7 +379,7 @@ z.string.groupCreationParticipantsActionSkip = 'Skip';
 z.string.groupCreationParticipantsHeader = 'Add people';
 z.string.groupCreationParticipantsHeaderWithCounter = 'Add people ({{number}})';
 z.string.groupCreationParticipantsPlaceholder = 'Search by name';
-z.string.groupSizeInfo = 'Up to {{count}} people can join a group conversation. Video calls work in groups of 4 or less.';
+z.string.groupSizeInfo = 'Up to {{count}} people can join a group conversation. Video calls work with up to 3 other people and you.';
 
 // Guest room
 z.string.guestRoomConversationName = 'Guest room';


app/script/localization/webapp.js
```diff
@@ -379,7 +379,7 @@ z.string.groupCreationParticipantsActionSkip = 'Skip';
 z.string.groupCreationParticipantsHeader = 'Add people';
 z.string.groupCreationParticipantsHeaderWithCounter = 'Add people ({{number}})';
 z.string.groupCreationParticipantsPlaceholder = 'Search by name';
-z.string.groupSizeInfo = 'Up to {{count}} people can join a group conversation. Video calls work in groups of 4 or less.';
+z.string.groupSizeInfo = 'Up to {{count}} people can join a group conversation. Video calls work with up to 3 other people and you.';
```
gregor, line 382: @infotexture Just to clarify: We will still show 128 respectively 256 as the number for people in a conversation here and not change that as well?

infotexture: Yes, since that change is not yet live/official, we'll leave it for now and make that change separately. 

Since the number is implemented here with the `{{count}}` placeholder, that change may not even require copy revisions.

gregor: Not talking about 128 vs 256 as that will indeed not require a change here.

The string will show the group size limit while the counter of people in the group is group participants - 1 (you). It has the same problem as the call limit.

gregor: I will go ahead and merge this for now. We can always adjust later.

infotexture: > counter of people in the group

@gregor Yes, that also stays the same for now. Discussed with design, they plan to propose changes for that in a separate step.

PR #3764, opened 2018-06-22
feat: add support for years

AndyLnd:


app/script/util/TimeUtil.js
```diff
@@ -37,6 +37,12 @@ z.util.TimeUtil = {
   formatDuration: (duration, maximumUnits = 1) => {
     const momentDuration = moment.duration(duration);
     const units = [
+      {
+        plural: z.string.ephemeralUnitsYears,
+        singular: z.string.ephemeralUnitsYear,
+        unit: 'y',
+        value: momentDuration.years(),
+      },
```
gregor, line 45: @AndyLnd If we introduce years we should probably also limit the smaller units.

months -> `value: momentDuration.months() % 12`
weeks -> 'value: ....'

AndyLnd: This only affects days, as they are given as 0..31.
Months get clamped to 11 automatically.

This PR is obsolete anyway, as I am changing the implementation right now to only support units up to weeks (and they are going to be rounded as well).

PR #3771, opened 2018-06-22
feat: add support for rounded time units

AndyLnd:


bennycode:
Code works fine until 26 weeks (`1000 * 60 * 60 * 24 * 7 * 26`). When reaching 27 weeks (`1000 * 60 * 60 * 24 * 7 * 26`), then it shows 1 year.

app/script/util/TimeUtil.js
```diff
@@ -28,67 +28,73 @@ z.util.TimeUtil = {
     return Date.now() - timeOffset;
   },
 
+  durationUnits: [
+    {
+      plural: 'ephemeralUnitsYears',
+      singular: 'ephemeralUnitsYear',
+      unit: 'y',
+      value: 1000 * 60 * 60 * 24 * 365,
```
gregor, line 36: @bennyn If we agreed on a new duration for a 365 in milliseconds we should use it everywhere. The validation in the handler still uses 365.25.

PR #3774, opened 2018-06-22
feat: Set and clear ephemeral timer

gregor:


app/script/conversation/ConversationEphemeralHandler.js
```diff
@@ -30,48 +30,65 @@ z.conversation.ConversationEphemeralHandler = class ConversationEphemeralHandler
...
 
-    this.timedMessages = [];
-    window.setInterval(this._updateTimedMessages.bind(this), ConversationEphemeralHandler.CONFIG.INTERVAL_TIME);
+    this.timedMessages = ko.observableArray([]);
+
+    let updateIntervalId = undefined;
+    this.timedMessages.subscribe(messageEntities => {
+      const shouldClearInterval = !messageEntities.length && updateIntervalId;
+      if (shouldClearInterval) {
+        return window.clearInterval(updateIntervalId);
```
AndyLnd: `updateIntervalId` should be set to a falsy value as well.

gregor: Upps.. thanks for the catch

app/script/conversation/ConversationEphemeralHandler.js
```diff
@@ -30,48 +30,68 @@ z.conversation.ConversationEphemeralHandler = class ConversationEphemeralHandler
...
 
   _updateTimedMessages() {
-    const now = Date.now();
-    for (let index = this.timedMessages.length - 1; index >= 0; index--) {
-      const messageEntity = this.timedMessages[index];
+    const currentTimestamp = Date.now();
+    const removeTimedMessage = (messageEntity, index) => {
+      amplify.publish(z.event.WebApp.CONVERSATION.EPHEMERAL_MESSAGE_TIMEOUT, messageEntity);
+      this._timeoutEphemeralMessage(messageEntity);
+      this.timedMessages().splice(index, 1);
```
AndyLnd, line 65: This might be a problem. You are mutating the array inside the observable directly, so the subscribtion won't trigger.

PR #3775, opened 2018-06-22
fix: Update conversation message timer on rejoin

bennyn:

### DIFF
--- a/app/script/conversation/ConversationMapper.js
+++ b/app/script/conversation/ConversationMapper.js
@@ -53,7 +53,7 @@ z.conversation.ConversationMapper = class ConversationMapper {
   update_properties(conversationEntity, conversationData) {
     Object.entries(conversationData).forEach(([key, value]) => {
       if (key !== 'id') {
-        if (value !== undefined) {
+        if (value !== undefined && conversationEntity.hasOwnProperty(key)) {
           if (ko.isObservable(conversationEntity[key])) {
             conversationEntity[key](value);
           } else {
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -616,6 +616,13 @@ z.conversation.ConversationRepository = class ConversationRepository {
     this.updateConversations(this.conversations_unarchived());
   }
 
+  updateConversationFromBackend(conversationEntity) {
+    this.conversation_service.get_conversation_by_id(conversationEntity.id).then(conversationData => {
+      this.conversation_mapper.update_properties(conversationEntity, conversationData);
+      this.conversation_mapper.update_self_status(conversationEntity, conversationData);
+    });
+  }
+
   /**
    * Get users and events for conversations.
    *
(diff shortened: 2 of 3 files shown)


test/unit_tests/conversation/ConversationMapperSpec.js
```diff
@@ -103,6 +103,23 @@ describe('Conversation Mapper', () => {
       expect(updated_conversation_et.id).not.toBe(data.id);
       expect(updated_conversation_et.creator).toBe(data.creator);
     });
+
+    it('only updates existing properties', () => {
```
atomrc, line 107: :+1:

app/script/conversation/ConversationRepository.js
```diff
@@ -616,6 +616,13 @@ z.conversation.ConversationRepository = class ConversationRepository {
     this.updateConversations(this.conversations_unarchived());
   }
 
+  updateConversationFromBackend(conversationEntity) {
+    this.conversation_service.get_conversation_by_id(conversationEntity.id).then(conversationData => {
+      this.conversation_mapper.update_properties(conversationEntity, conversationData);
+      this.conversation_mapper.update_self_status(conversationEntity, conversationData);
```
atomrc: I suggest we add the `name` property handling in the `self_update_status` and remove the call the `update_properties`. Those two lines are very confusing I think, it seems each line is doing the same job.

Since this `update_self_status` is ever only called with data from backend, I would also suggest that we rename it `updateFromBackend` (so that it's clear that it makes the translation between properties from the backend and local properties of the entity).
WDYT @bennyn @gregor ?

gregor:
Nice. In theory we can close a few other bugs with this PR as well. The changed name after rejoin, changed link after rejoin, etc. I will try to find and link them.

app/script/conversation/ConversationMapper.js
```diff
@@ -70,7 +70,7 @@ z.conversation.ConversationMapper = class ConversationMapper {
    * Update the membership properties of a conversation.
    *
    * @param {Conversation} conversation_et - Conversation to be updated
-   * @param {Object} self_state - Conversation self data
+   * @param {Object} self_state - Conversation self data from the database
```
gregor: @bennyn This isn't neccessarily from the database.

app/script/conversation/ConversationRepository.js
```diff
@@ -618,8 +618,9 @@ z.conversation.ConversationRepository = class ConversationRepository {
 
   updateConversationFromBackend(conversationEntity) {
     this.conversation_service.get_conversation_by_id(conversationEntity.id).then(conversationData => {
-      this.conversation_mapper.update_properties(conversationEntity, conversationData);
-      this.conversation_mapper.update_self_status(conversationEntity, conversationData);
+      const {name, message_timer} = conversationData;
+      this.conversation_mapper.update_properties(conversationEntity, {name});
+      this.conversation_mapper.update_self_status(conversationEntity, {message_timer});
```
gregor: @bennyn What we are missing now is a participants list update. People that got removed or added while we were not part of the conversation are not handled.

app/script/conversation/ConversationMapper.js
atomrc: To add a little context in there, what we discovered last Friday is that we handle 3 different format that can represent a conversation.
- the conversation entity (which mainly uses camelcase, and has some computed properties and observables)
- the raw conversation data in database
- the raw conversation data from backend

Both the backend and database differs on a few properties, that's why we though it would be a good idea to start typing, as much as we can, the format of data consumed by the functions.
This is also the reason we decided to cherry-pick the properties updated by each method (one of this method takes database formatted data while the other takes backend formatted data).

PR #3776, opened 2018-06-22
feat: new icon in preferences, svg icons demo, clean out css

AndyLnd:


app/demo/template/icons.htm
```diff
@@ -0,0 +1,33 @@
...
+    <mute-icon></mute-icon>
+    <people-icon></people-icon>
+    <pickup-icon></pickup-icon>
+    <plus-icon></plus-icon>
+    <profile-icon></profile-icon>
+    <screenshare-icon></screenshare-icon>
+    <settings-icon></settings-icon>
+    <undo-icon></undo-icon>
+    <verified-icon class="override-color"></verified-icon>
+</div>
```
gregor: Please add empty line at eof

gregor: Do we have any header in the html files?

app/style/fonts/zeta-neue.css
```diff
@@ -267,21 +213,12 @@
...
 }
-.icon-checkbox:before {
-  content: "\e600";
-}
-.icon-checkbox-checked:before {
-  content: "\e601";
-}
 .icon-contacts:before {
   content: "\e602";
 }
```
gregor, line 224: We shouldn't really manually edit this file. Design creates it as the output when we generated a new icon font. How much work left to remove it completely?

app/demo/template/icons.htm
AndyLnd: These files don't have any headers as they are just fragments.

app/style/fonts/zeta-neue.css
AndyLnd: This file won't see any automatic updates anymore. All new icons now come as svg. Removing the obsolete ones is a good way of keeping track of missing svg implementations and to-be-replaced icons.

gregor: I don't fully agree but let's go with it. I see removing the icon font as one task that will be done at the end if one chance of breaking stuff. Now we possibly break things everytime we add a new SVG based icon and forget about an instance of the old icon font based one.

PR #3794, opened 2018-06-25
chore: Add JSDoc type "ConversationData"

bennyn:


app/script/conversation/ConversationMapper.js
```diff
@@ -22,21 +22,53 @@
...
- * @typedef {object} SelfStatusUpdate
- * @property {number=} archived_timestamp
- * @property {number=} cleared_timestamp
- * @property {number=} ephemeral_timer
- * @property {number=} message_timer
- * @property {number=} last_event_timestamp
- * @property {number=} last_read_timestamp
- * @property {number=} last_server_timestamp
- * @property {boolean=} otr_archived
+ * @typedef {object} ConversationData
```
atomrc: Shouldn't we precise if it is data from the backend or the db?

app/script/conversation/ConversationMapper.js
```diff
@@ -22,21 +22,53 @@
...
+ */
+
+/**
+ * @typedef {object} OtherMember
+ * @property {string} id
+ * @property {number} status
+ */
+
+/**
+ * @typedef {object} SelfStatusUpdate
```
atomrc: This one seems to be incomplete, I believe we can pass a few more properties to be updated in the self status

bennycode: Good catch! Fixed.

app/script/conversation/ConversationMapper.js
bennycode: #changed

PR #3802, opened 2018-06-25
 fix: send ephemeral messages to own devices (WEBAPP-5186)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -2133,11 +2133,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
 
   send_generic_message_to_conversation(conversation_id, generic_message) {
     return this.sending_queue.push(() => {
-      const skip_own_clients = generic_message.content === z.cryptography.GENERIC_MESSAGE_TYPE.EPHEMERAL;
-
-      return this.create_recipients(conversation_id, skip_own_clients).then(recipients => {
```
gregor, line 2137: Let's check if we can remove the `skip_own_clients` functionality then.

atomrc: For the record, it's used in the `send_confirmation_status` method, so no

gregor: Thanks for checking

PR #3805, opened 2018-06-25
fix: Never add expired ephemeral messages to view (WEBAPP-5192)

gregor:


app/script/conversation/ConversationEphemeralHandler.js
```diff
@@ -226,6 +197,16 @@ z.conversation.ConversationEphemeralHandler = class ConversationEphemeralHandler
     this.logger.info(`Obfuscated text message '${messageEntity.id}'`);
   }
 
+  _timeoutEphemeralMessage(messageEntity) {
+    if (!messageEntity.is_expired()) {
+      if (messageEntity.user().is_me) {
+        return this._obfuscateMessage(messageEntity);
+      }
+
+      amplify.publish(z.event.WebApp.CONVERSATION.EPHEMERAL_MESSAGE_TIMEOUT, messageEntity);
```
atomrc, line 206: I would suggest some kind of convention that limits the usage of event dispatchers in leave components/handlers/...

Since the repository is the interface with the external world (mostly views), I would say that it is also the one responsible for warning the external world that something happened (in that case, firing an event).

Therefor, I would suggest to restore the explicit callback passing to the `handler`. If you find the API confusing (passing the callback in the constructor), I believe we can find something better.
Another idea would be 
`handler.onMessageTimeout(this.handleMessageTimeout)` 
for example.

WDYT?

atomrc: The ideal scenario (in my opinion) would be that handlers are local event emitters (not responsible for side effects) and that a parent component could listen to the events it is interested in.

gregor: Look at the previous code though. We called ourselves...

atomrc: Not sure I understand your comment?

gregor:
The initial implementation was too complicated. There where a handful of oversights and confusions in the code because of the somewhat circular structure. I cleaned that up. We definitely need the amplify for the outside world to react. Instead of two routes we now have one consistent one.

In the future the message deletion will be in its own handler. It will be a true sibling to the ephemeral handler. We can't and shouldn't pass it in the constructor.

The other option is to move the code from the repo to the handler and fire events that the message should be deleted. I suppose that would be the semantically correct way.

app/script/conversation/ConversationEphemeralHandler.js
atomrc: you mean the `this.onMessageTimeout`? This is actually a function that comes from the `ConversationRepo`, it is given in the constructor

PR #3817, opened 2018-06-26
fix: delete own pings after timer ends (WEBAPP-5188)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -3392,16 +3392,19 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
     }
+
+    this.get_conversation_by_id(messageEntity.conversation_id).then(conversationEntity => {
+      if (conversationEntity.is_group()) {
+        const user_ids = _.union([this.selfUser().id], [messageEntity.from]);
+        return this.delete_message_everyone(conversationEntity, messageEntity, user_ids);
+      }
+
+      return this.delete_message_everyone(conversationEntity, messageEntity);
+    });
```
gregor: This is so short. Let stick with the positive case and not do and early return. There isn't much indentation either.

app/script/conversation/ConversationRepository.js
```diff
@@ -3392,16 +3392,19 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
+      return;
     }
+
+    this.get_conversation_by_id(messageEntity.conversation_id).then(conversationEntity => {
+      if (conversationEntity.is_group()) {
+        const user_ids = _.union([this.selfUser().id], [messageEntity.from]);
+        return this.delete_message_everyone(conversationEntity, messageEntity, user_ids);
+      }
+
+      return this.delete_message_everyone(conversationEntity, messageEntity);
```
gregor: This is an obsolete return

PR #3821, opened 2018-06-26
fix: Adds accent color to placeholder when local timer is on (WEBAPP-5165)

atomrc:


app/style/content/conversation/input-bar.less
```diff
@@ -53,6 +53,9 @@
     padding-left: 8px;
     color: fade(@graphite, 48%);
   }
+  &--accent::placeholder {
```
gregor: missing empty line above

PR #3826, opened 2018-06-26
fix: Do not use accent color with global timer (WEBAPP-5205)

atomrc:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -248,6 +248,11 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     }
   }
 
+  hasLocalEphemeralTimer() {
```
gregor: This has to be a computed

atomrc: Good catch!! Fixed now

PR #3840, opened 2018-06-27
feat: Rely on IntersectionObserver for avatar viewport check

gregor:


app/script/components/participantAvatar.js
```diff
@@ -142,13 +142,24 @@ z.components.ParticipantAvatar = class ParticipantAvatar {
...
+            this.avatarEnteredViewport = true;
+            _loadAvatarPicture();
+            intersectionObserver.disconnect();
+          }
+        });
+      };
+
+      const options = {root: null, rootMargin: '0px', threshold: 0.0};
+      const intersectionObserver = new IntersectionObserver(onIntersect, options);
+      intersectionObserver.observe(this.element[0]);
```
atomrc: What would you think of using directly the `componentInfo.element` here and start slowly deprecating the jquery element `this.element`?
For the sake of brevity, we can even pass the element as a parameter of the `_createObserver` function

PR #3842, opened 2018-06-27
fix: Trim email input on login (WEBAPP-5122)

Yserz:


app/script/auth/page/Login.js
```diff
@@ -172,7 +172,8 @@ class Login extends React.PureComponent {
         }
       })
       .then(() => {
-        const {email, password, persist} = this.state;
+        const {password, persist} = this.state;
+        const email = this.state.email.trim();
```
lipis, line 176: `toLowerCase()` as well?

Yserz: Let's stick to only trim for now.

PR #3847, opened 2018-06-27
fix: Keeps track of the pie's animation state (WEBAPP-5209)

atomrc:


app/script/components/ephemeralTimer.js
```diff
@@ -25,21 +25,28 @@ window.z.components = z.components || {};
...
+    const dial = componentInfo.element.querySelector('.ephemeral-timer__dial');
 
-    this.remainingTime = messageEntity.ephemeral_expires() - Date.now();
-    this.initialDashOffset = dashLength - (this.remainingTime / duration) * -dashLength;
+    const animatePie = () => {
+      const remainingTime = messageEntity.ephemeral_expires() - Date.now();
+      dial.style.strokeDashoffset = dashLength - (remainingTime / duration) * -dashLength;
+      this.animationFrameCallback = window.requestAnimationFrame(animatePie);
+    };
+    this.animationFrameCallback = window.requestAnimationFrame(animatePie);
```
atomrc, line 37: @gregor I was thinking, since we handle the animation with JavaScript now, maybe we could use a less intensive timeout than `requestAnimationFrame` for our timer?
For example for days long timer we could update it only a few times an hour. I believe we could come up with a small formula to get a smooth animation.

I am going to merge this anyway, maybe an improvement we can make in another PR

PR #3849, opened 2018-06-27
fix: limits pie animation rate depending on total duration

atomrc:


app/script/components/ephemeralTimer.js
```diff
@@ -29,16 +29,21 @@ z.components.EphemeralTimer = class EphemeralTimer {
     const dashLength = 12.6;
     const dial = componentInfo.element.querySelector('.ephemeral-timer__dial');
 
+    const numberOfAnimationSteps = 40;
```
gregor, line 32: Did you test this value and check with design?

app/script/components/ephemeralTimer.js
```diff
@@ -29,16 +29,21 @@ z.components.EphemeralTimer = class EphemeralTimer {
...
-      dial.style.strokeDashoffset = dashLength - (remainingTime / duration) * -dashLength;
-      this.animationFrameCallback = window.requestAnimationFrame(animatePie);
+      const newDashoffset = dashLength - (remainingTime / duration) * -dashLength;
+      dial.style.strokeDashoffset = Math.max(newDashoffset, dashLength);
+      if (newDashoffset >= dashLength) {
+        this.animationTimeout = window.setTimeout(animatePie, animationRate);
+      }
     };
-    this.animationFrameCallback = window.requestAnimationFrame(animatePie);
+    this.animationTimeout = window.setTimeout(animatePie, animationRate);
```
gregor: Do we really want to set timeouts for this? I suppose this could break for long durations. Every timeout greater then 24.5 days will execute immediately. So it should event fit with the max of 1 year. But in the long run? Definitely a potential place of failure that we tried to get rid of during this sprint.

app/script/components/ephemeralTimer.js
mythsunwind: I sat next to Thomas and we did a empirical approach with directly checking it on different values. We think it looks very smooth.

app/script/components/ephemeralTimer.js
gregor: Let's add a upper threshold to prevent this and we are good.

app/script/components/ephemeralTimer.js
```diff
@@ -29,16 +29,22 @@ z.components.EphemeralTimer = class EphemeralTimer {
...
+        this.animationInterval = undefined;
+      }
     };
-    this.animationFrameCallback = window.requestAnimationFrame(animatePie);
+    this.animationInterval = window.setInterval(animatePie, animationIntervalValue);
   }
 
   dispose() {
-    window.cancelAnimationFrame(this.animationFrameCallback);
+    window.clearTimeout(this.animationInterval);
```
gregor: clearInterval

PR #3859, opened 2018-06-28
fix: Check matching element from observer

gregor:


app/script/ui/ViewportObserver.js
```diff
@@ -38,10 +38,12 @@ z.ui.ViewportObserver = (() => {
...
       entries.forEach(entry => {
         if (entry.isIntersecting) {
-          const trackedElement = _getObservedElement(entry.target);
-          trackedElement.callback();
-          intersectionObserver.unobserve(trackedElement.element);
-          _removeObservedElement(trackedElement);
+          const observedElement = _getObservedElement(entry.target);
+          if (observedElement) {
+            observedElement.callback();
+            intersectionObserver.unobserve(observedElement.element);
```
atomrc: We might want to unobserve the `entry.target` still (even if there is no matching element)

PR #3869, opened 2018-06-29
fix: Icon color of rename and timer update system messages (WEBAPP-5224)

gregor:
Applicability to timer update system message still needs to be confirmed with design.


gregor:
Approved by design

PR #3878, opened 2018-06-29
 fix: start timer only if element is not overlayed (WEBAPP-5142) 

atomrc:


app/script/view_model/bindings/CommonBindings.js
```diff
@@ -490,52 +490,63 @@ ko.bindingHandlers.removed_from_view = {
...
+          }
+        });
       }
+    });
+  };
 
-      function _dispose() {
-        z.util.ArrayUtil.removeElement(listeners, _checkElement);
-      }
+  const mutationObserver = new MutationObserver(checkOverlayedElements);
```
atomrc: @gregor I was thinking, probably we can extract this to a UI util somewhere with a nice and clean API (like you did with the IntersectionObserver). But then I am having troubles with naming it. Is it a 
`PaintedObserver` (checks that the element is painted on screen and not hidden behind another
`OverlayedObserver`
`VisibleObserver` (which might not be absolutely true, not sure yet)

app/page/template/_dist/app.htm
```diff
@@ -409,6 +409,7 @@
 <!-- View -->
 <script src="/script/ui/AvailabilityContextMenu.js"></script>
 <script src="/script/ui/ContextMenu.js"></script>
+<script src="/script/ui/OverlayedObserver.js"></script>
 <script src="/script/ui/ShortcutType.js"></script>
 <script src="/script/ui/Shortcut.js"></script>
 <script src="/script/ui/WindowHandler.js"></script>
```
gregor: 🙈 Looks like my ABC is not too good. Can you please change `WindowHandler` and `ViewportObserver` as well.

app/script/ui/OverlayedObserver.js
```diff
@@ -0,0 +1,83 @@
...
+   */
+  const isOverlayed = domElement => {
+    const box = domElement.getBoundingClientRect();
+    const elementAtPoint = document.elementFromPoint(box.x, box.y);
+    return (
+      elementAtPoint &&
+      !elementAtPoint.contains(domElement) &&
+      !domElement.contains(elementAtPoint) &&
+      domElement !== elementAtPoint
+    );
```
gregor: Can we split this into two named variables that explain a little more what you are checking?

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -514,27 +514,23 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
...
-    } else {
-      const start_timer_on_focus = this.conversation.id;
-
-      $(window).one('focus', () => {
-        if (start_timer_on_focus === this.conversation.id) {
-          this.conversation_repository.checkMessageTimer(message_et);
+    return () => {
+      const previouslyActiveConversation = this.conversation.id;
+      const startTimer = () => {
+        if (previouslyActiveConversation === this.conversation.id) {
```
gregor: We can actually simplify the implementation here. No need to keep track of the old conversation ID. All we want to know is whether the message is part of the currently visible conversation to ensure we haven't switched to something else. The ko disposal should already take care of this but to be extra safe we can check `messageEntity.conversation_id === this.conversation.id`

atomrc: Totally true, your condition is far better indeed.

Regarding ko disposal, there is none for this function. The `focus` event is added but never disposed (cf. the else condition here https://github.com/wireapp/wire-webapp/pull/3878/files#diff-7ee8a49d06b25a42ae9cb7821c077920R532 )
Or maybe there is something I do not see, but I think this check is not an extra safety

PR #3879, opened 2018-06-29
chore: Debug utility to check if message has been sent to other own clients

bennyn:


app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,39 @@ z.util.DebugUtil = class DebugUtil {
...
+  haveISentThisMessageToMyOtherClients(messageId) {
+    const userId = wire.app.repository.user.self().id;
+    const conversation = wire.app.repository.conversation.active_conversation();
+    const conversationId = conversation.id;
+    const clientId = wire.app.repository.client.currentClient().id;
+    const message = conversation.get_message_by_id(messageId);
+    const dateTime = new Date(message.timestamp());
+    let amountOfMessagesSent = 0;
+
+    const isOTRMessage = notification => notification.type === 'conversation.otr-message-add';
```
gregor: Let's use the const for this. `z.event.Backend...`

app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,39 @@ z.util.DebugUtil = class DebugUtil {
       .then(() => this.logger.log(`Corrupted Session ID '${sessionId}'`));
   }
 
+  haveISentThisMessageToMyOtherClients(messageId) {
+    const userId = wire.app.repository.user.self().id;
+    const conversation = wire.app.repository.conversation.active_conversation();
+    const conversationId = conversation.id;
+    const clientId = wire.app.repository.client.currentClient().id;
+    const message = conversation.get_message_by_id(messageId);
```
gregor: Let's use the Promise of `get_message_in_conversation_by_id` in ConversationRepository. Then the message does not even have to be visible. All we need is conversation Id and message Id.

app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,39 @@ z.util.DebugUtil = class DebugUtil {
...
+    const wasSentByOurCurrentClient = notification =>
+      notification.from === userId && (notification.data && notification.data.sender === clientId);
+    const hasExpectedTimestamp = notification => notification.time === dateTime.toISOString();
+
+    return wire.app.repository.event.notificationService
+      .getNotifications(undefined, undefined, 10000)
+      .then(({notifications}) =>
+        notifications
+          .map(notification => notification.payload)
+          .reduce((acc, payload) => acc.concat(payload))
```
gregor: Chars are free. acc -> accumulator

app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,39 @@ z.util.DebugUtil = class DebugUtil {
...
+          .filter(isOTRMessage)
+          .filter(isInCurrentConversation)
+          .filter(wasSentByOurCurrentClient)
+          .filter(hasExpectedTimestamp)
```
gregor: Shouldn't we merge these into a single filter function instead of touching every message 4 times?

app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,44 @@ z.util.DebugUtil = class DebugUtil {
...
+    let dateTime = undefined;
+
+    const clientId = wire.app.repository.client.currentClient().id;
+    const userId = wire.app.repository.user.self().id;
+
+    const isOTRMessage = notification => notification.type === z.event.Backend.CONVERSATION.OTR_MESSAGE_ADD;
+    const isInCurrentConversation = notification => notification.conversation === conversationId;
+    const wasSentByOurCurrentClient = notification =>
+      notification.from === userId && (notification.data && notification.data.sender === clientId);
+    const hasExpectedTimestamp = notification => notification.time === dateTime.toISOString();
```
atomrc: I see temporal coupling here with the `dateTime` variable.
If, for some reason, latter we move the order of the `then` callbacks or someone overwrite the `dateTime` variable, things will break (silently or not).

app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,44 @@ z.util.DebugUtil = class DebugUtil {
...
+
+    return wire.app.repository.conversation
+      .get_conversation_by_id(conversationId)
+      .then(conversation => {
+        return wire.app.repository.conversation.get_message_in_conversation_by_id(conversation, messageId);
+      })
+      .then(message => {
+        dateTime = new Date(message.timestamp());
+        return wire.app.repository.event.notificationService.getNotifications(undefined, undefined, 10000);
```
atomrc: One idea to fix the temporal coupling from above, would be to return the message in a `then` of the `getNotification`.

Something like 

```
getNotification(...).then(({ notification }) => ({ notification, message }))
```

atomrc: Then the next function gets the message injected and can check stuff in it

PR #3880, opened 2018-06-29
fix: Map additional known users to conversation participants (WEBAPP-5226)

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -594,27 +593,30 @@ z.conversation.ConversationRepository = class ConversationRepository {
   }
 
   /**
-   * Update users and events for archived conversations currently visible.
+   * Update conversations all conversations.
```
bennycode: `Update conversations all conversations.` -> `Update all conversations.`

PR #3891, opened 2018-07-02
fix: Tooltip texts for timed messages

AndyLnd:


yarn.lock
```diff
@@ -696,8 +696,8 @@ ajv@^6.1.0:
     uri-js "^3.0.2"
 
 ajv@^6.5.0:
-  version "6.5.1"
-  resolved "https://registry.yarnpkg.com/ajv/-/ajv-6.5.1.tgz#88ebc1263c7133937d108b80c5572e64e1d9322d"
+  version "6.5.2"
```
atomrc, line 699: Is that normal that there is a diff on the `yarn.lock`?

AndyLnd: This happened while updating my local environment. It might not hurt to have this patch update in there.

app/script/util/TimeUtil.js
```diff
@@ -80,41 +96,51 @@ z.util.TimeUtil = {
...
 
+  /**
+   * Generate a human readable string of the remaining time
+   * @param {number} duration - the remaining time in milliseconds
+   * @returns {string} readable representation of the remaining time
+   */
+  formatDurationCaption: duration => {
+    const mappedUnits = z.util.TimeUtil.mapUnits(duration, false);
+    const hours = mappedUnits.find(unit => unit.unit === 'h');
+    const minutes = mappedUnits.find(unit => unit.unit === 'm');
```
gregor: Let`s find a better naming here: `unit.unit` is not clear.

app/script/util/TimeUtil.js
```diff
@@ -80,41 +96,51 @@ z.util.TimeUtil = {
...
+    const hasHours = hours.value > 0;
+    const validUnitStrings = [];
+    for (let index = 0; index < mappedUnits.length; index++) {
```
gregor: Any particular reason we are not using a `for...of` here?

AndyLnd: A couple lines down I need the index to get the following element.

PR #3892, opened 2018-07-02
fix: guest link box should not be scrollable (WEBAPP-4980)

AndyLnd:


app/script/view_model/panel/GuestOptionsViewModel.js
```diff
@@ -50,13 +50,19 @@ z.viewModel.panel.GuestOptionsViewModel = class GuestOptionsViewModel {
 
     this.conversationEntity.subscribe(conversationEntity => this._updateCode(this.isVisible(), conversationEntity));
     this.isVisible.subscribe(isVisible => this._updateCode(isVisible, this.conversationEntity()));
+    this.hasAccessCode.subscribe(hasAccessCode => {
```
atomrc: Do we really need both the `afterRender` and the `subscribe` to resize the textarea?
It seems to me that only the afterRender should do the trick but there might be a scenario I am missing?

PR #3895, opened 2018-07-02
fix: respect link case in link previews (WEBAPP-5174)

atomrc:


app/script/util/URLUtil.js
```diff
@@ -75,7 +75,6 @@ z.util.URLUtil = (() => {
    */
   const _getDomainName = (url = '') => {
     return url
-      .toLowerCase()
```
atomrc: This function is only used to display the link preview url. So, unless I am missing something very important, I guess it's pretty safe to remove that lowercasing here...
@gregor @bennyn do you know the reason for lowercasing this?

bennycode: If tests are not failing, you are good to go! 🚦

gregor: The lowercasing was placed to catch cases like `WWW.WIRE.COM` and make sure we show `wire.com` instead. As tlds are case insensitive we should handle this properly and lowercase that part of the url.

app/script/util/URLUtil.js
```diff
@@ -74,11 +74,14 @@ z.util.URLUtil = (() => {
    * @returns {string} Plain URL
    */
   const _getDomainName = (url = '') => {
-    return url
-      .toLowerCase()
-      .replace(/.*?:\/\//, '') // remove protocol
-      .replace(/\/$/, '') // remove trailing slash
-      .replace('www.', '');
+    const isMissingProtocol = url.length > 0 && !url.match(/^https?:\/\//i);
+    url = isMissingProtocol ? `http://${url}` : url;
```
AndyLnd: could also be written in a single line as
```
url = url.replace(/^(?!https?:\/\/)/i, 'http://');
```

PR #3902, opened 2018-07-03
chore: use css animations again for ephemeral timer

AndyLnd:


app/script/components/ephemeralTimer.js
```diff
@@ -23,43 +23,25 @@ window.z = window.z || {};
...
-      dial.style.strokeDashoffset = Math.max(newDashoffset, dashLength);
-      if (newDashoffset === dashLength) {
-        window.clearInterval(this.animationInterval);
-        this.animationInterval = undefined;
-      }
-    };
-    this.animationInterval = window.setInterval(animatePie, animationIntervalValue);
-    animatePie();
+  constructor({message}) {
+    this.started = message.ephemeral_started();
```
atomrc: I think we should keep the name `messageEntity` for consistence purposes

PR #3903, opened 2018-07-03
fix: Release MediaStream initiated after call ended (WEBAPP-3896)

gregor:


app/script/calling/CallingRepository.js
```diff
@@ -113,6 +113,7 @@ z.calling.CallingRepository = class CallingRepository {
    */
   shareCallStates() {
     this.mediaRepository.streamHandler.calls = this.calls;
+    this.mediaRepository.streamHandler.getCallById = this.getCallById.bind(this);
```
atomrc: We cannot let such an important part of the responsibilities of the `StreamHandler` be an optional property set.
Plus, in my opinion, the StreamHandler should not be the one calling `getCallById`.

Can the caller of the function `initiateMediaStream` be the also the one responsible for getting the call?
Why do we actually need the call in there, it doesn't seem to be used in the body of the function?
We need to find another solution for this.

app/script/media/MediaStreamHandler.js
```diff
@@ -445,13 +437,26 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
...
    *
    * @private
+   * @param {string} conversationId - ID of conversation to initiate MediaStream for
    * @param {z.media.MediaStreamInfo} mediaStreamInfo - Type of requested MediaStream
    * @returns {undefined} No return value
    */
-  _initiateMediaStreamSuccess(mediaStreamInfo) {
+  _initiateMediaStreamSuccess(conversationId, mediaStreamInfo) {
     if (mediaStreamInfo) {
+      const conversationNeedsMedia = this.currentCalls.get(conversationId);
```
atomrc, line 446: Do we have the guarantee that the `conversationId` is the same as the `callId`?
Cause here, we compare call ids with conversation ids.

app/script/calling/CallingRepository.js
```diff
@@ -112,8 +112,17 @@ z.calling.CallingRepository = class CallingRepository {
...
-    this.mediaRepository.streamHandler.calls = this.calls;
-    this.mediaRepository.streamHandler.joinedCall = this.joinedCall;
+    this.calls.subscribe(callEntities => {
+      const currentCalls = new Map();
+      callEntities.forEach(callEntity => {
+        const hasPreJoinVideo = callEntity.isIncoming() && callEntity.isRemoteVideoCall();
+        const needsMediaStream = callEntity.selfClientJoined() || hasPreJoinVideo;
+        currentCalls.set(callEntity.id, needsMediaStream && !callEntity.isOngoingOnAnotherClient());
+      });
+      this.mediaRepository.streamHandler.currentCalls = currentCalls;
```
atomrc: Let's convert this to a proper setter instead of a property setting. 
It could ease future debug and has the benefit of throwing errors in case of typos.

app/script/media/MediaStreamHandler.js
gregor: Yes

app/script/calling/CallingRepository.js
```diff
@@ -112,8 +112,17 @@ z.calling.CallingRepository = class CallingRepository {
...
   shareCallStates() {
-    this.mediaRepository.streamHandler.calls = this.calls;
-    this.mediaRepository.streamHandler.joinedCall = this.joinedCall;
+    this.calls.subscribe(callEntities => {
+      this.mediaStreamHandler.clearCurrentCalls();
+      callEntities.forEach(callEntity => {
+        const hasPreJoinVideo = callEntity.isIncoming() && callEntity.isRemoteVideoCall();
+        const hasActiveCall = callEntity.selfClientJoined() || hasPreJoinVideo;
+        const needsMediaStream = hasActiveCall && !callEntity.isOngoingOnAnotherClient();
+        this.mediaStreamHandler.addCurrentCall(callEntity.id, needsMediaStream);
```
gregor, line 121: We can later extend this to include the media type of the call for a more granular check in the `needsMediaStream` method.

PR #3908, opened 2018-07-04
 fix: obfuscate unread timed messages in non focused windows (WEBAPP-5231)

AndyLnd:


gregor:
Addresses concerns raised in https://github.com/wireapp/wire-desktop/issues/928

PR #3920, opened 2018-07-05
 fix: Clears single instance interval check on window unload (WEBAPP-5222)

atomrc:


app/script/main/SingleInstanceHandler.js
```diff
@@ -0,0 +1,148 @@
...
+    _startSingleInstanceCheck() {
+      if (checkInterval) {
+        this._stopSingleInstanceCheck();
+      }
+      checkInterval = window.setInterval(this._checkSingleInstance.bind(this), CONFIG.INTERVAL);
+    }
+
+    _stopSingleInstanceCheck() {
+      window.clearInterval(checkInterval);
+    }
```
gregor: While this works I would be a little more explicit:

Lets move the `if (checkInterval)` check to `_stopSingleInstanceCheck` and also reset the `checkInterval` to undefined. I would also rename `checkInterval` to `checkIntervalId` as that is what this is keeping track off.

app/script/main/app.js
```diff
@@ -559,48 +559,18 @@ z.main.App = class App {
...
-
-    const shouldBlockTab = !singleInstanceCookie || singleInstanceCookie.appInstanceId !== this.instanceId;
-    if (shouldBlockTab) {
-      return this._redirectToLogin(z.auth.SIGN_OUT_REASON.MULTIPLE_TABS);
+  _registerSingleInstance() {
+    if (this.singleInstanceHandler.registerInstance(this.instanceId)) {
+      this._registerSingleInstanceCleaning();
+      return Promise.resolve();
     }
+    return Promise.reject(new Error('App already running in another instance'));
```
gregor: Rejecting with wrong error

gregor: Maybe we should even move the Promise based signature to the handler

gregor:
General comment on the JSDoc. Title ends with ., Desciprtion strings start with uppercase.

app/script/main/app.js
atomrc: Yep, good idea

atomrc: Humm actually, I have second thoughts about this idea.
It seems overkill. If the argument for this is that 'it is convenient for the `App` class, then I think it's not a valid one. 
I think we should use the most primitive types possible for those leaf handlers. 
I would say if it is not async, then no need for promises.
WDYT?

app/script/main/SingleInstanceHandler.js
```diff
@@ -0,0 +1,146 @@
...
+
+  return class SingleInstanceHandler {
+    constructor() {
+      this.instanceId = undefined;
+    }
+
+    /**
+     * Set the cookie to verify we are running a single instace tab.
+     * returns true if the instance has been registered successfully
+     * returns false if the app is already running in another instance
```
gregor: Full stop

app/script/main/SingleInstanceHandler.js
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @param {string} instanceId - the instance id to register
+     * @returns {boolean} - has the app being registered successfully
```
gregor: Uppercase text start

app/script/main/SingleInstanceHandler.js
```diff
@@ -0,0 +1,146 @@
...
+      }
+      Cookies.set(cookieName, {appInstanceId: this.instanceId});
+      return true;
+    }
+
+    /**
+     * Removes the cookie that keeps track of the running instance.
+     *
+     * @param {boolean} forceRemoval - do not check that the instance removing it is the current instance
+     * @returns {void} - returns nothing
```
gregor: Uppercase message start

app/script/main/app.js
```diff
@@ -559,48 +558,19 @@ z.main.App = class App {
...
-
-  _checkSingleInstanceOnInterval() {
-    const singleInstanceCookie = Cookies.getJSON(App.CONFIG.TABS_CHECK.COOKIE_NAME);
-
-    const shouldBlockTab = !singleInstanceCookie || singleInstanceCookie.appInstanceId !== this.instanceId;
-    if (shouldBlockTab) {
-      return this._redirectToLogin(z.auth.SIGN_OUT_REASON.MULTIPLE_TABS);
+  _registerSingleInstance() {
+    if (this.singleInstanceHandler.registerInstance(this.instanceId)) {
+      this.singleInstanceHandler.addExtraInstanceStartedListener(this._onExtraInstanceStarted);
```
gregor: We can further simplify the API if the register instance takes the listener as param right away. No need for an additional step. Currently we don't even need to support multiple listeners. It is overkill or do you see that need for us?

app/script/main/app.js
```diff
@@ -559,48 +558,19 @@ z.main.App = class App {
...
+  _registerSingleInstanceCleaning(singleInstanceCheckIntervalId) {
     $(window).on('beforeunload', () => {
-      const singleInstanceCookie = Cookies.getJSON(App.CONFIG.TABS_CHECK.COOKIE_NAME);
-
-      const isOwnInstanceId = singleInstanceCookie && singleInstanceCookie.appInstanceId === this.instanceId;
-      if (isOwnInstanceId) {
-        Cookies.remove(App.CONFIG.TABS_CHECK.COOKIE_NAME);
-      }
+      this.singleInstanceHandler.removeExtraInstanceStartedListener(this._onExtraInstanceStarted);
+      this.singleInstanceHandler.deregisterInstance();
```
gregor: We can also pass the listener(s) we intend to deregister with this. No need to an extra API call.

app/script/main/app.js
atomrc: No, your totally right. I first implemented this thinking that the login view would also need to listen to this. 
This is overkill indeed.

I think I can come up with something better. Give me a few minutes :D

PR #3936, opened 2018-07-06
chore: Debug utility to retrieve recent messages from database

bennyn:
Open any conversation and execute `await wire.app.util.debug.getLastMessagesFromDatabase()` to retrieve the last 10 messages from the database (latest is on top). The amount of messages and targeted conversation can be changed.

### DIFF
--- a/app/script/util/DebugUtil.js
+++ b/app/script/util/DebugUtil.js
@@ -58,6 +58,13 @@ z.util.DebugUtil = class DebugUtil {
       .then(() => this.logger.log(`Corrupted Session ID '${sessionId}'`));
   }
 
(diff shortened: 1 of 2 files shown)


app/script/util/DebugUtil.js
```diff
@@ -58,6 +58,13 @@ z.util.DebugUtil = class DebugUtil {
       .then(() => this.logger.log(`Corrupted Session ID '${sessionId}'`));
   }
 
+  getLastMessagesFromDatabase(amount = 10, conversationId = wire.app.repository.conversation.active_conversation().id) {
+    return wire.app.repository.storage.storageService.db.events.toArray(records => {
```
atomrc: Would be better to inject the `storageService` in the constructor of the Util instead on relying on global variable

PR #3937, opened 2018-07-06
feat: migrate google maps api to backend proxy (WEBAPP-5242)

atomrc:


app/script/location/LocationRepository.js
```diff
@@ -0,0 +1,30 @@
...
+'use strict';
+
+window.z = window.z || {};
+window.z.location = window.z.location || {};
+
+z.location.LocationRepository = class LocationRepository {
+  constructor(locationService) {
+    this.getMapsUrl = locationService.getMapsUrl.bind(locationService);
+    this.getLocation = locationService.getLocation.bind(locationService);
+  }
```
gregor: Let's bind them in the location service so that if we are covered if ever call them from elsewhere

atomrc: I would argue that this is the responsibility of the consumer to bind the function to the right object.
The class should not be aware how its methods are going to be used.

AndyLnd: Let's go with 
```
this.locationService = locationService;
```
and just call the methods directly from the service where they are needed. So it is clear where these methods belong and which context they are using without the need to awkwardly bind them anywhere.

PR #3950, opened 2018-07-09
 feat: add highlighting capability to the user list (WEBAPP-5240)

AndyLnd:


app/script/view_model/panel/ConversationParticipantsViewModel.js
```diff
@@ -35,6 +35,7 @@ z.viewModel.panel.ConversationParticipantsViewModel = class ConversationParticip
       }
       return [];
     });
+    this.highlightedUsers = ko.observable([]);
```
atomrc, line 38: Do you really need the power of observables here? It seems to me that a native array would do the trick. Am I wrong?

AndyLnd: The renderer needs this observable.

PR #3956, opened 2018-07-10
fix: uses the locationRepository to compute the GoogleMaps url (WEBAPP-5250)

atomrc:


app/page/template/partials/template-message.htm
```diff
@@ -134,7 +134,7 @@
         <file-asset class="message-asset" data-bind="css: {'ephemeral-asset-expired icon-file': message.isObfuscated()}" params="message: message"></file-asset>
       <!-- /ko -->
       <!-- ko if: asset.is_location() -->
-        <location-asset params="asset: asset"></location-asset>
+        <location-asset params="asset: asset, locationRepository: $parents[1].locationRepository"></location-asset>
```
bennycode, line 137: `$parents[1].locationRepository` looks fragile to me. Can we pass in the location repository somehow else?

atomrc: Yeah it's definitely awful. Unfortunately I don't see much options in there (not saying that they don't exist). That's how viewModels are given to the view.
If someone is aware of a better solution, I am all ears

PR #3959, opened 2018-07-10
fix: check connection status before handling ws events (WEBAPP-5245) 

atomrc:


app/script/event/WebSocketService.js
```diff
@@ -115,6 +115,13 @@ z.event.WebSocketService = class WebSocketService {
       };
 
       this.socket.onmessage = event => {
+        if (this._mightHaveExperiencedConnectionIssue()) {
+          /*
+           * If we detected that there might have been connection issues lately,
+           * no need to handle the notifications here, they will be handled upon next reconnection
+           */
+          return;
```
gregor: Let's trigger the reconnect right now and here. And also add a log message...

app/script/event/WebSocketService.js
```diff
@@ -115,6 +116,11 @@ z.event.WebSocketService = class WebSocketService {
       };
 
       this.socket.onmessage = event => {
+        if (this._pingHasExperiencedSuspiciousInactivity()) {
+          const secondsSinceLastPing = (Date.now() - this.lastPingTime) / z.util.TimeUtil.UNITS_IN_MILLIS.SECOND;
+          this.logger.warn(`Message received but ping was inactive for ${secondsSinceLastPing} seconds, reconnecting.`);
```
bennycode: Let's wrap `${secondsSinceLastPing}` inside quotes to indicate that's a variable value.

app/script/event/WebSocketService.js
```diff
@@ -208,20 +214,30 @@ z.event.WebSocketService = class WebSocketService {
...
-      const pingIntervalDifference = this.lastPingTime - currentTime;
-
-      const maxDifference = WebSocketService.CONFIG.PING_INTERVAL + WebSocketService.CONFIG.PING_INTERVAL_THRESHOLD;
-      if (pingIntervalDifference > maxDifference) {
+      if (this._pingHasExperiencedSuspiciousInactivity()) {
         this.logger.warn('Ping interval check failed');
         return this.reconnect(WebSocketService.CHANGE_TRIGGER.PING_INTERVAL);
       }
       this.logger.info('Sending ping to WebSocket');
+      this.lastPingTime = Date.now();
```
bennycode, line 222: Is the `lastPingTime` interval also reset when `reconnect` is being called?

atomrc: No, reconnecting will not do anything with the interval.
Only a `reset` will stop the interval.
In our case, it will continue running and checking the connection every 30s.

app/script/event/WebSocketService.js
atomrc: Not sure it is relevant in that case since it's a number of seconds that I compute ONLY for the warning message (it not a value in the class, and it make the line too long for prettier :D)

app/script/event/WebSocketService.js
atomrc: Ok, so after investigations, the interval is actually killed first when we reconnect then restarted with the new websocket connection.

PR #3961, opened 2018-07-10
feat: Store all team member state on group creation

gregor:


gregor:
Adds logic needed for visualization of this change

PR #3965, opened 2018-07-10
 feat: new system messages (WEBAPP-5238, WEBAPP-5239, WEBAPP-5240, WEBAPP-5241)

AndyLnd:


app/script/entity/message/MemberMessage.js
```diff
@@ -33,6 +41,16 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
     this.userIds = ko.observableArray();
     this.name = ko.observable('');
 
+    this.exceedsMaxVisibleUsers = ko.observable(false);
+    this.visibleUsers = ko.observable([]);
+    this.hiddenUserCount = ko.observable([]);
```
gregor: Default should be 0

app/script/entity/message/MemberMessage.js
```diff
@@ -42,6 +60,21 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
...
 
+    this.joinedUserEntities.subscribe(joinedUserEntities => {
+      const visibleUsers = joinedUserEntities.slice();
+      this.exceedsMaxVisibleUsers(visibleUsers.length > MemberMessage.CONFIG.MAX_USERS_VISIBLE);
+      if (this.exceedsMaxVisibleUsers()) {
+        visibleUsers.splice(MemberMessage.CONFIG.REDUCED_USERS_COUNT);
+        const selfUser = visibleUsers.find(userEntity => userEntity.is_me);
+        if (selfUser) {
+          visibleUsers.pop();
+          visibleUsers.push(selfUser);
```
gregor: The logic seems off. We need to make sure that the self user is never added twice to the visible users but always added as the last enttiy.

app/script/entity/message/MemberMessage.js
```diff
@@ -75,18 +104,47 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
 
         case z.message.SystemMessageType.CONVERSATION_CREATE: {
           if (this.name().length) {
-            return z.l10n.text(z.string.conversationCreateWith, this._generateNameString(z.string.Declension.DATIVE));
+            const exceedsMaxTeam = this.joinedUserEntities().length > MemberMessage.CONFIG.MAX_WHOLE_TEAM_USERS_VISIBLE;
+            if (this.allTeamMembers && exceedsMaxTeam) {
+              const guestCount = this.joinedUserEntities().filter(user => user.isGuest()).length;
+              if (guestCount === 0) {
```
gregor: `!guestCount`

app/script/entity/message/MemberMessage.js
```diff
@@ -75,18 +104,47 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
...
+                return z.l10n.text(z.string.conversationCreateTeam);
+              }
+              if (guestCount === 1) {
+                return z.l10n.text(z.string.conversationCreateTeamGuest);
+              }
+              return z.l10n.text(z.string.conversationCreateTeamGuests, guestCount);
+            }
```
gregor:
````
return guestCount === 1
  ? z.l10n.text(z.string.conversationCreateTeamGuest)
  : z.l10n.text(z.string.conversationCreateTeamGuests, guestCount);
````

File: app/script/entity/message/MemberMessage.js [hunk 5]
```diff
@@ -33,6 +41,16 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
     this.userIds = ko.observableArray();
     this.name = ko.observable('');
 
+    this.exceedsMaxVisibleUsers = ko.observable(false);
```
gregor: Let's change this to a computed

app/page/template/partials/template-message.htm
```diff
@@ -231,14 +228,6 @@
           </div>
         <!-- /ko -->
       </div>
-      <div class="message-body" data-bind="foreach: remoteUserEntities">
```
gregor, line 234: Do we still use remoteUserEntities elsewhere or can we remove it from member messages.

app/script/entity/message/MemberMessage.js
```diff
@@ -42,6 +62,18 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
         .map(userEntity => userEntity);
     });
 
+    this.joinedUserEntities.subscribe(joinedUserEntities => {
```
gregor, line 65: Are they supposed to be sorted alphabetically somehow? This might still be done by the name string.

app/page/template/partials/template-message.htm
AndyLnd: ConversationCellState still relies on it.

app/script/entity/message/MemberMessage.js
```diff
@@ -75,18 +103,46 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
 
         case z.message.SystemMessageType.CONVERSATION_CREATE: {
           if (this.name().length) {
-            return z.l10n.text(z.string.conversationCreateWith, this._generateNameString(z.string.Declension.DATIVE));
+            const exceedsMaxTeam = this.joinedUserEntities().length > MemberMessage.CONFIG.MAX_WHOLE_TEAM_USERS_VISIBLE;
+            if (this.allTeamMembers && exceedsMaxTeam) {
+              const guestCount = this.joinedUserEntities().filter(user => user.isGuest()).length;
```
gregor: Let's stick to the userEntity naming.

app/script/entity/message/MemberMessage.js
```diff
@@ -98,16 +154,24 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
...
 
           const userJoinedStringId = this.user().is_me
-            ? z.string.conversationMemberJoinYou
-            : z.string.conversationMemberJoin;
-
-          return z.l10n.text(userJoinedStringId, this._generateNameString());
+            ? this.exceedsMaxVisibleUsers()
+              ? z.string.conversationMemberJoinedYouMore
+              : z.string.conversationMemberJoinedYou
+            : this.exceedsMaxVisibleUsers()
```
gregor: Conjoined ternaries are hard to read. Can we pull these apart?

app/script/entity/message/MemberMessage.js
AndyLnd: The userEntities are already sorted (https://github.com/wireapp/wire-webapp/blob/dev/app/script/conversation/ConversationRepository.js#L3515)

app/script/util/LocalizerUtil.js
```diff
@@ -23,13 +23,13 @@ window.z = window.z || {};
...
-  joinNames: (userEntities, declension = z.string.Declension.ACCUSATIVE) => {
+  joinNames: (userEntities, declension = z.string.Declension.ACCUSATIVE, skipAnd = false) => {
     const firstNames = userEntities
       .map(userEntity => z.util.getFirstName(userEntity, declension))
       .sort((userNameA, userNameB) => z.util.StringUtil.sortByPriority(userNameA, userNameB));
 
     const numberOfNames = firstNames.length;
-    if (numberOfNames >= 2) {
+    if (!skipAnd && numberOfNames >= 2) {
       const [secondLastName, lastName] = firstNames.splice(firstNames.length - 2, 2);
```
gregor, line 37: Here we sort the names. You/du is the first name of self and will be sorted accordingly. Thus self will not always be last..

AndyLnd: You're right, we should be able to skip this

app/script/entity/message/MemberMessage.js
AndyLnd: We could put these in if-blocks, i will try if that makes it more readable.

PR #3967, opened 2018-07-10
refactor: First name string generation

gregor:


app/script/util/LocalizerUtil.js
```diff
@@ -24,16 +24,21 @@ window.z.util = z.util || {};
 
 z.util.LocalizerUtil = {
   joinNames: (userEntities, declension = z.string.Declension.ACCUSATIVE, skipAnd = false) => {
-    const selfUser = userEntities.find(userEntity => userEntity.is_me);
-    const otherUsers = userEntities.filter(userEntity => !userEntity.is_me);
-    const firstNames = otherUsers.map(userEntity => z.util.getFirstName(userEntity));
+    const containsSelfUser = !!userEntities.find(userEntity => userEntity.is_me);
```
AndyLnd: ```userEntities.some(userEntity => userEntity.is_me)```

app/script/util/util.js
```diff
@@ -512,26 +512,26 @@ z.util.murmurhash3 = (key, seed) => {
...
+    case z.string.Declension.DATIVE:
+      stringId = z.string.conversationYouDative;
+      break;
+    case z.string.Declension.ACCUSATIVE:
+      stringId = z.string.conversationYouAccusative;
+      break;
   }
 
-  return userEt.first_name();
+  return z.l10n.text(stringId);
```
AndyLnd: Maybe we could write it like
```
const youDeclensions = {
  [z.string.Declension.NOMINATIVE]: z.string.conversationYouNominative,
  [z.string.Declension.DATIVE]: z.string.conversationYouDative,
  [z.string.Declension.ACCUSATIVE]: z.string.conversationYouAccusative,
};
return z.l10n.text(youDeclensions[declension]);
```
With less boilerplate it feels more concise.

app/script/util/SanitizationUtil.js
```diff
@@ -0,0 +1,86 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+'use strict';
+
+window.z = window.z || {};
+window.z.util = z.util || {};
+
+z.util.SanitizationUtil = (() => {
```
bennycode, line 25: The `SanitizationUtil` uses this structure:

```js
z.util.SanitizationUtil = (() => {
  const _escapeString = string => _.escape(string);
  return {
    escapeString: _escapeString,
  }
}
```

Which can be written much simpler:

```js
z.util.SanitizationUtil = {
  escapeString: (string) => _.escape(string)
}
```

test/unit_tests/util/SanitizationUtilSpec.js
```diff
@@ -0,0 +1,38 @@
...
+
+describe('z.util.SanitizationUtil', () => {
+  describe('safeWindowOpen', () => {
+    let newWindow = undefined;
+    afterEach(() => {
+      if (newWindow) {
+        newWindow.close();
+      }
+    });
+
```
bennycode, line 67: Let's add a test for `getEscapedFirstName` (and all other methods of `SanitizationUtil`) too.

app/script/util/SanitizationUtil.js
bennycode: The pattern shown above will be a little complicated for methods which use other methods of the same object (like `_getEscapedSelfName`) but therefor we can use a class pattern with static methods.

atomrc: I would say that a `class` for that is too powerful (it allows its consumer to instantiate it, but we don't want that kind of power).
If we don't want something to be instantiated it's better if it is not possible to instantiate it by design (rather than by convention). 
To me, a simple JavaScript object is the perfect abstraction for this util

atomrc: To me, this is a good application of the `rule of least power`. (here it is about choosing the right language for the right job, but we can easily extend it to choosing the right abstraction).
https://www.w3.org/2001/tag/doc/leastPower.html

bennycode: **Very good reasoning with facts!** I would be very happy if we can reduce private functions (starting with the underscore pattern) to a minimum because this is what bothers me the most with the current pattern. Do you have an idea how we can achieve that by keeping the rule of least power?

atomrc: Yes, I might have a solution, thanks for asking :)
I believe here we actually don't need the `_` notation!!!

Let me explain. The `_` notation is here to say to consumer of our class "you should not use this method, please".
Here, because those functions are in an isolated scope (they are only published to the consumer if we decide to expose them), we actually don't need this notation. I would go like this:

```
z.util.SanitizationUtil = (() => {
  // here are the functions definition
  const escapeString = string => _.escape(string);
  return {
     // here you decide what API you publish
    escapeString
  }
}
```

AndyLnd: Which is basically how it is written now, except for the underscores (which i wouldn't miss at all) :)

PR #3970, opened 2018-07-11
fix: {{name}} was removed from team

AndyLnd:


app/script/entity/message/MemberMessage.js
```diff
@@ -199,7 +199,7 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
         }
 
         case z.event.Client.CONVERSATION.TEAM_MEMBER_LEAVE: {
-          return z.l10n.text(z.string.conversationTeamLeave);
```
gregor, line 202: Lets remove the old strng as well then,

PR #3973, opened 2018-07-11
fix: Explicitely update event in DB when receiving a link preview message (WEBAPP-4366)

atomrc:
Quick explanation:

# Initial behavior when a link preview is received

- the EventRepository merges the event with the previous text message already in DB (if found)
- the EventRepository save this "new" event in DB (so there are 2 duplicates in DB right now)
- the ConversationRepo takes the relay and delete the previous event (supposedly the text event).

Work well in most of the cases, but will actually delete the `linkPreview` event if there is no `textEvent` already in DB (which could occur if, say, a client send a `linkPreviewEvent` before the actual `textEvent`)

# New behavior

- the EventRepository merges the event with the previous text message already in DB (if found) ;
- the EventRepository **updates** the event in DB if found ;
- the ConversationRepo is no more responsible for touching the DB (deleting the previous event).

### DIFF
--- a/app/script/entity/Conversation.js
+++ b/app/script/entity/Conversation.js
@@ -335,16 +335,22 @@ z.entity.Conversation = class Conversation {
 
   /**
    * Adds a single message to the conversation.
-   * @param {z.entity.Message} message_et - Message entity to be added to the conversation
-   * @returns {undefined} No return value
+   * @param {z.entity.Message} messageEntity - Message entity to be added to the conversation.
+   * @param {boolean} replaceDuplicate - If a duplicate already exists, replace it with the new entity.
+   * @returns {undefined} No return value.
    */
-  add_message(message_et) {
-    message_et = this._checkForDuplicate(message_et);
-    if (message_et) {
-      this.update_timestamps(message_et);
-      this.messages_unordered.push(message_et);
-      amplify.publish(z.event.WebApp.CONVERSATION.MESSAGE.ADDED, message_et);
+  add_message(messageEntity, replaceDuplicate = false) {
+    this.update_timestamps(messageEntity);
+    if (replaceDuplicate) {
+      const duplicateEntity = this._findDuplicate(messageEntity);
+      if (duplicateEntity) {
+        const duplicateIndex = this.messages_unordered.indexOf(duplicateEntity);
+        this.messages_unordered.splice(duplicateIndex, 1, messageEntity);
+        return;
+      }
     }
+    this.messages_unordered.push(messageEntity);
+    amplify.publish(z.event.WebApp.CONVERSATION.MESSAGE.ADDED, messageEntity);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


app/script/entity/Conversation.js
```diff
@@ -461,24 +467,29 @@ z.entity.Conversation = class Conversation {
...
+    if (existingMessageEntity) {
+      const logData = {additionalMessage: messageEntity, existingMessage: existingMessageEntity};
+      this.logger.warn(`Filtered message '${messageEntity.id}' as duplicate in view`, logData);
+      return undefined;
     }
+    return messageEntity;
+  }
+
+  _findDuplicate(messageEntity) {
+    for (const existingMessageEntity of this.messages_unordered()) {
```
AndyLnd: Why not use ```this.messages_unordered().find()```?

atomrc: quick anwser: lazyness :D 

Good point :+1:

app/script/event/EventRepository.js
```diff
@@ -638,7 +638,13 @@ z.event.EventRepository = class EventRepository {
...
       if (storedEvent) {
         const {data: mappedData, from: mappedFrom, type: mappedType, time: mappedTime} = event;
-        const {data: storedData, from: storedFrom, type: storedType, time: storedTime} = storedEvent;
+        const {
+          data: storedData,
+          from: storedFrom,
+          type: storedType,
+          time: storedTime,
+          primary_key: storedPrimaryKey,
+        } = storedEvent;
```
gregor: We only use the primary key in the case all previous checks passed. Let's simplify the destructuring, remote the primary_key entry here so it fits into a single line and explicitly set it in L685

app/script/entity/Conversation.js
```diff
@@ -481,15 +481,18 @@ z.entity.Conversation = class Conversation {
...
     }
-    return undefined;
+    const areSameMessage = (messageEntity1, messageEntity2) => {
+      const sameId = messageEntity1.id && messageEntity2.id === messageEntity1.id;
+      const sameSender = messageEntity1.from === messageEntity2.from;
+      return sameId && sameSender;
+    };
+
+    return this.messages_unordered().find(existingMessageEntity =>
+      areSameMessage(messageEntity, existingMessageEntity)
```
gregor: Why do you create the overhead of moving the find functionality out of find?

gregor: No need for an early return otherwise.

app/script/entity/Conversation.js
```diff
@@ -335,16 +335,24 @@ z.entity.Conversation = class Conversation {
...
+    this.update_timestamps(messageEntity);
+    if (duplicateEntity && !replaceDuplicate) {
+      // If there is a duplicate in db, but we don't want to replace it our job here is done.
+      return;
     }
+    if (duplicateEntity) {
+      const duplicateIndex = this.messages_unordered.indexOf(duplicateEntity);
+      this.messages_unordered.splice(duplicateIndex, 1, messageEntity);
+      return;
+    }
```
gregor: We also still need the  existential check on the message entity here if I am not mistaken.  `update_timestamps()` and the others were previously safeguarded by this and now isn't anymore.

Please also move the early return into the `if(duplicateEntity)` check instead of doing the same thing twice.

PR #3976, opened 2018-07-11
chore: sortObjectByKeys, printDevicesId refactor

AndyLnd:


app/script/util/util.js
```diff
@@ -480,13 +473,13 @@ z.util.printDevicesId = id => {
   }
 
   const idWithPadding = z.util.zeroPadding(id, 16);
-  let prettifiedId = '';
+  const prettifiedId = [];
 
   for (const part of idWithPadding.match(/.{1,2}/g)) {
-    prettifiedId += `<span class='device-id-part'>${part}</span>`;
+    prettifiedId.push(`<span class='device-id-part'>${part}</span>`);
   }
```
gregor: If you want to do it modern you should do a map in the match return array.

gregor: The only problem with this is that match can return undefined...

PR #3977, opened 2018-07-11
fix: bold usernames in system messages, consistent oxford comma

AndyLnd:


app/script/util/LocalizerUtil.js
```diff
@@ -23,15 +23,19 @@ window.z = window.z || {};
...
       userEntities = userEntities.filter(userEntity => !userEntity.is_me);
     }
 
-    const firstNames = userEntities.map(userEntity => userEntity.first_name());
+    let firstNames = userEntities.map(userEntity => userEntity.first_name());
     firstNames.sort((userNameA, userNameB) => z.util.StringUtil.sortByPriority(userNameA, userNameB));
 
+    if (boldNames) {
+      firstNames = firstNames.map(name => `[bold]${name}[/bold]`);
+    }
```
gregor: Let's move this into the previous map. No need to iterate again...

````
const firstNames = userEntities.map(userEntity => {
  const firstName = userEntity.first_name();
  return boldNames ? `[bold]${firstName}[/bold]` : firstName );
});
````

gregor: Sorry for missing the sort.

PR #3980, opened 2018-07-11
chore: Add debug utility to reprocess OTR messages

bennyn:
1. Added `wire.app.util.debug.reprocessNotificationStream()` which fetches OTR messages for the current client and sends them to the processing chain
1. Seperated `EventRepository._handleEvent` into a verification step and processing step
1. Wrote test for the processing step


app/script/util/DebugUtil.js
```diff
@@ -262,4 +262,30 @@ z.util.DebugUtil = class DebugUtil {
     this.logger.log(`-- IndexedDB open: ${this.storageRepository.storageService.db.isOpen()}`);
     this.logger.log(`-- WebSocket ready state: ${window.wire.app.service.web_socket.socket.readyState}`);
   }
+
+  reprocessNotificationStream(conversationId = this.conversationRepository.active_conversation().id) {
+    const clientId = wire.app.repository.client.currentClient().id;
```
gregor: You have access to the repositories directly

app/script/util/DebugUtil.js
```diff
@@ -262,4 +262,30 @@ z.util.DebugUtil = class DebugUtil {
...
+          .reduce((accumulator, payload) => accumulator.concat(payload))
+          .filter(notification => {
+            return isOTRMessage(notification) && isInCurrentConversation(notification);
+          });
+      })
+      .then(events => {
+        this.logger.info(`Reprocessing "${events.length}" OTR messages...`);
+        for (const event of events) {
+          this.eventRepository._processEvent(event, z.event.EventRepository.SOURCE.STREAM);
+        }
```
gregor: Is there a particular reason for the preference for a `for. .. of` loop over `Array.forEach()`

app/script/event/EventRepository.js
```diff
@@ -578,7 +578,21 @@ z.event.EventRepository = class EventRepository {
...
+   * @private
+   * @param {JSON} event - Backend event extracted from notification stream
+   * @param {z.event.EventRepository.SOURCE} source - Source of event
+   * @returns {Promise} Resolves with the saved record or boolean true if the event was skipped
+   */
+  _processEvent(event, source) {
+    return Promise.resolve(event)
       .then(validatedEvent => {
         const isEncryptedEvent = validatedEvent.type === z.event.Backend.CONVERSATION.OTR_MESSAGE_ADD;
         return isEncryptedEvent ? this.cryptographyRepository.handleEncryptedEvent(event) : event;
```
gregor: What's the benefit of resolving the event if you continue to use the initial one like in the previous code. Let's skip that step and ensure we always use event instead.

PR #3988, opened 2018-07-11
feat: Improve event loading performance

gregor:
Thx to @atomrc for the debugging session looking into the bottle necks of our load performance.

This is a simple but powerful optimization:
Loading events from the database is a process that slows down over time with the database growing and containing more events per conversation.

If we load a specific event from the conversation we use the conversation index to get all events in that conversation and then filter them in JavaScript. The more events the conversation contains the slower this process will be.

This reverses this approach. Instead we rely on the message ID index. Message IDs are UUID that are unique within a conversation. Chances that different conversations contain the same IDs is slim by the entropy of UUIDs. Only then do we filter by conversation in JavaScript iterating over mostly one database record only.

In manual tests I have seen improvements in the call from 150ms down to 10ms only. 🎆

### DIFF
--- a/app/script/conversation/ConversationService.js
+++ b/app/script/conversation/ConversationService.js
@@ -470,18 +470,24 @@ z.conversation.ConversationService = class ConversationService {
   /**
    * Load conversation event.
    *
-   * @param {string} conversation_id - ID of conversation
-   * @param {string} message_id - ID of message to retrieve
+   * @param {string} conversationId - ID of conversation
+   * @param {string} messageId - ID of message to retrieve
    * @returns {Promise} Resolves with the stored record
    */
-  load_event_from_db(conversation_id, message_id) {
+  load_event_from_db(conversationId, messageId) {
+    if (!conversationId || !messageId) {
+      this.logger.error(`Cannot get event '${messageId}' in conversation '${conversation_id}' without IDs`);
+      throw new z.conversation.ConversationError(z.conversation.ConversationError.TYPE.MESSAGE_NOT_FOUND);
+    }
+
     return this.storageService.db[this.EVENT_STORE_NAME]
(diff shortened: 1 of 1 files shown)


gregor:
If you want to give the perf improvement a try:

````
const getMessageOld = (conversationId, messageId) => {
  const t0 = performance.now();
  return wire.app.service.storage.db[z.storage.StorageSchemata.OBJECT_STORE.EVENTS]
    .where('conversation')
    .equals(conversationId)
    .filter(record => messageId && record.id === messageId)
    .first()
    .then(message => {
      const t1 = performance.now();
      console.log('OLD: Loaded message in milliseconds', t1 - t0, message);
    });
};

const getMessageNew = (conversationId, messageId) => {
  const t0 = performance.now();
  return wire.app.service.storage.db[z.storage.StorageSchemata.OBJECT_STORE.EVENTS]
    .where('id')
    .equals(messageId)
    .filter(record => record.conversation === conversationId)
    .first()
    .then(message => {
      const t1 = performance.now();
      console.log('NEW: Loaded message in milliseconds', t1 - t0, message);
    });
};

let conversationEntity;
let conversationId;
let messageId;
````

````
conversationEntity = wire.app.repository.conversation.active_conversation()
conversationId = conversationEntity.id;
messageId = conversationEntity.messages()[conversationEntity.messages().length - 1].id;
getMessageOld(conversationId, messageId).then(() => getMessageNew(conversationId, messageId));
````

atomrc:
That's one hell of a performance improvement and this was a really exciting investigation session. 
Great team effort👏

gregor:
Just tested again on a machine with a much longer history and thus many more messages in the database:

594ms vs 2.7ms 🎆

PR #3991, opened 2018-07-12
chore: add the eslint no-undef rule

atomrc:
This should catch errors like this one at commit time https://github.com/wireapp/wire-webapp/pull/3990


gregor:
Thanks for adding this rule again.

PR #3995, opened 2018-07-12
feat: Show "create a group" to first-time users (WEBAPP-5266)

bennyn:
When being a new user on Wire, our app asks you to "Start a conversation or create a group" but there is no button to create a group if you have no contacts on Wire. 

**Status quo:**

a

Since we can create a group with ourselves, it makes sense to show a "Create a group" option so that first time users can see our feature set and start a playground conversation.


app/page/template/list/start-ui.htm
```diff
@@ -114,6 +114,12 @@
   </div>
 
   <!-- ko if: showNoContacts() -->
+    <ul class="start-ui-create-group left-list-items">
+      <li class="left-list-item left-list-item-clickable" data-bind="click: clickOnCreateGroup" data-uie-name="go-create-group">
+        <group-icon class="left-column-icon"></group-icon>
+        <div class="center-column" data-bind="l10n_text: z.string.searchCreateGroup"></div>
+      </li>
+    </ul>
```
gregor: Can we find a solution that does not require duplication of the html?

bennycode: The Andymator and I found a CSS solution to the problem. 👍

app/style/list/start-ui.less
```diff
@@ -63,13 +63,20 @@
   overflow: hidden;
   flex: 1 1;
   flex-direction: column;
+  &.split-view {
```
gregor: We usually have an empty line above new declarations

bennycode: 👍

PR #3996, opened 2018-07-12
feat: Add safe html binding (WEBAPP-5262)

gregor:


app/script/localization/Localizer.js
```diff
@@ -48,17 +48,52 @@ class Localizer {
...
 
     return string;
   };
 
+  const replaceWithObject = (string, substitutions) => replaceWithArray(Object.entries(substitutions));
+
+  const replaceWithString = (string, substitute) => string.replace(/{{\w+}}/, substitute);
+
   return {
+    safeHtml(value, unsafeSubstitute, safeSubstitutes = {}) {
```
atomrc: The substitutions should be a single object that has the keys `safe` and `unsafe`. Cause all those scenarii are valid:
- safeHTML without any substitution;
- safeHTML with only unsafeSubstitution;
- safeHTML with only safeSubstitution;
- safeHTML with both safe and unsafe substitutions.

app/script/localization/webapp.js
```diff
@@ -357,8 +357,8 @@ z.string.conversationToday = 'today';
 z.string.conversationVoiceChannelDeactivate = ' called';
 z.string.conversationVoiceChannelDeactivateYou = ' called';
 z.string.conversationYesterday = 'Yesterday';
-z.string.conversationUnableToDecrypt1 = 'A message from {{user}} was not received.';
-z.string.conversationUnableToDecrypt2 = '{{user}}´s device identity changed. Undelivered message.';
+z.string.conversationUnableToDecrypt1 = 'A message from [highlight]{{user}}[/highlight] was not received.';
```
AndyLnd, line 360: i think these could be just bold, right?

app/script/view_model/bindings/LocalizationBindings.js
```diff
@@ -48,10 +48,10 @@ ko.bindingHandlers.l10n_text = {
   },
 };
 
-ko.bindingHandlers.l10n_html = {
+ko.bindingHandlers.l10nSafeHtml = {
   update(element, valueAccessor = z.util.noop) {
-    const {id = valueAccessor(), substitute} = valueAccessor();
-    ko.utils.setHtml(element, z.l10n.text(id, substitute));
+    const {id = valueAccessor(), unsafeSubstitute, safeSubstitutes} = valueAccessor();
```
AndyLnd: this will fail with the default value of `noop`.

app/page/template/warning.htm
```diff
@@ -75,28 +75,28 @@
...
         </div>
       <!-- /ko -->
 
       <!-- ko if: visibleWarning() === z.viewModel.WarningsViewModel.TYPE.UNSUPPORTED_INCOMING_CALL -->
         <div class="warning-bar warning-bar-feature bg-theme">
           <!-- ko ifnot: z.util.Environment.browser.chrome -->
-            <div class="warning-bar-message"><span data-bind="l10n_text: {'id': z.string.warningCallUnsupportedIncoming, substitute: z.util.SanitizationUtil.escapeString(name())}"></span>&nbsp;<a class="warning-bar-link" rel="nofollow noopener noreferrer" target="_blank" data-bind="l10n_href: {support: z.config.SUPPORT.ID.CALLING}, l10n_text: z.string.warningLearnMore"></a></div>
+            <div class="warning-bar-message">
+              <span data-bind="l10n_text: {'id': z.string.warningCallUnsupportedIncoming, substitute: z.util.SanitizationUtil.escapeString(name())}"></span>&nbsp;<a class="warning-bar-link" rel="nofollow noopener noreferrer" target="_blank" data-bind="l10n_href: {support: z.config.SUPPORT.ID.CALLING}, l10n_text: z.string.warningLearnMore"></a>
```
atomrc: I believe you can even remove the `z.util.SanitizationUtil.escapeString` here (since it's now the default behavior of the binding)

app/script/entity/message/MemberMessage.js
```diff
@@ -108,42 +125,60 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
...
-          return z.l10n.text(createStringId, {
+
+          unsafeSubstitutions = {
             count: this.hiddenUserCount(),
             name: this.senderName(),
             users: this._generateNameString(this.exceedsMaxVisibleUsers(), z.string.Declension.DATIVE),
-          });
+          };
+
+          return z.l10n.safeHtml(createStringId, unsafeSubstitutions, tagSubstitutions);
```
atomrc: All the calls to `safeHtml` have the wrong signature.

should be 
```
safeHtml(string, { replace: {}, dangerouslyReplace: {}})
```

app/script/localization/Localizer.js
```diff
@@ -80,7 +115,7 @@ z.l10n = (() => {
       if (_.isObject(substitute)) {
```
atomrc, line 116: I believe text can rely on `safeHtml` for the `text` method, it's basically

```
text(value, substitutes) {
  return safeHtml(value, {replaceDangerously: substitues});
}
```

atomrc: could be done later though

PR #4006, opened 2018-07-12
fix: Recomputes message category when link preview is received (WEBAPP-5263)

atomrc:


app/script/event/EventRepository.js
```diff
@@ -698,6 +698,7 @@ z.event.EventRepository = class EventRepository {
       event.server_time = event.time;
       event.time = storedEvent.time;
       event.primary_key = storedEvent.primary_key;
+      event.category = z.message.MessageCategorization.categoryFromEvent(event);
```
atomrc, line 701: @gregor I hesitated to put this logic in the `update_event` method of the `conversationService` but not sure if it makes sense in all the cases. WDYT?

gregor: LGTM

PR #4021, opened 2018-07-13
 chore: refactor l10n

AndyLnd:


test/unit_tests/localization/LocalizerSpec.js
```diff
@@ -130,7 +130,7 @@ describe('l10n', () => {
           // no safe replacement
           expected: '<user>Felix</user>',
           params: {
-            substitutions: {replaceDangerously: {'\\/user': '</user>', user: '<user>'}},
+            substitutions: {replaceDangerously: {'/user': '</user>', user: '<user>'}},
```
atomrc, line 138: Could you just add the `number` test case for the `text` method, please. Then I think we are covered :+1:

app/script/localization/Localizer.js
```diff
@@ -48,51 +48,42 @@ class Localizer {
...
-    });
-
-    return string;
+  const replaceSubstitute = (string, regex, substitute) => {
+    if (isStringOrNumber(substitute)) {
+      return string.replace(regex, substitute);
+    }
+    return string.replace(
+      regex,
+      (found, content) => (substitute.hasOwnProperty(content) ? substitute[content] : found)
```
gregor: Can we move this to a const `replaceFunction` so that the actual replace call turns out to be a oneliner?

AndyLnd: prettier strikes again ...

gregor: Or maybe even:
````
const replacement = isStringOrNumber(substitute)
  ? substitute
  : (found, content) => (substitute.hasOwnProperty(content) ? substitute[content] : found);+      
return string.replace(regex, replacement);
````

PR #4026, opened 2018-07-13
fix: some system message icons are off center

AndyLnd:


app/style/content/conversation/message-list.less
```diff
@@ -137,10 +136,9 @@
     font-weight: @font-weight-bold;
     text-transform: capitalize;
   }
-
-  & > span,
-  & > a {
-    margin-top: -6px;
+  & > hr {
```
gregor, line 139: missing empty line

PR #4038, opened 2018-07-13
chore: enable automatic login redirect from app (WEBAPP-5164)

AndyLnd:


app/script/main/app.js
```diff
@@ -756,14 +756,15 @@ z.main.App = class App {
...
-        let url = `${baseUrl}${location.search}`;
+        let url = `${baseUrl}${window.location.search}`;
         const isImmediateSignOutReason = App.CONFIG.SIGN_OUT_REASONS.IMMEDIATE.includes(signOutReason);
         if (isImmediateSignOutReason) {
           url = z.util.URLUtil.appendParameter(url, `${z.auth.URLParameter.REASON}=${signOutReason}`);
         }
 
         const redirectToLogin = signOutReason !== z.auth.SIGN_OUT_REASON.NOT_SIGNED_IN;
-        if (redirectToLogin) {
+        const forcedAppLogin = z.util.Environment.desktop && z.window.location.hash === '#login';
```
atomrc: What is that `z.window` I don't find it in the codebase?

atomrc: Isn't it supposed to be a `||` here?
If the app is the desktop or that the current location contains `login`?

AndyLnd: Just a typo, already fixed

AndyLnd: I want to keep the '#login' only if it is in the context of the app.

app/script/main/app.js
```diff
@@ -756,14 +756,15 @@ z.main.App = class App {
...
-        let url = `${baseUrl}${location.search}`;
+        let url = `${baseUrl}${window.location.search}`;
         const isImmediateSignOutReason = App.CONFIG.SIGN_OUT_REASONS.IMMEDIATE.includes(signOutReason);
         if (isImmediateSignOutReason) {
           url = z.util.URLUtil.appendParameter(url, `${z.auth.URLParameter.REASON}=${signOutReason}`);
         }
 
         const redirectToLogin = signOutReason !== z.auth.SIGN_OUT_REASON.NOT_SIGNED_IN;
-        if (redirectToLogin) {
+        const forcedAppLogin = z.util.Environment.desktop && window.location.hash === '#login';
```
atomrc, line 766: Isn't it supposed to be a || here?
If the app is the desktop or that the current location contains login?

AndyLnd: Nope, see above.

AndyLnd:
This is related to https://github.com/wireapp/wire-desktop/pull/1607

app/script/main/app.js
gregor: Do we really need this dependency between wrapper and webapp? Would be nice to not have to think about this in the future with different versions of stuff here and there.

PR #4039, opened 2018-07-13
fix: don't obfuscate unsent messages (WEBAPP-5270)

AndyLnd:


app/script/entity/message/Message.js
```diff
@@ -65,7 +65,9 @@ z.entity.Message = class Message {
     });
 
     this.isObfuscated = ko.pureComputed(() => {
-      return this.ephemeral_status() === z.message.EphemeralStatusType.INACTIVE || this.is_expired();
+      const unsendStatus = [z.message.StatusType.SENDING, z.message.StatusType.FAILED];
+      const messageIsSent = !unsendStatus.includes(this.status());
```
gregor: We can simplify this check. The status is an enum based on increasing number for exactly this reason.

`const messageIsAtLeastSent = this.status() > z.message.StatusType.SENDING`

app/script/entity/message/Message.js
```diff
@@ -65,7 +65,9 @@ z.entity.Message = class Message {
...
+      return messageIsSent && (this.ephemeral_status() === z.message.EphemeralStatusType.INACTIVE || this.is_expired());
```
gregor: Let's pull the ephemeral_status check into a const

PR #4041, opened 2018-07-13
fix: scroll link preview into view when received (WEBAPP-5273)

atomrc:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -269,27 +269,31 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
 
   /**
    * Checks how to scroll message list and if conversation should be marked as unread.
-   * @param {Array} messages - Message entities
+   * @param {Array} messageDiffs - List of the messages that were added or removed from the list
    * @returns {undefined} No return value
    */
-  _on_message_add(messages) {
+  _scrollAddedMessagesIntoView(messageDiffs) {
```
gregor: Let's change that name. If is a single array containing all the changes -> messagesDiff. But the naming is weird to I would suggest `changedMessages`

atomrc: Good for me

PR #4042, opened 2018-07-13
Change guests option label per WEBAPP-5223

infotexture:
Align item label with other platforms and prepare for new **Options** section title.


infotexture:
@gregor Feel free to push additional commits to this branch to implement the **Options** section header per WEBAPP-5223.

gregor:
Changed as part of #3868

PR #4065, opened 2018-07-17
fix: Add safeguard to platform checks

### LINKED ISSUE #3685: platform check failed on `TypeError: window.platform.os.family is null`
Hello.

Looks like my browser is failing checks at `app/script/util/Environment.js`:

```
window.platform.ua:
"Mozilla/5.0 (X11; DragonFly x86_64; rv:60.0) Gecko/20100101 Firefox/60.0"
window.platform.os.family:
null
```

Hence check on the line 77 in Environment.js breaks due to `windows.platform.os.family` being null.

I don't know whether it's fine to treat my platform as Linux, but I think it should work fine anyway.

gregor:
Fixes #3685

### DIFF
--- a/app/script/util/Environment.js
+++ b/app/script/util/Environment.js
@@ -62,7 +62,8 @@ z.util.Environment = (() => {
   };
 
   const _getVersion = () => {
-    const [majorVersion] = window.platform.version.split('.');
+    const browserVersion = window.platform.version || [];
+    const [majorVersion] = browserVersion.split('.');
     return window.parseInt(majorVersion, 10);
   };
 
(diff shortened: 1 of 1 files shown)


app/script/util/Environment.js
```diff
@@ -62,7 +62,8 @@ z.util.Environment = (() => {
   };
 
   const _getVersion = () => {
-    const [majorVersion] = window.platform.version.split('.');
+    const browserVersion = window.platform.version || [];
+    const [majorVersion] = browserVersion.split('.');
```
atomrc: Array doesn't have a method `split`. Should be defaulting to an empty string.

Also you could inline it
```
const [majorVersion] = (window.platform.version || '').split('.');
```

PR #4069, opened 2018-07-17
chore: Use npm registry

ffflorian:
Yarn uses `registry.yarnpkg.com` as package registry by default.
This change lets yarn use only `registry.npmjs.com` as registry.


ffflorian:
@gregor To always rely on the "official" npm registry and not on private registries. We do the same in [wire-web-packages](https://github.com/wireapp/wire-web-packages/blob/master/.npmrc).

I believe it helps resolving issues like the recent [eslint-scope security incident](https://github.com/eslint/eslint-scope/issues/39) faster.

PR #4072, opened 2018-07-18
fix: don't escape text that isn't meant to (WEBAPP-5281)

AndyLnd:


app/script/localization/Localizer.js
```diff
@@ -101,7 +101,7 @@ z.l10n = (() => {
     text(value, substitute) {
       let string = ko.unwrap(value);
       string = replaceSubstitute(string, /{{(.+?)}}/g, substitute);
-      return z.util.SanitizationUtil.escapeString(string);
+      return string;
```
gregor, line 104: Let return replace... right away. No need for the extra line.

PR #4075, opened 2018-07-18
fix: remove deprecated mute call shortcut (WEBAPP-4210)

atomrc:
This shortcut is not supposed to be there anymore.
Fixes a conflict between the `mute call` and the `mute conversation` shortcuts.

### DIFF
--- a/app/script/ui/Shortcut.js
+++ b/app/script/ui/Shortcut.js
@@ -52,19 +52,6 @@ z.ui.Shortcut = (() => {
         },
       },
     },
-    [z.ui.ShortcutType.CALL_REJECT]: {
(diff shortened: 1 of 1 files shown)


app/script/ui/Shortcut.js
```diff
@@ -65,19 +65,6 @@ z.ui.Shortcut = (() => {
...
-        electron: {
-          macos: 'command + alt + m',
-          pc: 'ctrl + alt + m',
-        },
-        webapp: {
-          macos: 'command + alt + m',
-          pc: 'ctrl + alt + m',
-        },
-      },
-    },
```
gregor, line 80: I think we should also remove the reject shortcut as the last remaining call realted shortcut.

PR #4079, opened 2018-07-18
 fix: Let the EventRepository being responsible for saving events (WEBAPP-5283)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -63,8 +64,10 @@ z.conversation.ConversationRepository = class ConversationRepository {
     giphy_repository,
     link_repository,
     team_repository,
-    user_repository
+    user_repository,
+    eventRepository
```
atomrc: We now need the `eventRepository` in there, since we need feedback on when events are processed by this repo.
To me, it's a relevant dependency as the `ConversationRepository` is aware that events exists and is dependent on how events are handled.

gregor: Let's keep the repositories in alphabetical order.

PR #4080, opened 2018-07-18
fix: BackendError constructor is still needed for synthetic errors

Yserz:


app/script/auth/module/action/BackendError.js
```diff
@@ -18,6 +18,15 @@
  */
 
 export default class BackendError extends Error {
+  constructor(params) {
+    super();
+    this.name = this.constructor.name;
+    this.stack = new Error().stack;
```
bennycode: We can drop the `new Error().stack` pattern when inheriting from `Error`.

PR #4084, opened 2018-07-18
feat: Add clipboard util (WEBAPP-5286)

gregor:


app/script/components/copyToClipboard.js
```diff
@@ -0,0 +1,47 @@
...
+  constructor({text, trigger}) {
+    this.text = text;
+
+    trigger(() => z.util.ClipboardUtil.copyText(document.getElementById('copy-to-clipboard__source').innerText));
+  }
+
+  onClick() {
+    if (window.getSelection) {
+      const selectionRange = document.createRange();
+      selectionRange.selectNode(document.getElementById('copy-to-clipboard__source'));
```
atomrc: Humm when might need to have dynamic ids. If there are 2 `copy-text` component in the page, the selector will always select the first element (and you might end up copying the wrong text)

And idea could be let `ko` inject the root element of the component in the constructor and then select by a class only in that root DOM element.
Using the `createViewModel factory function` method would be to right way to do so I believe http://knockoutjs.com/documentation/component-registration.html

AndyLnd: That's the way to go. We also shouldn't use an ```id``` here.

PR #4093, opened 2018-07-19
fix: Handle events that fail validation

gregor:
Events there did not pass validation because they are being ignored (`conversation.is-typing`) or because of outdated timestamps were no longer properly handled and resulted in lots of errors in the console.

### DIFF
--- a/app/script/event/EventRepository.js
+++ b/app/script/event/EventRepository.js
@@ -587,7 +587,7 @@ z.event.EventRepository = class EventRepository {
           throw error;
         }
 
-        return true;
+        return event;
       });
   }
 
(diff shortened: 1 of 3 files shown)


app/script/event/EventRepository.js
```diff
@@ -579,39 +579,37 @@ z.event.EventRepository = class EventRepository {
...
-    );
+    return this._handleEventValidation(event, source)
+      .then(validatedEvent => this.processEvent(validatedEvent, source))
+      .catch(error => {
+        const isIgnoredError = EventRepository.CONFIG.IGNORED_ERRORS.includes(error.type);
+        if (!isIgnoredError) {
+          throw error;
+        }
+
+        return true;
```
atomrc: Can we actually return the `event` there?
Else we need to change the consumer side to properly handle that `true` (I had the case in my current devs where I triggers some other crashes because the consumer always expect an event

gregor: Good point. Tried to not change anything from the previous version in that regard but that does not make much sense.

PR #4094, opened 2018-07-19
feat: Add option to copy text message (WEBAPP-5286)

gregor:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -570,6 +570,13 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
       });
     }
 
+    if (message_et.has_asset_text()) {
+      entries.push({
+        click: () => message_et.copy(),
```
bennycode, line 575: I would just go with `click: message_et.copy`.

atomrc: Then you would need to go `click: message_et.copy.bind(message_et)` cause it needs `this` inside the function.
I guess the arrow function is shorter in that case.

PR #4100, opened 2018-07-19
 refactor: move DB updates from ConversationRepository to the EventRepository

atomrc:
Ok so, this is a first step toward freeing the `ConversationRepository` from persisting events in the DB.

Here I migrated the `edit message` functionality. Now the `EventRepository` is responsible for checking if there are duplicates in the DB and updates them accordingly. 

This avoids multiple writes to the DB when an event arrives. Previously the `EventRepo` would save most of the event, and, eventually, the `ConversationRepo` would remove them from there (which triggered the bug fixed here https://github.com/wireapp/wire-webapp/pull/4079).

There is still one place where the `ConversationRepository` stores events, for the `asset` handling. 
I might need the help of @gregor on this one, since I am not sure how to reproduce the case where it happens. https://github.com/wireapp/wire-webapp/blob/59d4a22f7873abe72f6c734870f8720fa1f58f8c/app/script/conversation/ConversationRepository.js#L3249

### DIFF
--- a/app/script/entity/Conversation.js
+++ b/app/script/entity/Conversation.js
@@ -335,16 +335,18 @@ z.entity.Conversation = class Conversation {
   /**
    * Adds a single message to the conversation.
    * @param {z.entity.Message} messageEntity - Message entity to be added to the conversation.
-   * @param {boolean} replaceDuplicate - If a duplicate already exists, replace it with the new entity.
+   * @param {boolean} replaceDuplicate - If a duplicate (or a message that should be replaced) already exists, replace it with the new entity.
    * @returns {undefined} No return value.
    */
   add_message(messageEntity, replaceDuplicate = false) {
     if (messageEntity) {
-      const duplicateEntity = this._findDuplicate(messageEntity);
+      const entityToReplace =
+        this._findDuplicate(messageEntity.id, messageEntity.from) ||
+        this._findDuplicate(messageEntity.replacing_message_id, messageEntity.from);
       this.update_timestamps(messageEntity);
-      if (duplicateEntity) {
+      if (entityToReplace) {
         if (replaceDuplicate) {
-          const duplicateIndex = this.messages_unordered.indexOf(duplicateEntity);
+          const duplicateIndex = this.messages_unordered.indexOf(entityToReplace);
           this.messages_unordered.splice(duplicateIndex, 1, messageEntity);
         }
         // The duplicated message has been treated (either replaced or ignored). Our job here is done.
@@ -475,7 +477,7 @@ z.entity.Conversation = class Conversation {
    */
   _checkForDuplicate(messageEntity) {
     if (messageEntity) {
-      const existingMessageEntity = this._findDuplicate(messageEntity);
+      const existingMessageEntity = this._findDuplicate(messageEntity.id, messageEntity.from);
       if (existingMessageEntity) {
         const logData = {additionalMessage: messageEntity, existingMessage: existingMessageEntity};
         this.logger.warn(`Filtered message '${messageEntity.id}' as duplicate in view`, logData);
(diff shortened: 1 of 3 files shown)


app/script/event/EventRepository.js
```diff
@@ -647,7 +647,8 @@ z.event.EventRepository = class EventRepository {
    * @returns {Promise} Resolves with the saved event
    */
   _handleEventSaving(event, source) {
-    const {conversation: conversationId, id: eventId} = event;
+    const conversationId = event.conversation;
+    const eventId = (event.data || {}).replacing_message_id || event.id;
```
gregor: Let's move this back to the object destructuring so we can set a default there:

````
const {conversation: conversationId, data: eventData = {}} = event;
const eventId = eventData.replacing_message_id || event.id;
````

File: app/script/event/EventRepository.js [hunk 2]
```diff
@@ -672,7 +673,8 @@ z.event.EventRepository = class EventRepository {
 
       const mappedIsMessageAdd = mappedType === z.event.Client.CONVERSATION.MESSAGE_ADD;
       const storedIsMessageAdd = storedType === z.event.Client.CONVERSATION.MESSAGE_ADD;
-      const userReusedId = !mappedIsMessageAdd || !storedIsMessageAdd || !mappedData.previews.length;
+      const userReusedId =
+        !mappedIsMessageAdd || !storedIsMessageAdd || !(mappedData.previews.length || mappedData.replacing_message_id);
```
gregor: Let's explain what we filter for in `!(mappedData.previews.length || mappedData.replacing_message_id)` by pulling it out.

app/script/event/EventRepository.js
```diff
@@ -686,22 +688,42 @@ z.event.EventRepository = class EventRepository {
...
+     * The only two valid cases for a duplicate event are:
+     *  - a link preview has been received (the text message already being in DB) ;
+     *  - a message has been updated
+     */
+    if (newData && newData.replacing_message_id) {
+      updates = Object.assign({}, newEvent, primaryKeyUpdate, {
+        edited_time: newEvent.time,
+        time: originalEvent.time,
+      });
+    } else if (newData && newData.previews && newData.previews.length) {
```
gregor: Let's pull these into `const isMessageEdit` and is `const isLinkPreviewUpdate`

app/script/entity/Conversation.js
```diff
@@ -335,16 +335,18 @@ z.entity.Conversation = class Conversation {
...
-   * @param {boolean} replaceDuplicate - If a duplicate already exists, replace it with the new entity.
+   * @param {boolean} replaceDuplicate - If a duplicate (or a message that should be replaced) already exists, replace it with the new entity.
    * @returns {undefined} No return value.
    */
   add_message(messageEntity, replaceDuplicate = false) {
     if (messageEntity) {
-      const duplicateEntity = this._findDuplicate(messageEntity);
+      const entityToReplace =
+        this._findDuplicate(messageEntity.id, messageEntity.from) ||
+        this._findDuplicate(messageEntity.replacing_message_id, messageEntity.from);
```
gregor: Let's move these into separate const that detail what kind of duplicate we retrieved.

app/script/event/EventRepository.js
```diff
@@ -686,22 +688,52 @@ z.event.EventRepository = class EventRepository {
...
+      case isLinkPreviewUpdate:
+        // case of a link preview
+        updates = Object.assign({}, newEvent, primaryKeyUpdate, {
+          category: z.message.MessageCategorization.categoryFromEvent(newEvent),
+          server_time: newEvent.time,
+          time: originalEvent.time,
+        });
+        break;
+      default:
+      // throw error?
```
atomrc: ping @gregor

gregor: Yeah. Throwing an error seems like a good idea:

````
this.logger.error(`Unhandled ID duplication by event '${logMessage}' of type '${newEvent.type}'`, event);
const errorMessage = 'Event validation failed: Unhandled event duplicate';
throw new z.event.EventError(z.event.EventError.TYPE.VALIDATION_FAILED, errorMessage);
````

PR #4104, opened 2018-07-20
fix: Uses the middle point of an element to detect overlays (WEBAPP-5290)

atomrc:
Quick explanation:

To detect if an element (say a timed message) is visible to the user, we check if there is an element overlaying it.
To do that, we used to use the top left corner of the element and ask the browser what is at this location.

Because of floating point rounding, under certain conditions, the element given by the browser would be another element (not the timed message). Thus the algorithm replies that the element is overlayed (thus not visible to the user).

The fix, just uses the middle point of the element instead of an edge point.


app/script/ui/OverlayedObserver.js
```diff
@@ -53,7 +53,9 @@ z.ui.OverlayedObserver = (() => {
    */
   const isOverlayed = domElement => {
     const box = domElement.getBoundingClientRect();
-    const elementAtPoint = document.elementFromPoint(box.left, box.top);
+    const middlePointX = box.left + (box.right - box.left) / 2;
```
AndyLnd: left + (right - left)/2
= left + right/2 - left/2
= left - left/2 + right/2
= left/2 + right/2
= (left+right)/2

atomrc: nice catch

PR #4107, opened 2018-07-20
test: OverlayObserver tests

atomrc:


test/unit_tests/ui/OverlayObserverSpec.js
```diff
@@ -0,0 +1,100 @@
...
+      const element = document.createElement('div');
+      element.style.height = '10px';
+      document.body.appendChild(element);
+
+      overlayObserver.onElementVisible(element, callbackSpy.onVisible);
+
+      expect(callbackSpy.onVisible).toHaveBeenCalled();
+    });
+
+    it('does not calls the callback if the element is overlayed', () => {
```
bennycode: `does not calls` -> `does not call`

test/unit_tests/ui/OverlayObserverSpec.js
```diff
@@ -0,0 +1,100 @@
...
+      spyOn(callbackSpy, 'onVisible');
+
+      const overlay = document.createElement('div');
+      overlay.style.position = 'absolute';
+      overlay.style.top = '0';
+      overlay.style.left = '0';
+      overlay.style.height = '100px';
+      overlay.style.width = '100px';
+      const element = document.createElement('div');
+      element.style.position = 'absolute';
```
bennycode, line 54: Would this also work? 

```js
const element = document.createElement('div');
element.style = {
  position: 'absolute',
  top: '10px',
  left: '10px',
  height: '10px',
  width: '10px',
}
```

test/unit_tests/ui/OverlayObserverSpec.js
```diff
@@ -0,0 +1,100 @@
...
+      document.body.appendChild(element);
+      document.body.appendChild(overlay);
+
+      overlayObserver.onElementVisible(element, callbackSpy.onVisible);
+
+      expect(callbackSpy.onVisible).not.toHaveBeenCalled();
+
+      document.body.removeChild(overlay);
+
+      setTimeout(() => {
```
bennycode: Let's use `window.setTimeout` to make clear that this method comes from the window namespace.

test/unit_tests/ui/OverlayObserverSpec.js
atomrc: I guess it would, let's try :)

atomrc: Ah no it doesn't work. I guess it's because the `style` object has a special type (`CSS2Properties`) that cannot just be replaced with a vanilla Object

PR #4171, opened 2018-07-27
fix: Repair flaky window title update test

bennyn:


test/unit_tests/view_model/WindowTitleViewModelSpec.js
```diff
@@ -192,24 +193,25 @@ describe('z.viewModel.WindowTitleViewModel', () => {
...
-      let expected_title = `(${waiting_people}) · ${message} · ${suffix}`;
+      let expectedWaitingPeople = '1';
+      let expected_title = `(${expectedWaitingPeople}) · ${message} · ${suffix}`;
       title_view_model.initiateTitleUpdates();
       expect(window.document.title).toBe(expected_title);
 
-      // Test multiple connect request messages
+      // Test multiple connect request messages and observe the title change
+      title_view_model.userRepository.connect_requests.subscribe(() => {
+        window.setTimeout(() => {
```
gregor: Should we set a timeout in tests or rely on jasmine clock functionality for this?

bennycode: What would be the benefit of using `jasmine.clock` here? Looking at [its source code](https://github.com/jasmine/jasmine/blob/74287c578cb7a9c5a775fba849f8e7c71f6c8194/src/core/Clock.js), it also just uses `setTimeout` underneath.

atomrc: Oh that could actually be a good idea. It would avoid slowing down our tests. 

With it you can basically tell jasmine to "jump" 250ms later in time instantly. This could serve as a base example for a time based test

bennycode: Ok, will add it.

PR #4180, opened 2018-07-30
fix: Differentiate link preview updates and edit messages (WEBAPP-5300)

atomrc:
Here is a little graph of the workflow when a new event arrives in the event repository handler.
screenshot from 2018-07-31 15-26-08

### DIFF
--- a/app/script/event/EventRepository.js
+++ b/app/script/event/EventRepository.js
@@ -657,8 +657,7 @@ z.event.EventRepository = class EventRepository {
     return findEventToReplacePromise.then(eventToReplace => {
       const hasLinkPreview = mappedData.previews && mappedData.previews.length;
       if (!eventToReplace && mappedData.replacing_message_id && !hasLinkPreview) {
-        const errorMessage = 'Event validation failed: edit event is missing original event';
-        throw new z.event.EventError(z.event.EventError.TYPE.VALIDATION_FAILED, errorMessage);
+        this._throwValidationError('Edit event without original event');
       }
 
       const handleEvent = newEvent => {
@@ -681,13 +680,10 @@ z.event.EventRepository = class EventRepository {
 
   _handleEventReplacement(originalEvent, newEvent) {
     const newData = newEvent.data || {};
-    const logMessage = `Ignored '${newEvent.type}' (${newEvent.id}) in '${newEvent.conversation}' from '${
-      newEvent.from
-    }':'`;
     if (originalEvent.data.from !== newData.from) {
-      this.logger.warn(`${logMessage} ID previously used by user '${newEvent.from}'`, newEvent);
-      const errorMessage = 'Event validation failed: ID reused by other user';
-      throw new z.event.EventError(z.event.EventError.TYPE.VALIDATION_FAILED, errorMessage);
+      const logMessage = `ID previously used by user '${newEvent.from}'`;
+      const errorMessage = 'ID reused by other user';
+      this._throwValidationError(errorMessage, logMessage);
     }
     const primaryKeyUpdate = {primary_key: originalEvent.primary_key};
     const isLinkPreviewEdit = newData.previews && !!newData.previews.length;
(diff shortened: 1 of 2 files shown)


app/script/event/EventRepository.js
```diff
@@ -648,7 +648,15 @@ z.event.EventRepository = class EventRepository {
...
     const {conversation: conversationId, data: mappedData} = event;
-    const eventId = (mappedData && mappedData.replacing_message_id) || event.id;
+    /*
+     * We can detect an edit event if the event has a replacing_message_id and
+     * no preview yet.
+     * This way, we avoid collisions with link preview of edited messages which have
+     * replacing_message_id and a non empty array of previews.
+     */
+    const hasLinkPreview = mappedData && mappedData.previews && !!mappedData.previews.length;
+    const isEditEvent = mappedData && !hasLinkPreview && mappedData.replacing_message_id;
```
gregor: A message can be both: A link preview and an edited message at the same time.

We need to handle all of these cases. With the change we also noticed that an edited message that is not a link preview should only ever replace an existing message. Message edits do not appear retroactively if the device did not exist at the time of the original message.

app/script/event/EventRepository.js
```diff
@@ -647,79 +647,86 @@ z.event.EventRepository = class EventRepository {
...
-    return loadPromise.then(storedEvent => {
-      if (!storedEvent) {
-        return this.conversationService.save_event(event);
+    const conversationId = event.conversation;
+    const mappedData = event.data || {};
+
+    //first check if a message that should be replaced exists in DB
+    const findEventToReplacePromise = mappedData.replacing_message_id
+      ? this.conversationService.load_event_from_db(mappedData.replacing_message_id)
+      : Promise.resolve(undefined);
```
bennycode: Why not just go with `Promise.resolve()` instead of `Promise.resolve(undefined)`?

app/script/event/EventRepository.js
```diff
@@ -647,79 +647,86 @@ z.event.EventRepository = class EventRepository {
...
+
+    return findEventToReplacePromise.then(eventToReplace => {
+      const hasLinkPreview = mappedData.previews && mappedData.previews.length;
+      if (!eventToReplace && mappedData.replacing_message_id && !hasLinkPreview) {
```
bennycode: I suggest to also create a boolean variable for `!eventToReplace && mappedData.replacing_message_id`.

app/script/event/EventRepository.js
```diff
@@ -647,79 +647,86 @@ z.event.EventRepository = class EventRepository {
...
+        this._throwValidationError('Edit event without original event');
       }
 
-      const {from: mappedFrom, type: mappedType} = event;
-      const {from: storedFrom, type: storedType} = storedEvent;
-
-      const logMessage = `Ignored '${mappedType}' (${eventId}) in '${conversationId}' from '${mappedFrom}':'`;
+      const handleEvent = newEvent => {
+        // check for duplicates (same id)
+        const loadPromise = newEvent.id
```
bennycode: `loadPromise` -> `loadEvent`

app/script/event/EventRepository.js
```diff
@@ -647,79 +647,86 @@ z.event.EventRepository = class EventRepository {
...
+      return eventToReplace ? this._handleEventReplacement(eventToReplace, event) : handleEvent(event);
     });
   }
 
-  _handleDuplicateEvents(originalEvent, newEvent) {
-    const newData = newEvent.data;
+  _handleEventReplacement(originalEvent, newEvent) {
+    const newData = newEvent.data || {};
+    if (originalEvent.data.from !== newData.from) {
+      const logMessage = `ID previously used by user '${newEvent.from}'`;
```
bennycode, line 685: `logMessage` -> `successMessage`

app/script/event/EventRepository.js
atomrc: Good catch

app/script/event/EventRepository.js
atomrc: Humm, it is not a success message.

app/script/event/EventRepository.js
```diff
@@ -646,79 +646,88 @@ z.event.EventRepository = class EventRepository {
...
-      if (userReusedId) {
-        this.logger.warn(`${logMessage} ID previously used by same user`, event);
-        const errorMessage = 'Event validation failed: ID reused by same user';
-        throw new z.event.EventError(z.event.EventError.TYPE.VALIDATION_FAILED, errorMessage);
-      }
+        return loadEventPromise.then(storedEvent => {
+          if (storedEvent) {
+            return this._handleLinkPreviewUpdate(storedEvent, newEvent);
+          }
+          return this.conversationService.save_event(newEvent);
```
gregor:
````
return storedEvent
  ? this._handleLinkPreviewUpdate(storedEvent, newEvent)
  : this.conversationService.save_event(newEvent);
````

PR #4182, opened 2018-07-30
fix: Only pre-validate SSO code on desktop app

Yserz:


app/script/auth/page/SingleSignOn.js
```diff
@@ -113,8 +113,10 @@ class SingleSignOn extends React.PureComponent {
         if (errors.length) {
           throw errors[0];
         }
+        if (isDesktopApp()) {
+          this.props.validateSSOCode(this.stripPrefix(this.state.code));
```
gregor: you need to return it to be blocking...

PR #4192, opened 2018-07-30
fix: Improve safeguard for active conversation in panel

gregor:


app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -141,24 +143,30 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
...
+        : false;
     });
 
-    this.showActionGuestOptions = ko.pureComputed(() => this.activeConversation().inTeam());
+    this.showActionGuestOptions = ko.pureComputed(() => {
+      return this.activeConversation() && this.activeConversation().inTeam();
+    });
     this.showActionTimedMessages = ko.pureComputed(() => {
-      return this.activeConversation() && this.activeConversation().is_group() && !this.activeConversation().isGuest();
+      return this.activeConversation()
```
atomrc, line 167: Not sure what is the point of that change?

gregor: No real point. I just aligned them. If there is a single condition i do the && if there are multiple conditions I use the ternary instead.

PR #4201, opened 2018-07-31
chore: Update translations

Yserz:


gregor:
We will have to do this again in a bit. German translations are being done right now. Maybe we should have waited a tiny bit...

PR #4209, opened 2018-07-31
fix: Check conversations affected by device changes (WEBAPP-5301)

gregor:


app/script/entity/Conversation.js
```diff
@@ -105,10 +105,14 @@ z.entity.Conversation = class Conversation {
     this.is_cleared = ko.pureComputed(() => this.last_event_timestamp() <= this.cleared_timestamp());
     this.is_muted = this.muted_state;
     this.is_verified = ko.pureComputed(() => {
-      if (this.self && (this.participating_user_ets().length || !this.participating_user_ids().length)) {
-        const all_users = [this.self].concat(this.participating_user_ets());
-        return all_users.every(user_et => user_et.is_verified());
+      const hasMappedUsers = this.participating_user_ets().length || !this.participating_user_ids().length;
```
atomrc, line 108: Quick question out of the scope of this PR, but I don't really understand this condition.

Why do we consider that there are some `mappedUsers` when either there are some user entities in the `participating_user_ets` array **or** if the `participating_user_ids` is **empty**?

app/script/conversation/ConversationVerificationStateHandler.js
```diff
@@ -166,18 +171,31 @@ z.conversation.ConversationVerificationStateHandler = class ConversationVerifica
   }
 
   /**
-   * Get all conversation where self user is active
+   * Get all conversation where self user and the given users are active.
+   *
    * @private
+   * @param {Array<string>} userIds - Multiple user IDs (can include self user ID)
    * @returns {Array<z.entity.Conversation>} Array of conversation entities
```
atomrc: Return signature needs to be updated

app/script/conversation/ConversationVerificationStateHandler.js
```diff
@@ -166,18 +171,31 @@ z.conversation.ConversationVerificationStateHandler = class ConversationVerifica
...
+          const selfUserId = this.conversationRepository.selfUser().id;
+          const userIdsInConversation = conversationEntity.participating_user_ids().concat(selfUserId);
+          const matchingUserIds = _.intersection(userIdsInConversation, userIds);
+
+          if (!!matchingUserIds.length) {
+            return {conversationEntity, userIds: matchingUserIds};
+          }
+        }
+      })
+      .filter(activeConversationInfo => activeConversationInfo);
```
atomrc: Let's explicitly cast it into a boolean `!!activeConversationInfo`

app/script/client/ClientRepository.js
```diff
@@ -545,22 +540,47 @@ z.client.ClientRepository = class ClientRepository {
...
+
+  /**
+   * Update clients of a user with the given backend data.
+   *
+   * @private
+   * @param {string} userId - ID of user whose clients are updated
+   * @param {Object} clientsData - Clients data from backend
+   * @returns {Promise} Resolves when the clients have been updated
+   */
+  _updateClientsByUserId(userId, clientsData) {
```
atomrc: I find it quite confusion to have a `_updateClientByUserId` and `_updateClientForUser` that both have exactly the same signature and that basically do the same stuff (except that one also `amplify.publish`).

Can't we unify those 2 (eventually add an optional param that says if the method should silent or not regarding `amplify`).
WDYT?

gregor: Good point

PR #4213, opened 2018-07-31
fix: improve perceived performance by un-parallelizing crypto operations (WEBAPP-5328)

atomrc:
Before the change we parallelized the generation of the optimistic event and the actual sending of the message.
Thing is, because of the single-threaded nature of JavaScript, if a crypto operation started before the optimistic event is generated, it would block this same generation.

Instead of parallelizing, we now generate the optimistic event, add it to the view and only then start sending the message (which will trigger all the crypto operations needed)

Time to show optimistic message (on my machine) when from `~100ms` down to `~20ms`

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -2148,48 +2148,37 @@ z.conversation.ConversationRepository = class ConversationRepository {
     });
   }
 
-  _send_and_inject_generic_message(conversation_et, generic_message, sync_timestamp = true) {
+  _send_and_inject_generic_message(conversationEntity, genericMessage, syncTimestamp = true) {
     return Promise.resolve()
       .then(() => {
-        if (conversation_et.removed_from_conversation()) {
+        if (conversationEntity.removed_from_conversation()) {
           throw new Error('Cannot send message to conversation you are not part of');
         }
 
-        const optimistic_event = z.conversation.EventBuilder.buildMessageAdd(conversation_et, this.timeOffset);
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationRepository.js
```diff
@@ -2170,15 +2170,21 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
+          return this.send_generic_message_to_conversation(conversation_et.id, generic_message).then(sentPayload => {
+            return {
+              event: injectedEvent,
+              sentPayload,
+            };
+          });
+        });
 
         /**
          * We will, in parallel, inject events to the repo (where they will be processed and saved in DB)
```
AndyLnd: The comment needs to be updated. :)

app/script/conversation/ConversationRepository.js
```diff
@@ -2148,42 +2148,37 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
     return Promise.resolve()
       .then(() => {
-        if (conversation_et.removed_from_conversation()) {
+        if (conversationEntity.removed_from_conversation()) {
           throw new Error('Cannot send message to conversation you are not part of');
         }
 
-        const optimistic_event = z.conversation.EventBuilder.buildMessageAdd(conversation_et, this.timeOffset);
-        return this.cryptography_repository.cryptographyMapper.mapGenericMessage(generic_message, optimistic_event);
+        const optimistic_event = z.conversation.EventBuilder.buildMessageAdd(conversationEntity, this.timeOffset);
```
bennycode: `optimistic_event` -> `optimisticEvent`

PR #4219, opened 2018-08-01
fix: Case sensitive SSO-Code (WEBAPP-5326)

Yserz:


app/script/auth/page/SingleSignOn.js
```diff
@@ -313,7 +313,12 @@ class SingleSignOn extends React.PureComponent {
       : '';
   };
 
-  stripPrefix = code => code && code.trim().replace(SingleSignOn.SSO_CODE_PREFIX, '');
+  stripPrefix = code =>
+    code &&
+    code
+      .toLowerCase()
+      .trim()
```
gregor: No a real performance issue here but I would always `trim` first and then do the `toLowerCase` on less chars.

app/script/auth/page/SingleSignOn.js
```diff
@@ -313,7 +313,12 @@ class SingleSignOn extends React.PureComponent {
...
+      .replace(SingleSignOn.SSO_CODE_PREFIX, '');
```
atomrc: Also I wonder if we really want to remove the prefix *after* the lowercasing.

I know that, for now, the prefix is a lowercase one, but what if someday there is an uppercase in there? It will break that code.
Do you think we can first remove the prefix and then do the other operations?

gregor: @atomrc Good point but I somewhat disagree. I think the safe way to go is to also run a `toLowerCase` on the prefix const. This way all we need to be aware of is that the whole string is case insensitive.

gregor: We had a short f2f discussion and concluded to not add further changes. Lowercasing first before replacing the prefix will sure that we we are fully case insensitive. The prefix is under our control and not expected to ever change. Thus adding a `toLowercase` on the prefix const is computational overhead that we will not add.

PR #4228, opened 2018-08-01
chore: Refactor versioned changes message update

gregor:


app/script/conversation/ConversationService.js
```diff
@@ -709,42 +709,72 @@ z.conversation.ConversationService = class ConversationService {
...
-                  update_version: changes.version,
-                };
-              }
-
-              Raygun.send('Failed sequential database update', custom_data);
-              throw new z.storage.StorageError(z.storage.StorageError.TYPE.NON_SEQUENTIAL_UPDATE);
-            });
-          });
-        }
+  updateMessageInDb(messageEntity, changes = {}, conversationId) {
```
atomrc: The `conversationId` is never used in the body

app/script/conversation/ConversationService.js
```diff
@@ -709,42 +709,72 @@ z.conversation.ConversationService = class ConversationService {
...
+              databaseVersion: databaseVersion,
+              updateVersion: changes.version,
+            };
+          }
+
+          const logMessage = 'Failed sequential database update';
+          this.logger.error(logMessage, customData);
+
+          Raygun.send(new Error(logMessage), customData);
+          throw new z.storage.StorageError(z.storage.StorageError.TYPE.NON_SEQUENTIAL_UPDATE);
```
atomrc: Maybe we should differentiate between `non sequential updates` and `updates to a non existent entity`. I think the logs would be more meaningful like this.
WDYT?

app/script/conversation/ConversationService.js
```diff
@@ -709,42 +709,72 @@ z.conversation.ConversationService = class ConversationService {
...
+
+  /**
+   * Update a message in the database.
+   *
+   * @param {Message} messageEntity - Message event to update in the database
+   * @param {Object} [changes={}] - Changes to update message with
+   * @param {string} conversationId - ID of conversation
+   * @returns {Promise} Resolves when the message was updated in database
+   */
+  updateMessageInDbByVersion(messageEntity, changes = {}, conversationId) {
```
atomrc: I would suggest adding a few test cases for that new method

atomrc: Also I am not sure about the name of the method. What would you think of something like `sequentiallyUpdateMessageInDb`?

PR #4279, opened 2018-08-07
refactor: Less verbose unit test logs

atomrc:
Before
screen shot 2018-08-08 at 09 55 58

After
screen shot 2018-08-08 at 09 56 30

### DIFF
--- a/test/api/TestFactory.js
+++ b/test/api/TestFactory.js
@@ -21,18 +21,6 @@
 
 'use strict';
 
-const proxyLogger = loggerLevel => {
-  const LoggerClass = z.util.Logger;
-  class LoggerProxy extends LoggerClass {
-    constructor(name, options) {
-      options.level = loggerLevel;
-      options.domains = {};
-      super(`fake-${name}`, options);
-    }
-  }
-  z.util.Logger = LoggerProxy;
-};
-
 /**
  * @param {function} [logger_level] - A function returning the logger level.
  * @returns {Window.TestFactory} A TestFactory instance.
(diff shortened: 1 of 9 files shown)


test/unit_tests/tracking/EventTrackingRepositorySpec.js
```diff
@@ -24,18 +24,14 @@
...
-      .exposeTrackingActors()
-      .then(() => {
-        TestFactory.tracking_repository._is_domain_allowed_for_tracking = () => true;
-        done();
-      })
-      .catch(done.fail);
+  beforeEach(() => {
+    return test_factory.exposeTrackingActors().then(() => {
+      TestFactory.tracking_repository._is_domain_allowed_for_tracking = () => true;
+    });
```
gregor, line 30: Learned something new here. Didn't know that jasmine gained proper Promise in version 2.7 support making the `done` schema obsolete.

https://jasmine.github.io/tutorials/async

atomrc: I already removed a few in some of our tests.
The cool thing is, you don't need to do `.catch(done.fail)` if you return the promise to `jasmine`.
So no more silent error when you forget that line :)

test/api/TestFactory.js
```diff
@@ -28,7 +28,7 @@
  */
 window.TestFactory = function(logger_level) {
   if (!logger_level) {
-    logger_level = z.util.Logger.prototype.levels.ERROR;
+    logger_level = z.util.Logger.prototype.levels.OFF;
```
bennycode, line 31: How verbose is the logging if the log level would still be `ERROR`? Because error is a case that we might want to have reporting.

atomrc: The thing is, when we log errors, it's that we handle the error in the code and we are probably testing the fact that we handle it.
What we want test is unexpected errors and those are reported directly by `jasmine`

PR #4294, opened 2018-08-09
feat: Get or create 1to1 conversation with service (WEBAPP-5346)

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -825,36 +825,46 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
    * Get conversation with a user.
-   * @param {User} user_et - User entity for whom to get the conversation
-   * @param {string} [team_id] - Team ID in which the conversation should be searched
+   * @param {User} userEntity - User entity for whom to get the conversation
    * @returns {Promise} Resolves with the conversation with requested user
    */
-  get_1to1_conversation(user_et, team_id = this.team().id) {
-    for (const conversation_et of this.conversations()) {
-      const with_expected_user = user_et.id === conversation_et.participating_user_ids()[0];
+  get1to1Conversation(userEntity) {
```
atomrc: We might want to name it `get1To1Conversation` in order to be consistent (have an uppercase after a number)

app/script/conversation/ConversationRepository.js
```diff
@@ -825,36 +825,46 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
+        const isActiveConversation = !conversationEntity.removed_from_conversation();
+        if (!isActiveConversation) {
+          // Disregard coversations that self is no longer part of
+          continue;
+        }
+
+        const [userId] = conversationEntity.participating_user_ids();
+        const withExpectedUser = userEntity.id === userId;
+        if (withExpectedUser) {
+          return Promise.resolve(conversationEntity);
```
atomrc: How about an implementation like this?

```
const conv = this.conversations().find(/*....*/)

return conv ? conv : this.createGroupConv()
```

app/script/integration/IntegrationRepository.js
```diff
@@ -132,6 +134,41 @@ z.integration.IntegrationRepository = class IntegrationRepository {
...
   }
 
+  /**
+   * Get conversation with a service.
+   * @param {Service} serviceEntity - Service entity for whom to get the conversation
+   * @param {string} [method] - Method used to trigger integration setup
+   * @returns {Promise} Resolves with the conversation with requested service
+   */
+  get1to1ConversationWithService(serviceEntity, method) {
+    for (const conversationEntity of this.conversationRepository.conversations()) {
```
atomrc: We could also benefit from the tenary return condition here

app/script/integration/IntegrationRepository.js
```diff
@@ -117,9 +118,10 @@ z.integration.IntegrationRepository = class IntegrationRepository {
         throw new z.conversation.ConversationError(z.conversation.ConversationError.TYPE.CONVERSATION_NOT_FOUND);
       })
       .then(conversationEntity => {
-        if (conversationEntity) {
+        if (conversationEntity && showConversation) {
```
atomrc: Quick proposal, instead of `amplifying` in a nested repository, how about the consumer of the method decides if it need to `amplify` or not?

This way the view layer decides if it need to be shown are not and we can remove view specific logic from the repository.

Something like 

```
repo.create1to1ConversationWithService(service, method).then(conv => amplify.publish(/*...*/))
```

Or (if we don't want to show the conv)


```
repo.create1to1ConversationWithService(service, method)
```

gregor: Yeah. Good point. We had in it here and I wanted to avoid writing the code again. Doing it from the consumer is much better.

PR #4302, opened 2018-08-10
refactor: Use length caching in for-loops

bennyn:


gregor:
Technical question out of curiosity: Why doesn't the linter complain? The length we cache is never reassinged. It should be `const` but is a `let`. It should trip of the linter but apparently does not. Why?

app/script/util/Crypto.js
```diff
@@ -29,7 +29,7 @@ z.util.Crypto = {
       let hash = uint32.toUint32(0);
       const key = string.toLowerCase();
 
-      for (let index = 0; index <= key.length - 1; index++) {
```
AndyLnd, line 32: ```index < key.length```

app/script/util/ArrayUtil.js
```diff
@@ -54,7 +54,7 @@ z.util.ArrayUtil = {
       return undefined;
     }
 
-    for (let index = nextIndex; index <= array.length; index++) {
```
AndyLnd, line 57: ```index < array.length```
otherwise the last will be out of bounds and therefore ```undefined```

bennycode:
Closed in favor of other array performance optimizations like:
- https://github.com/wireapp/wire-webapp/pull/4307

PR #4303, opened 2018-08-10
feat: add manage services to start ui (WEBAPP-5345)

AndyLnd:


app/script/view_model/list/StartUIViewModel.js
```diff
@@ -200,6 +202,7 @@ z.viewModel.list.StartUIViewModel = class StartUIViewModel {
       .extend({notify: 'always', rateLimit: 500});
 
     this._initSubscriptions();
+    this.clickOnManageServices = this.clickOnInviteMember;
```
gregor: This is weird. If we use the same function for both click events we should rename the function instead.

PR #4309, opened 2018-08-10
test: Don't create new array on every test run for ArrayUtil.chunk

bennyn:
## Type of change

Since `ArrayUtil.chunk` does not modify the source array, we don't need to create a new source array anymore before every spec run.


test/unit_tests/util/ArrayUtilSpec.js
```diff
@@ -23,11 +23,7 @@
 
 describe('z.util.ArrayUtil', () => {
   describe('chunk', () => {
-    let array = null;
-
-    beforeEach(() => {
-      array = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
-    });
+    const array = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
```
gregor, line 26: Should we do it in a beforeAll then? And how are tests decoupled in this case. While it does not brake right now it might in the future.

/@atomrc wdyt?

atomrc: I have no strong opinion on this. While `beforeAll` seems to be the right semantic for the thing (something we want to do before the tests), it would require a variable defined in the upperscope and a set in the `beforeAll`.

So I would say just defining a variable like @bennyn did, is the right way to go in my opinion.

If `beforeAll` could inject something into the test, I would be totally up for using that, but I don't think this is in Jasmine.

```
beforeAll(() => [1,2,3]);

it('does stuff', (testedArray) => {
  /// tests
})

PR #4325, opened 2018-08-13
feat: Preprocess services related events (WEBAPP-5353)

atomrc:
Simple mechanism to be able to plug custom event processors before the EventRepository actually processes the event

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -1144,8 +1144,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
       .then(response => {
         const event = response ? response.event : undefined;
         if (event) {
-          this.eventRepository.injectEvent(response.event, z.event.EventRepository.SOURCE.BACKEND_RESPONSE);
           this.logger.debug(`Successfully added bot to conversation '${conversationEntity.display_name()}'`, response);
+          return this.eventRepository.injectEvent(response.event, z.event.EventRepository.SOURCE.BACKEND_RESPONSE);
         }
 
         return event;
--- a/app/script/event/EventRepository.js
+++ b/app/script/event/EventRepository.js
@@ -144,20 +144,20 @@ z.event.EventRepository = class EventRepository {
     this.lastNotificationId = ko.observable();
     this.lastEventDate = ko.observable();
 
-    this.eventProcessMiddlewares = [];
+    this.eventProcessMiddleware = undefined;
 
     amplify.subscribe(z.event.WebApp.CONNECTION.ONLINE, this.recoverFromStream.bind(this));
   }
 
   /**
-   * Will set an array of middlewares to run before the EventRepository actually processes the event.
-   * Middlewares are just functions with the following signature (Event) => Event.
+   * Will set a middleware to run before the EventRepository actually processes the event.
+   * Middleware is just a function with the following signature (Event) => Promise<Event>.
    *
-   * @param {Array<Function>} middlewares - middlewares to run when a new event is about to be processed
+   * @param {Function} middleware - middleware to run when a new event is about to be processed
    * @returns {void} - returns nothing
    */
-  setEventProcessMiddlewares(middlewares) {
-    this.eventProcessMiddlewares = middlewares;
+  setEventProcessMiddleware(middleware) {
+    this.eventProcessMiddleware = middleware;
   }
 
   //##############################################################################
@@ -627,10 +627,7 @@ z.event.EventRepository = class EventRepository {
 
     return mapEvent
       .then(mappedEvent => {
-        return this.eventProcessMiddlewares.reduce(
-          (decoratedEvent, middleWare) => middleWare(decoratedEvent),
-          mappedEvent
-        );
+        return this.eventProcessMiddleware ? this.eventProcessMiddleware(mappedEvent) : mappedEvent;
       })
       .then(mappedEvent => {
         const shouldSaveEvent = z.event.EventTypeHandling.STORE.includes(mappedEvent.type);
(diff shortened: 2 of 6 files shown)


app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,59 @@
...
+
+    this.logger.info(`Preprocessing event of type ${event.type}`);
+    return this.conversationRepository.get_conversation_by_id(event.conversation).then(conversation => {
+      const hasBots = conversation.getNumberOfBots() > 0;
+      if (hasBots) {
+        return Object.assign({}, event, {
+          hasBots: true,
+        });
+      }
+      return event;
```
gregor: We can definitely turn this into a ternary. Simple enough for three lines only.

**BUT** it might not even be what we want to do. There are only a few very specific cases where we need to display the warning. It is not added to every member join message.

atomrc: Oh yeah totally, this is a work in progress, the logic to handle the events is not yet ready. 
Sorry for the misunderstanding, this was more a way to keep track of the current state of the event preprocessing archi that I was building

app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,70 @@
...
+
+  _processConversationCreationEvent(event) {
+    this.logger.info(`Preprocessing event of type ${event.type}`);
+    return this._containsBots(event.data.userIds).then(containsBots => {
+      return !containsBots ? event : Object.assign({}, event, {hasBots: true});
+    });
+  }
+
+  _containsBots(userIds) {
+    const userPromises = userIds.map(userId => this.userRepository.get_user_by_id(userId));
```
gregor: We can do this in a single request with multiple user IDs and avoid multiple backend requests if those users are new.

app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,70 @@
...
+   */
+  constructor(userRepository) {
+    this.userRepository = userRepository;
+    this.logger = new z.util.Logger('z.event.preprocessor.ServiceMiddleware', z.config.LOGGER.OPTIONS);
+  }
+
+  processEvent(event) {
+    switch (event.type) {
+      case z.event.Client.CONVERSATION.ONE2ONE_CREATION:
+      case z.event.Client.CONVERSATION.GROUP_CREATION:
```
gregor: Not for groups

app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,83 @@
...
+
+      default:
+        return Promise.resolve(event);
+    }
+  }
+
+  _processMemberJoinEvent(event) {
+    this.logger.info(`Preprocessing event of type ${event.type}`);
+
+    if (event.data.user_ids.includes(this.userRepository.self().id)) {
```
gregor: Can we pull this out a little bit so that we don't go 3 levels deep on a single call?

app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,86 @@
...
+window.z.event.preprocessor = z.event.preprocessor || {};
+
+z.event.preprocessor.ServiceMiddleware = class ServiceMiddleware {
+  /**
+   * Construct a new ServiceMiddleware.
+   *
+   * @param {z.conversation.UserRepository} userRepository - Repository to handle user related tasks
+   * @param {z.conversation.ConverationRepository} conversationRepository - Repository to handle conversation related tasks
+   */
+  constructor(userRepository, conversationRepository) {
```
gregor: We aim to pass repositories in alphabetical order

app/script/event/preprocessor/ServiceMiddleware.js
```diff
@@ -0,0 +1,86 @@
...
+    this.userRepository = userRepository;
+    this.conversationRepository = conversationRepository;
+    this.logger = new z.util.Logger('z.event.preprocessor.ServiceMiddleware', z.config.LOGGER.OPTIONS);
+  }
+
+  processEvent(event) {
+    switch (event.type) {
+      case z.event.Client.CONVERSATION.ONE2ONE_CREATION:
+        return this._process1To1ConversationCreationEvent(event);
+      case z.event.Backend.CONVERSATION.MEMBER_JOIN:
```
gregor: Can we add an empty line above

app/script/event/EventRepository.js
```diff
@@ -613,6 +626,9 @@ z.event.EventRepository = class EventRepository {
       : Promise.resolve(event);
 
     return mapEvent
+      .then(mappedEvent => {
+        return this.eventProcessMiddleware ? this.eventProcessMiddleware(mappedEvent) : mappedEvent;
+      })
```
gregor: This should fit into a single line now

app/script/conversation/ConversationRepository.js
```diff
@@ -1144,8 +1144,8 @@ z.conversation.ConversationRepository = class ConversationRepository {
       .then(response => {
         const event = response ? response.event : undefined;
         if (event) {
-          this.eventRepository.injectEvent(response.event, z.event.EventRepository.SOURCE.BACKEND_RESPONSE);
           this.logger.debug(`Successfully added bot to conversation '${conversationEntity.display_name()}'`, response);
+          return this.eventRepository.injectEvent(response.event, z.event.EventRepository.SOURCE.BACKEND_RESPONSE);
```
gregor, line 1148: Does `injectEvent` return the event JSON as well? If not we should make sure we do so here in order to not change the return value of the function.

atomrc: yes

PR #4331, opened 2018-08-14
feat: services warning in template

AndyLnd:


app/page/template/partials/template-message.htm
```diff
@@ -228,6 +228,9 @@
           </div>
         <!-- /ko -->
       </div>
+      <!-- ko if: showServicesWarning -->
+        <div class="message-header-warning" data-bind="l10n_text: z.string.conversationServicesWarning "></div>
+      <!-- /ko -->
```
gregor: This does not account for the case of 1to1 creation messages

PR #4333, opened 2018-08-14
fix: Hide member join message in fake 1to1

gregor:


app/script/conversation/EventMapper.js
```diff
@@ -293,23 +293,14 @@ z.conversation.EventMapper = class EventMapper {
...
-      const isAcceptedConnection = messageFromCreator && singleUserAdded;
-      if (isAcceptedConnection) {
-        messageEntity.memberMessageType = z.message.SystemMessageType.CONNECTION_ACCEPTED;
-        userIds = conversationEntity.participating_user_ids();
-      } else {
-        messageEntity.visible(false);
-      }
+    const isSingleModeConversation = conversationEntity.is_one2one() || conversationEntity.is_request();
+    if (isSingleModeConversation) {
+      messageEntity.visible(false);
```
atomrc: What do you think of removing the `if` and doing `messageEntity.visible(!isSingleModeConversation)`?

PR #4339, opened 2018-08-14
chore: Improve styling of conversation details panel (WEBAPP-5350)

AndyLnd:


app/script/view_model/panel/AddParticipantsViewModel.js
```diff
@@ -111,7 +111,7 @@ z.viewModel.panel.AddParticipantsViewModel = class AddParticipantsViewModel exte
   }
 
   shouldSkipTransition() {
-    return true;
+    return false;
   }
```
gregor: If we are going to the default, shouldn't we simply remove it?

app/style/panel/timed-messages.less
```diff
@@ -21,4 +21,7 @@
   &__disclaimer {
     margin: 16px;
   }
+  .panel__action-item:first-child {
```
gregor: Empty line above

PR #4340, opened 2018-08-14
chore: Conversation details service style (WEBAPP-5349)

AndyLnd:


app/style/components/participant-avatar.less
```diff
@@ -139,13 +139,6 @@ participant-avatar {
     .avatar-size(xxs);
   }
 
-  &.avatar-l,
-  &.avatar-xl {
-    .avatar-border {
-      display: none;
-    }
-  }
```
gregor, line 147: Is this change also valid for users?

AndyLnd: The border is barely visible as it is. I'd rather be consistent.

PR #4346, opened 2018-08-15
refactor: Move event deletion methods

atomrc:


app/script/event/EventService.js
```diff
@@ -134,6 +133,31 @@ z.event.EventService = class EventService {
...
+  /**
+   * Delete an event from a conversation. Duplicates are delete as well.
+   *
+   * @param {string} conversationId - ID of conversation to remove message from
+   * @param {string} eventId - ID of the actual message
+   * @returns {Promise} Resolves with the number of deleted records
+   */
+  deleteEvent(conversationId, eventId) {
+    return this.storageService.db[this.EVENT_STORE_NAME]
+      .where('id')
```
atomrc, line 188: Minor pref improvements here (selecting first using the event id and then filtering with the right conversation, before it was the other way around).

test/unit_tests/event/EventServiceSpec.js
```diff
@@ -274,6 +274,93 @@ describe('z.event.EventService', () => {
     });
   });
 
+  describe('deleteEvent', () => {
```
atomrc, line 338: These are new tests

test/unit_tests/event/EventServiceSpec.js
```diff
@@ -274,6 +274,132 @@ describe('z.event.EventService', () => {
...
+          expect(newEvents.length).toBe(2);
+        })
+        .then(() => eventService.loadPrecedingEvents('other-conversation-id'))
+        .then(otherConversationEvents => {
+          expect(otherConversationEvents.length).toBe(1);
+        });
+    });
+  });
+
+  describe('deleteEvents', () => {
```
atomrc, line 387: These are new tests

app/script/event/EventService.js
```diff
@@ -134,6 +133,45 @@ z.event.EventService = class EventService {
...
+    return this.storageService.db[this.EVENT_STORE_NAME].delete(primaryKey);
+  }
+
+  /**
+   * Delete all events of a conversation.
+   * @param {string} conversationId - Delete events for this conversation
+   * @param {string} [iso_date] - Date in ISO string format as upper bound which events should be removed
+   * @returns {Promise} Resolves when the events was deleted
+   */
+  deleteEvents(conversationId, iso_date) {
```
gregor: Let's camelCase the `isoDate` as well

app/script/event/EventService.js
```diff
@@ -134,6 +133,45 @@ z.event.EventService = class EventService {
...
+      .delete();
+  }
+
+  /**
+   * Delete an event from a conversation with the given primary.
+   *
+   * @param {string} primaryKey - ID of the actual message
+   * @returns {Promise} Resolves with the number of deleted records
+   */
+  deleteEventWithKey(primaryKey) {
```
gregor: Just a thought: Is this really deletion "with" key or rather "by" key

app/script/conversation/ConversationRepository.js
```diff
@@ -3523,7 +3523,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
    */
   _delete_message(conversation_et, message_et) {
```
gregor, line 3524: Seems like we have to build a proper EventRepository in the next step to move the event service related functions out of the ConversationRepository. Perfect start to slim that one down as well.

app/script/event/EventService.js
atomrc: Good catch

app/script/conversation/ConversationRepository.js
atomrc: I cannot agree more with that statement. That said, we already have an `EventRepository` which is... not really small :D

gregor: That one should get a different name. It's more of a `NotificationRepository` - which we also have. So maybe a `InboundBackendTrafficNotificationHandlingRepository`

PR #4347, opened 2018-08-15
chore: update translations

AndyLnd:


gregor:
@AndyLnd Please pull the translations one more time. We made an additional change to the German translations.

PR #4349, opened 2018-08-15
fix: Service panel style (WEBAPP-5349)

AndyLnd:


app/script/view_model/panel/GroupParticipantServiceViewModel.js
```diff
@@ -57,7 +57,7 @@ z.viewModel.panel.GroupParticipantServiceViewModel = class GroupParticipantServi
     });
 
     this.shouldUpdateScrollbar = ko
-      .computed(() => this.selectedService() && this.isVisible())
+      .computed(() => this.selectedService() && this.selectedService().providerName() && this.isVisible())
```
atomrc, line 60: Seems like `this.selectedService().providerName()` will always be true because the default value is not `' '` and not `''` (which evaluates to `true` in JavaScript)

atomrc: ... Well except if we set a `falsy` value at some which might be the use of this condition :)

PR #4350, opened 2018-08-15
refactor: Move the sequential update of messages

atomrc:


app/script/conversation/ConversationService.js
```diff
@@ -603,52 +603,4 @@ z.conversation.ConversationService = class ConversationService {
...
-   * Update a message in the database.
-   *
-   * @param {Message} messageEntity - Message event to update in the database
-   * @param {Object} [changes={}] - Changes to update message with
-   * @param {string} conversationId - ID of conversation
-   * @returns {Promise} Resolves when the message was updated in database
-   */
-  sequentiallyUpdateMessageInDb(messageEntity, changes = {}, conversationId) {
-    return Promise.resolve(messageEntity.primary_key).then(primaryKey => {
-      const hasChanges = !!Object.keys(changes).length;
```
atomrc, line 617: I decided to remove this check that is already handled both in the method that calls the function (the `update_reaction` method returns either an object with keys or `undefined` and the `undefined` case is handled on the consumer side) and in the test just below that checks if there is a `version` attribute.

PR #4312, opened 2018-08-10
refactor: Move event handling logic to a separate service

atomrc:
This is a pure refactoring, no behavior changes at all.

The basic idea is to start moving event related logci from the `ConversationService` to a dedicated `EventService`.

I also took care of migrating the `ConversationServiceNoCompound` to ensure compatibility with `edge`.

And I added/cleaned a few tests

Current status of the moved methods:

| ConversationService  | EventService |
|-----|----|
| `load_event_from_db` | `loadEvent` |
| `load_event_with_category_from_db` | `loadEventWithCategory` |
| `load_preceding_events_from_db` |  `loadPrecedingEvent` |
| `save_event` | `saveEvent` |
| `update_event` | `updateEvent` |
| `updateMessageInDb` |  `updateMessage` |

Other methods that could be moved in the future

- `post_encrypted_message`
- `delete_message_from_db`
- `delete_message_with_key_from_db`
- `delete_messages_from_db`
- `load_events_with_types`
- `load_subsequent_events_from_db`
- `update_asset_as_uploaded_in_db`
- `update_asset_as_failed_in_db`
- `sequentiallyUpdateMessageInDb`
- `update_message_as_upload_complete`

PR #4354, opened 2018-08-15
refactor: Move methods to load events (WEBAPP-5372)

atomrc:
continuation of #4312

### DIFF
--- a/app/script/conversation/ConversationService.js
+++ b/app/script/conversation/ConversationService.js
@@ -427,18 +427,6 @@ z.conversation.ConversationService = class ConversationService {
   }
 
   /**
-   * Load conversation events by event type.
-   * @param {Array<strings>} event_types - Array of event types to match
-   * @returns {Promise} Resolves with the retrieved records
-   */
-  load_events_with_types(event_types) {
-    return this.storageService.db[this.EVENT_STORE_NAME]
-      .where('type')
-      .anyOf(event_types)
-      .sortBy('time');
-  }
-
-  /**
    * Load conversation events starting from the upper bound to the present until the limit is reached
    *
    * @param {string} conversation_id - ID of conversation
--- a/app/script/conversation/ConversationService.js
+++ b/app/script/conversation/ConversationService.js
@@ -427,37 +427,6 @@ z.conversation.ConversationService = class ConversationService {
   }
 
   /**
-   * Load conversation events starting from the upper bound to the present until the limit is reached
-   *
-   * @param {string} conversation_id - ID of conversation
-   * @param {Date} upper_bound - Load until this date (excluded)
-   * @param {number} [limit=Number.MAX_SAFE_INTEGER] - Amount of events to load
-   * @param {boolean} [include_upper_bound=true] - Should upper bound be part of the message
-   * @returns {Promise} Resolves with the retrieved records
-   */
-  load_subsequent_events_from_db(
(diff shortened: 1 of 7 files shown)


app/script/event/EventService.js
```diff
@@ -79,32 +79,19 @@ z.event.EventService = class EventService {
...
 
-    if (lowerBound.getTime() > upperBound.getTime()) {
-      const errorMessage = `Lower bound (${lowerBound.getTime()}) cannot be greater than upper bound (${upperBound.getTime()}).`;
-      throw new Error(errorMessage);
-    }
-
-    return this.storageService.db[this.EVENT_STORE_NAME]
-      .where('[conversation+time]')
-      .between([conversationId, lowerBound.toISOString()], [conversationId, upperBound.toISOString()], true, false)
+    return this._loadEventsInDateRange(conversationId, fromDate, toDate, limit, includeParams)
```
atomrc: Ok, this is a pretty serious refactoring (I merged the logic of 2 methods into a single one, namely `_loadEventsInDateRange`), but we can be pretty confident the behavior is the same because we have a pretty good test coverage of both those methods and they are all green :+1:

app/script/conversation/ConversationService.js
```diff
@@ -426,49 +426,6 @@ z.conversation.ConversationService = class ConversationService {
     return this.storageService.getAll(this.CONVERSATION_STORE_NAME);
   }
 
-  /**
-   * Load conversation events by event type.
-   * @param {Array<strings>} event_types - Array of event types to match
-   * @returns {Promise} Resolves with the retrieved records
-   */
-  load_events_with_types(event_types) {
```
atomrc, line 434: This one is not used anymore. Last occurrence removed with this commit c53019ec

app/script/event/EventService.js
gregor: Does this also still work with no compound version of the service?

atomrc: Oh very good point, I forgot to migrate this method in the `noCompound` class!!
On it!

app/script/event/EventServiceNoCompound.js
```diff
@@ -44,40 +44,27 @@ z.event.EventServiceNoCompound = class EventServiceNoCompound extends z.event.Ev
...
     }
 
-    lowerBound = lowerBound.getTime();
-    upperBound = upperBound.getTime();
-
     return this.storageService.db[this.EVENT_STORE_NAME]
       .where('conversation')
       .equals(conversationId)
-      .reverse()
-      .sortBy('time')
```
gregor, line 74: Instead of removing sorting here we should make sure we add it on the compound version of the service instead.

gregor:
I removed the do not merge label since this fixes a bug for MS Edge users on production. Once reviewed we will make this part of the upcoming release.

app/script/event/EventService.js
```diff
@@ -92,7 +92,7 @@ z.event.EventService = class EventService {
 
     return this._loadEventsInDateRange(conversationId, fromDate, toDate, limit, includeParams)
       .reverse()
-      .toArray()
+      .sortBy('time')
       .catch(error => {
         this.logger.error(
           `Failed to load events for conversation '${conversationId}' from database: '${error.message}'`
```
gregor: Can we pull this into a const please?

PR #4369, opened 2018-08-17
fix: Accept dates in the future to load following messages

atomrc:


app/script/event/EventService.js
```diff
@@ -114,8 +114,11 @@ z.event.EventService = class EventService {
       includeFrom,
       includeTo: true,
     };
+    const toTimestamp = _.isDate(fromDate) ? Math.max(fromDate.getTime() + 1, Date.now()) : 0;
 
-    return this._loadEventsInDateRange(conversationId, fromDate, new Date(), limit, includeParams).sortBy('time');
+    return this._loadEventsInDateRange(conversationId, fromDate, new Date(toTimestamp), limit, includeParams).sortBy(
```
gregor: Please create a `const toDate` to avoid breaking this into three lines. Also we should crash here if input was not a Date.

PR #4370, opened 2018-08-17
refactor: Remove traces of the Message entity in the EventService

atomrc:
The idea is to have a little dumber `EventService`. 
To do that I removed the knowledge of the `Message` entity class from the API of the `EventService`. 

My view on this would be that `Entities` should stay on the application level (view + repositories) and low level classes should not be aware of those classes (if possible of course)

Here is the current state of the `EventService` API

screenshot_2018-08-17 jsdoc class eventservice

### DIFF
--- a/app/script/event/EventService.js
+++ b/app/script/event/EventService.js
@@ -152,7 +152,7 @@ z.event.EventService = class EventService {
   }
 
   /**
-   * Update an unencrypted conversation event.
+   * Update an unencrypted event.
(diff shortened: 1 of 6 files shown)


app/script/event/EventService.js
```diff
@@ -34,7 +35,7 @@ z.event.EventService = class EventService {
   }
 
   /**
-   * Load event from DB
+   * load event from database.
```
gregor: Uppercase

PR #4392, opened 2018-08-21
fix: Verified icon condition in conversation details (WEBAPP-5385)

AndyLnd:


app/script/components/panel/userDetails.js
```diff
@@ -59,6 +59,7 @@ ko.components.register('panel-user-details', {
   viewModel: class {
     constructor(params) {
       this.participant = params.participant;
+      this.isVerified = params.hasOwnProperty('isVerified') ? params.isVerified() : this.participant().is_verified();
```
gregor: This needs to be an observable as it might change during the existence of the component.

PR #4448, opened 2018-08-29
fix: Video calling UI fixes (WEBAPP-5405, WEBAPP-5406, WEBAPP-5407)

AndyLnd:


app/page/template/video-calling.htm
```diff
@@ -1,33 +1,31 @@
 <div id="video-calling" data-bind="hide_controls: z.viewModel.VideoCallingViewModel.CONFIG.HIDE_CONTROLS_TIMEOUT, css:{'video-calling--active': showFullscreen(), 'video-calling--minimized': multitasking.isMinimized()}" class="video-calling">
   <!-- ko if: showFullscreen() -->
-    <!-- ko if: showRemote() -->
```
gregor, line 3: Seems like this observable is no longer needed in the view model either

PR #4457, opened 2018-08-30
feat: Limit TURN server count on Firefox to 4 (WEBAPP-5415)

gregor:


app/script/calling/CallingService.js
```diff
@@ -33,11 +33,18 @@ z.calling.CallingService = class CallingService {
 
   /**
    * Retrieves a calling config from the backend.
+   *
+   * @see https://staging-nginz-https.zinfra.io/swagger-ui/tab.html#!//getCallsConfigV2
+   *
+   * @param {number} [limit] - Limit the number of TURNs servers in the resoinse (range 1, 10)
```
AndyLnd: ```response```

gregor:
Turns out Firefox not only aggressively warns you agains the use of 5 or more TURN servers but also notifies you that using more than two TURN servers will slow down route discoery.

We will run some tests looking at the duration of the ICE gathering process in order to make a decision, where to make a cut. We deem 3 TURN servers to be necessary for the different types of connections they support.

atomrc:
So after a bunch of testing here are a few results I could gather.
With the current number of servers (6), to gather the `ICE candidates`, it takes:

- for Chrome `~300-400ms` 
- for Firefox  `~800-13s` (most of the time around `13s` probably due to a timeout mechanism in FF https://bugzilla.mozilla.org/show_bug.cgi?id=1170616)

With 3 servers:

- for Chrome `~200-300ms` 
- for Firefox `~300-13s` (still the timeout occurring).

So, most of the time, on Firefox, we will run into our own timeout and send the SDP anyway even if the gathering is not over.

**BUT** what I also discovered on Firefox nothing happens between the time we have gathered the last `ICE candidate`  and the timeout FF triggers. 
What I would suggest is to improve a little bit the timeout condition and go like this.

- first timeout of `1s`
- if it times out AND there are **more** than 2 `relay` ICE candidates gathered, then we send the `SDP` anyway.
- if it times out AND there are  **less** than 3 `relay` ICE candidates gathered, then we relaunch the timeout.
- if the timeout if fired more than `5` times, we just send the SDP.

WDYT. Any idea is welcome

PR #4459, opened 2018-08-30
fix: Keep state of manual video fit, auto scale on screenshare (WEBAPP-5409)

AndyLnd:


app/script/components/groupVideoGrid.js
```diff
@@ -52,6 +52,7 @@ z.components.GroupVideoGrid = class GroupVideoGrid {
     });
 
     this.scaleVideos = this.scaleVideos.bind(this, rootElement);
+    this.doubleClickedOnVideo = this.doubleClickedOnVideo.bind(this);
```
gregor: Let's put this to the top of the constructor

app/script/components/groupVideoGrid.js
```diff
@@ -77,9 +90,12 @@ z.components.GroupVideoGrid = class GroupVideoGrid {
     });
   }
 
-  doubleClickedOnVideo(data, event) {
-    const childVideo = event.currentTarget.querySelector('video');
-    childVideo.classList.toggle(GroupVideoGrid.CONFIG.CONTAIN_CLASS);
+  doubleClickedOnVideo(_, {currentTarget}) {
```
gregor: Please keep a proper naming here. Even the linter should complain. First param is the view model in knockout.

PR #4461, opened 2018-08-30
 fix: Back arrow position in team creation flow (WEBAPP-5342)

AndyLnd:


app/script/auth/page/TeamName.js
```diff
@@ -55,13 +56,21 @@ class TeamName extends Component {
   state = {
     enteredTeamName: this.props.teamName || '',
     error: null,
+    isMobile: mediaMatcher.isMobile(),
     isValidTeamName: !!this.props.teamName,
   };
 
   componentDidMount() {
     this.props.enterTeamCreationFlow();
+    window.addEventListener('resize', this.updateIsMobile);
```
gregor, line 65: Should we debounce this listener? How often does it get called when you resize the screen?

AndyLnd: This listener is cheap, we don't really need to debounce it.

Issue #4414, opened 2018-08-26
How to start a web application


sd166:
When starting from the outside port 8888 is available and there it is only possible to authorize by phone number. Port 8088 from the outside is not available, but the application itself is running on it at the address http://IP_ADDDRES:8088/ aut and here it is possible to register authorization by phone or e-mail, etc.

help me please

bennycode:
@sd166 Can you login with your Wire account on http://localhost:8080/page/auth.html?env=prod#login? If login doesn't work on the first try (showing "Unexpected Error"), then just try again.

Please report back if that works.

sd166:
This is if you try to authorize by phone at the address http://IP_ADDR:8888/login/#login
When I try to log in, I get an error in the console:

POST http://IP_ADDR:8080/login/send 403 (Forbidden)

sd166:
If you go to the address http://IP_addr:8080/page/auth.html, it displays the following error in the browser console:

![wire_webaap](https://user-images.githubusercontent.com/10991740/45079248-b2084880-b0fa-11e8-9903-cd37cce5a7a8.jpg)

When entering the user registration:

![wireapppppp](https://user-images.githubusercontent.com/10991740/45079369-11feef00-b0fb-11e8-9252-979e1f8b00cc.jpg)

After entering the name, email, password. Errors, or what does not output and falls on the verification page where you need to enter a 6-digit code. In the server console:

```
[nginz] 87.239.27.145 - - [05/Sep/2018:11:02:07 +0300] "OPTIONS /activate/send HTTP/1.1" 204 0 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36" "-" - 6 0.000 - - - - aacff3613f4e46926eab44b7ac1258a5
[brig] D, logger=cassandra.brig, client.connect=127.0.0.1:9042#18
[nginz] 87.239.27.145 - - [05/Sep/2018:11:02:07 +0300] "POST /activate/send HTTP/1.1" 200 31 "http://IP_ADDR:8080/page/auth.html" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.106 Safari/537.36" "-" - 6 0.092 0.092 - - - d703ec19a6b112cc4966eb42f15e20e7
```
[...]

PR #4466, opened 2018-08-31
feat: Update to unicode 11 standard for inclusion of addtional emoji

maximbaz:
Hello, it's your Emoji Ambassador again 😜 

It's time to update to Unicode 11 standard and add whooping [66 new emojis](https://emojipedia.org/unicode-11.0/) to Wire 🎉 

Update procedure didn't change, let me just recap it for documentation purposes:
- Pull [emojione/eac.json](https://github.com/emojione/emojione/blob/master/extras/alpha-codes/eac.json) file, format it with prettier and save in `app/image/emoji.json`.
- Follow the instructions in `app/script/util/EmojiUtil.js` on how to update Unicode ranges in both `app/script/util/EmojiUtil.js` and `app/style/common/variables.less`.

Although I'm pretty confident in this PR, unfortunately I couldn't test this and I couldn't add some screenshots, I'm having troubles authenticating in local instance just like #4414:

* http://localhost:8888/login/?env=prod#login only accepts phone number and my account doesn't have a phone number attached.
* http://localhost:8888/auth/?env=prod#login has a bunch of errors in console and shows white screen.

I would appreciate a hint on how to login with email and password so that I could test this 😉


gregor:
@maximbaz Thanks for updating this.

Seems like our documentation on running the app locally is outdated. Since we added a react login page we have to make use of a proxy. I will update the documentation.

The page you should navigate to is http://localhost:8080/page/auth.html. Login with a public account will work with http://localhost:8080/page/auth.html?env=prod. Maybe not all redirects are working as intended. Once you did log in, you can go to http://localhost:8080/app?env=prod if the redirect does not work for the production account.

gregor:
The mentioned issue with our documentation is addressed by #4468

max-baz:
Thanks for the hint on how to run the app locally.

Seems good, ready for merge, as usual ping me if you discover some bugs in your own testing 😉 

wire-emoji11

PR #4479, opened 2018-08-31
fix: Secondary line usernames unnecessary escape

AndyLnd:


app/script/util/SanitizationUtil.js
```diff
@@ -38,10 +38,11 @@ z.util.SanitizationUtil = (() => {
 
     escapeString: string => _.escape(string),
 
-    getEscapedFirstName: (userEntity, declension) => {
-      return userEntity.is_me
-        ? _getEscapedSelfName(declension)
-        : z.util.SanitizationUtil.escapeString(userEntity.first_name());
+    getEscapedFirstName: (userEntity, declension, escaped = true) => {
```
gregor, line 41: Let's think about renaming this. If escaping is not an option the method name is confusing.

PR #4487, opened 2018-09-03
chore: Refactor modal default texts

AndyLnd:


AndyLnd:
The test suite doesn't support this syntax yet. So I will tag as ```Do not merge``` for not

gregor:
@AndyLnd Can we close this for now? I doubt the test suite will get default value in object destructuring support soon.

PR #4489, opened 2018-09-03
fix: Send the complete mediaStream state when syncing props (WEBAPP-5181)

atomrc:


test/unit_tests/calling/CallMessageBuilderSpec.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    it('forces video state if given', () => {
+      const state = {audioSend: () => true, screenSend: () => false, videoSend: () => false};
+
+      const propSyncMessageWithVideo = callMessageBuilder.createPropSync(state, {}, true);
+      expect(propSyncMessageWithVideo.properties.videosend).toEqual('true');
+
+      const propSyncMessageWithoutVideo = callMessageBuilder.createPropSync(state, {}, false);
+      expect(propSyncMessageWithoutVideo.properties.videosend).toEqual('false');
```
bennycode: Would be nice to have a test where `videoSend` is `true` in the beginning and then forced to be `false`.

app/script/calling/CallMessageBuilder.js
```diff
@@ -90,47 +90,20 @@ z.calling.CallMessageBuilder = (() => {
    * Create properties payload for call events.
    *
    * @param {Object} selfState - Current self state
-   * @param {z.media.MediaType|boolean} payloadType - Media type of property change or forced videosend state
    * @param {Object} additionalPayload - Optional additional payload to be added
+   * @param {boolean|undefined} forceVideoState - Forces the videosend property to be this value instead of the one in the selfState
```
bennycode: Optionals are defined in brackets: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

PR #4490, opened 2018-09-03
 feat: show modal if camera is not available (WEBAPP-5401)

AndyLnd:


app/script/media/MediaStreamHandler.js
```diff
@@ -663,6 +663,9 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    * @returns {undefined} No return value
    */
   _showPermissionDeniedHint(mediaType) {
+    if (mediaType === z.media.MediaType.AUDIO_VIDEO || mediaType === z.media.MediaType.VIDEO) {
+      this.mediaRepository.showNoCameraModal();
```
atomrc: Shouldn't you return here? You are going to show 2 modals in that case, is it normal?

AndyLnd: Nope, the other is just a log.

AndyLnd: Of course, it would have also shown the warning banner.

atomrc: Humm not sure I follow. The following doesn't seem to be a log, it sends a `z.event.WebApp.WARNING.MODAL` event, which let me think that it's going to show a modal

app/script/localization/webapp.js
```diff
@@ -614,6 +614,9 @@ z.string.preferencesAVMicrophone = 'Microphone';
 z.string.preferencesAVPermissionDetail = 'Enable from your browser Preferences';
 z.string.preferencesAVSpeakers = 'Speakers';
 z.string.preferencesAVTemporaryDisclaimer = 'Guests can’t start video conferences. Select the camera to use if you join one.';
+z.string.preferencesAVNoCamera = 'Wire doesn’t have access to the camera.<br><a href="https://support.wire.com/hc/articles/202935412" data-uie-name="go-no-camera-faq" target="_blank" rel="noopener noreferrer">Read this support article</a> to find out how to fix it.';
```
atomrc: It doesn't seem right to have html markup in the translations (with knowledge of the `uie-name`). I believe we have something to replace fake markups to html in the template, isn't it?

app/page/template/content/preferences-av.htm
```diff
@@ -82,13 +82,19 @@
...
             <div class="preferences-detail" data-bind="l10n_text: z.string.preferencesAVTemporaryDisclaimer"></div>
           <!-- /ko -->
-          <video class="preferences-av-video"
-                 autoplay
-                 playsinline
-                 data-bind="css: {'preferences-av-video-disabled': permissionDenied},
-                            muteMediaElement: mediaStream(),
-                            sourceStream: mediaStream()">
-          </video>
+          <!-- ko if: permissionDenied() -->
```
gregor, line 85: This should only be shown on `NotReadableError`

PR #4492, opened 2018-09-03
chore: Move react-intl-po to devDependencies

Yserz:
Moved `react-intl-po` to devDependencies. Everything else is sorting.


gregor:
@Yserz Your sorting is weird...

PR #4510, opened 2018-09-05
feat: Show "no camera access" in preview space (WEBAPP-5401, WEBAPP-5421)

AndyLnd:


app/script/components/list/conversationListCallingCell.js
```diff
@@ -88,6 +88,15 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
       return !hasOtherOngoingCalls && isInMinimizedState && (hasPreJoinVideo || isOngoingVideoCall);
     });
 
+    this.showNoCameraPreview = ko.computed(() => {
+      const permissionRepository = this.mediaRepository.streamHandler.permissionRepository;
```
gregor, line 92: It is weird to access the reposutiry like this.

app/script/components/list/conversationListCallingCell.js
```diff
@@ -88,6 +88,15 @@ z.components.ConversationListCallingCell = class ConversationListCallingCell {
...
+      return (
+        !this.showVideoPreview() &&
+        !this.isConnected() &&
+        permissionRepository.permissionState.camera() !== z.notification.PermissionState.GRANTED
+      );
```
gregor, line 97: Hard to read. Mabe we should pull the last condition into const for this to be much easier to read.

PR #4513, opened 2018-09-05
Use Personal/Pro on account/team creation screen

infotexture:
See related ticket WEBAPP-5424 & 71f6418 for color change.


bennycode:
Thanks @infotexture. We will do the color change in a separate PR.

PR #4522, opened 2018-09-06
chore: Update translations

AndyLnd:


app/i18n/webapp-el.json
```diff
@@ -2,8 +2,8 @@
   "footer.wireLink": "wire.com",
   "footer.copy": "© Wire Swiss GmbH",
   "cookiePolicyStrings.bannerText": "",
-  "index.createAccountForPersonalUse": "Για προσωπική χρήση",
-  "index.createAccountForOrganizations": "Για οργανισμούς",
+  "index.createAccountForPersonalUse": "",
```
bennycode: Let's prefer the previous translations over having no translations. This is goes for all translations where it is just `""`.

PR #4532, opened 2018-09-07
fix: Pass correct parameters to knockout (WEBAPP-5435)

AndyLnd:


app/script/view_model/bindings/CommonBindings.js
```diff
@@ -532,7 +532,7 @@ ko.bindingHandlers.clickOrDrag = {
   init(element, valueAccessor, allBindings, viewModel, bindingContext) {
     const isMacDesktop = z.util.Environment.electron && z.util.Environment.os.mac;
     const context = bindingContext.$data;
-    const callback = valueAccessor().bind(context);
+    const callback = valueAccessor().bind(context, context);
```
gregor, line 535: @AndyLnd Binding the same value twice? Shouldn't we rename one of them?

PR #4534, opened 2018-09-07
refactor: Simplify the sdp sending timeout protocol

atomrc:
Previous behavior:

- waiting 5 seconds for ICE candidate gathering
- after 5 seconds, if the SDP contains relays ICE candidate we send the SDP else we wait 1 second and replay this timeout as long as we don't have relay candidates

New behavior:

- we only wait 1 second for the ICE candidate gathering (and loop until we have a bunch of relays candidate in the SDP).

This PR also adds a log with the types of candidates we have gathered

### DIFF
--- a/app/script/calling/entities/FlowEntity.js
+++ b/app/script/calling/entities/FlowEntity.js
@@ -27,6 +27,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
   static get CONFIG() {
     return {
       DATA_CHANNEL_LABEL: 'calling-3.0',
+      MAX_ICE_CANDIDATE_GATHERING_ATTEMPTS: 5,
       NEGOTIATION_THRESHOLD: 0.5 * z.util.TimeUtil.UNITS_IN_MILLIS.SECOND,
       RECONNECTION_TIMEOUT: 2.5 * z.util.TimeUtil.UNITS_IN_MILLIS.SECOND,
       RENEGOTIATION_TIMEOUT: 30 * z.util.TimeUtil.UNITS_IN_MILLIS.SECOND,
@@ -85,6 +86,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
     //##############################################################################
 
     this.peerConnection = undefined;
+    this.iceCandidatesGatheringAttempts = 0;
     this.pcInitialized = ko.observable(false);
 
     this.mediaStream = this.callEntity.localMediaStream;
@@ -499,6 +501,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
   _createPeerConnection() {
     return this._createPeerConnectionConfiguration().then(pcConfiguration => {
       this.peerConnection = new window.RTCPeerConnection(pcConfiguration);
+      this.iceCandidatesGatheringAttempts = 0;
       this.telemetry.time_step(z.telemetry.calling.CallSetupSteps.PEER_CONNECTION_CREATED);
       this.signalingState(this.peerConnection.signalingState);
 
@@ -860,8 +863,11 @@ z.calling.entities.FlowEntity = class FlowEntity {
             connectionConfig,
             iceCandidates
           );
-          if (!isValidGathering) {
-            const logMessage = `Not enough ICE candidates gathered. Restarting timeout\n${iceCandidates}`;
+          const attempts = this.iceCandidatesGatheringAttempts;
+          const hasReachMaxGatheringAttempts = attempts >= FlowEntity.CONFIG.MAX_ICE_CANDIDATE_GATHERING_ATTEMPTS;
+          if (!hasReachMaxGatheringAttempts && !isValidGathering) {
+            const logMessage = `Not enough ICE candidates gathered (attempt '${attempts}'). Restarting timeout\n${iceCandidates}`;
+            this.iceCandidatesGatheringAttempts++;
             this.callLogger.warn(logMessage);
             return this._setSendSdpTimeout();
           }
--- a/app/script/calling/entities/FlowEntity.js
+++ b/app/script/calling/entities/FlowEntity.js
@@ -867,20 +867,10 @@ z.calling.entities.FlowEntity = class FlowEntity {
           }
         }
 
-        const iceCandidateTypes = iceCandidates.reduce((types, candidateStr) => {
-          const typeMatches = candidateStr.match(/typ (\w+)/);
-          if (!typeMatches) {
-            return types;
-          }
-          const candidateType = typeMatches[1];
-          types[candidateType] = types[candidateType] ? types[candidateType] + 1 : 1;
-          return types;
-        }, {});
+        const iceCandidateTypes = z.util.PeerConnectionUtil.getIceCandidatesTypes(iceCandidates);
 
         const iceCandidateTypesLog = Object.keys(iceCandidateTypes)
-          .map(candidateType => {
-            return `${iceCandidateTypes[candidateType]} ${candidateType}`;
-          })
+          .map(candidateType => `${iceCandidateTypes[candidateType]} ${candidateType}`)
           .join(', ');
 
         const logMessage = {
@@ -899,7 +889,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
               this.callLogger.obfuscateSdp(this.localSdp().sdp),
             ],
           },
-          message: `Sending local '{0}' SDP containing '{1}' ICE candidates ('{2}') for flow with '{3}'\n{4}`,
+          message: `Sending local '{0}' SDP containing '{1}' ICE candidates ({2}) for flow with '{3}'\n{4}`,
         };
         this.callLogger.debug(logMessage);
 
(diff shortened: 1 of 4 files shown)


gregor:
@atomrc The issue with this is that sending the SDP after a 5 second timeout if gathering did not complete is a behavior that is agreed upon and synched with the AVS team.

// @z-dule @c-g-owen

atomrc:
I am actually not sure what is the link with AVS?

To me, as long as the SDP contains some relay servers, AVS should be happy, am I wrong?
On chrome for example it almost never take more than 500ms, and still the connection works.
What is the difference here?

gregor:
The link with AVS is that we want to align the behavior as much as possible. The whole calling protocol is timeout driven and clients are supposed to behave the same.

You are not wrong in your assumption that the call "should" work. Firefox is just really slow in gathering the ICE candidate routes. It "should" be as fast as Chrome. But it is not and gathers more routes. There is no point in having 15 routes in the SDP. That's why they warn against the use of more than 2 TURN serves and strictly advice against the use of 5 or more TURN servers.

The slowness is exactly the reason why we have the 5 second timeout BUT still check that there is a TURN based candidate in the SDP at that point. If we want to change the behavior we will need to align in with AVS in order to not diverge the protocol and client behavior down the line.

atomrc:
Also, I ran a bunch of tests last week and I discovered that everytime Firefox finds all the candidates very quick (actually as quickly as Chrome or just a bit slower), but just take a looooottt of time confirming that info (sending the `null` iceCandidate).

the timeline looks like this :

- couple of `ms`
- found candidate 1
- couple of `ms`
- found candidate 2
- couple of `ms`
- found candidate 3
- [...]
- found candidate > 10  (from that point something like 500ms has passed)
- **nothing happens for 13 seconds**
- send the `null` candidate

So two things:
- in all the cases I tried I always had all the candidates in less than 1 second;
- if no relays are there we still have a safety that relaunchs the timeout (we can even be more secure on this and check that we have more than, say, 3 relays ...)

gregor:
@thomas But you were always on a decent network. The gathering will and can take much longer in other network conditions. Just tether your computer with a mobile device or similar. We need an approach that suits all environments. That's were the 5 second timeout from AVS on mobile stems from.

I fully agree that we can and should decrease it but we have to be clever on checking that the SDP contains not just any TURN candidate but the ones we want it to contain we do not have a subsequent ICE candidate trickling in place.

atomrc:
Bug submitted to Mozilla for the 13 seconds timeout https://bugzilla.mozilla.org/show_bug.cgi?id=1490672

app/script/calling/entities/FlowEntity.js
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
 
         const isModeDefault = this.negotiationMode() === z.calling.enum.SDP_NEGOTIATION_MODE.DEFAULT;
         if (isModeDefault && sendingOnTimeout) {
-          if (!this._containsRelayCandidate(iceCandidates)) {
-            const logMessage = `No relay ICE candidates in local SDP. Timeout reset\n${iceCandidates}`;
-            this.callLogger.warn(logMessage, iceCandidates);
-            return this._setSendSdpTimeout(false);
+          const MIN_NUMBER_OF_RELAY_SERVERS = 2;
```
gregor: Let's also move this to the config at the topic at the file

app/script/calling/entities/FlowEntity.js
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+            return this._setSendSdpTimeout();
           }
         }
 
+        const iceCandidateTypes = iceCandidates.reduce((types, candidateStr) => {
+          const typeMatches = candidateStr.match(/typ (\w+)/);
+          if (!typeMatches) {
+            return types;
+          }
+          const candidateType = typeMatches[1];
```
gregor: `const [, candidateType] = typeMatches;`

app/script/calling/entities/FlowEntity.js
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+          types[candidateType] = types[candidateType] ? types[candidateType] + 1 : 1;
```
gregor: No need for a ternary:

`types[candidateType] = types[candidateType] + 1 || 1;` should give you the same result as `underfined + 1` is `NaN` which is not truthy.

app/script/calling/entities/FlowEntity.js
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+          return types;
+        }, {});
+
+        const iceCandidateTypesLog = Object.keys(iceCandidateTypes)
+          .map(candidateType => {
+            return `${iceCandidateTypes[candidateType]} ${candidateType}`;
+          })
```
gregor: Single line?

````
.map(candidateType => `${iceCandidateTypes[candidateType]} ${candidateType}`)
````

File: app/script/calling/entities/FlowEntity.js [hunk 5]
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+            ],
             obfuscated: [
               localSdp.type,
               iceCandidates.length,
               this.callLogger.obfuscate(this.remoteUser.id),
               this.callLogger.obfuscateSdp(this.localSdp().sdp),
             ],
           },
-          message: `Sending local '{0}' SDP containing '{1}' ICE candidates for flow with '{2}'\n{3}`,
+          message: `Sending local '{0}' SDP containing '{1}' ICE candidates ('{2}') for flow with '{3}'\n{4}`,
```
gregor: Let's get rid of the `'` from within `('{2}')` as it would be a double indication of the expected variable content.

app/script/calling/entities/FlowEntity.js
```diff
@@ -929,18 +951,17 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
-    for (const iceCandidate of iceCandidates) {
-      if (iceCandidate.toLowerCase().includes('relay')) {
-        return true;
-      }
-    }
+  _getNumberOfRelayCandidates(iceCandidates) {
+    return iceCandidates.reduce((count, iceCandidate) => {
+      const isRelay = iceCandidate.toLowerCase().includes('relay');
+      return isRelay ? count + 1 : count;
+    }, 0);
```
gregor: Do we really need a reduce here or does `Array.filter().length` do the trick as well?

````
return iceCandidates.filter(iceCandidate => iceCandidate.toLowerCase().includes('relay')).length;
````

File: app/script/calling/entities/FlowEntity.js [hunk 7]
```diff
@@ -856,24 +855,47 @@ z.calling.entities.FlowEntity = class FlowEntity {
 
         const isModeDefault = this.negotiationMode() === z.calling.enum.SDP_NEGOTIATION_MODE.DEFAULT;
         if (isModeDefault && sendingOnTimeout) {
-          if (!this._containsRelayCandidate(iceCandidates)) {
-            const logMessage = `No relay ICE candidates in local SDP. Timeout reset\n${iceCandidates}`;
-            this.callLogger.warn(logMessage, iceCandidates);
-            return this._setSendSdpTimeout(false);
+          const MIN_NUMBER_OF_RELAY_SERVERS = 2;
+          if (this._getNumberOfRelayCandidates(iceCandidates) < MIN_NUMBER_OF_RELAY_SERVERS) {
```
gregor: Let's also pull this into a const

app/script/calling/entities/FlowEntity.js
atomrc: Very true

app/script/calling/entities/FlowEntity.js
atomrc: Ok works, but feels a little cryptic...

app/script/calling/entities/FlowEntity.js
```diff
@@ -500,6 +501,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
   _createPeerConnection() {
     return this._createPeerConnectionConfiguration().then(pcConfiguration => {
       this.peerConnection = new window.RTCPeerConnection(pcConfiguration);
+      this.iceCandidatesGatheringAttempts = 0;
```
gregor: Let's reset this to 0 in the reset method and put a default in the constructor.

app/script/calling/entities/FlowEntity.js
```diff
@@ -856,24 +858,44 @@ z.calling.entities.FlowEntity = class FlowEntity {
...
+          const connectionConfig = this.peerConnection.getConfiguration();
+          const isValidGathering = z.util.PeerConnectionUtil.isValidIceCandidatesGathering(
+            connectionConfig,
+            iceCandidates
+          );
+          const attempts = this.iceCandidatesGatheringAttempts;
+          const hasReachMaxGatheringAttempts = attempts >= FlowEntity.CONFIG.MAX_ICE_CANDIDATE_GATHERING_ATTEMPTS;
+          if (!hasReachMaxGatheringAttempts && !isValidGathering) {
+            const logMessage = `Not enough ICE candidates gathered (attempt '${attempts}'). Restarting timeout\n${iceCandidates}`;
+            this.iceCandidatesGatheringAttempts++;
```
gregor, line 869: Now it would send it after 6 seconds and not 5

PR #4535, opened 2018-09-07
chore: Update translations

Yserz:


app/script/localization/translations/webapp-de.js
```diff
@@ -97,8 +97,9 @@ z.string.de.callAccept = 'Annehmen';
 z.string.de.callJoin = 'Beitreten';
 z.string.de.callChooseSharedScreen = 'Wähle einen Bildschirm aus';
 z.string.de.callParticipants = '{{number}} im Anruf';
+z.string.de.callNoCameraAccess = 'Kein Kamerazugriff';
 
-z.string.de.videoCallOverlayFitVideoLabel = 'Füllmodus der Videos kann per Doppelklick geändert werden';
+z.string.de.videoCallOverlayFitVideoLabel = 'Das Video per Doppelklick anpassen';
```
gregor, line 102: @Astrid-Wire @infotexture This seems to be a weird German copy. This is supposed to be a hint and the chosen copy is not really clear to me. Now it looks like a somewhat random comment that you see when you move your mouse that is not connected to any UI need or user action,

app/script/localization/translations/webapp-de.js
```diff
@@ -274,6 +275,9 @@ z.string.de.modalUserUnblockAction = 'Freigeben';
 z.string.de.modalUserUnblockHeadline = 'Freigeben?';
 z.string.de.modalUserUnblockMessage = '{{user}} wird dich wieder kontaktieren und zu Gruppen hinzufügen können.';
 
+z.string.de.modalNoCameraTitle = 'Kein Kamerazugriff';
+z.string.de.modalNoCameraMessage = 'Wire hat keinen Zugriff auf die Kamera.[br][faqLink]Zur Behebung bitte diesen Support Artikel anschauen[/faqLink].';
+
```
gregor, line 280: @infotexture Also somehwat weird in my opinion.

Astrid-Wire: Gregor, thanks for the feedback. As discussed on Friday in German localization chat: It is not ideal,  I will look into it.

PR #4540, opened 2018-09-10
fix: Update mediastream when changing audio/video preferences (WEBAPP-5416)

atomrc:


app/script/media/MediaStreamHandler.js
```diff
@@ -268,11 +268,6 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
     return constraintsPromise
       .then(streamConstraints => this.requestMediaStream(mediaType, streamConstraints))
       .then(mediaStreamInfo => {
-        if (!this.mediaStreamInUse()) {
```
atomrc, line 271: @gregor I cannot really find a reason for that condition. Maybe you have insights regarding this.
Seems like it was introduced here https://github.com/wireapp/wire-webapp/pull/3938

This prevents the preferences page to update the stream shown when switching inputs

gregor: The condition for this is switching the stream and then ending the call where the stream is returned after the call was ended. In such cases the new stream would never be released and the camera access kept indefinitely. The right approach is to consider the preferences as a user of a `MediaStream` and ensure that `mediaStreamInUse` will return true if you have a joined call, an incoming video call, or the a/v preferences.

app/script/media/MediaStreamHandler.js
```diff
@@ -241,7 +241,7 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
    * @returns {Promise} Resolves when the input source has been replaced
    */
   replaceInputSource(mediaType) {
-    const isPreferenceChange = !this.mediaStreamInUse();
+    const isPreferenceChange = this.currentCalls.size === 0;
```
gregor, line 244: Where is `size` coming from? Never seen it in the knockout documentation

app/script/media/MediaStreamHandler.js
```diff
@@ -268,7 +268,8 @@ z.media.MediaStreamHandler = class MediaStreamHandler {
     return constraintsPromise
       .then(streamConstraints => this.requestMediaStream(mediaType, streamConstraints))
       .then(mediaStreamInfo => {
-        if (!this.mediaStreamInUse()) {
+        if (!isPreferenceChange && !this.mediaStreamInUse()) {
```
gregor: I still believe we just adjust this condition as we would fix a bunch of additional scenarios. Currently the stream will be erroneously released if you are in a video call, open the preferences, and then end the call from the remote side.

app/script/media/MediaStreamHandler.js
atomrc: That's because it's a `Map` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Properties_2

gregor:
Let's go with this fix but clearly mark it as a temporary workaround until we can dedicate the time to address the `mediaSteamInUse` capabilities in a better way.

PR #4541, opened 2018-09-10
chore: Update url-search-params-polyfill to version 5.0.0

gregor:


gregor:
Breaking as in something that used to not work now works....

PR #4551, opened 2018-09-11
fix: webapp landing page on wrapper

atomrc:
The problem came from the upgrade to `babel7`. 

The problem is, the `clearImmediate` polyfill is not included anymore in the build by Babel  (but the `setImmediate` is ..??!!).


atomrc:
Closed as we are on the verge of finding a cleaner solution

PR #4554, opened 2018-09-11
test: Bootstrap test cases for the SDPMapper class

atomrc:


app/script/calling/SDPMapper.js
```diff
@@ -124,6 +124,6 @@ z.calling.SDPMapper = {
 
     sessionDescription = sdpLines.join('\r\n');
     const sdpInit = {sdp: sessionDescription, type};
-    return Promise.resolve({iceCandidates, sdp: sdpInit});
```
atomrc, line 127: Noticeable change, I decide to remove the promise wrapping of the result.
Since it is not needed here (nothing async), it feels more natural and the api is easier to use

test/unit_tests/calling/SDPMapperSpec.js
```diff
@@ -0,0 +1,193 @@
...
+      const rtcSdp = {
+        sdp: sdpStr,
+        type: z.calling.rtc.SDP_TYPE.OFFER,
+      };
+
+      const flowEntity = {
+        negotiationMode: () => '',
+      };
+
+      // local SDP
```
bennycode: I think we can remove that comment because `{sdp: localSdp}` make already clear that it is the local SDP. Same applies to the `remoteSdp`.

gregor:
LGTM

PR #4560, opened 2018-09-12
feat: Parse mentions from message input (WEBAPP-5443)

AndyLnd:
This is just the method for parsing the mentions from the input. Nothing happens with that for now.

### DIFF
--- a/app/script/view_model/content/InputBarViewModel.js
+++ b/app/script/view_model/content/InputBarViewModel.js
@@ -360,6 +360,26 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     }
   }
 
+  parseForMentions(messageText) {
+    const mentionRegexp = /\B@(.+?)\b/g;
+    const mentions = [];
+    const users = this.conversationEntity().participating_user_ets();
+
+    let find;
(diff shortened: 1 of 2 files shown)


atomrc:
Ah no, it's not a pure function, but definitely something we want to have tests for

AndyLnd:
I will make it pure and add tests.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -356,10 +356,31 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
   sendMessage(messageText) {
     if (messageText.length) {
+      const mentions = this.parseForMentions(messageText, this.conversationEntity().participating_user_ets());
+      this.logger.info(`Found ${mentions.length} mentions in outgoing message.`);
       this.conversationRepository.send_text_with_link_preview(messageText, this.conversationEntity());
     }
   }
 
+  parseForMentions(messageText, userEntities) {
+    const mentionRegexp = /\B@(.+?)\b/g;
```
atomrc, line 366: Are you sure about the `\B`. Shouldn't it be `\b` (word boundary)?

gregor:
LGTM

PR #4568, opened 2018-09-13
fix: Cancel call when user decide not to call because of a degraded state (WEBAPP-5449)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -1849,7 +1849,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
           throw error;
         }
 
-        amplify.publish(z.event.WebApp.CALL.STATE.DELETE, callMessageEntity.id);
+        amplify.publish(z.event.WebApp.CALL.STATE.DELETE, callMessageEntity.conversationId);
```
atomrc, line 1852: On a side note, this should not be handled by an event. 
The promise chain passes by the `CallingRepo` so this guys is totally able to handle degradation error and end the call on his side.

gregor: Agreed. All the calling state changes are historically event based though.

PR #4570, opened 2018-09-13
chore: Use Protocol Buffers that support mentions

bennyn:
This PR brings Protocol Buffers that prevent clients from crashing when receiving messages with mentions. Please merge it to "dev" first because @marcoconti83 has ideas how to make it work for "dev" even without running with the latest Protocol Buffers specification.


gregor:
Superseded by #4579

PR #4579, opened 2018-09-14
 feat: Add mentions to outgoing text messages (WEBAPP-5445)

gregor:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -95,17 +95,22 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+
+      const pieces = this.parseForMentions(input, participatingUserEntities)
+        .reverse()
+        .reduce(
+          (currentPieces, mentionEntity) => {
+            const {end: mentionEnd, start: mentionStart} = mentionEntity;
+
+            const currentPiece = currentPieces.shift();
+            currentPieces.unshift(currentPiece.substr(mentionEnd));
+            currentPieces.unshift(currentPiece.substr(mentionStart, mentionEnd - mentionStart));
```
AndyLnd: Now that we have an entity for the mentions, we could put ```static get length() {return this.end-this.start}``` in there. WDYT?

gregor: Sure. Good idea. Will add that.

PR #4580, opened 2018-09-14
feat: Supports mentions on incoming text messages (WEBAPP-5456)

gregor:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -101,12 +101,10 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
         .reduce(
           (currentPieces, mentionEntity) => {
-            const {end: mentionEnd, start: mentionStart} = mentionEntity;
-
             const currentPiece = currentPieces.shift();
-            currentPieces.unshift(currentPiece.substr(mentionEnd));
-            currentPieces.unshift(currentPiece.substr(mentionStart, mentionEntity.getLength()));
-            currentPieces.unshift(currentPiece.substr(0, mentionStart));
+            currentPieces.unshift(currentPiece.substr(mentionEntity.getEndIndex()));
+            currentPieces.unshift(currentPiece.substr(mentionEntity.startIndex, mentionEntity.getEndIndex()));
```
AndyLnd: should be ```mentionEntity.length``` instead of ```mentionEntity.getEndIndex()```

AndyLnd: Or we use ```substring(start, end)``` instead, which is the favorable method arcording to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr

app/script/message/MentionEntity.js
```diff
@@ -22,31 +22,57 @@
...
 
-    this.userId = userId;
+    this.userId = mention.user_id || '';
+  }
+
+  get endIndex() {
+    return this.startIndex + this.length - 1;
+  }
+
+  isValid(messageText = '') {
```
atomrc: Some tests for this method would be nice (non-blocking, could be done in another PR)

gregor: Fair point. I will try to look into that later today. Would want to merge this now though so that @AndyLnd can continue his work on the rich input bar.

atomrc: :+1:

app/script/message/MentionEntity.js
```diff
@@ -22,31 +22,57 @@
...
+    this.startIndex = mention.start || 0;
+    this.length = mention.length || 1;
+    this.type = mention.mention_type || z.cryptography.PROTO_MESSAGE_TYPE.MENTION_TYPE_USER_ID;
 
-    this.userId = userId;
+    this.userId = mention.user_id || '';
+  }
+
+  get endIndex() {
+    return this.startIndex + this.length - 1;
```
AndyLnd: Let's lose the `- 1` so it will be consistent with text indices in JS.

app/script/message/MentionEntity.js
```diff
@@ -22,31 +22,57 @@
...
+  }
+
+  isValid(messageText = '') {
+    if (!typeof this.startIndex === 'number' || !typeof this.length === 'number' || !typeof this.userId === 'string') {
+      return false;
+    }
+
+    const isValidStartIndex = this.startIndex >= 0;
+    const isValidLength = this.length >= 1;
+    const isValidEnd = messageText.length && this.endIndex <= messageText.length - 1;
```
AndyLnd: after the above change `this.endIndex <= messageText.length`

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -101,12 +101,10 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
         .reverse()
         .reduce(
           (currentPieces, mentionEntity) => {
-            const {end: mentionEnd, start: mentionStart} = mentionEntity;
-
             const currentPiece = currentPieces.shift();
-            currentPieces.unshift(currentPiece.substr(mentionEnd));
-            currentPieces.unshift(currentPiece.substr(mentionStart, mentionEntity.getLength()));
-            currentPieces.unshift(currentPiece.substr(0, mentionStart));
+            currentPieces.unshift(currentPiece.substr(mentionEntity.endIndex + 1));
```
AndyLnd: After the change to `MentionEntity` => `currentPiece.substr(mentionEntity.endIndex)`

PR #4586, opened 2018-09-17
feat: Search algorithm for mentions

atomrc:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -93,8 +94,19 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       },
     });
 
+    this.mentionSuggestions = ko.pureComputed(() => {
+      const inputValue = this.input();
+      const incompleteMention = inputValue.match(/@(\w+)$/);
```
atomrc, line 101: changes in this file are basically test code (will be overridden in the future). No need to be too picky on this one

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
...
+        return weightValue + propertyImportance;
+      }, 0);
+
+      // add a weight bonus for properties that matched from the start
+      const positionBonus = matchedProperties.filter(property => matches(userEntity, property, true)).length * 100;
+      return results.concat({user: userEntity, weight: weight + positionBonus});
+    }, []);
+
+    return weightedResults
+      .slice(0)
```
AndyLnd: No need for a parameter.

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
...
+   *    the order of the properties in the array indicates the priorities by which results will be sorted
+   * @returns {Array<z.entity.User>} the filtered list of users
+   */
+  searchUserInSet(term, userEntities, properties = ['first_name', 'last_name', 'username']) {
+    const excludedEmojis = Array.from(term).filter(char => {
+      return z.util.EmojiUtil.UNICODE_RANGES.includes(char);
+    });
+
+    const matches = (userEntity, property, fromStart) => {
+      const value = userEntity[property]() || '';
```
AndyLnd: You might be able to use ko.unwrap() here.

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
...
+
+    const weightedResults = userEntities.reduce((results, userEntity) => {
+      const matchedProperties = properties.filter(property => matches(userEntity, property, false));
+
+      if (!matchedProperties.length) {
+        return results;
+      }
+
+      // add weight to the result based on the properties that matched and the position of the property in the property list
+      const weight = matchedProperties.reduce((weightValue, property, index) => {
```
AndyLnd: index is not used

app/script/util/StringUtil.js
```diff
@@ -25,9 +25,11 @@ window.z.util = z.util || {};
 z.util.StringUtil = {
   capitalizeFirstChar: (string = '') => `${string.charAt(0).toUpperCase()}${string.substring(1)}`,
 
-  compareTransliteration: (nameA, nameB, excludedChars = []) => {
+  compareTransliteration: (name, query, excludedChars = [], fromStart = false) => {
```
gregor: If we clean this up then we should change the naming properly. `name` does not have to be a name. It is part of the `StringUtil` and can be any random input string.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -95,8 +96,19 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+      const candidates = this.conversationEntity().participating_user_ets();
+      if (incompleteMention) {
+        return this.searchRepository.searchUserInSet(incompleteMention[1], candidates);
+      }
+      return [];
```
gregor, line 106: Easily turn this into a ternary

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
...
+        const propertyImportance = properties.length - properties.indexOf(property);
```
AndyLnd: ```const propertyImportance = properties.reverse().indexOf(property);```
might be more readable and gives you a lowest value of 0 instead of 1.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -112,7 +124,10 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
           [input]
         );
       const mentionAttrs = ' class="input-mention" data-uie-name="item-input-mention"';
-      return pieces.map((piece, index) => `<span${index % 2 ? mentionAttrs : ''}>${piece}</span>`).join('');
+      return (
+        pieces.map((piece, index) => `<span${index % 2 ? mentionAttrs : ''}>${piece}</span>`).join('') +
+        suggestions.map(user => user.username()).join(', ')
+      );
```
gregor, line 130: Let's turn this into something more readable

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
     this.logger = new z.util.Logger('z.search.SearchRepository', z.config.LOGGER.OPTIONS);
   }
 
+  /**
+   * Search for a user in the given user list and given a search term
+   * Doesn't sort the results and keep the initial order of the given user list
+   * @param {string} term - the search term
```
gregor: Let's clean up the JSDoc a bit.

- end name with .
- add empty like below name
- prefix info

app/script/search/SearchRepository.js
```diff
@@ -56,6 +56,49 @@ z.search.SearchRepository = class SearchRepository {
...
+   * @param {Array<z.entity.User>} userEntities - entities to match the search term against
+   * @param {Array<string>} properties=['username', 'first_name', 'last_name'] - list of properties that will be matched against the search term
+   *    the order of the properties in the array indicates the priorities by which results will be sorted
+   * @returns {Array<z.entity.User>} the filtered list of users
+   */
+  searchUserInSet(term, userEntities, properties = ['first_name', 'last_name', 'username']) {
```
gregor: Let's create a type for the possible properties instead of using the strings.

app/script/util/StringUtil.js
atomrc: true!

app/script/view_model/content/InputBarViewModel.js
AndyLnd: This is basically just interims code. The names will be rendered completely different. So it is okay to leave it like this for now.

app/script/view_model/content/InputBarViewModel.js
gregor: Still valid

app/script/search/SearchRepository.js
```diff
@@ -56,6 +61,52 @@ z.search.SearchRepository = class SearchRepository {
...
+
+      if (!matchedProperties.length) {
+        return results;
+      }
+
+      // add weight to the result based on the properties that matched and the position of the property in the property list
+      const weight = matchedProperties.reduce((weightValue, property) => {
+        const propertyImportance = properties.reverse().indexOf(property);
+        return weightValue + propertyImportance;
+      }, 0);
```
gregor, line 95: I think the reversing should not be done in the set. This is part of the visualization logic that we won't need for other applications of this search capability.

app/script/search/SearchRepository.js
```diff
@@ -56,6 +61,52 @@ z.search.SearchRepository = class SearchRepository {
...
+   * @param {Array<z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS>} properties=['first_name', 'last_name', 'username'] - list of properties that will be matched against the search term
+   *    the order of the properties in the array indicates the priorities by which results will be sorted
+   * @returns {Array<z.entity.User>} the filtered list of users
+   */
+  searchUserInSet(term, userEntities, properties) {
+    const SEARCHABLE_FIELDS = z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS;
+    properties = properties || [SEARCHABLE_FIELDS.FIRST_NAME, SEARCHABLE_FIELDS.LAST_NAME, SEARCHABLE_FIELDS.USERNAME];
+    const excludedEmojis = Array.from(term).filter(char => {
+      return z.util.EmojiUtil.UNICODE_RANGES.includes(char);
+    });
```
gregor: Fits a single line

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -95,8 +96,19 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       },
     });
 
+    this.mentionSuggestions = ko.pureComputed(() => {
+      const inputValue = this.input();
+      const incompleteMention = inputValue.match(/@(\w+)$/);
+      const candidates = this.conversationEntity().participating_user_ets();
+      if (incompleteMention) {
+        return this.searchRepository.searchUserInSet(incompleteMention[1], candidates);
```
gregor, line 104: Let's name what you are pushing in here. => `const [, searchTerm] = incompleteMention`

atomrc: Again, this is just placeholder code, I don't think we should spend to much time on this as this is going to be replaced in the near future

app/script/search/SearchRepository.js
atomrc: Not sure I get your point. The point of the reverse here is to get the importance of the field that matched in order to assign a weight to a search result. This has nothing to do with visualization (well we do sort in the end, that is true)

PR #4588, opened 2018-09-17
feat: Use the local user search algorithm everywhere (WEBAPP-5475)

atomrc:
Main changes:
- we can match handles with patterns that are in between the handle (`ichel` matches `michel`)
- the results are now sorted with the following priorities
  - first name matches from the start
  - last name matches from the start
  - first name matches in between
  - last name matches in between

Here are a few examples that give different results

### `@test_benny` matches because `benny` is somewhere in the handle

screenshot from 2018-09-17 16-30-13

### `Node` is before `Dusan Stevanovic` because the first name has the priority

screenshot from 2018-09-17 16-41-06

### DIFF
--- a/app/script/view_model/list/StartUIViewModel.js
+++ b/app/script/view_model/list/StartUIViewModel.js
@@ -568,7 +568,7 @@ z.viewModel.list.StartUIViewModel = class StartUIViewModel {
       const localSearchFields = isHandle ? [SEARCHABLE_FIELDS.USERNAME] : undefined;
 
       const contactSearchResults = this.searchRepository.searchUserInSet(
-        trimmedQuery.replace(/^@/, ''),
+        normalizedQuery,
         localSearchSources,
         localSearchFields
       );
--- a/app/script/view_model/list/StartUIViewModel.js
+++ b/app/script/view_model/list/StartUIViewModel.js
@@ -565,7 +565,7 @@ z.viewModel.list.StartUIViewModel = class StartUIViewModel {
         : this.userRepository.connected_users();
 
       const SEARCHABLE_FIELDS = z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS;
-      const localSearchFields = trimmedQuery.startsWith('@') ? [SEARCHABLE_FIELDS.USERNAME] : [SEARCHABLE_FIELDS.NAME];
+      const localSearchFields = isHandle ? [SEARCHABLE_FIELDS.USERNAME] : undefined;
 
       const contactSearchResults = this.searchRepository.searchUserInSet(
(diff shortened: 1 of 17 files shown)


gregor:
Why would `Gregor @NoClients` be prioritized over `Dusan Stevanovic` though? They are both an includes match on the last name and should thus simply be alphabetically ordered.

atomrc:
@gregor humm good point. This is because we generate transliterations to match strings, and the `@` of `@NoClients` gets removed, thus the matching from the start is positive, giving it a higher rank... Do you think it's a problem?

gregor:
Yes.

atomrc:
Any suggestions to fix this? 

I might have an idea, maybe we can add a new step in the matching algo:
- match raw string (with transliteration), if it matches, it then gives a boost
- match transliterated strings  (which has a lower impact on the score).

IMO it might be a little overkill to take this scenario into account but you might have some other insights.
WDYT?

gregor:
It is not overkill. Expecting something to be sorted alphabetically is bear minimum. IMHO it is a blocking regression.

atomrc:
Here is the new sorting 
screenshot from 2018-10-02 16-30-53

bennycode:
I have a test case with `Alex` written in cyrillic letters which will be `Аlех`. 

Here is how the cyrillic Alex looks in an Hex-Editor:

image

And that's the latin Alex (both in UTF-8):

image

atomrc:
@bennyn thanks, will add a test case with that then :+1:

bennycode:
@atomrc: According to @red-avtovo the cyrillic version of `Аlех` can be transliterated to "Aleh".

red-avtovo:
Google translate tells, that cyrillic `х` is closer to Latin `kh`

atomrc:
In this algorithm we are using a speakingurl that does the transliteration for us, no need for us to do it manually

app/script/view_model/list/StartUIViewModel.js
```diff
@@ -560,12 +560,20 @@ z.viewModel.list.StartUIViewModel = class StartUIViewModel {
...
-      }
+      const localSearchSources = this.isTeam()
+        ? this.teamRepository.teamUsers()
+        : this.userRepository.connected_users();
+
+      const SEARCHABLE_FIELDS = z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS;
+      const localSearchFields = isHandle ? [SEARCHABLE_FIELDS.USERNAME] : undefined;
+
+      const contactSearchResults = this.searchRepository.searchUserInSet(
+        trimmedQuery.replace(/^@/, ''),
```
gregor: Let's use the normalized query.

PR #4605, opened 2018-09-19
feat: Upgrade search algorithm to tokenize properties (WEBAPP-5452)

atomrc:


app/script/search/SearchRepository.js
```diff
@@ -67,44 +66,69 @@ z.search.SearchRepository = class SearchRepository {
    *
    * @param {string} term - the search term
    * @param {Array<z.entity.User>} userEntities - entities to match the search term against
-   * @param {Array<z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS>} properties=['first_name', 'last_name', 'username'] - list of properties that will be matched against the search term
+   * @param {Array<z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS>} properties=['name', 'username'] - list of properties that will be matched against the search term
```
gregor: The jsdoc should also contain the constants. Do the defaults even belong here now since they are no longer part of the signature?

app/script/search/SearchRepository.js
```diff
@@ -67,44 +66,69 @@ z.search.SearchRepository = class SearchRepository {
...
+          return weight + propertyMatchWeight * propertyWeight;
+        }, 0);
+
+      return matchWeight === 0 ? results : results.concat({user: userEntity, weight: matchWeight});
     }, []);
 
     return weightedResults
       .slice()
-      .sort((res1, res2) => res2.weight - res1.weight)
+      .sort((res1, res2) => {
```
gregor: Chars are free: result1 and result2

app/script/search/SearchRepository.js
atomrc: I would keep them here (even if they are not in the signature) because we set them programmatically now but the behavior stays exactly the same from the user point of view

app/script/search/SearchRepository.js
```diff
@@ -76,38 +75,63 @@ z.search.SearchRepository = class SearchRepository {
       return userEntities;
     }
     const SEARCHABLE_FIELDS = z.search.SearchRepository.CONFIG.SEARCHABLE_FIELDS;
```
gregor: Since you are in the same file you don't need the `z.search` -> `CONFIG.SEARCHABLE_FIELDS` will do

atomrc: Ahhh, good to know :+1:

PR #4613, opened 2018-09-20
fix: make the drag detection less sensitive

AndyLnd:


gregor:
Is this supposed to address WEBAPP-5457? Then please tag accordingly and move ticket.

AndyLnd:
There is no ticket for this. This is a minor tweak.

PR #4616, opened 2018-09-20
feat: Bold conversation titles, transparent mute icon (WEBAPP-5460)

ffflorian:
This PR will
* set the conversation cell titles in bold when there is a description
* set the mute icon to 80 % visibility


app/style/list/list.less
```diff
@@ -178,10 +178,6 @@
 
     > .icon-silence {
       pointer-events: none;
-
-      &.conversation-muted {
-        opacity: 1;
-      }
```
ffflorian, line 184: This was not in use so I removed it and used the name for the badge style.

app/script/components/list/conversationListCell.js
```diff
@@ -68,7 +68,7 @@ ko.components.register('conversation-list-cell', {
           </availability-state>
         <!-- /ko -->
         <!-- ko ifnot: conversation.is_one2one() && conversation.self.inTeam() -->
-          <span class="conversation-list-cell-name" data-bind="text: conversation.display_name(), css: {'text-theme': is_selected(conversation)}"></span>
+          <span class="conversation-list-cell-name" data-bind="text: conversation.display_name(), css: {'text-theme': is_selected(conversation), 'conversation-list-cell-name--bold': !!cell_state().description}"></span>
```
gregor, line 71: Should we add a boolean value to the cell state instead of abusing the description?

ffflorian: I think that would add unnecessary logic which would be needed only for this one line.

PR #4618, opened 2018-09-20
feat: Insert and keep track of mentions (WEBAPP-5466)

atomrc:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -217,8 +215,30 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
 
   addMention(userEntity) {
-    // TODO
+    const editedMention = this.editedMention();
+    const mentionData = {
+      length: userEntity.name().length + 1,
+      start: editedMention.start,
+      user_id: userEntity.id,
+    };
+    const mention = new z.message.MentionEntity(mentionData);
```
gregor: MentionEntity take a proto mention in the constructor. For this the proper approach would be

````
const mentionEntity = new z.message.MentionEntity().setUserIdMention(this.editedMention().start, userEntity.name().length + 1, userEntity.id);
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 2]
```diff
@@ -461,34 +481,15 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
     if (messageText.length) {
-      const participatingUserEntities = this.conversationEntity().participating_user_ets();
-      const mentionEntities = this.parseForMentions(messageText, participatingUserEntities);
-      this.conversationRepository.sendTextWithLinkPreview(messageText, this.conversationEntity(), mentionEntities);
+      const mentionEntities = this.currentMentions;
+      this.conversationRepository
+        .sendTextWithLinkPreview(messageText, this.conversationEntity(), mentionEntities)
+        .then(() => {
+          this.currentMentions = [];
+        });
```
gregor: Single line

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -216,9 +214,30 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+      length: userEntity.name().length + 1,
+      start: editedMention.start,
+      user_id: userEntity.id,
+    };
+    const mentionEntity = new z.message.MentionEntity(mentionData);
+
+    // keep track of what is before and after the mention being edited
+    const beforeMentionStr = this.input().slice(0, mentionEntity.startIndex);
+    const afterMentionStr = this.input()
+      .slice(mentionEntity.startIndex + editedMention.term.length + 1, this.input().length)
```
AndyLnd: No need for the second parameter.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -216,8 +214,29 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
-  addMention(userEntity) {
-    // TODO
+  addMention(userEntity, inputElement) {
+    const editedMention = this.editedMention();
+    const mentionData = {
+      length: userEntity.name().length + 1,
+      start: editedMention.start,
+      user_id: userEntity.id,
+    };
+    const mentionEntity = new z.message.MentionEntity(mentionData);
```
gregor: Still a valid comment:

````
const mentionEntity = new z.message.MentionEntity().setUserIdMention(this.editedMention().start, userEntity.name().length + 1, userEntity.id);
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 5]
```diff
@@ -461,34 +480,13 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
   sendMessage(messageText) {
     if (messageText.length) {
-      const participatingUserEntities = this.conversationEntity().participating_user_ets();
-      const mentionEntities = this.parseForMentions(messageText, participatingUserEntities);
-      this.conversationRepository.sendTextWithLinkPreview(messageText, this.conversationEntity(), mentionEntities);
+      const mentionEntities = this.currentMentions;
+      this.conversationRepository
+        .sendTextWithLinkPreview(messageText, this.conversationEntity(), mentionEntities)
+        .then(() => (this.currentMentions = []));
```
gregor: No reason to wait for the then. If something fails, mentions will not be unset.

````
this.conversationRepository.sendTextWithLinkPreview(messageText, this.conversationEntity(), this.currentMentions);
this.currentMentions = [];
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 6]
```diff
@@ -216,8 +214,29 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+    // keep track of what is before and after the mention being edited
+    const beforeMentionStr = this.input().slice(0, mentionEntity.startIndex);
+    const afterMentionStr = this.input()
+      .slice(mentionEntity.startIndex + editedMention.term.length + 1, this.input().length)
+      .replace(/^ /, '');
+
+    // insert the mention in between
+    this.input(`${beforeMentionStr}@${userEntity.name()} ${afterMentionStr}`);
+
+    const caretPosition = mentionEntity.startIndex + mentionEntity.length + 1;
```
AndyLnd: ```const caretPosition = mentionEntity.endIndex + 1;```

app/script/view_model/content/InputBarViewModel.js
atomrc: Then we have a problem with the API of the `MentionEntity`. Why is there 2 ways to feed data in there?

I would rather go for the constructor API (rather than a setter method) so that we can ensure that we build a correct `MentionEntity` from the start. WDYT?

app/script/view_model/content/InputBarViewModel.js
AndyLnd: I would argue that they could even be unset at the same position as the input is set to `''` in `onInputEnter`.

atomrc: @gregor @AndyLnd fair points. I think I'll go for the `onInputEnter` solution :+1:

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -216,8 +214,24 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
-  addMention(userEntity) {
-    // TODO
+  addMention(userEntity, inputElement) {
+    const editedMention = this.editedMention();
+    const mentionEntity = new z.message.MentionEntity(editedMention.start, userEntity.name().length + 1, userEntity.id);
+    this.currentMentions.push(mentionEntity);
+
+    // keep track of what is before and after the mention being edited
+    const beforeMentionStr = this.input().slice(0, mentionEntity.startIndex);
+    const afterMentionStr = this.input()
```
gregor: No need for abbreviations

test/unit_tests/view_model/InputBarViewModelSpec.js
```diff
@@ -22,6 +22,7 @@
 // grunt test_init && grunt test_run:view_model/InputBarViewModel
 
 describe('z.viewModel.content.InputBarViewModel', () => {
+  /*
```
gregor, line 25: What's this? If we remove the tests we can remove it.

atomrc: I wanted to keep this init code somewhere for the moment we actually write some tests.
Unfortunatly because the only test there was a method that was removed, this code is dead and eslint complains about it.

My point here, is that I want people who will want to write test in the future to be able to write them quickly without having to think about how to instantiate the ViewModel.

PR #4619, opened 2018-09-20
fix: Guard undefined OS

Yserz:


app/script/auth/Runtime.js
```diff
@@ -55,10 +55,10 @@ export const SUPPORTED_BROWSERS = {
 };
 
 const getPlatform = () => platform;
-const getOs = () => getPlatform().os;
+const getOs = () => getPlatform().os || {};
 const getUserAgent = () => getPlatform().ua.toLowerCase();
 
-const getOsFamily = () => getOs().family.toLowerCase();
+const getOsFamily = () => (getOs().family || '').toLowerCase();
```
gregor: Maybe we should put the guard as the default object in l58 instead of touching many lines

Yserz: That doesn't cater for cases where `os` is defined but single values are null.

PR #4620, opened 2018-09-20
fix: Add trigger to member message notification (WEBAPP-5472)

gregor:


app/script/notification/NotificationRepository.js
```diff
@@ -564,31 +564,19 @@ z.notification.NotificationRepository = class NotificationRepository {
...
-          amplify.publish(z.event.WebApp.CONTENT.SWITCH, z.viewModel.ContentViewModel.STATE.CONNECTION_REQUESTS);
-        };
-      }
-
-      default: {
-        const message = `No notification trigger for message '${messageEntity.id} in '${conversationId}'.`;
-        this.logger.log(this.logger.levels.OFF, message);
-      }
-    }
+    return () => amplify.publish(z.event.WebApp.CONVERSATION.SHOW, conversationEntity || conversationId);
```
atomrc, line 579: Now that we removed the whitelist, we might run into publishing `CONVERSATION.SHOW` events for things we don't want. Is that expected?

gregor: Yes. That is actually the bug. Check https://github.com/wireapp/wire-webapp/pull/4595. My PR from two days ago refactored some code and I missed the fact that the switch case had a fall through. The conversation show was the default, despite some exceptions doing already the exact same thing. This simplifies the code a lot and fixes a bug from the original code that was masked by the fallthough.

atomrc: Ok then, if you feel it's safe, then LGTM. thanks for the explanation.

gregor: Testing the trigger in the unit tests seems complicated. Maybe we should add that at some point. For now we were only looking at the stringified content.

PR #4622, opened 2018-09-20
chore: Simplify MentionEntity constructor

gregor:


app/script/conversation/EventMapper.js
```diff
@@ -661,7 +661,10 @@ z.conversation.EventMapper = class EventMapper {
   _mapAssetMentions(mentions, messageText) {
     return mentions
       .map(encodedMention => z.proto.Mention.decode64(encodedMention))
-      .map(mention => new z.message.MentionEntity(mention))
+      .map(protoMention => {
+        const {length, start: startIndex, user_id: userId} = protoMention;
```
AndyLnd: I really don't think you need to rename those here, as they are consumed on the next line and it doesn't really improve readability.

gregor: I am mapping proto naming to our camelCase specification. Once we enable that eslint flag it should actually throw here without. I am using the chance to align the name to the naming used within the mentionEntity.

app/script/conversation/EventMapper.js
```diff
@@ -661,7 +661,7 @@ z.conversation.EventMapper = class EventMapper {
   _mapAssetMentions(mentions, messageText) {
     return mentions
       .map(encodedMention => z.proto.Mention.decode64(encodedMention))
-      .map(mention => new z.message.MentionEntity(mention))
+      .map(protoMention => new z.message.MentionEntity(protoMention.start, protoMention.length, protoMention.user_Id))
```
gregor, line 664: @AndyLnd Do you prefer this?

PR #4623, opened 2018-09-20
feat: Add mentions to conversation description (WEBAPP-5460)

ffflorian:
This PR will
* add "1 mention" / "2 mentions" to the secondary line of a group
* change "1 new message" / "2 new messages" to "1 message" / "2 messages" in the secondary line of a group
* add a @ symbol next to the group name when the user was mentioned


app/script/entity/message/Message.js
```diff
@@ -202,6 +202,12 @@ z.entity.Message = class Message {
     );
   }
 
+  isMention() {
+    return (
+      this.has_asset_text() && this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.mentions().length)
```
gregor: Only self mentions are counted. There is already a `isSelfMentioned()` in place on the message to take care of that. You can add the some assets check there since it potentially fails in my code I belive.

ffflorian: Yes, thanks, I just noticed, too 🤦‍♂️

app/script/entity/message/Message.js
```diff
@@ -144,6 +144,12 @@ z.entity.Message = class Message {
     return this.is_content() ? this.assets().some(assetEntity => assetEntity.is_text()) : false;
   }
 
+  hasSelfMention() {
+    return (
+      this.has_asset_text() && this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.isSelfMentioned())
```
gregor: Again: Check the `ContentMessage`. There already is a `isSelfMentioned` for this purpose. It just needs to be adjusted to not only check the first asset (despite that being enough in the wild).

app/script/localization/webapp.js
```diff
@@ -354,6 +354,8 @@ z.string.conversationResetTimerYou = ' turned off the message timer';
 z.string.conversationResume = 'Start a conversation with {{users}}';
 z.string.conversationPing = ' pinged';
 z.string.conversationPingYou = ' pinged';
+z.string.conversationMention = ' mentioned';
+z.string.conversationMentionYou = ' mentioned';
```
gregor: Why are you adding these?

ffflorian: Thanks, they should not be part of this PR.

ffflorian:
Updated the PR to include non-muted groups, too.

app/script/conversation/ConversationCellState.js
```diff
@@ -33,6 +34,7 @@ z.conversation.ConversationCellState = (() => {
     const activities = {
       [ACTIVITY_TYPE.CALL]: 0,
       [ACTIVITY_TYPE.MESSAGE]: 0,
+      [ACTIVITY_TYPE.MENTION]: 0,
       [ACTIVITY_TYPE.PING]: 0,
```
gregor: I suppose that the string is supposed to be build in a certain order. The order of this object is mapped to the string by https://github.com/wireapp/wire-webapp/pull/4623/files#diff-e5ae7a9620ad8c48c4dccef5b8bae678R58. Looking at the spec this is not the expected order. Please clarify the order of priority within the string with design.

app/script/conversation/ConversationCellState.js
```diff
@@ -100,6 +111,10 @@ z.conversation.ConversationCellState = (() => {
           return z.conversation.ConversationStatusIcon.UNREAD_PING;
         }
 
+        if (lastAlertMessage.isSelfMentioned()) {
+          return z.conversation.ConversationStatusIcon.UNREAD_MENTION;
+        }
```
gregor: I am not sure on this one reading the spec. Please clarify with design. It seems that we currently show the icon matching the latest alert in the conversation. I believe that alerts have a priority. Thus it should always show the call icon if there is any call; if not it should always show the mention icon if there is any mention, and so on. Relates to above comment about priority.

app/script/entity/message/ContentMessage.js
```diff
@@ -109,7 +109,9 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
   }
 
   isSelfMentioned() {
-    return this.has_asset_text() && this.get_first_asset().isSelfMentioned();
+    return (
+      this.has_asset_text() && this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.isSelfMentioned())
+    );
```
gregor: Sorry to be a little bit annoying but I am not a fan of single lines if it doesn't fit the single line 🤷‍♂️ 

I would prefer:

````
return this.has_asset_text()
  ? this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.isSelfMentioned())
  : false;
````

or

````
if (this.has_asset_text()) {
  return this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.isSelfMentioned());
}
````

Where the second one would return void instead of false in case there is no text asset.

File: app/script/conversation/ConversationCellState.js [hunk 7]
```diff
@@ -93,13 +104,17 @@ z.conversation.ConversationCellState = (() => {
...
+        return z.conversation.ConversationStatusIcon.UNREAD_MENTION;
+      }
+
       const hasMissedCall = conversationEntity
         .unreadEvents()
         .some(messageEntity => messageEntity.is_call() && messageEntity.was_missed());
       if (hasMissedCall) {
         return z.conversation.ConversationStatusIcon.MISSED_CALL;
       }
-
```
gregor: Let's not remove this.

app/script/entity/message/ContentMessage.js
```diff
@@ -109,7 +109,9 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
   }
 
   isSelfMentioned() {
-    return this.has_asset_text() && this.get_first_asset().isSelfMentioned();
+    return this.has_asset_text()
+      ? this.assets().some(assetEntity => assetEntity.mentions().some(mentionEntity => mentionEntity.isSelfMentioned()))
+      : false;
```
gregor:
````
return this has_asset_text()
  ? this.assets().some(assetEntity => assetEntity.is_text() && assetEntity.isSelfMentioned())
  : false;
````

File: app/script/entity/message/ContentMessage.js [hunk 9]
```diff
@@ -109,7 +109,12 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
   }
 
   isSelfMentioned() {
-    return this.has_asset_text() && this.get_first_asset().isSelfMentioned();
+    return this.has_asset_text()
+      ? this.assets().some(
+          assetEntity =>
+            assetEntity.is_text() && assetEntity.mentions().some(mentionEntity => mentionEntity.isSelfMentioned())
+        )
```
gregor:
````
  ? this.assets().some(assetEntity => {
    return assetEntity.is_text() && assetEntity.mentions().some(mentionEntity => mentionEntity.isSelfMentioned());
  })
````

PR #4624, opened 2018-09-20
feat: Style suggestion popup (WEBAPP-5461)

AndyLnd:


app/script/components/mentionSuggestions.js
```diff
@@ -111,12 +111,18 @@ z.components.MentionSuggestions = class MentionSuggestions {
...
   template: `
   <!-- ko if: isVisible() -->
-    <div class="conversation-input-bar-mention-suggestion" data-uie-name="list-mention-suggestions" data-bind="style: positions()">
-      <ul class="mention-suggestion-list" data-bind="foreach: {data: suggestions, as: 'suggestion'}">
-        <li data-bind="style: {color: suggestion === $parent.selectedSuggestion() ? 'green' : ''}">
-          <span data-bind="text: suggestion.username()"></span>
-        </li>
-      </ul>
+    <div class="
+    " data-uie-name="list-mention-suggestions" data-bind="style: positions()">
```
gregor: This is weird...

app/style/components/mention-suggestions.less
```diff
@@ -17,7 +17,62 @@
...
+
+      svg {
+        width: 12px;
+        height: 12px;
+        path {
+          fill: fade(@graphite-dark, 40%);
+        }
+      }
+    }
+  }
```
gregor, line 77: I am missing the full width definition with right margin of the complete list that is mentioned in the spec.

AndyLnd: We are not doing the full width anymore. And this should be reflected in the specs.

gregor: So the width will jump depending on which suggestions get shown in the list?

AndyLnd: Yes, we will see how that works out and might introduce a maximum width.

PR #4625, opened 2018-09-20
fix: Priortize types in secondary line alert state

ffflorian:
This PR will
* correct the priority order of icons shown next to the conversation title in the conversation list
* correct the priority order of the subtitle string in the conversation list

Correct order:
1. Missed call
2. Ping
3. Normal message

### DIFF
--- a/app/script/conversation/ConversationCellState.js
+++ b/app/script/conversation/ConversationCellState.js
@@ -65,17 +65,17 @@ z.conversation.ConversationCellState = (() => {
               break;
             }
 
-            case ACTIVITY_TYPE.PING: {
-              stringId = activityCountIsOne
-                ? z.string.conversationsSecondaryLinePing
-                : z.string.conversationsSecondaryLinePings;
-              break;
-            }
-
             case ACTIVITY_TYPE.MESSAGE: {
               stringId = activityCountIsOne
                 ? z.string.conversationsSecondaryLineNewMessage
                 : z.string.conversationsSecondaryLineNewMessages;
+              break;
+            }
+
+            case ACTIVITY_TYPE.PING: {
+              stringId = activityCountIsOne
+                ? z.string.conversationsSecondaryLinePing
+                : z.string.conversationsSecondaryLinePings;
               break;
             }
 
--- a/app/script/conversation/ConversationCellState.js
+++ b/app/script/conversation/ConversationCellState.js
@@ -93,17 +93,16 @@ z.conversation.ConversationCellState = (() => {
   const _getStateAlert = {
     description: conversationEntity => _accumulateActivity(conversationEntity),
     icon: conversationEntity => {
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationCellState.js
```diff
@@ -96,14 +96,14 @@ z.conversation.ConversationCellState = (() => {
...
-        }
-
         const isMissedCall = lastAlertMessage.is_call() && lastAlertMessage.was_missed();
         if (isMissedCall) {
           return z.conversation.ConversationStatusIcon.MISSED_CALL;
         }
+
+        if (lastAlertMessage.is_ping()) {
+          return z.conversation.ConversationStatusIcon.UNREAD_PING;
+        }
```
gregor: This does not fix the icon but only moves code around without a behavior change. Intended?

We currently check the unread messages for the latest ping or call. The icon will be mapped to the latest alert. It should rather check all unread messages for the highest priority alert message: `call > ping > normal text message`

app/script/conversation/ConversationCellState.js
```diff
@@ -65,17 +65,17 @@ z.conversation.ConversationCellState = (() => {
...
               break;
             }
 
-            case ACTIVITY_TYPE.PING: {
+            case ACTIVITY_TYPE.MESSAGE: {
               stringId = activityCountIsOne
-                ? z.string.conversationsSecondaryLinePing
-                : z.string.conversationsSecondaryLinePings;
+                ? z.string.conversationsSecondaryLineNewMessage
+                : z.string.conversationsSecondaryLineNewMessages;
```
gregor: The change above unnecessary and should be reverted. Sorting a switch case alphabetically is highly perfered. This only moves code around with no functional change.

app/script/conversation/ConversationCellState.js
ffflorian: Why without behaviour change? If there was a ping and a missed call the function will now return and display the missed call icon instead of displaying the ping icon.

gregor: No, it will not.

`const lastAlertMessage = conversationEntity.unreadEvents().find(_isAlert);` will get the last unread message that is of any type of alert. The code you changed is essentialy a switch case deciding what type of message is connected to that. There it does not matter in which order the lines of code are. It is only checking one message, the same one. With the same result.

gregor: What you need to do instead: 

````
const hasMissedCall = conversationEntity.unreadEvents().some(messageEntity => messageEntity .is_call() && messageEntity .was_missed());
if (hasMissedCall) {
  return z.conversation.ConversationStatusIcon.MISSED_CALL;
}

const hasPing = ...
````

ffflorian: Alright, thanks. Need to get used to this project... :see_no_evil:

PR #4633, opened 2018-09-21
fix: Non team accounts should not see guest banner (WEBAPP-5438)

gregor:


app/script/entity/Conversation.js
```diff
@@ -81,7 +81,8 @@ z.entity.Conversation = class Conversation {
     this.is_self = ko.pureComputed(() => this.type() === z.conversation.ConversationType.SELF);
 
     this.hasGuest = ko.pureComputed(() => {
-      return this.is_group() && this.participating_user_ets().some(userEntity => userEntity.isGuest());
+      const hasGuestUser = this.participating_user_ets().some(userEntity => userEntity.isGuest());
+      return hasGuestUser && this.is_group() && this.self.inTeam();
```
gregor, line 85: Logical change that checks whether the self user is a team account.

app/script/conversation/ConversationType.js
```diff
@@ -25,7 +25,7 @@ window.z.conversation = z.conversation || {};
 // Enum of different conversation types.
 z.conversation.ConversationType = {
   CONNECT: 3,
+  GROUP: 0,
   ONE2ONE: 2,
-  REGULAR: 0,
```
gregor, line 29: I have tripped over this naming many times. This makes a lot more sense.

PR #4636, opened 2018-09-21
feat: Preserve mentions from input in local storage (WEBAPP-5474)

ffflorian:
This PR will
* save the current mentions from input to the local storage
* apply the loaded mentions from the local storage to input
* safe-guard old local storage data


app/page/template/content/conversation/input-bar.htm
```diff
@@ -42,7 +42,7 @@
             </gif-icon>
           <!-- /ko -->
 
-          <!-- ko ifnot: input().length -->
+          <!-- ko ifnot: input().text.length -->
```
gregor: Does this actually work? Text is no longer an observable. Changes to it should remain undetected by knockout.

ffflorian: But `input()` still is an observable. Before `input()` just returned a string, now it returns an object.

ffflorian: And yes, it works :slightly_smiling_face:

gregor: Weird. Because the object rarely changes. It stays the same object and only parts of it change.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -75,20 +75,29 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
-          return this.editInput();
+          inputObject = this.editInput();
+        } else if (this.conversationEntity() && this.conversationEntity().input()) {
+          inputObject = this.conversationEntity().input();
         }
 
-        if (this.conversationEntity()) {
-          return this.conversationEntity().input() || '';
-        }
+        this.currentMentions = inputObject.mentions.map(
```
atomrc: There is a weird relation between the `input` and the `currentMentions`. It seems to be an unecessary information duplication.

I we go for a structure that has both the text value and the mentions, I would go for fully using this structure and remove completely the `currentMentions`.
Probably I would rename the `input` observable `editedMessage`, and this will be the ground truth of our text input.
WDYT?

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -91,11 +91,12 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
         } else if (this.conversationEntity() && this.conversationEntity().input()) {
           inputObject = this.conversationEntity().input();
-        }
 
-        this.currentMentions = inputObject.mentions.map(
-          mention => new z.message.MentionEntity(mention.startIndex, mention.length, mention.userId)
-        );
+          this.currentMentions = inputObject.mentions.map(
+            mention => new z.message.MentionEntity(mention.startIndex, mention.length, mention.userId)
+          );
```
gregor:
`````
this.currentMentions = inputObject.mentions.map(mention => {
  return new ....
}):
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 4]
```diff
@@ -75,28 +76,42 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
         if (this.isEditing()) {
-          return this.editInput();
-        }
+          inputObject = this.editInput();
+        } else if (this.conversationEntity() && this.conversationEntity().input()) {
+          inputObject = this.conversationEntity().input();
 
-        if (this.conversationEntity()) {
-          return this.conversationEntity().input() || '';
+          this.currentMentions = inputObject.mentions;
```
atomrc: Let's talk 1:1 to find a proper solution to completely get rid of this duplication of information

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -133,7 +138,12 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     });
 
     this.editedMention = ko.observable(undefined);
-    this.currentMentions = [];
+    this.currentMentions = ko.observableArray();
```
atomrc: I actually think you don't need the power of Observables here. Nothing seem to react to that array changes, you can go for a good ol' javascript Array

gregor:
@ffflorian Once this is merged I will open a tiny PR with some unrelated refactoring in this area. Please ping  me on merge.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -361,7 +361,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       this.sendMessage(messageText);
     }
 
-    this.currentMentions.removeAll();
+    this.currentMentions = [];
```
gregor: Let's not always create a new array. You can do `this.currentMentions.length = 0`

PR #4637, opened 2018-09-21
fix: Edge character behavior, deletion fix, rendering fix (WEBAPP-5463)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -41,6 +41,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   constructor(mainViewModel, contentViewModel, repositories) {
     this.addedToView = this.addedToView.bind(this);
     this.addMention = this.addMention.bind(this);
+    this.updateSelectionState = this.updateSelectionState.bind(this);
```
gregor: ABC

PR #4639, opened 2018-09-21
fix: Add missing content type checks on incoming messages

ffflorian:
This PR will
* always check if a message is of type "content" before checking if it has self mentions.

### DIFF
--- a/app/script/conversation/ConversationCellState.js
+++ b/app/script/conversation/ConversationCellState.js
@@ -311,9 +311,9 @@ z.conversation.ConversationCellState = (() => {
   };
 
   const _isAlert = messageEntity => {
-    const isSelfMenionted = messageEntity.is_content() && messageEntity.isSelfMentioned();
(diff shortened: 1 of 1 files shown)


app/script/conversation/ConversationCellState.js
```diff
@@ -310,7 +312,7 @@ z.conversation.ConversationCellState = (() => {
 
   const _isAlert = messageEntity => {
     const isMissedCall = messageEntity.is_call() && messageEntity.was_missed();
-    return isMissedCall || messageEntity.is_ping() || messageEntity.isSelfMentioned();
+    return isMissedCall || messageEntity.is_ping() || (messageEntity.is_content() && messageEntity.isSelfMentioned());
```
gregor:
````
const isSelfMenionted = messageEntity.is_content() && messageEntity.isSelfMentioned():
const isMissedCall = messageEntity.is_call() && messageEntity.was_missed();
return isSelfMenionted || isMissedCall || messageEntity.is_ping();
````

PR #4648, opened 2018-09-24
feat: Better mention flow detection, reentering flow (WEBAPP-5463)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -398,24 +398,24 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     const textarea = document.querySelector('#conversation-input-bar-text');
```
atomrc, line 398: How about we call that function `handleMentionFlow` to be more specific

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -489,7 +489,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
   onInputKeyUp(data, keyboardEvent) {
     this.emojiInput.onInputKeyUp(data, keyboardEvent);
-    this.updateSelectionState();
+    if (keyboardEvent.key !== z.util.KeyboardUtil.KEY.ESC) {
```
atomrc, line 491: Shouldn't this be handled by the MentionSuggestion component? I can close itself when the `ESC` key is pressed

AndyLnd: To clarify: the closing is still handled by the MentionSuggestion component. This guard here just makes sure that it isn't immediatly reopened when `esc` is released.

PR #4651, opened 2018-09-24
feat: Handle mention actions (WEBAPP-5465)

atomrc:


app/style/content/conversation/message-list.less
```diff
@@ -93,6 +93,9 @@
       color: @graphite-dark;
       .accent-background-color(16%);
     }
+    &:not(.self-mention) {
```
gregor: Empty line above

PR #4652, opened 2018-09-24
fix: Mentioning the same user twice (WEBAPP-5483)

bennyn:
When constructing `z.message.MentionEntity`, its `userEntity` observable is `undefined` and can break the access on `this.userEntity().is_me`. 

That's why I added the check for `this.userEntity()`.


app/script/message/MentionEntity.js
```diff
@@ -33,7 +33,7 @@ z.message.MentionEntity = class Mention {
 
     this.isSelfMentioned = ko.pureComputed(() => {
       const isTypeUserId = this.type === z.cryptography.PROTO_MESSAGE_TYPE.MENTION_TYPE_USER_ID;
-      return isTypeUserId && this.userEntity().is_me;
+      return isTypeUserId && (this.userEntity() && this.userEntity().is_me);
```
gregor: No need for additional parenthesis.

PR #4656, opened 2018-09-24
fix: Edit messages with mentions (WEBAPP-5482, WEBAPP-5486)

ffflorian:


gregor:
This history looks weird. Can you squash it and rebase it on the current edge please?

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -296,6 +297,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       this.editMessageEntity(messageEntity);
       this.editMessageEntity().isEditing(true);
       this.input(this.editMessageEntity().get_first_asset().text);
+      this.currentMentions = this.editMessageEntity()
+        .get_first_asset()
+        .mentions();
       if (inputElement) {
```
gregor: I suppose we can simplify things a bit if we apply the state to the plain `messageEntity` - instead of getting it from the observable every time.

````
messageEntity.isEditing(true);
this.input(messageEntity.get_first_asset().text);
this.currentMentions = messageEntity.get_first_asset().mentions();
this.editMessageEntity(messageEntity);
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 2]
```diff
@@ -524,6 +529,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   }
 
   sendMessageEdit(messageText, messageEntity) {
+    const mentionEntities = [...this.currentMentions];
     this.cancelMessageEditing();
```
ffflorian: We need to make a copy here, because `this.cancelMessageEditing()` will delete the array.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -524,6 +527,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   }
 
   sendMessageEdit(messageText, messageEntity) {
+    const mentionEntities = [...this.currentMentions];
```
gregor: Fancy but I think a simple `this.currentMentions.slice()` will do in this case. No need for spread operator.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -285,7 +286,8 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       this.cancelMessageEditing();
       this.editMessageEntity(messageEntity);
       this.editMessageEntity().isEditing(true);
```
gregor: This is also still the same message....

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -179,6 +174,13 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       this.conversationHasFocus(true);
       this.pastedFile(null);
       this.cancelMessageEditing();
+      if (this.conversationEntity()) {
+        const mentions = this.conversationEntity().input().mentions;
+        this.currentMentions = mentions;
+        if (this.currentMentions.length) {
+          this.input(`${this.input()} `);
```
ffflorian: We need to add a space here or the current mentions index will be off by 1.

gregor: Can you please explain why? This sounds weird and hacky.

ffflorian: No, sorry, I can't explain this but probably @AndyLnd can.

ffflorian: Ok seems like it's not related to the space. The index is still sometimes off by 1 when loading from local storage.

atomrc: My guess would be that this is probably to trigger an observable change somewhere and the side effects that we want with it. Depending on what this is we might want to explicitly trigger them..?

PR #4674, opened 2018-09-25
fix: Use local peerConnection config to check for valid ICE candidates gathering (WEBAPP-5494)

atomrc:


gregor:
Can we implement this in a way that it is feature aware? Once the getConfiguration function is there, we will use it. The saved state is only a temporary fallback. And we add a comment that this is fixed in Chrome 70?

atomrc:
@gregor yes, let me update the PR

app/script/calling/entities/FlowEntity.js
```diff
@@ -862,7 +869,9 @@ z.calling.entities.FlowEntity = class FlowEntity {
 
         const isModeDefault = this.negotiationMode() === z.calling.enum.SDP_NEGOTIATION_MODE.DEFAULT;
         if (isModeDefault && sendingOnTimeout) {
-          const connectionConfig = this.peerConnection.getConfiguration();
+          const connectionConfig = this.peerConnection.getConfiguration
+            ? this.peerConnection.getConfiguration()
+            : this.peerConnectionConfiguration;
```
gregor: I would propose the following in case anything else is wrong in the implementation:

````
const connectionConfig = this.peerConnection.getConfiguration && this.peerConnection.getConfiguration() || this.peerConnectionConfiguration;
````

PR #4679, opened 2018-09-25
fix: Disable syntax highlight when mentions are in the code block (WEBAPP-5497)

atomrc:


app/script/util/util.js
```diff
@@ -324,6 +324,15 @@ z.util.renderMessage = (message, mentionEntities = []) => {
...
   mentionlessText = marked(mentionlessText, {
     highlight: function(code) {
+      const containsMentions = mentionEntities.some(mention => {
+        const hash = createMentionHash(mention);
+        return code.includes(hash);
+      });
+      if (containsMentions) {
+        // disable code highlighting if there is a mention in there
+        // hightlighting will be wrong anyway because this is not valid code
+        return code;
```
ffflorian, line 334:
```
` ` `
console.log('@atomrc')
` ` `
```
Isn't that valid code?

atomrc: It's valid if it is raw text. A mention is not something you can put into code

atomrc: But, just to be clear, your example will get syntax highligh

PR #4680, opened 2018-09-25
fix: Update selection state after loading mentions (WEBAPP-5489)

ffflorian:


app/script/entity/Conversation.js
```diff
@@ -43,7 +43,7 @@ z.entity.Conversation = class Conversation {
     const inputStorageKey = `${z.storage.StorageKey.CONVERSATION.INPUT}|${this.id}`;
     this.input = ko.observable(this._getStorageInputData(inputStorageKey));
     this.input.subscribe(({mentions, text}) => {
-      return z.util.StorageUtil.setValue(inputStorageKey, {mentions, text: text.trim()});
+      return z.util.StorageUtil.setValue(inputStorageKey, {mentions, text: text.trimEnd()});
```
ffflorian, line 46: If we trim the whitespace in the beginning, the loaded mentions will be in the wrong position.

atomrc: :+1: for the explanation here.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -114,6 +114,8 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     });
 
     this.richTextInput = ko.pureComputed(() => {
+      this.updateSelectionState();
```
atomrc, line 117: This side effect doesn't belong here in a pureComputed. What is it achieving?

ffflorian: It updates the internal mention ranges (in `this.currentMentions`) . As I was told by @AndyLnd this is needed to ensure that the mentions are up-to-date and the correct beginnings and endings of the mentions are saved.

PR #4681, opened 2018-09-25
fix: correct size for suggestion box, show temporary guests (WEBAPP-5502)

AndyLnd:


app/script/components/mentionSuggestions.js
```diff
@@ -58,10 +58,12 @@ z.components.MentionSuggestions = class MentionSuggestions {
     this.shouldUpdateScrollbar = ko.pureComputed(() => this.suggestions()).extend({notify: 'always', rateLimit: 100});
     this.shouldUpdateScrollbar.subscribe(() => {
       z.util.afterRender(() => {
-        const item = document.querySelector('.mention-suggestion-list__item');
+        const items = document.querySelectorAll('.mention-suggestion-list__item');
         const wrapper = document.querySelector('.conversation-input-bar-mention-suggestion');
-        if (item && wrapper) {
-          wrapper.style.width = `${item.offsetWidth}px`;
+        if (items.length && wrapper) {
+          const itemWidths = items.maps(item => item.offsetWidth);
```
gregor: `maps()` - I suppose you want to do `map()`? You are trying to get the highest width value. How about `reduce()`?

PR #4682, opened 2018-09-25
fix: Do no fail silently if it is not possible to determine if mention is self mention (WEBAPP-5496)

atomrc:
This changes a bit the `MentionEntity` API.

It doesn't need a full userEntity now and has a new method that set who is going to read the message

It also crash if we ask it if it's a self mention while the reader is not set

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -3564,9 +3564,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
 
           const textAsset = messageEntity.get_first_asset();
           if (textAsset.mentions && textAsset.mentions().length) {
-            textAsset.mentions().forEach(mentionEntity => {
-              mentionEntity.setReaderId(this.selfUser().id);
-            });
+            textAsset.mentions().forEach(mentionEntity => mentionEntity.setSelfId(this.selfUser().id));
           }
         }
       }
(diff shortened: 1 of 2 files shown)


app/script/conversation/ConversationRepository.js
```diff
@@ -3560,20 +3560,9 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
-                mentionedUserEntities.forEach(mentionedUserEntity => {
-                  const mentionEntity = textAsset.mentions().find(({userId}) => mentionedUserEntity.id === userId);
-                  mentionEntity.userEntity(mentionedUserEntity);
-                });
-                return messageEntity;
-              });
-            }
+            textAsset.mentions().forEach(mentionEntity => {
+              mentionEntity.setReaderId(this.selfUser().id);
+            });
```
gregor: Should fit single line

app/script/message/MentionEntity.js
```diff
@@ -43,10 +43,20 @@ z.message.MentionEntity = class MentionEntity {
...
 
-    this.isSelfMentioned = ko.pureComputed(() => {
-      const isTypeUserId = this.type === z.cryptography.PROTO_MESSAGE_TYPE.MENTION_TYPE_USER_ID;
-      return isTypeUserId && this.userEntity() && this.userEntity().is_me;
-    });
+    // id of the user who will read the mention
+    this.readerId = undefined;
+  }
+
+  setReaderId(userId) {
```
gregor: We refer to the reader as self in any case. So let's continue to do so here: `setSelfId`

PR #4685, opened 2018-09-25
fix: trimEnd polyfill (WEBAPP-5509)

ffflorian:


app/script/util/util.js
```diff
@@ -148,6 +148,19 @@ z.util.getFileExtension = filename => {
...
+ * Trim end of a string.
+ * @param {string} text - text to trim
+ * @returns {string} Trimmed string
+ */
+z.util.trimEnd = text => {
+  if (_.isString(text)) {
+    return text.replace(/\s*$/, '');
+  }
+
+  return '';
```
gregor: Can easily be place in a ternary

app/script/util/util.js
```diff
@@ -148,6 +148,19 @@ z.util.getFileExtension = filename => {
   return filename.substr(filename.lastIndexOf('.') + 1);
 };
 
+/**
+ * Trim end of a string.
+ * @param {string} text - text to trim
+ * @returns {string} Trimmed string
+ */
+z.util.trimEnd = text => {
+  if (_.isString(text)) {
```
atomrc: I would not put safeguards here. We probably want to know if we call this with an unexpected input

atomrc: Also how about do feature detection here.

```
return text.trimEnd ? text.trimEnd() : text.replace(...)
```

ffflorian:
@gregor @atomrc What about the safeguard? Yes or no?

bennycode:
Going forward we should investigate in using [Babel](https://babeljs.io/) for the webapp code.

app/script/util/util.js
```diff
@@ -148,6 +148,15 @@ z.util.getFileExtension = filename => {
...
 };
 
+/**
+ * Trim end of a string.
+ * @param {string} text - text to trim
+ * @returns {string} Trimmed string
+ */
+z.util.trimEnd = text => {
+  return _.isString(text) ? text.replace(/\s*$/, '') : '';
+};
```
gregor: Arrow function fits into single line.

ffflorian:
Should this go to `StringUtil`?

atomrc:
Yes

PR #4705, opened 2018-09-26
test: Implement additional mention validation cases

gregor:


test/unit_tests/message/MentionEntitySpec.js
```diff
@@ -76,7 +71,15 @@ describe('MentionEntity', () => {
 
     it('should return true on validation', () => {
       const mentionEntity = new z.message.MentionEntity(14, 10, userId);
-      expect(mentionEntity.validate(textMessage)).toBe(true);
+      expect(mentionEntity.validate(textMessage)).toBeTruthy();
```
bennycode: `validate` always returns a `boolean`, right? If so, then `toBe(true)` is a better match here.

PR #4706, opened 2018-09-26
refactor: Move to a stateless MentionEntity (WEBAPP-5516)

atomrc:
We now explicitly pass the selfUserId when we need to compare against it and not rely on previous piece of code to have set it.

This will allow us to avoid running into temporal coupling.

### DIFF
--- a/app/script/entity/Conversation.js
+++ b/app/script/entity/Conversation.js
@@ -394,7 +394,9 @@ z.entity.Conversation = class Conversation {
     return this.unreadEvents()
       .slice()
       .reverse()
-      .find(messageEntity => messageEntity.visible() && messageEntity.is_content() && messageEntity.isSelfMentioned());
+      .find(messageEntity => {
+        return messageEntity.visible() && messageEntity.is_content() && messageEntity.isSelfMentioned(this.self.id);
+      });
   }
 
   get_last_known_timestamp(time_offset) {
--- a/app/script/entity/Conversation.js
+++ b/app/script/entity/Conversation.js
@@ -395,7 +395,7 @@ z.entity.Conversation = class Conversation {
       .slice()
       .reverse()
       .find(messageEntity => {
-        return messageEntity.visible() && messageEntity.is_content() && messageEntity.isSelfMentioned(this.self.id);
+        return messageEntity.visible() && messageEntity.is_content() && messageEntity.isUserMentioned(this.self.id);
       });
   }
 
--- a/app/script/entity/message/ContentMessage.js
+++ b/app/script/entity/message/ContentMessage.js
@@ -108,9 +108,9 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
     }
   }
 
-  isSelfMentioned(selfId) {
+  isUserMentioned(userId) {
     return this.has_asset_text()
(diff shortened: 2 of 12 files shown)


app/script/message/MentionEntity.js
```diff
@@ -41,22 +41,11 @@ z.message.MentionEntity = class MentionEntity {
...
-
-  setSelfId(userId) {
-    this.selfId = userId;
   }
 
-  isSelfMentioned() {
-    if (this.selfId === undefined) {
-      throw new Error('Cannot determine if mention is a self mention if selfId is not set');
-    }
+  isSelfMentioned(selfId) {
```
gregor: This is no longer a check specific for self. We are in a mention anyway. Let's rename this to maybe `isUserWithId`

atomrc: Fair point

app/script/entity/Conversation.js
```diff
@@ -394,7 +394,9 @@ z.entity.Conversation = class Conversation {
     return this.unreadEvents()
       .slice()
       .reverse()
-      .find(messageEntity => messageEntity.visible() && messageEntity.is_content() && messageEntity.isSelfMentioned());
+      .find(messageEntity => {
+        return messageEntity.visible() && messageEntity.is_content() && messageEntity.isSelfMentioned(this.self.id);
```
atomrc: @gregor Unfortunately, there is not much we can do about this line break. 
The previous version was already 2 characters away from breaking and I don't think we want to rename our convention `messageEntity` for the sake of formatting.
Any other ideas?

app/script/entity/message/Text.js
```diff
@@ -49,12 +49,12 @@ z.entity.Text = class Text extends z.entity.Asset {
...
-  render() {
-    const message = z.util.renderMessage(this.text, this.mentions());
+  render(selfId) {
+    const message = z.util.renderMessage(this.text, selfId, this.mentions());
     return !this.previews().length ? z.media.MediaParser.renderMediaEmbeds(message, this.theme_color) : message;
   }
 
-  isSelfMentioned() {
-    return this.mentions().some(mentionEntity => mentionEntity.isSelfMentioned());
+  isSelfMentioned(selfId) {
```
gregor: This still isn't specific to the self user. => `isUserMentioned`.

app/script/entity/message/ContentMessage.js
```diff
@@ -108,11 +108,9 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
     }
   }
 
-  isSelfMentioned() {
+  isSelfMentioned(selfId) {
```
gregor: This isn't specific to self any longer => `isUserMentioned()`

app/script/notification/NotificationRepository.js
```diff
@@ -209,7 +212,7 @@ z.notification.NotificationRepository = class NotificationRepository {
     if (messageEntity.has_asset_text()) {
       for (const assetEntity of messageEntity.assets()) {
         if (assetEntity.is_text()) {
-          const notificationText = assetEntity.isSelfMentioned()
+          const notificationText = assetEntity.isUserMentioned(this.userRepository.self().id)
```
gregor: We use self three times already in this file and it causes a line to break. Fine to keep a reference by adding `this.selfUser = this.userRepository.self`

gregor:
LGTM

PR #4708, opened 2018-09-26
fix: Enable clicking links in messages with mentions (WEBAPP-5506)

gregor:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -610,14 +610,10 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
...
 
-  handleClickOnMessage(message, event) {
-    const hasMentions = message.mentions().length > 0;
-    if (!hasMentions) {
-      // need to return `true` because knockout will prevent default if we return anything else (including undefined)
-      return true;
-    }
+  handleClickOnMessage(messageEntity, event) {
+    const hasMentions = messageEntity.mentions().length;
     const mentionElement = event.target.closest('.message-mention');
```
atomrc: How about modifying this to 

```
const isClickOnMention = hasMentions && event.target.closest('.message-mention')
```

To avoid querying the DOM if we don't need to?

PR #4715, opened 2018-09-26
fix: Resize rich text properly (WEBAPP-5515)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -153,6 +153,10 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
         .join('');
     });
 
+    this.richTextInput.subscribe(() => {
+      $('.shadow-input').trigger('input');
+    });
```
gregor: Could be single line

app/page/template/content/conversation/input-bar.htm
```diff
@@ -29,7 +29,7 @@
                   data-uie-name="input-message"
                   dir="auto">
         </textarea>
-        <div class="shadow-input" data-bind="html: richTextInput, scrollSync: '#conversation-input-bar-text', resize: input" data-uie-name="input-message-rich-text"></div>
+        <div class="shadow-input" data-bind="html: richTextInput, scrollSync: '#conversation-input-bar-text', resize: input, resize_callback: scrollMessageList" data-uie-name="input-message-rich-text"></div>
```
bennycode: `resize_callback` -> `resizeCallback` if possible

app/script/view_model/content/InputBarViewModel.js
atomrc: Let's go for this solution for now. But please can we try to remove DOM querying and jQuery in the near future?

app/page/template/content/conversation/input-bar.htm
gregor: Then we have to rename in other places as well. This was missed during the renaming of binding and is decared in `CommonBindings.js`

gregor:
LGTM

app/script/view_model/content/InputBarViewModel.js
AndyLnd: I'm totally with you. This is a real band aid solution.

PR #4716, opened 2018-09-26
refactor: Remove localstorage handling from the Conversation Entity (WEBAPP-5489)

atomrc:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -305,11 +327,14 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       this.cancelMessageEditing();
       messageEntity.isEditing(true);
       this.editMessageEntity(messageEntity);
-      this.currentMentions(messageEntity.get_first_asset().mentions());
+
+      // clean mentions before setting text in order to prevent mentions offset recomputing
+      this.currentMentions.removeAll();
```
atomrc: This is an actual bug that was introduced since #4702 (the mentions where shifted because we set the mentions before setting the input)

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -227,6 +194,22 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       return conversationEntity.localMessageTimer() && !conversationEntity.hasGlobalMessageTimer();
     });
 
+    this.editedMessage = ko
```
atomrc: This is the new stuff around. The idea is we have a single source of truth to write on the localStorage. Whenever this changes (and we are not editing) we write the content to the localStorage

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -240,6 +223,45 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+    }
+    const previousSessionData = this._loadEditState(conversationEntity);
+    this.input(previousSessionData.text);
+    this.currentMentions(previousSessionData.mentions);
+  }
+
+  _saveEditState(conversationEntity, text, mentions) {
+    if (!this.isEditing()) {
+      // we only save state for newly written messages
+      const storageKey = `${z.storage.StorageKey.CONVERSATION.INPUT}|${conversationEntity.id}`;
```
atomrc: The key generation should be factorized in another function

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -227,6 +194,22 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+    this.editedMessage = ko
+      .pureComputed(() => {
+        const text = this.input();
+        const mentions = this.currentMentions();
+        return {mentions, text};
+      })
+      .extend({rateLimit: {method: 'notifyWhenChangesStop', timeout: 1}});
+
+    this.conversationEntity.subscribe(this.loadInitialStateForConversation.bind(this));
+    this.editedMessage.subscribe(message => {
```
gregor: I consider this naming confusing. The `editedMessage` calls `_saveEditState` to persist the state but not if `isEditing` is true. We need to clearly seperate the naming regarding editing an existing message (changing it) and drafting a new message.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -79,30 +78,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
-      },
-      write: value => {
-        if (this.isEditing()) {
-          return this.editInput(value);
-        }
-
-        if (this.conversationEntity()) {
-          const mentions = this.currentMentions();
-
-          this.conversationEntity().input({
```
atomrc, line 99: This was the main pain point that I wanted to remove. The circular dependency between `conv.input` and `inputBarViewModel.input`

app/script/view_model/content/InputBarViewModel.js
atomrc: Oh true. `draft` sounds great actually, I would go for `draftMessage`, WDYT?

app/script/view_model/content/InputBarViewModel.js
gregor: Is this related to reopening WEBAPP-5489?

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -129,7 +113,8 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
     this.richTextInput = ko.pureComputed(() => {
       const mentionAttributes = ' class="input-mention" data-uie-name="item-input-mention"';
-      const pieces = this.currentMentions
+      const {text, mentions} = this.draftMessage();
```
atomrc, line 116: Because this now depends on a slightly delayed `draftMessage` observable, we will compute the new rich text based on a consistent state

atomrc: Removes timing issues with setting the `input` or the `currentMentions` one before another

app/script/view_model/content/GiphyViewModel.js
```diff
@@ -134,9 +134,9 @@ z.viewModel.content.GiphyViewModel = class GiphyViewModel {
     }
   }
 
-  showGiphy() {
+  showGiphy(query) {
     this.sendingGiphyMessage = false;
-    this.query(this.conversationRepository.active_conversation().input().text);
+    this.query(query);
```
atomrc, line 139: I tested this scenario, works as expected!

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -227,6 +203,14 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
       return conversationEntity.localMessageTimer() && !conversationEntity.hasGlobalMessageTimer();
     });
 
+    this.conversationEntity.subscribe(this.loadInitialStateForConversation.bind(this));
+    this.draftMessage.subscribe(message => {
+      if (!this.conversationEntity()) {
+        return;
+      }
+      this._saveDraftState(this.conversationEntity(), message.text, message.mentions);
```
gregor: Not a favor of an early return on the negative case on this one. Positive approach is better as that is what we are looking for and not for what is not doing:

````
 if (this.conversationEntity()) {
  this._saveDraftState(this.conversationEntity(), message.text, message.mentions);
}
````

File: app/script/view_model/content/InputBarViewModel.js [hunk 9]
```diff
@@ -240,6 +224,49 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     });
   }
 
+  loadInitialStateForConversation(conversationEntity) {
+    this.conversationHasFocus(true);
+    this.pastedFile(null);
+    this.cancelMessageEditing();
+    if (!conversationEntity) {
+      return;
+    }
```
gregor, line 234: Early return is not really beneficial in these three lines either.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -306,13 +333,16 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
       this.editMessageEntity(messageEntity);
 
-      // clean mentions before setting text in order to prevent mentions offset recomputing
-      this.currentMentions.removeAll();
       this.input(messageEntity.get_first_asset().text);
+      this.currentMentions(
+        messageEntity
+          .get_first_asset()
+          .mentions()
+          .slice()
```
gregor: Let's pull tehse into a const and then put them into the observable. This is hard to read.

test/unit_tests/cache/CacheRepositorySpec.js
```diff
@@ -29,16 +29,13 @@ describe('z.cache.CacheRepository', () => {
...
-      conversation.input({mentions: [], text: 'Hello World!'});
-
       amplify.store(z.storage.StorageKey.AUTH.SHOW_LOGIN, true);
       amplify.store(TEMP_KEY, true);
     });
 
     it('deletes cached keys', () => {
       const deleted_keys = cache_repository.clearCache(false);
-      expect(deleted_keys.length).toBe(2);
+      expect(deleted_keys.length).toBe(1);
```
gregor: There is a reason for this test. We should not take the shortcut but rather add the conversation input key without using the conversation entity.

test/unit_tests/cache/CacheRepositorySpec.js
```diff
@@ -29,9 +29,7 @@ describe('z.cache.CacheRepository', () => {
     beforeEach(() => {
       cache_repository.clearCache();
 
-      const conversation = new z.entity.Conversation();
-      conversation.input({mentions: [], text: 'Hello World!'});
-
+      amplify.store(z.storage.StorageKey.CONVERSATION.INPUT, 'text');
```
gregor: This is not a "regular" storage key. We should fake one that will look like what we write into the storage.

gregor:
````
const inputStorageKey = `${z.storage.StorageKey.CONVERSATION.INPUT}|${z.util.createRandom....}`;
amplify.store(inputStorageKey, {});
````

gregor: Let's put a proper key and a proper input there. Just generate a random UUID using the util function. And place the object in the store and not just a string. => `{mentions: [], text: 'test'}` or whatever

PR #4724, opened 2018-09-27
chore: Log event details on mapping error

bennyn:


gregor:
We already have those in the object that is logged

bennycode:
But the stringified version of that object in a log file is just `Array(1)`.

PR #4728, opened 2018-09-27
test: Add test for multibyte characters in mentions

bennyn:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -445,11 +455,15 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
       const term = `${wordBeforeSelection.replace(/^@/, '')}${textInSelection}${wordAfterSelection}`;
       const startIndex = selectionStart - wordBeforeSelection.length;
-      this.editedMention({startIndex, term});
-    } else {
-      this.editedMention(undefined);
+      return {startIndex, term};
```
atomrc, line 458: WDYT of an early return that explicitly `return undefined` (instead of the implicit one)?

bennycode: Also thought of that. Especially since we mention it in the `@returns` section. Will update my PR.

gregor: That is now an explicit return but not and early one.

bennycode: Talked with @atomrc and we agreed on having it not early.

PR #4729, opened 2018-09-27
fix: Update selection state before computing new mentions state

atomrc:


atomrc:
Not a good idea. Will close it

PR #4733, opened 2018-09-27
fix: Sync scrolling on long texts (WEBAPP-5533)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -138,6 +138,18 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
         .join('');
     });
 
+    this.richTextInput.subscribe(() => {
+      const textarea = document.querySelector('#conversation-input-bar-text');
```
bennycode: We have the ID `#conversation-input-bar-text` now in three different places. Let's do a constant for it, so we can update it in a single place.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -489,15 +505,15 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   }
 
   handleMentionFlow() {
-    const textarea = document.querySelector('#conversation-input-bar-text');
+    const textarea = this.getTextArea();
     const {selectionStart, selectionEnd, value} = textarea;
```
gregor: I suppose we no longer need the const then and can simply do:

`const {selectionStart, selectionEnd, value} = this.getTextArea();`

PR #4735, opened 2018-09-27
fix: Editing from the message menu and adding char offsets mentions (WEBAPP-5541, WEBAPP-5542)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -341,6 +341,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
         .mentions()
         .slice();
       this.currentMentions(newMentions);
+      const inputElement = document.querySelector('#conversation-input-bar-text');
```
gregor: Seems like this should use the `getTextArea()` from #4733

AndyLnd: Once the other one is merged, I will change that

PR #4736, opened 2018-09-27
fix: Reset edited draft when gif is sent (WEBAPP-5538)

atomrc:


gregor:
Let's create a helper that reset this. We have the same two lines of code in 3 places now. All they do is reset the input which we can make more obvious.

atomrc:
Yeah, will do in a follow up PR I think

PR #4738, opened 2018-09-27
fix: Don't insert html when pressing shift+enter (WEBAPP-5543)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -113,10 +113,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
     this.richTextInput = ko.pureComputed(() => {
       const mentionAttributes = ' class="input-mention" data-uie-name="item-input-mention"';
-
-      const text = this.draftMessage().text.replace(/[\r\n]$/, '<br>&nbsp;');
-      const pieces = this.draftMessage()
-        .mentions.slice()
+      const {text, mentions} = this.draftMessage();
```
gregor: Let's do the ABC here

PR #4742, opened 2018-09-27
fix: ESC while suggestions are open cancels message (WEBAPP-5546)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -321,7 +321,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   }
 
   cancelMessageEditing() {
-    this.emojiInput.removeEmojiPopup();
```
marcoconti83, line 324: Is it OK to delete this line?

PR #4744, opened 2018-09-27
fix: a better solution to the cancle message problem (WEBAPP-5546)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -451,7 +451,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
         }
 
         case z.util.KeyboardUtil.KEY.ESC: {
-          if (this.pastedFile()) {
+          if (this.mentionSuggestions().length) {
+            this.endMentionFlow();
+          } else if (this.pastedFile()) {
             this.pastedFile(null);
           } else {
             this.cancelMessageEditing();
```
gregor, line 457: We still need to check that there is a message being edited. Otherwise this will clear the input if there is no suggestion and no pasted file.

PR #4746, opened 2018-09-27
fix: Use copy of mentions array when sending link preview (WEBAPP-5545)

gregor:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -612,7 +612,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
 
   sendMessage(messageText) {
     if (messageText.length) {
-      const mentionEntities = this.currentMentions();
+      const mentionEntities = this.currentMentions.slice();
```
gregor, line 615: Replicating the behavior from message editing in L621

PR #4757, opened 2018-09-28
fix: Cache DOM elements for performance (WEBAPP-5548)

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -233,6 +235,11 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     });
   }
 
+  _setElements(nodes) {
```
gregor: If you are calling it from the html it should not be prefixed with _

app/page/template/content/conversation/input-bar.htm
```diff
@@ -25,6 +25,7 @@
                              attr: {'placeholder': inputPlaceholder},
                              css: {'conversation-input-bar-text--accent': hasLocalEphemeralTimer()},
                              paste_file: onPasteFiles"
+
```
gregor: No need for empty line here

PR #4759, opened 2018-09-28
refactor: Run textarea resizing asynchronously (WEBAPP-5548)

atomrc:
Will prevent reflow operations to block the thread while typing in the textarea

### DIFF
--- a/app/script/view_model/bindings/CommonBindings.js
+++ b/app/script/view_model/bindings/CommonBindings.js
@@ -175,12 +175,8 @@ ko.bindingHandlers.resize = (function() {
           element,
           {
             event: {
-              focus() {
-                throttledResizeTextarea(element);
-              },
-              input() {
-                throttledResizeTextarea(element);
-              },
+              focus: () => throttledResizeTextarea(element),
+              input: () => throttledResizeTextarea(element),
             },
           },
           context
--- a/app/script/view_model/bindings/CommonBindings.js
+++ b/app/script/view_model/bindings/CommonBindings.js
@@ -132,7 +132,7 @@ ko.bindingHandlers.resize = (function() {
   let callback = null;
   let syncElement = null;
 
-  const resizeTextarea = _.throttle(element => {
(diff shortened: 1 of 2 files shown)


app/script/view_model/bindings/CommonBindings.js
```diff
@@ -175,10 +176,10 @@ ko.bindingHandlers.resize = (function() {
           {
             event: {
               focus() {
-                resizeTextarea(element);
+                throttledResizeTextarea(element);
```
gregor: Any reason why these are not single line arrow functions? `focus: () => throttledResizeTextarea(element),`

atomrc: The only reason I can think of is legacy.

app/script/view_model/bindings/CommonBindings.js
```diff
@@ -191,8 +188,9 @@ ko.bindingHandlers.resize = (function() {
       const params = ko.unwrap(valueAccessor()) || {};
       triggerValue = params.triggerValue;
       syncElement = document.querySelector(params.syncElement);
+      throttledResizeTextarea = _.throttle(resizeTextarea, 100, {leading: !params.delayedResize});
```
atomrc: There might a problem here, I should create the throttled version in the `init` function instead

gregor: Yeah, that actually makes a lot of sense.

app/script/view_model/bindings/CommonBindings.js
```diff
@@ -188,11 +186,10 @@ ko.bindingHandlers.resize = (function() {
     },
 
     update(element, valueAccessor) {
-      const params = ko.unwrap(valueAccessor()) || {};
-      triggerValue = params.triggerValue;
-      syncElement = document.querySelector(params.syncElement);
```
atomrc, line 193: Another perf improvement here, we now don't query the DOM on every single input (it is done when the binding is initiated)

PR #4761, opened 2018-09-28
fix: Prevent triggering user details from click on self mention (WEBAPP-5547)

gregor:


app/script/user/UserRepository.js
```diff
@@ -681,13 +681,17 @@ z.user.UserRepository = class UserRepository {
...
+   * @param {string} userId - User ID
    * @returns {Promise<z.entity.User>} Resolves with the matching user entity
    */
-  findUserById(user_id) {
-    for (const user_et of this.users()) {
-      if (user_et.id === user_id) {
-        return Promise.resolve(user_et);
+  findUserById(userId) {
+    if (!userId) {
+      return Promise.reject(new z.user.UserError(z.user.UserError.TYPE.MISSING_PARAMETER));
```
gregor: Previously any requestor of `get_user_by_id` would have to handle undefined as a potential return value. This should never have been the case.

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -613,7 +613,9 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
   handleClickOnMessage(messageEntity, event) {
     const hasMentions = messageEntity.mentions().length;
     const mentionElement = hasMentions && event.target.closest('.message-mention');
-    if (mentionElement) {
+    const userId = mentionElement.dataset.userId;
+
+    if (mentionElement && userId) {
```
gregor: A self mention does not contain a user id. The click should not trigger the show user details flow.

app/script/user/UserRepository.js
```diff
@@ -681,13 +681,17 @@ z.user.UserRepository = class UserRepository {
...
+    }
+
+    for (const userEntity of this.users()) {
```
atomrc: How about using `array.find` for that use case?

gregor: Ah. Yes that is definitely possible these day 🙈

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -613,7 +613,9 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
   handleClickOnMessage(messageEntity, event) {
     const hasMentions = messageEntity.mentions().length;
     const mentionElement = hasMentions && event.target.closest('.message-mention');
-    if (mentionElement) {
+    const userId = mentionElement.dataset.userId;
```
atomrc: Will crash if `mentionElement` is not found/defined/null..

atomrc: Should probably be moved inside the `if (mentionElement)` condition

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -613,7 +613,9 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
   handleClickOnMessage(messageEntity, event) {
     const hasMentions = messageEntity.mentions().length;
     const mentionElement = hasMentions && event.target.closest('.message-mention');
-    if (mentionElement) {
+    const userId = mentionElement && mentionElement.dataset.userId;
+
+    if (userId) {
       this.userRepository
         .get_user_by_id(mentionElement.dataset.userId)
```
atomrc: And then you can use `userId` here

gregor: Obviously...

PR #4765, opened 2018-09-28
chore: Remove “on” from device activation timestamp string

infotexture:
This better supports relative timestamps like “just now”.


gregor:
I suppose we also have this copy in the react world somewhere and should align.

PR #4768, opened 2018-09-28
fix: Skip mention confirmation on shift enter (WEBAPP-5530)

gregor:


app/script/components/mentionSuggestions.js
```diff
@@ -103,8 +106,14 @@ z.components.MentionSuggestions = class MentionSuggestions {
     this.onSelectionValidated(data, this.targetInput);
   }
 
-  validateSelection() {
+  validateSelection(keyboardEvent) {
+    const skipValidation = z.util.KeyboardUtil.isEnterKey(keyboardEvent) && keyboardEvent.shiftKey;
```
bennycode: Would maybe call it `isShiftEnter` instead of `skipValidation` as we have precise return values which indicate the invalid/skipped state.

PR #4772, opened 2018-10-01
fix: Do not share state between resize bindings (WEBAPP-5551)

atomrc:


app/script/view_model/bindings/CommonBindings.js
```diff
@@ -126,73 +126,63 @@ ko.virtualElements.allowedBindings.stopBinding = true;
 /**
  * Resize textarea according to the containing text.
  */
-ko.bindingHandlers.resize = (function() {
-  let lastHeight = null;
-  let triggerValue = null;
-  let callback = null;
```
atomrc, line 132: The problem was that this IIFE is only called once during the lifecycle of the app.
Which means that this statefull variables (lastHeight, callback, triggerValue) are shared across all the instances of this binding.
Most of the time it works but whenever we create a new textarea with this binding, everything goes wrong

app/script/view_model/bindings/CommonBindings.js
```diff
@@ -126,73 +126,63 @@ ko.virtualElements.allowedBindings.stopBinding = true;
...
+    }).bind(null, element);
+    const throttledResizeTextarea = _.throttle(resizeTextarea, 100, {leading: !params.delayedResize});
+
+    throttledResizeTextarea();
+    if (triggerValue === undefined) {
+      return ko.applyBindingsToNode(
+        element,
+        {
+          event: {
+            focus: throttledResizeTextarea,
```
atomrc, line 175: Here I bound the `resizeTextarea` function to the element at creation time (no need to pass it for those callbacks anymore)

PR #4773, opened 2018-10-01
feat: Change self mention highlight color (WEBAPP-5534)

gregor:


gregor:
https://github.com/wireapp/wire-theme/blob/master/app/style/wire-theme.less#L36

PR #4786, opened 2018-10-02
fix: Use the shadow input as the ground truth for textarea height

atomrc:
This will be a game changer in term of performance.

Resizing textarea to its content is pretty expensive because the way to do that is:
- resize the textarea to 0 ;
- compute the scrollHeight (trigger a complete reflow) ;
- set the new size to the computed scrollHeight (will trigger another reflow) ;

Because the visible content is displayed by a div, which do not need any reflow to know their size, the idea is to use this div to know how to size the textarea:
- when text change, we read the height of the div (basically a free operation) ;
- if the size is different than the size of the textarea, we set the textarea with the new size.

So now we trigger a reflow (and some costy operations) **only** when a new line is added to the textarea

### DIFF
--- a/app/script/view_model/bindings/CommonBindings.js
+++ b/app/script/view_model/bindings/CommonBindings.js
@@ -185,7 +185,7 @@ ko.bindingHandlers.heightSync = {
 
     // initial resize
     resizeTarget();
-    const valueSubscription = triggerValue.subscribe(resizeTarget);
+    const valueSubscription = triggerValue.subscribe(() => window.requestAnimationFrame(resizeTarget));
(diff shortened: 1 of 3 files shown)


app/script/view_model/bindings/CommonBindings.js
```diff
@@ -130,64 +130,69 @@ ko.bindingHandlers.resize = {
   init(element, valueAccessor, allBindings, data, context) {
     const params = ko.unwrap(valueAccessor()) || {};
 
-    const callback = params.callback;
-    const syncElement = document.querySelector(params.syncElement);
```
atomrc, line 134: We can safely remove those as this binding is now only used for a simple textarea

PR #4789, opened 2018-10-02
feat: Direct links to team settings feature pages (WEBAPP-5567)

gregor:


app/page/template/content/preferences-account.htm
```diff
@@ -123,13 +123,13 @@
           <hr class="preferences-separator">
 
           <!-- ko if: isTeamManager() -->
-            <div class="preferences-link text-theme" data-bind="click: clickOnManage, l10n_text: z.string.preferencesAccountManageTeam" data-uie-name="do-manage-team"></div>
+            <div class="preferences-link text-theme" data-bind="click: clickToOpenManageTeam, l10n_text: z.string.preferencesAccountManageTeam" data-uie-name="do-manage-team"></div>
```
atomrc: I don't think I understand the semantic of this naming.
Why `clickToOpen...`? I sounds like a text we would display to a user to know what will happen when he clicks on the button.

I would generally for a more imperative semantic that explain what the function is going to do.
I mean, we already know that it's bound to the click event because it's bound on the `click` listener.

I would go for a `click: openManageTeam`. WDYT?

gregor: Yes and no. The `clickTo...` naming shows that this is used in knockout click events when looking at the view model code. Those events always have a certain set of parameters to expect. Avoids breaking the method when reusing it outside of that scope.

PR #4790, opened 2018-10-02
chore: Update yarn lock with new integrity check

gregor:


.travis.yml
```diff
@@ -29,6 +29,8 @@ node_js:
 
 # http://docs.travis-ci.com/user/gui-and-headless-browsers/
 before_install:
+  - curl https://yarnpkg.com/install.sh -sSfL | bash -s --
```
atomrc: Isn't this going to slow down every build on travis since it will always install this?

gregor: Yes, Copied the behavior from `wire-web-packages`. Just saw that the new version is used by @dependabot without those lines in the travis file as well. Let's try without.

PR #4793, opened 2018-10-02
fix: Use object for excluded emojis to avoid number replacements (WEBAPP-5536)

atomrc:


app/script/util/StringUtil.js
```diff
@@ -34,13 +34,13 @@ z.util.StringUtil = {
    * @param {boolean} fromStart=false - should the query match the string from the beginning of the string
    * @returns {boolean} does the string matches the query
    */
-  compareTransliteration: (string, query, excludedChars = [], fromStart = false) => {
+  compareTransliteration: (string, query, excludedChars = {}, fromStart = false) => {
```
gregor, line 37: Please update JSDoc

PR #4799, opened 2018-10-04
feat: Open notification settings from context menu (WEBAPP-5559)

AndyLnd:
This PR doesn't take care of the shortcuts yet. It has to be determined how they should behave.

### DIFF
--- a/app/script/notification/NotificationRepository.js
+++ b/app/script/notification/NotificationRepository.js
@@ -570,7 +570,12 @@ z.notification.NotificationRepository = class NotificationRepository {
 
     const containsSelfMention = messageEntity.is_content() && messageEntity.isUserMentioned(this.selfUser().id);
     if (containsSelfMention) {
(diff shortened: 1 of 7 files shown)


app/script/notification/NotificationRepository.js
```diff
@@ -570,7 +570,12 @@ z.notification.NotificationRepository = class NotificationRepository {
 
     const containsSelfMention = messageEntity.is_content() && messageEntity.isUserMentioned(this.selfUser().id);
     if (containsSelfMention) {
-      return () => amplify.publish(z.event.WebApp.CONVERSATION.SHOW, conversationEntity, messageEntity, true);
+      return () => {
+        amplify.publish(z.event.WebApp.CONVERSATION.SHOW, conversationEntity, {
+          messageEntity,
```
gregor: I will change the naming of the options object slightly as it is not clear what the messageEntity does.

PR #4808, opened 2018-10-04
fix: Set a status to ping messages (WEBAPP-5544)

atomrc:


app/script/conversation/EventMapper.js
```diff
@@ -200,7 +200,7 @@ z.conversation.EventMapper = class EventMapper {
     messageEntity.type = type;
     messageEntity.version = version || 1;
 
-    if (messageEntity.is_content()) {
+    if (messageEntity.is_content() || messageEntity.is_ping()) {
```
gregor, line 203: I suppose we do this somewhere else as well. When sending a message. Does the check include pings there as well?

atomrc: For sent messages, the status is set to `SENT` once it has hit the backend (no distinction).
So we are already good for the sending side

PR #4809, opened 2018-10-04
feat: Adjust window title to self mentions (WEBAPP-5558)

ffflorian:


app/script/entity/Conversation.js
```diff
@@ -193,6 +193,12 @@ z.entity.Conversation = class Conversation {
       }).length;
     });
 
+    this.unreadSelfMentionsCount = ko.pureComputed(() => {
+      return this.unreadEvents().filter(messageEntity => {
+        return messageEntity.is_content() && messageEntity.isUserMentioned(this.self.id);
+      }).length;
+    });
```
gregor, line 228: This is all of computational overhead. Let's use the unread events observable and set a flag instead of iterating of the array again for every new message. We don't need the number anyway. A boolean will do.

app/script/entity/Conversation.js
```diff
@@ -167,36 +167,42 @@ z.entity.Conversation = class Conversation {
     this.hasActiveCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().isActiveState() : false));
     this.hasJoinableCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().canJoinState() : false));
 
+    this.unreadEventsCount = ko.pureComputed(0);
+    this.unreadMessagesCount = ko.pureComputed(0);
+    this.hasUnreadSelfMention = ko.observable(false);
```
gregor: All of need to be observables. You cannot set a pureComputed

app/script/entity/Conversation.js
```diff
@@ -167,36 +167,42 @@ z.entity.Conversation = class Conversation {
...
+
     this.unreadEvents = ko.pureComputed(() => {
```
gregor: As discussed: For this to trigger without a subscriber we need to turn it into a bare computed.

app/script/view_model/WindowTitleViewModel.js
```diff
@@ -61,10 +61,9 @@ z.viewModel.WindowTitleViewModel = class WindowTitleViewModel {
...
 
             const notifySelfMentions = conversationEntity.showNotificationsOnlyMentions();
-            const selfMentions = conversationEntity.unreadSelfMentionsCount();
 
             if (notifySelfMentions) {
-              return selfMentions > 0;
+              return conversationEntity.hasUnreadSelfMention();
             }
 
             return conversationEntity.unreadMessagesCount() > 0 || conversationEntity.hasJoinableCall();
```
gregor: Ternary

app/script/entity/Conversation.js
ffflorian: They are already observables..?

ffflorian: Oh, I think you commented before I added my last commit?

app/script/view_model/WindowTitleViewModel.js
```diff
@@ -54,9 +54,17 @@ z.viewModel.WindowTitleViewModel = class WindowTitleViewModel {
...
-            const isIgnored = conversationEntity.is_request() || conversationEntity.is_muted();
-            const hasActions = conversationEntity.unreadMessagesCount() || conversationEntity.hasJoinableCall();
-            return hasActions && !isIgnored;
+            const isIgnored = conversationEntity.is_request() || conversationEntity.showNotificationsNothing();
+
+            if (isIgnored) {
+              return false;
+            }
+
+            const notifySelfMentions = conversationEntity.showNotificationsOnlyMentions();
```
gregor: No need for this const

app/script/entity/Conversation.js
```diff
@@ -167,30 +167,42 @@ z.entity.Conversation = class Conversation {
     this.hasActiveCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().isActiveState() : false));
     this.hasJoinableCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().canJoinState() : false));
 
-    this.unreadEvents = ko.pureComputed(() => {
+    this.unreadEventsCount = ko.observable(0);
+    this.unreadMessagesCount = ko.observable(0);
+    this.hasUnreadSelfMention = ko.observable(false);
+
+    this.unreadEvents = ko.computed(() => {
```
atomrc: Let's not create a `computed` here. 
As a rule of thumb, you can keep in mind that `subscribe` or `computed` should not live in an entity.
They should live in the upper layers of the app as much as possible (viewmodel for example).

The problem with subscribing in an entity is that we don't have the mechanism to tear down the subscription, and it will live on until the browser is reloaded.

I would suggest that unreadEvents is a `pureComputed` and that `unreadEventsCount`, `unreadMessageCount` and `hasUnreadSelfMention` are also `pureComputed` that depends on `unreadEvents`.

app/script/view_model/WindowTitleViewModel.js
```diff
@@ -54,9 +54,17 @@ z.viewModel.WindowTitleViewModel = class WindowTitleViewModel {
         const unreadConversations = this.conversationRepository
           .conversations_unarchived()
           .filter(conversationEntity => {
-            const isIgnored = conversationEntity.is_request() || !conversationEntity.showNotificationsEverything();
-            const hasActions = conversationEntity.unreadMessagesCount() || conversationEntity.hasJoinableCall();
-            return hasActions && !isIgnored;
+            const {unreadMessages, unreadSelfMentions} = conversationEntity.unreadState();
+
+            const isIgnored = conversationEntity.is_request() || conversationEntity.showNotificationsNothing();
```
atomrc, line 59: There might be a logical issue here 
Those two are not equivalent:

`!conversationEntity.showNotificationsEverything()`
`conversationEntity.showNotificationsNothing()`

atomrc: Maybe this is an intended change?

ffflorian: This is an intended change to return early if all notifications are disabled. Correct me if I'm wrong but I see that as logical.

app/script/entity/Conversation.js
```diff
@@ -176,30 +176,46 @@ z.entity.Conversation = class Conversation {
     this.hasActiveCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().isActiveState() : false));
     this.hasJoinableCall = ko.pureComputed(() => (this.hasLocalCall() ? this.call().canJoinState() : false));
 
-    this.unreadEvents = ko.pureComputed(() => {
+    this.unreadState = ko.pureComputed(() => {
       const unreadEvents = [];
       const messages = this.messages();
+      const unreadSelfMentions = [];
+      const unreadTextMessages = [];
```
gregor: These are not necessarily test messages. The name is confusing. Files, images, videos would also fall in this category.

app/script/entity/Conversation.js
```diff
@@ -384,12 +401,9 @@ z.entity.Conversation = class Conversation {
...
   getFirstUnreadSelfMention() {
-    return this.unreadEvents()
-      .slice()
-      .reverse()
-      .find(messageEntity => {
-        return messageEntity.visible() && messageEntity.is_content() && messageEntity.isUserMentioned(this.self.id);
-      });
+    return this.unreadState()
+      .selfMentions.slice()
+      .reverse();
```
gregor: This should return a single message only

PR #4815, opened 2018-10-05
feat: Differentiate mute behavior based on account type

gregor:


app/script/view_model/ListViewModel.js
```diff
@@ -290,15 +291,31 @@ z.viewModel.ListViewModel = class ListViewModel {
   onContextMenu(conversationEntity, event) {
     const entries = [];
 
-    const canSetNotifications = !conversationEntity.is_request() && !conversationEntity.removed_from_conversation();
-    if (canSetNotifications) {
-      const notificationsShortcut = z.ui.Shortcut.getShortcutTooltip(z.ui.ShortcutType.NOTIFICATIONS);
+    const isMutable = !conversationEntity.is_request() && !conversationEntity.removed_from_conversation();
```
atomrc: I see this boolean operation (at least) twice (also in `ConversationDetailsViewModel`). Maybe would be a good idea to factorize it in a single place (either a util or on the entity).

gregor: Good point

PR #4823, opened 2018-10-05
refactor: Migrate conversation archiving to camelCase

gregor:


app/script/entity/Conversation.js
```diff
@@ -153,12 +153,6 @@ z.entity.Conversation = class Conversation {
     this.isLeavable = ko.pureComputed(() => this.is_group() && !this.removed_from_conversation());
     this.isMutable = ko.pureComputed(() => !this.is_request() && !this.removed_from_conversation());
 
-    this.removed_from_conversation.subscribe(is_removed => {
-      if (!is_removed) {
-        return this.archived_state(false);
```
atomrc: This is a behavior change, right? Else, I don't see where it is handled now.

gregor: Thanks for pointing out. This was not supposed to be part of this PR. Will revert.

PR #4824, opened 2018-10-05
feat: Adjust unarchive behavior to notification states (WEBAPP-5566)

gregor:


test/unit_tests/entity/ConversationSpec.js
```diff
@@ -794,44 +794,115 @@ describe('Conversation', () => {
...
-      expect(conversation_et.shouldUnarchive()).toBeFalsy();
+    it('returns false if conversation is not archived', () => {
+      conversationEntity.archivedState(false);
+      expect(conversationEntity.shouldUnarchive()).toBe(false);
+    });
 
-      conversation_et.last_event_timestamp(time + 100);
-      expect(conversation_et.shouldUnarchive()).toBeFalsy();
+    it('returns false if conversation is in no notification state', () => {
+      conversationEntity.notificationState(z.conversation.NotificationSetting.STATE.NOTHING);
```
atomrc, line 860: Would be good to still add a message in the conversation (let say the highest priority type of message, namely `mention`), just to check that the test is not successful only because there are no messages in there

PR #4838, opened 2018-10-08
feat: Map muted state boolean based on account type (WEBAPP-5574)

gregor:


app/script/conversation/ConversationRepository.js
```diff
@@ -339,23 +339,28 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
-      .then(([local_conversations, remote_conversations = []]) => {
-        if (remote_conversations.length) {
-          const conversations = this.conversationMapper.mergeConversation(local_conversations, remote_conversations);
-          return this.conversation_service.save_conversations_in_db(conversations);
+    return Promise.all([this.conversation_service.load_conversation_states_from_db(), remoteConversationsSromise])
+      .then(([localConversations, remoteConversations = []]) => {
+        const shouldMigrateData = localConversations.some(({muted_state}) => muted_state !== undefined);
+        if (shouldMigrateData) {
+          localConversations = this.conversationMapper.migrateConversationsData(localConversations, this.isTeam());
         }
```
gregor, line 352: We check for the presence of some outdated data. Not sure if this is the best way to do it. It should only be true once once every machine. Maybe it is reliable enough to check the first object in the array instead of doing a some on the full array every time. WDYT?

app/script/conversation/ConversationMapper.js
```diff
@@ -99,16 +99,17 @@ z.conversation.ConversationMapper = class ConversationMapper {
    *
    * @param {Array} conversationsData - Conversation data
    * @param {number} [timestamp=1] - Initial timestamp for conversation
+   * @param {boolean} [isTeam] - Type of account for state mapping
    * @returns {Array<Conversation>} Mapped conversation entities
    */
-  mapConversations(conversationsData, timestamp = 1) {
+  mapConversations(conversationsData, timestamp = 1, isTeam) {
```
ffflorian: Let's set a default value for `isTeam`, otherwise we'll have a required value after an optional value.

gregor: Check the JSDoc. It is optional. It will work without setting it. The default is set where we need it further down the chain.

gregor: Since we only call it from one place and the value is set. I will change the order and make it non optional. That should also solve your comment.

app/script/conversation/ConversationRepository.js
ffflorian: I like this solution. Two things:
`remoteConversationsSromise` => `remoteConversationsPromise`
`muted_state !== undefined` => `typeof muted_state !== 'undefined'`

gregor: The second one should not pass the linter. Unless we have changed the rules: https://eslint.org/docs/rules/valid-typeof

atomrc: The only problem I can imagine with only checking the first item would be, if the migration crashes at some point, then we would have conversations that are never migrated.

It also depends on how often do we call this method. If it's only at app load time and then never again during the lifecycle of the app, I guess it's totally acceptable to do a `some`...

atomrc: Even if the array has thousands of items, we are talking about a < 10ms optimization I believe.
I think I would go for the safer solution here that could prevent very unexpected bugs

gregor:
I will go ahead, merge this implementation for now and think about a solution that will map on the getter side in order to forgo the migration altogether.

PR #4841, opened 2018-10-08
fix: Remove location from client activation info (WEBAPP-5518)

AndyLnd:


app/script/view_model/content/PreferencesDevicesViewModel.js
```diff
@@ -90,11 +70,8 @@ z.viewModel.content.PreferencesDevicesViewModel = class PreferencesDevicesViewMo
 
   updateDeviceInfo() {
     if (this.currentClient() && !this.localFingerprint().length) {
-      const {location, time} = this.currentClient();
+      const {time} = this.currentClient();
```
atomrc: You probably don't need a tmp variable for that. The two lines should fit into one

app/script/view_model/content/PreferencesDeviceDetailsViewModel.js
```diff
@@ -37,56 +37,32 @@ z.viewModel.content.PreferencesDeviceDetailsViewModel = class PreferencesDeviceD
     this.conversationRepository = repositories.conversation;
     this.cryptographyRepository = repositories.cryptography;
     this.logger = new z.util.Logger('z.viewModel.content.PreferencesDeviceDetailsViewModel', z.config.LOGGER.OPTIONS);
-    this.locationRepository = repositories.location;
```
atomrc, line 40: Is the locationRepo used somewhere else? Maybe we can totally remove it?

AndyLnd: It is still used.

gregor: It should not be used anywhere. We can remove geolocation and CSP headers related to it.

gregor:
Do we display the location on the auth pages as well?

AndyLnd:
We don't display it on the auth pages.

gregor:
Why against the edge branch though? This is unrelated to the new feature and should show up in the commit history after a squash merge from the edge branch.

PR #4843, opened 2018-10-08
test: Bootstrap ConversationCellState tests

atomrc:
These are not exhaustive tests.
It covers the basic scenarios for 1:1 conversations but is missing a lot of combinations.

### DIFF
--- a/app/script/conversation/ConversationCellState.js
+++ b/app/script/conversation/ConversationCellState.js
@@ -323,7 +323,7 @@ z.conversation.ConversationCellState = (() => {
       }
     },
     icon: () => z.conversation.ConversationStatusIcon.UNREAD_MESSAGES,
-    match: conversationEntity => conversationEntity.unreadState().allEvents.length > 0,
+    match: conversationEntity => conversationEntity.unreadState().allMessages.length > 0,
   };
 
   const _getStateUserName = {
--- a/test/unit_tests/conversation/ConversationCellStateSpec.js
+++ b/test/unit_tests/conversation/ConversationCellStateSpec.js
@@ -47,7 +47,7 @@ describe('z.conversation.ConversationCellState', () => {
     });
   });
 
-  describe('Second line description for 1:1 conversation', () => {
+  describe('Second line description for conversations', () => {
     const defaultUnreadState = {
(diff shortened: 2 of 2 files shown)


app/script/conversation/ConversationCellState.js
```diff
@@ -323,7 +323,7 @@ z.conversation.ConversationCellState = (() => {
       }
     },
     icon: () => z.conversation.ConversationStatusIcon.UNREAD_MESSAGES,
-    match: conversationEntity => conversationEntity.unreadState().allEvents.length > 0,
+    match: conversationEntity => conversationEntity.unreadState().allMessages.length > 0,
```
atomrc, line 326: This generator never consumes the `allEvents` array (only the `allMessages`). So I think it's semantically more correct to use it as a condition to start the computation.

I guess, right now, we can have a scenario (where we only have system events for example), where we go through the computation loop for nothing.

test/unit_tests/conversation/ConversationCellStateSpec.js
```diff
@@ -0,0 +1,136 @@
...
+        expect(state).toEqual(expectedOne2One);
+      });
+    });
+
+    conversation.isGroup = () => true;
+    tests.forEach(({description, expected, unreadState}) => {
+      const expectedGroup = expected.group || expected;
+      conversation.unreadState = () => unreadState;
+      const state = conversationCellState.generate(conversation);
+      it(`${description} (group)`, () => {
```
gregor, line 133: Can we add an empty line before the `it` block please?

test/unit_tests/conversation/ConversationCellStateSpec.js
```diff
@@ -0,0 +1,136 @@
...
+        }),
+      },
+    ];
+
+    conversation.isGroup = () => false;
+    tests.forEach(({description, expected, unreadState}) => {
+      const expectedOne2One = expected.one2one || expected;
+      conversation.unreadState = () => unreadState;
+      const state = conversationCellState.generate(conversation);
+      it(`${description} (1:1)`, () => {
```
gregor, line 122: Can we add an empty line before the `it` block please?

PR #4845, opened 2018-10-08
fix: Ignore leave events when parsing missing users in a conversation (WEBAPP-5439)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -2780,6 +2780,12 @@ z.conversation.ConversationRepository = class ConversationRepository {
       return conversationEntity;
     }
 
+    const isLeaveEvent = eventJson.type === z.event.Backend.CONVERSATION.MEMBER_LEAVE;
+    if (isLeaveEvent) {
+      // we ignore leave event as they come from the person leaving (thus not being in the conversation participants anymore)
```
ffflorian:
```
// we ignore the leave event as it comes from the person leaving (thus not being in list of conversation participants anymore)
```

app/script/conversation/ConversationRepository.js
```diff
@@ -2780,6 +2780,12 @@ z.conversation.ConversationRepository = class ConversationRepository {
       return conversationEntity;
     }
 
+    const isLeaveEvent = eventJson.type === z.event.Backend.CONVERSATION.MEMBER_LEAVE;
+    if (isLeaveEvent) {
+      // we ignore leave events as they come from the person leaving (thus not being in the list of conversation's participants anymore)
+      return conversationEntity;
```
gregor, line 2786: Please recheck that. This is only true if sender and leaving user are identical.

PR #4847, opened 2018-10-08
fix: Use map on input data when migrating

gregor:


app/script/conversation/ConversationMapper.js
```diff
@@ -401,7 +401,7 @@ z.conversation.ConversationMapper = class ConversationMapper {
   }
 
   migrateConversationsData(conversationsDate, isProAccount) {
-    return this.migrateConversationsData.map(conversationData => {
+    return conversationsDate.map(conversationData => {
```
atomrc: you probably meant `conversationDatA`?

atomrc: And the name of the parameter should be changed also

PR #4848, opened 2018-10-08
fix: Only ignore leave events that are from the user leaving (WEBAPP-5439)

atomrc:


app/script/conversation/ConversationRepository.js
```diff
@@ -2780,19 +2780,26 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
     const {from: sender, id, type, time} = eventJson;
 
     if (sender) {
       const allParticipantIds = conversationEntity.participating_user_ids().concat(this.selfUser().id);
       const isFromUnknownUser = !allParticipantIds.includes(sender);
 
       if (isFromUnknownUser) {
+        const leavingEventTypes = [
+          z.event.Backend.CONVERSATION.MEMBER_LEAVE,
+          z.event.Client.CONVERSATION.TEAM_MEMBER_LEAVE,
```
atomrc, line 2792: As seen here, the `TEAM_MEMBER_LEAVE` event is also fed with the leaving user id
https://github.com/wireapp/wire-webapp/blob/5374df02b545e02794834d14fdf01edb089b7416/app/script/conversation/EventBuilder.js#L189-L201

So we also need to account for this event.

PR #4851, opened 2018-10-09
chore: Remove support for JSX

bennyn:
We use TSX instead of JSX now and ESLint doesn't work for TSX, so there is no need to add TSX but also no need to keep JSX.

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,7 +163,7 @@
     }
   },
   "lint-staged": {
-    "*.{js,jsx}": [
+    "*.js": [
       "eslint --fix",
       "git add"
     ],


bennycode:
@ffflorian There shouldn't be any `.jsx` file left but yes, we can remove all other hints to `jsx` as well.

ffflorian:
* [`app/script/auth/page/Root.test.jsx`](https://github.com/wireapp/wire-webapp/blob/dev/app/script/auth/page/Root.test.jsx)
* [`app/script/auth/page/TeamName.test.jsx`](https://github.com/wireapp/wire-webapp/blob/dev/app/script/auth/page/TeamName.test.jsx)
* [`app/script/auth/component/AccountForm.test.jsx`](https://github.com/wireapp/wire-webapp/blob/dev/app/script/auth/component/AccountForm.test.jsx)

bennycode:
Ditching this for now because we still have test files that are of type JSX. We need to convert those first before this PR here can have a chance to get merged.

PR #4860, opened 2018-10-09
fix: Hide options header if no options are available

gregor:


app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -132,32 +132,32 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
...
-
     this.showActionLeave = ko.pureComputed(() => this.activeConversation() && this.activeConversation().isLeavable());
 
     this.showActionMute = ko.pureComputed(() => {
       return this.activeConversation() && this.activeConversation().isMutable() && !this.isTeam();
     });
 
-    this.showActionNotificationSettings = ko.pureComputed(() => {
+    this.showOptionGuests = ko.pureComputed(() => {
+      return this.isActiveGroupParticipant() && this.activeConversation().inTeam();
```
atomrc, line 142: Is it safe to remove the `&& this.activeConversation()`?

gregor: I believe so. The `isActiveGroupParticipant` checks for it as well and should thus never return true if active conversation is not set.

atomrc: Ah ok, it's included in `isActiveGroupParticipant`

app/script/view_model/panel/ConversationDetailsViewModel.js
```diff
@@ -132,32 +132,32 @@ z.viewModel.panel.ConversationDetailsViewModel = class ConversationDetailsViewMo
...
+    this.showOptionNotifications1To1 = ko.pureComputed(() => {
+      return this.hasAdvancedNotifications() && !this.activeConversation().isGroup();
     });
 
-    this.showActionTimedMessages = this.isActiveGroupParticipant;
+    this.showOptionTimedMessages = this.isActiveGroupParticipant;
 
     this.showSectionOptions = ko.pureComputed(() => {
-      return this.showActionGuestOptions() || this.showActionNotificationSettings() || this.showActionTimedMessages();
+      return this.showOptionGuests() || this.showOptionNotificationsGroup() || this.showOptionTimedMessages();
```
gregor, line 160: Fix to hide the section header

PR #4873, opened 2018-10-10
refactor: Simplify error classes

gregor:


app/script/auth/AuthError.js
```diff
@@ -23,37 +23,33 @@ window.z = window.z || {};
...
-      case AuthError.TYPE.MULTIPLE_TABS:
-        this.message = 'Cannot open in multiple tabs simultaneously';
-        break;
-      case AuthError.TYPE.PRIVATE_MODE:
-        this.message = 'Unsupported Private Mode';
-        break;
-      default:
-        this.message = 'Unknown AuthError';
-    }
+    this.message = message || AuthError.MESSAGE[this.type] || AuthError.MESSAGE.UNKNOWN;
```
atomrc: I see this pattern quite some times in this PR. Maybe having a super class that does that could be a good idea. 
How about a `WireError` and make all of those inherit from it?
WDYT?

gregor: Yeah. Sounds like a good idea.

app/script/error/AccessTokenError.js
```diff
@@ -24,8 +24,7 @@ window.z.error = z.error || {};
 
 z.error.AccessTokenError = class AccessTokenError extends z.error.BaseError {
   constructor(type, message) {
-    const errorName = 'AccessTokenError';
-    super(errorName, type, message);
+    super(type, message);
```
atomrc: In those cases, you can remove the `constructor` at all

app/script/error/BaseError.js
```diff
@@ -28,7 +28,12 @@ z.error.BaseError = class BaseError extends Error {
 
     this.name = this.constructor.name;
     this.stack = new Error().stack;
-    this.type = type || BaseError.TYPE.UNKNOWN;
+
+    const specificTypes = Object.values(z.error[this.name].TYPE);
+    const knownTypes = Object.vales(BaseError.TYPE).concat(specificTypes);
```
atomrc: Typo : Object.vale

PR #4893, opened 2018-10-15
chore: Migrate /tracking to camelCase

gregor:


app/script/tracking/EventTrackingRepository.js
```diff
@@ -244,14 +242,14 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
 
-  _is_domain_allowed_for_tracking() {
+  _isDomainAllowedForTracking() {
     if (!z.util.URLUtil.getParameter(z.auth.URLParameter.TRACKING)) {
-      for (const domain of EventTrackingRepository.CONFIG.USER_ANALYTICS.DISABLED_DOMAINS) {
+      return !EventTrackingRepository.CONFIG.USER_ANALYTICS.DISABLED_DOMAINS.some(domain => {
         if (z.util.StringUtil.includes(window.location.hostname, domain)) {
           this.logger.debug(`Tracking is disabled for domain '${window.location.hostname}'`);
-          return false;
+          return true;
```
ffflorian, line 250: The `for..of` solution was better because it returned as soon as there was a disabled domain. This change lets the code always run through the whole array.

app/script/tracking/Helpers.js
```diff
@@ -25,59 +25,17 @@ window.z.tracking = z.tracking || {};
...
-          return z.tracking.attribute.MessageType.PING;
-        case !message_et.has_asset():
-          return z.tracking.attribute.MessageType.FILE;
-        case !message_et.has_asset_image():
-          return z.tracking.attribute.MessageType.IMAGE;
-        case !message_et.has_asset_location():
-          return z.tracking.attribute.MessageType.LOCATION;
-        default:
-          return z.tracking.attribute.MessageType.TEXT;
-      }
```
ffflorian, line 60: This function is gone now?

app/script/tracking/EventTrackingRepository.js
gregor: `Array.prototype.some()` does exactly that as well.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

app/script/tracking/Helpers.js
gregor: It was not used anywhere in the code

app/script/tracking/EventTrackingRepository.js
ffflorian: Oh. TIL :sun_with_face:

PR #4894, opened 2018-10-15
fix: Don't double-escape names in system messages (WEBAPP-5601)

AndyLnd:


app/script/entity/message/MemberMessage.js
```diff
@@ -83,7 +83,7 @@ z.entity.MemberMessage = class MemberMessage extends z.entity.SystemMessage {
 
     this.senderName = ko.pureComputed(() => {
       const isTeamMemberLeave = this.type === z.event.Client.CONVERSATION.TEAM_MEMBER_LEAVE;
-      return isTeamMemberLeave ? this.name() : z.util.SanitizationUtil.getFirstName(this.user());
+      return isTeamMemberLeave ? this.name() : z.util.SanitizationUtil.getFirstName(this.user(), undefined, true);
```
gregor: Let's be specific and pass in the default declension value as it is the one we expect anyway in this case as we set the third param:

````
return isTeamMemberLeave ? this.name() : z.util.SanitizationUtil.getFirstName(this.user(), z.string.Declension.NOMINATIVE, true);
````

PR #4905, opened 2018-10-16
feat: Make suggestion popup full width (WEBAPP-5517)

AndyLnd:
This introduces simplebar (not dead) as an alternative to antiscroll (dead). It is used for now on the suggestions popup and achieves what antiscroll couldn't: allow me to make it full width.
Simplebar will be explored further in coming PRs and eventually replace antiscroll everywhere.


app/script/components/mentionSuggestions.js
```diff
@@ -56,7 +56,7 @@ z.components.MentionSuggestions = class MentionSuggestions {
     });
 
     this.shouldUpdateScrollbar = ko.pureComputed(() => this.suggestions()).extend({notify: 'always', rateLimit: 100});
-    this.shouldUpdateScrollbar.subscribe(() => this.setWrapperSize(), null, 'beforeChange');
+    /*this.shouldUpdateScrollbar.subscribe(() => this.setWrapperSize(), null, 'beforeChange');
```
gregor: Let's remove code that has been commented out

app/script/components/mentionSuggestions.js
```diff
@@ -117,24 +117,26 @@ z.components.MentionSuggestions = class MentionSuggestions {
   }
 
   updateScrollPosition(selectedNumber) {
-    const suggestionList = document.querySelector('.mention-suggestion-list');
-    if (!suggestionList) {
+    const scrollContent = document.querySelector(
+      '.conversation-input-bar-mention-suggestion > .simplebar-scroll-content'
```
gregor: Put in const if too long for single line: 3 => 2

app/style/common/mixins.less
```diff
@@ -242,10 +242,6 @@
   visibility: hidden;
 }
 
-.visible {
-  visibility: visible;
-}
-
```
gregor, line 248: Why removing a mixin?

package.json
```diff
@@ -60,6 +60,7 @@
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
     "sdp-transform": "2.4.1",
+    "simplebar": "3.0.0-beta.5",
```
gregor, line 63: Beta?

app/style/common/mixins.less
AndyLnd: It is not used anywhere and actually clashes with simplebar.

app/script/components/mentionSuggestions.js
```diff
@@ -117,24 +107,25 @@ z.components.MentionSuggestions = class MentionSuggestions {
   }
 
   updateScrollPosition(selectedNumber) {
-    const suggestionList = document.querySelector('.mention-suggestion-list');
-    if (!suggestionList) {
+    const scrollContentSelector = '.conversation-input-bar-mention-suggestion > .simplebar-scroll-content';
```
atomrc: Let's try not to depend on classes we don't manage ourselves. Here we create an implicit dependency to `simplebar`.
There might be some cleaner way to get the scroll height

package.json
AndyLnd: Yes, this has performance improvements over the latest stable version.
https://github.com/Grsmto/simplebar/releases/tag/simplebar%403.0.0-beta.0

app/script/components/mentionSuggestions.js
```diff
@@ -56,16 +57,6 @@ z.components.MentionSuggestions = class MentionSuggestions {
     });
 
     this.shouldUpdateScrollbar = ko.pureComputed(() => this.suggestions()).extend({notify: 'always', rateLimit: 100});
```
atomrc: Can you try `ko.pureComputed(this.suggestions)./*...*/`?

PR #4907, opened 2018-10-16
feat: Add caching mechanism to load image once

atomrc:
I believe service worker does cache the images. But, at app's load time, we send a bunch of requests to the server (could be a lot), and I don't think service worker helps on this.... (?)


app/script/assets/AssetRemoteData.js
```diff
@@ -130,7 +136,10 @@ z.assets.AssetRemoteData = class AssetRemoteData {
         if (!isExpectedError) {
           throw error;
         }
-      });
+      })
+      .finally(() => (this.loadPromise = undefined));
```
gregor: This is only available in Chrome 63 and above. We are not there yet...

PR #4938, opened 2018-10-19
fix: Remove pre guestroom release condition

Yserz:


app/script/conversation/ConversationMapper.js
```diff
@@ -384,15 +384,10 @@ z.conversation.ConversationMapper = class ConversationMapper {
   }
 
   mapAccessCode(conversationEntity, accessCode) {
-    const {code, key, uri} = accessCode;
+    const {uri} = accessCode;
```
gregor: No need to destructure. We only use it once.

app/script/conversation/ConversationMapper.js
```diff
@@ -384,15 +384,10 @@ z.conversation.ConversationMapper = class ConversationMapper {
...
     const isTeamConversation = conversationEntity && conversationEntity.team_id;
 
     if (uri && isTeamConversation) {
-      if (z.util.Environment.frontend.isInternal()) {
-        const accessLink = `${z.config.URL.WEBAPP.INTERNAL}/join/?key=${key}&code=${code}`;
-        return conversationEntity.accessCode(accessLink);
-      }
-
       conversationEntity.accessCode(uri);
```
gregor:
```suggestion
      conversationEntity.accessCode(accessCode.uri);
```

PR #4945, opened 2018-10-22
feat: Lazily migrate users's avatar to the assets v3 API (WEBAPP-5369)

atomrc:


app/script/assets/AssetRemoteData.js
```diff
@@ -127,7 +127,7 @@ z.assets.AssetRemoteData = class AssetRemoteData {
           ? z.assets.AssetCrypto.decryptAesAsset(buffer, this.otrKey.buffer, this.sha256.buffer)
           : buffer;
       })
-      .then(plaintext => new Blob([new Uint8Array(plaintext)], {mime_type: type}))
+      .then(plaintext => new Blob([new Uint8Array(plaintext)], {type}))
```
atomrc, line 130: This one was wrong (see https://developer.mozilla.org/en-US/docs/Web/API/Blob/Blob)

app/script/assets/AssetMapper.js
```diff
@@ -23,31 +23,57 @@ window.z = window.z || {};
...
-  mapProfileAssets: (entity, assets) => {
-    return assets.filter(asset => asset.type === 'image').map(asset => {
+  /**
+   * Creates asset entities out of raw asset data.
+   *
+   * @param {string} userId - the id of the user that will hold the asset
+   * @param {Array<Object>} assets - the assets to map
+   * @returns {MappedAssets} Object containing the mapped assets
+   */
+  mapProfileAssets: (userId, assets) => {
```
atomrc, line 33: I needed to split the mapper's functions in two pure functions that do the mapping and a non pure function that actually put the result of the mapping into a user entity.

This way we have exactly the same signature for both methods, and they can be used without a user entity lying around

app/script/assets/AssetMapper.js
```diff
@@ -23,31 +23,57 @@ window.z = window.z || {};
...
       switch (asset.size) {
         case 'preview':
-          entity.previewPictureResource(assetRemoteData);
+          return Object.assign({}, mappedAssets, {preview: assetRemoteData});
           break;
         case 'complete':
-          entity.mediumPictureResource(assetRemoteData);
-          break;
-        default:
+          return Object.assign({}, mappedAssets, {medium: assetRemoteData});
```
gregor: Let's dissolve the `switch` case. It's just two possible values so we can handle that with an two `if` checks.

app/script/user/UserRepository.js
```diff
@@ -427,6 +428,30 @@ z.user.UserRepository = class UserRepository {
...
+    const hasPicture = userData.picture.length;
+    const hasAsset = userData.assets.length;
+
+    if (hasPicture) {
+      if (!hasAsset) {
+        // if there are no assets, just upload the old picture to the new api
+        const {medium} = z.assets.AssetMapper.mapProfileAssetsV1(userData.id, userData.picture);
+        medium.load().then(imageBlob => this.change_picture(imageBlob));
+      } else {
+        // if an asset is already there, just delete it on the user's profile
```
gregor: Let's clarify the comment here. We do not delete it but the old v1 asset.

app/script/user/UserRepository.js
```diff
@@ -732,7 +757,7 @@ z.user.UserRepository = class UserRepository {
           {key: mediumImageKey, size: 'complete', type: 'image'},
         ];
         return this.selfService
-          .putSelf({assets})
+          .putSelf({assets, picture: undefined})
```
gregor: :+1:

PR #4960, opened 2018-10-23
fix: Attach error reporting despite URL param (WEBAPP-5604)

gregor:
PR looks more complex than it is. The underlying issue was that we called `detach()` on Raygun without ever calling `attach()` in case the domain was blocked from analytics. Due to the difference in `init()` and `updatePrivacyPreference()` in the former Raygun was not attached while later when re-enabling Raygun after toggling the setting it would be attached. Thus https://github.com/MindscapeHQ/raygun4js/issues/302 would only occur when first initiating Rayung on a domain where reporting was disabled.

I aligned the enabling/disabling behavior for a more structured approach.

### DIFF
--- a/app/script/tracking/EventTrackingRepository.js
+++ b/app/script/tracking/EventTrackingRepository.js
@@ -202,14 +202,12 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
     const isDisabledEvent = EventTrackingRepository.CONFIG.USER_ANALYTICS.DISABLED_EVENTS.includes(eventName);
     if (isDisabledEvent) {
       this.logger.info(`Skipped sending disabled event of type '${eventName}'`);
-      return Promise.resolve();
+    } else {
+      const logAttributes = attributes ? `with attributes: ${JSON.stringify(attributes)}` : 'without attributes';
+      this.logger.info(`Tracking event '${eventName}' ${logAttributes}`);
+
+      // Send event if provider API available
     }
-
-    const logAttributes = attributes ? `with attributes: ${JSON.stringify(attributes)}` : 'without attributes';
-    this.logger.info(`Tracking event '${eventName}' ${logAttributes}`);
-
-    // Send event if provider API available
-    return Promise.resolve();
   }
 
   _unsubscribeFromAnalyticsEvents() {
--- a/app/script/tracking/EventTrackingRepository.js
+++ b/app/script/tracking/EventTrackingRepository.js
@@ -76,44 +76,93 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
    */
   init(privacyPreference) {
     this.privacyPreference = privacyPreference;
-    this.logger.info(`Initialize tracking and error reporting: ${this.privacyPreference}`);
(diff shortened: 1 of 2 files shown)


app/script/tracking/EventTrackingRepository.js
```diff
@@ -76,44 +76,88 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
   init(privacyPreference) {
     this.privacyPreference = privacyPreference;
-    this.logger.info(`Initialize tracking and error reporting: ${this.privacyPreference}`);
-
-    return Promise.resolve()
-      .then(() => {
-        if (this._isDomainAllowedForTracking() && this.privacyPreference) {
-          this._enableErrorReporting();
-          return this._initTracking();
-        }
```
gregor, line 86: Compare this....

app/script/tracking/EventTrackingRepository.js
```diff
@@ -76,44 +76,88 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
+    const hasPreferenceChanged = privacyPreference !== this.privacyPreference;
+    if (hasPreferenceChanged) {
       this.privacyPreference = privacyPreference;
 
-      if (privacyPreference) {
-        this._enableErrorReporting();
-        if (this._isDomainAllowedForTracking()) {
-          this._reEnableTracking();
-        }
-      } else {
```
gregor, line 102: ... with this.

app/script/tracking/EventTrackingRepository.js
```diff
@@ -76,44 +76,88 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
-        this._disableErrorReporting();
-        this._disableTracking();
-      }
+      return this.privacyPreference
+        ? this._enableServices().then(() => this._trackEvent(z.tracking.EventName.SETTINGS.OPTED_IN_TRACKING))
+        : this._trackEvent(z.tracking.EventName.SETTINGS.OPTED_OUT_TRACKING).then(() => this._disableServices());
```
atomrc: In order not to create a single point of failure with whatever tracking system we will use in the future, let's not make the call to `_disableServices` dependent on the result of tracking the event

app/script/tracking/EventTrackingRepository.js
```diff
@@ -129,95 +173,47 @@ z.tracking.EventTrackingRepository = class EventTrackingRepository {
...
-      ? `Tracking event '${eventName}' with attributes: ${JSON.stringify(attributes)}`
-      : `Tracking event '${eventName}' without attributes`;
-    this.logger.info(logMessage);
-
     const isDisabledEvent = EventTrackingRepository.CONFIG.USER_ANALYTICS.DISABLED_EVENTS.includes(eventName);
-    if (!isDisabledEvent) {
-      // send event to provider
+    if (isDisabledEvent) {
+      this.logger.info(`Skipped sending disabled event of type '${eventName}'`);
+      return Promise.resolve();
```
atomrc: Should return `void`, we don't care about the result of the tracking

gregor: We do care about it. We need to ensure an event was sent before we announce to the ouside world that we are free to disable.

PR #4962, opened 2018-10-23
fix: Avoid recomputing input bar rich text container on init (WEBAPP-5621)

atomrc:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -261,6 +268,7 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
         }
 
         $('.conversation').css({opacity: 1});
+        this.capture_scrolling_event = true;
```
atomrc, line 272: ~~This is the actual fix, enabling scroll handling **after** the programmatic scroll have occurred....~~

On a side note, this will probably annoy us again in the future, so we might need to think of a better way to do that.
A different API to explicitly scroll into the messages without triggering listeners could be a solution.... Thinking in progress

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -216,6 +218,13 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
       });
   }
 
+  _getMessagesContainer() {
```
atomrc: This is bonus. As the messages list element doesn't change during the lifecycle of a conversation, we can safely cache it.

app/script/view_model/content/MessageListViewModel.js
atomrc: Wow, I am actually not even sure this will fix the issue.
As we throttle the handling of scroll events, we might end up handling a past event with the boolean set to `true`....

app/script/view_model/bindings/CommonBindings.js
```diff
@@ -174,7 +174,7 @@ ko.bindingHandlers.heightSync = {
 
     const resizeTarget = () => {
       const sourceHeight = element.scrollHeight;
-      const targetHeight = parseInt(targetElement.style.height, 10);
+      const targetHeight = targetElement.offsetHeight;
```
atomrc, line 177: Ok, here is the **actual** fix.

The problem here is that we would trigger resizing on the first render of the input bar (because no CSS property has been set yet) and thus trigger a forced scrollToBottom on the message list.

Using the actual height (not the CSS property) avoids triggering the callback and the scrolling. 

On a side note: This addresses this specific bug but is actually the consequence of multiple view models accessing the same DOM element directly, we **should never do that** in the future.

PR #4976, opened 2018-10-24
feat: Rename notification setting to mentions and replies (WEBAPP-5632)

gregor:


app/script/entity/Conversation.js
```diff
@@ -122,15 +122,15 @@ z.entity.Conversation = class Conversation {
 
       const knownNotificationStates = Object.values(z.conversation.NotificationSetting.STATE);
       if (knownNotificationStates.includes(this.mutedState())) {
-        const isStateOnlyMentions = this.mutedState() === z.conversation.NotificationSetting.STATE.ONLY_MENTIONS;
-        const isInvalidState = isStateOnlyMentions && !this.selfUser().inTeam();
+        const isExpectedState = this.mutedState() === z.conversation.NotificationSetting.STATE.MENTIONS_AND_REPLIES;
```
atomrc: I don't quite understand this renaming. Why is it expected that the state is `MENTIONS_AND_REPLIES`?

gregor: Thanks for brining this up. I cleaned up the naming.

PR #4979, opened 2018-10-24
feat: Add replies to notification behavior (WEBAPP-5631)

gregor:


app/script/notification/NotificationRepository.js
```diff
@@ -832,7 +851,7 @@ z.notification.NotificationRepository = class NotificationRepository {
       return isEventToNotify;
     }
 
-    const isSelfMentioned = messageEntity.is_content() && messageEntity.isUserMentioned(userId);
-    return isEventToNotify && isSelfMentioned;
+    const isSelfMentionOrReply = messageEntity.is_content() && messageEntity.isResponseToUser(userId);
```
atomrc: I guess the condition is a bit off here. Should probably be

```suggestion
    const isSelfMentionOrReply = messageEntity.is_content() && ( messageEntity.isUserMentioned(userId) || messageEntity.isResponseToUser(userId));
```

atomrc: Ok, ignore this comment. It is included in the `isResponseToUser` actually

app/script/entity/message/ContentMessage.js
```diff
@@ -117,7 +117,11 @@ z.entity.ContentMessage = class ContentMessage extends z.entity.Message {
   }
 
   isReplyToUser(userId) {
-    return this.quote() && this.quote().isQuoteFromUser(userId);
+    return this.quote() ? this.quote().isQuoteFromUser(userId) : false;
+  }
+
+  isResponseToUser(userId) {
```
atomrc: Not sure about the naming of this function. 
`isReply` and `isResponse` are very close and hard to distinguish. 

I believe we can come up with a better name. 

WDYT of `isTargetingUser`?

ffflorian: As agreed f2f, will change it to `isUserTargeted`.

PR #4985, opened 2018-10-25
feat: Message hashing (WEBAPP-5633)

ffflorian:


test/unit_tests/message/MessageHashingSpec.js
```diff
@@ -0,0 +1,272 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+// grunt test_init && grunt test_run:message/MessageHashing
+
+'use strict';
+
+describe('z.message.MessageHashing', () => {
+  describe('_getTimestampArray', () => {
```
atomrc: Private functions should not be tested as they are not supposed to be exposed

app/script/message/MessageHashing.js
```diff
@@ -0,0 +1,115 @@
...
+
+'use strict';
+
+const Long = window.dcodeIO.Long;
+const crypto = window.crypto;
+
+window.z = window.z || {};
+window.z.message = z.message || {};
+
+z.message.MessageHashing = {
```
atomrc: How about MessageHasher for the naming?

atomrc: Let's try to expose a clean public API with just the tthreewo methods that should be used outside (namely `getAssetMessageHash`, 'getLocationMessageHash` and `getTextMessageHash`).

To do that, we could use an IIFE and create functions in there. This IIFE will then return an object with the public API

app/script/entity/message/Text.js
```diff
@@ -36,6 +36,9 @@ z.entity.Text = class Text extends z.entity.Asset {
     // Array of z.message.MentionEntity instances
     this.mentions = ko.observableArray();
 
+    // Quote object
+    this.quote = ko.observable();
```
atomrc: Let's remove this, it is not used in this PR

PR #5009, opened 2018-10-30
fix: Revert phone login to own ajax call

gregor:


atomrc:
Could you give a little context to why it fails exactly?

gregor:
No. I honestly don't know at this point in time. But I am fairly sure that this change from a few days ago is at fault. You will see in the run results in the morning.

PR #5010, opened 2018-10-30
fix: Pad client IDs with zeros on auth page (WEBAPP-5476)

ffflorian:


app/script/auth/component/ClientItem.tsx
```diff
@@ -72,7 +72,11 @@ class ClientItem extends React.Component<CombinedProps, State> {
     validationError: null,
   };
 
-  formatId = (id = '?') => id.toUpperCase().replace(/(..)/g, '$1 ');
+  formatId = (id = '?', outputLength = 16) => {
```
bennycode, line 75: This functionality can be moved to our `commons` package.

ffflorian: As agreed f2f, will use `String.prototype.padStart()` instead.

PR #5015, opened 2018-10-31
feat: Restore reply state when switching conversations (WEBAPP-5640)

ffflorian:


app/page/template/content/conversation/message-list.htm
```diff
@@ -9,7 +9,7 @@
                       attr: {'data-uie-uid': message.id, 'data-uie-value': message.super_type, 'data-uie-expired-status': message.ephemeral_expires, 'data-uie-send-status': message.status}"
            data-uie-name="item-message">
         <div data-bind="template: {name: 'timestamp'}"></div>
-        <div data-bind="template: {name: message.super_type}, css: {'message-isreplying': message.isReplying()}"></div>
+        <div data-bind="template: {name: message.super_type}, css: {'message-isreplying': message.id === $root.inputBar.replyMessageId()}"></div>
```
ffflorian, line 12: As agreed with @atomrc, we need to get rid of observables on the message entity which are reading states from the input bar.

app/script/view_model/content/InputBarViewModel.js
```diff
@@ -645,13 +660,13 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
   sendMessage(messageText, replyMessageEntity) {
     if (messageText.length) {
       const mentionEntities = this.currentMentions.slice();
-      const quoteEntity = replyMessageEntity && new z.message.QuoteEntity(replyMessageEntity.id, '');
+      const replyEntity = replyMessageEntity && new z.message.QuoteEntity(replyMessageEntity.id, '');
```
atomrc: Let's keep it as the name of the class of the entity (`quoteEntity` which is an instance of `QuoteEntity`)

PR #5016, opened 2018-10-31
feat: Remove password check for SSO users when removing devices from a full device list (WEBAPP-5638)

bennyn:


app/script/auth/component/ClientItem.tsx
```diff
@@ -144,6 +145,18 @@ class ClientItem extends React.Component<CombinedProps, State> {
     this.props.onClick(event);
   };
 
+  handlePasswordlessClientDeletion = (event: React.FormEvent) => {
+    event.preventDefault();
+
+    return Promise.resolve()
+      .then(() => this.props.onClientRemoval())
```
ffflorian, line 152:
```js
return this.props.onClientRemoval())
```

app/script/auth/component/ClientItem.tsx
```diff
@@ -144,6 +145,18 @@ class ClientItem extends React.Component<CombinedProps, State> {
...
+      .catch(error => {
+        if (!error.label) {
```
ffflorian, line 154: I think it would be good to also check what's in the label, what do you think?

app/script/auth/component/ClientItem.tsx
bennycode: It's not possible because `onClientRemoval` is not a promise so it needs to be wrapped in a promise.

app/script/auth/component/ClientItem.tsx
bennycode: The check for `error.label` is just an indicator to know if the error should be thrown or not. No need to check what's inside the label.

PR #5040, opened 2018-11-02
feat: Add error object to quote event (WEBAPP-5647)

ffflorian:


app/script/event/preprocessor/QuotedMessageMiddleware.js
```diff
@@ -55,17 +55,23 @@ z.event.preprocessor.QuotedMessageMiddleware = class QuotedMessageMiddleware {
 
     return this.eventService.loadEvent(event.conversation, quote.quoted_message_id).then(quotedMessage => {
       if (!quotedMessage) {
-        this.logger.warn('Quoted message not found');
-        // TODO add error metadata (to discuss with UI team)
+        this.logger.warn(`Quoted message with ID "${quote.quoted_message_id}" not found.`);
+        event.error = {
```
atomrc: Let's rather put that content in the `event.quote.error` instead of `event.error`.

app/script/event/preprocessor/QuotedMessageMiddleware.js
```diff
@@ -55,17 +55,23 @@ z.event.preprocessor.QuotedMessageMiddleware = class QuotedMessageMiddleware {
...
+          message: z.message.QuoteEntity.ERROR.MESSAGE_NOT_FOUND,
```
atomrc: Let's use an error type (some kind of identifier, could be string identifiers) rather than a message here.
The message part is for the UI side to decide.

PR #5045, opened 2018-11-02
fix: input bar behaviour and appearance

AndyLnd:
- your name is shown as a name instead of 'You' in the reply preview
- the message list now respects the height of the input bar including the reply preview
- input gets focus when clicking 'reply'


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -639,7 +644,7 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     amplify.unsubscribeAll(z.event.WebApp.SHORTCUT.PING);
   }
 
-  scrollMessageList(newListHeight, previousListHeight) {
+  scrollMessageList(newListHeight = 0, previousListHeight = 0) {
```
atomrc, line 647: Oh crap. I missed this one on my scroll refacto :/

PR #5046, opened 2018-11-02
feat: Do not scroll to bottom on input focus (WEBAPP-5648)

atomrc:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -137,8 +138,6 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
         this.conversation().isActiveParticipant() && this.conversation().inTeam() && this.conversation().isGuestRoom()
       );
     });
-
-    amplify.subscribe(z.event.WebApp.CONVERSATION.INPUT.CLICK, this.on_conversation_input_click.bind(this));
```
ffflorian, line 141: What do you think about removing `z.event.WebApp.CONVERSATION.INPUT.CLICK` since we are not using it anymore?

atomrc: Absolutely!

PR #5047, opened 2018-11-02
fix: Auth tests

Yserz:


karma.conf.react.js
```diff
@@ -45,7 +45,7 @@ module.exports = function(config) {
         flags: ['--no-sandbox'],
       },
     },
-    files: [`${test}main.test.js`, `${dist}dexie.js`, `${dist}script.js`, `${dist}test.js`],
```
ffflorian, line 48: What do you think about setting `dist` to `'aws/static/min'` and add a `/` to each string? It makes it more obvious that the final string will be a path.

e.g.:
```js
const dist = 'aws/static/min';
...
files: [..., `${dist}/dexie.js`, `${dist}/script.js`, `${dist}/test.js`],
```

ffflorian: Same goes for `test`.

PR #5050, opened 2018-11-03
feat: Added detection for SSO-compatible wrapper

pwnsdx:


ffflorian:
@pwnsdx can we delete this branch?

PR #5053, opened 2018-11-05
feat: Basic support for all message types in quotes (WEBAPP-5627)

AndyLnd:
Adds (very) basic support for all types of quotes that are available.
Not included: 
- detail view for images
- special treatment of videos (smaller preview, full size on play)
- mention interaction in quoted text
- "scroll to original message" line

### DIFF
--- a/app/script/components/messageQuote.js
+++ b/app/script/components/messageQuote.js
@@ -23,7 +23,7 @@ window.z = window.z || {};
 window.z.components = z.components || {};
 
 z.components.MessageQuote = class MessageQuote {
-  constructor({conversation, quote, conversationRepository, locationRepository}) {
+  constructor({conversation, conversationRepository, locationRepository, quote}) {
     this.updateCanShowMore = this.updateCanShowMore.bind(this);
     this.toggleShowMore = this.toggleShowMore.bind(this);
 


app/script/components/messageQuote.js
```diff
@@ -0,0 +1,100 @@
...
+
+window.z = window.z || {};
+window.z.components = z.components || {};
+
+z.components.MessageQuote = class MessageQuote {
+  constructor({quote, conversationRepository, locationRepository}) {
+    this.updateCanShowMore = this.updateCanShowMore.bind(this);
+    this.toggleShowMore = this.toggleShowMore.bind(this);
+
+    const activeConversation = conversationRepository.active_conversation;
```
atomrc: Since we don't need the power of the `conversationRepository` I would argue that we could just inject a `conversationEntity` as a param of the component.

The less the component knows about the app architecture the more future proof we will be

app/script/components/messageQuote.js
```diff
@@ -0,0 +1,100 @@
...
+    this.locationRepository = locationRepository;
+    this.message = ko.observable();
+    conversationRepository.get_message_in_conversation_by_id(activeConversation(), quote().messageId).then(message => {
```
atomrc: Also here, we could pass the message as a parameter

app/script/components/messageQuote.js
atomrc: Here is a suggestion of signature that could work if I am not mistaken

```
<message-quote param="message: message, locationRepo: locationRepo"></message-quote>
```

You should have all the info you need with that

app/script/components/messageQuote.js
AndyLnd: This is not the same message as the the one containing the quote, but the one we are quoting. Therefore we need the conversationRepo in here and can't just pass the message down.

app/script/components/messageQuote.js
```diff
@@ -0,0 +1,99 @@
...
+ *
+ */
+
+'use strict';
+
+window.z = window.z || {};
+window.z.components = z.components || {};
+
+z.components.MessageQuote = class MessageQuote {
+  constructor({conversation, quote, conversationRepository, locationRepository}) {
```
ffflorian: Sorting

ffflorian:
@AndyLnd let's ignore the flaky test and merge.

PR #5055, opened 2018-11-05
fix: Prevent .env file from loading in production environment

Yserz:


package.json
```diff
@@ -224,8 +224,8 @@
     "postinstall": "node ./bin/postinstall.js",
     "prettier": "prettier --ignore-path .gitignore \"**/*.{json,md}\"",
     "start": "yarn build:server && yarn grunt build_dev && concurrently \"yarn start:server:dev\" \"yarn grunt watch\"",
-    "start:server:dev": "node ./aws/index.js",
-    "start:prod": "pm2 start \"./index.js\" --name \"Webapp\" && pm2 logs",
+    "start:server:dev": "cross-env NODE_ENV=development node ./aws/index.js",
+    "start:prod": "cross-env NODE_ENV=production pm2 start \"./aws/index.js\" --name \"Webapp\" && pm2 logs",
```
ffflorian: `NODE_ENV=production` is not needed here since we set `nodeEnvironment` to `production` if `NODE_ENV` is not set anyway.

Additionally, `cross-env` is only in the `devDependencies`, so it will not be available in a production environment.

```suggestion
    "start:prod": "pm2 start \"./aws/index.js\" --name \"Webapp\" && pm2 logs",
```

PR #5057, opened 2018-11-05
feat: Add method to scroll to specific message (WEBAPP-5648)

atomrc:


app/script/entity/Conversation.js
```diff
@@ -667,15 +667,13 @@ z.entity.Conversation = class Conversation {
...
+   * @param {string} messageId - ID of message to be retrieved
    * @returns {z.entity.Message|undefined} Message with ID or undefined
    */
-  get_message_by_id(id) {
-    for (const message_et of this.messages()) {
-      if (message_et.id === id) {
-        return message_et;
-      }
-    }
+  getMessage(messageId) {
```
ffflorian, line 675: Why not `getMessageById`?

atomrc: Mostly because I believe it's extra verbosity that can be deduced from the signature.
But open to debate (you can have a look at the `EventService` for example, that never mentions `Id` in the methods name)

ffflorian: > you can have a look at the EventService for example, that never mentions Id in the methods name

True. On the other hand, we have [many functions](https://github.com/wireapp/wire-webapp/search?q=by+id+OR+byid) that are called `...byId()` or `...by_id()`.

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -380,20 +372,34 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
...
+          });
+
+    loadMessagePromise.then(() => {
+      z.util.afterRender(() => {
+        const message_element = $(`.message[data-uie-uid="${messageId}"]`);
+
+        if (message_element.length) {
+          message_element.removeClass('message-marked');
+          const message_list_element = this._getMessagesContainer();
+          message_list_element.scrollBy(message_element.offset().top - message_list_element.height() / 2);
```
AndyLnd: You might want to look into https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView for this.

atomrc: Good idea but unfortunately, this scrolls the whole app :D

PR #5058, opened 2018-11-05
refactor: Remove knowledge of the message list from the input bar

atomrc:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -170,6 +172,17 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
     }
   }
 
+  _handleInputResize(inputSizeDiff) {
+    const antiscroll = $('.message-list').data('antiscroll');
+    if (antiscroll) {
+      antiscroll.rebuild();
+    }
+
+    if (inputSizeDiff !== 0) {
```
AndyLnd: ```if (inputSizeDiff)``` should be enough as it also implies that the parameter can be neglected.

PR #5066, opened 2018-11-06
feat: Show error message for invalid quotes

AndyLnd:


app/script/message/QuoteEntity.js
```diff
@@ -30,10 +30,11 @@ z.message.QuoteEntity = class QuoteEntity {
     };
   }
 
-  constructor(messageId, userId, hash) {
+  constructor({error, hash, messageId, userId}) {
```
atomrc, line 33: :+1: to the config object here

PR #5069, opened 2018-11-06
feat: Styling and behaviour of quoted content (WEBAPP-5627)

AndyLnd:


app/script/components/asset/videoAsset.js
```diff
@@ -81,7 +80,8 @@ z.components.VideoAssetComponent = class VideoAssetComponent {
     this.logger.error('Video cannot be played', jquery_event);
   }
 
-  on_play_button_clicked() {
+  onPlayButtonClicked() {
+    this.displaySmall(false);
```
atomrc, line 84: Don't you need to revert this state at some point?

AndyLnd: No, the video should stay in the expanded format once it is played.

PR #5074, opened 2018-11-06
fix: Catch invalid URIs on link preview generation (WEBAPP-5661)

bennyn:
Our wrappers can generate link previews. To make this work, the wrapper exposes an open graph library to the `window.openGraph` namespace. This library is then being used from the webapp code when calling `wire.app.repository.links.getLinkPreview` which calls the underlying `_fetchOpenGraphData` function.

In some cases (for example when entering an invalid URI), the `window.openGraph` library (which uses a Promise interface) throws an error but we never catched that error inside of `_fetchOpenGraphData`. Now we do.

This addresses the following crash report:
- https://app.raygun.com/crashreporting/8785p7/errors/3014612384

### DIFF
--- a/test/unit_tests/links/LinkPreviewRepositorySpec.js
+++ b/test/unit_tests/links/LinkPreviewRepositorySpec.js
@@ -19,6 +19,7 @@
 
 'use strict';
 
+// grunt test_init && grunt test_run:links/LinkPreviewRepository
 describe('z.links.LinkPreviewRepository', () => {
   let link_preview_repository = null;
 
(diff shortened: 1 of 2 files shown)


test/unit_tests/links/LinkPreviewRepositorySpec.js
```diff
@@ -70,5 +85,20 @@ describe('z.links.LinkPreviewRepository', () => {
           done();
         });
     });
+
+    it('catches errors that are raised by the openGraph lib when invalid URIs are parsed', done => {
+      window.openGraph = function() {
+        return Promise.reject(new Error('Invalid URI'));
+      };
```
ffflorian:
```js
window.openGraph = () => Promise.reject(new Error('Invalid URI'));
```

PR #5076, opened 2018-11-06
runfix: Decouple messageQuote from hierarchy

AndyLnd:


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -45,6 +45,8 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
     this.should_hide_user_avatar = this.should_hide_user_avatar.bind(this);
     this.showUserDetails = this.showUserDetails.bind(this);
     this._handleWindowResize = this._handleWindowResize.bind(this);
+    this._focusMessage = this._focusMessage.bind(this);
```
atomrc: `_focusMessage` can be marked as public now since it is used in the view

PR #5077, opened 2018-11-06
fix: Improve extra messages loading behavior (WEBAPP-5621)

atomrc:
This PR changes the scroll behavior quite a bit.

The basic idea is:

- if the container is not scrollable (the content is smaller than the container), then the `mousewheel` event handles loading the messages depending on the direction the user wants to scroll.
- if the container is scrollable, we just load messages depending on whether we hit the top of the bottom of the message list

### DIFF
--- a/app/script/view_model/content/MessageListViewModel.js
+++ b/app/script/view_model/content/MessageListViewModel.js
@@ -82,41 +82,43 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
 
     this.recalculate_timeout = undefined;
 
-    // Should we scroll to bottom when new message comes in
-    this.should_scroll_to_bottom = true;
-
-    // Check if the message container is to small and then pull new events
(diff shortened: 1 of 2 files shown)


app/script/view_model/content/MessageListViewModel.js
```diff
@@ -80,41 +80,43 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
...
-          if (this._conversationHasExtraMessages(this.conversation())) {
-            this._push_messages();
-          }
+      return true;
+    }, 50);
 
-          this._mark_conversation_as_read_on_focus(this.conversation());
-        }
-
-        this.should_scroll_to_bottom = scroll_position > scrollEnd - z.config.SCROLL_TO_LAST_MESSAGE_THRESHOLD;
```
atomrc, line 119: This has been moved to a getter that returns true when we need to stick to the bottom

PR #5091, opened 2018-11-07
runfix: Keep quotes in edited messages

AndyLnd:


app/script/view_model/content/InputBarViewModel.js
```diff
@@ -686,23 +692,26 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     this._resetDraftState();
   }
 
+  _generateQuotePromise(replyMessageEntity) {
```
atomrc: You can drop the `Promise` from the function name.

PR #5094, opened 2018-11-07
runfix: Convert message timestamp to seconds before hashing

ffflorian:
Internally we use milliseconds for the event date, but all teams agreed to use seconds for hashing.

### DIFF
--- a/app/script/message/MessageHasher.js
+++ b/app/script/message/MessageHasher.js
@@ -64,7 +64,11 @@ z.message.MessageHasher = (() => {
    * @returns {number[]} the timestamp as long endian bytes
    * @private
    */
-  const getTimestampBytes = event => Long.fromInt(new Date(event.time).getTime()).toBytesBE();
+  const getTimestampBytes = event => {
+    const unixTimestamp = new Date(event.time).getTime();
+    const timestampSeconds = Math.floor(unixTimestamp / 1e3);
+    return Long.fromInt(timestampSeconds).toBytesBE();
+  };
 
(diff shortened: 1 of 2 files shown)


app/script/message/MessageHasher.js
```diff
@@ -64,7 +64,11 @@ z.message.MessageHasher = (() => {
    * @returns {number[]} the timestamp as long endian bytes
    * @private
    */
-  const getTimestampBytes = event => Long.fromInt(new Date(event.time).getTime()).toBytesBE();
+  const getTimestampBytes = event => {
+    const unixTimestamp = new Date(event.time).getTime();
+    const timestampSeconds = Math.floor(unixTimestamp / 1e3);
```
ffflorian, line 69: `1e3 = 1000`

test/unit_tests/message/MessageHasherSpec.js
```diff
@@ -37,19 +37,19 @@ describe('z.message.MessageHasher', () => {
       it('correctly hashes text events', () => {
         const tests = [
           {
-            event: createTextEvent('This has **markdown**', 1540213965),
+            event: createTextEvent('This has **markdown**', 1540213965000),
```
ffflorian: Need to fake milliseconds to match our internal format.

app/script/message/MessageHasher.js
atomrc: Was the `floor` agreed with all the teams?

ffflorian: Yes.

PR #5102, opened 2018-11-08
runfix: Edit own message with quote (WEBAPP-5664)

ffflorian:


app/script/event/preprocessor/QuotedMessageMiddleware.js
```diff
@@ -75,13 +75,16 @@ z.event.preprocessor.QuotedMessageMiddleware = class QuotedMessageMiddleware {
...
       this.logger.info(`Updating '${replies.length}' replies to updated message '${originalMessageId}'`);
+
       replies.forEach(reply => {
         reply.data.quote.message_id = event.id;
         this.eventService.replaceEvent(reply);
       });
-      return event;
+
+      const decoratedData = Object.assign(event.data, {quote: originalEvent.data.quote});
+      return Object.assign(event, {data: decoratedData});
```
AndyLnd: You are assigning the `{quote: ...}` to `event.data` already. So I don't think it's necessary to do `Object.assign(event, {data: decoratedData})`.

PR #5138, opened 2018-11-12
runfix: Always load users with quotes (WEBAPP-5706)

AndyLnd:


app/script/conversation/ConversationRepository.js
```diff
@@ -403,20 +403,40 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
       if (event) {
         return this.event_mapper.mapJsonEvent(event, conversationEntity);
       }
       throw new z.error.ConversationError(z.error.ConversationError.TYPE.MESSAGE_NOT_FOUND);
     });
+
+    if (ensureUser) {
+      return messagePromise.then(message => {
+        if (!message.user().id) {
+          return this.user_repository.get_user_by_id(message.from).then(userEntity => {
```
atomrc, line 428: I believe certain messages can have no `from` (system messages, for eg). We probably want to add this safeguard

PR #5145, opened 2018-11-13
refactor: Componentify the message view model

atomrc:
Two benefits from this PR:
- performance;
- isolation (which will lead to an easier migration later on).

# Performance

Previously, we were using ko templates, which as quite slow (https://jsperf.com/knockout-template-binding-performance/17)
Now we use a component, so the template part is compiled by ko only once.
Which give us those results:

| | initial load of messages | subsequent load of messages |
| ---- | ------ | ----- |
| with template | ~500-700ms | ~200-300ms |
| with component | ~100-200ms | ~50-100ms |

# component

This allows isolation (not relying on globally defined classes), that will greatly help us when we want to migrate to another tech (all the dependencies are now explicitly listed)


app/script/components/message.js
```diff
@@ -176,172 +270,277 @@
...
+      <!-- /ko -->
+      <!-- ko if: message.isTypeUnverified() -->
+        <span class="message-header-sender-name" data-bind="text: message.unsafeSenderName()"></span>
+        <span class="ellipsis" data-bind="l10n_text: z.string.conversationDeviceUnverified"></span>
+        <span class="message-verification-action text-theme" data-bind="click: () => clickOnDevice(message), text: message.captionUnverifiedDevice" data-uie-name="go-devices"></span>
+      <!-- /ko -->
+      <!-- ko if: message.isTypeNewDevice() -->
+        <span class="message-header-plain-sender-name" data-bind='text: message.captionUser'></span>
+        <span class="ellipsis" data-bind="text: message.captionStartedUsing"></span>
+        <span class="message-verification-action text-theme" data-bind="click: () => clickOnDevice(message), text: message.captionNewDevice" data-uie-name="go-devices"></span>
```
atomrc: TODO use imperative semantic (showDevice)

PR #5146, opened 2018-11-13
runfix: Update quotes on visible messages (WEBAPP-5712)

AndyLnd:


app/script/conversation/ConversationRepository.js
```diff
@@ -3491,6 +3491,20 @@ z.conversation.ConversationRepository = class ConversationRepository {
         if (conversationEntity && messageEntity) {
           const replacedEntity = conversationEntity.add_message(messageEntity, true);
           if (replacedEntity) {
+            const messages = conversationEntity.messages_unordered();
+
+            const updatedMessages = messages.map(message => {
+              const hasEditedQuote =
+                message.quote && message.quote() && message.quote().messageId === replacedEntity.id;
```
ffflorian, line 3498: A `Message` always has `message.quote`, so we don't need to check it.

```suggestion
                message.quote() && message.quote().messageId === replacedEntity.id;
```

atomrc: I believe the check is legit. We should not remove it

app/script/conversation/ConversationRepository.js
```diff
@@ -3491,6 +3491,20 @@ z.conversation.ConversationRepository = class ConversationRepository {
         if (conversationEntity && messageEntity) {
           const replacedEntity = conversationEntity.add_message(messageEntity, true);
           if (replacedEntity) {
+            const messages = conversationEntity.messages_unordered();
```
atomrc, line 3494: Possible optimization, you can use the ordered messages list (namely `messages` I believe) and do stop when messages are before the timestamp of the edited message

app/script/conversation/ConversationRepository.js
ffflorian: :+1:

PR #5170, opened 2018-11-16
fix: Check app instance cookie existence before deleting

Yserz:


atomrc:
I think this is a proper `fix:` not a `runfix:`

PR #5175, opened 2018-11-16
fix: Link embedding

### LINKED ISSUE #529: Soundcloud links validation
https://soundcloud.com/charts/top — is not treated as valid link. Wire tries to generate composition preview.

ffflorian:
This PR will
* add link preview rendering for static SoundCloud, YouTube, Vimeo and Spotify pages
* allow embedding all vimeo channel videos, not only from staff picks
* close #529

### DIFF
--- a/app/script/links/LinkPreviewBlackList.js
+++ b/app/script/links/LinkPreviewBlackList.js
@@ -30,6 +30,7 @@ z.links.LinkPreviewBlackList = (() => {
     'discover',
     'discussion',
     'featured',
+    'home',
     'messages',
     'mobile',
     'pages',
@@ -46,7 +47,7 @@ z.links.LinkPreviewBlackList = (() => {
     `soundcloud.com/(?!(${soundcloudStatic.join('|')}))`,
     'spotify.com/(?![A-z]{2}/)',
     'youtu.be',
-    'youtube(-nocookie)?.com/watch',
+    'youtube(-nocookie)?.com/(watch|embed)',
     'vimeo.com/(channels(/[^/]+)?/)?[0-9]+',
   ];
 
--- a/app/script/links/LinkPreviewBlackList.js
+++ b/app/script/links/LinkPreviewBlackList.js
@@ -23,7 +23,7 @@ window.z = window.z || {};
 window.z.links = z.links || {};
 
 z.links.LinkPreviewBlackList = (() => {
-  const soundcloudStatic = [
+  const soundCloudStatic = [
     'about',
     'channels',
     'charts',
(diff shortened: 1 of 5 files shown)


test/unit_tests/links/LinkPreviewBlackListSpec.js
```diff
@@ -25,36 +25,66 @@ describe('is_blacklisted', () => {
...
-  it('should return true if link is vimeo', () => {
-    const url = 'vimeo.com';
+  it('should return false if link is Spotify static', () => {
+    const url = 'https://www.spotify.com/en/legal';
 
-    expect(z.links.LinkPreviewBlackList.isBlacklisted(url)).toBeTruthy();
+    expect(z.links.LinkPreviewBlackList.isBlacklisted(url)).toBe(false);
+  });
+
+  it('should return true if link is SoundCloud', () => {
```
bennycode: Let's phrase these a little less technical:
```
 it('should return true if link is SoundCloud',
```
->
```
 it('blacklists SoundCloud links',
```

app/script/links/LinkPreviewBlackList.js
```diff
@@ -23,9 +23,35 @@ window.z = window.z || {};
 window.z.links = z.links || {};
 
 z.links.LinkPreviewBlackList = (() => {
-  const BLACKLIST = ['soundcloud', 'spotify', 'youtu[.]?be', 'vimeo'];
+  const soundcloudStatic = [
```
bennycode: `soundcloudStatic` -> `soundCloudStatic`

AndyLnd: I would stick with florian's style, as soundcloud is a single name.

app/script/media/MediaEmbeds.js
```diff
@@ -70,8 +70,8 @@ z.media.MediaEmbeds = (function() {
     // example: http://regexr.com/3ase5
     soundcloud: /(https?:\/\/(?:www\.|m\.)?)?soundcloud\.com(\/[\w-]+){2,3}/g,
```
atomrc, line 71: Let's try to rely on this list to actually generate the linkPreview black list.

app/script/links/LinkPreviewBlackList.js
bennycode: ...written with a capital C. 💃

app/script/media/MediaEmbeds.js
ffflorian: Unfortunately the regex needed is set in `MediaEmbeds`, but needed in `LinkPreviewBlacklist`.
Since `LinkPreviewBlacklist` is loaded before `MediaEmbeds`, we would need to change the order of loading these files. I am not sure which other implications that would have.

PR #5194, opened 2018-11-22
chore: Bump @wireapp/core to 7.2.14

ffflorian:


ffflorian:
Does not fix WEBAPP-5732. So I changed the title.

PR #5198, opened 2018-11-22
 fix: Remove any traces of gmail import (WEBAPP-5733)

AndyLnd:


app/script/main/app.js
```diff
@@ -61,7 +61,6 @@ z.main.App = class App {
     this.util = this._setup_utils();
 
     // @todo Added for wrapper backwards compatibility. Remove after uptake of version > 3.2.
```
ffflorian:
```suggestion
```

PR #5199, opened 2018-11-22
fix: Appearance of options

AndyLnd:


app/script/view_model/panel/NotificationsViewModel.js
```diff
@@ -40,8 +38,18 @@ z.viewModel.panel.NotificationsViewModel = class NotificationsViewModel extends
 
     this.isRendered = ko.observable(false).extend({notify: 'always'});
 
-    this.currentNotificationSetting = ko.pureComputed(() => {
-      return this.activeConversation() && this.activeConversation().notificationState();
+    this.currentNotificationSetting = ko.observable();
+
+    this.activeConversation.subscribe(conversation => {
```
atomrc, line 43: Why didn't you keep a pure computed? Now you need to unsubscribe to those!
I believe we can come up with a cleaner solution.

AndyLnd: Not really, this is because of the `checked` binding that is now used.

PR #5202, opened 2018-11-23
chore: Use webpack to compile legacy sources

atomrc:


.gitignore
```diff
@@ -1,42 +1,20 @@
 !/.idea/codeStyleSettings.xml
 !/.idea/modules.xml
-!/aws/templates/
-.DS_Store
```
Yserz, line 4: please keep the DS_Store

atomrc: We should not have system specific ignore here. This should be in `.gitignoreglobal` of each user. We cannot account for every user/platform/editor/... here

app/script/cryptography/CryptographyRepository.js
```diff
@@ -520,7 +521,7 @@ z.cryptography.CryptographyRepository = class CryptographyRepository {
     const customData = {
       clientLocalClass: this.currentClient().class,
       clientLocalType: this.currentClient().type,
-      cryptoboxVersion: cryptobox.version,
+      cryptoboxVersion: cryptoboxVersion,
```
Yserz:
```suggestion
      cryptoboxVersion,
```

Gruntfile.js
```diff
@@ -19,45 +19,28 @@
...
-      script: 'app/script',
       style: 'app/style',
-      templateDist: 'app/page/template/_dist',
     },
-    aws_: 'aws',
-    aws: {
-      s3: 'aws/s3',
-      static: 'aws/static',
-      templates: 'aws/templates',
+    dist: {
```
atomrc, line 34: We don't have the 2 steps compilation anymore
- from src to deploy
- and from deploy to aws.

We now just have a single compilation target called `dist` (the `aws` folder renamed to be more generic). All the sources are compiled there

test/unit_tests/view_model/bindingsSpec.js
```diff
@@ -30,17 +36,15 @@ describe('ko.bindingHandlers', () => {
...
+      handler = {on_enter: () => 'yay'};
 
       // we need the callFake since the spyOn will overwrite the on_enter property
       spyOn(handler, 'on_enter').and.callFake(() => () => 'yay');
 
       binding.init(element, handler.on_enter);
     });
 
-    it('can execute callback when enter is pressed', () => {
+    xit('can execute callback when enter is pressed', () => {
```
Yserz: Is this exclusion on purpose?

webpack.config.common.js
```diff
@@ -70,8 +76,22 @@ module.exports = {
...
   },
-  plugins: [new webpack.IgnorePlugin(/^.\/locale$/, /moment$/)],
+  plugins: [
+    new webpack.IgnorePlugin(/^.\/locale$/, /moment$/),
+    new CopyWebpackPlugin([{from: './node_modules/@wireapp/protocol-messaging/proto/messages.proto', to: 'proto'}]),
+    new webpack.ProvidePlugin({
+      $: 'jquery',
+      _: 'underscore',
+      jQuery: 'jquery',
+      jQuery: 'jquery',
```
Yserz: duplicate

test/unit_tests/view_model/bindingsSpec.js
atomrc: Yeah the problem of this one is a bit deep. 
Since we include `knockout` on the test side AND on the code side, they are 2 different knockout.
Thus the code cannot detect that an observable is an observable because it checks for the prototype which are 2 different object.

Previously we use the knockout that is on `window`. 
This will be fixed once we don't use the window.knockout at all anymore

atomrc: Edit : this one is with jQuery (same problem than explained earlier). I have another test ignored with `ko` also

atomrc: I'll add a comment

.gitignore
```diff
@@ -1,42 +1,17 @@
...
 /doc
 /docs/auth-coverage
 /docs/coverage
 /keys
-/nbproject
-/node_modules
+node_modules
 /package-lock.json
 /tags
-/temp
```
Yserz: `temp` should be ignored and is application specific (we put translations in there)

Yserz:
Livin' it!

PR #5203, opened 2018-11-23
feat: Catch low disk space errors on login (WEBAPP-5719)

bennyn:


app/script/strings.ts
```diff
@@ -523,6 +524,10 @@ export const errorHandlerStrings = defineMessages({
     defaultMessage: 'Something went wrong. Please contact your team administrator for details (Error 0).',
     id: 'BackendError.LABEL.SSO_GENERIC_ERROR',
   },
+  [LabeledError.GENERAL_ERRORS.LOW_DISK_SPACE]: {
+    defaultMessage: 'Not enough quota',
```
ffflorian: Hmm, everywhere it says "Low disk space" but here you use the term "quota". I'd rather see a message like "Not enough disk space available" or something like this.

app/script/auth/module/action/LabeledError.ts
```diff
@@ -0,0 +1,42 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export default class LabeledError extends Error {
+  public label: string;
+
+  constructor(label: string, message: string);
+  constructor(label: string, error: Error);
+  constructor(label: string, messageOrError: string | Error) {
```
ffflorian, line 25: 👍

app/script/strings.ts
bennycode: Message is just temporary. Astrid and Roger will provide something nicer.

PR #5210, opened 2018-11-27
fix: remove wrong sdp rewrite offer condition (WEBAPP-5738)

atomrc:


app/script/calling/SDPMapper.js
```diff
@@ -85,13 +85,9 @@ z.calling.SDPMapper = {
 
     if (isFirefox) {
       if (isLocalSdp) {
-        sessionDescription = isOffer
```
ffflorian: If we are not using `isOffer` anymore, we can remove it.

PR #5238, opened 2018-11-29
refactor: Migrate PropertiesRepository to module system

bennyn:


karma.conf.js
```diff
@@ -106,6 +106,7 @@ module.exports = function(config) {
     },
 
     webpackMiddleware: {
+      logLevel: 'error',
```
ffflorian, line 110: Hm, is this part of the PR? The [default](https://github.com/webpack/webpack-dev-middleware#loglevel) is `'info'`.

bennycode: It's good to have: https://stackoverflow.com/a/53543001/451634

PR #5240, opened 2018-11-29
fix: Removal of Pako (WEBAPP-4292)

pwnsdx:
Removing Pako as:
- The browser already manage compression
- See ticket WEBAPP-4292


ffflorian:
@pwnsdx can you please update the yarn lockfile before merging? :slightly_smiling_face:

PR #5254, opened 2018-11-30
feat: Insert system message when the read receipt setting changes (WEBAPP-5757)

AndyLnd:


app/script/entity/message/Message.js
```diff
@@ -350,4 +350,7 @@ z.entity.Message = class Message {
     }
     return false;
   }
-};
+}
+
+export default Message;
+z.entity.Message = Message;
```
ffflorian, line 356: Is this still needed?

AndyLnd: Yes. All other message classes still extend from the globally scoped message class.

PR #5275, opened 2018-12-03
feat: Add account level setting for read receipts (WEBAPP-5751)

bennyn:


app/script/properties/PropertiesRepository.js
```diff
@@ -98,22 +104,43 @@ class PropertiesRepository {
...
+      .getPropertiesByKey(property.key)
+      .then(value => {
+        this.setProperty(property.key, value);
       })
-      .then(() => this._publishProperties());
+      .catch(() => {
+        this.logger.warn(
+          `Property "${property.key}" doesn't exist for this account. Continuing with the default value of "${
+            property.defaultValue
+          }".`
```
ffflorian: Can we create a variable with the message beforehand? Makes it more readable :slightly_smiling_face:

app/script/user/UserRepository.js
```diff
@@ -32,21 +32,14 @@ z.user.UserRepository = class UserRepository {
...
-   * @class z.user.UserRepository
-   * @param {z.user.UserService} user_service - Backend REST API user service implementation
-   * @param {z.assets.AssetService} asset_service - Backend REST API asset service implementation
-   * @param {z.self.SelfService} selfService - Backend REST API self service implementation
-   * @param {z.client.ClientRepository} client_repository - Repository for all client interactions
-   * @param {z.time.ServerTimeRepository} serverTimeRepository - Handles time shift between server and client
-   */
-  constructor(user_service, asset_service, selfService, client_repository, serverTimeRepository) {
-    this.user_service = user_service;
+  constructor(user_service, asset_service, selfService, client_repository, serverTimeRepository, propertyRepository) {
```
ffflorian, line 45: Why did you remove the JSDoc?

app/script/properties/PropertiesRepository.js
```diff
@@ -98,22 +104,43 @@ class PropertiesRepository {
     return this.selfUser().isTemporaryGuest() ? this._initTemporaryGuestAccount() : this._initActivatedAccount();
   }
 
-  _initActivatedAccount() {
+  _fetchWebAppAccountSettings() {
+    return this.propertiesService
+      .getPropertiesByKey(PropertiesRepository.CONFIG.WEBAPP_ACCOUNT_SETTINGS)
+      .then(properties => {
+        $.extend(true, this.properties, properties);
```
atomrc, line 113: Is the `deep` boolean really important here? 
Else we could remove the call to jquery and use native `Object.assign`

app/script/user/UserRepository.js
bennycode: It's back

app/script/properties/PropertiesRepository.js
bennycode: I tried with `Object.assign` but we really need the deep merge here. In the future the whole `PropertiesRepository` will get an overhaul and then we can look again at getting rid off `$.extend`.

app/script/properties/PropertiesRepository.js
```diff
@@ -34,6 +39,7 @@ class PropertiesRepository {
 
     this.properties = new z.properties.PropertiesEntity();
     this.selfUser = ko.observable();
+    this.enableReadReceipts = ko.observable(PropertiesRepository.CONFIG.ENABLE_READ_RECEIPTS.defaultValue);
```
atomrc, line 44: Let's use the `propertiesEntity` that is responsible for holding all the user's properties

app/script/properties/PropertiesRepository.js
atomrc: :+1:

app/script/properties/PropertiesRepository.js
```diff
@@ -168,9 +194,38 @@ class PropertiesRepository {
...
+        break;
+    }
+  }
+
+  // Map a property and set it into our state
+  setProperty(key, value) {
+    switch (key) {
+      case PropertiesRepository.CONFIG.ENABLE_READ_RECEIPTS.key:
+        value = JSON.parse(value);
+        this.logger.info(`Setting key "${key}" to value "${value}".`);
```
AndyLnd: Do you want to move this line out of the switch/case? So it will show up for all (future) cases.

PR #5277, opened 2018-12-03
fix: Add Chromium to supported browsers list

Yserz:


ffflorian:
Can we rename this to something like "chore: Bump @wireapp/commons to 0.10" and add the fix message in the description?

PR #5282, opened 2018-12-03
refactor: Switch to WIRE_RECEIPT_MODE

bennyn:


app/script/view_model/content/PreferencesAccountViewModel.js
```diff
@@ -384,7 +384,8 @@ z.viewModel.content.PreferencesAccountViewModel = class PreferencesAccountViewMo
 
   onReadReceiptsChange(viewModel, event) {
     const enableReadReceipts = event.target.checked;
-    this.propertiesRepository.saveReadReceipts(enableReadReceipts);
+    const receiptMode = enableReadReceipts ? 1 : 0;
+    this.propertiesRepository.saveReadReceipts(receiptMode);
```
atomrc: We should probably update the name of the method on the `propertieRepo` too.
How about `saveReceiptMode`?

PR #5283, opened 2018-12-03
feat: Parse read receipts and update local messages

atomrc:


app/script/event/preprocessor/ReadReceiptMiddleware.js
```diff
@@ -0,0 +1,63 @@
...
+    this.eventService = eventService;
+    this.logger = new z.util.Logger('ReadReceiptMiddleware', z.config.LOGGER.OPTIONS);
+  }
+
+  /**
+   * Handles validation of the event if it contains a quote.
+   * If the event does contain a quote, will also decorate the event with some metadata regarding the quoted message
+   *
+   * @param {Object} event - event in the DB format
+   * @returns {Object} event - the original event if no quote is found (or does not validate). The decorated event if the quote is valid
```
bennycode: It returns a `Promise<Object>`, doesn't it?

test/unit_tests/event/preprocessor/ReadReceiptMiddlewareSpec.js
```diff
@@ -0,0 +1,75 @@
...
+  const noop = () => {};
+  let readReceiptMiddleware;
+  const eventService = {loadEvent: noop, replaceEvent: noop};
+
+  beforeEach(() => {
+    readReceiptMiddleware = new ReadReceiptMiddleware(eventService);
+  });
+
+  describe('processEvent', () => {
+    it('Ignores read receipt for which original message is not found', () => {
```
bennycode: `Ignores` with lowercase `i`, so that it can be read like `it ignores`

test/unit_tests/event/preprocessor/ReadReceiptMiddlewareSpec.js
```diff
@@ -0,0 +1,75 @@
...
+      spyOn(eventService, 'loadEvent').and.returnValue(Promise.resolve(undefined));
+      spyOn(eventService, 'replaceEvent');
+
+      const event = {
+        conversation: 'conversattionid',
```
bennycode: `conversattionid` -> `conversationid`

I would also put a real UUIDv4 there instead of `conversattionid` so that it looks more like an actual payload from the real world.

app/script/event/preprocessor/ReadReceiptMiddleware.js
atomrc: Very true, sir

PR #5297, opened 2018-12-04
chore: Disable flaky camera tests (WEBAPP-5772)

AndyLnd:


bennycode:
We only need to disable one of them. I will fix it.

PR #5304, opened 2018-12-04
Feat: Message read indicator (WEBAPP-5770)

AndyLnd:
There are still some conditions that need to be handled once everything is implemented.


app/script/components/message.js
```diff
@@ -66,6 +66,23 @@ class Message {
     this.showLikes = ko.observable(false);
 
     this.bindShowMore = this.bindShowMore.bind(this);
+
+    this.readReceiptTooltip = ko.pureComputed(() => {
+      const receipts = this.message.readReceipts();
+      if (!receipts.length || !this.conversation().is1to1()) {
+        return '';
+      }
+      return moment(receipts[0].time).format('DD.MM.YY · HH:mm');
```
atomrc, line 77: Could you explicitly import `moment` in that file, please.

app/script/view_model/bindings/MessageListBindings.js
```diff
@@ -60,17 +60,15 @@ ko.bindingHandlers.focus_on_keydown = {
...
-        .getBoundingClientRect();
-      const show_timestamps = event.clientX > rect.right - 64 && event.clientX < rect.right;
-      $('.time').toggleClass('show-timestamp', show_timestamps);
-    });
+    const toggleShowTimeStamp = force => {
+      const times = document.querySelectorAll('.time');
+      times.forEach(time => time.classList.toggle('show-timestamp', force));
+    };
+
+    element.addEventListener('mousemove', () => toggleShowTimeStamp(true));
```
atomrc: I believe `mousemove` is a quite expensive event (firing a looottt). Maybe we should debounce it.

PR #5307, opened 2018-12-04
feat: Make read receipt setting based on account preference (WEBAPP-5762)

bennyn:
Test Code:

```js
wire.app.repository.conversation.shouldSendReadReceipt(wire.app.repository.conversation.active_conversation());
```

### DIFF
--- a/app/script/user/UserRepository.js
+++ b/app/script/user/UserRepository.js
@@ -55,9 +55,6 @@ z.user.UserRepository = class UserRepository {
     this.should_set_username = false;
 
     this.self = ko.observable();
-    this.self.subscribe(user => {
-      user.receiptMode = this.propertyRepository.receiptMode;
-    });
     this.users = ko.observableArray([]);
 
     this.connect_requests = ko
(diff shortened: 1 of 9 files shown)


app/script/user/UserRepository.js
```diff
@@ -55,6 +55,9 @@ z.user.UserRepository = class UserRepository {
     this.should_set_username = false;
 
     this.self = ko.observable();
+    this.self.subscribe(user => {
+      user.receiptMode = this.propertyRepository.receiptMode;
```
atomrc: I don't think it's a good idea to create multiple entry point to the `receiptMode` value.
We are, kind of, creating two mutable sources of truth of the value which can lead to confusion in the future.

I would rather use the `propertyRepository` as the unique source of truth for that information and do the check whether we should send the receipt or not in an upper layer of the application (not on the entity level)

bennycode: Ok, I can make the check inside of the `ConversationRepository`.

PR #5330, opened 2018-12-06
feat: Show read receipt system message on group creation (WEBAPP-5783)

bennyn:


app/script/components/message.js
```diff
@@ -66,6 +66,9 @@ class Message {
     this.actionsViewModel = actionsViewModel;
 
     this.showLikes = ko.observable(false);
+    this.hasReadReceiptsTurnedOn =
+      this.conversationRepository.active_conversation() &&
```
atomrc: Let's use the `conversation` parameter that is injected in the constructor instead of relying on the current active convo (which might not be the one linked to this message at some point in time)

atomrc:
```suggestion
this.hasReadReceiptsTurnedOn = conversation.expectsReadConfirmation()
```

PR #5331, opened 2018-12-06
feat: Enable read receipt for all supported message types (WEBAPP-5782)

atomrc:


app/script/conversation/EventMapper.js
```diff
@@ -725,4 +714,17 @@ z.conversation.EventMapper = class EventMapper {
 
     return assetEntity;
   }
-};
+}
+
+function addReadReceiptData(entity, event) {
+  const {data: eventData, read_receipts} = event;
+  if (eventData.expects_read_confirmation) {
+    entity.expectsReadConfirmation = eventData.expects_read_confirmation;
```
AndyLnd: Is this condition really necessary?

atomrc: fair point

PR #5341, opened 2018-12-07
fix: Group message timer cannot be turned off (WEBAPP-5794)

AndyLnd:


ffflorian:
This is a `runfix`, isn't it?

PR #5360, opened 2018-12-10
fix: Don't receive receipts for other user's messages (WEBAPP-5801)

AndyLnd:


app/script/conversation/EventMapper.js
```diff
@@ -717,6 +717,11 @@ export default class EventMapper {
 }
 
 function addReadReceiptData(entity, event) {
+  const isFromMe = entity.user && entity.user().is_me;
```
atomrc: This will not work (user will not be set during the lifetime of the event in the `EventMapper`)

PR #5376, opened 2018-12-11
runfix: Add flag to message to determine if it was expecting read receipts

bennyn:
This PR fixes the following use case.

**Scenario**

- Read receipts are turned ON
- User A goes offline
- User B sends message 1
- User B sends message 2
- User B sends message 3
- User B turns read receipts OFF
- User B sends message 4
- User B turns read receipts ON
- User B sends message 5
- User A goes online

**Expected result**

- User A replies with a read receipt to message 1, 2, 3 & 5
- User A does not reply with a read receipt to message 4

### DIFF
--- a/app/script/conversation/ConversationRepository.js
+++ b/app/script/conversation/ConversationRepository.js
@@ -3819,7 +3819,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
     if (conversationEntity.is1to1()) {
       return !!this.propertyRepository.receiptMode();
     }
-    return conversationEntity.expectsReadConfirmation();
+    return true;
   }
 
   /**
--- a/app/script/view_model/content/MessageListViewModel.js
+++ b/app/script/view_model/content/MessageListViewModel.js
@@ -32,7 +32,7 @@ window.z.viewModel.content = z.viewModel.content || {};
  * Message list rendering view model.
  *
  * @todo Get rid of the participants dependencies whenever bubble implementation has changed
- * @todo Remove all jquery selectors
+ * @todo Remove all jQuery selectors
  */
 z.viewModel.content.MessageListViewModel = class MessageListViewModel {
   constructor(mainViewModel, contentViewModel, repositories) {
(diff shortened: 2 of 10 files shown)


app/script/conversation/ConversationRepository.js
```diff
@@ -1817,7 +1817,7 @@ z.conversation.ConversationRepository = class ConversationRepository {
           protoAsset.set(z.cryptography.PROTO_MESSAGE_TYPE.ASSET_PREVIEW, assetPreview);
           protoAsset.set(
             z.cryptography.PROTO_MESSAGE_TYPE.EXPECTS_READ_CONFIRMATION,
-            this.shouldSendReadReceipt(conversationEntity)
+            this.expectReadReceipt(conversationEntity)
```
atomrc: we could use `option.expectsReadConfirmation` here, since we already have computed that value earlier.

app/script/view_model/content/MessageListViewModel.js
```diff
@@ -636,7 +636,7 @@ z.viewModel.content.MessageListViewModel = class MessageListViewModel {
     const isUnreadMessage = messageTimestamp > conversationTimestamp;
     const isNotOwnMessage = !messageEntity.user().is_me;
     const expectsConfirmation = messageEntity.expectsReadConfirmation;
-    const shouldSendReadReceipt = this.conversation_repository.shouldSendReadReceipt(conversationEntity);
+    const shouldSendReadReceipt = this.conversation_repository.expectReadReceipt(conversationEntity);
```
atomrc: We should remove this condition here. This is checked at receiving time now, no need to recheck at viewing time.

I think with that, the following scenario is not working:
- RR are ON
- A logs out
- B sends msg
- B turns RR OFF
- A logs in

=> I believe A will not send a RR (because the last state of the convo for RR is OFF)

atomrc: Now I remember why we needed this method to return `true` in case of a group + team conversation.

We might want to create another method then...

bennycode: Ok, let's talk about creating another method.

PR #5415, opened 2018-12-13
feat: Dark Mode

pwnsdx:


app/style/common/mixins.less
```diff
@@ -17,6 +17,38 @@
...
+    .iter(length(@list));
+    .iter(@i) when (@i > 0) {
+      .iter(@i - 1);
+      @pair: extract(@list, @i);
+      @key: extract(@pair, 1);
+      --@{key}: ~'@{@{key}}';
+    }
+  }
+
+  .setVariables(primary secondary);
```
AndyLnd: I am still not too fond of the naming here. Those aren't `primary` and `secondary` colors, but more like `background` and `foreground`. `primary` and `secondary` can be confusing later on.

app/script/view_model/ThemeViewModel.js
```diff
@@ -0,0 +1,51 @@
...
+    );
+  }
+
+  setTheme(newTheme) {
+    const classesWithoutTheme = document.body.className
+      .split(' ')
+      .filter(elementClass => !elementClass.startsWith(THEMES_CLASS_PREFIX))
+      .join(' ');
+    const theme = `${THEMES_CLASS_PREFIX}${newTheme}`;
+    document.body.className = `${classesWithoutTheme} ${theme}`;
```
AndyLnd:
```
     const classes = document.body.className
      .split(' ')
      .filter(elementClass => !elementClass.startsWith(THEMES_CLASS_PREFIX))
      .concat(`${THEMES_CLASS_PREFIX}${newTheme}`);
    document.body.className = classes.join(' ');
```
might be a bit more straight forward

PR #5422, opened 2018-12-14
runfix: Keep read receipts on edited messages (WEBAPP-5825)

ffflorian:


bennyn:
Let's add a test case. 💪

PR #5431, opened 2018-12-14
feat: Feature flag ENABLE_USERNAME_LOGIN (WEBAPP-5675)

Yserz:


Yserz:
Turns out we don't need it ATM. Let's keep this for a while and then close it if we really don't need it.

bennycode:
> Turns out we don't need it ATM. Let's keep this for a while and then close it if we really don't need it.

Time has told us to close this issue. Let's take a minute of silence before we continue with our regular work. 🕒

PR #5436, opened 2018-12-14
runfix: Add phone login flag to main login

Yserz:


Yserz:
Turns out we don't need it ATM. Let's keep this for a while and then close it if we really don't need it.

PR #5453, opened 2018-12-17
runfix: Point to local unsupported page instead of website

Yserz:


package.json
```diff
@@ -22,6 +22,7 @@
     "jszip": "3.1.5",
     "keyboardjs": "2.5.1",
     "knockout": "3.4.2",
+    "lint-staged": "8.1.0",
```
Yserz, line 25: I had an issue with this dependency ending up reinstalling it. https://github.com/okonet/lint-staged/issues/533

PR #5517, opened 2018-12-21
refactor: Use dexie-observable to handle message deletion and update

atomrc:


app/script/storage/StorageSchemata.js
```diff
@@ -310,6 +310,10 @@ z.storage.StorageSchemata = class StorageSchemata {
         },
         version: 15,
       },
+      {
+        schema: {},
```
atomrc: We need an empty change in the DB cause dexie-observable is creating some extra tables and indexes

app/script/storage/StorageService.js
```diff
@@ -72,42 +78,25 @@ class StorageService {
...
-            );
-        });
-      }
-    };
-    if (!this.db) {
-      // waiting for the DB to be initialized
-      return (this._afterDbInit = initHook);
-    }
-    initHook();
+    this.dbListeners.push({callback, store: storeName, type: 2});
```
bennycode: What's `type: 2`?

atomrc: This is the type of change we want to listen to. These are value sent by Dexie.
Do you think I should create a named variable to make that more explicit?

bennycode: That would be great.

PR #5523, opened 2019-01-02
feat: Enable TURN servers logging when fetching calling config

atomrc:


app/script/calling/CallingRepository.js
```diff
@@ -1604,8 +1604,10 @@ z.calling.CallingRepository = class CallingRepository {
         const expirationDate = new Date(Date.now() + timeout);
         callingConfig.expiration = expirationDate;
 
-        const logMessage = `Updated calling configuration expires on '${expirationDate.toISOString()}'`;
-        this.callLogger.info(logMessage, callingConfig);
+        const turnServersConfig = (callingConfig.ice_servers || []).map(server => server.urls.join('\n')).join('\n');
+        const logMessage = `Updated calling configuration expires on '${expirationDate.toISOString()}' with servers :
```
ffflorian:
```suggestion
        const logMessage = `Updated calling configuration expires on '${expirationDate.toISOString()}' with servers:
```

app/script/calling/CallingRepository.js
```diff
@@ -1604,8 +1604,10 @@ z.calling.CallingRepository = class CallingRepository {
         const expirationDate = new Date(Date.now() + timeout);
         callingConfig.expiration = expirationDate;
 
-        const logMessage = `Updated calling configuration expires on '${expirationDate.toISOString()}'`;
-        this.callLogger.info(logMessage, callingConfig);
+        const turnServersConfig = (callingConfig.ice_servers || []).map(server => server.urls.join('\n')).join('\n');
```
ffflorian, line 1607: Shouldn't we rather join the server array with `, `? That way the log won't be so large.

atomrc: It makes the log way easier to read though, you can easily identify which are tpc and udp servers with the `\n`

PR #5539, opened 2019-01-03
fix: Typecasting for incoming login errors

bennyn:
The `catch` clause receives all kinds of errors but from the implementation it thinks that they are all of kind `BackendError` and do have a `label` which is not true. 

So far the TypeScript code assumed that error is of type `any` which we always want to avoid, so I type-casted the errors because I ran into a scenario where I received an error in the `catch` clause which did NOT have a `label` and caused another error because the `default` case of the `switch` block failed when executing `backendError.label.endsWith(errorType)`.

### DIFF
--- a/app/script/auth/page/Login.tsx
+++ b/app/script/auth/page/Login.tsx
@@ -236,39 +236,44 @@ class Login extends React.Component<CombinedProps, State> {
           : this.props.doLogin(login);
       })
       .then(this.navigateChooseHandleOrWebapp)
-      .catch(error => {
-        switch (error.label) {
-          case BackendError.LABEL.NEW_CLIENT: {
-            this.props.resetAuthError();
-            /**
-             * Show history screen if:
-             *   1. database contains at least one event
-             *   2. there is at least one previously registered client
-             *   3. new local client is temporary
-             */
-            return this.props.doGetAllClients().then(clients => {
-              const shouldShowHistoryInfo = this.props.hasHistory || clients.length > 1 || !this.state.persist;
-              return shouldShowHistoryInfo
-                ? this.props.history.push(ROUTE.HISTORY_INFO)
-                : this.navigateChooseHandleOrWebapp();
-            });
-          }
-          case BackendError.LABEL.TOO_MANY_CLIENTS: {
-            this.props.resetAuthError();
-            return this.props.history.push(ROUTE.CLIENTS);
-          }
-          case BackendError.LABEL.INVALID_CREDENTIALS:
-          case LabeledError.GENERAL_ERRORS.LOW_DISK_SPACE: {
-            return;
-          }
-          default: {
-            const isValidationError = Object.values(ValidationError.ERROR).some(errorType =>
-              error.label.endsWith(errorType)
-            );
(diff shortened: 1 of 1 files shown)


app/script/auth/page/Login.tsx
```diff
@@ -236,39 +236,44 @@ class Login extends React.Component<CombinedProps, State> {
...
-          }
-          default: {
-            const isValidationError = Object.values(ValidationError.ERROR).some(errorType =>
-              error.label.endsWith(errorType)
-            );
-            if (!isValidationError) {
-              throw error;
+      .catch((error: Error | BackendError) => {
+        if ((error as BackendError).label) {
+          const backendError: BackendError = error as BackendError;
```
ffflorian:
```suggestion
          const backendError = error as BackendError;
```

app/script/auth/page/Login.tsx
```diff
@@ -236,39 +236,44 @@ class Login extends React.Component<CombinedProps, State> {
...
+              this.props.resetAuthError();
+              return this.props.history.push(ROUTE.CLIENTS);
+            }
+            case BackendError.LABEL.INVALID_CREDENTIALS:
+            case LabeledError.GENERAL_ERRORS.LOW_DISK_SPACE: {
+              return;
+            }
+            default: {
+              const isValidationError = Object.values(ValidationError.ERROR).some(errorType =>
+                backendError.label.endsWith(errorType)
```
atomrc, line 268: No sure I see the fix here. Because you cast it in TypeScript doesn't mean you will have a `label` on your error at runtime. 
Am I missing something here?

app/script/auth/page/Login.tsx
```diff
@@ -236,39 +236,44 @@ class Login extends React.Component<CombinedProps, State> {
...
-          case LabeledError.GENERAL_ERRORS.LOW_DISK_SPACE: {
-            return;
-          }
-          default: {
-            const isValidationError = Object.values(ValidationError.ERROR).some(errorType =>
-              error.label.endsWith(errorType)
-            );
-            if (!isValidationError) {
-              throw error;
+      .catch((error: Error | BackendError) => {
```
atomrc, line 239: If you accept Error and BackendError as parameter of that function, you should handle both case

app/script/auth/page/Login.tsx
bennycode: It crashed on `backendError.label.endsWith(errorType)` when `backendError` didn't have a `label`. 

The fix for this is some lines above where I do the following ` if ((error as BackendError).label) `.

app/script/auth/page/Login.tsx
bennycode: I handle both cases. If it's a `BackendError` (detected by the `label` property), then it goes into the `switch` block. If it is a usual `Error`, then this error is thrown.

app/script/auth/page/Login.tsx
atomrc: Ah true, sorry I read this line too fast, I though it was the `switch` statement. My mistake.

PR #5589, opened 2019-01-09
fix: Background color of the emoji modal (WEBAPP-5871)

pwnsdx:


AndyLnd:
Does the highlighting still look okay?

ghost:
> Does the highlighting still look okay?

Yes

PR #5603, opened 2019-01-10
fix: Sync account preferences (WEBAPP-5846)

bennyn:


test/unit_tests/user/UserRepositorySpec.js
```diff
@@ -33,7 +33,127 @@ describe('z.user.UserRepository', () => {
...
+              notifications: 'on',
+              permissions: {
+                notifications: 'default',
+              },
+              previews: {
+                send: true,
+              },
+              privacy: {
+                enable_read_receipts: false,
+                improve_wire: true,
```
atomrc, line 51: You should limit the data in the test to the data actually turning on error reporting. namely the path `value.privacy.improve_wire`. 
We don't need to account for the other values here

src/script/properties/PropertiesRepository.js
```diff
@@ -208,10 +208,16 @@ class PropertiesRepository {
 
   // Map a property and set it into our state
   setProperty(key, value) {
-    this.logger.info(`Setting key "${key}" to value "${value}"...`);
+    this.logger.info(`Setting key "${key}"...`, value);
+
     switch (key) {
+      case PropertiesRepository.CONFIG.WEBAPP_ACCOUNT_SETTINGS:
+        if (this.properties.version === value.version) {
+          this.properties = {...this.properties, ...value};
```
atomrc: Object spread is not supported in the webapp. You need to use `Object.assign` here

test/unit_tests/user/UserRepositorySpec.js
bennycode: I can reduce the values but then we won't have an original payload from the real world anymore which means that we test with synthetic values. I can also not reduce every value because some other parts of the code check for the attributes and will fail if they are not set which affects my test case.

PR #5611, opened 2019-01-11
fix: Reset ice candidate generation attemps when sending SDP 

atomrc:


src/script/calling/entities/FlowEntity.js
```diff
@@ -901,6 +903,7 @@ z.calling.entities.FlowEntity = class FlowEntity {
             obfuscated: [
               transformedSdp.type,
               iceCandidates.length,
+              iceCandidateTypesLog,
```
atomrc, line 906: I also fixed the production log to include the types of ice candidates that we send

PR #5635, opened 2019-01-17
refactor: Migrate team permission utils to modules

atomrc:
This is only a refactoring PR to prepare for the next feature. 

It doesn't include any behavior change. 

I bootstrapped tests for the Permission handler and refactored the tested method. 

I also resolved a cyclic dependency between `permission` and `role` handlers. We might want to re-split those later on but let's do it properly when we can take a step back on what we need exactly.


src/script/team/TeamPermission.js
```diff
@@ -17,89 +17,101 @@
  *
  */
 
-window.z = window.z || {};
-window.z.team = z.team || {};
+/**
+ * Enum for different team permissions.
+ * @returns {z.team.TeamPermission.PERMISSION} Enum of team permissions
+ */
+const PERMISSION = {
```
AndyLnd, line 25: Can we sort that by value and not alphabetical, please?

src/script/team/TeamPermission.js
```diff
@@ -17,89 +17,101 @@
...
+
+  return detectedRole || ROLE.INVALID;
+}
+
+function combinePermissions(permissions) {
+  let result = 0;
+  for (const permission of permissions) {
+    result = result | permission;
+  }
+  return result;
```
AndyLnd: ```return permissions.reduce((acc, permission) => acc | permission, 0);```

test/unit_tests/team/TeamPermissionSpec.js
```diff
@@ -0,0 +1,47 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {ROLE, roleFromPermissions} from 'src/script/team/TeamPermission';
+
+const memberPermissionBitmask = 1587;
```
AndyLnd: Can we please write those in binary notation, so they make more sense as a bitmask?
```suggestion
const memberPermissionBitmask = 0b11000110011;
```

test/unit_tests/team/TeamPermissionSpec.js
```diff
@@ -0,0 +1,47 @@
...
+const ownerPermissionBitmask = 8191;
```
AndyLnd:
```suggestion
const ownerPermissionBitmask = 0b1111111111111;
```

test/unit_tests/team/TeamPermissionSpec.js
```diff
@@ -0,0 +1,47 @@
...
+const adminPermissionBitmask = 5951;
```
AndyLnd:
```suggestion
const adminPermissionBitmask = 0b1011100111111;
```

src/script/team/TeamPermission.js
atomrc: I wouldn't say value is a better sorting algo than feature name....
Why do you feel it's better?

AndyLnd: Maybe it's just my OCD kicking in, but I feel that with bitflags the value is way more important than the name that you chose for it.

PR #5653, opened 2019-01-18
refactor: Adjusted Dockerfile and configure.js to use external directories if necessary

tiago-loureiro:
This changeset makes it easier for build systems to pull configs from the local filesystem if given


tiago-loureiro:
> Note: configure.js will be removed once wireapp/wire-web-packages#1477 is in place. The environment variables will stay the same, but you might need some additional variables then.

Will that be across all web apps?

ffflorian:
> Will that be across all web apps?

Yes

PR #5688, opened 2019-01-24
fix: Improve computation of asset load progress

atomrc:


src/script/components/asset/fileAsset.js
```diff
@@ -120,7 +106,7 @@ ko.components.register('file-asset', {
   `,
   viewModel: {
     createViewModel(params, component_info) {
-      return new z.components.FileAssetComponent(params, component_info);
+      return new FileAssetComponent(params, component_info);
```
ffflorian: Maybe rename `component_info` to `componentInfo`? Not a must, but a nice-to-have.

PR #5708, opened 2019-01-29
Detect network changes (e.g. establishing VPN connection)

maximbaz:
This implements client-side support for https://github.com/wireapp/wire-server/pull/561 and fixes https://github.com/wireapp/wire-desktop/issues/1129

Finally no need to restart Wire after connecting to VPN! 🎉

### DIFF
--- a/src/script/event/WebSocketService.js
+++ b/src/script/event/WebSocketService.js
@@ -41,7 +41,7 @@ z.event.WebSocketService = class WebSocketService {
 
   static get CONFIG() {
     return {
-      PING_INTERVAL: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 30,
+      PING_INTERVAL: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 5,
       PING_INTERVAL_THRESHOLD: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 2,
       RECONNECT_INTERVAL: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 15,
     };
@@ -64,7 +64,7 @@ z.event.WebSocketService = class WebSocketService {
     this.onNotification = undefined;
 
     this.pingIntervalId = undefined;
-    this.lastPingTime = undefined;
+    this.lastPongTime = undefined;
 
     this.reconnectTimeoutId = undefined;
     this.reconnectCount = 0;
@@ -117,13 +117,19 @@ z.event.WebSocketService = class WebSocketService {
 
       this.socket.onmessage = event => {
         if (this._pingHasExperiencedSuspiciousInactivity()) {
-          const secondsSinceLastPing = (Date.now() - this.lastPingTime) / z.util.TimeUtil.UNITS_IN_MILLIS.SECOND;
-          this.logger.warn(`Message received but ping was inactive for "${secondsSinceLastPing}" sec, reconnecting.`);
+          const secondsSinceLastPong = (Date.now() - this.lastPongTime) / z.util.TimeUtil.UNITS_IN_MILLIS.SECOND;
+          this.logger.warn(`Message received but there was no pong for "${secondsSinceLastPong}" sec, reconnecting.`);
           return this.reconnect(WebSocketService.CHANGE_TRIGGER.LONG_INACTIVITY);
         }
         if (event.data instanceof Blob) {
           const blobReader = new FileReader();
-          blobReader.onload = () => onNotification(JSON.parse(blobReader.result));
+          blobReader.onload = () => {
+            if (blobReader.result === 'pong') {
+              this.lastPongTime = new Date();
+            } else {
+              onNotification(JSON.parse(blobReader.result));
+            }
+          };
           blobReader.readAsText(event.data);
         }
       };
@@ -199,7 +205,7 @@ z.event.WebSocketService = class WebSocketService {
       this.socket.close();
       window.clearInterval(this.pingIntervalId);
       window.clearTimeout(this.reconnectTimeoutId);
-      this.lastPingTime = undefined;
+      this.lastPongTime = undefined;
     }
 
     if (reconnect) {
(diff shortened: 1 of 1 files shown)


src/script/event/WebSocketService.js
```diff
@@ -41,7 +41,7 @@ z.event.WebSocketService = class WebSocketService {
 
   static get CONFIG() {
     return {
-      PING_INTERVAL: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 30,
+      PING_INTERVAL: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 5,
       PING_INTERVAL_THRESHOLD: z.util.TimeUtil.UNITS_IN_MILLIS.SECOND * 2,
```
max-baz: Please note, I increased the amount of 'ping' requests, because we have to wait for the `PING_INTERVAL_THRESHOLD` before Wire detects network loss, and it was a whole minute. Now it is only 10 seconds, which I personally find acceptable.

The amount of traffic is not increased as much, because we also changed from sending 36 bytes per message (`Wire is so much nicer with internet!`) to only 4 (`ping`).

src/script/event/WebSocketService.js
```diff
@@ -117,13 +117,19 @@ z.event.WebSocketService = class WebSocketService {
...
-          this.logger.warn(`Message received but ping was inactive for "${secondsSinceLastPing}" sec, reconnecting.`);
+          const secondsSinceLastPong = (Date.now() - this.lastPongTime) / z.util.TimeUtil.UNITS_IN_MILLIS.SECOND;
+          this.logger.warn(`Message received but there was no pong for "${secondsSinceLastPong}" sec, reconnecting.`);
           return this.reconnect(WebSocketService.CHANGE_TRIGGER.LONG_INACTIVITY);
         }
         if (event.data instanceof Blob) {
           const blobReader = new FileReader();
-          blobReader.onload = () => onNotification(JSON.parse(blobReader.result));
+          blobReader.onload = () => {
+            if (blobReader.result === 'pong') {
```
atomrc, line 120: Although the idea of listening to `pong` replies instead of sent `ping` is really good, it doesn't seem that the backend is sending any `pong` message. 
Are you sure this condition is met at some point?

max-baz: Of course, see [discussion](https://github.com/wireapp/wire-server/issues/560#issuecomment-458139018) and [implementation](https://github.com/wireapp/wire-server/pull/561) 🙂

atomrc: @maximbaz oh awesome, I missed that discussion :) 

I am not sure on which environment it is right now, and what is the plan for putting that on production. 
@bennyn any insights on this, maybe?

max-baz: The backend part is already on production as per last comment in the discussion thread 🙂 

> @maximbaz now this ticket is legitimately closed, and #561 is in production.

bennycode:
@maximbaz Thanks for the PR. 

Have you considered the following implementation strategy?

1. Build a very small stack in which you store a PING when you send a PING and where you remove a PING if you receive a PONG
2. Send PINGs in a predefined interval and stack them
3. Show the "no internet connection" bar when there are 2 PINGs in the stack (because ideally there can only be one at maximum when the backend responds with a PONG to every outgoing PING)

max-baz:
The current implementation is essentially the same: 

1. PING is being sent every `PING_INTERVAL` (5s)
1. Every time we send a ping we check if a PONG was received within `PING_INTERVAL_THRESHOLD` (which equals to `PING_INTERVAL * 2`)

So essentially when we are about to send the 3rd PING and there still have been no PONG, we show the "no internet connection" bar and attempt to reconnect.

bennycode:
Could the stack idea replace the "received within `PING_INTERVAL_THRESHOLD`" check?

src/script/event/WebSocketService.js
```diff
@@ -215,30 +214,16 @@ z.event.WebSocketService = class WebSocketService {
   sendPing() {
     const isReadyStateOpen = this.socket.readyState === 1;
     if (isReadyStateOpen) {
-      if (this._pingHasExperiencedSuspiciousInactivity()) {
+      if (this.hasAlreadySentUnansweredPing) {
         this.logger.warn('Ping interval check failed');
         return this.reconnect(WebSocketService.CHANGE_TRIGGER.PING_INTERVAL);
```
max-baz, line 219: Done @bennyn, this is now based on a counter (used boolean for readability), a second ping in the stack now causes reconnect as you can see right in this snippet

src/script/event/WebSocketService.js
```diff
@@ -116,14 +114,15 @@ z.event.WebSocketService = class WebSocketService {
       };
 
       this.socket.onmessage = event => {
-        if (this._pingHasExperiencedSuspiciousInactivity()) {
-          const secondsSinceLastPing = (Date.now() - this.lastPingTime) / z.util.TimeUtil.UNITS_IN_MILLIS.SECOND;
-          this.logger.warn(`Message received but ping was inactive for "${secondsSinceLastPing}" sec, reconnecting.`);
-          return this.reconnect(WebSocketService.CHANGE_TRIGGER.LONG_INACTIVITY);
-        }
```
max-baz, line 123: The only downside of this flag-based approach that I can see is that now it's impossible to detect that `pingInterval` has not been started at all, previously based on Dates we could see that PINGs were not even being sent, now we can't. I don't immediately see how this situation can occur, just pointing out the difference 🙂

bennycode: Great. Thanks for taking the time rewriting it. The code is easier to read now and removes more lines than it adds. I like. 😺

bennycode: I think you can see that pings are not send when [inspecting the WebSocket](https://kaazing.com/inspecting-websocket-traffic-with-chrome-developer-tools/).

src/script/event/WebSocketService.js
atomrc: Quick question that I cannot seem to answer just looking at the code. 
Will the non-time-based solution work in case of the laptop being put in sleep mode?

Previously the time would also tell us if the computer is not active (and thus not sending any ping). This was why time is an important component here. 
I believe with this solution, if the computer is put to sleep we will not reconnect automatically. 

Now this scenario might actually be covered by the fact that we are going to send pings that are not going to be replied to. Can you confirm this @maximbaz ?

max-baz: I've just confirmed that both putting computer to sleep and simply shutting down wifi doesn't affect Wire, when Internet is back Wire does reconnect as expected. However I can also see that Wire is detecting "connection lost" and "connection regained", so when connection is lost, WebsocketService is reset and no pings are being sent. Once "connection regained" occurs, pings are being sent again as expected.

atomrc: Awesome!! This is a great PR @maximbaz Thanks a lot :+1:

PR #5718, opened 2019-02-04
 chore: Happy new year 🎄

lipis:


bennycode:
We don't need to update the copyright: https://stackoverflow.com/a/2391555/451634

PR #5746, opened 2019-02-07
refactor: File-based environment variables

Yserz:


.gitignore
```diff
@@ -20,5 +20,6 @@ CHANGELOG.md
 
 # Configuration imported from https://github.com/wireapp/wire-web-config-default/
 .env
+.env.defaults
```
atomrc, line 23: Am I wrong when I say that this will break our local dev instances?

Yserz: It just got moved to this lovely place: https://github.com/wireapp/wire-web-config-default/blob/staging/wire-webapp/.env.defaults :)

PR #5748, opened 2019-02-07
fix: Skip HTML escapes for notifications

AndyLnd:


test/unit_tests/util/LocalizerUtilSpec.js
```diff
@@ -127,6 +127,21 @@ describe('LocalizerUtil', () => {
       expect(result3).toBe('Hello Rick, you are &lt;script&gt;a pickle&lt;/script&gt;');
     });
 
+    it("doesn't escape substitutions when it should skip the escaping", () => {
```
atomrc, line 130: :+1:

PR #5764, opened 2019-02-08
feat: Ability to display user's profile as read only

atomrc:


src/script/user/UserMapper.js
```diff
@@ -97,6 +97,7 @@ export default class UserMapper {
       assets,
       email,
       expires_at: expirationDate,
+      can_edit_profile,
```
atomrc: This will need to change when the property is available on the backend

src/style/content/preferences/account.less
```diff
@@ -168,11 +168,12 @@
     font-size: 12px;
   }
 
-  &:focus,
-  &:hover {
+  &:not([^readonly]):focus,
```
AndyLnd:
```suggestion
  &:not([readonly]):focus,
```

src/style/content/preferences/account.less
```diff
@@ -168,11 +168,12 @@
...
+  &:not([^readonly]):hover {
```
AndyLnd:
```suggestion
  &:not([readonly]):hover {
```

PR #5765, opened 2019-02-08
refactor: Use logdown instead of console logging

bennyn:


src/script/auth/module/action/SelfAction.ts
```diff
@@ -19,10 +19,13 @@
 
 import {ConsentType, Self} from '@wireapp/api-client/dist/commonjs/self';
 import * as config from '../../config';
+import {getLogger} from '../../LogProvider';
 import {ThunkAction} from '../reducer';
 import {SelfActionCreator} from './creator/';
 
 export class SelfAction {
+  private static readonly LOGGER = getLogger('SelfAction');
```
atomrc: Quick suggestion. The logger could be instantiated in a `const` outside of the class instead of a private static readonly property of the class. 
WDYT?

bennycode: Ok: https://github.com/wireapp/wire-webapp/pull/5765/commits/1625ae16138f74c02777ff12a12d2f42aafcfbab

PR #5767, opened 2019-02-08
chore: Use Babel for complete app

bennyn:
This will give us the possibility to finally use `async` & `await`, [Object destructuring](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and many other sweets while shipping for MS Edge 17.

### DIFF
--- a/src/script/util/util.js
+++ b/src/script/util/util.js
@@ -19,12 +19,12 @@
 
 import {Decoder, Encoder} from 'bazinga64';
 import UUID from 'uuidjs';
-import marked from './marked.js';
 import hljs from 'highlightjs';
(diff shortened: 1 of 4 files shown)


src/script/util/util.js
```diff
@@ -332,7 +332,7 @@ z.util.renderMessage = (message, selfId, mentionEntities = []) => {
       );
     }, message);
 
-  mentionlessText = marked(mentionlessText, {
+  mentionlessText = window.marked(mentionlessText, {
```
atomrc, line 335: Why adding `window` here. Is the `import` broken somehow?

PR #5779, opened 2019-02-11
chore: Trigger staging releases based on tags

bennyn:
[Tagging suggestion](https://github.com/wireapp/wire-webapp/releases/new) from GitHub:

> If the tag isn’t meant for production use, add a pre-release version after the version name. Some good pre-release versions might be v0.2-alpha or v5.9-beta.3.

In the long run we would like to name releases like `v1.5.7-staging` but for now we can go with `2019-02-11-staging`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,8 +88,8 @@ deploy:
     env: wire-webapp-rc
     bucket_name: wire-webapp-builds
     on:
-      repo: wireapp/wire-webapp
-      branch: staging
+      tags: true
+      condition: $TRAVIS_TAG =~ *-staging*
 
   # https://wire-webapp-staging.zinfra.io/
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -88,8 +88,8 @@ deploy:
     env: wire-webapp-rc
     bucket_name: wire-webapp-builds
     on:
-      repo: wireapp/wire-webapp
-      branch: staging
+      tags: true
+      condition: $TRAVIS_TAG =~ *-staging*
```
ffflorian, line 92: But does this really work with Travis?

bennycode: Documentation sounds very promising: https://docs.travis-ci.com/user/deployment#conditional-releases-with-on

PR #5818, opened 2019-02-14
refactor: Expose the dependencies graph from the resolver

atomrc:
The API of the resolver changes a bit.
It now also exports all the dependencies (classes) that can be resolved. 

This way, the app can completely ignore the file structure (asking where the `userRepo` is for example) and has everything it needs to resolve a class.


src/script/config/dependenciesGraph.js
```diff
@@ -25,23 +25,50 @@ import BackendClient from '../service/BackendClient';
 import BackupService from '../backup/BackupService';
 import GiphyService from '../extension/GiphyService';
 import GiphyRepository from '../extension/GiphyRepository';
+import MediaRepository from '../media/MediaRepository';
```
atomrc, line 28: I think I'll sort those by name

PR #5828, opened 2019-02-15
test: Add tests for the SelfService class

atomrc:


bennycode:
@atomrc It looks like coveralls is just slow! 😋

PR #5831, opened 2019-02-15
refactor: Use @wireapp/protocol-messaging (WEBAPP-5904)

ffflorian:


src/script/conversation/ConversationRepository.js
```diff
@@ -2083,15 +2122,11 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
-    const genericMessage = new z.proto.GenericMessage(z.util.createRandomUuid());
-    const protoLocation = new z.proto.Location(
-      longitude,
-      latitude,
-      name,
-      zoom,
-      this.expectReadReceipt(conversationEntity)
-    );
-    genericMessage.set(z.cryptography.GENERIC_MESSAGE_TYPE.LOCATION, protoLocation);
+    const protoLocation = new Location(longitude, latitude, name, zoom, this.expectReadReceipt(conversationEntity));
```
atomrc: I believe the `Location` constructor should also take an object as a single parameter, isn't it?

ffflorian: Yes of course, good find!

PR #5845, opened 2019-02-18
refactor: Export ClipboardUtil as module

ffflorian:


src/script/util/ClipboardUtil.js
```diff
@@ -17,10 +17,7 @@
  *
  */
 
-window.z = window.z || {};
-window.z.util = z.util || {};
-
-z.util.ClipboardUtil = {
+const ClipboardUtil = {
```
atomrc: You could even export individual named functions here (instead of an object as the default export).
We would benefit from tree shaking with this method (BTW it seems the `pasteText` method is never used, if not, it can be removed)

src/script/view_model/panel/GuestsAndServicesViewModel.js
```diff
@@ -20,6 +20,7 @@
 import Logger from 'utils/Logger';
 
 import BasePanelViewModel from './BasePanelViewModel';
+import ClipboardUtil from '../../util/ClipboardUtil';
```
atomrc: There is an alias to access utils classes/functions.

```suggestion
import ClipboardUtil from 'utils/ClipboardUtil';
```

src/script/entity/message/ContentMessage.js
```diff
@@ -20,6 +20,7 @@
 import moment from 'moment';
 import ko from 'knockout';
 
+import ClipboardUtil from '../../util/ClipboardUtil';
```
atomrc:
```suggestion
import ClipboardUtil from 'utils/ClipboardUtil';
```

PR #5875, opened 2019-02-20
runfix: Correctly map file and audio data

ffflorian:


src/script/conversation/ConversationRepository.js
```diff
@@ -1798,16 +1798,14 @@ z.conversation.ConversationRepository = class ConversationRepository {
         return undefined;
       })
       .then(metadata => {
-        let assetOriginal = undefined;
+        const assetOriginal = new Asset.Original({mimeType: file.type, name: file.name, size: file.size});
```
atomrc, line 1801: Nice factorization :+1:

PR #5878, opened 2019-02-20
test: Add tests for the media button

atomrc:


src/script/components/asset/controls/mediaButton.js
```diff
@@ -32,37 +33,28 @@ class MediaButtonComponent {
...
-    this.on_cancel_button_clicked = function() {
-      if (typeof params.cancel === 'function') {
-        params.cancel();
-      }
-    };
+    const noop = () => {};
+
+    this.on_play_button_clicked = typeof params.play === 'function' ? () => params.play() : noop;
+    this.on_pause_button_clicked = typeof params.pause === 'function' ? () => params.pause() : noop;
+    this.on_cancel_button_clicked = typeof params.cancel === 'function' ? () => params.cancel() : noop;
```
lipis: what about making that cameCase while we at it?

atomrc: yep, sounds like a good idea ;)

PR #5886, opened 2019-02-21
fix: Use the proper tools for simple text styles (WEBAPP-5946)

AndyLnd:


src/script/util/LocalizerUtil.js
```diff
@@ -87,9 +87,9 @@ const LocalizerUtil = {
 
     const replaceDangerously = Object.assign(
       {
-        '/bold': '</b>',
+        '/bold': '</strong>',
         '/italic': '</i>',
-        bold: '<b>',
+        bold: '<strong>',
```
lipis, line 92: nice!

src/script/util/LocalizerUtil.js
```diff
@@ -87,9 +87,9 @@ const LocalizerUtil = {
 
     const replaceDangerously = Object.assign(
       {
-        '/bold': '</b>',
+        '/bold': '</strong>',
         '/italic': '</i>',
```
lipis, line 91: I think we can leave the `<i>` for now.. but if you guys want to read more about it: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#%3Ci%3E_vs._%3Cem%3E

PR #5898, opened 2019-02-22
refactor: Export PROPERTY_STATE directly

ffflorian:


src/script/components/groupVideoGrid.js
```diff
@@ -162,7 +164,7 @@ ko.components.register('group-video-grid', {
                   <micoff-icon></micoff-icon>
                 </div>
               <!-- /ko -->
-              <!-- ko if: streamInfo.videoSend() === z.calling.enum.PROPERTY_STATE.PAUSED -->
+              <!-- ko if: streamInfo.videoSend() === PROPERTY_STATE.PAUSED -->
```
atomrc, line 169: You probably want to add `PROPERTY_STATE` to the scope of the component. 

somewhere in the constructor `this.PROPERTY_STATE = PROPERTY_STATE`.

I don't think it works like this

PR #5913, opened 2019-02-25
feat: Enable logging via parameter

bennyn:
You can now supply a namespace which will activate all [logdown](https://github.com/caiogondim/logdown.js) instances (and [webapp-module-logger](https://github.com/wireapp/webapp-module-logger) instances) with that namespace.

Example:

- http://localhost:8888/auth/?env=staging&enableLogging=@wireapp


src/script/main/app.js
```diff
@@ -42,8 +42,9 @@ import {t} from 'utils/LocalizerUtil';
 /* eslint-disable no-unused-vars */
 import globals from './globals';
 import auth from './auth';
-import {getWebsiteUrl} from '../externalRoute';
 /* eslint-enable no-unused-vars */
+import {getWebsiteUrl} from '../externalRoute';
+import {enableLogging} from '../util/LoggerUtil';
```
atomrc, line 47: I think you can move those outside of the `eslint` comment (since the variables are used)

bennycode: They are below "no-unused" :)

PR #5920, opened 2019-02-26
feat: Enable logging on auth page via parameter

bennyn:
This PR uses a JS file (`/util/LoggerUtil.js`) inside a TSX file (`/auth/main.tsx`). Because of that we need to tell the `tsconfig.json` to use JS (`"allowJs": true`).

With `"allowJs": true` the TypeScript compiler checks our JavaScript code (which is a good thing!) and fails if it sees type mismatches. I resolved the type mismatches and made the PR compile.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "wire-webapp",
   "license": "GPL-3.0",
   "dependencies": {
+    "@types/js-cookie": "2.2.1",
     "@bower_components/normalize-css": "necolas/normalize.css#8.0.1",
     "@bower_components/webapp-module-bubble": "wireapp/webapp-module-bubble#1.1.1",
     "@bower_components/webapp-module-logger": "wireapp/webapp-module-logger#v1.1.2",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
   "name": "wire-webapp",
   "license": "GPL-3.0",
   "dependencies": {
-    "@types/js-cookie": "2.2.1",
     "@bower_components/normalize-css": "necolas/normalize.css#8.0.1",
     "@bower_components/webapp-module-bubble": "wireapp/webapp-module-bubble#1.1.1",
     "@bower_components/webapp-module-logger": "wireapp/webapp-module-logger#v1.1.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
   "name": "wire-webapp",
   "license": "GPL-3.0",
   "dependencies": {
+    "@types/js-cookie": "2.2.1",
```
ffflorian: Let's move it to the `devDependencies`.

src/script/main/SingleInstanceHandler.js
```diff
@@ -90,9 +91,8 @@ z.main.SingleInstanceHandler = (() => {
      * Does not check for the id of the running instance and thus cannot be
      * invoked once the registering of the current instance has been done.
      *
-     * @param {Function} listener - A listener to be removed.
      * @throws {Error} - When the current app has already been registered.
-     * @returns {void} - Returns nothing.
+     * @returns {boolean} - Returns nothing.
```
ffflorian: Should return something now :slightly_smiling_face:

src/script/util/TimeUtil.js
```diff
@@ -21,7 +21,7 @@
 
 import moment from 'moment';
 
-import {t} from 'utils/LocalizerUtil';
+import {t} from './LocalizerUtil';
```
ffflorian, line 24: Will we change this everywhere?

src/script/main/SingleInstanceHandler.js
bennyn: Why "now"? It has always returned something. Just the JSdoc annotation has been wrong.

src/script/util/TimeUtil.js
bennycode: Talked with @atomrc about it. I have a couple of reasons why we should use relative imports. I will sum them up but for now I would like to change only the ones which prevented `tsc` from compiling.

atomrc: We could also use absolute path, that would solve the `..` issue and is also compatible with TS. 
https://hackernoon.com/react-pro-tip-use-absolute-imports-for-better-readability-and-easy-refactoring-2ad5c7f2f957
Let's talk about it altogether

PR #5922, opened 2019-02-26
fix: Add a proper binding to handle infinite scroll (WEBAPP-5949)

atomrc:


src/script/view_model/bindings/MessageListBindings.js
```diff
@@ -76,6 +79,66 @@ ko.bindingHandlers.showAllTimestamps = {
...
+      disableScrollListener();
+    };
+
+    const onScroll = event => {
+      if (!scrollListenerEnabled) {
+        return;
+      }
+      const element = event.target;
+
+      // On some HDPI screen scrollTop returns a floating point number instead of an integer
```
AndyLnd:
```suggestion
      // On some HiDPI screens scrollTop returns a floating point number instead of an integer
```

src/script/view_model/bindings/MessageListBindings.js
```diff
@@ -76,6 +79,66 @@ ko.bindingHandlers.showAllTimestamps = {
...
+      } else if (hitBottom) {
+        onHitBottom();
+      }
+    };
+
+    const onMouseWheel = event => {
+      /**
+       * To avoid reacting to programmatically triggered scroll, we just want to enable the scroll listener when there is a mouse event before it.
+       * The strategy here is to enable the scroll listener when there is a mouse event and automatically disable the listener after 500ms of inactivity of the mouse.
+       */
```
AndyLnd: I think it might be problematic to only react to the `mousewheel` event. There are other user interactions that will cause scrolling, selecting chunks of text and dragging the scrollbar for example.

src/script/view_model/bindings/MessageListBindings.js
```diff
@@ -76,6 +79,66 @@ ko.bindingHandlers.showAllTimestamps = {
...
+      if (hitTop) {
+        onHitTop();
+      } else if (hitBottom) {
+        onHitBottom();
+      }
+    };
+
+    const onMouseWheel = event => {
+      /**
+       * To avoid reacting to programmatically triggered scroll, we just want to enable the scroll listener when there is a mouse event before it.
```
AndyLnd:
```suggestion
       * To avoid reacting to programmatically triggered scrolls, we just want to enable the scroll listener when there is a mouse event before it.
```

PR #5932, opened 2019-02-27
runfix: Enforce latest React types

Yserz:


package.json
```diff
@@ -230,5 +230,8 @@
     "@bower_components/wire-theme": {
       "less": "app/style/{wire-theme,variables,input}.less"
     }
+  },
+  "resolutions": {
```
bennycode, line 234: Why do we need that?

PR #5933, opened 2019-02-27
chore: update ui kit dependency

AndyLnd:


package.json
```diff
@@ -53,7 +53,6 @@
     "wire-web-config-default-staging": "https://github.com/wireapp/wire-web-config-default#v0.13.2"
   },
   "devDependencies": {
-    "@types/js-cookie": "2.2.1",
```
ffflorian, line 56: Why did you remove them?

AndyLnd: It was in there twice (line 66)

PR #5936, opened 2019-02-27
runfix: Handle rich profile fetching errors (WEBAPP-5964)

atomrc:


src/script/components/panel/enrichedFields.js
```diff
@@ -37,11 +37,16 @@ ko.components.register('enriched-fields', {
...
+    this.richProfileRepository
+      .getUserRichProfile(ko.unwrap(userId))
+      .then(richProfile => {
+        if (richProfile.fields) {
+          this.fields(richProfile.fields);
+          onFieldsLoaded(this.fields());
+        }
+      })
+      .catch(() => {
+        this.fields([]);
```
bennycode: `this.fields` is already initialized with an empty array so is it necessary to set it again?

ffflorian: If the error occurs in `onFieldsLoaded()` then it would be necessary I think.

atomrc: Yeah good point @ffflorian we don't want to change the fields back to the empty array if the parent function fails. 
I'll just put a `noop` function in the catch then.

PR #5962, opened 2019-03-01
refactor: Centralize handling of conversations' last read timestamp (WEBAPP-5967)

atomrc:
As of the Read Receipt feature (https://github.com/wireapp/wire-webapp/pull/5411) we have a mechanism to reliably detect when a message has been read.

This is a perfect opportunity to rethink the way the `last_read_timestamp` of the conversation is handled.

This PR brings the power of the read detection to update the timestamp. Which means we will now update this timestamp live while the user scrolls the message list.

Another benefit of this is that we have a single place where we send the message that synchronizes all the clients.

### DIFF
--- a/src/script/conversation/ConversationRepository.js
+++ b/src/script/conversation/ConversationRepository.js
@@ -1112,19 +1112,34 @@ z.conversation.ConversationRepository = class ConversationRepository {
   }
 
   /**
-   * Mark conversation as read.
+   * Sends a message to backend that the conversation has been fully read.
+   * The message will allow all the self clients to synchronize conversation read state.
+   *
    * @param {Conversation} conversationEntity - Conversation to be marked as read
    * @returns {undefined} No return value
    */
   markAsRead(conversationEntity) {
-    if (conversationEntity) {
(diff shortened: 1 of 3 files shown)


src/script/conversation/ConversationRepository.js
```diff
@@ -1112,19 +1112,34 @@ z.conversation.ConversationRepository = class ConversationRepository {
...
+    const eventInfoEntity = new z.conversation.EventInfoEntity(genericMessage, this.self_conversation().id);
+    this.sendGenericMessageToConversation(eventInfoEntity)
+      .then(() => {
+        this.logger.info(`Marked conversation '${conversationId}' as read on '${new Date(timestamp).toISOString()}'`);
+      })
+      .catch(error => {
+        const errorMessage = 'Failed to update last read timestamp';
+        this.logger.error(`${errorMessage}: ${error.message}`, error);
+        Raygun.send(new Error(errorMessage), {label: error.label, message: error.message});
+      });
```
AndyLnd, line 1143: Looks good, but this seems to be missing: `amplify.publish(z.event.WebApp.NOTIFICATION.REMOVE_READ);`

PR #5963, opened 2019-03-01
runfix: Fix detection of the first unread message

atomrc:


src/page/template/content/conversation.htm
```diff
@@ -1,12 +1,11 @@
 <div id="conversation" class="conversation" data-bind="drop_file: $root.inputBar.onDropFiles, css: { loading: !$root.messageList.conversationLoaded() }">
   <!-- ko if: $root.conversationRepository.active_conversation() -->
-    <div class="conversation-loading" >
-      <div class="icon-spinner spin text-theme"></div>
-    </div>
-    
     #include('content/conversation/title-bar.htm')
     #include('content/conversation/message-list.htm')
     #include('content/conversation/input-bar.htm')
+    <div class="conversation-loading" >
```
atomrc, line 6: This will fix a bug where we would trigger the callbacks supposed to be triggered when a message is viewed by the user (updating conversation last_read_timestamp, sending read receipts...) while the conversation was loading (and thus the messages not visible to the user).

src/page/template/content/conversation/message-list.htm
```diff
@@ -7,7 +7,7 @@
                     attr: {'data-uie-uid': message.id, 'data-uie-value': message.super_type, 'data-uie-expired-status': message.ephemeral_expires, 'data-uie-send-status': message.status}"
          data-uie-name="item-message">
 
-      <div class="message-header message-timestamp" data-bind="css: get_timestamp_class($data)">
+      <div class="message-header message-timestamp" data-bind="css: get_timestamp_class(message)">
```
atomrc: This is probably broken since the upgrade to knockout 3.5

PR #5992, opened 2019-03-05
refactor: Use openGraphAsync if possible

ffflorian:
Introducing an async version of `openGraph()` (https://github.com/wireapp/wire-desktop/pull/2269)


src/script/links/LinkPreviewRepository.js
```diff
@@ -143,6 +143,20 @@ z.links.LinkPreviewRepository = class LinkPreviewRepository {
    * @returns {Promise} Resolves with the retrieved open graph data
    */
   _fetchOpenGraphData(link) {
+    if (typeof window.openGraphAsync === 'function') {
+      return window
+        .openGraphAsync(link)
+        .then(data => {
+          if (data) {
```
atomrc: The logic seems to be exactly the same for both handler. Can we extract it into a new function

PR #5995, opened 2019-03-05
fix: Don't autocomplete user search field (WEBAPP-5976)

ffflorian:


src/script/components/userInput.js
```diff
@@ -82,20 +82,21 @@ z.components.UserInput = class UserInput {
...
-        <div class="search-inner"">
-          <div class="search-icon icon-search"></div>
-          <!-- ko ifnot: noSelectedUsers-->
-            <!-- ko foreach: selectedUsers -->
-              <span data-bind="text: first_name()" data-uie-name="item-selected"></span>
-            <!-- /ko -->
-          <!-- /ko -->
-          <input type="text" style="display:none" /> <!-- prevent chrome from autocomplete -->
-          <input autocomplete="off" maxlength="128" required spellcheck="false" class="search-input" type="text" data-bind="textInput: input, hasFocus: hasFocus, attr: {placeholder: placeholder}, css: {'search-input-show-placeholder': placeholder}, event: {keydown: onKeyDown}, enter: onEnter" data-uie-name="enter-users">
+    <form autocomplete="off">
```
atomrc: Maybe we can try replacing the wrapping `div` with the `form` instead of adding an empty wrapping `form`

PR #6053, opened 2019-03-11
refactor: Use app default logger everywhere 

atomrc:


src/script/calling/entities/FlowEntity.js
```diff
@@ -60,8 +61,8 @@ z.calling.entities.FlowEntity = class FlowEntity {
     this.conversationId = this.callEntity.id;
     this.messageLog = this.participantEntity.messageLog;
 
-    const loggerName = 'z.calling.entities.FlowEntity';
-    this.callLogger = new z.telemetry.calling.CallLogger(loggerName, this.id, z.config.LOGGER.OPTIONS, this.messageLog);
+    const loggerName = 'FlowEntity';
```
Yserz: If it's not used anywhere else let's make this one inline

src/script/calling/entities/ParticipantEntity.js
```diff
@@ -51,7 +52,7 @@ z.calling.entities.ParticipantEntity = class ParticipantEntity {
     this.sessionId = undefined;
 
     const loggerName = 'z.calling.entities.ParticipantEntity';
-    this.callLogger = new z.telemetry.calling.CallLogger(loggerName, this.id, z.config.LOGGER.OPTIONS, this.messageLog);
+    this.callLogger = new CallLogger(loggerName, this.id, this.messageLog);
```
Yserz: If it's not used anywhere else let's make this one inline

src/script/entity/message/Message.js
```diff
@@ -18,6 +18,7 @@
  */
 
 import ko from 'knockout';
+import moment from 'moment';
```
Yserz, line 21: why?

atomrc: It was published on the global scope before (the old logger needed it). 
Now every class needs to import it manually

test/api/TestFactory.js
```diff
@@ -30,23 +32,10 @@ window.testConfig = {
...
-
-  const initialLoggerOptions = z.config.LOGGER.OPTIONS;
-  Object.keys(initialLoggerOptions.domains).forEach(domain => {
-    initialLoggerOptions.domains[domain] = logger_level;
-  });
-  initialLoggerOptions.level = logger_level;
-
-  return this;
-};
+window.TestFactory = function() {};
```
Yserz, line 38: Can we remove that?

atomrc: unfortunalty no, we still do `new TestFactory` everywhere. 
But in the long term, yes we will remove that class entirely :)

PR #6065, opened 2019-03-12
runfix: Trigger a last read timestamp update when scroll reaches bottom of conversation (WEBAPP-5997)

atomrc:


src/script/view_model/bindings/MessageListBindings.js
```diff
@@ -83,9 +83,7 @@ ko.bindingHandlers.infinite_scroll = {
     const {onHitTop, onHitBottom, onInit} = params();
     onInit(scrollingElement);
 
-    const onScroll = event => {
-      const element = event.target;
-
+    const onScroll = ({target: element}) => {
```
bennyn: Is it an `event` or an `element` that comes in? Because before it seemed to be `event.target`.

src/script/view_model/content/MessageListViewModel.js
```diff
@@ -582,6 +575,18 @@ class MessageListViewModel {
     };
   }
 
+  updateConversationLastRead(conversationEntity, messageEntity) {
+    const conversationLastRead = conversationEntity.last_read_timestamp();
+    const lastKnowTimestamp = conversationEntity.get_last_known_timestamp(
```
bennycode: typo: `lastKnownTimestamp`

atomrc: Good catch

PR #6066, opened 2019-03-12
chore: Log OpenGraph fetching errors

ffflorian:
Connected to https://github.com/wireapp/wire-desktop/pull/2328.


src/script/links/LinkPreviewRepository.js
```diff
@@ -148,19 +148,23 @@ class LinkPreviewRepository {
       return window
         .openGraphAsync(link)
         .then(mergeOpenGraphData)
-        .catch(() => Promise.resolve());
+        .catch(error => this.logger.info(`Error while fetching OpenGraph data: ${error.message}`));
```
atomrc: let's use `warn` instead of `info`

PR #6080, opened 2019-03-13
refactor: Clean the lifecycle API and add tests to it

atomrc:
This class will be completely removed in the near future. 
**but** this PR will serve as an example of how tests can allow/force us to write simpler code. 

The main takeaways from that PR are:
- Avoid side effects from utility modules (in this case, calling `amplify.publish` and reading the version from the module)
- Avoid injecting too much power into utility modules (in this case, `UserRepository` was injected which gives way to much power to the actual module)
- Avoid too complex abstraction for simple tasks (when a single function can do the trick, use a single function)
- jasmine.clock is a really powerful tool


src/script/lifecycle/newVersionHandler.js
```diff
@@ -0,0 +1,59 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
AndyLnd:
```suggestion
 * Copyright (C) 2019 Wire Swiss GmbH
```

test/unit_tests/lifecycle/newVersionHandlerSpec.js
```diff
@@ -0,0 +1,66 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
AndyLnd:
```suggestion
 * Copyright (C) 2019 Wire Swiss GmbH
```

PR #6083, opened 2019-03-13
refactor: Read env config from the most recent config spec

atomrc:
This PR does a couple of improvements:
- removes the old `auth` entry point (which is now handled by React);
- makes `app.js` and `login.js` proper standalone entry points;
- makes more use of the config coming from the server (instead of guessing the config from the url).


src/script/auth/Environment.ts
```diff
@@ -27,9 +27,11 @@ export enum ENVIRONMENT {
   PRODUCTION = 'PRODUCTION',
 }
 
-export const APP_ENVIRONMENT = getEnvironmentFromQuery();
-checkEnvironment();
-export function getEnvironmentFromQuery() {
+export function isLocalhost() {
+  return window.location.hostname.includes('localhost') || window.location.hostname.startsWith('192.168.');
```
ffflorian, line 31: localhost could also be `127.0.0.1` or another IP..

atomrc: I don't think we need to change this behavior right now. This will be completely removed in the near future

PR #6101, opened 2019-03-14
refactor: Use resolver where possible

atomrc:


src/script/util/DebugUtil.js
```diff
@@ -297,13 +297,6 @@ export default class DebugUtil {
     this.callingRepository.printLog();
   }
 
-  logConnectionStatus() {
-    this.logger.log('Online Status');
-    this.logger.log(`-- Browser online: ${window.navigator.onLine}`);
-    this.logger.log(`-- IndexedDB open: ${this.storageRepository.storageService.db.isOpen()}`);
-    this.logger.log(`-- WebSocket ready state: ${window.wire.app.service.web_socket.socket.readyState}`);
```
atomrc, line 304: I checked, this is not used by QA (and this service `web_socket` doesn't not exist anymore)

PR #6103, opened 2019-03-14
refactor: Use CSS variables for accent colors

AndyLnd:


src/style/common/colors.less
```diff
@@ -17,65 +17,15 @@
...
-.text-blue {
-  .text(@w-blue);
-}
-.text-green {
-  .text(@w-green);
-}
-.text-yellow {
-  .text(@w-yellow);
-}
 .text-red {
```
atomrc, line 25: Why do we still have this one?

AndyLnd: it is still used for some warnings

atomrc: I think you removed it already (in the `device-remove.less`

PR #6115, opened 2019-03-15
refactor: First step to resolve cyclic dependencies

atomrc:
This is a first step to factorizing the way we send messages in Wire. 

Right now, we have 2 methods that, basically, have the same logic `ConversationRepository._sendEncryptedMessage` and `BroadcastRepository._sendEncryptedMessage`. 

Harmonizing this will be a long journey of identifying patterns and resolving cyclic dependencies but we will get there.

### DIFF
--- a/src/script/broadcast/BroadcastRepository.js
+++ b/src/script/broadcast/BroadcastRepository.js
@@ -136,7 +136,7 @@ z.broadcast.BroadcastRepository = class BroadcastRepository {
         return accumulator + userEntity.devices().length;
       }
       return accumulator + z.client.ClientRepository.CONFIG.AVERAGE_NUMBER_OF_CLIENTS;
-    });
+    }, 0);
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/script/broadcast/BroadcastRepository.js
```diff
@@ -32,47 +32,62 @@ z.broadcast.BroadcastRepository = class BroadcastRepository {
    * @param {ClientRepository} clientRepository - Repository for client interactions
    * @param {ConversationRepository} conversationRepository - Repository for conversation interactions
    * @param {CryptographyRepository} cryptographyRepository - Repository for all cryptography interactions
-   * @param {UserRepository} userRepository - Repository for all user interactions
```
atomrc, line 35: This is the main goal of this PR: removing the dependency to the UserRepository from the BroadcastRepository.

src/script/message/MessageSender.js
```diff
@@ -0,0 +1,33 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import PromiseQueue from 'utils/PromiseQueue';
+
+export class MessageSender {
```
atomrc, line 22: I have great plans for this `MessageSender`. In the future it will be the only class knowing how to send a message on Wire. 
Its reponsabilities will include:
- encrypting the messages 
- handling 412 responses from the backend and warn the app that something needs to be done. 

Right now it's mostly an empty shell around the `PromiseQueue` but it will evolve with some followup PR

atomrc:
> LGTM but as I remember our inital plan was to integrate the `@wireapp/core` module someday. Isn't this unnecessary work then?

It will surely make the work of integrating `@wireapp/core` easier. 
Quick question regarding this: Does the core handles 412 responses and does it have a way to warn that a 412 response was received?

PR #6155, opened 2019-03-20
feat: handle markdown links

AndyLnd:


src/script/util/util.js
```diff
@@ -302,15 +303,25 @@ markdownit.renderer.rules.link_open = (tokens, idx, options, env, self) => {
...
-    const email = href.replace(/^mailto:/, '');
+    const email = SanitizationUtil.escapeString(href.replace(/^mailto:/, ''));
     link.attrPush(['onclick', `z.util.SanitizationUtil.safeMailtoOpen(event, '${email}')`]);
   } else {
     link.attrPush(['target', '_blank']);
     link.attrPush(['rel', 'nofollow noopener noreferrer']);
   }
   if (link.markup !== 'linkify') {
-    link.attrPush(['class', 'dangerLink']);
+    link.attrPush(['class', 'markdown-link']);
```
atomrc: Let's try to use dataset (if allowed by markdownit). This way we know it not for styling and we won't be tempted to remove it in the future.

PR #6176, opened 2019-03-22
refactor: Optimize loading view performances

atomrc:
In the current state of the app, we instantiate **all** the views of the app at bootstrap time. 
This part of the app's bootstrap is pretty expensive (because it does all the bindings between the view models and the real DOM).

**But** when the app loads the data it needs (conversations, users, crypto...) we just need a single view model (namely the `LoadingViewModel`).

This PR will start loading the data of the app **before** loading all the other views.

### DIFF
--- a/src/script/main/app.js
+++ b/src/script/main/app.js
@@ -22,6 +22,7 @@ import Logger from 'utils/Logger';
 import platform from 'platform';
 import {Config} from '../auth/config';
 import {startNewVersionPolling} from '../lifecycle/newVersionHandler';
+import {LoadingViewModel} from '../view_model/LoadingViewModel';
 import PreferenceNotificationRepository from '../notification/PreferenceNotificationRepository';
 import * as UserPermission from '../user/UserPermission';
 import UserRepository from '../user/UserRepository';
@@ -86,7 +87,6 @@ class App {
 
     this.service = this._setupServices();
     this.repository = this._setupRepositories();
-    this.view = this._setupViewModels();
     this.util = this._setup_utils();
 
     this._publishGlobals();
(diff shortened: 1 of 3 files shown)


src/script/main/app.js
```diff
@@ -290,29 +282,30 @@ class App {
    * @returns {undefined} No return value
    */
   initApp(isReload = this._isReload()) {
+    const loadingView = new LoadingViewModel(this.repository.user.self);
```
atomrc: Nice benefit, since the loadingView is never used anywhere else, we just declare it here and then the garbage collector will take care of removing it from memory. 

(before, it stayed in memory forever)

src/script/view_model/LoadingViewModel.js
```diff
@@ -18,23 +18,23 @@
...
 
-    ko.applyBindings(this, document.getElementById(this.elementId));
+    const elementId = 'loading-screen';
+    this.element = document.getElementById(elementId);
+    ko.applyBindings(this, this.element);
   }
 
   removeFromView() {
-    $(`#${this.elementId}`).remove();
+    ko.cleanNode(this.element);
```
atomrc, line 35: This is also needed for the gc to collect that class

src/script/view_model/LoadingViewModel.js
```diff
@@ -18,23 +18,23 @@
...
 
-window.z = window.z || {};
-window.z.viewModel = z.viewModel || {};
-
-z.viewModel.LoadingViewModel = class LoadingViewModel {
-  constructor(mainViewModel, repositories) {
-    this.elementId = 'loading-screen';
-    this.userRepository = repositories.user;
+export class LoadingViewModel {
+  constructor(userObservable) {
```
atomrc: This class just needs the user (to display the user name in the message).
This can be optimized further by only injecting the user name but at the time it is instantiated, the user is undefined.

Will see if I can find a better option

PR #6208, opened 2019-03-27
runfix: Show conversation when sending a connection request or unblock a user (WEBAPP-6034, WEBAPP-6035)

atomrc:


src/script/components/panel/userActions.js
```diff
@@ -136,7 +136,7 @@ ko.components.register('user-actions', {
           return isNotMe && isNotConnectedUser && canConnect;
         },
         item: {
-          click: () => actionsViewModel.sendConnectionRequest(user).then(() => onAction(Actions.SEND_REQUEST)),
+          click: () => actionsViewModel.sendConnectionRequest(user, true).then(() => onAction(Actions.SEND_REQUEST)),
```
atomrc: Just for the record: I might create a follow up PR that removes the fact that the `ConnectionRepository` knows how to control the UI!!!

PR #6237, opened 2019-03-29
refactor: Limit scope of variables in root components of the app

atomrc:
This will allow those variables to be garbage collected and help reading/refactoring the code in the future

### DIFF
--- a/src/script/main/app.js
+++ b/src/script/main/app.js
@@ -18,6 +18,7 @@
  */
 
 import Logger from 'utils/Logger';
+import ko from 'knockout';
 
 import platform from 'platform';
 import {Config} from '../auth/config';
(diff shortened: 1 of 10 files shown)


src/script/main/app.js
```diff
@@ -598,7 +602,8 @@ class App {
    * @returns {undefined} No return value
    */
   _showInterface() {
-    const mainView = new z.viewModel.MainViewModel(this.repository);
+    const mainView = new MainViewModel(this.repository);
+    ko.applyBindings(mainView, this.domContainer);
```
atomrc, line 606: Now, only the App class knows where to bootstrap the application.
The MainViewModel has no idea about the DOM structure anymore.

src/script/view_model/MainViewModel.js
```diff
@@ -60,15 +57,8 @@ z.viewModel.MainViewModel = class MainViewModel {
   }
 
   constructor(repositories) {
-    this.closePanel = this.closePanel.bind(this);
```
atomrc, line 63: All those functions do not need to be manually bound (as they are always consumed directly `view.method()` and never given `const method = view.method`).
(expect for `togglePanel` and `closePanelOnClick`, for them I used the `method = () => {}` technique (in order to cleanup the constructor)

src/script/main/app.js
```diff
@@ -845,13 +850,13 @@ class App {
 
 $(() => {
   enableLogging(Config.FEATURE.ENABLE_DEBUG);
-  if ($('#wire-main-app').length !== 0) {
+  if (document.getElementById('wire-main-app') !== null) {
```
atomrc, line 853: So long jQuery

PR #6241, opened 2019-03-29
refactor: Clear some strong DOM coupling in App class

atomrc:


src/page/template/wire-main.htm
```diff
@@ -63,7 +63,7 @@
         <!-- ko if: state() === z.viewModel.ContentViewModel.STATE.WATERMARK -->
           <div class="watermark">
             <watermark-icon class="absolute-center" data-uie-name="no-conversation"></watermark-icon>
-          </div>        
+          </div>
```
atomrc, line 66: Without any doubt the most important change of this PR!!!

src/script/main/app.js
```diff
@@ -852,13 +852,14 @@ class App {
 
 $(() => {
   enableLogging(Config.FEATURE.ENABLE_DEBUG);
-  if (document.getElementById('wire-main-app')) {
+  const appContainer = document.getElementById('wire-main');
+  if (appContainer) {
```
atomrc, line 855: The test is slightly different (not on the same DOM element), but the 2 elements have the exact same semantic so the behavior is the same.

PR #6242, opened 2019-03-29
 feat: Update group creation design

AndyLnd:


src/script/util/KeyboardUtil.js
```diff
@@ -142,3 +142,19 @@ z.util.KeyboardUtil = (() => {
...
+  }
+});
+
+export const listenToEscKey = (identifier, handler) => {
+  escKeyListeners[identifier] = handler;
+};
+
+export const unlistenToEscKey = identifier => {
+  delete escKeyListeners[identifier];
+};
```
AndyLnd, line 161: I am going to tackle the rest in a separate PR

src/script/view_model/ContentViewModel.js
```diff
@@ -74,7 +76,7 @@ z.viewModel.ContentViewModel = class ContentViewModel {
       repositories,
       z.message.MessageHasher
     );
-    this.groupCreation = new z.viewModel.content.GroupCreationViewModel(mainViewModel, this, repositories);
+    this.groupCreation = new GroupCreationViewModel(mainViewModel, this, repositories);
```
atomrc: Can you check if we can reduce the amount of stuff we import in there?

src/script/util/KeyboardUtil.js
atomrc: Can you add tests for those new functions.
Should be quite simple

PR #6243, opened 2019-03-29
refactor: Inline small-single-used private methods of App

atomrc:


src/script/main/app.js
```diff
@@ -90,15 +90,16 @@ class App {
 
     this.service = this._setupServices();
     this.repository = this._setupRepositories();
-    this.util = this._setup_utils();
+    if (z.config.FEATURE.ENABLE_DEBUG) {
+      this.util = {debug: new DebugUtil(this.repository)};
```
atomrc, line 94: Would be even nicer to have `this.debug = new DebugUtil` but it requires QA to change all occurences of `wire.app.util.debug` to `wire.app.debug` on their end. 

Will discuss this with @mythsunwind on Monday

PR #6292, opened 2019-04-04
refactor: Migrate call components to modules

atomrc:


src/script/calling/CallingRepository.js
```diff
@@ -960,38 +961,39 @@ export class CallingRepository {
...
-   * @param {string} conversationId - ID of conversation to join call in
+   * @param {Conversation} conversationEntity - conversation to join call in
    * @param {z.media.MediaType} mediaType - Media type for this call
    * @param {CALL_STATE} callState - Current state of call
    * @returns {Promise} Resolves when conversation supports calling
    */
-  _checkCallingSupport(conversationId, mediaType, callState) {
-    return this.conversationRepository.get_conversation_by_id(conversationId).then(conversationEntity => {
+  _checkCallingSupport(conversationEntity, mediaType, callState) {
+    return new Promise((resolve, reject) => {
```
atomrc, line 978: I reduced the dependency to the `conversationRepo` here by injecting the conversationEntity directly into the method.
In the near future, the conversationRepo should be completely gone from the CallingRepo

src/script/calling/CallMessageBuilder.js
```diff
@@ -19,107 +19,102 @@
...
-    buildSetup: _buildSetup,
-    buildUpdate: _buildUpdate,
-    createPayload: _createPayload,
-    createPropSync: _createPropSync,
-  };
-})();
+  return additionalPayload ? Object.assign(payload, additionalPayload) : payload;
+};
+
+export const CallMessageBuilder = {
```
ffflorian, line 107: Why not also refactor this by removing the unnecessary renaming?

atomrc: Fair point :+1:

PR #6297, opened 2019-04-04
refactor: warning modals

AndyLnd:


src/script/view_model/ModalsViewModel.js
```diff
@@ -60,169 +76,119 @@ z.viewModel.ModalsViewModel = class ModalsViewModel {
...
-    const actionElement = $(type).find('.modal__button--confirm');
-    const messageElement = $(type).find('.modal-text');
-    const titleElement = $(type).find('.modal-title');
+  showModal = (type, options = {}) => {
+    if (!Object.values(ModalsViewModel.TYPE).includes(type)) {
+      return this.logger.warn(`Modal of type '${type}' is not supported`);
+    }
+
+    const {text = {}, preventClose = false, close = noop, action = noop, secondary = noop, data} = options;
+    this.titleText(text.title);
```
atomrc: I really like what I am seeing with this refactoring. Although, I think you have over-engineered that part. 
Having all those observables seems not efficient, performance-wise (each single modification will trigger some DOM updates), and pretty hard to maintain and evolve.

Since those observables can only change all at once (and never independently) I would go with a single obs that contains one config object with everything inside:
Benefits are:
- just 1 DOM update per modal appearing/disappearing
- if the consumer passes on the config with the right format you have a minimal amount of work to do in this class

PR #6299, opened 2019-04-05
feat: Enable screensharing in Chrome v72+ (WEBAPP-6083)

bennyn:

### DIFF
--- a/src/script/media/MediaConstraintsHandler.js
+++ b/src/script/media/MediaConstraintsHandler.js
@@ -39,7 +39,7 @@ export default class MediaConstraintsHandler {
             },
           },
         },
-        FIREFOX: {
+        USER_MEDIA: {
           frameRate: 30,
           height: {exact: 720},
           mediaSource: 'screen',
@@ -135,23 +135,23 @@ export default class MediaConstraintsHandler {
       return Promise.resolve(streamConstraints);
     }
 
-    if (z.util.Environment.browser.firefox) {
-      this.logger.info('Enabling screen sharing from Firefox');
-
-      const streamConstraints = {
-        audio: false,
-        video: MediaConstraintsHandler.CONFIG.SCREEN_CONSTRAINTS.FIREFOX,
-      };
-
-      return Promise.resolve(streamConstraints);
-    }
-
-    if (!!navigator.mediaDevices.getDisplayMedia) {
+    if (navigator.mediaDevices.getDisplayMedia) {
       this.logger.info('Enabling screen sharing from Chrome');
 
       const streamConstraints = {
         audio: false,
         video: MediaConstraintsHandler.CONFIG.SCREEN_CONSTRAINTS.DISPLAY_MEDIA,
+      };
+
+      return Promise.resolve(streamConstraints);
+    }
+
+    if (z.util.Environment.browser.firefox) {
+      this.logger.info('Enabling screen sharing from Firefox');
+
+      const streamConstraints = {
+        audio: false,
+        video: MediaConstraintsHandler.CONFIG.SCREEN_CONSTRAINTS.USER_MEDIA,
       };
 
       return Promise.resolve(streamConstraints);
--- a/src/script/media/MediaConstraintsHandler.js
+++ b/src/script/media/MediaConstraintsHandler.js
@@ -138,6 +138,17 @@ export default class MediaConstraintsHandler {
       return Promise.resolve(streamConstraints);
     }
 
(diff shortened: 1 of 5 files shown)


src/script/view_model/VideoCallingViewModel.js
```diff
@@ -195,15 +195,18 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
     this.TimeUtil = TimeUtil;
   }
 
-  chooseSharedScreen(conversationId) {
+  async chooseSharedScreen(conversationId) {
```
ffflorian: Not needed and not supported
```suggestion
  chooseSharedScreen(conversationId) {
```

src/script/util/Environment.js
```diff
@@ -117,7 +117,10 @@ z.util.Environment = (() => {
     const requestPermissionNotSupported = window.Notification.requestPermission === undefined;
     return requestPermissionNotSupported ? false : document.visibilityState !== undefined;
   };
-  const _supportsScreenSharing = () => (window.desktopCapturer ? true : _isFirefox());
+  const _supportsScreenSharing = () => {
+    const hasScreenCaptureAPI = window.desktopCapturer || !!navigator.mediaDevices.getDisplayMedia;
+    return hasScreenCaptureAPI ? true : _isFirefox();
```
ffflorian:
```suggestion
    return hasScreenCaptureAPI || _isFirefox();
```

src/script/media/MediaConstraintsHandler.js
```diff
@@ -24,13 +24,21 @@ export default class MediaConstraintsHandler {
...
         },
+        DISPLAY_MEDIA: {
+          video: {
+            height: {
+              ideal: 1080,
+              max: 1080,
+            },
+          },
+        },
         FIREFOX: {
```
atomrc: Let's maybe rename this to `GET_USER_MEDIA` to be consistent with the other two, WDYT?

src/script/media/MediaConstraintsHandler.js
```diff
@@ -138,6 +146,17 @@ export default class MediaConstraintsHandler {
       return Promise.resolve(streamConstraints);
     }
 
+    if (!!navigator.mediaDevices.getDisplayMedia) {
```
atomrc: Since Firefox also implements this method, we could move that condition up (before the firefox condition) so that we go into the legacy `getUserMedia` only if the preferred methods are not available

src/script/media/MediaDevicesHandler.js
```diff
@@ -251,7 +251,8 @@ export default class MediaDevicesHandler {
   }
 
   /**
-   * Update list of available Screens.
+   * TODO: Make it usable for chrome because currently it's a wrapper-only thing.
```
atomrc: This todo can be removed. Chrome should handle the screen selection popup itself, we don't need to hook into it

src/script/media/MediaStreamHandler.js
```diff
@@ -565,8 +565,13 @@ export default class MediaStreamHandler {
       this._schedulePermissionHint(mediaType);
     }
 
-    return navigator.mediaDevices
-      .getUserMedia(mediaStreamConstraints)
+    const isScreenSharingFromChrome = mediaType === 'screen' && !!navigator.mediaDevices.getDisplayMedia;
```
atomrc:
```suggestion
    const supportsGetDisplayMedia = mediaType === 'screen' && !!navigator.mediaDevices.getDisplayMedia;
```

src/script/view_model/VideoCallingViewModel.js
```diff
@@ -197,13 +197,16 @@ z.viewModel.VideoCallingViewModel = class VideoCallingViewModel {
...
+        this.selfStreamState.screenSend() ||
+        z.util.Environment.browser.firefox ||
+        navigator.mediaDevices.getDisplayMedia;
       if (skipScreenSelection) {
         amplify.publish(z.event.WebApp.CALL.MEDIA.TOGGLE, conversationId, z.media.MediaType.SCREEN);
         return;
       }
 
-      if (z.util.Environment.desktop) {
+      if (window.desktopCapturer) {
```
atomrc, line 209: :+1:

src/script/media/MediaConstraintsHandler.js
bennycode: Yeah, that would be more consistent then.

PR #6309, opened 2019-04-08
feat: Show email if supplied by the backend

AndyLnd:


src/script/components/panel/enrichedFields.js
```diff
@@ -33,18 +34,28 @@ ko.components.register('enriched-fields', {
...
-    this.richProfileRepository
-      .getUserRichProfile(ko.unwrap(userId))
-      .then(richProfile => {
-        if (richProfile.fields) {
-          this.fields(richProfile.fields);
-          onFieldsLoaded(this.fields());
-        }
-      })
-      .catch(() => {});
+    ko.computed(() => {
```
atomrc: Please dispose of this subscription when the component is being destroyed

atomrc: Please dispose of this subscription when the component is being destroyed

PR #6372, opened 2019-04-16
 feat: Do not notify when user is away (WEBAPP-6090)

AndyLnd:


src/script/notification/NotificationRepository.js
```diff
@@ -470,22 +478,20 @@ z.notification.NotificationRepository = class NotificationRepository {
    * @returns {Promise} Resolves with the notification message body
```
ffflorian: Does it still return a Promise?

src/script/notification/NotificationRepository.js
```diff
@@ -470,22 +478,20 @@ z.notification.NotificationRepository = class NotificationRepository {
...
+        return this._createBodyContent(messageEntity);
+      case z.message.SuperType.MEMBER:
+        return this._createBodyMemberUpdate(messageEntity, connectionEntity, conversationEntity);
+      case z.message.SuperType.PING:
+        return this._createBodyPing();
+      case z.message.SuperType.REACTION:
+        return this._createBodyReaction(messageEntity);
+      case z.message.SuperType.SYSTEM:
+        return this._createBodySystem(messageEntity);
+    }
```
ffflorian, line 494: No default?

PR #6389, opened 2019-04-17
fix: Adjust call messages time with the server time (WEBAPP-6129)

atomrc:
In order to consume the serverTimeRepo from the message builder, I needed to make the former a proper singleton.

### DIFF
--- a/src/script/calling/CallMessageBuilder.js
+++ b/src/script/calling/CallMessageBuilder.js
@@ -17,6 +17,7 @@
  *
  */
 
+import {serverTimeRepository} from '../time/ServerTimeRepository';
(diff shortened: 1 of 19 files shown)


bennycode:
Can `import {serverTimeRepository} from 'src/script/time/ServerTimeRepository'` be turned into `import {ServerTimeRepository} from 'src/script/time/ServerTimeRepository'`?

atomrc:
@bennyn yeah, I also hesitated. But since the file doesn't export a class anymore I was actually thinking about renaming the file and the export to something different.
Like `serverTimeHandler` (not capitalized). WDYT?

bennycode:
> I was actually thinking about renaming the file and the export to something different

Ok!

PR #6398, opened 2019-04-18
runfix: Remove duplicate error message (WEBAPP-6108)

atomrc:
This is the proper fix that should have been done here https://github.com/wireapp/wire-webapp/pull/6387


src/script/calling/CallingRepository.js
```diff
@@ -945,7 +945,6 @@ export class CallingRepository {
           if (mediaType === z.media.MediaType.VIDEO || mediaType === z.media.MediaType.AUDIO_VIDEO) {
             this.mediaRepository.showNoCameraModal();
           }
-          this.callLogger.error(`Failed to toggle media of type '${mediaType}'`, error);
```
atomrc, line 948: This error message is already prompted as a `warn` in the lower layers

PR #6406, opened 2019-04-18
refactor: Export assets as modules

bennyn:


src/script/assets/AssetCrypto.js
```diff
@@ -17,21 +17,18 @@
  *
  */
 
-window.z = window.z || {};
-window.z.assets = z.assets || {};
-
-z.assets.AssetCrypto = (() => {
+export default class AssetCrypto {
```
atomrc, line 20: I believe this will not work. Previously we exported an object, not a class (which means we don't do `new` on the other side).
I think an object is the right abstraction for this. 

side note: we are moving away from default exports to named ones

bennycode:
Superseded by https://github.com/wireapp/wire-webapp/pull/6428

PR #6409, opened 2019-04-18
refactor: Export connect as modules

ffflorian:


src/script/connect/ConnectService.ts
```diff
@@ -17,33 +17,39 @@
  *
  */
 
-import Logger from 'utils/Logger';
+import * as logdown from 'logdown';
```
atomrc: Please do not use logdown directly but rather the abstraction we have in the webapp (utils/Logger)

ffflorian: It's only imported for the typings, see [line 28](https://github.com/wireapp/wire-webapp/pull/6409/files/711f6b48481707e79ac5869ad4e6c146167bed72#diff-adce89f8e8b09cf98e55a20b88d9cd2fR28).

PR #6428, opened 2019-04-23
refactor: Modularize assets

bennyn:


src/script/assets/AssetCrypto.js
```diff
@@ -17,21 +17,18 @@
  *
  */
 
-window.z = window.z || {};
-window.z.assets = z.assets || {};
-
-z.assets.AssetCrypto = (() => {
+class AssetCrypto {
```
AndyLnd: Why is this a class? As far as I can tell there are just static methods in there which could be exported by itself.

ffflorian: Not sure if we need a class here since all functions are static.

src/script/assets/AssetMapper.js
```diff
@@ -17,18 +17,15 @@
  *
  */
 
-window.z = window.z || {};
-window.z.assets = z.assets || {};
-
-z.assets.AssetMapper = {
+class AssetMapper {
```
ffflorian: Same here with the class.

src/script/assets/AssetCrypto.js
atomrc: I think a class is not the right abstraction here, it is too powerful for what we want to achieve. I would go for exporting the 2 public functions with the 
```
export const decrypt.. = () => {}
```

thus making the other functions private by design

src/script/assets/AssetMapper.js
AndyLnd: I believe this also doesn't need to be a class but rather individually exported functions.

src/script/assets/AssetURLCache.ts
```diff
@@ -19,33 +19,29 @@
 
 import {LRUCache} from '@wireapp/lru-cache';
 
-window.z = window.z || {};
-window.z.assets = z.assets || {};
+class AssetURLCache {
```
AndyLnd: There is also no class necessary here.

PR #6432, opened 2019-04-23
 chore: Use @wireapp/tslint-config

lipis:


src/script/auth/util/handleUtil.ts
```diff
@@ -17,6 +16,7 @@
  * along with this program. If not, see http://www.gnu.org/licenses/.
  *
  */
+/* @flow */
```
lipis: the no header was complaining..

PR #6456, opened 2019-04-24
refactor: Modularize integration

ffflorian:


src/script/assets/AssetMapper.ts
```diff
@@ -48,13 +48,7 @@ export const mapProfileAssetsV1 = (userId: string, pictures: Asset[]): MappedAss
...
 
 // TODO: Fake user entity signature until "User" has been converted to TypeScript
-export const updateUserEntityAssets = (
-  userEntity: {
-    previewPictureResource: ko.Observable<AssetRemoteData>;
-    mediumPictureResource: ko.Observable<AssetRemoteData>;
-  },
-  mappedAssets: MappedAsset = {}
-) => {
+export const updateUserEntityAssets = (userEntity: any, mappedAssets: MappedAsset = {}) => {
```
ffflorian, line 51: @bennyn had to set this to `any`, otherwise the IntegrationMapper, which also uses `updateUserEntityAssets`, won't work with the fake interface.

bennycode: Ok!

PR #6465, opened 2019-04-25
refactor: Migrate AssetMapper to TypeScript

bennyn:


src/script/entity/message/Asset.ts
```diff
@@ -73,10 +59,6 @@ export class Asset {
     return false;
   }
 
-  /**
-   * Check if asset is a audio.
-   * @returns {boolean} Is asset of type audio
-   */
   is_audio() {
```
ffflorian: For all functions:

```suggestion
   is_audio(): boolean {
```

bennycode: 👍

PR #6466, opened 2019-04-25
refactor: Move client service init to client repo

AndyLnd:


src/script/client/ClientRepository.js
```diff
@@ -45,8 +47,9 @@ z.client.ClientRepository = class ClientRepository {
     return 'local_identity';
   }
 
-  constructor(clientService, cryptographyRepository) {
-    this.clientService = clientService;
+  constructor(backendClient, cryptographyRepository) {
+    const storageService = resolve(graph.StorageService);
```
atomrc: Please resolve from the `app.js` side and inject it in the constructor.

src/script/view_model/AuthViewModel.js
```diff
@@ -80,7 +81,7 @@ class AuthViewModel {
       this.cryptography_service,
       this.storage_repository
     );
-    this.client_service = new z.client.ClientService(backendClient, this.storageService);
+    this.client_service = new ClientService(backendClient, this.storageService);
```
atomrc: This line can be removed, you need to change the parameter of the `clientRepo` on the next line.

PR #6469, opened 2019-04-25
refactor: Rewrite telemetry and some utils in TS

ffflorian:


src/script/telemetry/calling/CallSetupTimings.ts
```diff
@@ -59,9 +81,9 @@ class CallSetupTimings {
     return timings;
   }
 
-  time_step(step) {
+  time_step(step: CallSetupSteps) {
     if (this[step] === 0) {
-      this[step] = window.parseInt(window.performance.now() - this.started);
+      this[step] = Math.floor(window.performance.now() - this.started);
```
ffflorian, line 86: Since `window.performance.now()` and `this.started` are both numbers, but `parseInt()` expects a string, we can't use `parseInt()` anymore.

Besides parsing a string, `parseInt()` also floors the parsed number so we can also just use `Math.floor` to achieve the same result.

src/script/telemetry/calling/CallSetupTimings.ts
```diff
@@ -70,8 +92,8 @@ class CallSetupTimings {
 
     this._steps_order().forEach(step => {
       if (this.hasOwnProperty(step)) {
-        const placeholder_key = Array.from(Math.max(26 - step.length, 1)).join(' ');
-        const placeholder_value = Array.from(Math.max(6 - this[step].toString().length, 1)).join(' ');
+        const placeholder_key = new Array(Math.max(26 - step.length, 1)).join(' ');
+        const placeholder_value = new Array(Math.max(6 - this[step].toString().length, 1)).join(' ');
```
ffflorian: `Array.from` actually needs an array-like parameter. What we want to achieve here is a new array with a specified length and fill it.

test/unit_tests/util/MomentSpec.js
```diff
@@ -18,9 +18,7 @@
  */
 
 import moment from 'moment';
-/* eslint-disable no-unused-vars */
-import {isSameDay, isSameMonth, isCurrentYear, isToday} from '../../../src/script/util/moment';
-/* eslint-enable no-unused-vars */
```
ffflorian, line 23: These are used, so why disable `no-unused-vars`?

src/script/util/Crypto.ts
```diff
@@ -19,9 +19,9 @@
 
 import uint32 from 'uint32';
```
ffflorian, line 20: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34990

src/script/telemetry/calling/CallSetupTimings.ts
AndyLnd: Then how about `' '.repeat()`?

src/script/telemetry/calling/CallSetupTimings.ts
atomrc: Thanks for the explanation :+1:

test/unit_tests/util/MomentSpec.js
atomrc: Probably something that evolved organically... There might have been a reason at some point (probably for other files...)

PR #6470, opened 2019-04-25
test: Check profile asset mapping

bennyn:


src/script/assets/AssetMapper.ts
```diff
@@ -20,12 +20,9 @@
 import {Asset} from '../entity/message/Asset';
 import {AssetRemoteData} from './AssetRemoteData';
 
-export interface MappedAsset {
-  medium?: AssetRemoteData;
-  preview?: AssetRemoteData;
-}
+export type MappedAsset = {[index: string]: AssetRemoteData};
```
atomrc, line 23: Why loosening this one up?

bennycode: Because there is `medium`, `preview` and `complete` already and it's not clear yet if there are any others too so I didn't want to limit the possibilities.

ffflorian: Hint: next time use `Record<string, AssetRemoteData>` :sunglasses:

atomrc: cool one, indeed :)

PR #6506, opened 2019-04-30
refactor: Rewrite File Module in TS

ffflorian:


src/script/entity/message/File.ts
```diff
@@ -91,35 +104,35 @@ export class File extends Asset {
...
         this.logger.info(`Downloaded asset in ${download_duration} seconds`);
+        return blob;
       })
       .catch(error => this.logger.error('Failed to download asset', error));
   }
 
-  cancel_download() {
+  cancel_download(): void {
     this.status(AssetTransferState.UPLOADED);
     return this.original_resource().cancelDownload();
```
atomrc: Let's maybe remove the `return` here that doesn't serve any purposes (as shown by the signature)

PR #6507, opened 2019-05-01
refactor: TS-ify ArrayUtil

AndyLnd:


src/script/util/ArrayUtil.ts
```diff
@@ -17,7 +17,7 @@
  *
  */
 
-export const chunk = (array, size) => {
+export const chunk = <T>(array: T[] | Float32Array, size: number) => {
```
ffflorian, line 20: Why do we need a `Float32Array` if we have `T`?

PR #6530, opened 2019-05-03
 refactor: Modularize client

ffflorian:


src/script/client/ClientEntity.ts
```diff
@@ -53,62 +62,56 @@ class ClientEntity {
...
+   * @returns Client ID in pairs of two as an array
    */
-  formatId() {
+  formatId(): string[] {
     return zeroPadding(this.id, 16).match(/.{1,2}/g);
   }
 
   /**
-   * @returns {boolean} True, if the client is the self user's permanent client.
+   * @returns `true`, if the client is the self user's permanent client.
```
bennycode, line 84: There is no check for `isSelfClient`, so it will also be `true` if it is the client of someone else.

src/script/client/ClientEntity.ts
```diff
@@ -53,62 +62,56 @@ class ClientEntity {
...
    */
-  isPermanent() {
+  isPermanent(): boolean {
     return this.type === ClientType.PERMANENT;
   }
 
   /**
-   * @returns {boolean} - True, if it is NOT the client of the self user.
+   * @returns `true`, if it is **not** the client of the self user.
```
bennycode, line 91: We should check for `this.isSelfClient` if we want to be sure that it only affects self clients.

src/script/client/ClientEntity.ts
```diff
@@ -53,62 +62,56 @@ class ClientEntity {
...
    */
-  isRemote() {
+  isRemote(): boolean {
     return !this.isPermanent() && !this.isTemporary();
   }
 
   /**
-   * @returns {boolean} - True, if the client is the self user's temporary client.
+   * @returns `true`, if the client is the self user's temporary client.
```
bennycode, line 98: Sorry for repeating but a check for `this.isSelfClient` should be mandatory to fulfill this requirement.

src/script/client/ClientEntity.ts
ffflorian: This is a refactor and I don't think I want to introduce behaviour changes here. Maybe we can do it in a second PR?

src/script/client/ClientEntity.ts
AndyLnd: I am with @ffflorian on this one and would rather open a separate PR with the changed behaviour.

test/api/TestFactory.js
```diff
@@ -159,19 +159,19 @@ window.TestFactory.prototype.exposeClientActors = function() {
...
+    currentClient.address = '62.96.148.44';
+    currentClient.class = 'desktop';
+    currentClient.cookie = 'webapp@2153234453@temporary@1470926647664';
+    currentClient.id = '132b3653b33f851f';
+    currentClient.label = 'Windows 10';
+    currentClient.location = {lat: 52.5233, lon: 13.4138};
+    currentClient.meta = {is_verified: true, primary_key: 'local_identity'};
+    currentClient.model = 'Chrome (Temporary)';
+    currentClient.time = '2016-10-07T16:01:42.133Z';
+    currentClient.type = 'temporary';
```
AndyLnd: This could also be done with `Object.assign` to avoid some repetition.

src/script/client/ClientEntity.ts
ffflorian: https://github.com/wireapp/wire-webapp/pull/6558

PR #6539, opened 2019-05-06
runfix: Use && instead of AND in condition

Yserz:


.travis.yml
```diff
@@ -117,7 +117,7 @@ deploy:
     bucket_name: wire-webapp-builds
     on:
       tags: true
-      condition: $TRAVIS_TAG =~ .*-staging.* AND $DISTRIBUTION = wire
+      condition: $TRAVIS_TAG =~ .*-staging.* && $DISTRIBUTION = wire
```
ffflorian, line 120: Shouldn't it be `==`?

```suggestion
      condition: $TRAVIS_TAG =~ .*-staging.* && $DISTRIBUTION == "wire"
```

.travis.yml
```diff
@@ -133,7 +133,7 @@ deploy:
...
```
ffflorian, line 136:
```suggestion
      condition: $TRAVIS_TAG =~ .*-staging.* && $DISTRIBUTION == "wire"
```

.travis.yml
Yserz: I haven't figured out all the details but it seems to be a Travis thing here: https://docs.travis-ci.com/user/conditions-v1

PR #6548, opened 2019-05-07
runfix: Make room for localised timestamps (WEBAPP-6170)

AndyLnd:


src/style/content/conversation/message-list.less
```diff
@@ -243,17 +243,17 @@
...
 }
 
 // MESSAGE - TIMESTAMP
 .message-timestamp {
   height: 48px;
   padding-top: 8px; // TODO margin top is not working because of collapsing margins
-  margin-bottom: 16px;
-  border-bottom: 1px solid @separator-color;
+  
+  border-bottom: 1px solid @separator-color;margin-bottom: 16px;
```
ffflorian: What happened here? :eyes:

AndyLnd: 😾 on save autoformat

PR #6558, opened 2019-05-08
chore: Clean-up ClientEntity

ffflorian:


src/script/client/ClientEntity.ts
```diff
@@ -84,21 +81,21 @@ export class ClientEntity {
...
+    return this.isSelfClient && this.type === ClientType.PERMANENT;
   }
 
   /**
    * @returns `true`, if it is **not** the client of the self user.
    */
   isRemote(): boolean {
-    return !this.isPermanent() && !this.isTemporary();
+    return !this.isSelfClient && !this.isPermanent() && !this.isTemporary();
   }
```
AndyLnd: I just checked and it seems as if this method is not used anywhere besides tests.

ffflorian:
I removed the JSDoc because after making clear that it's not necessarily the self user's client, they would just state what's already readable from the function signature.

PR #6605, opened 2019-05-15
refactor: Make finding local users synchronous

atomrc:


src/script/user/UserRepository.js
```diff
@@ -442,14 +442,7 @@ export class UserRepository {
    * @returns {Promise<User>} Resolves with the matching user entity
```
ffflorian: Doesn't return a promise anymore

atomrc: woups :D

PR #6612, opened 2019-05-16
feat: Legal hold modal

AndyLnd:
This does not include all the changes yet.
I would rather open it for reviews now, before it gets to unwieldy.


src/script/components/userList.ts
```diff
@@ -0,0 +1,147 @@
...
+              user.username() === normalizedQuery
+            );
+          });
+        }
+      }
+      resultUsers = userEntities().filter(user => {
+        return connectedUsers.includes(user) || teamRepository.isSelfConnectedTo(user.id);
+      });
+
+      // make sure the selfuser is the first one in the list
```
bennycode:
```suggestion
      // make sure the self user is the first one in the list
```

PR #6632, opened 2019-05-22
feat: Indicate self user in participant item

AndyLnd:


src/script/components/availabilityState.ts
```diff
@@ -17,37 +17,41 @@
...
-    this.label = params.label;
-    this.showArrow = params.showArrow || false;
-    this.theme = params.theme || false;
-    this.AvailabilityType = AvailabilityType;
-  }
-};
+interface AvailabilityStateParams {
+  availability: () => AvailabilityType;
+  label: string;
+  showArrow: boolean;
```
ffflorian:
```suggestion
  showArrow?: boolean;
```

src/script/components/availabilityState.ts
```diff
@@ -17,37 +17,41 @@
...
+  theme: boolean;
```
ffflorian:
```suggestion
  theme?: boolean;
```

PR #6654, opened 2019-05-29
fix: Disable av preference page when a call is ongoing

atomrc:


src/page/template/content/preferences-av.htm
```diff
@@ -2,6 +2,10 @@
   <div class="preferences-titlebar" data-bind="text: t('preferencesAV')"></div>
   <div class="preferences-content" data-bind="fadingscrollbar">
 
+  <!-- ko if: hasOngoingCall() -->
+    Changing the settings during the call is temporarily unavailable
```
atomrc, line 6: This key is temporary and will be removed soon, we agreed to hard code it and only have it in english in the meantime

PR #6663, opened 2019-05-31
fix: Decorrelate AV Preferences' mediaStream from the one used in a call (WEBAPP-6227)

atomrc:
Previously, mediaStream were shared accross the preferences page and a potential current call. Which lead to hard-to-find mediaStream leaks (because it's hard to know if you need to release the tracks of the media stream from the call and/or the preference page).

Now, we have discinct workflow for both the page and a call. Which means they are both responsible for cleaning their own mediaStream, but never the media stream of someone else. It makes the dataflow easier to understand and to scale in the future

### DIFF
--- a/src/script/media/MediaStreamHandler.js
+++ b/src/script/media/MediaStreamHandler.js
@@ -196,11 +196,14 @@ export class MediaStreamHandler {
 
   /**
    * Replace the MediaStream after a change of the selected input device.
-   * @param {MediaStreamInfo} mediaStreamInfo - Info about new MediaStream
-   * @returns {undefined} No return value
+   * @param {MediaStream} mediaStream - the new mediastream (tracks will be clone, this mediaStream can be disposed of once changed)
+   * @param {MediaType} mediaType - Type of the media that needs to be changed
+   * @returns {Promise<void>} Resolves when the local media stream and the call have been updated
    */
-  changeMediaStream(mediaStreamInfo) {
-    const mediaStream = mediaStreamInfo.stream;
+  changeMediaStream(mediaStream, mediaType) {
+    if (!this.localMediaStream()) {
+      return Promise.resolve();
+    }
 
     const logMessage = `Received new MediaStream containing '${mediaStream.getTracks().length}' track/s`;
     const logObject = {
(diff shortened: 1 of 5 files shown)


src/script/media/MediaStreamHandler.js
```diff
@@ -275,16 +284,9 @@ export class MediaStreamHandler {
 
     return constraintsPromise
       .then(streamConstraints => this.requestMediaStream(mediaType, streamConstraints))
-      .then(mediaStreamInfo => {
-        // FIXME: the mediaStreamInUse should be more intelligent and handle all scenarios where the stream is actually needed
-        if (!isPreferenceChange && !this.mediaStreamInUse()) {
```
atomrc, line 280: This is a typical example of a place where the `mediaStreamHandler` needed to "guess" where the update came from (preferences or in the call). Now, because the `localMediaStream` of the `streamHandler` is **only** used for the current call, we will never need to care about preferences changes in here

src/script/media/MediaStreamHandler.js
```diff
@@ -807,21 +808,6 @@ export class MediaStreamHandler {
   // Media handling
   //##############################################################################
 
-  /**
-   * Check for active calls that need a MediaStream.
-   * @returns {boolean} Returns true if an active media stream is needed for at least one call
-   */
-  mediaStreamInUse() {
```
atomrc, line 814: This method also becomes obsolete because the mediaStream is not shared anymore. No need to check if it's "in use" by some other part of the code, only this class really "needs" it

src/script/media/MediaStreamHandler.js
```diff
@@ -196,51 +196,58 @@ export class MediaStreamHandler {
 
   /**
    * Replace the MediaStream after a change of the selected input device.
-   * @param {MediaStreamInfo} mediaStreamInfo - Info about new MediaStream
-   * @returns {undefined} No return value
+   * @param {MediaStream} mediaStream - the new mediastream (tracks will be clone, this mediaStream can be disposed of once changed)
```
ffflorian:
```suggestion
   * @param {MediaStream} mediaStream - the new mediastream (tracks will be cloned, this mediaStream can be disposed of once changed)
```

PR #6679, opened 2019-06-07
feat: Query legalhold state from backend

bennyn:


src/script/main/app.js
```diff
@@ -529,24 +529,32 @@ class App {
    * Initiate the self user by getting it from the backend.
    * @returns {Promise<User>} Resolves with the self user entity
    */
-  _initiateSelfUser() {
-    return this.repository.user.getSelf().then(userEntity => {
-      this.logger.info(`Loaded self user with ID '${userEntity.id}'`);
+  async _initiateSelfUser() {
```
ffflorian, line 532: We can't use `async` in the webapp AFAIK

ffflorian: Seems like we can now :slightly_smiling_face:

PR #6713, opened 2019-06-13
feat: Add remote fingerprint calculation by PreKey

bennyn:


test/unit_tests/cryptography/CryptographyRepositorySpec.js
```diff
@@ -100,6 +100,56 @@ describe('CryptographyRepository', () => {
...
+        '51',
+        '31',
+        'a6',
+        'f7',
+        'b8',
+        '6c',
+        '01',
+        '81',
+        'b9',
+        '1a',
```
ffflorian: Let's add an eslint exception here so that we can have the array in a single line

PR #6724, opened 2019-06-17
fix: Chunk user requests to the backend (WEBAPP-6240)

AndyLnd:


src/script/user/UserService.js
```diff
@@ -126,13 +127,19 @@ export class UserService {
...
    */
   getUsers(userIds) {
-    return this.backendClient.sendRequest({
-      data: {
-        ids: userIds.join(','),
-      },
-      type: 'GET',
-      url: UserService.URL.USERS,
-    });
+    const maxRequestCount = 50;
```
Yserz: `chunkSize`

PR #6742, opened 2019-06-19
feat: Show loading spinners

AndyLnd:


src/script/main/globals.js
```diff
@@ -82,7 +82,7 @@ import userInputGlobal from 'Components/userInput.js';
 import guestModeToggleGlobal from 'Components/guestModeToggle.js';
 import infoToggleGlobal from 'Components/infoToggle.js';
 import iconsGlobal from 'Components/icons.js';
-import modalGlobal from 'Components/modal.js';
+import modalGlobal from 'Components/modal.ts';
```
ffflorian:
```suggestion
import modalGlobal from 'Components/modal';
```

And let's move this somewhere up or down to clarify that this is actually needed globally.

PR #6753, opened 2019-06-20
feat: Decrypt but ignore messages from legal hold clients

bennyn:
It looks like `_distributeEvent` is a "fire and forget" method which is why I am not awaiting it in `_handleEventDistribution` of `EventRepository.js`. Or have I overlooked something?

### DIFF
--- a/src/script/event/EventRepository.js
+++ b/src/script/event/EventRepository.js
@@ -565,14 +565,22 @@ export class EventRepository {
    * @param {EventRepository.SOURCE} source - Source of notification
    * @returns {undefined} No return value
    */
-  _distributeEvent(event, source) {
+  async _distributeEvent(event, source) {
     const {conversation: conversationId, from: userId, type} = event;
 
     const hasIds = conversationId && userId;
-    const logMessage = hasIds
-      ? `Distributed '${type}' event for conversation '${conversationId}' from user '${userId}'`
-      : `Distributed '${type}' event`;
-    this.logger.info(logMessage, event);
+    if (hasIds) {
+      const clientEntity = await this.userRepository.client_repository.loadClientFromDb(userId, event.from_client);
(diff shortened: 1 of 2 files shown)


src/script/event/EventRepository.js
```diff
@@ -636,6 +644,7 @@ export class EventRepository {
         }, Promise.resolve(mappedEvent));
       })
       .then(mappedEvent => {
+        mappedEvent.from_client = event.data && event.data.sender ? event.data.sender : '';
```
ffflorian:
```suggestion
        mappedEvent.from_client = event.data && event.data.sender || '';
```

src/script/event/EventRepository.js
```diff
@@ -565,14 +565,22 @@ export class EventRepository {
    * @param {EventRepository.SOURCE} source - Source of notification
    * @returns {undefined} No return value
```
ffflorian:
```suggestion
   * @returns {Promise<void>} Empty promise
```

ffflorian:
Even though `_distributeEvent` is a "fire-and-forget" method, it should still be awaited.

bennycode:
@ffflorian It depends. Useless awaits can also make the app slower but to be on a safe side I put an `await` now. :)

ffflorian:
> Useless awaits can also make the app slower

`await` should be non-blocking, do you have documentation on it slowing apps down?

bennycode:
> await should be non-blocking, do you have documentation on it slowing apps down?

Yes, I do: https://medium.com/front-end-weekly/async-await-is-not-about-making-asynchronous-code-synchronous-ba5937a0c11e

Check the article's sample with:

```
const data1 = await downloadFromService1();
const data2 = await downloadFromService2();
const data3 = await downloadFromService3();
```

bennycode:
Requirements have changed. Functionality is not needed at this point in time.

PR #6793, opened 2019-06-24
refactor: Remove useless returns

lipis:


src/script/auth/module/reducer/selfReducer.ts
```diff
@@ -68,9 +68,11 @@ export function selfReducer(state: SelfState = initialState, action: AppActions)
     case SELF_ACTION.CONSENT_GET_SUCCESS: {
       return {
         ...state,
-        consents: action.payload.reduce((consentAccumulator, consent) => {
-          return {...consentAccumulator, [consent.type]: consent.value};
-        }, {}),
+        consents: action.payload.reduce(
+          (consentAccumulator, consent) => ({...consentAccumulator, [consent.type]: consent.value}),
+          {},
```
ffflorian: The original code was much more readable. While I agree with the rest of the PR, let's keep this as it was.

AndyLnd: I really don't think that it was more readable before. I prefer this concise style.

PR #6811, opened 2019-06-25
feat: Conversation legal hold state

AndyLnd:


src/script/entity/message/LegalHoldMessage.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
bennycode: 2019

PR #6816, opened 2019-06-26
feat: Add LH status to messages (WEBAPP-6266)

ffflorian:


src/script/assets/AssetService.ts
```diff
@@ -39,7 +39,8 @@ export interface CompressedImage {
 }
 
 export interface AssetUploadOptions {
-  expectsReadConfirmation: boolean;
+  expectsReadConfirmation?: boolean;
```
bennycode: So far `expectsReadConfirmation` has been mandatory. Why do you want to make it optional? Isn't it good to have it always being `true` or `false`?

ffflorian: The default will be falsy, so we can omit it, too. Just wanted to make the API easier :slightly_smiling_face:

bennycode: I would rather have a pure `false` than working with falsy to be explicit.

ffflorian: Done!

PR #6875, opened 2019-07-04
feat: Use storage engine for other StorageService calls

pwnsdx:


bennycode:
We will continue here https://github.com/wireapp/wire-webapp/pull/6902

PR #6902, opened 2019-07-08
refactor: Use StoreEngine in StorageService where possible

bennyn:


src/script/storage/StorageService.ts
```diff
@@ -185,35 +185,20 @@ export class StorageService {
...
+  async deleteDatabase(): Promise<boolean> {
+    try {
+      await this.engine.purge();
+      this.logger.info(`Clearing IndexedDB '${this.dbName}' successful`);
+    } catch (error) {
+      this.logger.error(`Clearing IndexedDB '${this.dbName}' failed`);
+      throw error;
     }
-    this.logger.error(`IndexedDB '${this.dbName}' not found`);
     return Promise.resolve(true);
```
ffflorian:
```suggestion
    return true;
```

PR #6917, opened 2019-07-09
refactor: Cleanup currency and availability

ffflorian:


src/script/cryptography/CryptographyMapper.js
```diff
@@ -234,23 +233,21 @@ export class CryptographyMapper {
...
-            case Availability.Type.AWAY:
-              return AvailabilityType.AWAY;
-            case Availability.Type.BUSY:
-              return AvailabilityType.BUSY;
-            default:
-              const message = 'Unhandled availability type';
-              throw new z.error.CryptographyError(z.error.CryptographyError.TYPE.UNHANDLED_TYPE, message);
-          }
-        })(),
+        availability: availability.type,
```
atomrc, line 250: I believe the types are different here.

`Availability.Type` vs `AvailabilityType`.

Do they resolve to the same values?

ffflorian: Yes, they are the same:

https://github.com/wireapp/wire-webapp/blob/c13f05233c2a4e0b1a1d4dfc2ced055ad3725a0b/src/script/user/AvailabilityType.ts#L20-L25

[`@wireapp/protocol-messaging@1.24.1/web/Protobuf.d.ts`](https://unpkg.com/@wireapp/protocol-messaging@1.24.1/web/Protobuf.d.ts):
```ts
export namespace Availability {
  enum Type {
    NONE = 0,
    AVAILABLE = 1,
    AWAY = 2,
    BUSY = 3
  }
}
```

PR #6930, opened 2019-07-10
runfix: Avoid race condition when fetching users (WEBAPP-6282)

atomrc:
At app load time, we update the conversations (fetch the users and make sure everything is there). But since we did not wait for this to be over when loading the app, we would also load those users again when displaying the conversation. 
This resulted in 2 requests being sent in parallel and 2 user entities being created for the same user

### DIFF
--- a/src/script/conversation/ConversationRepository.js
+++ b/src/script/conversation/ConversationRepository.js
@@ -759,13 +759,15 @@ export class ConversationRepository {
 
   /**
    * Update all conversations on app init.
-   * @returns {undefined} No return value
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationRepository.js
```diff
@@ -782,7 +784,7 @@ export class ConversationRepository {
    * @returns {undefined} No return value
```
bennycode: Now it returns something?

atomrc: True that

PR #7058, opened 2019-07-29
refactor: Convert user to TypeScript

ffflorian:


src/script/user/UserPermission.ts
```diff
@@ -19,45 +19,40 @@
...
+enum PUBLIC_FEATURES {
+  CREATE_GROUP_CONVERSATION = 1 << 13,
+  CREATE_GUEST_ROOM = 1 << 14,
+  UPDATE_CONVERSATION_SETTINGS = 1 << 15,
+  UPDATE_GROUP_PARTICIPANTS = 1 << 16,
+  MANAGE_SERVICES = 1 << 17,
+  MANAGE_TEAM = 1 << 18,
+  INVITE_TEAM_MEMBERS = 1 << 19,
+  CHAT_WITH_SERVICES = 1 << 20,
+  SEARCH_UNCONNECTED_USERS = 1 << 21,
```
ffflorian: It's way better readable like this.

src/script/user/UserRepository.ts
```diff
@@ -186,41 +232,42 @@ export class UserRepository {
             this.logger.log(`Loaded state of '${users.length}' users from database`, users);
 
             const mappingPromises = users.map(user => {
-              return this.get_user_by_id(user.id).then(userEntity => userEntity.availability(user.availability));
+              return this.get_user_by_id(user.id).then(userEntity => userEntity.availability(user.availability()));
```
ffflorian, line 223: Small fix

src/script/user/UserRepository.ts
```diff
@@ -536,18 +575,17 @@ export class UserRepository {
...
+  _upgradePictureAsset<T extends UserUpdate>(userData: T): T {
     const hasPicture = userData.picture.length;
     const hasAsset = userData.assets.length;
 
     if (hasPicture) {
       if (!hasAsset) {
         // if there are no assets, just upload the old picture to the new api
         const {medium} = mapProfileAssetsV1(userData.id, userData.picture);
-        medium.load().then(imageBlob => this.change_picture(imageBlob));
+        medium.load().then(imageBlob => this.change_picture(imageBlob as Blob));
```
ffflorian, line 551: Not sure if we should do a falsy check here since `load()` can return `void`.

src/script/user/UserRepository.ts
```diff
@@ -646,43 +683,43 @@ export class UserRepository {
...
 
-    return Promise.all([getLocalUser(userId), this.user_service.getUser(userId)])
+    return Promise.all([getLocalUser(), this.user_service.getUser(userId)])
       .then(([localUserEntity, updatedUserData]) =>
         this.user_mapper.updateUserFromObject(localUserEntity, updatedUserData),
       )
       .then(userEntity => {
         if (this.isTeam()) {
-          this.mapGuestStatus([userEntity]);
+          this.mapGuestStatus([userEntity || undefined]);
```
ffflorian: Just casting `void` to `undefined`.

src/script/user/UserPermission.ts
lipis: the other option had some weird beauty.. can't decide :)

lipis: and no magic numbers with the old approach

ffflorian: These numbers are not magic, they are just incerementing. The old approach was using a changing variable to generate a static, which is dangerous IMO.

ffflorian: Now you can even sort the object and will not destroy its functionality.

src/script/assets/AssetMetaDataBuilder.ts
```diff
@@ -112,7 +112,7 @@ const buildMetadataVideo = (videoFile: File | Blob): Promise<Asset.VideoMetaData
  * This needs to be done because error events are not standardized between browser implementations.
  * @private
  * @param {Event} event - Error event
- * @returns {MediaError} Error object
```
atomrc, line 115: I think you can remove the jsdoc comment altogether. It doesn't add any value now :)
(same for the others)

src/script/user/UserPermission.ts
atomrc: Please don't sort this one!! It's gonna be awful and very error prone to add new values, else!!!

atomrc: I think we took the decision to use a variable with @AndyLnd because it leaves less room for error from a human. You can just copy paste a line and create the new entry without worrying about the current state of the enum

ffflorian: > You can just copy paste a line and create the new entry without worrying about the current state of the enum

But you can easily paste it into the wrong line without knowing directly what the resulting number will be since you need to count how often `bitsCounter` is incremented. And you also have to take into consideration that it is incrementing **after** it is written into the object...

src/script/user/UserRepository.ts
atomrc: wow! Maybe we never saw it cause ko resolves the observable automatically... (?)

ffflorian: Maybe, but then the knockout types would be wrong.

src/script/user/UserRepository.ts
```diff
@@ -482,10 +489,10 @@ export class UserRepository {
         });
     };
 
-    const chunksOfUserIds = chunk(userIds, z.config.MAXIMUM_USERS_PER_REQUEST);
+    const chunksOfUserIds = chunk(userIds, (z as any).config.MAXIMUM_USERS_PER_REQUEST) as string[][];
```
atomrc: If you are courageous enough, you can import `config.js` directly (but can be done later, no problem)

src/script/user/UserRepository.ts
atomrc: We probably handle the `undefined` case on the UI level... (just a guess).
I would not change anything as of now

src/script/user/UserRepository.ts
```diff
@@ -587,24 +586,22 @@ export class UserRepository {
...
+    const _find_user = (user_id: string) => {
       return this.findUserById(user_id) || user_id;
     };
 
     const find_users = user_ids.map(user_id => _find_user(user_id));
 
     return Promise.all(find_users).then(resolve_array => {
-      const known_user_ets = resolve_array.filter(array_item => !_.isString(array_item));
-      const unknown_user_ids = resolve_array.filter(array_item => _.isString(array_item));
+      const known_user_ets = resolve_array.filter(array_item => typeof array_item !== 'string') as User[];
```
atomrc: Please keep the underscore check, it does more than just `typeof`. 

In this case a `new String` will return `false`.

```
typeof "test"; // => "string"
typeof new String("test"); // => object
```

atomrc: Please keep the underscore check, it does more than just `typeof`. 

In this case a `new String` will return `false`.

```
typeof "test"; // => "string"
typeof new String("test"); // => object
```

src/script/user/UserRepository.ts
```diff
@@ -616,23 +613,19 @@ export class UserRepository {
 
   /**
    * Is the user the logged in user.
-   * @param {User|string} user_id - User entity or user ID
-   * @returns {boolean} Is the user the logged in user
    */
-  is_me(user_id) {
-    if (!_.isString(user_id)) {
+  is_me(user_id: User | string): boolean {
+    if (typeof user_id !== 'string') {
```
atomrc: Please keep the underscore check, it does more than just `typeof`. 

In this case a `new String` will return `false`.

```
typeof "test"; // => "string"
typeof new String("test"); // => object
```

src/script/user/UserMapper.ts
```diff
@@ -146,7 +135,7 @@ export class UserMapper {
       if (this.serverTimeHandler.timeOffset() !== undefined) {
         setAdjustedTimestamp();
       } else {
-        this.serverTimeHandler.timeOffset.subscribe_once(setAdjustedTimestamp);
+        (this.serverTimeHandler.timeOffset as any).subscribe_once(setAdjustedTimestamp);
```
ffflorian, line 139: `subscribe_once` usually doesn't exist in knockout, and I couldn't find a simple solution to add a type to knockout that reflects adding a additional function.

src/script/user/UserPermission.ts
atomrc: The thing is you never (or very rarely?) need to know what is the value is.
It is true that you can copy at the wrong place, I am open to using fixed values.

Just the sorting part, I really think it should not be done.

src/script/user/UserRepository.ts
ffflorian: I don't see where anyone would use `new String()` but I get your point.

src/script/user/UserMapper.ts
atomrc: Yeah that's a custom function that we add in the `commonBindings` file. 
There should be a way in TS to define those but I am no expert with this.
I am fine with the `any` here.

atomrc: Maybe a cleaner solution would be to make the `subscribe_once` function a raw function (not one that is dynamically added to the prototype). 

Something like `subscribe_once(observable, subscribeFn)` This way we import the file everywhere we need it (instead of relying on another file to have imported it before) and we won't have to tweak typescript.

src/script/user/UserPermission.ts
ffflorian: > Just the sorting part, I really think it should not be done.

I agree.

src/script/user/UserRepository.ts
atomrc: Yeah, we don't right now in our code. But it could come from a library or another developer in the future. Better safe than sorry

src/script/user/UserMapper.ts
ffflorian: I'll add a comment so we can do this later.

test/unit_tests/user/RichProfileRepositorySpec.js
```diff
@@ -35,7 +35,7 @@ describe('RichProfileRepository', () => {
 
       return richProfileRepository.getUserRichProfile(userId).then(richProfile => {
         expect(richProfileRepository.backendClient.sendRequest).toHaveBeenCalledTimes(1);
-        expect(richProfile).toBe(richProfile);
```
atomrc, line 38: Why? This test has lost value with this change.

atomrc: Maybe `toEqual` is the one you are looking for (if the problem is that it's a new object)

PR #7059, opened 2019-07-30
feat: Optional 'public computer' checkbox checked by default (WEBAPP-6352)

AndyLnd:


server/ServerConfig.ts
```diff
@@ -10,6 +10,7 @@ export interface ServerConfig {
     ENVIRONMENT: string;
     FEATURE: {
       CHECK_CONSENT: boolean;
+      DEFAULT_TEMPORARY_DEVICE: boolean;
```
Yserz: DEFAULT_LOGIN_TEMPORARY_DEVICE

PR #7082, opened 2019-08-08
Check media devices support before calling them

backspac:
## Type of change

Fixes crash on startup when media devices are disabled.


atomrc:
LGTM. Thanks for the fix @backspac

PR #7085, opened 2019-08-09
 feat: Add hooks for the MemoryEngine

ffflorian:


src/script/storage/StorageService.ts
```diff
@@ -391,7 +416,14 @@ export class StorageService {
   async update<T = Object>(storeName: string, primaryKey: string, changes: T): Promise<number> {
     try {
       if (this.isTemporaryAndNonPersistant) {
+        const oldRecord = await this.engine.read<unknown>(storeName, primaryKey);
         await this.engine.update(storeName, primaryKey, changes);
+        const newRecord = await this.engine.read<unknown>(storeName, primaryKey);
+
+        this.dbListeners
```
atomrc: Let's try to not have this code duplicated.
Maybe writing a function `warnListeners(store, eventType, oldRecord, newRecord)` could abstract the implementation away.

src/script/storage/StorageService.ts
```diff
@@ -358,12 +369,26 @@ export class StorageService {
       throw new z.error.StorageError(z.error.StorageError.TYPE.NO_DATA);
     }
 
-    try {
-      const newKey = await this.engine.updateOrCreate(storeName, primaryKey, entity);
-      return newKey;
-    } catch (error) {
-      this.logger.error(`Failed to put '${primaryKey}' into store '${storeName}'`, error);
-      throw error;
+    if (this.isTemporaryAndNonPersistant) {
```
atomrc: Can we switch on the lower level engine type rather than on the high level feature flag.
The condition to manually trigger the event listeners is not whether the client is temporary but rather if the engine used doesn't support hooks natively.
Let's reflect that idea in the code

bennycode: Good idea, we can determine that using a check like `this.engine instanceof MemoryEngine`.

PR #7114, opened 2019-08-14
feat: Save conversation input in memory when using in-memory client (WEBAPP-6382)

ffflorian:


src/script/view_model/content/InputBarViewModel.js
```diff
@@ -278,9 +280,9 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
     });
 
     this.conversationEntity.subscribe(this.loadInitialStateForConversation.bind(this));
-    this.draftMessage.subscribe(message => {
+    this.draftMessage.subscribe(async message => {
       if (this.conversationEntity()) {
-        this._saveDraftState(this.conversationEntity(), message.text, message.mentions, message.reply);
+        await this._saveDraftState(this.conversationEntity(), message.text, message.mentions, message.reply);
```
atomrc: Why awaiting here?

src/script/view_model/content/InputBarViewModel.js
```diff
@@ -328,22 +330,46 @@ z.viewModel.content.InputBarViewModel = class InputBarViewModel {
...
+    return data;
+  }
+
+  /**
+   * @param {string} storageKey - The primary key for the data
+   * @param {any} data - The data to save
+   * @returns {Promise<void>} No return value
+   */
+  async _storeValue(storageKey, data) {
+    if (isTemporaryClientAndNonPersistent()) {
```
atomrc: you could use the same pattern as the `_loadValue` function

```
return isTemporaryClient() ? await ... : storeValue()
```

atomrc: Oh wait no, this might change the signature of the function

src/script/view_model/content/InputBarViewModel.js
ffflorian: Because I don't like floating promises.

src/script/view_model/content/InputBarViewModel.js
ffflorian: Yeah then it would be returning data where it shouldn't be.

PR #7116, opened 2019-08-14
refactor: Export ParticipantAvatar as module

ffflorian:


src/script/components/mentionSuggestions.js
```diff
@@ -179,7 +182,7 @@ ko.components.register('mention-suggestions', {
           event: { mouseenter: onMouseEnter},
           css: {'mention-suggestion-list__item--highlighted': suggestion === selectedSuggestion()},
           attr: {'data-uie-value': suggestion.id, 'data-uie-selected': suggestion === selectedSuggestion()}" data-uie-name="item-mention-suggestion">
-          <participant-avatar params="participant: suggestion, size: z.components.ParticipantAvatar.SIZE.XXX_SMALL"></participant-avatar>
+          <participant-avatar params="participant: suggestion, size: ParticipantAvatar.SIZE.XXX_SMALL"></participant-avatar>
```
atomrc, line 185: How about we try something?

Instead of publishing the `ParticipanAvatar` on the scope of the component, how about we use the string literal and just statically replace in the string. 

```
`
... size: ${ParticipantAvatar.SIZE.XXX} ...
`
```

WDYT?

atomrc: Humm actually this might be a problem cause if it's a string we would need to wrap it in quotes...
Let's forget about this idea, maybe

PR #7125, opened 2019-08-19
feat: Add the single value store 

atomrc:
This will allow passing values from the login page to the rest of the app without any writings on the disk

### DIFF
--- a/src/script/auth/page/Login.tsx
+++ b/src/script/auth/page/Login.tsx
@@ -50,6 +50,7 @@ import {Redirect, RouteComponentProps, withRouter} from 'react-router';
 import {noop} from 'Util/util';
 import {isValidEmail, isValidPhoneNumber, isValidUsername} from 'Util/ValidationUtil';
 
+import {save} from 'Util/valueStore';
 import {loginStrings, logoutReasonStrings} from '../../strings';
 import AppAlreadyOpen from '../component/AppAlreadyOpen';
 import RouterLink from '../component/RouterLink';
@@ -237,6 +238,11 @@ class Login extends React.Component<CombinedProps, State> {
         return hasKeyAndCode
           ? this.props.doLoginAndJoin(login, this.state.conversationKey, this.state.conversationCode)
           : this.props.doLogin(login);
+      })
+      .then(() => {
+        const secretKey = new Uint32Array(64);
+        self.crypto.getRandomValues(secretKey);
+        return save(secretKey);
       })
       .then(this.navigateChooseHandleOrWebapp)
       .catch((error: Error | BackendError) => {
--- a/src/script/util/valueStore.ts
+++ b/src/script/util/valueStore.ts
@@ -21,7 +21,7 @@ import {Config} from '../auth/config';
 
 /**
  * The value store util allows storing a single value across page navigation.
- * This value will never be stored on the disk, which makes it a good candidate for storing cryptographic key.
+ * This value will never be stored on the disk, which makes it a good candidate for storing cryptographic key material.
  * There are no guarantees regarding the time the value will stay in memory. Should not be used as a reliable store.
  * Main use case is to pass on a value from a page to another page.
  */
@@ -55,8 +55,8 @@ function sendMessage(worker: ServiceWorker, action: any): Promise<any> {
     };
     try {
       worker.postMessage(action, [messageChannel.port2]);
-    } catch (err) {
-      reject(err);
+    } catch (error) {
+      reject(error);
     }
   });
 }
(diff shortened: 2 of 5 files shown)


src/script/util/valueStore.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * The value store util allows storing a single value across page navigation.
+ * This value will never be stored on the disk, which makes it a good candidate for storing cryptographic key.
+ * There are no guarantees regarding the time the value will stay in memory. Should not be used as a reliable store.
+ * Main use case is to pass on a value from a page to another page.
+ */
+export interface ValueStoreApi {
+  get: () => Promise<any>;
+  save: (value: any) => Promise<any>;
```
ffflorian: Is it possible to type this?

src/script/util/valueStore.ts
```diff
@@ -0,0 +1,60 @@
...
+function sendMessage(worker: ServiceWorker, action: any): Promise<any> {
+  const messageChannel = new MessageChannel();
+  return new Promise((resolve, reject) => {
+    messageChannel.port1.onmessage = event => {
+      resolve(event.data);
+    };
+    try {
+      worker.postMessage(action, [messageChannel.port2]);
+    } catch (e) {
+      reject(e);
```
ffflorian: Let's use `error` instead of `e`.

src/script/util/valueStore.ts
atomrc: You can literally store anything you want. I think `any` is the proper type here. Any suggestions ?

ffflorian: The question is, does it return the value? Then we could replace `any` with a generic.

atomrc: Oh, right !

src/script/util/valueStore.ts
```diff
@@ -25,7 +25,7 @@
  */
 export interface ValueStoreApi {
   get: () => Promise<any>;
-  save: (value: any) => Promise<any>;
+  save: <T>(value: T) => Promise<T>;
```
ffflorian: Nice!

src/script/auth/page/Login.tsx
```diff
@@ -238,6 +239,11 @@ class Login extends React.Component<CombinedProps, State> {
           ? this.props.doLoginAndJoin(login, this.state.conversationKey, this.state.conversationCode)
           : this.props.doLogin(login);
       })
+      .then(() => {
+        const secretKey = new Uint32Array(64);
+        self.crypto.getRandomValues(secretKey);
+        return save(secretKey);
```
atomrc, line 245: This is mostly demo code (since we are not sure yet how this is going to be consumed)

src/script/util/valueStore.ts
```diff
@@ -0,0 +1,62 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Config} from '../auth/config';
+
+/**
+ * The value store util allows storing a single value across page navigation.
+ * This value will never be stored on the disk, which makes it a good candidate for storing cryptographic key.
```
bennycode:
```suggestion
 * This value will never be stored on the disk, which makes it a good candidate for storing cryptographic key material.
```

src/script/util/valueStore.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function sendMessage(worker: ServiceWorker, action: any): Promise<any> {
+  const messageChannel = new MessageChannel();
+  return new Promise((resolve, reject) => {
+    messageChannel.port1.onmessage = event => {
+      resolve(event.data);
+    };
+    try {
+      worker.postMessage(action, [messageChannel.port2]);
+    } catch (err) {
```
bennycode:
```suggestion
    } catch (error) {
```

src/script/util/valueStore.ts
```diff
@@ -0,0 +1,62 @@
...
+      reject(err);
```
bennycode:
```suggestion
      reject(error);
```

src/worker/sw-value-store.js
```diff
@@ -0,0 +1,50 @@
...
+    event.waitUntil(self.clients.claim());
+    keepAlive();
+  });
+
+  global.addEventListener('install', event => event.waitUntil(global.skipWaiting()));
+  global.addEventListener('message', event => {
+    const client = event.ports[0];
+    const action = actions[event.data.action];
+    if (!action) {
+      console.error(`Action '${event.data.action}' doesn't exist in valueStoreSw`);
```
bennycode: What`s `valueStoreSw`? I guess value store service worker but let's use a less technical term here.

PR #7163, opened 2019-08-22
runfix: Revert invalid client check to fix login loop

bennyn:
This reverts https://github.com/wireapp/wire-webapp/pull/7124.


bennycode:
False positive. It's not the fix we are looking for.

PR #7189, opened 2019-08-26
runfix: Do not show join notification when call is ongoing (WEBAPP-6397)

atomrc:


src/script/calling/CallingRepository.ts
```diff
@@ -335,7 +335,12 @@ export class CallingRepository {
     switch (type) {
       case CALL_MESSAGE_TYPE.SETUP:
       case CALL_MESSAGE_TYPE.GROUP_START:
-        this.injectActivateEvent(conversationId, userId, time, source);
+        const activeCall = this.findCall(conversationId);
+        const ignoreNotificationStates = [CALL_STATE.MEDIA_ESTAB, CALL_STATE.ANSWERED, CALL_STATE.OUTGOING];
+        if (!activeCall || !ignoreNotificationStates.includes(activeCall.state())) {
+          // we want to ignore call start event that already have an active call (wether it's ringing or connected).
```
ffflorian:
```suggestion
          // we want to ignore call start events that already have an active call (whether it's ringing or connected).
```

PR #7194, opened 2019-08-27
feat: Do not block AV preferences when only one device fails (WEBAPP-6401)

atomrc:


src/script/view_model/content/PreferencesAVViewModel.js
```diff
@@ -155,14 +163,36 @@ export class PreferencesAVViewModel {
         this.devicesHandler.refreshMediaDevices();
         return stream;
       })
+      .catch(error => {
+        // if getting the streams all together failed, we try to get them separatly.
```
ffflorian:
```suggestion
        // if getting the streams all together failed, we try to get them separately.
```

PR #7195, opened 2019-08-27
feat: Enable persistent storage

ffflorian:
Follow-up to https://github.com/wireapp/wire-web-packages/pull/2224.


src/script/storage/StorageService.ts
```diff
@@ -91,9 +91,10 @@ export class StorageService {
    * Initialize the IndexedDB for a user.
    *
    * @param userId - User ID
+   * @param requestPersistentStorage - if a persistent storage should be requested
    * @returns Resolves with the database name
    */
-  async init(userId = this.userId): Promise<string> {
+  async init(userId = this.userId, requestPersistentStorage = false): Promise<string> {
```
bennycode:
```suggestion
  async init(userId: string = this.userId, requestPersistentStorage: boolean = false): Promise<string> {
```

PR #7199, opened 2019-08-28
refactor: Turn TeamName into functional component

AndyLnd:


src/script/auth/page/TeamName.tsx
```diff
@@ -51,7 +51,7 @@ import {ROUTE} from '../route';
 import {parseError, parseValidationErrors} from '../util/errorUtil';
 import Page from './Page';
 
-interface Props extends React.HTMLAttributes<TeamName>, RouteComponentProps<{}> {}
+interface Props extends RouteComponentProps<{}> {}
```
Yserz:
```suggestion
interface Props extends React.HTMLProps<HTMLDivElement>, RouteComponentProps<{}> {}
```

src/script/auth/page/TeamName.tsx
```diff
@@ -64,37 +64,32 @@ interface DispatchProps {
...
-}
-
-class TeamName extends React.Component<Props & ConnectedProps & DispatchProps & InjectedIntlProps, State> {
-  private readonly teamNameInput: React.RefObject<any> = React.createRef();
-  state: State = {
-    enteredTeamName: this.props.teamName || '',
-    error: null,
-    isValidTeamName: !!this.props.teamName,
-  };
+const TeamName = ({intl: {formatMessage: _}, ...props}: Props & ConnectedProps & DispatchProps & InjectedIntlProps) => {
```
Yserz, line 67: Let's destructure the props here

AndyLnd: Most of the time they are used just once and i would argue that it is easier to read and understand where the values are coming from if we address them by `props.value`.

Yserz: I would argue it's more readable that way because the actual code is more concise. Also that's what we do in new code.

PR #7204, opened 2019-08-29
debug: Add the call debug tool

atomrc:
This is a tool to help debugging and monitoring calls. 
Will help QA and developers understanding what audio/video legs are working (or not). 

It can be toggled on/off by pressing `ctrl+alt+c` and is available when debugging is enabled

This is what is looks like.
screenshot


src/script/util/DebugUtil.js
```diff
@@ -31,6 +32,9 @@ import {BackendEvent} from '../event/Backend';
 import {StorageSchemata} from '../storage/StorageSchemata';
 import {EventRepository} from '../event/EventRepository';
 
+import {h, init} from 'snabbdom';
```
atomrc: It uses `snabbdom` as a lightweight virtual dom lib. If you prefer another vdom lib, we can consider other options.
The vanilla JS version is, either, very very slow or unreadable (when optimized). Thus the idea of using a vdom lib.

src/script/calling/CallingRepository.ts
```diff
@@ -695,6 +695,7 @@ export class CallingRepository {
 
     if (Object.keys(missingStreams).length === 0) {
       // we have everything in cache, just return the participant's stream
+      // (FIXME fix for Chrome v73) return navigator.mediaDevices.getUserMedia({audio: true}).then(() => selfParticipant.getMediaStream());
```
ffflorian: Should this be commented?

src/script/util/DebugUtil.js
ffflorian: Maybe import `h` and `init` with other names (like `snabbdomInit`?)

src/script/calling/CallingRepository.ts
atomrc: Oh shit, this is not supposed to be in this PR. Good catch

src/script/util/DebugUtil.js
atomrc: Good idea for `init`. Will keep `h` as is, it's a convention more than anything else, I think it has value to keep it this way

ffflorian: What value would that be? It's hard to figure out what `h` is supposed to mean.

atomrc: It comes from `hyperscript`, it's a, kind of, standard in the vdom dev community
https://github.com/hyperhype/hyperscript

atomrc: The idea, I believe, is to make it short, because you are going to use it (potentially) a lot. Would you prefer a 2 letters word like `el`?

ffflorian: I understand that it makes sense when you use it a lot. But since we use it only once, I'd call it `snabbdomHelper`, since `h()` is [documented](https://github.com/snabbdom/snabbdom#inline-example) as a "helper function".

atomrc: What do you think of `createElement` then. At least it explains what it does? (while `snabbdomHelper` doesn't really describe this)

ffflorian: Sounds good!

PR #7205, opened 2019-08-29
debug: Add access to raw call logs

atomrc:


src/script/util/DebugUtil.js
```diff
@@ -309,8 +309,8 @@ export class DebugUtil {
    * Print call log to console.
```
ffflorian: Not valid anymore
```suggestion
   * Return the whole call log as string
```

src/script/util/DebugUtil.js
```diff
@@ -309,8 +309,8 @@ export class DebugUtil {
...
    * @returns {undefined} No return value
```
ffflorian:
```suggestion
   * @returns {string} The call log
```

PR #7225, opened 2019-09-02
fix: Catch client error on logout

bennyn:


src/script/main/app.js
```diff
@@ -734,7 +734,16 @@ class App {
...
 
-      const keepPermanentDatabase = this.repository.client.isCurrentClientPermanent() && !clearData;
+      let keepPermanentDatabase = !clearData;
+
+      try {
+        keepPermanentDatabase = this.repository.client.isCurrentClientPermanent() && !clearData;
+      } catch (error) {
+        if (error.type === z.error.ClientError.TYPE.CLIENT_NOT_SET) {
+          keepPermanentDatabase = false;
+        }
```
ffflorian, line 744: If error is not of type `CLIENT_NOT_SET`, should we throw it then?

bennycode: No, then we should go with whatever is set to `keepPermanentDatabase` in this stage.

PR #7228, opened 2019-09-02
feat: Include user details in backup file

bennyn:


src/script/backup/BackupRepository.js
```diff
@@ -172,7 +174,7 @@ class BackupRepository {
     const zip = new JSZip();
 
     // first write the metadata file
-    zip.file(BackupRepository.CONFIG.FILENAME.METADATA, JSON.stringify(metaData));
+    zip.file(BackupRepository.CONFIG.FILENAME.METADATA, JSON.stringify(metaData, null, 2));
```
ffflorian, line 177: I am against this. While this will be more readable, it will increase the backup size a lot when you have many conversations.

```suggestion
    zip.file(BackupRepository.CONFIG.FILENAME.METADATA, JSON.stringify(metaData);
```

bennycode: The size won't increase if you have many conversations. I had this in mind when applying my change, that's why I only added the beautification to the meta data file which has just 7 properties.

ffflorian: Whoops right, my bad.

PR #7234, opened 2019-09-03
runfix: Gracefully degrade the EphemeralValueStore when ServiceWorkers are not available

atomrc:


src/script/util/ephemeralValueStore.ts
```diff
@@ -38,6 +38,9 @@ export async function save<T>(value: T): Promise<T> {
 }
 
 async function getWorker(): Promise<ServiceWorker> {
```
ffflorian:
```suggestion
async function getWorker(): Promise<ServiceWorker | undefined> {
```

PR #7235, opened 2019-09-03
refactor: Remove z.components

ffflorian:


src/script/components/asset/controls/seekBar.js
```diff
@@ -118,15 +115,15 @@ z.components.SeekBarComponent = class SeekBarComponent {
...
+}
 
 ko.components.register('seek-bar', {
   template: `
     <input type="range" value="0" max="100" data-bind="css: {'show-seek-bar-thumb': show_seek_bar_thumb, 'element-disabled': disabled}">
   `,
   viewModel: {
     createViewModel(params, component_info) {
-      return new z.components.SeekBarComponent(params, component_info);
+      return new SeekBarComponent(params, component_info);
```
ffflorian, line 126: Maybe it's possible to replace
```js
viewModel: {
  createViewModel(params, component_info) {
    return new SeekBarComponent(params, component_info);
  }
}
```
with
```js
viewModel: SeekBarComponent;
```
?

AndyLnd: Unfortunately it is not that easy, as the `componentInfo` is only passed to the `createViewModel` method

PR #7240, opened 2019-09-04
feat: Allow switching to conversation when clicking the callUI conversation name

atomrc:


src/script/router/routeGenerator.test.ts
```diff
@@ -0,0 +1,32 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {generateConversationUrl} from './routeGenerator';
+
+describe('generateConversationUrl', () => {
+  it('generates an URL that contains the given conversation ID', () => {
+    const conversationIds = ['conversationId'];
```
bennycode, line 24:
```suggestion
    const conversationIds = ['8924e7f8-e04e-4546-aa70-a3496c20be8b'];
```

Let's use a value which looks more "realistic".

PR #7241, opened 2019-09-04
refactor: Migrate "NotificationService" to TypeScript

bennyn:


src/script/event/EventRepository.js
```diff
@@ -442,10 +442,12 @@ export class EventRepository {
 
   _missedEventsFromStream() {
     this.notificationService.getMissedIdFromDb().then(notificationId => {
-      const lastNotificationIdEqualsMissedId = this.lastNotificationId() === notificationId;
-      if (!lastNotificationIdEqualsMissedId) {
-        amplify.publish(WebAppEvents.CONVERSATION.MISSED_EVENTS);
-        this.notificationService.saveMissedIdToDb(this.lastNotificationId());
+      if (notificationId) {
```
ffflorian: Not sure about this, it changes the behavior, doesn't it?

src/script/event/EventRepository.js
```diff
@@ -283,11 +283,11 @@ export class EventRepository {
...
         .getNotifications(this.currentClient().id, notificationId, limit)
         .then(_gotNotifications)
-        .catch(errorResponse => {
+        .catch(async errorResponse => {
           // When asking for notifications with a since set to a notification ID that does not belong to our client ID,
           // we will get a 404 AND notifications
           if (errorResponse.notifications) {
-            this._missedEventsFromStream();
+            await this._missedEventsFromStream();
             return _gotNotifications(errorResponse);
```
ffflorian: Let's not mix `.then()` and `await`.
Use e.g.
```ts
if (errorResponse.notifications) {
  return this._missedEventsFromStream()
    .then(() => _gotNotifications(errorResponse));
}
```

PR #7249, opened 2019-09-06
feat: Find notification ID by message ID

bennyn:


src/script/event/NotificationService.ts
```diff
@@ -169,4 +168,33 @@ export class NotificationService {
       value: notificationId,
     });
   }
+
+  getNotificationIdByMessageId(messageId: string, clientId: string): Promise<string | void> {
+    return new Promise(async resolve => {
```
ffflorian: Please do not mix `Promise` and `async`, rather make the function `async` and use `for ... of` with `return` in line 187 instead.

src/script/event/NotificationService.ts
```diff
@@ -169,4 +168,33 @@ export class NotificationService {
...
+
+      const {notifications} = await this.getNotifications(clientId);
+
+      notifications.forEach(notification => {
+        notification.payload.forEach(event => {
+          if (event.type === CONVERSATION_EVENT.OTR_MESSAGE_ADD) {
+            const eventLooksLikeMessage =
+              event.time === message.time && event.from === message.from && event.conversation === message.conversation;
+            if (eventLooksLikeMessage) {
+              resolve(notification.id);
```
AndyLnd: The loop will keep on looping and there is nothing you can do ...
You might want to use a `.find()` here (in place of both instances of `forEach`) or two `for .. of` loops and a return.

src/script/event/NotificationService.ts
```diff
@@ -169,4 +168,34 @@ export class NotificationService {
...
+      message = await this.storageService.db
+        .table(StorageSchemata.OBJECT_STORE.EVENTS)
+        .where({id: messageId})
+        .first();
+    }
+
+    const {notifications} = await this.getNotifications(clientId);
+
+    for (const notification of notifications) {
+      const matchedEvent = notification.payload.find(event => {
```
ffflorian, line 187: I think this won't work, because every loop iteration will overwrite `matchedEvent`.

ffflorian: Whoops, that's still inside the loop, so all good!

PR #7264, opened 2019-09-11
fix: Improve redirect after reloading a temporary client (WEBAPP-6419)

atomrc:


src/script/main/app.js
```diff
@@ -595,10 +596,8 @@ class App {
    * @returns {boolean} `true` if it is a page refresh
    */
   _isReload() {
-    const isReload = isSameLocation(document.referrer, window.location.href);
-    const log = `App reload: '${isReload}', Referrer: '${document.referrer}', Location: '${window.location.href}'`;
-    this.logger.debug(log);
-    return isReload;
+    const NAVIGATION_TYPE_RELOAD = 1;
+    return window.performance.navigation.type === NAVIGATION_TYPE_RELOAD;
```
atomrc, line 600: This method is way safer to check for reload (in place of string comparison) and compatibility-wise, we are good.

ffflorian: Good find!

PR #7265, opened 2019-09-11
fix: Safer localStorage access from logger (WEBAPP-6423)

atomrc:
This could be due to a Firefox regression where, when cookies are not enabled, firefox does not throw an error but instead gives back a null object

### DIFF
--- a/src/script/util/LoggerUtil.js
+++ b/src/script/util/LoggerUtil.js
@@ -20,14 +20,15 @@
 import 'url-search-params-polyfill';
 
 function enableLogging(force = false, search = window.location.search) {
-  /**
-   * If users disable cookies in their browsers, they won't have access to the localStorage API.
-   * The following check will fix this error:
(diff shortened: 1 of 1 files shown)


src/script/util/LoggerUtil.js
```diff
@@ -29,7 +29,8 @@ function enableLogging(force = false, search = window.location.search) {
 
   try {
     localStorage = window.localStorage;
-  } catch (error) {
+  } catch (error) {}
+  if (!localStorage) {
```
bennycode, line 34: Let's add your comments ("This could be due to a Firefox regression...") as a code comment to not forget why we check for `!localStorage`.

atomrc: I moved the comment a few lines up here and added the note. Does it make sense?

bennycode: Yes, thanks for adding the details.

PR #7267, opened 2019-09-11
fix: Display conversation not found modal at the right time

ffflorian:
Right now, every error in `showConversation` causes the "conversation not found" modal to appear and is being swallowed then.
To be able to find and fix errors within this function, we will only swallow the `CONVERSATION_NOT_FOUND` error and surface the others. Since the promise inside the function is floating, it should not break the flow of other functions.

### DIFF
--- a/src/script/conversation/ConversationRepository.js
+++ b/src/script/conversation/ConversationRepository.js
@@ -1025,9 +1025,8 @@ export class ConversationRepository {
    * @returns {boolean} Is the conversation active
    */
   is_active_conversation(conversationEntity) {
-    if (this.active_conversation()) {
-      return this.active_conversation().id === conversationEntity.id;
-    }
+    const activeConversation = this.active_conversation();
+    return !!activeConversation && activeConversation.id === conversationEntity.id;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationRepository.js
```diff
@@ -1025,9 +1025,8 @@ export class ConversationRepository {
    * @returns {boolean} Is the conversation active
    */
   is_active_conversation(conversationEntity) {
-    if (this.active_conversation()) {
-      return this.active_conversation().id === conversationEntity.id;
-    }
```
ffflorian, line 1030: This was actually returning `boolean | undefined` and not just `boolean`.

src/script/view_model/ContentViewModel.js
```diff
@@ -211,7 +211,10 @@ export class ContentViewModel {
 
     conversationPromise
       .then(conversationEntity => {
-        const isActiveConversation = conversationEntity === this.conversationRepository.active_conversation();
+        if (!conversationEntity) {
+          throw new z.error.ConversationError(z.error.ConversationError.TYPE.CONVERSATION_NOT_FOUND);
+        }
+        const isActiveConversation = this.conversationRepository.is_active_conversation(conversationEntity);
```
ffflorian, line 217: Why not use `is_active_conversation` if we have it?

PR #7271, opened 2019-09-11
fix(server): Remove config whitespace

ffflorian:


server/routes/config/ConfigRoute.ts
```diff
@@ -26,8 +26,7 @@ export const ConfigRoute = (config: ServerConfig) =>
       APP_BASE: config.SERVER.APP_BASE,
     };
 
-    res.type('application/javascript').send(`
-      window.wire = window.wire || {};
-      window.wire.env = ${JSON.stringify(clientConfig)};
-    `);
+    res
```
atomrc: How about?

```
const payload = `window.wire = window.wire || {};
window.wire.env = ${JSON.stringify(clientConfig)};`

res.send(payload);
```

I feel it's slightly easier to read

atomrc: The `\n` makes things slightly hard to read

ffflorian: It would then look like this:
```ts
    const payload = `window.wire = window.wire || {};
window.wire.env = ${JSON.stringify(clientConfig)};`
```

ffflorian: But yeah, the `\n` is annoying...

PR #7281, opened 2019-09-12
test: Unit tests for the conversation calling cell

atomrc:


test/unit_tests/components/list/conversationListCallingCellSpec.js
```diff
@@ -0,0 +1,102 @@
...
+    });
+  });
+});
+
+function createCall(state) {
+  const call = new Call();
+  call.state(state);
+  call.selfParticipant = new Participant();
+  return call;
+}
```
ffflorian: Let's move this before the `describe` block so it's easier to find.

PR #7282, opened 2019-09-13
refactor: Convert IntegrationRepository to TypeScript

ffflorian:


src/script/integration/IntegrationRepository.ts
```diff
@@ -177,31 +201,33 @@ export class IntegrationRepository {
...
         return IntegrationMapper.mapServiceFromObject(serviceData);
       }
+      return undefined;
     });
   }
 
-  getServices(tags, start) {
-    const tagsArray = _.isArray(tags) ? tags.slice(0, 3) : [ServiceTag.INTEGRATION];
+  getServices(tags: ServiceTag | ServiceTag[], start: string): Promise<ServiceEntity[]> {
+    const tagsArray = isArray(tags) ? tags.slice(0, 3) : [ServiceTag.INTEGRATION];
```
bennycode: Time to replace `underscore` with ES6 standards: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

src/script/integration/IntegrationRepository.ts
```diff
@@ -17,32 +17,52 @@
...
    * Trim query string for search.
-   * @param {string} query - Service search string
-   * @returns {string} Normalized service search query
+   * @param query Service search string
+   * @returns Normalized service search query
    */
-  static normalizeQuery(query) {
-    if (!_.isString(query)) {
+  static normalizeQuery(query: string): string {
+    if (!isString(query)) {
```
bennycode: Let's go with a simple `typeof query === 'string'` check.

src/script/integration/IntegrationRepository.ts
```diff
@@ -56,41 +76,45 @@ export class IntegrationRepository {
...
+   * @param entity Service or user to resolve to ServiceEntity
+   * @returns Resolves with the ServiceEntity
    */
-  getServiceFromUser(entity) {
+  getServiceFromUser(entity: ServiceEntity | User): Promise<ServiceEntity> {
     if (entity instanceof ServiceEntity) {
       return Promise.resolve(entity);
     }
-    const {providerId, serviceId} = entity;
+    const {providerId, serviceId} = entity as any;
```
bennycode: What else is inside of `entity`? Can we type it to be `User` or `ServiceEntity`?

src/script/integration/IntegrationRepository.ts
```diff
@@ -210,12 +236,12 @@ export class IntegrationRepository {
...
-   * @param {User} userEntity - Service user to be removed from the conversation
-   * @returns {Promise} Resolves when service was removed from the conversation
+   * @param conversationEntity Conversation to remove service from
+   * @param userEntity Service user to be removed from the conversation
+   * @returns Resolves when service was removed from the conversation
    */
-  removeService(conversationEntity, userEntity) {
-    const {id: userId, serviceId} = userEntity;
+  removeService(conversationEntity: Conversation, userEntity: User): Promise<any> {
+    const {id: userId, serviceId} = userEntity as any;
```
bennycode: Let's connect our brains to find a solution on how to dissolve `any`.

ffflorian: `User` is not typed yet so it doesn't make sense to create an interface here.

src/script/integration/IntegrationRepository.ts
ffflorian: `entity` is of type `User` here, but since `User` is not typed yet so it doesn't make sense to create an interface here.

PR #7305, opened 2019-09-16
runfix: Return fetched notifications

bennyn:


src/script/event/NotificationService.ts
```diff
@@ -68,6 +68,26 @@ export class NotificationService {
...
+    return notificationList.time;
+  }
+
+  getAllNotificationsForClient(clientId: string, notificationId?: string): Promise<Notification[]> {
+    const notifications: Notification[] = [];
+
+    const collectNotifications = async (lastNotificationId?: string): Promise<Notification[]> => {
+      const notificationList = await this.getNotifications(clientId, lastNotificationId);
+      const newNotifications = notificationList.notifications;
+      lastNotificationId = newNotifications[newNotifications.length - 1].id;
```
AndyLnd: Could it happen that newNotifications is an empty array?

PR #7309, opened 2019-09-18
Revert "refactor: Turn util.js into TS"

AndyLnd:
Reverts wireapp/wire-webapp#7308


bennycode:
We will squash the commits later.

PR #7476, opened 2019-10-17
refactor: Functional SSO components

Yserz:


package.json
```diff
@@ -31,7 +31,7 @@
     "raygun4js": "2.18.0",
     "react": "16.10.2",
     "react-dom": "16.10.2",
-    "react-intl": "2.9.0",
+    "react-intl": "3.3.2",
```
bennycode, line 34: When updating `react-intl` we should also update `@types/react-intl`.

src/script/auth/page/SingleSignOn.tsx
```diff
@@ -32,31 +32,32 @@ import {
...
 import Page from './Page';
 import SingleSignOnForm from './SingleSignOnForm';
 
-interface Props extends React.HTMLAttributes<SingleSignOn>, RouteComponentProps<{}> {}
+interface Props extends React.HTMLAttributes<HTMLDivElement> {}
 
-interface State {
-  isOverlayOpen: boolean;
-}
+let ssoWindow: Window = undefined;
```
AndyLnd: useRef

src/script/auth/page/SingleSignOn.tsx
```diff
@@ -176,82 +177,93 @@ class SingleSignOn extends React.PureComponent<Props & InjectedIntlProps, State>
     const top = parentHeight / 2 - childHeight / 2 + screenTop;
     return {left, top};
   };
+  const focusChildWindow = () => ssoWindow && ssoWindow.focus();
```
AndyLnd: focusSsoWindow

package.json
Yserz: We can't do that until we moved all components to functional components. Reason for that is that react-intl moved the types to the package itself and the 3.x types are not working anymore.

Yserz: Also the temporary solution for using the hook won't work.

PR #7478, opened 2019-10-17
refactor: Migrate "PropertiesRepository" to TypeScript

bennyn:


src/script/properties/PropertiesRepository.ts
```diff
@@ -292,15 +320,7 @@ class PropertiesRepository {
...
-  /**
-   * Set the preference of specified type
-   *
-   * @private
-   * @param {PROPERTIES_TYPE} propertiesType - Type of preference to set
-   * @param {*} changedPreference - New preference to set
-   * @returns {undefined} No return value
-   */
-  _setPreference(propertiesType, changedPreference) {
+  _setPreference(propertiesType: string, changedPreference: {}): void {
```
ffflorian:
```suggestion
  private _setPreference(propertiesType: string, changedPreference: {}): void {
```

bennycode: Good reminder. I will also remove the leading underscore then.

PR #7506, opened 2019-10-21
runfix: Throw validation error

AndyLnd:


src/script/auth/page/SingleSignOnForm.tsx
```diff
@@ -108,18 +108,16 @@ const SingleSignOnForm = ({
       return;
     }
     codeInput.current.value = codeInput.current.value.trim();
+    const currentValidationError = codeInput.current.checkValidity()
```
bennycode, line 111: How about destructuring `codeInput.current`? Might be worth it since it is used quite a lot here.

src/script/auth/page/SingleSignOnForm.tsx
```diff
@@ -108,18 +108,16 @@ const SingleSignOnForm = ({
...
-        : ValidationError.handleValidationState(codeInput.current.name, codeInput.current.validity),
-    );
-
+    setValidationError(currentValidationError);
     setIsCodeInputValid(codeInput.current.validity.valid);
 
     try {
-      if (validationError) {
-        throw validationError;
+      if (currentValidationError) {
```
bennycode, line 119: This does not need to be inside of `try`, does it?

PR #7511, opened 2019-10-21
feat: Save encryption key for temporary clients

bennyn:


bennycode:
Closed because of https://github.com/wireapp/wire-webapp/pull/7620

PR #7543, opened 2019-10-23
refactor: Migrate CopyToClipboard to TypeScript

lipis:

### DIFF
--- a/src/script/components/copyToClipboard.ts
+++ b/src/script/components/copyToClipboard.ts
@@ -18,14 +18,17 @@
  */
 
 class CopyToClipboard {
-  constructor(params) {
+  // tslint:disable-next-line
+  private readonly text: string;
+
+  constructor(params: {text: string}) {
     this.text = params.text;
   }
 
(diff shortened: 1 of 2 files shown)


src/script/components/copyToClipboard.ts
```diff
@@ -18,14 +18,17 @@
  */
 
 class CopyToClipboard {
-  constructor(params) {
+  // tslint:disable-next-line
```
ffflorian: Why disable TSLint here?

lipis: `'text' is declared but its value is never read.`

It's used in the template below, but TypeScript can't see that.. if you have anything better, we are open :)

lipis: Travis is still complaining about it..

ffflorian: Make it public :)

lipis: It was that easy! Thanks

PR #7554, opened 2019-10-25
fix: Add error message to device deletion (WEBAPP-6484)

AndyLnd:


src/script/view_model/ActionsViewModel.js
```diff
@@ -112,23 +114,33 @@ export class ActionsViewModel {
...
+          action: async password => {
+            if (!isSending) {
+              isSending = true;
+              try {
+                await this.clientRepository.deleteClient(clientEntity.id, password);
+                this.modalsViewModel.hide();
+                resolve();
+              } catch (error) {
+                this.modalsViewModel.errorMessage(expectedErrors[error.label] || error.message);
+                isSending = false;
```
Yserz, line 142: finally {isSending = false}

src/page/template/modals.htm
```diff
@@ -20,15 +20,18 @@
...
         </form>
       <!-- /ko -->
       <!-- ko if: hasInput() -->
-        <form data-bind="submit: () => doAction(confirm)">
+        <form data-bind="submit: () => doAction(confirm, content().closeOnConfirm)">
           <input maxlength="64" class="modal__input" data-bind="textInput: inputValue, attr: {placeholder: content().inputPlaceholder}, hasFocus: inputFocus"/>
         </form>
       <!-- /ko -->
+      <!-- ko if: errorMessage() -->
+      <div class="modal__input__error" data-bind="text: errorMessage"></div>
```
ffflorian:
```suggestion
        <div class="modal__input__error" data-bind="text: errorMessage"></div>
```

src/script/view_model/ActionsViewModel.js
AndyLnd: we really don't care about isSending anymore once this is successful, as we then leave the scope. if the method errors, however, you should be able to send the request again.

ffflorian: But @Yserz is right, it also makes sense when reading it.

Yserz: It's true, technically this doesn't change anything. But for a reader it's more clear what should happen.

ffflorian, line 142:
```suggestion
              } finally {
                isSending = false;
```

PR #7610, opened 2019-10-30
refactor: Sequential login flow

Yserz:
login-flow2

Changes:
When logging in you will always pass `history screen`, `choose handle` and `initial invites` but all pages will continue if certain conditions are matched.

- `history screen` will be skipped if the current client is no new client or current client is the first (...)
- `choose handle` will be skipped if the user already has a handle
- `initial invites` will skip if it's not the initial creation of a team


Yserz:
Can be merged after tomorrows staging bump

src/script/auth/page/HistoryInfo.tsx
```diff
@@ -27,24 +28,35 @@ import {Config} from '../config';
...
+   *   1. database contains at least one event
+   *   2. there is at least one previously registered client
+   *   3. new local client is temporary
+   */
+  const shouldShowHistoryInfo =
+    isNewCurrentSelfClient &&
+    (hasHistory || clients.length > 1 || (currentSelfClient && currentSelfClient.type === ClientType.TEMPORARY));
+
+  if (!shouldShowHistoryInfo) {
+    history.push(ROUTE.CHOOSE_HANDLE);
```
AndyLnd, line 57:
```suggestion
    history.push(ROUTE.CHOOSE_HANDLE);
    return null;
```

PR #7621, opened 2019-11-04
refactor: Functional ConversationJoin component

Yserz:


src/script/auth/page/ConversationJoin.tsx
```diff
@@ -59,124 +59,91 @@ import {parseError, parseValidationErrors} from '../util/errorUtil';
...
-  };
-
-  readAndUpdateParamsFromUrl = (nextProps: CombinedProps) => {
-    if (this.isPwaEnabled()) {
-      this.setState({forceNewTemporaryGuestAccount: true});
-    }
-    const conversationCode = getURLParameter(QUERY_KEY.CONVERSATION_CODE);
-    const conversationKey = getURLParameter(QUERY_KEY.CONVERSATION_KEY);
-    const expiresIn = parseInt(getURLParameter(QUERY_KEY.JOIN_EXPIRES), 10) || undefined;
+  useEffect(() => {
```
AndyLnd, line 91: This might cause issues, as the original lifecycle event was `componentWillReceiveProps`. This useEffect will now only trigger once in the beginning as it has the second parameter set to `[]`;

PR #7645, opened 2019-11-07
refactor: Async/await in UserAction

Yserz:


src/script/auth/module/action/UserAction.ts
```diff
@@ -23,23 +23,22 @@ import {UserActionCreator} from './creator/';
 
 export class UserAction {
   checkHandles = (handles: string[]): ThunkAction<Promise<string>> => {
-    return (dispatch, getState, {apiClient}) => {
-      return apiClient.user.api.postHandles({handles, return: 1}).then(result => result[0]);
+    return async (dispatch, getState, {apiClient}) => {
+      const availableHandles = await apiClient.user.api.postHandles({handles, return: 1});
+      return availableHandles[0];
```
ffflorian: Here you could also destructure:
```ts
      const [availableHandle] = await apiClient.user.api.postHandles({handles, return: 1});
      return availableHandle;
```

PR #7650, opened 2019-11-07
runfix: Enforce resolve of MediaStream promise (WEBAPP-6523)

Yserz:


Yserz:
We might not need it. I think this bug fixed itself :P

PR #7710, opened 2019-11-12
runfix: Correctly support legacy boolean mutedState (WEBAPP-6537)

AndyLnd:


src/script/components/list/groupedConversationHeader.ts
```diff
@@ -42,10 +42,14 @@ ko.components.register('grouped-conversation-header', {
     this.badge = ko.pureComputed(
       () =>
         conversationLabel.conversations().filter(conversation => {
-          if (conversation.mutedState() === NOTIFICATION_STATE.NOTHING) {
+          const mutedState = conversation.mutedState();
+          if (typeof mutedState === 'boolean') {
+            return mutedState ? false : conversation.hasUnread();
```
ffflorian:
```suggestion
            return mutedState === true ? false : conversation.hasUnread();
```

PR #7712, opened 2019-11-13
fix: Speed up MD5 hashing by using js-md5 (WEBAPP-6538)

ffflorian:
This dramatically speeds up MD5 hashing by using [`js-md5`](https://www.npmjs.com/package/js-md5) instead of [`crypto-js`](https://www.npmjs.com/package/crypto-js) (and saves some disk space since `js-md5` is 30 kilobytes smaller :smiley:)

---

### Hashing a 100 MB file with `CryptoJS` (~7 seconds)

before

---

### Hashing a 100 MB file with `js-md5` (~0.5 seconds)
after

---

Tested with Chrome 78 on an Intel(R) Core(TM) i7-7500U @ 2.70GHz.


src/script/util/util.ts
```diff
@@ -259,9 +259,6 @@ export const phoneNumberToE164 = (phoneNumber: string, countryCode: string): str
 
 export const createRandomUuid = (): string => UUID.genV4().hexString;
 
-export const encodeSha256Base64 = (text: string | CryptoJS.LibWordArray): string =>
```
ffflorian, line 262: This was not used at all.

Issue #549, opened 2016-12-09
Unable to scroll history in conversations without using a mouse (i.e. with PageUp/PageDown)


max-baz:
Open an Otto the Bot chat.
Exchange a few messages so that the history does not fit one screen and requires scrolling.
Try to scroll the history without using a mouse (natural thought: PageUp / PageDown).

Nothing happens, the page does not scroll.

----

I had a quick look at the code, it doesn't seem to be a problem with javascript - the keydown events are bubbled and not cancelled. I believe the problem is with css, thought I don't have a proof. I found elements with classes that are called `antiscroll-inner` and `antiscroll-wrap`, but simple removal of these classes does not make scrolling work.

bennycode:
We are supporting navigating with Page Up / Page Down keys now in our webapp.

PR #7714, opened 2019-11-13
Scroll conversation view without using a mouse

maximbaz:
This is an attempt to help with #549, although I couldn't make it work 100%, so feedback and ideas are welcome 🙂 

**What works:** Click somewhere in the middle of the conversation view. Pressing PageUp and PageDown will scroll through the history. Press "hello" and it will be entered in the text area. Press PageUp / PageDown again and it will scroll conversation history again.

**What doesn't work:** Click in the text area. Pressing PageUp / PageDown will **not** scroll conversation history 🙁 I was trying different things, focusing on `#conversation` div or `body`, adding `tabindex`, nothing helps 😞 

In my mind, even this state is already a significant improvement, annoying, but better than nothing.

Let me know your thoughts.

P.S. I noticed a small leak, in `MessageListBindings.js` function `focus_on_keydown` gets executed multiple times if you switch between contacts in the sidebar, so if you switch back and forth 20 times, that callback will be called 20 times now. I'm not sure how to fix it (and it doesn't belong to this PR maybe), but maybe you will know how to do it 🙂


lipis:
@maximbaz Something went completely wrong with my attempt to merge conflicts..

PR #7740, opened 2019-11-18
fix: Speed up base64 to array conversion

ffflorian:
This PR will
- use `libsodium` to convert arrays to base64 and vice-versa
- make the conversion functions asynchronous so they don't block the whole I/O
  - thus refactoring some normal functions to `async`
- refactor some promise-returning functions to `async` to make them more readable
- rename the synchronous version of `base64ToArray()` to `base64ToArraySync()` since we need it in the `StorageSchemata`

Best viewed [without whitespace changes](https://github.com/wireapp/wire-webapp/pull/7740/files?w=1).

### DIFF
--- a/src/script/assets/AssetCrypto.ts
+++ b/src/script/assets/AssetCrypto.ts
@@ -17,6 +17,12 @@
  *
  */
 
+export interface EncryptedAsset {
+  cipherText: ArrayBuffer;
+  keyBytes: ArrayBuffer;
+  sha256: ArrayBuffer;
+}
+
 /**
  * @param cipherText - Encrypted plaintext
  * @param keyBytes - AES key used for encryption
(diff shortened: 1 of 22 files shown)


src/script/assets/AssetCrypto.ts
```diff
@@ -17,6 +17,12 @@
  *
  */
 
+export interface EncryptedAsset {
```
bennycode, line 20: Can we use `EncryptedAsset` from [here](https://github.com/wireapp/wire-web-packages/blob/16579f5a41c36769d8e4cd1a502dfae32dbd31b2/packages/core/src/main/cryptography/EncryptedAsset.ts)?

src/script/assets/AssetCrypto.ts
```diff
@@ -28,51 +34,32 @@ export const decryptAesAsset = async (
...
-
-      throw new Error('Encrypted asset does not match its SHA-256 hash');
-    })
-    .then(key => {
-      const iv = cipherText.slice(0, 16);
-      const assetCipherText = cipherText.slice(16);
-      return window.crypto.subtle.decrypt({iv: iv, name: 'AES-CBC'}, key, assetCipherText);
-    });
+  const computedSha256 = await window.crypto.subtle.digest('SHA-256', cipherText);
+  if (!_equalHashes(referenceSha256, computedSha256)) {
```
bennycode: Let's make `_equalHashes` a private function without underscore.

src/script/assets/AssetCrypto.ts
```diff
@@ -28,51 +34,32 @@ export const decryptAesAsset = async (
...
+    throw new Error('Encrypted asset does not match its SHA-256 hash');
+  }
+  const key = await window.crypto.subtle.importKey('raw', keyBytes, 'AES-CBC', false, ['decrypt']);
+  const iv = cipherText.slice(0, 16);
```
bennycode: Is there a more descriptive name for `iv`?

src/script/assets/AssetCrypto.ts
ffflorian: Unfortunately not, since we use `ArrayBuffer` instead of `Buffer`.

src/script/assets/AssetCrypto.ts
ffflorian: Yes, actually it's an initialization vector.

PR #7744, opened 2019-11-19
chore(deps): Revert "bump geolite2 from 1.3.0 to 2.0.0 in /server"

Yserz:
Reverts wireapp/wire-webapp#7719


bennycode:
Geolite2 update is blocked by: https://github.com/runk/node-geolite2/issues/13

PR #7755, opened 2019-11-20
runfix: Show partners on exact match (WEBAPP-6547)

AndyLnd:


src/script/components/userList.ts
```diff
@@ -41,6 +41,7 @@ interface UserListParams {
   filter: ko.Observable<string>;
   selected: ko.ObservableArray<User>;
   searchRepository: SearchRepository;
+  skipSearch: boolean;
```
ffflorian, line 44: This should be optional

PR #7797, opened 2019-11-26
refactor: Use singleton BackendClient and remove dependency graph

lipis:


bennycode:
Will be tackled with https://github.com/wireapp/wire-webapp/pull/8009

PR #7852, opened 2019-12-06
runfix: Throw error if client type is unspecified

bennyn:
Although `isTemporaryClientAndNonPersistent` is written in TypeScript there might be some JavaScript code calling that method with an `undefined` value, so I am throwing an error now in this particular case.

### DIFF
--- a/src/script/main/app.js
+++ b/src/script/main/app.js
@@ -115,6 +115,22 @@ import {GiphyService} from '../extension/GiphyService';
 import {PermissionRepository} from '../permission/PermissionRepository';
 import {loadValue} from 'Util/StorageUtil';
 
+function doRedirect(signOutReason) {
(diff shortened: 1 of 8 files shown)


src/script/main/app.js
```diff
@@ -848,7 +861,7 @@ class App {
 
   /**
    * Redirect to the login page after internet connectivity has been verified.
-   * @param {SIGN_OUT_REASON} signOutReason - Redirect triggered by session expiration
+   * @param {SIGN_OUT_REASON} signOutReason - Redirect triggered by session expiration_redirectToLogin
```
ffflorian:
```suggestion
   * @param {SIGN_OUT_REASON} signOutReason - Redirect triggered by session expiration
```

PR #7875, opened 2019-12-09
fix: Disable click when image is still loading (WEBAPP-6564)

ffflorian:


src/script/view_model/bindings/MessageListBindings.js
```diff
@@ -141,7 +141,6 @@ ko.bindingHandlers.background_image = {
       assetLoader
         .load()
         .then(blob => {
-          $(element).removeClass('image-loading');
```
ffflorian, line 144: This class was not in use at all.

PR #7881, opened 2019-12-09
runfix: Do not always redirect to landing page

bennyn:


src/script/main/app.js
```diff
@@ -875,7 +859,19 @@ class App {
         return window.location.replace(url);
       }
 
-      doRedirect(signOutReason);
+      let url = `/auth/${location.search}`;
```
Yserz, line 862: I would prefer to keep the named function because it's adding information what this big block of code does.

PR #7892, opened 2019-12-10
chore(deps): Revert "bump geolite2 from 1.3.0 to 2.0.2 in /server"

Yserz:
Reverts wireapp/wire-webapp#7818


ffflorian:
Reverted because of https://github.com/runk/node-geolite2/issues/13.

PR #7895, opened 2019-12-11
runfix: Don't throw error on logout with permanent device

bennyn:

### DIFF
--- a/src/script/main/app.js
+++ b/src/script/main/app.js
@@ -716,7 +716,8 @@ class App {
       this.repository.calling.destroy();
 
       if (this.repository.user.isActivatedAccount()) {
-        if (isTemporaryClientAndNonPersistent(loadValue(StorageKey.AUTH.PERSIST))) {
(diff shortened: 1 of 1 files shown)


src/script/main/app.js
```diff
@@ -716,7 +716,8 @@ class App {
       this.repository.calling.destroy();
 
       if (this.repository.user.isActivatedAccount()) {
-        if (isTemporaryClientAndNonPersistent(loadValue(StorageKey.AUTH.PERSIST))) {
+        const persist = loadValue(StorageKey.AUTH.PERSIST);
+        if (persist && isTemporaryClientAndNonPersistent(persist)) {
```
ffflorian: That somehow doesn't make sense ... `persist && isTemporaryClientAndNonPersistent` 🤔
Maybe we can rename `persist` again?

bennycode: Any suggestion? `persist` can be undefined that's why I have that `persist &&` check.

ffflorian: What about `shouldPersist`?

PR #7897, opened 2019-12-12
chore: Add tag release script

ffflorian:


bin/release_tag.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+  default: {
+    logger.error(`No or invalid target specified. Valid targets are: staging, production`);
+    logger.info(usageText);
+    process.exit(1);
+  }
+}
+
+if (!commitId) {
+  logger.error(`No commit ID specified`);
```
Yserz, line 70: should be `logger.error('No commit ID specified');` (backticks)

bin/release_tag.js
```diff
@@ -0,0 +1,105 @@
...
+    logger.info(usageText);
+    process.exit();
+  }
+  case 'production':
+  case 'staging': {
+    target = firstArgument;
+    break;
+  }
+  default: {
+    logger.error(`No or invalid target specified. Valid targets are: staging, production`);
```
Yserz, line 63: backticks

PR #7923, opened 2019-12-16
refactor: Use static readonly members instead of getter

ffflorian:
Best viewed [without whitespace changes](https://github.com/wireapp/wire-webapp/pull/7923/files?w=1).


ffflorian:
Since we are relying on the getters I will close this for now. We should convert more files to TS first.

PR #7930, opened 2019-12-17
refactor: Lazy loaded routes

Yserz:
Improves initial page load

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "forceConsistentCasingInFileNames": true,
     "jsx": "react",
     "lib": ["dom", "es2017"],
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "forceConsistentCasingInFileNames": true,
     "jsx": "react",
     "lib": ["dom", "es2017"],
+    "module": "esnext",
```
bennycode, line 12: Is this required or will it also work with other module formats, i.e. `commonjs`?

Yserz: It seems `commonjs` and `esnext` works

PR #8010, opened 2020-01-10
refactor: Convert release script to TypeScript

bennyn:


bin/release_tag.ts
```diff
@@ -92,31 +92,19 @@ logger.info(`Fetching base "${origin}" ...`);
 exec(`git fetch ${origin}`);
 
 /**
- * @param {number} index - The tag name index
- * @returns {string} The new tag name
+ * @param index The tag name index
+ * @returns The new tag name
```
ffflorian: You can completely remove these comments as they were only necessary to have valid JSDoc. :slightly_smiling_face:

PR #8040, opened 2020-01-13
Rephrase error message about local browser storage not being available

lucendio:
## Type of change

error message phrasing

## Reason

I just ran into the error message after deploying the wire-server
in demo mode for learning purposes. I found the phrasing a bit
misleading, since it could have referred to a broken server
installation instead. I admit, it't a very edgy case, but
I guess a bit more explicit phasing doesnt do any harm.

## NOTE

in case of acceptance, there is still some work to do,
because the i18n is missing

### DIFF
--- a/src/script/strings.ts
+++ b/src/script/strings.ts
@@ -293,7 +293,7 @@ export const unsupportedStrings = defineMessages({
     id: 'unsupported.headlineCookies',
   },
   headlineIndexedDb: {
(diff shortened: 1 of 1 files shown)


bennycode:
Thanks for your contribution. We will forward this change to our translation team. 👍

lucendio:
@bennyn green light from translation team, so feel free to merge it.

PR #8044, opened 2020-01-15
refactor: AudioSeekBar to TS

AndyLnd:


src/script/components/asset/controls/audioSeekBar.ts
```diff
@@ -16,41 +16,46 @@
...
 
 import {interpolate} from 'Util/ArrayUtil';
 import {clamp} from 'Util/NumberUtil';
 
+interface Params {
+  src: HTMLAudioElement;
+  disabled: ko.Subscribable<boolean>;
+
+  // TODO: replace with proper Type once they are defined
+  asset: {meta: {loudness: number[]}};
```
ffflorian: If this is not properly typed, let's use `any` for the time being.
```suggestion
  asset: any;
```

src/script/components/asset/controls/audioSeekBar.ts
```diff
@@ -16,41 +16,46 @@
...
           this.element.classList.toggle('element-disabled', params.disabled());
         }
       },
-      {disposeWhenNodeIsRemoved: componentInfo.element},
+      {disposeWhenNodeIsRemoved: element},
     );
 
     const assetMeta = params.asset.meta;
-    if (assetMeta !== null && assetMeta.loudness !== null) {
+    if (assetMeta?.loudness !== null) {
```
ffflorian, line 58: Before we checked only for `null` and not for `undefined`, does this have any implications?

AndyLnd: It does not. It will still not work for null and undefined would have errored before.

PR #8050, opened 2020-01-17
chore: Send GIFs when releasing

ffflorian:
gif


bin/release_tag.ts
```diff
@@ -127,27 +170,31 @@ const announceRelease = async (tagName: string, commitId: string): Promise<void>
     const message = `Released tag "${tagName}" based on commit ID "${commitId}".`;
     const payload = account.service.conversation.messageBuilder.createText(WIRE_CONVERSATION, message).build();
     await account.service.conversation.send(payload);
+    await sendRandomGif(account, WIRE_CONVERSATION, 'in the oven');
```
bennycode, line 172: Let's send this one before sending `Released tag "${tagName}" based on commit ID "${commitId}"`. Otherwise you might miss the important part as it's not the latest message.

PR #8092, opened 2020-01-21
refactor: Convert audioAsset to TS

AndyLnd:


src/script/components/asset/audioAsset.ts
```diff
@@ -17,85 +17,91 @@
...
 
-import {AbstractAssetTransferStateTracker} from './AbstractAssetTransferStateTracker';
 import {AssetTransferState} from '../../assets/AssetTransferState';
+import {ContentMessage} from '../../entity/message/ContentMessage';
+import {File as FileAsset} from '../../entity/message/File';
+import {AbstractAssetTransferStateTracker} from './AbstractAssetTransferStateTracker';
+
+interface Params {
+  message: ContentMessage;
+  header: boolean;
```
ffflorian, line 34: Let's add a comment explaining what this does

src/script/entity/message/File.ts
```diff
@@ -31,7 +31,10 @@ import {AssetTransferState} from '../../assets/AssetTransferState';
 import {AssetType} from '../../assets/AssetType';
 import {AssetUploadFailedReason} from '../../assets/AssetUploadFailedReason';
 
-type AssetMetaData = ProtobufAsset.IAudioMetaData | ProtobufAsset.IImageMetaData | ProtobufAsset.IVideoMetaData;
+type AssetMetaData = (ProtobufAsset.IAudioMetaData | ProtobufAsset.IImageMetaData | ProtobufAsset.IVideoMetaData) & {
+  loudness?: number[];
+  duration?: number;
+};
```
ffflorian, line 37: Why do you need this? We are relying on Protobuf types here, and `loudness` or `duration` don't exist as keys in the protobuf spec.

AndyLnd: These are added later by us in the cryptographyMapper. This isn't ideal but we have to work with that for now.

PR #8194, opened 2020-02-07
docs: Add translation process

bennyn:


bennycode:
@lipis does Excalidraw support magentic arrows which stick to objects? Otherwise it's not an option for us.

lipis:
No sticky arrows yet.. :)

lipis:
But the process is not changing that often I would say :)

PR #8243, opened 2020-02-19
feat: Close modals on ESC key

AndyLnd:


src/script/view_model/ModalsViewModel.js
```diff
@@ -220,7 +221,24 @@ export class ModalsViewModel {
     }
     this.content(content);
     this.state(States.OPEN);
+    if (!preventClose) {
+      onEscKey(this.hide);
+    }
+    window.addEventListener('keydown', this.handleEnterKey);
     afterRender(() => this.inputFocus(true));
+    document.getElementById(this.elementId).focus();
```
lipis: My attempt to send the focus to the modal.. seems to be working but by pressing Tab it goes only on the first button (or checkbox, then button) but not to the second one. Strange 🙅‍♂ :)

PR #8250, opened 2020-02-21
Add JoyPixels to the list of local emoji fonts

maximbaz:
EmojiOne was rebranded to JoyPixels almost a year ago.

To be honest, I would have just removed the `local('Emoji One')`, but I wasn't sure if you prefer to keep it for some legacy reasons?


bennycode:
Hey Maxim, thanks for the update. I removed `local('Emoji One')` in your PR. Is it OK like this?

max-baz:
It's perfect 👍

PR #8313, opened 2020-03-03
feat: Add GitHub actions for posting changelog

lipis:


bennycode:
Great investigation but we will keep going with Travis CI because of pricing.

PR #8423, opened 2020-03-25
refactor: Simplify Cryptobox initialization

bennyn:


src/script/cryptography/CryptographyRepository.js
```diff
@@ -131,6 +77,11 @@ export class CryptographyRepository {
       const {userId, clientId} = ClientEntity.dismantleUserClientId(sessionId);
       amplify.publish(WebAppEvents.CLIENT.ADD, userId, {id: clientId}, true);
     });
+
+    if (useExistingCryptobox === true) {
```
lipis:
```suggestion
    if (useExistingCryptobox) {
```

src/script/main/app.js
```diff
@@ -388,7 +388,7 @@ class App {
       telemetry.time_step(AppInitTimingsStep.VALIDATED_CLIENT);
       telemetry.add_statistic(AppInitStatisticsValue.CLIENT_TYPE, clientEntity.type);
 
-      await cryptographyRepository.loadCryptobox(this.service.storage.db);
+      await cryptographyRepository.initCryptobox(true);
```
lipis: default value is `true`
```suggestion
      await cryptographyRepository.initCryptobox();
```

test/helper/TestFactory.js
```diff
@@ -118,11 +118,12 @@ export class TestFactory {
     this.cryptography_repository = new CryptographyRepository(this.cryptography_service, this.storage_repository);
     this.cryptography_repository.currentClient = ko.observable(currentClient);
 
-    if (mockCryptobox) {
+    if (mockCryptobox === true) {
```
lipis, line 121: was that an issue?

src/script/main/app.js
bennycode: I don't want to rely on default values. I am actually considering removing the default.

test/helper/TestFactory.js
bennycode: I want to be sure to check against a boolean. If `mockCryptobox` would be `'a'` the condition would still be valid but it should not.

src/script/cryptography/CryptographyRepository.js
bennycode: `useExistingCryptobox` should be a boolean, I don't want to wrongly accept numbers or characters.

PR #8453, opened 2020-03-31
fix: Use JSON mapping from SQLeet (WEBAPP-6765)

ffflorian:
This uses the fixed JSON mapping, see https://github.com/wireapp/wire-web-packages/pull/2913.

### DIFF
--- a/src/script/storage/StorageService.ts
+++ b/src/script/storage/StorageService.ts
@@ -303,7 +303,8 @@ export class StorageService {
    */
   async getAll<T = Object>(storeName: string): Promise<T[]> {
     try {
-      return await this.engine.readAll<T>(storeName);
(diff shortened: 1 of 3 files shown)


src/script/storage/StorageService.ts
```diff
@@ -303,21 +303,7 @@ export class StorageService {
...
-        serializedProperties.forEach(property => {
-          if (typeof (record as any)[property] === 'string') {
-            try {
-              (record as any)[property] = JSON.parse((record as any)[property]);
-            } catch (error) {}
-          }
-        });
-        return record;
-      });
+      return await this.engine.readAll<T>(storeName);
```
bennycode: We might still need `filter(Boolean)`.

ffflorian: Whoops, of course!

PR #8456, opened 2020-03-31
feat: Fetch team data when opening account settings (WEBAPP-6759)

ffflorian:


src/script/main/app.js
```diff
@@ -374,7 +374,7 @@ class App {
         user: userRepository,
       } = this.repository;
       await checkIndexedDb();
-      await this._registerSingleInstance();
```
ffflorian, line 377: Not needed

PR #8461, opened 2020-04-01
fix: Icon color in preferences, placement of (you) in user list

AndyLnd:


src/style/content/preferences/account.less
```diff
@@ -41,6 +41,15 @@
 
   edit-icon {
     opacity: 0;
+
+    svg path {
+      fill: currentColor;
+    }
+  }
+  animated-check-icon {
```
ffflorian, line 50:
```suggestion

  animated-check-icon {
```

PR #8510, opened 2020-04-07
feat: Show members found on backend in start UI

AndyLnd:


src/script/components/userList.ts
```diff
@@ -175,7 +177,7 @@ ko.components.register('user-list', {
     };
 
     // Filter all list items if a filter is provided
-    this.filteredUserEntities = ko.pureComputed(() => {
+    this.filteredUserEntities = ko.pureComputed(async () => {
```
ffflorian: Why is this `async` now?

src/script/view_model/list/StartUIViewModel.js
```diff
@@ -335,25 +336,14 @@ class StartUIViewModel {
     this.services.removeAll();
   }
 
-  _searchPeople(query) {
+  async _searchPeople(query) {
```
ffflorian, line 339: Why is this `async` now?

AndyLnd: There is an await now in line 373.

ffflorian: Ah, I see!

PR #8516, opened 2020-04-08
fix: Show modal when a deleted conversation was opened via deeplink (WEBAPP-6748)

ffflorian:


src/script/conversation/ConversationRepository.js
```diff
@@ -3178,6 +3178,15 @@ export class ConversationRepository {
           z.error.ConversationError.TYPE.CONVERSATION_NOT_FOUND,
         ];
 
+        const isConversationNotFound = error.label === BackendClientError.LABEL.ACCESS_DENIED;
```
bennycode: Let's use a different name to not confuse access denied with a real conversation not found error. How about `isRemovedFromConversation`?

PR #8517, opened 2020-04-08
runfix: Don't inject messages to remove users (WEBAPP-6777)

bennyn:


bennycode:
Replaced by https://github.com/wireapp/wire-webapp/pull/8518

PR #8556, opened 2020-04-15
refactor: Migrate AssetService to async/await

bennyn:


src/script/util/util.ts
```diff
@@ -135,12 +135,12 @@ export const loadUrlBuffer = (
...
 };
 
-export const loadImage = function (blob: Blob): Promise<GlobalEventHandlers> {
+export const loadImage = function (blob: Blob): Promise<HTMLImageElement> {
   return new Promise((resolve, reject) => {
     const object_url = window.URL.createObjectURL(blob);
     const img = new Image();
     img.onload = function (): void {
-      resolve(this);
+      resolve(img);
```
ffflorian, line 143: If this works, you can use an arrow function in the line above.

src/script/assets/AssetService.ts
```diff
@@ -231,23 +221,21 @@ export class AssetService {
...
-      })
-      .then(compressedBytes => {
-        return loadImage(new Blob([compressedBytes], {type: image.type})).then(compressedImage => ({
-          compressedBytes,
-          compressedImage,
-        }));
-      });
+    const buffer = await loadFileBuffer(image);
+    if (typeof filter === 'function' ? filter() : undefined) {
+      new Uint8Array(buffer as ArrayBuffer);
```
ffflorian: There is something missing?

bennycode: Good catch, I reverted it.

src/script/assets/AssetService.ts
```diff
@@ -231,23 +221,23 @@ export class AssetService {
...
-      })
-      .then(compressedBytes => {
-        return loadImage(new Blob([compressedBytes], {type: image.type})).then(compressedImage => ({
-          compressedBytes,
-          compressedImage,
-        }));
-      });
+    const buffer = await loadFileBuffer(image);
+    let compressedBytes: ArrayBuffer;
+    if (typeof filter === 'function' ? filter() : undefined) {
```
ffflorian: IMO it makes more sense to make clear that, if no `filter` is available, this should not run:

```suggestion
    if (typeof filter === 'function' ? filter() : false) {
```

What do you think?

PR #8557, opened 2020-04-16
runfix: Don't fail on bad requests when fetching users

bennyn:


src/script/user/UserRepository.ts
```diff
@@ -510,13 +511,14 @@ export class UserRepository {
       return Promise.resolve([]);
     }
 
-    const _getUsers = (chunkOfUserIds: string[]) => {
+    const _getUsers = (chunkOfUserIds: string[]): Promise<(void | User)[]> => {
```
ffflorian, line 514:
```suggestion
    const _getUsers = (chunkOfUserIds: string[]): Promise<Array<void | User>> => {
```

bennycode: I took the style from "mapUsersFromJson" to keep it consistent.

PR #8562, opened 2020-04-16
feat: Check for missing team member when no clients are left (WEBAPP-6760)

ffflorian:


src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -176,40 +184,39 @@ export class ClientMismatchHandler {
...
-            const timestamp = this.serverTimeHandler.toServerTimestamp();
-            const event = EventBuilder.buildMemberLeave(conversationEntity, userId, false, timestamp);
-            this.eventRepository.injectEvent(event);
+      if (noRemainingClients && typeof conversationEntity !== 'undefined' && conversationEntity.isGroup()) {
+        try {
+          const users = await this.userRepository.fetchUsersById([userId]);
+          if (users.length === 0) {
+            throw new Error('No users found');
           }
+        } catch (error) {
```
bennycode: Catching here is not a good idea as it might inject member leave messages when the network request fails. Just check if `users.length === 0` and then inject the event.

src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -176,40 +184,35 @@ export class ClientMismatchHandler {
...
-      if (noRemainingClients) {
-        if (conversationEntity !== undefined) {
-          if (conversationEntity.isGroup() && verifyDeletedUsers) {
-            const timestamp = this.serverTimeHandler.toServerTimestamp();
-            const event = EventBuilder.buildMemberLeave(conversationEntity, userId, false, timestamp);
-            this.eventRepository.injectEvent(event);
-          }
+      if (noRemainingClients && typeof conversationEntity !== 'undefined' && conversationEntity.isGroup()) {
+        const users = await this.userRepository.fetchUsersById([userId]);
+        if (users.length === 0 || users.every(user => user && !!user.isDeleted)) {
```
ffflorian: TODO: Use the UserService? Since the UserRepository even returns a "deleted" user back when it was deleted on the backend.

src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -84,27 +85,31 @@ export class ClientMismatchHandler {
    * @note Contains clients of which the backend is sure that they should not be recipient of a message and verified they no longer exist.
    * @param recipients User client map containing redundant clients
    * @param payload Payload of the request
-   * @param [conversationEntity] Conversation entity
+   * @param conversationEntity Conversation entity
```
bennycode, line 88:
```suggestion
   * @param [conversationEntity] Conversation entity
```
Conversation entity is still optional.

src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -176,40 +185,41 @@ export class ClientMismatchHandler {
...
    * @param conversationEntity Conversation entity
    * @param payload Initial payload resulting in a 412
-   * @param verifyDeletedUsers Verifies if users still exist on the backend by checking their active clients
-   * @returns Function array
    */
-  removePayload(
+  async removePayload(
     recipients: UserClients,
-    clientFn: Function,
+    clientFn: (userId: string, clientId: string) => any | Promise<any>,
```
bennycode:
```suggestion
    clientFn: (userId: string, clientId: string) => void | Promise<void>,
```

src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -84,27 +85,31 @@ export class ClientMismatchHandler {
...
+    recipients: UserClients,
+    payload: NewOTRMessage,
+    conversationEntity?: Conversation,
+  ): Promise<void> {
     if (Object.entries(recipients).length === 0) {
       return;
     }
     this.logger.debug(`Message contains deleted clients of '${Object.keys(recipients).length}' users`, recipients);
-    const removeDeletedClient = (userId: string, clientId: string) => {
+    const removeDeletedClient = async (userId: string, clientId: string) => {
```
bennycode: Return type missing

src/script/conversation/ClientMismatchHandler.ts
ffflorian: Yes, but no need to reflect that in the JSDoc since we are using TypeScript here.

ffflorian: At least in VS Code, adding square brackets around the parameter name is considered invalid in TypeScript:
screenshot

While removing the brackets highlights the parameter name correctly:
screenshot

bennycode: Ok, looks like TSDoc has no support for optionals: https://github.com/Microsoft/vscode/issues/66079 & https://github.com/Microsoft/tsdoc/issues/151#issuecomment-473423780

PR #8596, opened 2020-04-21
runfix: Handle call mute state when connecting

bennyn:


jspittka:
> Although I liked the idea that the button is disabled while connecting.

Allowing to mute during connecting has been a specific request by our investor(s) in the past.

PR #8598, opened 2020-04-21
refactor: Extend asset options from API client

bennyn:


ffflorian:
Can we also delete `assets/AssetRetentionPolicy.ts`?

bennycode:
Not yet

PR #8611, opened 2020-04-22
refactor: Convert components to TypeScript

ffflorian:


src/script/components/message.ts
```diff
@@ -169,12 +244,12 @@ class Message {
...
   };
 
-  showContextMenu(messageEntity, event) {
+  showContextMenu(messageEntity: ContentMessage, event: Event) {
     const entries = [];
 
     if (messageEntity.is_downloadable()) {
       entries.push({
-        click: () => messageEntity.download(),
+        click: () => (messageEntity as any).download(),
```
bennycode: A `ContentMessage` has the `download` property defined?!

ffflorian: Whoops, left over from trying to figure out if `message` is of type `Message` or `ContentMessage` :)

PR #8617, opened 2020-04-23
refactor: Use API client to send images

bennyn:


src/script/assets/AssetUploader.ts
```diff
@@ -61,56 +95,25 @@ export class AssetUploader {
...
-          sha256: new Uint8Array(sha256),
-        });
-        const protoAsset = new Asset({
-          [PROTO_MESSAGE_TYPE.ASSET_UPLOADED]: assetRemoteData,
-          [PROTO_MESSAGE_TYPE.EXPECTS_READ_CONFIRMATION]: options.expectsReadConfirmation,
-          [PROTO_MESSAGE_TYPE.LEGAL_HOLD_STATUS]: options.legalHoldStatus,
-        });
+      .then(async uploadedAsset => {
+        const protoAsset = this.buildProtoAsset(encryptedAsset, uploadedAsset, options);
+        if (isImage === true) {
```
AndyLnd, line 100:
```suggestion
        if (isImage) {
```

bennycode: I like the explicit boolean check more. People using the code from JavaScript should not have a chance when entering '1'. :D

PR #8669, opened 2020-04-29
refactor: Convert User to TypeScript

ffflorian:


src/script/entity/User.ts
```diff
@@ -158,42 +209,28 @@ class User {
...
   }
 
-  /**
-   * Check whether handle or name matches the given query
-   * @param {string} query Query
-   * @param {boolean} is_handle Query string is handle
-   * @param {array} excludedChars list of chars to exclude from getSlug
-   * @returns {undefined} No return value
-   */
-  matches(query, is_handle, excludedChars = []) {
```
ffflorian, line 182: This was not in use and was using `compareTransliteration` in a wrong way, too.

PR #8711, opened 2020-05-05
refactor: Cleanup dependencies and globals

ffflorian:


ffflorian:
> Why do you switch to `uuidjs`?

I switched only in two files, in the other files we are already using `uuidjs` and the `pure-uuid` dependency was missing in `package.json`. The `@types/uuid` dependency was wrong though.

PR #8726, opened 2020-05-05
fix: Remove locale from website URL (WEBAPP-6823)

bennyn:


Yserz:
I think it would be good to get rid of `addLocaleToUrl` completely. With this we might have fixed this single occurrence of the problem but not others.

bennycode:
I will remove more occurrences of `addLocaleToUrl` when I got feedback from Jens.

PR #8730, opened 2020-05-06
chore: Remove last bits of tslint

ffflorian:


server/package.json
```diff
@@ -32,19 +32,11 @@
     "nyc": "15.0.1",
     "rimraf": "3.0.2",
     "ts-node": "8.9.1",
-    "tslint": "5.20.1",
```
ffflorian, line 35: Not needed in the server part since `eslint` will check the files from the parent directory.

PR #8731, opened 2020-05-06
fix: Close user details on member removed discovery (WEBAPP-6846)

AndyLnd:


src/script/view_model/PanelViewModel.js
```diff
@@ -238,7 +238,10 @@ export class PanelViewModel {
 
     const panelStateElementId = this.subViews[state].getElementId();
     const exitPanel = $(`#${panelStateElementId}`);
-    exitPanel.removeClass('panel__page--visible panel__page--move-out--left panel__page--move-out--right');
+    exitPanel.removeClass('panel__page--move-out--left panel__page--move-out--right');
+    if (this.state() !== state) {
```
bennycode, line 242: Can you check for `isVisible` here?

AndyLnd: Not really, as `isVisible` is independent from this class name. This check here is really just for the rare case when we want to go back to the conversation details view while it is still animating to another view.

PR #8810, opened 2020-05-15
refactor: Use async-await in CallingRepository

bardiarastin98:


src/script/calling/CallingRepository.ts
```diff
@@ -302,22 +302,23 @@ export class CallingRepository {
...
+          call.selfParticipant.videoState(VIDEO_STATE.STARTED);
         }
-        return true;
-      })
-      .catch(() => false);
+      } else {
+        mediaStream.getTracks().forEach(track => track.stop());
+      }
+      return true;
+    } catch (_) {
```
bennycode, line 319:
```suggestion
    } catch () {
```

PR #8842, opened 2020-05-18
feat: Store Conversations/Folders view into backend

lipis:


src/script/properties/PropertiesType.ts
```diff
@@ -41,12 +41,17 @@ enum PROPERTIES {
   VERSION = 'version',
 }
 
+enum VIEW {
+  RECENT_CONVERSATIONS = 'settings.view.recent_conversations',
```
lipis: not sure what should be the label.

PR #8887, opened 2020-05-26
runfix: Respect deviceId in calls, use selfuser from avs callbacks (WEBAPP-6880)

AndyLnd:


src/script/calling/Call.ts
```diff
@@ -50,17 +49,23 @@ export class Call {
...
+  }
+
   addParticipant(participant: Participant): void {
-    this.participants.unshift(participant);
+    this.participants.push(participant);
   }
 
-  getParticipant(userId: string): Participant {
-    return this.participants().find(participant => participant.userId === userId);
+  getParticipant(userId: UserId, deviceId: DeviceId): Participant {
```
bennycode: Let's call it `clientId`.

test/unit_tests/components/list/conversationListCallingCellSpec.js
```diff
@@ -27,9 +27,9 @@ import 'src/script/components/list/conversationListCallingCell';
 import {User} from 'src/script/entity/User';
 
 function createCall(state) {
-  const call = new Call();
+  const selfParticipant = new Participant('selfId');
+  const call = new Call('', '', undefined, selfParticipant);
```
bennycode, line 32: Let's use `createRandomUuid` to pass IDs to the call so it looks a little more like on production.

src/script/calling/Call.ts
AndyLnd: I named it `deviceId` here, as the property on the `Participant` is also called `deviceId`

bennycode: So `DeviceId` should also be renamed.

AndyLnd: renamed everywhere

PR #8891, opened 2020-05-28
fix: Correctly stop screen sharing on stop screen button click

bardiarastin98:


src/script/calling/CallingRepository.ts
```diff
@@ -811,6 +811,12 @@ export class CallingRepository {
       .then(mediaStream => {
         this.mediaStreamQuery = undefined;
         const newStream = selfParticipant.updateMediaStream(mediaStream);
+        if (missingStreams.screen) {
+          // https://stackoverflow.com/q/25141080/7091035
```
bennycode:
```suggestion
          // https://stackoverflow.com/a/25179198/451634
```

It's better to link the answer than the question.

PR #8896, opened 2020-05-29
chore: Update translations

Yserz:


Aikatsui:
@Yserz @AndyLnd @bennyn anyone can help to add new language to wire in crowdin? i have sent email and request through crowdin but none of them respond. Only does thing is that approve me silently if request to join for translate existing language. Is it possible to enable Sinhala language and set access to @iStore for all projects?

PR #8977, opened 2020-06-10
fix: Broken CSS for audio wave forms (WEBAPP-6896)

lipis:


src/style/common/mixins.less
```diff
@@ -190,12 +190,6 @@
   justify-content: center;
 }
 
-.vertical-align() {
-  position: relative;
-  top: 50%;
-  transform: translateY(-50%);
-}
-
```
lipis, line 198: It was used only in the audio seek bar and since I'm overriding one value.. not sure if we want to keep unused code.

PR #9006, opened 2020-06-16
fix: Change the order of the badge icons in participant item

lipis:


src/style/components/list/participant-item.less
```diff
@@ -42,14 +42,6 @@ participant-item {
...
   }
 
-  verified-icon + guest-icon,
-  verified-icon + partner-icon,
-  verified-icon + camera-icon,
-  partner-icon + camera-icon,
-  guest-icon + camera-icon {
-    margin-left: 16px;
-  }
-
```
lipis, line 59: not used like this anymore

PR #9025, opened 2020-06-18
feat: Show unsupported browser pop-up (WEBAPP-6923) (WEBAPP-6927)

lipis:
Updated crowdin as well


bennycode:
At a later point in time feature detection will be done with:

```ts
const pt: any = RTCRtpSender.prototype;  
pc.insertable_legacy = !!pt.createEncodedVideoStreams;
pc.insertable_streams = !!pt.createEncodedStreams;

// const supportsInsertableStreams = !!RTCRtpSender.prototype.createEncodedVideoStreams;
```

PR #9070, opened 2020-06-26
runfix: Add missing ConversationRepository property to GroupCreationViewModel

thisisamir98:


src/script/view_model/content/GroupCreationViewModel.ts
```diff
@@ -72,6 +73,7 @@ export class GroupCreationViewModel {
     private readonly userRepository: UserRepository,
   ) {
     this.isTeam = this.teamRepository.isTeam;
+    this.ConversationRepository = ConversationRepository;
```
bennycode, line 76: Can't you use the existing `this.conversationRepository`?

thisisamir98: can't access static method on an instance :(

PR #9101, opened 2020-06-29
chore: Auto open correct local address

AndyLnd:


server/index.ts
```diff
@@ -28,7 +28,7 @@ server
   .then(port => {
     console.info(`[${formatDate()}] Server is running on port ${port}.`);
     if (config.SERVER.DEVELOPMENT) {
-      require('opn')(`http://localhost:${config.SERVER.PORT_HTTP}`);
+      require('opn')(`https://local.zinfra.io:${config.SERVER.PORT_HTTP}`);
```
bennycode: Let's use the `APP_BASE` from `.env` here. You use `process.env.APP_BASE`.

PR #9111, opened 2020-07-01
refactor: Improve initial message sending performance in big conversations

Yserz:


src/script/conversation/ConversationRepository.js
```diff
@@ -2755,17 +2755,21 @@ export class ConversationRepository {
...
 
         if (noRemainingClients) {
-          const backendUser = await this.userRepository.getUserFromBackend(recipientId);
-          const isDeleted = backendUser?.deleted === true;
+          userIdsWithoutClients.push(recipientId);
+        }
+      }
+      const userList = await this.userRepository.getUsersById(userIdsWithoutClients);
+      for (const user of userList) {
+        const isDeleted = user?.deleted === true;
```
AndyLnd: Shouldn't it be `user?.isDeleted`?

Yserz: I think that's a value from the backend. Either way it's just what it was before :>

Yserz: Mh, looking a bit deeper in the code...this stuff seems to be pretty broken ;P

PR #9118, opened 2020-07-02
feat: Add 'file type restricted' system message (WEBAPP-6980)

AndyLnd:


src/script/entity/message/FileTypeRestrictedMessage.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ */
+
+import {t} from 'Util/LocalizerUtil';
+
+import {SuperType} from '../../message/SuperType';
+import {Message} from './Message';
+
+/**
+ * Legal hold system message
```
bennycode: Wrong comment?

test/main.test.js
```diff
@@ -23,6 +23,7 @@ window.wire.env = {
   BACKEND_REST: 'http://localhost',
   BACKEND_WS: 'wss://localhost',
   FEATURE: {
+    ALLOWED_FILE_UPLOAD_EXTENSIONS: ['*'],
```
Yserz, line 26: I think for tests the default should be restrictive and should be more relaxed in specific test cases with some mocking.

E.g.:
```
spyOn<{getConfig: () => TypeUtil.RecursivePartial<Configuration>}>(Config, 'getConfig').and.returnValue({
      BACKEND_NAME: 'Test',
      FEATURE: {
        ENABLE_ACCOUNT_REGISTRATION: true,
      },
    });
```

src/script/util/util.ts
```diff
@@ -177,6 +179,35 @@ export const trimFileExtension = (filename: string): string => {
   return '';
 };
 
+export const allowedImageTypes = ['image/bmp', 'image/gif', 'image/jpeg', 'image/jpg', 'image/png', '.jpg-large'];
```
Yserz, line 182: Mh, not a big fan of pulling this out of the config and into a util. Config seems to be the correct place for this kind of stuff.

test/main.test.js
AndyLnd: Very good point, I will address this in a later PR.

PR #9157, opened 2020-07-06
fix: Video icon in participant list for user that started the video call from different client

lipis:


src/script/components/list/conversationListCallingCell.ts
```diff
@@ -174,11 +174,11 @@ class ConversationListCallingCell {
...
     });
 
-    this.users = ko.pureComputed(() =>
+    this.participants = ko.pureComputed(() =>
       call
         .participants()
-        .map(({user}) => user)
-        .sort(sortUsersByPriority),
+        .slice()
+        .sort((a, b) => sortUsersByPriority(a.user, b.user)),
```
bennycode:
```suggestion
        .sort((participantA, participantB) => sortUsersByPriority(a.user, b.user)),
```

PR #9169, opened 2020-07-07
fix: Remember quoted message in Inputbar (WEBAPP-6991)

AndyLnd:


src/script/view_model/content/InputBarViewModel.ts
```diff
@@ -71,6 +71,12 @@ type DraftMessage = {
   text: string;
 };
 
+interface Draft {
+  mentions: MentionEntity[];
+  reply: {messageId: string} | {};
```
ffflorian:
```suggestion
  reply: {messageId?: string};
```

PR #9176, opened 2020-07-08
refactor: Stricter check for some undefined cases

ffflorian:


src/page/template/list/conversations.htm
```diff
@@ -20,7 +20,7 @@
   </div>
   <!-- ko using: callingViewModel -->
     <!-- ko foreach: {data: activeCalls, as: 'activeCall', noChildContext: true} -->
-      <!-- ko if: !activeCall.reason() -->
+      <!-- ko if: activeCall.reason() === CALL_REASON.NORMAL -->
```
ffflorian, line 23: Since `activeCall.reason()` can be `undefined` and `0`, we should check what we actually want here.

src/script/view_model/CallingViewModel.ts
```diff
@@ -114,13 +116,13 @@ export class CallingViewModel {
...
       const soundId = sounds[initialCallState];
-      if (!soundId || call.reason() !== undefined) {
+      if (!soundId || typeof call.reason() !== 'undefined') {
         return;
       }
 
       audioRepository.loop(soundId).then(() => {
         const stateSubscription = ko.computed(() => {
-          if (call.state() !== initialCallState || call.reason() !== undefined) {
+          if (call.state() !== initialCallState || typeof call.reason() !== 'undefined') {
```
ffflorian, line 125: From https://stackoverflow.com/a/4725697:

> For undeclared variables, `typeof foo` will return the string literal `"undefined"`, whereas the identity check `foo === undefined` would trigger the error _"foo is not defined"_.
>
> For local variables (which you know are declared somewhere), no such error would occur, hence the identity check.

But since `call.reason()` can be `undefined`, we can run into an error here.

PR #9182, opened 2020-07-08
feat: Don't show "outdated version" banner for 15 minutes when accepted (WEBAPP-7002)

AndyLnd:


src/script/calling/CallingRepository.ts
```diff
@@ -120,6 +123,17 @@ export class CallingRepository {
       return this.activeCalls().find(call => call.state() === CALL_STATE.MEDIA_ESTAB);
     });
 
+    this.acceptedVersionWarnings = ko.observableArray<string>();
+    this.acceptVersionWarning = (conversationId: string) => {
+      this.acceptedVersionWarnings.push(conversationId);
+      window.setTimeout(() => this.acceptedVersionWarnings.remove(conversationId), TIME_IN_MILLIS.MINUTE * 15);
```
thisisamir98, line 129: is it necessary to explicitly use window here?

AndyLnd: It is not necessary but it is more explicit.

PR #9189, opened 2020-07-09
refactor: Migrate FaviconViewModel to Typescript

thisisamir98:


src/script/view_model/MainViewModel.js
```diff
@@ -96,7 +97,7 @@ export class MainViewModel {
 
     this.lightbox = new z.viewModel.ImageDetailViewViewModel(this, repositories);
     this.title = new WindowTitleViewModel(this, repositories);
-    this.favicon = new z.viewModel.FaviconViewModel(window.amplify);
+    this.favicon = new FaviconViewModel(window.amplify);
```
bennycode: It it possible to import `amplify` here instead of using `window.amplify`?

thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/9189/commits/d9e9b2dc71b4ebe8edde654396e840a4476b6ae3

PR #9263, opened 2020-07-22
feat: Send targeted messages

thisisamir98:


src/script/calling/CallingRepository.ts
```diff
@@ -698,7 +708,18 @@ export class CallingRepository {
       return 0;
     }
 
-    const options = this.targetMessageRecipients(payload, destinationUserId, destinationClientId);
+    let options: MessageSendingOptions;
+
+    if (typeof targets === 'string') {
+      const parsedTargets: SendMessageTarget = JSON.parse(targets);
```
franziskuskiefer, line 714: `JSON.parse` might throw. I don't see that handled.

bennycode: It's an async callback which cannot be fulfilled when `JSON.parse` throws which is okay since we then get notified about the error.

PR #9271, opened 2020-07-24
feat: Sync setting for SFT calling beta program

bennyn:


src/script/calling/CallingRepository.ts
```diff
@@ -98,6 +98,7 @@ export class CallingRepository {
   private readonly logger: Logger;
   private readonly callLog: string[];
   private readonly cbrEncoding: ko.Observable<number>;
+  private readonly useSftCalling: ko.Observable<boolean>;
```
AndyLnd, line 101: Please also add this to the constructor with a default value.

bennycode: Good catch. I was hoping for TypeScript to catch uninitialized paramters.

PR #9300, opened 2020-07-27
feat: Enable Beta program SFT calling via `enable_sft_calling` URL parameter (WEBAPP-7025)

bennyn:
**How to use the parameter?**
- Use a URL like `https://local.zinfra.io:8081/?enable_sft_calling=true`
- Check with JavaScript code `wire.app.repository.calling.useSftCalling()` if it worked

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -64,6 +64,8 @@ import {ClientId, Participant, UserId} from './Participant';
 import type {Recipients} from '../cryptography/CryptographyRepository';
 import type {Conversation} from '../entity/Conversation';
 import {UserRepository} from '../user/UserRepository';
+import {UrlUtil} from '@wireapp/commons';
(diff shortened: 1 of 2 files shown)


src/script/calling/CallingRepository.ts
```diff
@@ -153,7 +155,14 @@ export class CallingRepository {
   }
 
   toggleSftCalling(enableSftCalling: boolean) {
-    this.useSftCalling(this.supportsConferenceCalling && enableSftCalling);
+    const urlSetting = UrlUtil.getURLParameter(QUERY_KEY.ENABLE_SFT_CALLING);
+    if (urlSetting) {
```
ffflorian, line 159: This will trigger if `enable_sft_calling` is set to `'false'`, right?

```suggestion
    if (urlSetting === 'true') {
```

bennyn: Yes, but it should trigger whenever the param is set.

PR #9306, opened 2020-07-29
refactor: Detect conference calling support via webapp

bennyn:


src/script/calling/CallingRepository.ts
```diff
@@ -363,8 +363,16 @@ export class CallingRepository {
     }
   }
 
+  /**
+   * Checks if browser supports all WebRTC APIs which are required for conference calling. Users with Chrome 83 need to enable "Experimental Web Platform features" (--enable-experimental-web-platform-features) to use all required APIs.
```
lipis, line 367: maybe breaking this line into multiple ones for easier reading.

bennycode: The linter didn't complain. So I will keep it like this.

PR #9309, opened 2020-07-31
runfix: Show video call button based on sft enabled state

AndyLnd:


src/script/Config.ts
```diff
@@ -42,7 +42,8 @@ export class Configuration {
   readonly ENVIRONMENT = env.ENVIRONMENT || 'production';
   readonly FEATURE = env.FEATURE;
   readonly MAX_GROUP_PARTICIPANTS = env.MAX_GROUP_PARTICIPANTS || 500;
-  readonly MAX_VIDEO_PARTICIPANTS = env.MAX_VIDEO_PARTICIPANTS || 4;
+  readonly MAX_SFT_VIDEO_PARTICIPANTS = env.MAX_VIDEO_PARTICIPANTS || 4;
+  readonly MAX_VIDEO_PARTICIPANTS = Math.min(4, this.MAX_SFT_VIDEO_PARTICIPANTS);
```
thisisamir98, line 46: here we're doing MAX_VIDEO_PARTICIPANTS and using Math.min? 🤔

src/script/Config.ts
```diff
@@ -42,7 +42,8 @@ export class Configuration {
   readonly ENVIRONMENT = env.ENVIRONMENT || 'production';
   readonly FEATURE = env.FEATURE;
   readonly MAX_GROUP_PARTICIPANTS = env.MAX_GROUP_PARTICIPANTS || 500;
-  readonly MAX_VIDEO_PARTICIPANTS = env.MAX_VIDEO_PARTICIPANTS || 4;
+  readonly MAX_SFT_VIDEO_PARTICIPANTS = env.MAX_VIDEO_PARTICIPANTS || 4;
```
thisisamir98, line 45: shouldn't this be env.MAX_SFT_VIDEO_PARTICIPANTS instead of env.MAX_VIDEO_PARTICIPANTS? what if we set env.MAX_VIDEO_PARTICIPANTS to 40 and someone without beta program enabled would start a call?

PR #9337, opened 2020-08-04
runfix: check for degraded conversations before calls (WEBAPP-6988, WEBAPP-7015)

AndyLnd:


src/script/calling/CallingRepository.ts
```diff
@@ -245,15 +247,59 @@ export class CallingRepository {
     try {
       await this.apiClient.conversation.api.postOTRMessage(this.selfClientId, conversationId);
     } catch (error) {
-      const mismatch = (error as AxiosError).response!.data;
-      const data: {clients: {clientid: string; userid: string}[]} = {
-        clients: [],
+      type ClientListEntry = [string, string];
```
ffflorian: Let's define this outside of the function

src/script/calling/CallingRepository.ts
```diff
@@ -245,15 +247,59 @@ export class CallingRepository {
...
+          [],
+        );
+
+      const isSameEntry = ([userA, clientA]: ClientListEntry, [userB, clientB]: ClientListEntry): boolean =>
+        userA === userB && clientA === clientB;
+
+      const fromClientList = (clientList: ClientListEntry[]): UserClients =>
+        clientList.reduce(
+          (acc, [userId, clientId]) => ({...acc, [userId]: [...(acc[userId] || []), clientId]}),
+          {} as UserClients,
```
ffflorian: Can we use a more readable reducer function? :slightly_smiling_face:

PR #9352, opened 2020-08-07
fix: Include user availability status in accountInfo (webapp-7046)

thisisamir98:


ffflorian:
I'd say this is more of a feature than a bugfix.

thisisamir98:
@ffflorian on the wrapper side it's a bug so I taught that `fix` would be okay

PR #9428, opened 2020-08-13
fix: Disable initial fullscreen video for self screen sharing (WEBAPP-6847)

thisisamir98:


src/script/view_model/CallingViewModel.ts
```diff
@@ -231,6 +231,12 @@ export class CallingViewModel {
         };
 
         this.mediaStreamHandler.selectScreenToShare(showScreenSelection).then(() => {
+          if (
+            [CALL_TYPE.NORMAL, CALL_TYPE.FORCED_AUDIO].includes(call.initialType) ||
```
bennycode: Let's put it into more readable variablies, i.e.

```
const isAudioCall = [CALL_TYPE.NORMAL, CALL_TYPE.FORCED_AUDIO].includes(call.initialType);
const isFullScreenVideoCall = (call.initialType === CALL_TYPE.VIDEO && !this.multitasking.isMinimized();

if (isAudioCall || isFullScreenVideoCall) { ... }
```

thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/9428/commits/02e64f1fe83cc75b3acc063dbe31703cb388c51a

PR #9429, opened 2020-08-13
runfix: Use "What's new link" from server config (WEBAPP-7041)

bennyn:


src/script/Config.ts
```diff
@@ -45,28 +45,7 @@ export class Configuration {
...
-      INDEX: 'https://support.wire.com/',
-      MICROPHONE_ACCESS_DENIED: 'https://support.wire.com/hc/articles/202590081',
-      SCREEN_ACCESS_DENIED: 'https://support.wire.com/hc/articles/202935412',
-    },
-    TEAMS_BASE: 'https://teams.wire.com',
-    TERMS_OF_USE_PERSONAL: 'https://wire-website-staging.zinfra.io/legal/terms/personal',
-    TERMS_OF_USE_TEAMS: 'https://wire-website-staging.zinfra.io/legal/terms/teams',
-    WEBSITE_BASE: 'https://wire.com',
-  };
+  readonly URL = env.URL;
```
ffflorian: Why was this removed?

bennycode: Because we use defaults from our config repo/files now.

PR #9434, opened 2020-08-14
runfix: Remove rendered lists in markdown for now

AndyLnd:


bennycode:
It's okay to keep the CSS because it will come back once we reactivate the feature.

PR #9440, opened 2020-08-17
fix: Fix the gaps between lists in markdown (WEBAPP-7056)

lipis:


lipis:
Was just about to say that it wasn't merged.. but it was merged today :D

bennycode:
Ok, we won't touch the Markdown lists for now.

PR #9524, opened 2020-08-28
chore: Migrate EventServiceNoCompound to TS

AndyLnd:


src/script/event/EventServiceNoCompound.ts
```diff
@@ -76,24 +87,24 @@ export class EventServiceNoCompound extends EventService {
...
-        const timestamp = new Date(record.time).getTime();
+        const timestamp = new Date(record.time);
         return (
           record.conversation === conversationId &&
           fromCompareFunction(fromDate, timestamp) &&
           toCompareFunction(toDate, timestamp)
         );
       })
-      .sort((a, b) => a.conversation - b.conversation)
+      .sort()
```
AndyLnd: As we are comparing strings here (conversationId) a comparator function is not necessary, as this is the default method for `sort()`

lipis (approved):

src/script/event/EventServiceNoCompound.ts
ffflorian: But the `record` is of type `DBEvent` and not of type `string`?

ffflorian (changes requested):

src/script/event/EventServiceNoCompound.ts
AndyLnd: You are absolutely right. 🤦🏻‍♂️

ffflorian (approved):
Function return values would be nice, but it's okay without it for now.

PR #9563, opened 2020-09-01
refactor: Remove redundant profile image worker

ffflorian:


src/worker/image-worker.js
```diff
@@ -1,24 +1,30 @@
...
+    MAX_FILE_SIZE = 1024 * 1024;
+    COMPRESSION = 80;
+  }
 
-  Jimp.read(event.data).then((image) => {
-
-    if ((image.bitmap.width > MAX_SIZE) || (image.bitmap.height > MAX_SIZE)) {
+  Jimp.read(event.data.buffer).then(image => {
+    if (event.data.useProfileImageSize) {
+      image.cover(MAX_SIZE, MAX_SIZE);
```
bennycode, line 38: Is this width and height? If yes, does it make sense to have one value for each?

bennyn (approved):

src/worker/image-worker.js
ffflorian: Not sure. I just copied this part over from the other worker.

PR #9600, opened 2020-09-07
runfix: Send call outgoing-incoming instead of call state

thisisamir98:


src/script/calling/CallingRepository.ts
```diff
@@ -682,7 +682,7 @@ export class CallingRepository {
       this.wCall.answer(this.wUser, call.conversationId, callType, this.cbrEncoding());
 
       this.sendCallingEvent(EventName.CALLING.JOINED_CALL, call, {
-        [Segmentation.CALL.DIRECTION]: call.state(),
+        [Segmentation.CALL.DIRECTION]: call.initiator === call.getSelfParticipant().user.id ? 'outgoing' : 'incoming',
```
bennycode: You are using this part of code quite a lot. How about making a util function for `call.initiator === call.getSelfParticipant().user.id ? 'outgoing' : 'incoming'`?

bennyn (changes requested):

src/script/calling/CallingRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/9600/commits/39b5d72b6a41c5a8f1ffac272400edc3de665152

PR #9608, opened 2020-09-08
runfix: Set correct environment version (WEBAPP-7081)

ffflorian:
Follow-up to https://github.com/wireapp/wire-webapp/pull/9553.

### DIFF
--- a/src/script/util/Environment.ts
+++ b/src/script/util/Environment.ts
@@ -28,25 +28,23 @@ const APP_ENV = {
 
 const getAppVersion = (): string => {
   const versionElement = document.head.querySelector("[property='wire:version']");
-  const hasVersion = versionElement && versionElement.hasAttribute('version');
-  return hasVersion ? versionElement.getAttribute('version').trim() : '';
+  return versionElement?.getAttribute('version').trim() || '';
 };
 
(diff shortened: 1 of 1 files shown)


src/script/util/Environment.ts
```diff
@@ -28,25 +28,23 @@ const APP_ENV = {
...
+  // [match, version]
+  const [, electronVersion] = /Wire(?:Internal)?\/(\S+)/.exec(userAgent) || [];
   return electronVersion;
 };
 
 const getFormattedAppVersion = (): string => {
-  const [year, month, day, hour, minute] = getAppVersion().split('-');
-  return `${year}.${month}.${day}.${hour}${minute}`;
+  const [year, month, day, hour, minute] = getAppVersion().split('.');
+  return `${year}.${month}.${day}.${hour}.${minute}`;
```
ffflorian, line 42: This is the format we are using everywhere now. Should we use it here, too?

thisisamir98: on webapp we do 
`Config.getConfig().VERSION` which gives -> 2020.09.07.22.34

thisisamir98 (approved):

ffflorian:
:rocket: :smiley:

PR #9615, opened 2020-09-11
refactor: Remove version from HTML

bennyn:
Since the introduction of the config file approach, there is no more need to read the "version" from our HTML code.

### DIFF
--- a/src/page/template/meta.htm
+++ b/src/page/template/meta.htm
@@ -1,7 +1,6 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <meta name="geoip" country="{{country}}" />
-    <meta property="wire:version" version="{{@config.VERSION}}" />
     {{#if @config.CHROME_ORIGIN_TRIAL_TOKEN}}<meta http-equiv="origin-trial" content="{{{@config.CHROME_ORIGIN_TRIAL_TOKEN}}}">{{/if}}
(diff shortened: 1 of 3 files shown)


Yserz (approved):

src/script/util/Environment.ts
```diff
@@ -62,11 +53,11 @@ export const Environment = {
     }
 
     if (doNotFormat) {
-      return getAppVersion();
+      return Config.getConfig().VERSION;
```
ffflorian: There is no formatting anymore so we should only check in this function if we want to show the electron version or not.

src/page/template/meta.htm
```diff
@@ -1,7 +1,6 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <meta name="geoip" country="{{country}}" />
-    <meta property="wire:version" version="{{@config.VERSION}}" />
```
ffflorian, line 4: Let's please keep this because it's quite handy to use.

ffflorian (changes requested):

src/page/template/meta.htm
Yserz: I don't think this is a good way of checking for the application version. Also it's way easier to check for `window.wire.env.VERSION`.

ffflorian: I didn't mean to continue checking from here but to just leave it here for convenience.

Yserz: In general you can find the version also via `/version`, in any resource query below or in `window.wire.env.VERSION` via the console. So there are plenty of places to take the version from.

ffflorian (approved):

PR #9617, opened 2020-09-11
Resizable sent images

Aisse-258:
When open sent image it's size increases on hover.


bennycode:
Thank you for your contribution but we don't plan to change the UX of image display at this point in time. We'll leave this decision to our designers.

PR #9618, opened 2020-09-12
Full password message

Aisse-258:


lipis:
I made these updates in our [Crowdin](https://crowdin.com/translate/wire-webapp/1224/en-ru) so they will be updated in our next release. Thank you!

PR #9645, opened 2020-09-14
fix: Check for lowercase User-Agent (WEBAPP-7046)

ffflorian:


src/script/util/Environment.ts
```diff
@@ -33,7 +33,7 @@ const getAppVersion = (): string => {
 
 const getElectronVersion = (userAgent: string): string => {
   // [match, version]
-  const [, electronVersion] = /Wire(?:Internal)?\/(\S+)/.exec(userAgent) || [];
+  const [, electronVersion] = /Wire(?:Internal)?\/(\S+)/i.exec(userAgent) || [];
```
Yserz, line 36: Now that we know it's always lowercase we could do:
```suggestion
  const [, electronVersion] = /wire(?:internal)?\/(\S+)/.exec(userAgent) || [];
```

ffflorian: Why is that better than using the `i` modifier?

Yserz (approved):

PR #9694, opened 2020-09-24
chore: Bring back babel-plugin-react-intl

Yserz:


src/i18n/en-US.json
```diff
@@ -922,7 +922,7 @@
   "ssoLogin.pasteButton": "Paste code",
   "ssoLogin.subhead": "Enter the company SSO access code.",
   "ssoLogin.subheadCode": "Please enter your SSO code",
-  "ssoLogin.subheadCodeOrEmail": "Please enter your email or SSO code",
+  "ssoLogin.subheadCodeOrEmail": "Please enter your email or SSO code.",
```
ffflorian, line 925: Not part of the PR I think :slightly_smiling_face:

Yserz: It is part of this PR because this is a change that wasn't propagated but should have.

ffflorian: Ah okay

ffflorian (approved):

PR #9695, opened 2020-09-24
chore: Use @formatjs/cli instead of babel-plugin-react-intl

Yserz:


bin/translations_extract_formatter.js
```diff
@@ -0,0 +1,29 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+const webappTranslations = require('../src/i18n/en-US.json');
```
ffflorian, line 20:
```suggestion

const webappTranslations = require('../src/i18n/en-US.json');
```

ffflorian (approved):

PR #9752, opened 2020-09-30
fix: Don't show 2nd modal after degradation (WEBAPP-7108, WEBAPP-7109)

AndyLnd:


ffflorian (approved):
LGTM but the tests are failing.

PR #9792, opened 2020-10-06
refactor: CryptographyRepository - Stronger encapsulation

Yserz:


src/script/cryptography/CryptographyRepository.ts
```diff
@@ -13,7 +13,7 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program. If not, see http://www.gnu.org/licenses/.
+ * along with this program. If not, ssee http://www.gnu.org/licenses/.
```
lipis:
```suggestion
 * along with this program. If not, see http://www.gnu.org/licenses/.
```

lipis (approved):

PR #9828, opened 2020-10-14
refactor: React participant-avatar

AndyLnd:


src/script/components/ParticipantAvatar.tsx
```diff
@@ -0,0 +1,156 @@
...
+  noBadge = false,
+  noFilter = false,
+  size = AVATAR_SIZE.LARGE,
+}) => {
+  const isUser = participant instanceof User && !participant.isService && !participant.isTemporaryGuest();
+  const isService = participant instanceof ServiceEntity || participant.isService;
+  const isTemporaryGuest = !isService && participant.isTemporaryGuest();
+
+  const avatarState = (() => {
+    switch (true) {
```
bennycode, line 93:
```suggestion
    switch (true) {
```

This will always hit the `default` cause. Is it intentional?

bennyn (changes requested):

src/script/components/ParticipantAvatar.tsx
```diff
@@ -0,0 +1,156 @@
...
+      case isService:
+        return STATE.NONE;
+      case participant.isMe:
+        return STATE.SELF;
+      case participant.isTeamMember():
```
bennycode, line 98: Can be put together with the case of `isService`.

src/script/components/participantAvatar/AvatarImage.tsx
```diff
@@ -0,0 +1,118 @@
...
+        const isCached = pictureResource.downloadProgress() === 100;
+        setShowTransition(!isCached && !isSmall);
+        try {
+          const url = await assetRepository.getObjectUrl(pictureResource);
+          if (url) {
+            setAvatarImage(url);
+          }
+          setAvatarLoadingBlocked(false);
+        } catch (error) {
+          console.warn('Failed to load avatar picture.', error);
```
bennycode, line 78: Can we use our `logger` here?

src/script/util/CSSMixin.ts
```diff
@@ -30,3 +30,35 @@ export const CSS_SQUARE: (size: number | string) => CSSObject = size => ({
...
+export const CSS_FLEX_CENTER: CSSObject = {
+  alignItems: 'center',
+  display: 'flex',
+  justifyContent: 'center',
+};
+
+export const CSS_ICON: (code: string, iconSize?: string | number) => CSSObject = (code, iconSize = '16px') => ({
+  MozOsxFontSmoothing: 'grayscale',
+
+  /* Better Font Rendering =========== */
```
bennycode: Comment is not needed

bennyn (changes requested):

src/script/components/ParticipantAvatar.tsx
Yserz: Note: We found out that `isService` needs to be called first in this switch because then the `participant` might be a `ServiceEntity` which does not have all the other accessors. But still if we ensure `isService` is called before `isTeamMember` everything should be fine.

src/script/components/ParticipantAvatar.tsx
AndyLnd: Yes, if none of the other cases are true, `STATE.NONE` should be returned.

bennyn (approved):

PR #9857, opened 2020-10-19
refactor: Add ClientRecord type

ffflorian:


bennyn (changes requested):
I like this typing.  Saving into the DB is part of our storage layer. For our events we have already a typing in `storage/EventRecord.ts`, so lets turn `ClientInDB` into `storage/ClientRecord.ts`.

bennyn (approved):

PR #9976, opened 2020-11-10
chore: Setup daily code scanning alerts at 10 AM UTC

bennycode:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,26 @@
...
+  schedule:
+    - cron: '0 10 * * *'
+
+jobs:
+  'Analyze':
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
Yserz, line 17: I guess we don't need a matrix for just one entry. Does it distinguish between js and ts btw?

Yserz (approved):

.github/workflows/codeql-analysis.yml
bennycode: Their JavaScript setting includes TypeScript: https://help.semmle.com/QL/learn-ql/javascript/introduce-libraries-ts.html#overview

I added the matrix because it was given in the template from GitHub.

lipis (approved):

PR #10061, opened 2020-11-26
refactor: De-duplicate connection requests

bennycode:


src/script/connection/ConnectionRepository.ts
```diff
@@ -187,7 +185,9 @@ export class ConnectionRepository {
    * @returns User connection entity
    */
   getConnectionByConversationId(conversationId: string): ConnectionEntity {
-    return this.connectionEntities().find(connectionEntity => connectionEntity.conversationId === conversationId);
+    const connectionEntities = Object.values(this.connectionEntities());
+    const match = connectionEntities.find(connectionEntity => connectionEntity.conversationId === conversationId);
+    return match ? this.getConnectionByUserId(match.userId) : undefined;
```
Yserz: Can the match be returned directly here?

lipis (approved):

src/script/connection/ConnectionRepository.ts
```diff
@@ -245,12 +248,9 @@ export class ConnectionRepository {
    * @returns Promise that resolves when all connections have been updated
    */
   private async updateConnections(connectionEntities: ConnectionEntity[]): Promise<ConnectionEntity[]> {
-    if (!connectionEntities.length) {
-      throw new ConnectionError(BaseError.TYPE.INVALID_PARAMETER, BaseError.MESSAGE.INVALID_PARAMETER);
-    }
-    this.connectionEntities.push(...connectionEntities);
+    connectionEntities.forEach(connectionEntity => this.addConnectionEntity(connectionEntity));
```
Yserz: That might trigger a lot of updates for knockout. WDYT about creating an object with the new connections, merge it with existing connections and then update the observable in one go? Alternatively this looks more like a `setAll` so we might not need an `update` and can just create an object in the shape of the oservable and push it in one go without respecting the connections already present in the observable.

src/script/connection/ConnectionRepository.ts
bennycode: No because it will return a copy and not the original observable. We need to guarantee identity here.

src/script/connection/ConnectionRepository.ts
bennycode: Good idea. How about this? https://github.com/wireapp/wire-webapp/pull/10061/commits/1f583106eed7142acf8ce6a1f0af456100c7a817

Yserz (approved):

PR #10063, opened 2020-11-27
fix: Don't fetch clients twice when checking for Legal Hold (SQCORE-239)

ffflorian:


src/script/conversation/ClientMismatchHandler.ts
```diff
@@ -55,15 +55,18 @@ export class ClientMismatchHandler {
     clientMismatch: ClientMismatch,
     payload?: NewOTRMessage,
   ): Promise<NewOTRMessage> {
-    const {deleted, missing, redundant} = clientMismatch;
+    const {deleted: deletedRecipients, missing: missingRecipients, redundant: redundantRecipients} = clientMismatch;
     // Note: Broadcast messages have an empty conversation ID
```
bennycode: I would keep the notice attached to the `eventInfoEntity.conversationId !== ''` check.

bennycode (approved):

PR #10088, opened 2020-11-30
refactor: Add missing types, fix JSDoc

ffflorian:


src/script/storage/EventRecord.ts
```diff
@@ -30,8 +32,8 @@ export interface EventRecord {
   conversation: string;
   data: any;
   edited_time?: string;
-  ephemeral_expires?: number;
-  ephemeral_started?: string;
+  ephemeral_expires?: boolean | string | number;
```
bennycode, line 42: Is it the event record than can be of type `boolean` or just the message entity which supports that? The record should only have one type. Let's try to fix that bug instead of widening the type.

bennycode (changes requested):

src/script/storage/EventRecord.ts
ffflorian: It supports `boolean`, `number` and `string`. See [`Message.ts`](https://github.com/wireapp/wire-webapp/blob/c9744c458c6525d348d0a6b9f39a34db94494e82/src/script/entity/message/Message.ts#L100-L112) (where you see that a message supports all three types) and [`ConversationEphemeralHandler.ts`](https://github.com/wireapp/wire-webapp/blob/c9744c458c6525d348d0a6b9f39a34db94494e82/src/script/conversation/ConversationEphemeralHandler.ts#L133-L134) (where you see that an `EventRecord` just takes the value).

bennycode (approved):

PR #10094, opened 2020-12-02
feat: Implement active speakers

thisisamir98:


src/script/components/calling/ParticipantMicOnIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const ParticipantMicOnIcon: React.FC<ParticipantMicOnIconProps> = ({
+  className,
+  isActive = false,
+  activeColor = '#fff',
+  ...props
+}) => {
+  return (
+    <span
+      css={css`
```
Yserz: No need to call extra `css` here. `css={{...}}` would be enough.

src/script/calling/CallingRepository.ts
```diff
@@ -1193,6 +1194,20 @@ export class CallingRepository {
     return this.mediaStreamQuery;
   };
 
+  private readonly updateActiveSpeakers = (wuser: number, conversationId: string, rawJson: string) => {
+    const call = this.findCall(conversationId);
+    if (!call) {
+      return;
+    }
+
+    const activeSpeakers = JSON.parse(rawJson);
```
Yserz, line 1203: `rawJson` is guaranteed to be at least `{"audio_levels": []}`, right?

src/script/components/calling/ParticipantMicOnIcon.tsx
thisisamir98: nice catch, https://github.com/wireapp/wire-webapp/pull/10094/commits/a4d957a9133cd531722a0027e338dc1060c57599

Yserz (approved):

src/script/calling/CallingRepository.ts
thisisamir98: avs said it's fine

PR #10106, opened 2020-12-04
refactor: Extract duration timer to separate component

thisisamir98:


src/script/components/calling/Duration.tsx
```diff
@@ -0,0 +1,29 @@
+import React, {useState, useEffect} from 'react';
```
ffflorian, line 20: Missing copyright header

ffflorian (changes requested):

src/script/components/calling/Duration.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    const props = {
+      startedAt: 0,
+    } as DurationProps;
+
+    const Duration = new DurationPage(props);
+
+    const now = Date.now();
+    jest.setSystemTime(now);
+    props.startedAt = Date.now();
+    Duration.setProps(props);
```
Yserz:
```suggestion
    const now = Date.now();
    const Duration = new DurationPage({
      startedAt: now,
    });

    jest.setSystemTime(now);
```

PR #10107, opened 2020-12-04
refactor: Remove unused return value from grant request

bennycode:


src/script/conversation/MessageRepository.ts
```diff
@@ -1239,7 +1239,7 @@ export class MessageRepository {
     const messageType = eventInfoEntity.getType();
     const allowedMessageTypes = ['cleared', 'clientAction', 'confirmation', 'deleted', 'lastRead'];
     if (allowedMessageTypes.includes(messageType)) {
-      return false;
+      return true;
```
Yserz: o.O what happened here? looks like a fixed bug! 🥳

bennycode: Yes and no. It would have been a bug but the return type actually is never read. I discovered that the function "grantOutgoingMessage" rejects if a message should not get through. Otherwise it just resolves. That's why I just did one more improvement: https://github.com/wireapp/wire-webapp/pull/10107/commits/7a15af0aaa7843a29c486e3083d6c4a1a6a98347 :)

src/script/conversation/MessageRepository.ts
```diff
@@ -1298,18 +1298,18 @@ export class MessageRepository {
       const shouldShowLegalHoldWarning =
         haveNewClientsChangeLegalHoldStatus && updatedLocalLegalHoldStatus === LegalHoldStatus.ENABLED;
 
-      return this.grantMessage(eventInfoEntity, consentType, userIds, shouldShowLegalHoldWarning);
+      await this.grantMessage(eventInfoEntity, consentType, userIds, shouldShowLegalHoldWarning);
     }
 
-    return this.grantMessage(eventInfoEntity, consentType, userIds, false);
+    await this.grantMessage(eventInfoEntity, consentType, userIds, false);
```
Yserz: Ah that's called twice for team users now.

bennycode: Ok, let me put a return after that statement.

Yserz (approved):

PR #10134, opened 2020-12-08
runfix: Fetch users if incoming message has no user IDs set

bennycode:


src/script/conversation/MessageRepository.ts
```diff
@@ -1336,7 +1336,10 @@ export class MessageRepository {
     return new Promise((resolve, reject) => {
       let sendAnyway = false;
 
-      userIds ||= conversationEntity.getUsersWithUnverifiedClients().map(userEntity => userEntity.id);
+      userIds =
+        userIds.length === 0
+          ? conversationEntity.getUsersWithUnverifiedClients().map(userEntity => userEntity.id)
+          : userIds;
```
ffflorian:
```suggestion
      if (userIds.length === 0) {
        userIds = conversationEntity.getUsersWithUnverifiedClients().map(userEntity => userEntity.id);
      }
```

ffflorian (changes requested):

PR #10232, opened 2021-01-04
refactor: Remove import shortcuts

ffflorian:
Reasons:
- The imports were inconsistent (e.g. why have a shortcut for "Util", but not for "Entity"?)
- We needed to keep two files (`tsconfig.json` and `jest.config.js`) in sync
- Auto-importing in VS Code doesn't use the shortcuts


ffflorian:
Too big for one PR

PR #10333, opened 2021-01-22
fix: Show same maximized video participant when switching to minimized view (SQCALL-34)

thisisamir98:


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -39,6 +39,7 @@ import type {Multitasking} from '../../notification/NotificationRepository';
 import Duration from './Duration';
 import NamedIcon from 'Components/NamedIcon';
 import ButtonGroup from 'Components/ButtonGroup';
+import type {Participant} from 'src/script/calling/Participant';
```
ffflorian, line 42:
```suggestion
import type {Participant} from './Participant';
```

ffflorian (approved):

src/script/components/calling/FullscreenVideoCall.tsx
thisisamir98: participant is not there, this is inside components/calling...

PR #10347, opened 2021-01-25
chore: Shorten PR build times

ffflorian:
Since we are neither building nor publishing from pull requests with `DISTRIBUTION != wire` we can also disable running `yarn` and `yarn test --coverage` in those action runs.

### DIFF
--- a/.github/workflows/test_build_deploy.yml
+++ b/.github/workflows/test_build_deploy.yml
@@ -74,17 +74,17 @@ jobs:
         uses: c-hive/gha-yarn-cache@v1
 
       - name: Authenticate git clone
-        if: env.DISTRIBUTION == 'wire' || github.event_name != 'pull_request'
+        if: env.DISTRIBUTION == 'wire' && github.event_name != 'pull_request'
         env:
           GH_TOKEN: ${{secrets.OTTO_THE_BOT_GH_TOKEN}}
         run: echo "machine github.com login ${GH_TOKEN}" > ~/.netrc
 
       - name: Install JS dependencies
-        if: env.DISTRIBUTION == 'wire' || github.event_name != 'pull_request'
+        if: env.DISTRIBUTION == 'wire' && github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/test_build_deploy.yml
```diff
@@ -74,15 +74,17 @@ jobs:
         uses: c-hive/gha-yarn-cache@v1
 
       - name: Authenticate git clone
+        if: env.DISTRIBUTION == 'wire' || github.event_name != 'pull_request'
         env:
           GH_TOKEN: ${{secrets.OTTO_THE_BOT_GH_TOKEN}}
-        run: echo -e "machine github.com\n  login ${GH_TOKEN}" > ~/.netrc
+        run: echo "machine github.com login ${GH_TOKEN}" > ~/.netrc
```
ffflorian, line 81: Next line is not needed, see https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html

.github/workflows/test_build_deploy.yml
```diff
@@ -74,15 +74,17 @@ jobs:
...
 
       - name: Install JS dependencies
+        if: env.DISTRIBUTION == 'wire' || github.event_name != 'pull_request'
         run: yarn --frozen-lockfile
 
       - name: Test
-        if: env.DISTRIBUTION == 'wire'
+        if: env.DISTRIBUTION == 'wire' || github.event_name != 'pull_request'
```
Yserz: No tests for PRs?

ffflorian: Why should we test everything twice in a PR? We are already testing with `DISTRIBUTION = wire` :)

Yserz: I don't get it what are we testing twice?

ffflorian: Resolved in 1:1

Yserz (approved):

PR #10350, opened 2021-01-25
 fix: Call controls overlay doesn't come up (SQCALL-133)

AndyLnd:


src/script/hooks/useHideElement.ts
```diff
@@ -1,57 +1,59 @@
...
+    };
+
+    const onMouseLeave = () => {
+      isMouseIn = false;
+      element.classList.add('hide-controls');
+    };
+
+    const startTimer = () => {
+      hideTimeout = window.setTimeout(() => {
+        element.classList.add('hide-controls');
```
bennycode: The `'hide-controls'` string is used 5x in this code. I suggest adding a `const` with that name and reusing this const.

bennycode (changes requested):

bennycode (approved):

PR #10395, opened 2021-02-02
refactor: Use @wireapp/core 17

ffflorian:


src/script/links/LinkPreviewRepository.ts
```diff
@@ -170,8 +170,8 @@ export class LinkPreviewRepository {
    * @param dataUri image data as base64 encoded data URI
    * @returns Resolves with the uploaded asset
    */
-  private async uploadPreviewImage(dataUri: string): Promise<Asset> {
-    const blob = await base64ToBlob(dataUri);
+  private uploadPreviewImage(dataUri: string): Promise<Asset> {
```
bennycode, line 173:
```suggestion
  private uploadPreviewImage(dataUri: string): Asset {
```

bennycode (changes requested):

src/script/links/LinkPreviewRepository.ts
ffflorian: `uploadFile` is returning a `Promise`, so the return type is correct here.

bennycode (approved):

PR #10436, opened 2021-02-04
refactor: Migrate device-card to react

thisisamir98:


src/script/components/DeviceCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export interface DeviceCardProps {
+  click?: (device: ClientEntity) => void;
+  device: ClientEntity;
+  showIcon?: boolean;
+  showVerified?: boolean;
+}
+
+const DeviceCard: React.FC<DeviceCardProps> = ({
+  click,
```
Yserz, line 39: Let's specify a noop click as default parameter here (noop exists in some util already)

src/script/components/DeviceCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+  device: wrappedDevice,
+  showVerified = false,
+  showIcon = false,
+}) => {
+  const clientEntity = wrappedDevice;
```
Yserz: Let's call it clientEntity directly

src/script/components/DeviceCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const {class: deviceClass = '?', id = '', label = '?', meta} = clientEntity;
+  const formattedId = id ? clientEntity.formatId() : [];
+  const name = clientEntity.getName();
+  const clickable = !!click;
+  const isVerified = meta.isVerified;
+  const showLegalHoldIcon = showIcon && deviceClass === ClientClassification.LEGAL_HOLD;
+  const showDesktopIcon = showIcon && deviceClass === ClientClassification.DESKTOP;
+  const showOtherIcon = showIcon && !showLegalHoldIcon && !showDesktopIcon;
```
Yserz: `showMobileIcon`

src/script/components/DeviceCard.tsx
thisisamir98: doing a noop here will break a part of this component that is rendering based on existence of this prop (that `noop` is a true value)
image

src/script/components/DeviceCard.tsx
```diff
@@ -0,0 +1,135 @@
...
+      )}
+      {showVerified && !isVerified() && (
+        <NamedIcon
+          width={14}
+          height={16}
+          className="not-verified-icon"
+          name="not-verified-icon"
+          data-uie-name="user-device-not-verified"
+        />
+      )}
```
Yserz: Might be a good candidate for a component

thisisamir98: TBH to me it does not look so much work to be a separate component

src/script/components/DeviceCard.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/10436/commits/a36edf0bf32b6ddd0f98816b08e300373fd92e76

src/script/components/DeviceCard.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/10436/commits/a36edf0bf32b6ddd0f98816b08e300373fd92e76

src/script/components/deviceCard.ts
```diff
@@ -1,111 +0,0 @@
...
-
-import ko from 'knockout';
-import {ClientClassification} from '@wireapp/api-client/src/client';
-
-import type {ClientEntity} from '../client/ClientEntity';
-
-interface DeviceCardParams {
-  click?: (device: ClientEntity) => void;
-  current?: boolean;
-  detailed?: boolean;
```
Yserz, line 28: What happened to those params? Are they not used anymore?

thisisamir98: nope, they are not used at all so I removed them.

src/script/components/DeviceCard.tsx
Yserz: We have various places where we want to show the verified icon in different states. So something like
```
<VerifiedIcon verified={isVerified()} />
```
would be nicer than repeating both icons over and over.

src/script/components/DeviceCard.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+class DeviceCardPage extends TestPage<DeviceCardProps> {
+  constructor(props?: DeviceCardProps) {
+    super(DeviceCard, props);
+  }
+
+  getDesktopIcon = () => this.get('svg[data-uie-name="status-desktop-device"]');
+  getMobileDeviceIcon = () => this.get('svg[data-uie-name="status-mobile-device"]');
+}
+
+describe('DeviceCard', () => {
```
Yserz, line 50: Let's add some more tests for:
it('shows disclose icon when component is clickable', ...)
it('shows verified icon', ...)
it('shows unverified icon', ...)

src/script/components/DeviceCard.tsx
AndyLnd: Or you could check like this:
```
const clickable = click !== noop;
```

src/script/components/DeviceCard.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/10436/commits/63ef9cca389f94481696d5b80c15e4185441a8fb

src/script/components/DeviceCard.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const formattedId = id ? clientEntity.formatId() : [];
+  const name = clientEntity.getName();
+  const clickable = !!click;
+  const isVerified = meta.isVerified;
+  const showLegalHoldIcon = showIcon && deviceClass === ClientClassification.LEGAL_HOLD;
+  const showDesktopIcon = showIcon && deviceClass === ClientClassification.DESKTOP;
+  const showMobileIcon = showIcon && !showLegalHoldIcon && !showDesktopIcon;
+
+  const clickOnDevice = () => {
+    if (typeof click === 'function') {
```
Yserz: This should be covered by TypeScript now

thisisamir98: it can be undefined so we have to check explicitly

src/script/components/DeviceCard.test.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/10436/commits/3ddd48e372f301af5a330484fe80fa8e184536af

Yserz (approved):

PR #10437, opened 2021-02-04
refactor: Migrate input level audiometer to React

bennycode:


src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const passedThreshold = level > bulletIndex / amountOfBullets.length;
+    if (passedThreshold) {
+      return 'input-level-bullet-active';
+    }
+
+    return '';
+  };
+
+  return (
+    <>
```
AndyLnd: Please use the `<ul>` as the wrapping element here, as we will still need it when we switch completely to react.

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+      ))}
+    </>
+  );
+};
+
+export default InputLevel;
+
+registerReactComponent('input-level', {
+  component: InputLevel,
+  template: '<ul class="input-level" data-bind="react: {disabled: ko.unwrap(disabled), level: ko.unwrap(level)}"></ul>',
```
AndyLnd: You can use `<div>` here an move the class to the `<ul>` in the jsx.

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+  level: number;
+}
+
+export const MAX_AUDIO_BULLETS = 20;
+
+/**
+ * Shows bullet indicators to visualize the audio input level.
+ *
+ * @param disabled - Don't show audio level indicator if set to `false`
+ * @param level - Audio input volume as floating point number, `1.0` is 100%
```
ffflorian:
```suggestion
 * @param disabled Don't show audio level indicator if set to `false`
 * @param level Audio input volume as floating point number, `1.0` is 100%
```

ffflorian (approved):

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+ */
+const InputLevel: React.FC<InputLevelProps> = ({disabled, level}) => {
+  const amountOfBullets = Array.from(Array(MAX_AUDIO_BULLETS).keys());
+
+  const isActiveBullet = (bulletIndex: number): string => {
```
AndyLnd:
```suggestion
  const getBulletClass = (bulletIndex: number): string => {
```

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+      return 'input-level-bullet-active';
+    }
+
+    return '';
+  };
+
+  return (
+    <>
+      {amountOfBullets.map(bulletIndex => (
+        <li key={bulletIndex} className={cx('input-level-bullet', isActiveBullet(bulletIndex))} />
```
AndyLnd:
```suggestion
        <li key={bulletIndex} className={cx('input-level-bullet', getBulletClass(bulletIndex))} />
```

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+ */
+const InputLevel: React.FC<InputLevelProps> = ({disabled, level}) => {
+  const amountOfBullets = Array.from(Array(MAX_AUDIO_BULLETS).keys());
+
+  const isActiveBullet = (bulletIndex: number): string => {
+    if (disabled) {
+      return 'input-level-bullet-disabled';
+    }
+
+    const passedThreshold = level > bulletIndex / amountOfBullets.length;
```
AndyLnd:
```suggestion
    const passedThreshold = level > bulletIndex / MAX_AUDIO_BULLETS;
```

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
thisisamir98, line 3:
```suggestion
 * Copyright (C) 2021 Wire Swiss GmbH
```

AndyLnd (changes requested):

thisisamir98 (approved):

src/script/components/InputLevel.tsx
thisisamir98: +1, my concern as well

src/script/components/InputLevel.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import TestPage from 'Util/test/TestPage';
+import InputLevel, {InputLevelProps, MAX_AUDIO_BULLETS} from 'Components/InputLevel';
+
+class InputLevelTestPage extends TestPage<InputLevelProps> {
+  constructor(props?: InputLevelProps) {
+    super(InputLevel, props);
+  }
+
+  getActiveInputLevelBullets = () => this.get('li.input-level-bullet-active');
```
Yserz: Let's decouple that from CSS and use data-uie-names

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+  disabled: boolean;
+  level: number;
+}
+
+export const MAX_AUDIO_BULLETS = 20;
+
+/**
+ * Shows bullet indicators to visualize the audio input level.
+ *
+ * @param disabled - Don't show audio level indicator if set to `false`
```
Yserz: Comment is misleading since the component is still shown.

AndyLnd:
Actually, can we change the use of `<ul>` and `<li>` to `<div>` instead. In the css we change the `<li>` to `display: block` either way. That way we don't need to use the `::before` pseudo selector. 
Oh, an can we change the naming scheme to BEM please? 😎

src/script/components/InputLevel.tsx
bennycode: Ok, I thought it's unnecessary wrapping.

src/script/components/InputLevel.test.tsx
bennycode: I don't like spamming the code with `data-uie-names` but I will do it this time and come up with a better solution next time.

src/script/components/InputLevel.tsx
bennycode: Indeed. Changed it. :)

src/script/components/InputLevel.test.tsx
Yserz: https://www.npmjs.com/package/babel-plugin-remove-jsx-attributes

Yserz: And then in `babel.config.js`
image

So these can be removed in the production build.

src/script/components/InputLevel.tsx
```diff
@@ -0,0 +1,67 @@
...
+  disabled: boolean;
+  level: number;
+}
+
+export const MAX_AUDIO_BULLETS = 20;
+
+/**
+ * Shows bullet indicators to visualize the audio input level.
+ *
+ * @param disabled Show audio meter with disabled bullets if set to `false`
```
Yserz:
```suggestion
 * @param disabled Show audio meter with disabled bullets if set to `true`
```

src/script/components/InputLevel.test.tsx
Yserz: I think there's nothing wrong in using `data-uie-name`s and `value`s and I don't see it as clutter. It makes it easy to select components, it makes it easy to see what is actually selected in a testcase, it doesn't break when you change the CSS and QA can use it as well.

Plus, how would you select a component that doesn't have a class or id attached?

bennycode:
@AndyLnd, let's do the naming scheme change and move from `ul` to `div` in another PR. The goal here is to move closer to React. I incorporated all the other feedback. For the sake of the Nightly run, I will merge it now. We both can fine tune it tomorrow.

src/script/components/InputLevel.test.tsx
bennycode: 1. I don't think that a selection with `li[data-uie-name="input-level-bullet-active"]` is easier than `li.input-level-bullet-active`. Both are easy to write.
1. It's more cluttered because `<li class="input-level-bullet-active">` now becomes `<li class="input-level-bullet-active" data-uie-name="input-level-bullet-active">`.
1. Our QA team doesn't need to select this component because we now have a test for it.
1. Instead, creating a `data-uie-name` can now implicitly mean that our QA team is using that component (even if they don't) which makes us afraid of changing its name or removing it.
1. Ofcourse the `data-uie-name` doesn't break when changing the CSS name but it breaks when changing the `data-uie-name` (which may be needed if the component gets renamed, i.e. from "input-level" to "audio-input-level").
1. Adding `data-uie-name` attributes to select components was an effort made to simplify the access for our QA team. Since our QA team doesn't need to access this component from an E2E test, we can use tools like [resq](https://github.com/baruchvlz/resq) to select a React component.

Yserz: But QA requested data-uie's for the exact same reason we are using them now. Decoupling from styling and classes and ids.

1. That's true for simple components but it makes a difference for complex ones.
2. Okay, but worth it IMHO
3. Having a unit test doesn't mean QA doesn't want to write an E2E test on top.
4. Good point, should we name our data attributes differently? (potentially even more attributes then)
5. Of course but changing CSS is more likely and that's why QA requested data-uies.
6. resq doesn't help because you I believe you can't transfer this object to enzyme.

I'm all up for a better solution or improvements but until then, data attributes seem to be state of the art.

Yserz: Ah damn, food got cold 😭

PR #10440, opened 2021-02-05
style: Use BEM in InputLevelComponent

AndyLnd:


lipis (approved):
`<div>` all the things

PR #10442, opened 2021-02-05
fix: Increase user handle limit to 256

bennycode:
Mobile clients allow up to 256 characters for the user handle so we should allow that same number.


lipis (approved):
Thanks for renaming it to `Handle` :)

PR #10443, opened 2021-02-05
chore: Fix attaching desktop builds to tags

ffflorian:


electron/package.json
```diff
@@ -7,9 +7,11 @@
   "devDependencies": {
     "@types/node": "12.19.16",
     "@wireapp/copy-config": "1.1.13",
+    "adm-zip": "0.5.2",
```
lipis, line 10: Why not your own: https://github.com/ffflorian/jszip-cli :)

lipis (approved):

.github/workflows/test_build_deploy.yml
```diff
@@ -421,24 +426,40 @@ jobs:
         working-directory: electron
         run: |
           echo "OS=macos" >> $GITHUB_ENV
-          yarn build:mac
+          yarn build:macos
+
+      - name: Zip Electron build
```
lipis:
```suggestion
      - name: Zip desktop build
```

lipis:
I would change all `Desktop` to `desktop`..

electron/package.json
```diff
@@ -34,7 +36,8 @@
     "dist": "yarn clear:ts && yarn build:ts",
     "postinstall": "copy-config",
     "prestart": "yarn dist",
-    "start": "electron . --no-sandbox"
+    "start": "electron . --no-sandbox",
+    "zip": "node bin/zip.js"
```
bennycode, line 40: Let's make use of `ts-node`.

bennycode (changes requested):

electron/package.json
ffflorian: I think it's too much overhead to use ts-node and a separate `tsconfig.json` for two files. Additionally, I am already using `// @ts-check`.

electron/package.json
ffflorian: Not the right use case. `jszip-cli` is supposed to be used as CLI.

PR #10455, opened 2021-02-08
runfix: Set default size for panel items to 16

ffflorian:


src/script/components/panel/PanelActions.tsx
```diff
@@ -37,10 +37,10 @@ export interface PanelActionsProps {
 const PanelActions: React.FC<PanelActionsProps> = ({items}) => {
   return (
     <Fragment>
-      {items.map(({click, identifier, icon, label, size = {}}) => (
+      {items.map(({click, identifier, icon, label, size}) => (
```
bennycode, line 40: How about this default: `size = {width: 16, height: 16}`?

bennycode (approved):

src/script/components/panel/PanelActions.tsx
ffflorian: But when you set only `width`, `height` will not be set.
(e.g. when `size` is set to `{width: 16}`)

PR #10456, opened 2021-02-08
runfix: Don't clear build directory before creating desktop installers

bennycode:
Our ASAR package is part of the `build` directory. When cleaning the `build` directory before calling `electron-builder`, we will get the following error:

image


bennycode:
`yarn dist` was missing

PR #10465, opened 2021-02-09
refactor: Use component's props as generic for optional params

ffflorian:


src/script/util/ComponentUtil.ts
```diff
@@ -20,30 +20,30 @@
...
-    optionalParams?: string[];
+    optionalParams?: (keyof Props)[];
     template: string;
   },
 ) {
   ko.components.register(name, {
-    template: template,
-    viewModel: function (knockoutParams: Record<string, any>) {
+    template,
+    viewModel: function (knockoutParams: Record<keyof Props, any>) {
```
Yserz: Would that also work?
```suggestion
    viewModel: function (knockoutParams: Props) {
```

Yserz (approved):

src/script/util/ComponentUtil.ts
ffflorian: Good idea!

PR #10474, opened 2021-02-11
refactor: Extract & Migrate CallMessage component to React

Yserz:


src/script/components/message/CallMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import NamedIcon from 'Components/NamedIcon';
+import React from 'react';
```
thisisamir98:
```suggestion
import React from 'react';
import NamedIcon from 'Components/NamedIcon';
```
I like to import global stuff first, but just my preference feel free to ignore

lipis (approved):

src/script/components/message/CallMessage.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+import TestPage from 'Util/test/TestPage';
+import {CallMessage as CallMessageEntity} from 'src/script/entity/message/CallMessage';
+import CallMessage, {CallMessageProps} from './CallMessage';
+
+class CallMessagePage extends TestPage<CallMessageProps> {
+  constructor(props?: CallMessageProps) {
+    super(CallMessage, props);
+  }
+
+  getCallMessage = (completed?: string) =>
```
bennycode:
```suggestion
  getCallMessage = (completed: 'completed' | 'not_completed') =>
```

I removed the optional because I don't see it being used without passing a string.

src/script/components/message/CallMessage.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const CallMessage: React.FC<CallMessageProps> = ({message}) => {
+  const isCompleted = message.was_completed();
+  const unsafeSenderName = message.unsafeSenderName();
+  const caption = message.caption();
+  const displayTimestampShort = message.displayTimestampShort();
+  const displayTimestampLong = message.displayTimestampLong();
+
+  // Equivalent for `ko.bindingHandlers.showAllTimestamps`
```
bennycode, line 37: How about removing the comment and putting it here? https://github.com/wireapp/wire-webapp/wiki/React-refactoring-goals-&-conventions#examples

bennycode (changes requested):

src/script/components/message/CallMessage.test.tsx
Yserz: The idea was to use the selector for 3 states. 
- Check that a call message in `completed` state exists
- Check that a call message in `not_completed` state exists
- And check that any call message exists

But can be change when needed.

src/script/components/message/CallMessage.tsx
Yserz: This is actually a thing that needs to be reconsidered when the whole `message.ts` is refactored to react because this is needed in every kind of message but the state is floating outside of the component.

(it makes all timestamps in the conversation visible when hovering just one)

bennycode (approved):

PR #10475, opened 2021-02-11
refactor: Migrate groupList to React

lipis:


thisisamir98 (approved):

src/script/components/GroupList.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <div className="search-list-item-header">{group.display_name()}</div>
+        </div>
+      ))}
+    </div>
+  );
+};
+
+export default GroupList;
+
+registerReactComponent<GroupListProps>('group-list', {
```
ffflorian:
```suggestion
registerReactComponent('group-list', {
```

src/script/components/GroupList.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <div
+          key={group.id}
+          data-uie-name="item-group"
+          className="search-list-item"
+          data-uie-uid={`${group.id}`}
+          onClick={() => {
+            click(group);
+            router.navigate(generateConversationUrl(group.id));
+          }}
+          data-uie-value={group.display_name()}
```
AndyLnd: It might be a good idea to have the single group item as a separate component and use the koSubscribable hook in them, as this seems to rely on observables inside the single groups (e.g. the display_name might change and it would not be reflected in the react component).

AndyLnd (changes requested):

AndyLnd (approved):

PR #10624, opened 2021-02-26
runfix: Show domain only when webapp allows it

bennycode:
Problem: Our staging backend currently always returns a `domain`, so this indicates to the webapp that the domain should be shown and that we are enabling federation. With the following PR we return the control to the webapp, so that the webapp decides if federated strings should be shown by setting an env var for `FEATURE_ENABLE_FEDERATION`.


lipis (approved):

thisisamir98 (approved):

server/ServerConfig.ts
```diff
@@ -19,6 +19,7 @@ export interface ServerConfig {
       ENABLE_ACCOUNT_REGISTRATION_ACCEPT_TERMS_AND_PRIVACY_POLICY: boolean;
       ENABLE_DEBUG: boolean;
       ENABLE_DOMAIN_DISCOVERY: boolean;
+      ENABLE_FEDERATION: boolean;
```
Yserz, line 22: At some point it might be interesting to have a separate section for `FEDERATION` outside of `FEATURE`.

kinda:
```
FEDERATION: {
  ENABLE_DOMAIN_DISCOVERY: boolean,
  FEDERATION_DOMAIN: string,
}
```

Yserz (approved):

src/script/entity/User.ts
```diff
@@ -213,7 +213,7 @@ export class User {
       /** Very old user accounts don't have a handle on Wire. */
       return '';
     }
-    return this.domain
+    return this.domain && Config.getConfig().FEATURE.ENABLE_FEDERATION
```
ffflorian, line 216: Won't this return a `boolean`?

ffflorian (changes requested):

ffflorian (approved):

server/ServerConfig.ts
bennycode: Yes, I am prefix all federation-related keys already with `FEDERATION_` (except `ENABLE_FEDERATION`) to identify these.

PR #10647, opened 2021-03-01
fix: Add missing sender's client ID

ffflorian:


bennycode (approved):

src/script/conversation/EventBuilder.ts
```diff
@@ -301,10 +303,15 @@ export const EventBuilder = {
   },
 
   buildMessageAdd(conversationEntity: Conversation, currentTimestamp: number): MessageAddEvent {
+    const selfUser = conversationEntity.selfUser();
+    const clientState = container.resolve(ClientState);
```
Yserz: Ideally that should be injected in an upper scope and not on every function call or it should be passed as param if possible. Idea here is to not overuse dependency injection.

Yserz (approved):

PR #10649, opened 2021-03-01
refactor: Extend logging for notification stream fetching

bennycode:


src/script/event/NotificationService.ts
```diff
@@ -176,7 +176,7 @@ export class NotificationService {
       message = await this.storageService.db.table(StorageSchemata.OBJECT_STORE.EVENTS).where({id: messageId}).first();
     }
 
-    const {notifications} = await this.apiClient.notification.api.getNotifications(clientId);
+    const notifications = await this.apiClient.notification.api.getAllNotifications(clientId);
```
ffflorian, line 179: What implications does this have? Seems like this is more than a `chore`, isn't it?

bennycode: Nope. It's a util function. The implication is that we get now all notifications instead of just the latest 1000.

ffflorian (approved):

PR #10736, opened 2021-03-19
refactor: Migrate EventRepository tests to TypeScript

bennycode:


src/script/event/EventRepositorySpec.ts
```diff
@@ -160,25 +163,25 @@ describe('EventRepository', () => {
     });
 
     it('should not update last notification id if transient is true', () => {
-      const notification_payload = {id: createRandomUuid(), payload: [], transient: true};
+      const notification_payload = {id: createRandomUuid(), payload: [], transient: true} as Notification;
```
Yserz, line 166: Suggestion: A const with `TypeUtil.RecursivePartial<Notification>` from commons might be a thing here. When the notification is used later this can be casted to `Notification`. The benefit here would be a partial type check of the `Notification` attributes that are used for the test.

src/script/event/EventRepositorySpec.ts
```diff
@@ -468,15 +491,19 @@ describe('EventRepository', () => {
 
     it('saves a text message with link preview with an ID previously used by the same user for a plain text message', () => {
       previously_stored_event = JSON.parse(JSON.stringify(event));
-      spyOn(testFactory.event_service, 'loadEvent').and.returnValue(Promise.resolve(previously_stored_event));
-      spyOn(testFactory.event_service, 'replaceEvent').and.returnValue(Promise.resolve(previously_stored_event));
+      jest
+        .spyOn(testFactory.event_service, 'loadEvent')
+        .mockImplementation(() => Promise.resolve(previously_stored_event));
```
Yserz: Suggestion: Can also be written as
```suggestion
        .mockImplementation(async () => previously_stored_event);
```

Yserz (approved):

PR #10810, opened 2021-03-30
runfix: Use child observables properly for user details and avatar

AndyLnd:


src/script/util/ComponentUtil.ts
```diff
@@ -69,3 +69,30 @@ export const useKoSubscribable = <T = any>(observable: ko.Subscribable<T>, defau
...
+  const getInitialState = (root: ParentOfObservables): ChildValues =>
+    children.reduce((acc, child) => {
+      acc[child] = root[child]();
+      return acc;
+    }, {} as ChildValues);
+
+  const [state, setState] = useState<ChildValues>(getInitialState(parent));
+  useEffect(() => {
+    setState(getInitialState(parent));
+    const subscriptions = [] as ko.Subscription[];
```
ffflorian:
```suggestion
    const subscriptions: ko.Subscription[] = [];
```

src/script/util/ComponentUtil.ts
```diff
@@ -69,3 +69,30 @@ export const useKoSubscribable = <T = any>(observable: ko.Subscribable<T>, defau
...
+    children.forEach(child => {
+      subscriptions.push(
```
ffflorian: Why not use a `map` here?

ffflorian (changes requested):

src/script/util/ComponentUtil.ts
AndyLnd: 🤦🏻‍♂️ Of course. I added the map later and didn't think of it at that point.

ffflorian: :)

thisisamir98 (approved):

ffflorian (approved):

PR #10924, opened 2021-04-15
test: Use ISO date when using `EventInfoEntity.setTimestamp()`

ffflorian:
before:
`new Date('1461926303002') -> Invalid Date`

after:
`new Date(1461926303002) -> 2016-04-29T10:38:23.002Z`

### DIFF
--- a/src/script/conversation/EventInfoEntity.ts
+++ b/src/script/conversation/EventInfoEntity.ts
@@ -62,8 +62,8 @@ export class EventInfoEntity {
     return this.type || (this.genericMessage && this.genericMessage.content);
   }
 
-  setTimestamp(time: string | number): void {
-    this.timestamp = new Date(Number(time)).getTime();
+  setTimestamp(isoDate: string): void {
+    this.timestamp = new Date(isoDate).getTime();
   }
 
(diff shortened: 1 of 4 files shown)


src/script/conversation/EventInfoEntity.ts
```diff
@@ -62,8 +62,8 @@ export class EventInfoEntity {
     return this.type || (this.genericMessage && this.genericMessage.content);
   }
 
-  setTimestamp(time: string): void {
-    this.timestamp = new Date(time).getTime();
+  setTimestamp(time: string | number): void {
```
bennycode: We should only allow strings or numbers here. If something else arrives, we need to solve the problem at the root.

bennycode (changes requested):

src/script/conversation/EventInfoEntity.ts
ffflorian: The problem is that the time arrives from the IndexDB where it is saved as a string.

ffflorian: Ah, I figured it out. The time is provided as ISO date string, the tests were just wrong.

bennycode (approved):

PR #10970, opened 2021-04-27
feat: Reload page if backend is unreachable on load (SQCORE-585)

ffflorian:


src/script/main/app.ts
```diff
@@ -543,19 +543,27 @@ class App {
         }
 
         default: {
-          this.logger.error(`Caused by: ${(error ? error.message : undefined) || error}`, error);
+          this.logger.warn(`Caused by: ${error?.message || 'unknown'}`, error);
+
+          const isNetworkError = error.name === 'NetworkError';
+          if (isNetworkError) {
+            this.logger.warn('Backend is not reachable. Reloading page.', error);
+            window.location.reload();
```
Yserz, line 551: As already discussed, this might DDoS our backend in case of an outage and the backend is coming up again.

PR #10982, opened 2021-04-29
refactor: Icon components

AndyLnd:


src/script/components/panel/PanelActions.tsx
```diff
@@ -40,7 +40,7 @@ const PanelActions: React.FC<PanelActionsProps> = ({items}) => {
       {items.map(({click, identifier, icon, label, size}) => (
```
Yserz: Can the `size` param be removed now?

Yserz (approved):

PR #11022, opened 2021-05-04
chore: CI change docker tag name scheme

lucendio:
*This change set aims to fix the naming scheme if container image tags with a focus on release versions.*

1. refactor: make script used to push Docker images a bit more readable

* remove dashes from values and add them inline when needed
* remove obsolete variables along the way

2. ci: change container image tag names to reflect

* before, image tags did not indicate actual releases
* before, `pkg.version` was a) meaningless and b) kept being static '0.2.0'

This way, it's a one-step lookup to the corresponding webapp release

If CI is triggered via release tag, the tags are, e.g.:

  * 2021-04-01-production.0-v0.28.3-254d51 - '${RELEASE_NAME}-${COMMIT_ID_SHORT}-${CONFIG_VERSION}'
  * edge - '${STAGE}'

If CI is triggered via dev or edge branch, the tag is, e.g.:

  * dev - '${STAGE}'

This allows to uniquely and readable pin the image version on one hand, and at
the same time use a dynamic tag to always fetch latest (e.g. `imagePullPolicy: Always`)

__NOTE:__ As a follow-up it is planned to create another step in CI to create a PR on https://github.com/wireapp/wire-server to update the container image tag in the Helm chart values whenever a new release is being created. This PR basically lays the groundwork for that.

/cc @Yserz

### DIFF
--- a/.github/workflows/test_build_deploy.yml
+++ b/.github/workflows/test_build_deploy.yml
@@ -239,9 +239,9 @@ jobs:
           DOCKER_USERNAME: ${{secrets.WEBTEAM_QUAY_USERNAME}}
         run: |
           if [ "$DISTRIBUTION" == "wire" ]; then
-            yarn docker '' staging
+            yarn docker '' staging "$TAG"
           else
-            yarn docker "$DISTRIBUTION" staging
+            yarn docker "$DISTRIBUTION" staging "$TAG"
           fi
 
       - name: Push production Docker image
@@ -251,9 +251,9 @@ jobs:
           DOCKER_USERNAME: ${{secrets.WEBTEAM_QUAY_USERNAME}}
         run: |
           if [ "$DISTRIBUTION" == "wire" ]; then
-            yarn docker '' production
+            yarn docker '' production "$TAG"
           else
-            yarn docker "$DISTRIBUTION" production
+            yarn docker "$DISTRIBUTION" production "$TAG"
           fi
 
       - name: Generate changelog
--- a/bin/push_docker.js
+++ b/bin/push_docker.js
@@ -32,16 +32,14 @@ try {
   currentBranch = execSync('git rev-parse HEAD').toString().trim();
 } catch (error) {}
 
-const distributionParam = process.argv[2];
-const stageParam = process.argv[3];
-const suffix = distributionParam ? `-${distributionParam}` : '';
-const stage = stageParam ? `-${stageParam}` : '';
+const distributionParam = process.argv[2] ? process.argv[2] : '';
+const stageParam = process.argv[3] ? process.argv[3] : '';
 const commitSha = process.env.GITHUB_SHA || 'COMMIT_ID';
 const commitShaLength = 7;
 const commitShortSha = commitSha.substring(0, commitShaLength - 1);
-const configurationEntry = suffix
-  ? `wire-web-config-default${suffix}`
-  : `wire-web-config-default${currentBranch === 'master' ? '-master' : '-staging'}`;
+const configurationEntry = distributionParam
+  ? `wire-web-config-default${distributionParam ? `-${distributionParam}` : ''}`
+  : `wire-web-config-default-${currentBranch === 'master' ? 'master' : 'staging'}`;
 const dependencies = {
   ...appConfigPkg.dependencies,
   ...appConfigPkg.devDependencies,
@@ -51,19 +49,21 @@ const dependencies = {
 
 const configVersion = dependencies[configurationEntry].split('#')[1];
 const dockerRegistryDomain = 'quay.io';
-const repository = `${dockerRegistryDomain}/wire/webapp${suffix}`;
+const repository = `${dockerRegistryDomain}/wire/webapp${distributionParam ? `-${distributionParam}` : ''}`;
 
 const dockerImageTag = `${repository}:${pkg.version}-${commitShortSha}-${configVersion}${stage}`;
-const dockerImageStageTag = stageParam ? `${repository}:${stageParam}` : '';
 
 const dockerCommands = [
   `echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin ${dockerRegistryDomain}`,
-  `docker build . -t ${dockerImageTag} ${dockerImageStageTag ? `-t ${dockerImageStageTag}` : ''}`,
+  `docker build . -t ${dockerImageTag}`,
   `docker push ${dockerImageTag}`,
 ];
 
-if (dockerImageStageTag) {
-  dockerCommands.push(`docker push ${dockerImageStageTag}`);
+if (stageParam) {
+  dockerCommands.push(...[
+    `docker tag ${dockerImageTag} ${repository}:${stageParam}`,
+    `docker push ${repository}:${stageParam}`
+  ]);
 }
 
 dockerCommands.push(`docker logout ${dockerRegistryDomain}`);
--- a/bin/push_docker.js
+++ b/bin/push_docker.js
@@ -64,10 +64,10 @@ const dockerCommands = [
   `docker build . --tag ${commitShortSha}`,
 ];
 
-tags.forEach( containerImageTag => {
+tags.forEach( containerImageTagName => {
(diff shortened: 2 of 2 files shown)


.github/workflows/test_build_deploy.yml
```diff
@@ -219,9 +219,8 @@ jobs:
           wait_for_deployment: false
           wait_for_environment_recovery: ${{env.DEPLOYMENT_RECOVERY_TIMEOUT_SECONDS}}
 
-      - name: Push master/dev/edge Docker image
+      - name: Push dev/edge Docker image
         if: |
-          env.BRANCH_NAME == 'master' ||
```
Yserz: Mh, I think that needs to stay because our dev builds are built on commit to dev. Our staging builds are built on commit to master and production builds are created when a tag is created.

bin/push_docker.js
```diff
@@ -51,20 +49,23 @@ const dependencies = {
...
 const configVersion = dependencies[configurationEntry].split('#')[1];
 const dockerRegistryDomain = 'quay.io';
-const repository = `${dockerRegistryDomain}/wire/webapp${suffix}`;
+const repository = `${dockerRegistryDomain}/wire/webapp${distributionParam ? `-${distributionParam}` : ''}`;
 
-const dockerImageTag = `${repository}:${pkg.version}-${commitShortSha}-${configVersion}${stage}`;
-const dockerImageStageTag = stageParam ? `${repository}:${stageParam}` : '';
+const tags = [];
+stageParam && tags.push(`${repository}:${stageParam}`);
+releaseParam && tags.push(`${repository}:${releaseParam}-${configVersion}-${commitShortSha}`);
```
Yserz: As discussed internally `pkg.version` is incremented by now. We are slowly transitioning to semver instead of a date based release version. But for now we can do whatever works best for you.

.github/workflows/test_build_deploy.yml
Yserz: Ah wait actually we create staging builds on tag as well. I confused it with another setup.

lucendio: You were right, but for different reasons :) - QA uses images tagged with those the branch names. So, I reverted the changes in that step.

bin/push_docker.js
lucendio: Agreed. As soon as the release name (or part of it) is reflected by `pkg.version`, we are going to adjust the image tag name, which is supposed to contain the release name.

Yserz (approved):

bin/push_docker.js
```diff
@@ -32,16 +31,15 @@ try {
   currentBranch = execSync('git rev-parse HEAD').toString().trim();
 } catch (error) {}
 
-const distributionParam = process.argv[2];
-const stageParam = process.argv[3];
-const suffix = distributionParam ? `-${distributionParam}` : '';
-const stage = stageParam ? `-${stageParam}` : '';
+const distributionParam = process.argv[2] ? process.argv[2] : '';
+const stageParam = process.argv[3] ? process.argv[3] : '';
+const releaseParam = process.argv[4] ? process.argv[4] : '';
```
ffflorian:
```suggestion
const distributionParam = process.argv[2] || '';
const stageParam = process.argv[3] || '';
const releaseParam = process.argv[4] || '';
```

bin/push_docker.js
```diff
@@ -32,16 +31,15 @@ try {
...
 const commitSha = process.env.GITHUB_SHA || 'COMMIT_ID';
 const commitShaLength = 7;
 const commitShortSha = commitSha.substring(0, commitShaLength - 1);
-const configurationEntry = suffix
-  ? `wire-web-config-default${suffix}`
-  : `wire-web-config-default${currentBranch === 'master' ? '-master' : '-staging'}`;
+const configurationEntry = distributionParam
+  ? `wire-web-config-default${distributionParam ? `-${distributionParam}` : ''}`
```
ffflorian: No need to double-check `distributionParam`

```suggestion
  ? `wire-web-config-default-${distributionParam}`
```

bin/push_docker.js
```diff
@@ -51,20 +49,24 @@ const dependencies = {
...
-  `docker build . -t ${dockerImageTag} ${dockerImageStageTag ? `-t ${dockerImageStageTag}` : ''}`,
-  `docker push ${dockerImageTag}`,
+  `docker build . --tag ${commitShortSha}`,
 ];
 
-if (dockerImageStageTag) {
-  dockerCommands.push(`docker push ${dockerImageStageTag}`);
-}
+tags.forEach(tag => {
+  dockerCommands.push(...[`docker tag ${commitShortSha} ${tag}`, `docker push ${tag}`]);
```
ffflorian:
```suggestion
  dockerCommands.push(`docker tag ${commitShortSha} ${tag}`, `docker push ${tag}`);
```

ffflorian (changes requested):

bin/push_docker.js
Yserz: When `distributionParam` is an empty string the result would be `wire-web-config-default-`.

ffflorian: But then this string wouldn't even be used. We are already checking if `distributionParam` is truthy in the line above.

Yserz: Ah right, got confused by the GH quick view.

lucendio: My own refactoring caught up wit me. thx!

bin/push_docker.js
```diff
@@ -51,20 +49,28 @@ const dependencies = {
...
   `echo "$DOCKER_PASSWORD" | docker login --username "$DOCKER_USERNAME" --password-stdin ${dockerRegistryDomain}`,
-  `docker build . -t ${dockerImageTag} ${dockerImageStageTag ? `-t ${dockerImageStageTag}` : ''}`,
-  `docker push ${dockerImageTag}`,
+  `docker build . --tag ${commitShortSha}`,
 ];
 
-if (dockerImageStageTag) {
-  dockerCommands.push(`docker push ${dockerImageStageTag}`);
-}
+// prettier-ignore
```
ffflorian: Why ignore prettier here?

ffflorian (changes requested):

bin/push_docker.js
lucendio: in order to be able to write the docker commands below each other. I find this more readable and natural to how scripts are being written and executed. [Prettier does not agree with me here](https://github.com/wireapp/wire-webapp/runs/2507970565?check_suite_focus=true#step:6:125).

ffflorian: Understandable, but I'd vote for prettier's approach here for the sake of consistency.

ffflorian (approved):

bin/push_docker.js
bennycode:
```suggestion
```

We don't accept `// prettier-ignore` statements in our PRs. 

[Prettier](https://prettier.io/), by its definition, is an opinionated code formatter and consequently it doesn't please everyone's style. That's okay because it has one big benefit: We don't lose time anymore by debating about coding styles in our projects and can invest the saved time into building new features and fixing existing bugs. Our projects also use auto formatting scripts, so that none of us wastes time indenting code. This helps us to focus on things that are really important by the end of the day. We are not going to start debating about styles and adding ignore rules here.

@lucendio, if you feel that Prettier's code styles doesn't feel "readable" and "natural", then please forward your concerns to [Prettier's issue tracker](https://github.com/prettier/prettier/issues).

bennycode (changes requested):

bin/push_docker.js
lucendio: > We don't accept // prettier-ignore statements in our PRs.

If that is the case, How come that an overall amount of 32 `// prettier-ignore` can be found in this very repository?

Anyway, linter is now happy. We shoud be good to go.

bennycode: @lucendio, I think those `// prettier-ignore` statements come from a time where we didn't enforce our rules consistently. We strive to get better at that.

Thanks for resolving the Prettier issue here. This protects us from 32 becoming 33. 👍

bennycode (approved):

PR #11067, opened 2021-05-10
chore: Auto-merge workflow for dependency updates

bennycode:


.github/auto-merge.yml
```diff
@@ -0,0 +1,3 @@
+- match:
+    dependency_type: all
+    update_type: 'semver:major'
```
ffflorian: Does that mean that major version updates will be merged, too? I'd vote against that since sometimes there are major updates with breaking changes which slip through the tests. So it's necessary to at least quickly have a look at the changelog.

.github/workflows/merge-dependencies.yml
```diff
@@ -0,0 +1,20 @@
...
+# https://github.com/ahmadnassri/action-dependabot-auto-merge/issues/60#issuecomment-806027389
+on: [pull_request_target]
+
+jobs:
+  auto-merge:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [12.x]
```
ffflorian: Matrix is not needed if it uses only one OS and one Node.js version

bennycode: But does it hurt having a matrix? It is more extensible.

.github/auto-merge.yml
bennycode: Ok, changed it to `semver:minor`.

.github/workflows/merge-dependencies.yml
ffflorian: Well, we never use a matrix if we don't need it and I find it confusing when there is a matrix without using it.

.github/workflows/merge-dependencies.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: 'Automerge dependency updates from Dependabot'
+        uses: ahmadnassri/action-dependabot-auto-merge@v2.4.0
+        # Guarantee that commit comes from Dependabot (don't blindly trust external GitHub Actions)
+        if: github.actor == 'dependabot[bot]'
```
ffflorian:
```suggestion
    # Guarantee that commit comes from Dependabot (don't blindly trust external GitHub Actions)
    if: github.actor == 'dependabot[bot]'
    steps:
      - name: 'Automerge dependency updates from Dependabot'
        uses: ahmadnassri/action-dependabot-auto-merge@v2.4.0
```

ffflorian (changes requested):

.github/workflows/merge-dependencies.yml
```diff
@@ -0,0 +1,16 @@
...
+  auto-merge:
+    runs-on: ubuntu-latest
+    # Guarantee that commit comes from Dependabot (don't blindly trust external GitHub Actions)
+    if: github.actor == 'dependabot[bot]'
+    steps:
+      - name: 'Automerge dependency updates from Dependabot'
+        uses: ahmadnassri/action-dependabot-auto-merge@v2.4.0
+        with:
+          github-token: ${{ secrets.GH_AUTOMERGE_TOKEN }}
+          node-version: 12.x
```
ffflorian: I don't think that works here, at least I don't see `node-version` as a property within [ahmadnassri/action-dependabot-auto-merge](https://github.com/ahmadnassri/action-dependabot-auto-merge)

ffflorian (changes requested):

PR #11071, opened 2021-05-10
fix: Improve UI performance of ephemeral messages

AndyLnd:


Yserz (approved):

ffflorian (approved):

src/page/template/modal/user-modal.htm
```diff
@@ -17,7 +17,7 @@
           params="user: user, actionsViewModel: actionsViewModel, onAction: onUserAction, isSelfActivated: isActivatedAccount()"
         ></user-actions>
       <!-- /ko -->
-      <!-- ko if: !user() && !userNotFound()  -->
+      <!-- ko if: isVisible() && !user() && !userNotFound()  -->
```
bennycode, line 20: Does it just hide the `loading-icon` SVG? If yes, can we completely remove it via JavaScript? Because it seems to be related to this report: https://github.com/wireapp/wire-desktop/issues/2507#issuecomment-833673808

AndyLnd: This is now there for exactly this reason. The `loading-icon` will not be in the DOM when it is not needed.

PR #11073, opened 2021-05-11
fix: Avoid leaking remote audio elements after a call (SQCALL-293)

c-g-owen:
Investigation into SQCALL-293 shows we are leaking the created audio elements for remote participants in conference calls, continuing to use CPU even after the call & accumulating CPU usage for each call until the tab is reloaded. The fix cleans up the call.audios list.

### DIFF
--- a/src/script/calling/Call.ts
+++ b/src/script/calling/Call.ts
@@ -100,7 +100,7 @@ export class Call {
   }
 
   removeAllAudio() {
-    Object.entries(this.audios).forEach(([audioId, a]) => {
+    Object.keys(this.audios).forEach(audioId => {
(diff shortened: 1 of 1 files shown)


src/script/calling/Call.ts
```diff
@@ -99,6 +99,12 @@ export class Call {
     delete this.audios[audioId];
   }
 
+  removeAllAudio() {
+    Object.entries(this.audios).forEach(([audioId, a]) => {
```
bennycode:
```suggestion
    Object.keys(this.audios).forEach(audioId => {
```

ffflorian (approved):

PR #11158, opened 2021-05-25
chore(ci): Add job to update image tag in corresponding Helm chart

lucendio:
This change-set is meant to keep the webapp Helm chart in wire-server
up-to-date. To accomplish that, it adds two new blocks of steps (one
of which actually incorporates a new job).

(1) following the naming scheme in ./bin/push_docker.js:59, a step
assembles the image tag name (not the release tag) - and so the
naming clash begins. A second one write the information to an 'artifact'.
This way, its possible to access the information from another job.

(2) this new job reads the image tag from the artifact and opens a
new PR over at wire-server. This PR is supposed to only update the
webapp chart - more precisely it only changes the value of 'image.tag'
in wire-server/charts/webapp/values.yaml

Finally, the artifact is removed by the general clean up job 'cleanup_artifacts'.

Furthermore, the change-set simplifies cleanup with the help of setting
'failOnError', which wont cause the workflow to fail, if a certain
artifact does not exist.

### DIFF
--- a/.github/workflows/test_build_deploy.yml
+++ b/.github/workflows/test_build_deploy.yml
@@ -117,6 +117,24 @@ jobs:
           env.BUILD_DESKTOP == 'true'
         run: yarn build:prod
 
+      - name: Assembling release information
+        if: contains(env.TAG, 'production') && matrix.DISTRIBUTION == 'DISTRIBUTION_0'
+        # NOTE: always using 'master' config, since release version is only consumed at
+        #       'production' stage for now
+        shell: bash
+        run: |
+          configVersion=$(cat ./app-config/package.json | jq -r '.dependencies["wire-web-config-default-master"]' | awk -F '#' '{ print $2 }')
+          packageVersion=$(cat ./package.json | jq -r '.version')
+          containerImageTag="${TAG:-${packageVersion}}-${configVersion}-${GITHUB_SHA::7}"
+          echo "{\"imageTag\": \"${containerImageTag}\", \"releaseName\": \"${TAG:-v${packageVersion}}\"}" > ./release-info.json
+
+      - name: Storing release information
+        if: contains(env.TAG, 'production') && matrix.DISTRIBUTION == 'DISTRIBUTION_0'
+        uses: actions/upload-artifact@v2
+        with:
+          name: release-info.json
+          path: ./release-info.json
+
       # https://wire-webapp-avs.zinfra.io/
       - name: Deploy avs build to Elastic Beanstalk
         if: env.BRANCH_NAME == 'avs' && matrix.DISTRIBUTION == 'DISTRIBUTION_0'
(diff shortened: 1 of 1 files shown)


.github/workflows/test_build_deploy.yml
```diff
@@ -323,6 +341,66 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    needs: test_build_deploy
+
+    steps:
+      - name: Obtaining release information artifact
+        id: release-info-artifact
+        uses: actions/download-artifact@v2
+        continue-on-error: true
+        with:
```
arianvp, line 355: I don't really follow why we go through the trouble of uploading and downloading an artifact to transfer this info. Why not set an `output` variable or an environment variable; and then just have  these steps follow immediately the steps  after line `120` ?    

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter

lucendio: FYI: we verified that step outputs are not accessible across jobs, hence the artifact approach. Rational being here to not put the PR creation into the build-test-deploy job. (As long as *deploying* in the web-team-context does not include `helm install/upgrade`).

.github/workflows/test_build_deploy.yml
```diff
@@ -450,24 +528,10 @@ jobs:
...
-        shell: bash
-        run: |
-          echo "BUILD_DESKTOP=${{contains(env.TAG, 'staging') || contains(env.TAG, 'production') || contains(env.PR_LAST_COMMIT_MESSAGE, '+Desktop')}}" >> $GITHUB_ENV
-
       - name: Delete WebApp artifacts
-        if: env.BUILD_DESKTOP == 'true'
         uses: geekyeggo/delete-artifact@v1
         with:
-          name: webapp-dist
+          failOnError: false
```
ffflorian, line 534: Good one!

.github/workflows/test_build_deploy.yml
```diff
@@ -323,6 +341,66 @@ jobs:
...
+          title: 'Update webapp version in Helm chart [skip ci]'
+          body: |
+            Image tag: `${{ fromJSON(steps.release-info-file.outputs.releaseInfo).imageTag }}`
+            Release: [`${{ fromJSON(steps.release-info-file.outputs.releaseInfo).releaseName }}`](${{ steps.change-helm-value.outputs.releaseUrl }})
+
+      - name: Printing Pull Request URL
+        if: ${{ steps.release-info-file.outputs.exists == 'true' }}
+        shell: bash
+        run: |
+          echo "PR: ${{ steps.create-pr.outputs.pull-request-url }}"
```
ffflorian, line 401:
```suggestion
        run: echo "PR: ${{ steps.create-pr.outputs.pull-request-url }}"
```

.github/workflows/test_build_deploy.yml
```diff
@@ -323,6 +341,66 @@ jobs:
...
+        with:
+          repository: 'wireapp/wire-server'
+          fetch-depth: 1
+
+      - name: Changing Helm value of the webapp chart
+        id: change-helm-value
+        if: ${{ steps.release-info-file.outputs.exists == 'true' }}
+        shell: bash
+        run: |
+          sed --in-place --expression="s/  tag: .*/  tag: \"${{ fromJSON(steps.release-info-file.outputs.releaseInfo).imageTag }}\"/" ./charts/webapp/values.yaml
```
ffflorian, line 378:
```suggestion
          sed -i -e "s/  tag: .*/  tag: \"${{ fromJSON(steps.release-info-file.outputs.releaseInfo).imageTag }}\"/" ./charts/webapp/values.yaml
```

ffflorian (approved):

arianvp (approved):

.github/workflows/test_build_deploy.yml
lucendio: using the short versions of the flag would decrease readability. I'll stick with the long ones :)

.github/workflows/test_build_deploy.yml
lucendio: Sadly, this wont work because of the colon after `PR:`. At least that is what my editor says. It might be a false positive though. I guess, there is no harm in keeping it that way.

.github/workflows/test_build_deploy.yml
ffflorian: Okay, but why are you using the long flags everywhere else then? ;)

PR #11161, opened 2021-05-27
fix: Check if errors are backend errors

bennycode:
Errors are by default of `any` type, so they can be `undefined` or of type `string` which will cause a check on `error.code` to fail.


Yserz (approved):

src/script/util/TypePredicateUtil.test.ts
```diff
@@ -0,0 +1,48 @@
+import {isAxiosError, isBackendError} from 'Util/TypePredicateUtil';
```
ffflorian, line 1: Missing copyright header

PR #11279, opened 2021-06-14
feat: Create a conversation with members from federated backends (SQCORE-773)

bennycode:


src/script/conversation/ConversationFilter.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export function isGroupWithFederatedUser(conversationEntity: Conversation, userEntity: User): boolean {
+  if (!conversationEntity.is1to1()) {
+    const [userId] = conversationEntity.participatingQualifiedUserIds();
+    return userEntity.id === userId?.id;
+  }
+  return false;
+}
+
+export function isInTeam(conversationEntity: Conversation, userEntity: User): boolean {
```
Yserz, line 47: looks like a duplicate of `isTeamConversation` https://github.com/wireapp/wire-webapp/pull/11279/files#diff-2ee7e1e0e8d5c3bf23bd4abe48ebba8253759a6006abfebaa1f56d5b307c0757R23

src/script/conversation/ConversationMapper.ts
```diff
@@ -316,8 +324,21 @@ export class ConversationMapper {
         const noOthers = !mergedConversation.others || !mergedConversation.others.length;
         if (isGroup || noOthers) {
           mergedConversation.others = othersStates
-            .filter(otherState => (otherState.status as number) === (ConversationStatus.CURRENT_MEMBER as number))
+            .filter(otherState => {
+              return (
+                (otherState.status as number) === (ConversationStatus.CURRENT_MEMBER as number) &&
```
Yserz, line 329: `!isRemovedFromConversation(otherState)`

src/script/conversation/ConversationFilter.ts
```diff
@@ -0,0 +1,49 @@
+/*
```
Yserz, line 1: WDYT about putting this into the `UserState` or `ConversationState`?

src/script/entity/User.ts
```diff
@@ -113,14 +113,15 @@ export class User {
     };
   }
 
-  constructor(id: string = '') {
+  constructor(id: string = '', domain: string | null) {
```
Yserz, line 116: Why do we put explicit `null`s instead of optional `domain` here?

bennycode: We wanted to identify all constructors which potentially need to supply a domain.

src/script/conversation/ConversationFilter.ts
bennycode: The `ConversationFilter` has only stateless functions so its functions are not a good fit for the `ConversationState` which is stateful.

src/script/conversation/ConversationFilter.ts
bennycode: The `isTeamConversation` checks if a conversation belongs to a team but `isInTeam` checks if a user is in the same team to which the conversation belongs to.

PR #11336, opened 2021-06-23
chore: Base tag information on script input rather than branch name

lucendio:
The container image tag information are now solely defined by the script input.
Due to the fact that branch information are not available during the release
process (because it indicated by a tag), it make sense to rather base the tag
determination on the stageParam, which is set i the respective pipeline step.

This also allows to further simplify the script assembling that tag and pushing
the images. Alongside, some variables have been renamed to clarify the meaning
of their values.
The pipelines uses 7 characters while the script only used 6 (second param of
substring behaves exclusive). This has been aligned.

*Owner: @arianvp*

### DIFF
--- a/bin/push_docker.js
+++ b/bin/push_docker.js
@@ -21,24 +21,16 @@
 
 const child = require('child_process');
 const appConfigPkg = require('../app-config/package.json');
-const {execSync} = require('child_process');
 
 require('dotenv').config();
 
-let currentBranch = '';
-
-try {
-  currentBranch = execSync('git rev-parse HEAD').toString().trim();
-} catch (error) {}
-
 const distributionParam = process.argv[2] || '';
 const stageParam = process.argv[3] || '';
-const releaseParam = process.argv[4] || '';
+const versionParam = process.argv[4] || '';
 const commitSha = process.env.GITHUB_SHA || 'COMMIT_ID';
-const commitShaLength = 7;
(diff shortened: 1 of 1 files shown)


arianvp (approved):
At first glance this looks ok to me

bin/push_docker.js
```diff
@@ -21,24 +21,16 @@
...
 const stageParam = process.argv[3] || '';
-const releaseParam = process.argv[4] || '';
+const versionParam = process.argv[4] || '';
 const commitSha = process.env.GITHUB_SHA || 'COMMIT_ID';
-const commitShaLength = 7;
-const commitShortSha = commitSha.substring(0, commitShaLength - 1);
+const commitShortSha = commitSha.substring(0, 7);
 const configurationEntry = `wire-web-config-default-${
-  distributionParam ? distributionParam : currentBranch === 'master' ? 'master' : 'staging'
+  distributionParam ? distributionParam : stageParam === 'production' ? 'master' : 'staging'
```
ffflorian:
```suggestion
  distributionParam || stageParam === 'production' ? 'master' : 'staging'
```

ffflorian (changes requested):

arianvp:
Thanks for merging ! I'll make sure everything is good once you push the new staging build and will report back if I still run into weird things.

PR #11337, opened 2021-06-24
runfix: Reset isMinimized on end call

AndyLnd:


src/script/components/calling/CallingOverlayContainer.tsx
```diff
@@ -77,11 +77,16 @@ const CallingContainer: React.FC<CallingContainerProps> = ({
     'state',
   ]);
   const currentCallState = state as CALL_STATE;
+
   useEffect(() => {
     if (currentCallState === CALL_STATE.MEDIA_ESTAB && joinedCall.initialType === CALL_TYPE.VIDEO) {
       multitasking.isMinimized(false);
     }
+    if (currentCallState === undefined) {
```
ffflorian, line 85:
```suggestion
    if (typeof currentCallState === 'undefined') {
```

ffflorian (changes requested):

ffflorian (approved):

src/script/components/calling/CallingOverlayContainer.tsx
AndyLnd: At this point I want to check the value of the variable. Using typeof would also return 'undefined' for a variable that is not declared. If that would be the case, the app should throw an error.

PR #11339, opened 2021-06-24
refactor: Migrate PingMessage to React

Yserz:


src/script/components/message/PingMessage.tsx
```diff
@@ -0,0 +1,88 @@
...
+      </div>
+    </div>
+  );
+};
+
+export default PingMessage;
+
+registerReactComponent('ping-message', {
+  component: PingMessage,
+  template:
```
thisisamir98: do you need an extra div? can't we use bindings here?

Yserz: Absolutely correct, I just got used to the old stuff :P

Yserz: 🦅

ffflorian (approved):

PR #11469, opened 2021-07-16
feat(services): Prevent conference calls based on feature config (SQSERVICES-562)

Yserz:


Yserz:
https://github.com/wireapp/wire-webapp/pull/11570

PR #11595, opened 2021-07-30
test: Adjust audio/video data-uies

Yserz:


src/script/components/asset/controls/SeekBar.tsx
```diff
@@ -23,6 +23,7 @@ import {clamp} from 'Util/NumberUtil';
 
 export interface SeekBarProps extends React.HTMLProps<HTMLDivElement> {
   dark?: boolean;
+  ['data-uie-name']?: string;
```
thisisamir98, line 26: isn't it better if we would just do dataUieName instead?

Yserz: Given that there's meaning behind `data-` attributes I would prefer to keep it like that: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

bennycode (approved):

thisisamir98 (approved):

PR #11624, opened 2021-08-04
feat: Display qualified users in federated conversation

ffflorian:


Yserz (approved):

src/script/entity/Conversation.ts
```diff
@@ -126,8 +127,12 @@ export class Conversation {
   public readonly messageTimer: ko.PureComputed<number>;
   public readonly name: ko.Observable<string>;
   public readonly notificationState: ko.PureComputed<number>;
+  /** Does not include self user */
   public readonly participating_user_ets: ko.ObservableArray<User>;
+  /** Does not include self user */
   public readonly participating_user_ids: ko.ObservableArray<string>;
+  /** Does not include self user */
```
bennycode:
```suggestion
  /** Contains only IDs from users with domains. Does not include self user. */
```

src/script/entity/Conversation.ts
```diff
@@ -126,8 +127,12 @@ export class Conversation {
   public readonly messageTimer: ko.PureComputed<number>;
   public readonly name: ko.Observable<string>;
   public readonly notificationState: ko.PureComputed<number>;
+  /** Does not include self user */
   public readonly participating_user_ets: ko.ObservableArray<User>;
+  /** Does not include self user */
```
bennycode:
```suggestion
  /** Contains only IDs from users without a domain. Does not include self user. */
```

src/script/entity/Conversation.ts
```diff
@@ -126,8 +127,12 @@ export class Conversation {
   public readonly messageTimer: ko.PureComputed<number>;
   public readonly name: ko.Observable<string>;
   public readonly notificationState: ko.PureComputed<number>;
+  /** Does not include self user */
```
bennycode:
```suggestion
  /** Includes all participating users (with & without domain) except the self user. */
```

src/script/entity/Conversation.ts
```diff
@@ -454,7 +464,7 @@ export class Conversation {
           return truncate(joinedNames, maxLength, false);
         }
 
-        const hasUserIds = !!this.participating_user_ids().length;
+        const hasUserIds = this.participating_user_ids().length + this.participatingQualifiedUserIds().length > 0;
```
bennycode: How about exposing this value (`this.participating_user_ids().length + this.participatingQualifiedUserIds().length`) in a getter for `Conversation.ts`?

bennycode (changes requested):

bennycode (approved):

PR #11632, opened 2021-08-06
runfix: Use string type for fingerprints (SQSERVICES-692)

AndyLnd:


src/script/view_model/content/LegalHoldModalViewModel.ts
```diff
@@ -134,7 +134,7 @@ export class LegalHoldModalViewModel {
           response.client.id,
           response.last_prekey,
         );
-        fingerprint = splitFingerprint(fingerprintString);
+        fingerprint = fingerprintString;
```
ffflorian: Just set `fingerprint` to the result in line 132.

ffflorian (changes requested):

src/script/view_model/content/LegalHoldModalViewModel.ts
AndyLnd: 🤦‍♀️ of course

ffflorian (approved):

PR #11667, opened 2021-08-10
runfix: Redirect to login when app config can not be loaded (SQSERVICES-710)

Yserz:
image

### DIFF
--- a/src/script/team/TeamService.ts
+++ b/src/script/team/TeamService.ts
@@ -67,6 +67,7 @@ export class TeamService {
   }
 
   getAllTeamFeatures(): Promise<FeatureList> {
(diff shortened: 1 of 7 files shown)


src/script/team/TeamService.ts
```diff
@@ -67,6 +67,7 @@ export class TeamService {
   }
 
   getAllTeamFeatures(): Promise<FeatureList> {
+    throw new Error('kaputt');
```
thisisamir98: Broken?

Yserz: Ooopsie ;P That was for testing the error :P

bennycode (approved):

PR #11673, opened 2021-08-11
refactor: Use String.replaceAll()

ffflorian:


AndyLnd (approved):

ffflorian:
`String.replaceAll()` is only supported with Node.js >=15 :roll_eyes:

PR #11675, opened 2021-08-12
feat: Display qualified users in federated conversation

bennycode:


bennycode:
Will be done with https://github.com/wireapp/wire-webapp/pull/11743

PR #11745, opened 2021-08-25
runfix: Allow maximizing tile only with more than one participants

thisisamir98:


src/script/components/calling/GroupVideoGrid.tsx
```diff
@@ -93,7 +93,9 @@ const GroupVideoGrid: React.FunctionComponent<GroupVideoGripProps> = ({
       return;
     }
     const participant = grid.grid.find(participant => participant?.doesMatchIds(userId, clientId));
-    setMaximizedParticipant(participant);
```
AndyLnd, line 96: You can return earlier. you could just change line 92 to 
```
if (grid.grid.length < 2) {
```

AndyLnd (changes requested):

src/script/components/calling/GroupVideoGrid.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/11745/commits/5d7d78140e1bec4fe8a446ca9cb6f7239035d707

AndyLnd (approved):

PR #11776, opened 2021-08-31
refactor(core): Fix missing conversation receipt mode typings

ffflorian:
Follow-up to https://github.com/wireapp/wire-webapp/pull/11748.

### DIFF
--- a/src/script/components/toggle/ReceiptModeToggle.test.ts
+++ b/src/script/components/toggle/ReceiptModeToggle.test.ts
@@ -17,7 +17,7 @@
  *
  */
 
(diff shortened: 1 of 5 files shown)


thisisamir98 (approved):

bennycode:
We have to be careful here! There is a receipt mode (on conversation level) and there are confirmation types (on message level). Every read receipt settings, that is not bound to group conversations, makes use of the confirmation type. Are you sure @ffflorian that your changes are just related to receipt modes on conversation-level?

ffflorian:
@bennycode Yes, I am pretty confident.

PR #11928, opened 2021-09-29
feat(core): Remove call to old conversations endpoint for federated backends

atomrc:
As discussed with backend, there is no need to query the old converation endpoint AND the new one. The new one already includes both local and remote conversations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "dependencies": {
     "@wireapp/antiscroll-2": "1.0.2",
     "@wireapp/avs": "7.2.91",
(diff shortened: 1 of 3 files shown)


bennycode (approved):
Thanks for changing. How about renaming `getRemoteConversations` in the `api-client` package because it doesn't only return the remote conversations?

atomrc:
> Thanks for changing. How about renaming `getRemoteConversations` in the `api-client` package because it doesn't only return the remote conversations?

yes, I was actually thinking about it :)

PR #11930, opened 2021-09-29
runfix(core): Use conversation entity as source of truth when muting/unmuting

atomrc:


src/script/conversation/ConversationRepository.ts
```diff
@@ -2392,8 +2392,12 @@ export class ConversationRepository {
    * @param eventJson JSON data of 'conversation.member-update' event
    * @returns Resolves when the event was handled
    */
-  private onMemberUpdate(conversationEntity: Conversation, eventJson: Partial<ConversationMemberUpdateEvent>) {
-    const {qualified_conversation: conversationId, data: eventData, from} = eventJson;
+  private onMemberUpdate(
+    conversationEntity: Conversation,
+    eventJson: Pick<ConversationMemberUpdateEvent, 'data' | 'from'>,
+  ) {
+    const {data: eventData, from} = eventJson;
```
atomrc, line 2399: We already had the bug before, but it was silent (since `conversation` was a string and not an object as opposed to qualified conversation id). So `undefined` was not a problem before but has become with the qualifiedId being accessed

AndyLnd (approved):

atomrc:
Closed in favor of https://github.com/wireapp/wire-webapp/pull/11941

PR #11963, opened 2021-10-05
[draft] typed event emitter lib

atomrc:


src/script/app/eventemitter.ts
```diff
@@ -0,0 +1,8 @@
+import mitt from 'mitt';
```
atomrc, line 1: This is the file that is responsible for initializing the event emitter and configure all the possible events and payload types.

src/script/app/eventemitter.ts
```diff
@@ -0,0 +1,8 @@
...
+import type {QualifiedId} from '@wireapp/api-client/src/user/';
+
+type Events = {
+  'conversation.delete': QualifiedId;
```
atomrc, line 8: We probably do not need a huge object to define all the possible events, we can stay with string identifiers since now TS checks them. Open for discussion

src/script/team/TeamRepository.ts
```diff
@@ -363,7 +364,7 @@ export class TeamRepository {
     const {
       data: {conv: conversationId},
     } = eventJson;
-    amplify.publish(WebAppEvents.CONVERSATION.DELETE, conversationId);
+    emitter.emit('conversation.delete', {domain: '', id: conversationId});
```
atomrc, line 367: Here I already found a problematic event. Thanks TS :)

atomrc:
We are slowly trying to get rid of amplify. 
Once we have cleaned all the place where we use it and be sure those left are absolutely needed, then we could migrate to a different event emitter solution

PR #12017, opened 2021-10-12
runfix(core): Fix how we check for federated env

atomrc:


src/script/conversation/ConversationService.ts
```diff
@@ -57,6 +57,10 @@ import {ConversationRecord} from '../storage/record/ConversationRecord';
 import {Config} from '../Config';
 import {QualifiedId} from '@wireapp/api-client/src/user';
 
+function isFederatedEnv() {
```
atomrc, line 60: I'll probably generalize this helper somewhere in a util in a followup PR (so that we don't do the mistake another time :) )

Yserz: This could be a simple value instead of a function as the config gets read on page load and is supposed to be readonly from there on.

atomrc: Yeah true that !

Yserz (approved):

PR #12022, opened 2021-10-14
chore(wire-server): Include changelog entry

jschaul:
We recently added a changelog process on wire-server. This adds a file on the automatic PRs with a changelog description that is then picked up with the rest of our current process. This is for the helm charts of the webapp, also currently located inside the wire-server repo. These are used as the defaults for on-premise installations.

The same would need to be done on the similar code active on team-settings and account pages.


atomrc (approved):

jschaul:
> PR Title Check / check (pull_request)

What needs to be done to pass this check? @atomrc could you change the title of the PR according to your guidelines and merge the PR? (I don't know what the processes in web look like). Thanks!

PR #12043, opened 2021-10-19
refactor(core): Use core to send pings

atomrc:
This would be the very first message we send with the core in the webapp. 
This means that both federated and regular env will use the core to send those messages. Which also means we need to initiate the core for every client.

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -27,7 +27,6 @@ import {
   Ephemeral,
   External,
   GenericMessage,
-  Knock,
   LastRead,
   LegalHoldStatus,
   MessageDelete,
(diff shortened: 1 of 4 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -223,31 +222,17 @@ export class MessageRepository {
...
-    const protoKnock = new Knock({
-      [PROTO_MESSAGE_TYPE.EXPECTS_READ_CONFIRMATION]: this.expectReadReceipt(conversationEntity),
-      [PROTO_MESSAGE_TYPE.LEGAL_HOLD_STATUS]: conversationEntity.legalHoldStatus(),
-      hotKnock: false,
-    });
-
-    let genericMessage = new GenericMessage({
-      [GENERIC_MESSAGE_TYPE.KNOCK]: protoKnock,
-      messageId: createRandomUuid(),
+  public async sendPing(conversation: Conversation): Promise<void> {
```
AndyLnd, line 225: 💕

AndyLnd (approved):

PR #12056, opened 2021-10-20
runfix(core): Improve payload of conversation creation api call

atomrc:


src/script/conversation/ConversationRepository.ts
```diff
@@ -320,25 +320,22 @@ export class ConversationRepository {
...
-      .filter(userEntity => !userEntity.isOnSameFederatedDomain())
-      .map(userEntity => userEntity.qualifiedId);
+    const isFederated = Config.getConfig().FEATURE.ENABLE_FEDERATION && Config.getConfig().FEATURE.FEDERATION_DOMAIN;
+    const userIds = isFederated ? userEntities.map(user => user.qualifiedId) : userEntities.map(user => user.id);
+
     let payload: NewConversation & {conversation_role: string} = {
       conversation_role: DefaultRole.WIRE_MEMBER,
       name: groupName,
-      qualified_users: otherFederatedDomainUserIds,
+      qualified_users: isFederated ? (userIds as QualifiedId[]) : undefined,
```
atomrc, line 329: Checked with backend, we do not need to split federated/non-federated users (as we are currently doing) to create a conversation. 
On a federated env, we can just send the `qualified_users` field, no need for the `users` field at all

AndyLnd (approved):

PR #12123, opened 2021-11-04
refactor(core): cleanup CryptographyRepository

atomrc:


src/script/cryptography/CryptographyRepository.ts
```diff
@@ -113,27 +107,12 @@ export class CryptographyRepository {
     });
   }
 
-  /**
-   * Generate all keys needed for client registration.
-   * @returns Resolves with an array of last resort key, pre-keys, and signaling keys
-   */
-  async generateClientKeys(): Promise<ClientKeys> {
```
atomrc, line 120: This is not used anywhere anymore

atomrc: since b6ca4dd0b

src/script/cryptography/CryptographyRepository.ts
```diff
@@ -457,9 +425,7 @@ export class CryptographyRepository {
     const cipherText = cipherTextArray.buffer;
     const sessionId = constructClientPrimaryKey(userId, eventData.sender);
 
-    const plaintext = isFederatedEnv
-      ? await this.core.service.cryptography.cryptobox.decrypt(sessionId, cipherText)
```
atomrc, line 446: Now that the cryptobox in the repo is the same as the one in the core, no need to differenciate

rohan-wire: I'm fine with this comment, but please keep in mind that things will be different when we have OpneMLS and cryptobox in use at the same time (for different conversations).

atomrc: Hey Rohan,

Yes, all the PR we create regarding federation are made with that fact in mind. 
We are currently in the process of moving message sending to our own web `core` lib. Which means that when MLS is here, we just have to integrate it into our web `core` to enable it in the webapp and all the tools that QA uses to send messages (namely the `ETS`). 

Once we have a company's `corelib` then integrating it into our web `core` is gonna be quite easy.

In the meantime I feel it's important that we remove code that could generate confusion. On this line, a developer could ask himself the question "Why do we need a condition here, does it mean we have 2 cryptoboxes in the webapp?". I believe it's important to prevent those confusions to happen.

AndyLnd (approved):

PR #12169, opened 2021-11-16
fix(core): Add domain to user search requests [FS-225]

atomrc:
----

- The **PR Title**
  - [x] answers the question: _If merged, this PR will: ..._ . E.g. `feature(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

- The **PR Description**
----

# What's new in this PR?

### Issues
When searching for federated users, we send the username and the domain in the `query` param of the search API.
We should split them and send the domain as a different param.

### Dependencies (Optional)

- [ ] https://github.com/wireapp/wire-web-packages/pull/4180

### DIFF
--- a/src/script/search/SearchRepository.ts
+++ b/src/script/search/SearchRepository.ts
@@ -169,8 +169,7 @@ export class SearchRepository {
     isHandle?: boolean,
     maxResults = SearchRepository.CONFIG.MAX_SEARCH_RESULTS,
   ): Promise<User[]> {
-    const domain = query.includes('@') ? query.substr(query.lastIndexOf('@') + 1) : undefined;
-    const name = domain ? query.substr(0, query.indexOf('@')) : query;
+    const [name, domain] = query.replace(/^@/, '').split('@');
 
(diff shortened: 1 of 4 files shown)


src/script/search/SearchRepository.ts
```diff
@@ -169,16 +169,16 @@ export class SearchRepository {
     isHandle?: boolean,
     maxResults = SearchRepository.CONFIG.MAX_SEARCH_RESULTS,
   ): Promise<User[]> {
+    const domain = query.includes('@') ? query.substr(query.lastIndexOf('@') + 1) : undefined;
+    const name = domain ? query.substr(0, query.indexOf('@')) : query;
```
AndyLnd:
```suggestion
   const [name, domain] = query.split('@');
```
i am pretty sure this would do the same in a more readable manner. WDYT?

AndyLnd (changes requested):

src/script/search/SearchRepository.ts
atomrc: It should indeed, let me run a few tests

AndyLnd (approved):

PR #12241, opened 2021-12-01
runfix(core): Send calling messages to other self devices

atomrc:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This was added with a misconception in mind. We actually need to send the calling message to the other clients but not inject the message into the self client

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -1875,7 +1875,6 @@ export class MessageRepository {
     return this.sendAndInjectGenericCoreMessage(message, conversation, {
       ...options,
       skipInjection: true,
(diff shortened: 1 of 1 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -1875,7 +1875,6 @@ export class MessageRepository {
     return this.sendAndInjectGenericCoreMessage(message, conversation, {
       ...options,
       skipInjection: true,
-      skipSelf: true, // We never want to forward calling messages to the self user
```
atomrc, line 1878: This was a misunderstanding. The trick was to no inject in the selfClient (see `skipInjection`) not to prevent sending to the other devices.

AndyLnd (approved):

PR #12297, opened 2021-12-14
runfix(core): Fix tweet link previews

atomrc:


src/script/components/asset/LinkPreviewAssetComponent.tsx
```diff
@@ -41,10 +40,10 @@ export interface LinkPreviewAssetProps {
 const LinkPreviewAssetComponent: React.FC<LinkPreviewAssetProps> = ({header = false, message}) => {
   const [preview] = (message.getFirstAsset() as Text).previews();
 
-  const isTypeTweet = preview?.meta_data_type === LinkPreviewMetaDataType.TWEET;
+  const isTypeTweet = preview?.tweet;
```
vovchenso: I think `isTypeTweet` supposed to be a boolean const. And `tweet` is object (or undefined). Maybe make sense to cast it to boolean type here with `!!`:
```
const isTypeTweet = !!preview?.tweet;
```

atomrc: Yeah, that makes sense, let's be consistent :)  Will update

vovchenso (approved):

PR #12317, opened 2021-12-16
fix(core): Properly warn user when calling in a degraded conversation

atomrc:


src/script/calling/CallingRepository.ts
```diff
@@ -303,9 +303,14 @@ export class CallingRepository {
         data.map(clientid => ({clientid, userid: this.serializeQualifiedId(userId)})),
       ),
     );
-    this.wCall.setClientsForConv(this.wUser, this.serializeQualifiedId(conversationId), JSON.stringify({clients}));
+    this.wCall.setClientsForConv(this.wUser, this.serializeQualifiedId(call.conversationId), JSON.stringify({clients}));
     // We warn the message repository that a mismatch has happened outside of its lifecycle (eventually triggering a conversation degradation)
-    this.messageRepository.handleClientMismatch(conversationId, {missing} as ClientMismatch);
+    const consentType =
```
atomrc, line 308: This shows the correct message to the user according to the call direction

src/script/calling/CallingRepository.ts
```diff
@@ -303,9 +303,14 @@ export class CallingRepository {
...
+      this.getCallDirection(call) === CALL_DIRECTION.INCOMING ? CONSENT_TYPE.INCOMING_CALL : CONSENT_TYPE.OUTGOING_CALL;
+    return (
+      !checkMismatch ||
+      this.messageRepository.handleClientMismatch(call.conversationId, {missing} as ClientMismatch, consentType)
```
AndyLnd:
```suggestion
      checkMismatch ? 
      this.messageRepository.handleClientMismatch(call.conversationId, {missing} as ClientMismatch, consentType) :
      true
```

It took me a while to parse this. I think the ternary would be easier to digest.

AndyLnd (changes requested):

AndyLnd (approved):

PR #12428, opened 2022-01-21
fix: Disable applock when feature is not enforced in the team anymore (SQSERVICES-1064)

Yserz:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1064" title="SQSERVICES-1064" target="_blank">BugSQSERVICES-1064</a>  [Web] When mandatory App Lock is disabled (after being enforced), you are still required to enter passphrase to unlock
  </td></table>
  <br />
 

Disables applock for the user when it is not enforced for the team anymore. It respects the previous local setting of the user in that it is still enabled when the user actively enabled applock locally before it was enforced on the team level.

### DIFF
--- a/src/script/page/AppLock.tsx
+++ b/src/script/page/AppLock.tsx
@@ -149,8 +149,10 @@ const AppLock: React.FC<AppLockProps> = ({
   }, [state, isVisible]);
 
   const showAppLock = () => {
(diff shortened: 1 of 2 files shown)


Yserz:
Can be merged after the next webapp release

tlebon (approved):

vovchenso (approved):

PR #12450, opened 2022-01-25
refactor(core): Extract client mismatch logic to a standalone module

atomrc:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Refactors the logic for handling mismatch and triggering conversation degradation or legal hold detection.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -189,11 +189,13 @@ export class ConversationRepository {
       deletedClients.forEach(({userId, clients}) =>
         clients.forEach(client => this.userRepository.removeClientFromUser(userId, client)),
       );
-      const removedTeamUserIds = emptyUsers.filter(user => user.inTeam()).map(user => user.id);
+      const removedTeamUserIds = emptyUsers.filter(user => user.inTeam()).map(user => user.qualifiedId);
 
       if (removedTeamUserIds.length) {
         // If we have found some users that were removed from the conversation, we need to check if those users were also completely removed from the team
-        const usersWithoutClients = await this.userRepository.getUserListFromBackend(removedTeamUserIds);
+        const usersWithoutClients = await this.userRepository.getUserListFromBackend(
+          conversation.isFederated() ? removedTeamUserIds : removedTeamUserIds.map(({id}) => id),
+        );
         await Promise.all(
           usersWithoutClients
             .filter(user => user.deleted)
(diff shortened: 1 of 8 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -767,25 +738,6 @@ export class MessageRepository {
       this.updateMessageAsSent(conversation, genericMessage.messageId, syncTimestamp ? sentTime : undefined);
     };
 
-    /**
-     * Once a message has been sent, check the state of the members of this conversation.
-     * If, in this conversation, there are still users that have 0 devices, that could mean they have been silently removed from the team
-     * We need to ask backend if those users have been deleted and, if so, trigger the teamMemberLeave event
-     */
-    const checkMissingTeamMembers = async (): Promise<void> => {
```
atomrc, line 787: This is now handled by the mismatch handler

src/script/conversation/MessageRepository.ts
```diff
@@ -1468,80 +1414,14 @@ export class MessageRepository {
...
-        } else {
-          await Promise.all(
-            Object.entries(usersMap).map(([userId, clients]) =>
-              Promise.all(
-                clients.map(client => this.userRepository.addClientToUser({domain: '', id: userId}, client, false)),
-              ),
-            ),
-          );
-        }
-      }
```
atomrc, line 1544: This all logic is also handled by the mismatch handler

vovchenso (approved):

src/script/conversation/ConversationRepository.ts
```diff
@@ -179,85 +173,66 @@ export class ConversationRepository {
...
+          usersWithoutClients
+            .filter(user => user.deleted)
+            .map(user =>
+              this.teamMemberLeave(
+                this.teamState.team().id,
+                {
+                  domain: this.teamState.teamDomain(),
+                  id: user.id,
+                },
+                new Date(mismatch.time).getTime() - 1,
```
vovchenso, line 211: Just for curiosity.. why we use `.getTime() - 1` in some places in this file?

atomrc: This is in order to inject the message before the message being sent. 
This is a not so nice hack to have message ordered the right way. 

In a not so distant future, I'd like to have a message buffer that takes care of putting the right timestamp to system messages, but for now we are stuck with this system

vovchenso: got it thanks for explanation

PR #12454, opened 2022-01-25
refactor(core): Send countly message via @wireapp/core

atomrc:


src/script/conversation/MessageRepository.ts
```diff
@@ -1583,30 +1582,8 @@ export class MessageRepository {
...
-    const genericMessage = new GenericMessage({
-      [GENERIC_MESSAGE_TYPE.DATA_TRANSFER]: protoDataTransfer,
-      messageId: createRandomUuid(),
-    });
-
-    const eventInfoEntity = new EventInfoEntity(genericMessage, selfConversation.qualifiedId);
-    try {
-      await this.sendGenericMessageToConversation(eventInfoEntity);
-      this.logger.info(`Sent countly sync message with ID ${countlyId}`);
-    } catch (error) {
```
tlebon, line 1480: wondering if we should still try catch an error (in case one appears)

atomrc: This is the strategy I applied here:
There could be a backend errors, in which case no message sending is going to go through anyway and the webapp is not going to work properly. This case should be handled in very high levels of the app and reflected in the UI. 
And there could be a logic/coding mistake somewhere, in which case I think it's good that the error is loud and clear and bubbles up rather than being hidden in a log. 

In general I try to catch error in rather high layers of the app, not too deep, but I am open to discussions about the right place to catch errors :)

tlebon: this is fine for me (and i appreciate hearing your thinking on the subject), just curious as the old version had the error handling

tlebon (approved):

PR #12456, opened 2022-01-26
refactor(core): Improve types for the MessageRepository [FS-418]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-418" title="FS-418" target="_blank">TaskFS-418</a>  [Web] cleanup of legacy created by M2 sprint
  </td></table>
  <br />

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -25,7 +25,6 @@ import {
   GenericMessage,
   LastRead,
   LegalHoldStatus,
-  Asset as ProtobufAsset,
   DataTransfer,
 } from '@wireapp/protocol-messaging';
 import {
@@ -138,15 +137,21 @@ type ClientMismatchHandlerFn = (
   consentType?: CONSENT_TYPE,
 ) => Promise<boolean>;
 
+/** A Quote that is meant to be sent in a message (and thus needs to have a valid hash) */
+type OutgoingQuote = QuoteEntity & {hash: Uint8Array};
+
 type TextMessagePayload = {
   conversation: Conversation;
   linkPreview?: LinkPreviewUploadedContent;
   mentions?: MentionEntity[];
   message: string;
   messageId?: string;
-  quote?: QuoteEntity;
+  quote?: OutgoingQuote;
 };
 type EditMessagePayload = TextMessagePayload & {originalMessageId: string};
+
+/** A message that has already been stored in DB and has a primary key */
+type StoredMessage = Message & {primary_key: string};
 
 export class MessageRepository {
   private readonly logger: Logger;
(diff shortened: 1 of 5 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -1709,7 +1712,7 @@ export class MessageRepository {
     let messageType = eventInfoEntity.getType();
 
     if (messageType === GENERIC_MESSAGE_TYPE.CONFIRMATION) {
-      messageType += ` (type: "${eventInfoEntity.genericMessage.confirmation.type}")`;
+      messageType += ` (type: "${eventInfoEntity.genericMessage.confirmation?.type}")`;
```
vovchenso, line 1716: Will it be ok if `confirmation` will be `undefined`? I mean in this case it will produce:
```
 (type: "undefined")
```

tlebon: ah nice catch, maybe it should cast to some value in that case

atomrc: It's no biggies since it's just a log, I didn't touch the logic at all in there. Theoretically it will never be undefined but if at some point it is, I think it's a good thing we see it in the log

src/script/conversation/MessageRepository.ts
```diff
@@ -138,16 +137,23 @@ type ClientMismatchHandlerFn = (
   consentType?: CONSENT_TYPE,
 ) => Promise<boolean>;
 
+/** A Quote that is meant to be sent in a message (and thus needs to have a valid hash) */
+type OutgoingQuote = QuoteEntity & {hash: Uint8Array};
```
tlebon: wondering if this should be exported and reused since it appears again InputBarViewModel

src/script/conversation/MessageRepository.ts
```diff
@@ -630,7 +643,7 @@ export class MessageRepository {
     return this.sendAndInjectGenericCoreMessage(payload, conversation);
   }
 
-  private isUserCancellationError(error: ConversationError): boolean {
+  private isUserCancellationError(error: any): boolean {
```
tlebon, line 646: why the decision to switch to an any type here? just curious

atomrc: That's a typescript constraint: Error are of type `unknown` and cannot be cast into a proper type. So you have  to cast it to `any` before you can do anything with it. 
 As of now the webapp compiles in non-strict mode, so TS doesn't complain about this, but in strict mode it doesn't allow an error to be cast.

src/script/conversation/MessageRepository.ts
atomrc: Yep, that's good call :+1:

vovchenso (approved):

PR #12459, opened 2022-01-26
fix(core): Avoid displaying domain twice in the Account preferences [SQSERVICES-1245]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1245" title="SQSERVICES-1245" target="_blank">BugSQSERVICES-1245</a>  WEB: Domain name is showing twice on profile section 
  </td></table>
  <br />
 

Since domain is already handled by our richField hook, we do not need to manually add it in the profile

### DIFF
--- a/src/script/page/preferences/AccountPreferences.tsx
+++ b/src/script/page/preferences/AccountPreferences.tsx
@@ -168,14 +168,6 @@ const AccountPreferences: React.FC<AccountPreferencesProps> = ({
             {isTeam && (
               <AccountInput label={t('preferencesAccountTeam')} value={teamName} readOnly data-uie-name="status-team" />
             )}
-            {isFederated && (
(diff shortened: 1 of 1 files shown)


src/script/page/preferences/AccountPreferences.tsx
```diff
@@ -168,14 +168,6 @@ const AccountPreferences: React.FC<AccountPreferencesProps> = ({
             {isTeam && (
               <AccountInput label={t('preferencesAccountTeam')} value={teamName} readOnly data-uie-name="status-team" />
             )}
-            {isFederated && (
```
atomrc, line 171: This field is already added by the `useEnrichedFields` hook above in this file

tlebon (approved):

PR #12486, opened 2022-02-01
feat(core): Restore message sending queueing [FS-418]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-418" title="FS-418" target="_blank">TaskFS-418</a>  [Web] cleanup of legacy created by M2 sprint
  </td></table>
  <br />
 

Message queue was lost when messages started being sent with @wireapp/core. Queueing messages will allow loadtime to increase (avoiding confirmation messages to be sent while the app is loading)


tlebon (approved):
seems fine

PR #12582, opened 2022-02-22
feat(core): Initialize APIClient with supported version [FS-423]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-423" title="FS-423" target="_blank">TaskFS-423</a>  [Web] Clients need to know if the backend support federated endpoints or not
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

This allows setting the supported version and let the APIClient decide the common version to use between the client and the backend. 

Now the condition to use federation or not is completely opaque to the consumer (the webapp) and we can remove the `ENABLE_FEDERATION` flag (it's now deduces by the api client depending on the version used)


src/script/conversation/ConversationRepository.ts
```diff
@@ -378,15 +376,14 @@ export class ConversationRepository {
...
+    const userIds = userEntities.map(user => user.qualifiedId);
 
     let payload: NewConversation & {conversation_role: string} = {
       conversation_role: DefaultRole.WIRE_MEMBER,
       name: groupName,
-      qualified_users: isFederated ? (userIds as QualifiedId[]) : undefined,
+      qualified_users: userIds,
       receipt_mode: null,
-      users: !isFederated ? (userIds as string[]) : [],
+      users: userIds.map(({id}) => id),
```
tlebon, line 386: cant you just map all of the userIds on line 379? 

`const userIds = userEntities.map(user => user.qualifiedId.id);`

tlebon: maybe im missing something

src/script/conversation/ConversationRepository.ts
```diff
@@ -1344,11 +1341,7 @@ export class ConversationRepository {
     const userIds = userEntities.map(userEntity => userEntity.qualifiedId);
 
     try {
-      const response = await this.conversation_service.postMembers(
-        conversationEntity.id,
-        userIds,
-        conversationEntity.isFederated(),
-      );
+      const response = await this.conversation_service.postMembers(conversationEntity.id, userIds);
```
tlebon, line 1344: obviously not something you did in this 1500 line file, but im wondering if conversation_service should be conversationService

src/script/user/UserRepository.ts
```diff
@@ -482,10 +459,8 @@ export class UserRepository {
       }
     };
 
-    const chunksOfUserIds = chunk<string | QualifiedId>(userIds, Config.getConfig().MAXIMUM_USERS_PER_REQUEST);
-    const resolveArray = await Promise.all(
-      chunksOfUserIds.map(userChunk => getUsers(userChunk as string[] | QualifiedId[])),
-    );
+    const chunksOfUserIds = chunk<QualifiedId>(userIds, Config.getConfig().MAXIMUM_USERS_PER_REQUEST);
+    const resolveArray = await Promise.all(chunksOfUserIds.map(userChunk => getUsers(userChunk)));
```
tlebon: does this work also? 

```suggestion
    const resolveArray = await Promise.all(chunksOfUserIds.map(getUsers));
```

tlebon (approved):
i left a few suggestions but i think it looks fine

src/script/conversation/ConversationRepository.ts
atomrc: I am using it as `QualifiedId[]` on line `384`. So I need both `QualifiedId[]` and `string[]` in the payload

src/script/conversation/ConversationRepository.ts
atomrc: Yeah, that's a bad call from me, actually. When I set up the core library in there, I aliased `core.service.conversation` to `conversationService`... Not realizing that the ConversationRepo already had a `conversation_service` (that will slowly be deprecated). I'll try to rename it in another PR, probably

src/script/user/UserRepository.ts
atomrc: LGTM

PR #12670, opened 2022-03-15
refactor(core): Migrate archive list to React

atomrc:
Migrates the archive list to React,
Also introduce the new `LeftListWrapper` component that is used to display both the preferences list and the archive list


src/script/page/preferences/Preferences.tsx
```diff
@@ -88,42 +105,28 @@ const Preferences: React.FC<PreferencesProps> = ({listViewModel, contentViewMode
...
-                  isSelected={contentState === item.id}
-                  uieName={item.uieName}
-                  IconComponent={item.IconComponent}
-                />
-              ))}
-          </ul>
-        </div>
-      </div>
-    </Transition>
+    <LeftListWrapper
```
atomrc, line 108: The wrapping elements for all the list on the left of the app are common,
So I extracted the left list logic to put it in the wrapper. We can use this component when migrating more lists to React

src/script/components/list/ConversationListCell.tsx
```diff
@@ -42,9 +42,9 @@ export interface ConversationListCellProps {
   conversation: Conversation;
   dataUieName?: string;
   index: number;
-  is_selected: (conversation: Conversation) => boolean;
```
tlebon, line 45: camelCase?

atomrc: Ahaha fair point :) 
But since this is a parameter that is already there and can be instantiated in ko templates, I won't have any typecheck there and I have to admit I'd rather not take that risk :upside_down_face:

tlebon (approved):

PR #12671, opened 2022-03-15
runfix(core): Fix displaying domain in conversation detail

atomrc:


src/script/view_model/PanelViewModel.ts
```diff
@@ -132,8 +133,10 @@ export class PanelViewModel {
     const conversationState = container.resolve(ConversationState);
 
     this.elementId = 'right-column';
+
     this.repositories = repositories;
     this.mainViewModel = mainViewModel;
+    this.isFederated = mainViewModel.isFederated;
```
atomrc, line 138: The bug was here: the view was trying to access a `isFederated` property which was `undefined`.

BartDyb (approved):

PR #12676, opened 2022-03-17
refactor(core): Migrate home conversations list to React

atomrc:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Migrates the conversation list to React

### Notes

This PR introduces 2 new concepts in the codebase:
- A state management tool (here we use https://github.com/pmndrs/zustand). See file https://github.com/wireapp/wire-webapp/pull/12676/files#diff-977472cfadb919e358405a4c5962b3eb2ac6e7fd717fab93e07769d9c2b9d9dc
- A typed EventTarget class (for communication between the GroupedConversationList and the label repository). It a good pattern to have in general and will unload our global usage of `amplify` in the future (see https://github.com/wireapp/wire-webapp/pull/12676/files#diff-506356865baee629dbb327f8c327e2653eb6aece885a3451ee2032a34ec69b22R83)


src/script/components/list/ConversationListCell.tsx
```diff
@@ -123,153 +111,144 @@ const ConversationListCell: React.FC<ConversationListCellProps> = ({
...
-            ) : (
-              <span className={cx('conversation-list-cell-name', {'accent-text': isSelected})}>{displayName}</span>
-            )}
-            <span className="conversation-list-cell-description" data-uie-name="secondary-line">
-              {cellState.description}
-            </span>
-          </div>
-          <div className="conversation-list-cell-right">
+    <li ref={setViewportElementRef}>
+      <div
```
tlebon, line 115: if this is role button maybe it should actually be a button element

src/script/components/list/ConversationListCell.tsx
```diff
@@ -123,153 +111,144 @@ const ConversationListCell: React.FC<ConversationListCellProps> = ({
...
+                >
+                  <Icon.Hangup className="svg-icon" />
+                </span>
+              )}
+              {cellState.icon === ConversationStatusIcon.MUTED && (
+                <span
+                  className="conversation-list-cell-badge cell-badge-dark conversation-muted"
+                  data-uie-name="status-silence"
+                  title={t('callStatusMutted')}
+                  aria-label={t('callStatusMutted')}
```
tlebon: typo, also update the translation string
```suggestion
                  title={t('callStatusMuted')}
                  aria-label={t('callStatusMuted')}
```

src/script/components/list/ConversationListCell.tsx
kamil-au: This one is correct as it has a div inside's do should stay like it is

kamil-au (approved):

src/script/components/list/ConversationListCell.tsx
atomrc: Since this is not a new key, I'd rather not change it. It might have side effects for the translations (like in crowding). I'm not sure

kamil-au: It's a new key.  I've added this one in the previous PR. I think it's safe to update this one, as translations to other languages were not processed. Sorry for the typo :(

atomrc: Oohhh right, then I'll update it :ok_hand:

atomrc: Fixed :)

src/script/util/TypedEventTarget.ts
```diff
@@ -0,0 +1,32 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export class TypedEventTarget<EventDef extends {type: any}> extends EventTarget {
+  public dispatch(event: EventDef): boolean {
+    return this.dispatchEvent(new Event(event.type));
```
tlebon, line 22: `this` in this case refers to eventTarget? this is one of those times I find `this` to be a little ambiguous

tlebon (approved):
LGTM!

src/script/util/TypedEventTarget.ts
atomrc: Yeah, it refers to the native class. Would you do something different? 
Notice that the children class method is `dispatch` VS the native method is `dispatchEvent`, maybe that's the source of the confusion?

tlebon: i think that is the source of the confusion for me, I don't know if I have any other solution though, I think its fine.

thisisamir98 (approved):

PR #12678, opened 2022-03-17
refactor(accessibility): Use button for conversation list cells

atomrc:


src/script/components/list/ConversationListCell.tsx
```diff
@@ -94,7 +93,7 @@ const ConversationListCell: React.FC<ConversationListCellProps> = ({
   const cellBottom = cellTop + cellHeight;
 
   const isInitiallyVisible = isVisibleFunc(cellTop, cellBottom);
-  const [viewportElementRef, setViewportElementRef] = useEffectRef<HTMLDivElement>();
+  const [viewportElementRef, setViewportElementRef] = useEffectRef<HTMLElement>();
```
BartDyb, line 96: why not `HTMLLiElement`?

atomrc: It's because I don't need that level of specificity. Would work with any kind of element `div` `li` or any other.

BartDyb: ahh, ok - it is used by the `useViewPortObserver` hook - it makes sense

atomrc:
Closed because there are too many problems with using a button here. 
The main one being: we cannot have `div` inside a button element and some components there are divs.

PR #12680, opened 2022-03-18
fix(accessibility): Video/Audio call fullscreen view #ACC17

kamil-au:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Focusable buttons controlled by the keyboard. Added focus trap inside fullscreen view and unhide buttons on tab key pressed not only on mouse move.


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -176,6 +178,27 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
 
   const totalPages = callPages.length;
 
+  const isSpaceOrEnterKey = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    return event.key === KEY.ENTER || event.key === KEY.SPACE;
```
atomrc, line 182: Why do you not use the helper functions in the KeyboardUtil (like `isSpaceKey`)?

atomrc (approved):

src/script/components/calling/FullscreenVideoCall.tsx
kamil-au: Removed this anyway as it's not necessary here. About KeyboardUtill there we are using a regular `KeyboardEvent` as arguments and didn't want to mix it up.

PR #12704, opened 2022-03-22
refactor(core): replace deprecated String.prototype.substr()

CommanderRoot:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### Solutions

I replaced all usages of `substr()` with `slice()` or `substring()`

### Testing

I have tested the return of each function to make sure it's the same as the current behaviour

----

----

----
##### References


Gruntfile.js
```diff
@@ -93,7 +93,7 @@ module.exports = grunt => {
 
     let user = grunt.config('gitinfo.local.branch.current.currentUser');
     if (user) {
-      user = user.substr(0, user.indexOf(' ')).toLowerCase();
+      user = user.substring(0, user.indexOf(' ')).toLowerCase();
```
tlebon, line 96: im not clear why we are using substring some places and slice some places.

CommanderRoot: `substr()`, `slice()` and `substring()` are very similar but act differently in [some situations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice), especially if the second parameter is negative.
If the second parameter is normally positve but can be negative you most likely want to use substring() as slice() would produce a different behaviour than substr().

Example:
```
"0123456789".substr(0, -1) == ""
"0123456789".slice(0, -1) == "012345678"
"0123456789".substring(0, -1) == ""
```

Sometimes you want the `slice()` behaviour but not in the cases in the code here.

It would be possible to use `slice()` everywhere if you add a check, like

`      user = user.substring(0, user.indexOf(' ') > 0 ? user.indexOf(' ') : 0).toLowerCase();`

Using substring() does look a bit nicer though if you don't mind mixing the two.

ffflorian:
Hi @CommanderRoot, thanks for the PR! I started the checks on it.
In general the PR looks good and I am in favor of merging it.

ffflorian:
@CommanderRoot looks like there are some linting errors. Can you merge from `dev` again?

Gruntfile.js
tlebon: ah sorry didnt read this close enough didn't realize the second param was indexOf, makes perfect sense then.

ffflorian (approved):
LGTM

atomrc (approved):

atomrc:
super seeded and merged here https://github.com/wireapp/wire-webapp/pull/12887 
Thanks @CommanderRoot for your contribution

PR #12782, opened 2022-04-04
runfix: Bypass auto unwrapping of registerStaticReactComponent in user-list

thisisamir98:


src/script/components/UserList.tsx
```diff
@@ -333,7 +330,7 @@ const UserList: React.FC<UserListProps> = ({
             .map(user => (
               <ParticipantItem
                 key={user.id}
-                noInteraction={noSelfInteraction && user.isMe}
+                noInteraction={!(noSelfInteraction && user.isMe)}
```
atomrc, line 334: why did you invert those?

src/script/components/UserList.tsx
```diff
@@ -52,44 +52,38 @@ export enum UserlistMode {
...
   conversationRepository: ConversationRepository;
   conversationState?: ConversationState;
-  filter: string;
   highlightedUsers: () => User[];
   infos: Record<string, string>;
   maxVisibleUsers: number;
   mode: UserlistMode;
   noSelfInteraction: boolean;
   noUnderline: boolean;
+  observers: {filter: ko.Observable<string>; users: ko.ObservableArray<User>};
```
atomrc: Let's rather call it `observables` rather than `observers`. 
Also `selected` can be put in this `observables` object :)

atomrc (changes requested):

src/script/components/UserList.tsx
thisisamir98: very complicated logic coming from old knockout codebase

src/script/components/UserList.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/12782/commits/3b2d685a2d98c9d3738fc1152b121df81e59f74a

atomrc (approved):

PR #12784, opened 2022-04-05
runfix: Correctly pass users to user-list

thisisamir98:


src/script/components/UserList.tsx
```diff
@@ -111,7 +111,7 @@ const UserList: React.FC<UserListProps> = ({
   const selfInTeam = userState.self().inTeam();
   const {self} = useKoSubscribableChildren(userState, ['self']);
   const {
-    users: userEntities,
+    users: userEntities = [],
```
atomrc: Hummm it feels weird to have a default value for the thing it's supposed to display? What justifies to use this component without a user list?

thisisamir98: Actually that can go, main issue was in the htm files https://github.com/wireapp/wire-webapp/pull/12784/commits/d868804aac5e2c58a2c797e5a802e47d1e2ab267

atomrc (approved):

PR #12858, opened 2022-04-19
chore: remove unused and outdated dependency react-intl-po

comawill:
react-intl-po was never used within wire-webapp and is also unmaintained for ~4 years.

It is safe to remove the dependency and we get rid of several other old dependencies.


atomrc (approved):
:ok_hand:  good catch

PR #12861, opened 2022-04-19
chore: Configure dependabot to upgrade github actions

atomrc:


.github/dependabot.yml
```diff
@@ -53,23 +58,9 @@ updates:
       - dependency-name: '@types/node'
         versions:
           - '>= 13.a'
-  - package-ecosystem: npm
-    directory: '/electron'
```
atomrc, line 57: This was the old `/electron` build that has be removed in #12860

PR #12890, opened 2022-04-20
chore(deps): replace wrtc with @koush/wrtc

comawill:
This PR replaces `wrtc` with `@koush/wrtc` to add support for tests on macOS/arm64.


atomrc (approved):
Nice, thanks. ping @thisisamir98 you are going to like this one ;)

PR #12894, opened 2022-04-21
refactor: Optimize ko-react components

atomrc:
This harmonizes and simplified the way to instantiate a react component inside ko.


src/script/util/ComponentUtil.ts
```diff
@@ -20,30 +20,62 @@
...
+
+/**
+ * Registers a react component against the ko world.
  *
  * @param name Name of the component to register. can be used a `<component-name>` directly in ko
- * @param {component}
+ * @param component The React component to register
  */
-export function registerStaticReactComponent<Props>(name: string, component: React.ComponentType<Props>) {
+export function registerReactComponent<Props>(name: string, component: React.ComponentType<Props>) {
```
atomrc, line 78: This is the main change of this PR. 
It removes the `registerStaticReactComponent` and replaces it with a single optimized `registerReactComponent`. 

It both handled regular component (that do not have observables as direct input) and ko dynamic components (with observables as direct inputs).

thisisamir98 (approved):

PR #12921, opened 2022-04-26
chore: Simplify semantic commit linting

atomrc:


.github/workflows/semantic-commit-lint.yml
```diff
@@ -1,7 +1,7 @@
 name: 'Semantic Commit Linting of PR titles'
 
 on:
-  pull_request:
+  pull_request_target:
```
atomrc, line 4: This will fix the errors when dependabot is creating PR (this creates a more secure context, that has also write property)
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

.github/workflows/semantic-commit-lint.yml
```diff
@@ -10,19 +10,13 @@ permissions:
 jobs:
   semantic-commit-pr-title-lint:
     runs-on: ubuntu-latest
-    env:
```
atomrc, line 13: For now, I'll remove the label, as we are not using it yet and slows down the feedback loop (as we need to pull the repo first). 
We can think about restoring it if we have a workflow that actually uses the label in the future

tlebon (approved):

PR #12929, opened 2022-04-27
runfix: message list correct key

kamil-au:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### DIFF
--- a/src/script/components/MessagesList/Message/index.tsx
+++ b/src/script/components/MessagesList/Message/index.tsx
@@ -107,12 +107,10 @@ const Message: React.FC<
 
   const content = <MessageWrapper {...props} hasMarker={markerType !== MessageMarkerType.NONE} />;
   const wrappedContent = props.onVisible ? <InViewport onVisible={props.onVisible}>{content}</InViewport> : content;
-
   return (
     <div
       className={`message ${isMarked ? 'message-marked' : ''}`}
       ref={messageElementRef}
-      key={message.id}
       data-uie-uid={message.id}
       data-uie-value={message.super_type}
       data-uie-expired-status={ephemeral_expires}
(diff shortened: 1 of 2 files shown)


src/script/components/MessagesList/index.tsx
```diff
@@ -179,9 +179,10 @@ const MessagesList: React.FC<MessagesListParams> = ({
     const isLastDeliveredMessage = lastDeliveredMessage?.id === message.id;
 
     const visibleCallback = getVisibleCallback(conversation, message);
+    const key = (message.id || 'message-') + message.timestamp();
```
kamil-au, line 182: For some messages `id` is undefined, and tine `timestamp()` is ko.observable so it has to be accessed this way not just `timestamp`.

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -107,12 +107,10 @@ const Message: React.FC<
 
   const content = <MessageWrapper {...props} hasMarker={markerType !== MessageMarkerType.NONE} />;
   const wrappedContent = props.onVisible ? <InViewport onVisible={props.onVisible}>{content}</InViewport> : content;
-
   return (
     <div
       className={`message ${isMarked ? 'message-marked' : ''}`}
       ref={messageElementRef}
-      key={message.id}
```
kamil-au, line 115: This one seems to be not required as we don't have a list here.

thisisamir98 (approved):

src/script/components/MessagesList/index.tsx
BartDyb: maybe ``message.id || `message-${message.timestamp()}` ``?

BartDyb (approved):

src/script/components/MessagesList/index.tsx
kamil-au: It will provide duplicated key, as the system messages have the same string as `id`

atomrc (commented):
Good catch, thanks for fixing it :pray:

PR #12940, opened 2022-04-29
runfix(accessibility): call accept hover

kamil-au:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### DIFF
--- a/src/style/components/list/conversation-list-calling-cell.less
+++ b/src/style/components/list/conversation-list-calling-cell.less
@@ -87,12 +87,12 @@
       background-color: var(--green-500) !important;
 
       &:hover {
-        border-color: var(--green-600) !important;
-        background-color: var(--green-600) !important;
+        border-color: @green-600 !important;
+        background-color: @green-600 !important;
 
         body.theme-dark & {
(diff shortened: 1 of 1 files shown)


tlebon (approved):

BartDyb (approved):

thisisamir98 (approved):

src/style/components/list/conversation-list-calling-cell.less
```diff
@@ -87,12 +87,12 @@
       background-color: var(--green-500) !important;
 
       &:hover {
-        border-color: var(--green-600) !important;
-        background-color: var(--green-600) !important;
+        border-color: @green-600 !important;
```
tlebon, line 90: just curious why some places this is less style vars and some CSS style? any particular reason for the split

kamil-au: we have `var(--color-500)` and `var(--accent-color-{colorIndex})` served from accent-colors.less, but here we just need a specific variant of `green` and `red` so instead of mapping every possible color and index to var we just use less variable.

PR #12968, opened 2022-05-02
refactor: share core and app db initialization between auth page and app

atomrc:
Right now, we have 2 classes that take care of instantiating databases:
- the Auth page that has the logic to bootstrap the database
- the App that has the same logic copied over to have access to the database created by the auth page 

Which means we generate the database name in different places. 

This PR harmonize the database creation and access. 
It let the core actually decide how the database should be name and let's it create it. 
We then get the reference of the database from the core directly and we can bootstrap the app on top of it.


src/script/auth/module/action/AuthAction.ts
```diff
@@ -356,9 +346,6 @@ export class AuthAction {
           throw new Error(`Could not find value for '${LocalStorageKey.AUTH.PERSIST}'`);
         }
         const clientType = persist ? ClientType.PERMANENT : ClientType.TEMPORARY;
-        if (isTemporaryClientAndNonPersistent(clientType === ClientType.PERMANENT)) {
```
atomrc, line 359: This logic has moved there https://github.com/wireapp/wire-webapp/pull/12968/files#diff-48632c70319182040bb02263c6e38bce17b438bcb0f2ba730020ad83f14ab7a5R32

src/script/storage/StorageService.ts
```diff
@@ -84,27 +78,13 @@ export class StorageService {
...
-    this.dbName = `wire@${Config.getConfig().ENVIRONMENT}@${userId}@${clientType}`;
+  init(userId: string = this.userId, requestPersistentStorage: boolean = false): string {
+    this.dbName = this.engine.storeName;
 
     try {
-      if (this.isTemporaryAndNonPersistent) {
-        this.logger.info(`Initializing Storage Service with encrypted database '${this.dbName}'`);
-        await this.engine.init(this.dbName);
-      } else {
-        this.db = new DexieDatabase(this.dbName);
```
atomrc, line 99: The creation of the database is now refactored in the `StoreEngineProvider`

thisisamir98 (approved):

PR #13000, opened 2022-05-09
runfix: Init the core as part of the App init [FS-647 FS-648]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-647" title="FS-647" target="_blank">BugFS-647</a>  [Web] Automatic login when all devices are deleted
  </td></table>
  <br />
 

In order to have proper error handling from the app.init function, we to include the `core.init` function in the `initApp` function. 

But since we need the storage engine (created by the core) in the `StorageService` at build time, this was not possible before.

This PR allows this by moving the moment we give the `StorageService` to when the `core` is actually initiated instead of in the constructor of the `StorageService`.

### DIFF
--- a/src/script/auth/AuthRepository.ts
+++ b/src/script/auth/AuthRepository.ts
@@ -17,13 +17,9 @@
  *
  */
 
-import type {Context} from '@wireapp/api-client/src/auth/';
-import {ClientType} from '@wireapp/api-client/src/client/';
 import {container} from 'tsyringe';
 
 import {Logger, getLogger} from 'Util/Logger';
-import {loadValue} from 'Util/StorageUtil';
-import {StorageKey} from '../storage/StorageKey';
 import {APIClient} from '../service/APIClientSingleton';
 
 export class AuthRepository {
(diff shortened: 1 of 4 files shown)


src/script/auth/AuthRepository.ts
```diff
@@ -49,12 +45,6 @@ export class AuthRepository {
     this.logger = getLogger('AuthRepository');
   }
 
-  init(): Promise<Context> {
```
atomrc, line 52: This is actually included in the `core.init` call in the `app.ts` file. 
This was redundant

src/script/main/app.ts
```diff
@@ -326,8 +328,8 @@ class App {
    * @param Encrypted database handler
    * @returns All services
    */
-  private _setupServices(encryptedEngine: SQLeetEngine) {
-    container.registerInstance(StorageService, new StorageService(encryptedEngine));
+  private _setupServices() {
+    container.registerInstance(StorageService, new StorageService());
```
atomrc, line 332: We cannot have access to the core database at this point. So the database will be given to the storage engine later on when we `init` the StorageEngine

src/script/main/app.ts
```diff
@@ -404,7 +405,9 @@ class App {
       this._registerSingleInstance();
       loadingView.updateProgress(2.5);
       telemetry.timeStep(AppInitTimingsStep.RECEIVED_ACCESS_TOKEN);
-      await authRepository.init();
+
+      await this.core.init(clientType);
```
atomrc, line 409: This is where we init the core and that the DB is actually going to be created

src/script/main/app.ts
```diff
@@ -606,7 +609,7 @@ class App {
       }
     }
 
-    await container.resolve(StorageService).init(userEntity.id);
+    await container.resolve(StorageService).init(this.core.storage);
```
atomrc, line 612: Later on, we can finally init the storage engine with the storage engine that the core has created

thisisamir98 (approved):

PR #13021, opened 2022-05-16
fix: escape constants in tests before using them in a regex

comawill:
Fix potentially unsafe usage of constants in RegExp expressions.


tlebon (approved):

tlebon:
@comawill theres a prettier linting issue in the file, please format and then it's good

atomrc (approved):

PR #13022, opened 2022-05-16
refactor: Use @wireapp/core to delete clients [FS-552]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-552" title="FS-552" target="_blank">TaskFS-552</a>  [Web] Register new client with MLS
  </td></table>
  <br />
 

With the integration of MLS we need to move all that is related to `clients` and `messages` to the core. 

This PR will let the core handle clients deletion (on backend and local db). In the future handling deletion of MLS clients will be completely transparent to the webapp

### DIFF
--- a/src/script/client/ClientRepository.ts
+++ b/src/script/client/ClientRepository.ts
@@ -287,10 +287,9 @@ export class ClientRepository {
    * @param password Password entered by user
    * @returns Resolves with the remaining user devices
    */
-  async deleteClient(clientId: string, password: string): Promise<ClientEntity[]> {
+  async deleteClient(clientId: string, password?: string): Promise<ClientEntity[]> {
     const selfUser = this.selfUser();
-    await this.clientService.deleteClient(clientId, password);
-    await this.deleteClientFromDb(selfUser.qualifiedId, clientId);
+    await this.core.service.client.deleteClient(clientId, password);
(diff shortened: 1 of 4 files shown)


tlebon (approved):

thisisamir98 (approved):
LGTM

src/script/client/ClientRepository.ts
```diff
@@ -287,10 +287,9 @@ export class ClientRepository {
    * @param password Password entered by user
    * @returns Resolves with the remaining user devices
    */
-  async deleteClient(clientId: string, password: string): Promise<ClientEntity[]> {
+  async deleteClient(clientId: string, password?: string): Promise<ClientEntity[]> {
     const selfUser = this.selfUser();
-    await this.clientService.deleteClient(clientId, password);
-    await this.deleteClientFromDb(selfUser.qualifiedId, clientId);
+    await this.core.service.client.deleteClient(clientId, password);
```
przemvs, line 292: This password can be undefined ?

atomrc: yes, that's on purpose. Temporary devices can be deleted without the password

przemvs: Okay, thanks!

przemvs (approved):

PR #13045, opened 2022-05-19
chore: Fix translation download job

atomrc:
Mostly aligns the translation download job with the one on Team Settings (see https://github.com/wireapp/wire-team-settings/blob/dev/.github/workflows/download_translations.yml)

### DIFF
--- a/.github/workflows/sync_translations.yml
+++ b/.github/workflows/sync_translations.yml
@@ -5,6 +5,7 @@ on:
     branches: [dev]
   schedule:
     - cron: '0 8 * * *'
+  workflow_dispatch:
 
 concurrency:
   group: translations-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/sync_translations.yml
```diff
@@ -36,12 +37,11 @@ jobs:
         uses: crowdin/github-action@1.4.8
         env:
           GITHUB_TOKEN: ${{secrets.OTTO_THE_BOT_GH_TOKEN}}
+          GITHUB_ACTOR: otto-the-bot
           CROWDIN_PROJECT_ID: 342359
           INPUT_DEBUG_MODE: false
         with:
           upload_sources: true
-          github_user_name: otto-the-bot
-          github_user_email: webapp+otto@wire.com
```
atomrc: A possible explanation for our translation update jobs to fail, would be that this user is not identified as a contributor with write access to the webapp repo... :shrug:

.github/workflows/sync_translations.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches: [dev]
   schedule:
     - cron: '0 8 * * *'
+  workflow_dispatch:
```
atomrc, line 8: This way we could also run it manually to do some testing

.github/workflows/sync_translations.yml
tlebon: i tried to switch it to otto before because it does have write access 😬

PR #13087, opened 2022-05-24
feat: Use @wireapp/core to decrypt incoming messages [FS-651]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-651" title="FS-651" target="_blank">TaskFS-651</a>  [Web] Use @wireapp/core for message decryption
  </td></table>
  <br />
 

This will use the Core to decrypt all incoming messages. This means the webapp now uses the core `listen` method to connect to the websocket (so all the lifecycle events of the websocket have been moved to the core). 

It also fixed a bug with the `no_internet` dismiss 

TODO:
- [x] Fix loading the webapp with some messages in the notification stream that are in a deleted converstation
- [x] missing events warning when device hasn't logged in since a long time

Needs:
- [x] https://github.com/wireapp/wire-web-packages/pull/4268


src/script/auth/module/action/WebSocketAction.ts
```diff
@@ -30,19 +29,6 @@ import * as SelfSelector from '../../module/selector/SelfSelector';
 export class WebSocketAction {
   private readonly logger = getLogger('WebSocketAction');
 
-  disconnect = (): ThunkAction => {
```
atomrc, line 33: This was never used

tlebon (approved):
LGTM! 🚀

PR #13120, opened 2022-06-02
chore: Improve dev documentation and default env config

atomrc:
Forces the dev default env config to hit `local.zinfra.io` to avoid CORS issues and remove section about this in the README (as it's not needed anymore)


tlebon (approved):
LGTM

PR #13158, opened 2022-06-08
runfix: Add a wrapper to group creation component to fix broken styles

thisisamir98:
fixes broken automation tests

### DIFF
--- a/src/script/components/UserSearchableList.tsx
+++ b/src/script/components/UserSearchableList.tsx
@@ -161,26 +161,32 @@ const UserSearchableList: React.FC<UserListProps> = ({onUpdateSelectedUsers, ...
   const userList = foundUserEntities();
 
   if (userList.length === 0) {
-    return users.length === 0 ? (
-      <div className="user-list__no-results" data-uie-name="status-all-added">
-        {t('searchListEveryoneParticipates')}
-      </div>
-    ) : (
-      <div className="user-list__no-results" data-uie-name="status-no-matches">
-        {t('searchListNoMatches')}
+    return (
+      <div className="user-list-wrapper">
+        {users.length === 0 ? (
+          <div className="user-list__no-results" data-uie-name="status-all-added">
+            {t('searchListEveryoneParticipates')}
+          </div>
+        ) : (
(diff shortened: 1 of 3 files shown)


atomrc (approved):

src/script/components/UserSearchableList.tsx
```diff
@@ -161,26 +161,32 @@ const UserSearchableList: React.FC<UserListProps> = ({onUpdateSelectedUsers, ...
...
+      <div className="user-list-wrapper">
+        {users.length === 0 ? (
+          <div className="user-list__no-results" data-uie-name="status-all-added">
+            {t('searchListEveryoneParticipates')}
+          </div>
+        ) : (
+          <div className="user-list__no-results" data-uie-name="status-no-matches">
+            {t('searchListNoMatches')}
+          </div>
+        )}
```
przemvs, line 174: I'll change this functionality. Make users.length === 0 as variable, eg
```
const hasUsers = !!users.length
```
Also make variables for data-uie-name and translation key

```
const dataUieName = hasUsers ? "status-no-matches" : "status-all-added";
const translationText = hasUsers ?  'searchListNoMatches' : 'searchListEveryoneParticipates';
```
And for component without ternary:
```
<div className="user-list__no-results" data-uie-name={dataUieName}>
  {t(translationText)}
</div>
```

przemvs (commented):
LGTM, we can consult about this functionality ;)

src/script/components/UserSearchableList.tsx
atomrc: Yes, that's is a valuable refactoring, I agree :+1: 

Also we could refactor to have the `<div className="user-list-wrapper">` only once in the component.

Like having something like 

```ts
const content = emptyUserList ? <div classname="noresults">...</div> : <UserList />

return <div className="user-list-wrapper">{content}</div>;
```

przemvs: @atomrc it's better to have if statement for return component, it's more readable ;)

thisisamir98: https://github.com/wireapp/wire-webapp/pull/13159

PR #13199, opened 2022-06-15
feat: Meaningful document title for all auth screens(ACC-42)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-42" title="ACC-42" target="_blank">StoryACC-42</a>  Meaningful document title (9.2.4.2)
  </td></table>
  <br />
 

----

- The **PR Title**
  - Meaningful document title for all auth screens

- The **PR Description**
  - added document title for all auth screens like login, welcome screen, forgot password, create account
----

# What's new in this PR?

### Issues

- Meaningful document title helps with better accessibility and search engine optimisation.

### Solutions

- added title for each component using useTitle hook, titles are translatable. 

#### How to Test

- Navigating between auth screens should display appropriate titles.


atomrc (approved):

src/script/auth/page/Login.tsx
```diff
@@ -65,8 +65,12 @@ import {parseError, parseValidationErrors} from '../util/errorUtil';
 import {UrlUtil} from '@wireapp/commons';
 import Page from './Page';
 import EntropyContainer from './EntropyContainer';
+import {t} from 'Util/LocalizerUtil';
 
-interface Props extends React.HTMLProps<HTMLDivElement> {}
+interface Props extends React.HTMLProps<HTMLDivElement> {
+  changeTitle: (title: string) => void;
+  title: string;
```
tlebon: i was wondering, since these same props are being added to every page here can we export them as an extra interface from somewhere general and just add it to the extends?

arjita-mitra: yes we can do that. I waned to talk about it.

src/script/auth/page/Index.test.tsx
```diff
@@ -42,7 +43,16 @@ class IndexPage {
     store: MockStoreEnhanced<TypeUtil.RecursivePartial<RootState>, ThunkDispatch<RootState, Api, AnyAction>>,
     history?: History<any>,
   ) {
-    this.driver = mountComponent(<Index />, store, history);
+    this.driver = mountComponent(
+      <Index
+        title=""
+        changeTitle={title => {
+          document.title = title;
+        }}
```
atomrc: The idea would be more to pass on a `mock` function here and in the test check that the mock function has been called. 
In the end the component doesn't really care what side effect is happening on the consumer side. 

It allows you to test your component as a unit: you just test that the component is calling the function when it's rendered :)

src/script/auth/page/Index.test.tsx
```diff
@@ -57,6 +67,25 @@ class IndexPage {
...
+        ...initialRootState,
+        runtimeState: {
+          hasCookieSupport: true,
+          hasIndexedDbSupport: true,
+          isSupportedBrowser: true,
+        },
+      }),
+    );
+    await waitForExpect(() => {
+      expect(global.window.document.title).toBe(indexTitle);
```
tlebon: does this work correctly? i wouldnt necessarily expect this to get the correct title from the index page

src/script/auth/page/Login.tsx
atomrc: Actually can we try this option as well, I feel like it's going to be much simpler
image

tlebon (approved):

src/script/auth/page/Root.tsx
```diff
@@ -87,6 +88,10 @@ const Root: React.FC<RootProps & ConnectedProps & DispatchProps> = ({
     return null;
   };
 
+  const setTitle = (title = 'Wire') => {
```
thisisamir98, line 91: this is hard coded, shouldn't be

PR #13206, opened 2022-06-17
fix: ignore outdated decryption errors

atomrc:
Hide Outdated and Duplicate event errors from the user. 

This is an issue since https://github.com/wireapp/wire-webapp/pull/13087/files#diff-6e9bc4673bf0a70fc0e1e6ce55adc18bf81ba2db7e0f4781b4dd0cd370af6061

The problem was that the original version of the decryption we ignored the 209 and 208 errors and didn't show them to the user. 
This logic was not correctly re-rewritten when the core was used for decryption. 

This fixes this misalignment

### DIFF
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -22,7 +22,6 @@ import {WebAppEvents} from '@wireapp/webapp-events';
 import {amplify} from 'amplify';
 import ko from 'knockout';
 import {USER_EVENT, CONVERSATION_EVENT} from '@wireapp/api-client/src/event/';
-import type {Notification} from '@wireapp/api-client/src/notification/';
 import {container} from 'tsyringe';
 import {getLogger, Logger} from 'Util/Logger';
 import {TIME_IN_MILLIS} from 'Util/TimeUtil';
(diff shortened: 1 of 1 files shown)


src/script/event/EventRepository.ts
```diff
@@ -57,18 +56,13 @@ import {PayloadBundleSource} from '@wireapp/core/src/main/conversation';
...
-  notificationsLoaded: ko.Observable<boolean>;
-  notificationsPromises: [(value?: unknown) => void, (value?: unknown) => void];
   notificationsTotal: number;
-  notificationsQueue: ko.ObservableArray<Notification>;
-  lastNotificationId: ko.Observable<string>;
-  notificationsBlocked: boolean;
-  webSocketBuffer: Notification[];
-  lastEventDate: ko.Observable<string>;
+  lastNotificationId: ko.Observable<string | undefined>;
+  lastEventDate: ko.Observable<string | undefined>;
```
atomrc, line 65: This is just some cleanup that I forgot to do in the original PR.

src/script/event/EventRepository.ts
```diff
@@ -186,6 +180,8 @@ export class EventRepository {
     await this.handleTimeDrift();
     const connect = () => {
       this.notificationHandlingState(NOTIFICATION_HANDLING_STATE.STREAM);
+      // We make sure there is only be a single active connection to the WebSocket.
+      this.disconnectWebSocket?.();
```
atomrc, line 184: To be on the safe side we want to make absolutely sure we are never connected to 2 websockets at the same time

tlebon (approved):

thisisamir98 (approved):

PR #13232, opened 2022-06-21
refactor: Migrate InviteModal to React

przemvs:
Migrate InviteModal to React


src/script/page/Modals/InviteModal/InviteModal.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+export default InviteModalComponent;
+
+let modalContainer: HTMLDivElement;
+let reactRoot: Root;
+
+const cleanUp = () => {
+  if (modalContainer) {
+    reactRoot.unmount();
+    document.body.removeChild(modalContainer);
```
thisisamir98: I think appending to document body will break some styles as it did in my PR can you check that and if the styles are broken do it to #wire-main instead like here: https://github.com/wireapp/wire-webapp/pull/13230/files#diff-95ed7641b4ddf02e5e622ac2d81f29df88d8a604c8b8226fdc5ee828a61db8d1R104-R115

przemvs: Fixed!

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
```diff
@@ -43,6 +43,7 @@ import {Conversation} from 'src/script/entity/Conversation';
 import {ServiceEntity} from 'src/script/integration/ServiceEntity';
 import {showUserModal} from '../../../../page/Modals/UserModal/UserModal';
 import {showServiceModal} from '../../../../page/Modals/ServiceModal/ServiceModal';
+import {showInviteModal} from '../../../../page/Modals/InviteModal/InviteModal';
```
atomrc: Can you try adding a `index.ts` at the root of `InviteModal` that just exposes the API, so that we can import it without having to do `InviteModal/InviteModal`?

przemvs: Hmm... I think good way is to specify aliases in webpack :)

atomrc: Actually, we already have an alias for `Components` maybe it's a good idea to move all the modals to `Components/Modals` :thinking: 

Still adding an `index.ts` at the root of the folder is good because you then import by doing
```js
import {Component} from 'Components/Modal/Component`
```
instead of
```js
import {Component} from 'Components/Modal/Component/Component`
```

src/script/page/Modals/ServiceModal/ServiceModal.tsx
```diff
@@ -95,24 +94,4 @@ const ServiceModal: React.FC<ServiceModalProps> = ({
...
-  const {integrationRepository, service} = props;
-  console.info('bardia showServiceModal', props);
-  cleanUp();
-  integrationRepository.addProviderNameToParticipant(service);
-  modalContainer = document.createElement('div');
-  document.getElementById('wire-main').appendChild(modalContainer);
-  reactRoot = createRoot(modalContainer);
-  reactRoot.render(<ServiceModal {...props} onClose={cleanUp} />);
-};
+export const showServiceModal = renderModal<ServiceModalProps>(ServiceModal);
```
tlebon, line 97: theres a TS error in startUI, this appears to be missing some props

thisisamir98 (approved):

tsconfig.json
```diff
@@ -28,6 +28,7 @@
     // Must be in sync with jest.config.js >> moduleNameMapper
     "paths": {
       "Components/*": ["./src/script/components/*"],
+      "Modals/*": ["./src/script/page/Modals/*"],
```
atomrc: I'd like to start a discussion about this. 
I don't think we should add aliases too lightly. Aliases are great for utility tools that we use across the entire app, not for isolated things we use in 2/3 different places. 
Aliases tie us a lot with the bundler/compilation step we use and we need to maintain those across all build steps (tsconfig, webpack config, test config...). 

In this case, I would rather not add a custom alias for this

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
przemvs: For me it's good idea! :) Modal is a component :D

tsconfig.json
przemvs: Sure! :)

atomrc (approved):

PR #13242, opened 2022-06-24
fix: conversation action hover states, like button color on dark and light mode, send button positioning(ACC-160)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-160" title="ACC-160" target="_blank">BugACC-160</a>  [Webapp] Incorrect Message Action Buttons States
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

The arrow in the send button is not center aligned.

### Solutions

- use margin to position it

#### How to Test
Send button arrow should be center aligned.


src/style/common/like.less
```diff
@@ -54,7 +54,7 @@
 
 .like-button-liked {
   > span {
-    color: var(--gray-70) !important;
+    color: @red-500 !important;
```
V-Gira: Didn't Wolfgang specify dark-red-500?

V-Gira (commented):
Gotta edit the title to "runfix: ..."

src/style/common/like.less
arjita-mitra: [Figma](https://www.figma.com/file/OrP9jAJdfttfvsm3nxnihP/Accessibility-Quick-Wins-(Web%2FDesktop)?node-id=924%3A14200), click on the heart icon and you can see the colour mentioned for the heart icon is
Light UI/Red 500 #C20013

If thats wrong then the figma needs to be updated.

V-Gira: gotcha, I imagine Wolfgang is using Dark Mode and is trying to say the heart color should be different depending on the theme. I agree it's a mess, there's no Dark Mode exemple in Figma and I interpreted Wolfgang comment as "heart is @dark-red-500 all the time".

You do need to change the title though, you can't merge a PR without semantic commit title

tlebon (approved):

V-Gira (approved):

atomrc (approved):

PR #13244, opened 2022-06-26
chore: Put eslint jsx-a11y/interactive-supports-focus rule to error

atomrc:
Since all the [jsx-a11y/interactive-supports-focus](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/interactive-supports-focus.md) errors have been fixed, we can now put the rule to `error` instead of `warn`


tlebon:
question: removing the line makes this an error? it wont just remove the rule? (I'm not clear this is included as error by default but I suppose it must be)

atomrc:
@tlebon yes, it's configured as `error` by default. We just overrode it to `warn` during the time we fixed those issues :)

tlebon (approved):

PR #13261, opened 2022-06-27
runfix: allow autofocus on group creation's text input acc-178

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-178" title="ACC-178" target="_blank">BugACC-178</a>  [Web] No autofocus on group name input when creating a group
  </td></table>
  <br />
 

----

## Needs review on how to proceed

> **This issue has been pushed as high priority by Rahil and he confirms that we want the autofocus**

### Issues

- The input for group creation doesn't autofocus and pressing enter doesn't proceed

### Solutions

- Add the react autoFocus prop to the input
- Add an onkeydown event when pressing enter

### However

Autofocus shouldn't be used for accessibility reasons, is there a way to control focus that plays nice with screen readers?
Trying to solve the eslint issue by using refs and focus() feels like it's just ignoring the problem.

Screenshot from 2022-06-21 15-54-35

### Compromise

We are discussing allowing the autoFocus React prop only on a case by case basis.

- It's important to rename the input prop "autofocus" to "autoFocus" so it's flagged by eslint in the parent component

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -44,7 +44,7 @@ import {User} from '../../../entity/User';
 import {ACCESS_STATE} from '../../../conversation/AccessState';
 import Icon from 'Components/Icon';
 import {t} from 'Util/LocalizerUtil';
-import {onEscKey, offEscKey, KEY} from 'Util/KeyboardUtil';
+import {onEscKey, offEscKey, handleKeyDown} from 'Util/KeyboardUtil';
 import {
   ACCESS_TYPES,
   teamPermissionsForAccessState,
@@ -339,9 +339,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
                   setGroupName(trimmedName);
                 }}
                 onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
-                  if (event.key === KEY.ENTER) {
-                    clickOnNext();
-                  }
+                  handleKeyDown(event, clickOnNext);
                 }}
                 value={groupName}
                 isError={nameError.length > 0}
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -320,12 +320,12 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
             )}
           </div>
         )}
-
+        {/* eslint jsx-a11y/no-autofocus : "off" */}
         {stateIsPreferences && (
           <>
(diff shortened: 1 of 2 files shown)


src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -336,6 +338,11 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
                   const trimmedName = value.trim();
                   setGroupName(trimmedName);
                 }}
+                onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
```
atomrc, line 342: Please use `handleKeyDown` from the `KeyboardUtil` file

src/script/components/TextInput/TextInput.tsx
```diff
@@ -96,7 +98,9 @@ const TextInput: React.ForwardRefRenderFunction<HTMLDivElement, UserInputProps>
         </span>
       )}
 
+      {/* eslint-disable */}
```
atomrc: humm this sounds bad

src/script/components/TextInput/TextInput.tsx
```diff
@@ -96,7 +98,9 @@ const TextInput: React.ForwardRefRenderFunction<HTMLDivElement, UserInputProps>
...
       <input
+        autoFocus={autofocus}
```
atomrc: I think you should use a `ref` + `useEffect` to trigger the focus at first render. Can you try playing with this?

atomrc (changes requested):

src/script/components/TextInput/TextInput.tsx
V-Gira: haha, there's a reason I flagged it as "Needs review", thanks for suggesting a way forward :)

tlebon: just disable one rule

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -282,6 +282,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
                   'accent-text': groupNameLength,
                   enabled: groupNameLength && !nameError.length,
```
V-Gira: This is actually still needed to give the enabled label correct style

atomrc: Since this condition and https://github.com/wireapp/wire-webapp/pull/13261/files#diff-374d17b2a9898c44a7be4d7d1e40a329f71a2d6b53c865e052296e2e4988646aR285 are the exact negation of one-another, let's extract this to a variable earlier in the file ;)

atomrc (approved):

PR #13267, opened 2022-06-29
runfix: Avoid triggering close event on modal first render [SQCORE-1327]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1327" title="SQCORE-1327" target="_blank">BugSQCORE-1327</a>  [Web] Blank app lock modal when refreshing page while app is locked
  </td></table>
  <br />
 

The problem is when the app lock modal is instantiated, it then creates the modal with `isVisible: false` which then triggers the ModalComponent `onClose` callback which resets the React state of the AppLock Component. 
For the initial rendering of the `ModalComponent` we do not need to trigger the `onClose` event (since the modal was never opened in the first place)

### DIFF
--- a/src/script/components/ModalComponent.tsx
+++ b/src/script/components/ModalComponent.tsx
@@ -18,7 +18,7 @@
  */
 
 import {CSSObject} from '@emotion/react';
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import {noop} from 'Util/util';
 import Icon from './Icon';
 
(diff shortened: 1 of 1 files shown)


arjita-mitra (approved):

src/script/components/ModalComponent.tsx
```diff
@@ -87,11 +87,20 @@ const ModalComponent: React.FC<ModalComponentProps> = ({
 }) => {
   const [displayNone, setDisplayNone] = useState<boolean>(!isShown);
   const hasVisibleClass = isShown && !displayNone;
+  const isMounting = useRef<boolean>(true);
```
thisisamir98, line 90: please use useIsMounted in src/script/util/useIsMounted.ts

PR #13273, opened 2022-06-29
runfix: add focus trap to modals acc-178

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-178" title="ACC-178" target="_blank">BugACC-178</a>  [Web] No autofocus on group name input when creating a group
  </td></table>
  <br />
 

----

# What's new in this PR?

### Issues

- Focus needs to be trapped inside modals for accessibility reasons
- Shift Tabbing doesn't work in preventFocusOutside

### Solutions

- Use the preventFocusOutside function from Util
- Hard code shift-tabbing into preventFocusOutside
Peek 2022-06-29 16-02

### DIFF
--- a/src/script/components/ModalComponent.tsx
+++ b/src/script/components/ModalComponent.tsx
@@ -18,7 +18,7 @@
  */
 
 import {CSSObject} from '@emotion/react';
-import React, {useEffect, useRef, useState} from 'react';
+import React, {useEffect, useId, useRef, useState} from 'react';
 import {noop, preventFocusOutside} from 'Util/util';
 import Icon from './Icon';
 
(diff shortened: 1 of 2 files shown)


src/script/components/ModalComponent.tsx
```diff
@@ -88,6 +88,23 @@ const ModalComponent: React.FC<ModalComponentProps> = ({
   const [displayNone, setDisplayNone] = useState<boolean>(!isShown);
   const hasVisibleClass = isShown && !displayNone;
   const isMounting = useRef<boolean>(true);
+  const trapId = useId();
+
+  const onKeyDown = (event: KeyboardEvent): void => {
```
atomrc: I think we should use a `useCallback` here, since we want the reference to the function to be stable over re-renders.

atomrc: Also please rename the `onKeyDown` to a more imperative style function name (like `trapFocus`) that would explain more what the function is actually doing. 

I know we have a history of lazy naming event handler as the name of the event they are reacting to, but I think it's a bad practice that completely hides what the purpose of the function is

atomrc (approved):

PR #13277, opened 2022-06-30
chore: add 'betterer' to prevent ts strict regressions (#WEBAPP-7144)

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WEBAPP-7144" title="WEBAPP-7144" target="_blank">TaskWEBAPP-7144</a>  Have a job that checks that the number of strict typescript errors is going down
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Our goal is to make our code compiled with TypeScript **strict** mode. As our codebase is pretty big and currently includes about 2300 ts strict errors (null checks, possible undefineds, etc.) It's not possible to just enable the `strict` mode.

We want to prevent us from pushing more typescript `strict` errors.

### Solutions

I've found pretty cool tool called [Betterer](https://phenomnomnominal.github.io/betterer/), which let us keep track of ts compile errors, and saves them as a snapshot. The tool does not allow to increase the number of specified errors, but each time we reduce the number, the snapshot gets updated 😄 

### Setup

Here is the recommended workflow by tool's authors: https://phenomnomnominal.github.io/betterer/docs/workflow. It seems to match our requirements. How it would basically work:

- there is pre-commit husky hook added which will run after `lint-staged`, so it won't allow us to push if we've added any more ts strict errors.
- there is a ci action that will run and make sure our snapshot is in sync with the errors.
- in case we got any merge conflicts in snapshot file, there is a `yarn betterer:conflict` script which should fix it
- in case something bad happens to the snapshot file, we can force its update with `yarn betterer -u`

### Notes

I've never used this tool before, and I need your opinion. I've tested it out locally and works great!

Our codebase is pretty big and `betterer` command can take long time (about 16s for me) on the first run, but thanks to `--cache` option added  next runs should be much faster.

----
##### References

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ CHANGELOG.md
 node_modules
 npm-debug.log
 yarn-error.log
+.betterer.cache
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no-install lint-staged
-yarn betterer precommit
+yarn betterer precommit --cache
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no-install lint-staged
+yarn betterer precommit
--- a/.betterer.results
+++ b/.betterer.results
@@ -1,0 +1,3018 @@
+// BETTERER RESULTS V2.
+// 
+// If this file contains merge conflicts, use `betterer merge` to automatically resolve them:
+// https://phenomnomnominal.github.io/betterer/docs/results-file/#merge
+//
+exports[`stricter compilation`] = {
+  value: `{
+    "src/script/Config.ts:3231864459": [
+      [51, 4, 40, "\'ACCOUNT_BASE\' is specified more than once, so this usage will be overwritten.", "2150385420"],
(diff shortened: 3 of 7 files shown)


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no-install lint-staged
+yarn betterer precommit
```
atomrc: How expensive is this command? (in term of time of execution). If it's too long I fear it's going to just have people do the `commit -n` and cancel all checks

PatrykBuniX: It's about 16s for me (m1 mac), It's a lot of files to check. Is it too long? 🤔

atomrc: Oh yeah, that's a bit long for a precommit... :grimacing:

.gitignore
```diff
@@ -16,3 +16,4 @@ CHANGELOG.md
 node_modules
 npm-debug.log
 yarn-error.log
+.betterer.cache
```
PatrykBuniX, line 19: Can this one live here, or we should use individual global ignores for this type of caching things? 🤔

atomrc: Since we are all going to use it locally, I think it makes sense to have it in the repo :ok_hand:

atomrc (approved):

PR #13304, opened 2022-07-04
chore: fix strict ts errors

PatrykBuniX:


thisisamir98 (approved):

src/script/page/MainContent/panels/preferences/accountPreferences/FileInput.tsx
```diff
@@ -20,30 +20,32 @@
...
-});
+const FileInput = React.forwardRef<HTMLInputElement>(
+  ({fileTypes = ['*'], onFileChange, ...rest}: FileInputProps, ref) => {
+    return (
+      <input
+        ref={ref}
+        type="file"
+        accept={fileTypes.join(',')}
+        onChange={({target: {files, value}}) => {
+          if (files && files.length > 0) {
```
atomrc:
```suggestion
          if (files?.length > 0) {
```

src/script/page/MainContent/panels/preferences/accountPreferences/FileInput.tsx
```diff
@@ -20,30 +20,32 @@
...
+            onFileChange?.(files);
+            window.setTimeout(() => {
+              value = '';
```
atomrc: :warning: This is a breaking change, since you are desctructuring, you just update the reference `value` points to but not the value of `target.value`. Please revert to `target.value = ''`

atomrc (changes requested):

src/script/page/MainContent/panels/preferences/accountPreferences/FileInput.tsx
PatrykBuniX: Good catch! Thats why I avoid mutations when its possible 😆

src/script/page/MainContent/panels/preferences/accountPreferences/FileInput.tsx
PatrykBuniX: Can't shorten it this way, `files?.length` might be `undefined`

atomrc: Ohh right, makes sense. My mistake :sweat_smile:

atomrc (approved):

PR #13307, opened 2022-07-05
runfix: Give fallback domain when showing user modal

atomrc:
With backend API v2, fetching a user requires the domain to be given. 
If the URL is not present in the url params, we then fallback to the domain configured in the apiClient

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -703,7 +703,7 @@ class App {
       '/preferences/av': () => mainView.list.openPreferencesAudioVideo(),
       '/preferences/devices': () => mainView.list.openPreferencesDevices(),
       '/preferences/options': () => mainView.list.openPreferencesOptions(),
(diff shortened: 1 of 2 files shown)


tlebon (approved):

src/script/main/app.ts
```diff
@@ -703,7 +703,7 @@ class App {
       '/preferences/av': () => mainView.list.openPreferencesAudioVideo(),
       '/preferences/devices': () => mainView.list.openPreferencesDevices(),
       '/preferences/options': () => mainView.list.openPreferencesOptions(),
-      '/user/:userId(/:domain)': (userId: string, domain?: string) => {
+      '/user/:userId(/:domain)': (userId: string, domain: string = this.apiClient.context!.domain ?? '') => {
```
tlebon, line 706: should this actually be a ! if we are still expecting a failure case? maybe ? is better

atomrc: Fair point, will fix in a follow up PR

PR #13310, opened 2022-07-06
runfix: Use asset API v2 for services

atomrc:
The asset API now needs to be fully qualified (with domain name). 
We need to pass the default domain for Services


src/script/integration/IntegrationRepository.ts
```diff
@@ -195,26 +194,14 @@ export class IntegrationRepository {
...
-
-    const {services: servicesData} = await this.integrationService.getServices(tagsArray.join(','), start);
-    return IntegrationMapper.mapServicesFromArray(servicesData);
-  }
-
-  async getServicesByProvider(providerId: string): Promise<ServiceEntity[]> {
-    const servicesData = await this.integrationService.getProviderServices(providerId);
-    return IntegrationMapper.mapServicesFromArray(servicesData);
-  }
-
```
atomrc, line 217: Those were never used :partying_face:

przemvs (approved):

PR #13311, opened 2022-07-06
feat: Send MLS messages (FS-560)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-560" title="FS-560" target="_blank">TaskFS-560</a>  [Web] Send message in an MLS group
  </td></table>
  <br />
 

send mls messages


src/script/conversation/ConversationRepository.ts
```diff
@@ -396,7 +397,7 @@ export class ConversationRepository {
      * we need to add this creator_client to conversation creation payload
      * for creating MLS conversations
      */
-    if (this.teamState.isMLSEnabled()) {
```
tlebon, line 399: do we still need to make sure MLS is enabled in the team features?

thisisamir98: image
this is already changed, now we're checking protocol from the options instead

tlebon: im asking if *also* need to check that its enabled for the team

atomrc: It's checked at another level. At this point, if the conversation is marked as MLS, we can trust that the team config is correct and we need to handle it as such :)

arjita-mitra (approved):

tlebon (approved):

PR #13316, opened 2022-07-08
runfix: add hover state to radios acc-79

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-79" title="ACC-79" target="_blank">SubtaskACC-79</a>  Preferences page changes
  </td></table>
  <br />
 

----

# What's new in this PR?

### Issues

- Radios are missing a hover status
- Some styles still differed from the figma reference

### Solutions

- Adjust border-color on hover
-  Adjust border size from 1 to 2 px


src/style/content/preferences/options.less
```diff
@@ -64,7 +64,7 @@
   label::before {
     width: 22px;
     height: 22px;
-    border: 1px solid var(--gray-80);
+    border: 2px solid var(--gray-80);
```
aweiss-dev: I don't understand why we have so many hardcoded pixel values. Arent borders and spacings streamlined for the app and therefore should be variables?

src/style/content/preferences/options.less
```diff
@@ -88,19 +91,21 @@ body.theme-dark {
     }
 
     label::before {
-      border: 1px solid var(--gray-60);
+      border: 2px solid var(--gray-60);
```
aweiss-dev: Since the change occurs here as well I  think its time for a variable

src/style/panel/panel.less
```diff
@@ -227,7 +232,7 @@
             width: 22px;
             height: 22px;
             box-sizing: border-box;
-            border: 1px solid var(--gray-80);
+            border: 2px solid var(--gray-80);
```
aweiss-dev: see above

src/style/content/preferences/options.less
przemvs: or prepare mixin, or smth like that where you can pass variable for color, because in this case you have different color than in 2 other comments ;)

src/style/content/preferences/options.less
V-Gira: Ain't nobody got time to have a conversation with me about what to do with css variables and theming going forward. There's no variable for radio-borders, should I had one.

aweiss-dev (approved):

PR #13335, opened 2022-07-11
fix: sidebar focus not clearly visible(ACC-204)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-204" title="ACC-204" target="_blank">SubtaskACC-204</a>  Sidebar focus is not visible
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Sidebar elements focus is not clearly indicated. Sidebar element hover and focus state was wrong.

### Solutions

create an outline to be clearly visible on focus. Fix hover states as described in the mock.

#### How to Test

- sidebar focus and hover state should match the mock. Sidebar focus should be clearly indicated.

----
##### References
1. https://www.figma.com/file/OrP9jAJdfttfvsm3nxnihP/Accessibility-Quick-Wins-(Web%2FDesktop)?node-id=293%3A2964

### DIFF
--- a/src/style/list/list.less
+++ b/src/style/list/list.less
@@ -146,8 +146,21 @@ body.theme-dark {
     text-align: left;
     transition: background-color 0.15s ease-in-out;
 
-    &:hover {
+    &:hover,
+    &:focus-visible,
+    &--active:focus-visible {
       background: @white;
+      color: @black;
+    }
+
+    &:focus-visible,
+    &--active:focus-visible {
+      outline: 1px solid var(--accent-color);
+      outline-offset: -2px;
+
+      .left-column-icon > svg {
+        fill: @black;
+      }
     }
 
     &--active,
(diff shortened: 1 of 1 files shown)


src/style/list/list.less
```diff
@@ -146,8 +146,21 @@ body.theme-dark {
...
+    &:focus-visible,
+    &--active:focus-visible {
       background: @white;
+      color: @black;
+    }
+
+    &:focus-visible,
+    &--active:focus-visible {
+      outline: 1px solid var(--accent-color);
+      outline-offset: -2px;
```
PatrykBuniX, line 159: It's negative because we want it inside active element, right?

PatrykBuniX (approved):

V-Gira (approved):

src/style/list/list.less
```diff
@@ -246,13 +259,24 @@ body.theme-dark {
   }
 
   .left-list-item-button {
-    &:hover {
-      background: @gray-95;
```
V-Gira, line 250: Good catch!

src/style/list/list.less
arjita-mitra: yes

PR #13338, opened 2022-07-13
fix: messages bigger than viewport not being read (#SQSERVICES-1558)

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1558" title="SQSERVICES-1558" target="_blank">BugSQSERVICES-1558</a>  [Web]Number badge doesn't go away
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

There is a bug where messages bigger than window's viewport are not being read. 

### Causes

Using `IntersectionObserver` we mark the messages as read when the whole message is visible in the viewport, so it does not work for messages higher that viewport. 

### Solutions

I suggest adding `allowBiggerThanViewport` boolean (optional, defaults to `false`) to the `<InViewport />` component's props. When its value is set to true, and observed element is higher than viewport, we treat this element as visible.

When element will be considered as visible:

`const isVisible = isIntersecting && (!fullyInView || isFullyInView() || isBiggerThanRoot());`

`isIntersecting`: The element must be partially visible (at least 1%), and:
- `!fullyInView` (`<InViewport />`'s prop) - we don't want the whole element to be visible, or
-  `isFullyInView()` - we want the whole element to be visible and it's fully in the view (doesn't work for higher messages)

after this PR is merged:

- `isBiggerThanRoot()` - we want the whole element to be visible, but we also allow it to be partially visible when it's bigger than viewport (only if `allowBiggerThanViewport` is set to `true`) 

This is how I check if element is bigger than viewport:
```ts
const isBiggerThanRoot = () => {
      return (
        allowBiggerThanViewport &&
        !!rootBounds &&
        (element.clientHeight > rootBounds.height || element.clientWidth > rootBounds.width)
      );
};
```

----

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References

### DIFF
--- a/src/script/components/MessagesList/Message/index.tsx
+++ b/src/script/components/MessagesList/Message/index.tsx
@@ -76,7 +76,7 @@ export interface MessageParams extends MessageActions {
 const Message: React.FC<
   MessageParams & {scrollTo?: (elm: {center?: boolean; element: HTMLElement}, isUnread?: boolean) => void}
 > = props => {
-  const {message, previousMessage, isMarked, lastReadTimestamp} = props;
+  const {message, previousMessage, isMarked, lastReadTimestamp, onVisible} = props;
   const messageElementRef = useRef<HTMLDivElement>();
   const {status, ephemeral_expires, timestamp} = useKoSubscribableChildren(message, [
     'status',
@@ -106,8 +106,8 @@ const Message: React.FC<
   };
 
   const content = <MessageWrapper {...props} hasMarker={markerType !== MessageMarkerType.NONE} />;
-  const wrappedContent = props.onVisible ? (
-    <InViewport allowBiggerThanViewport onVisible={props.onVisible}>
+  const wrappedContent = onVisible ? (
+    <InViewport allowBiggerThanViewport onVisible={onVisible}>
       {content}
     </InViewport>
   ) : (
--- a/src/script/components/MessagesList/Message/index.tsx
+++ b/src/script/components/MessagesList/Message/index.tsx
@@ -106,7 +106,13 @@ const Message: React.FC<
   };
 
   const content = <MessageWrapper {...props} hasMarker={markerType !== MessageMarkerType.NONE} />;
-  const wrappedContent = props.onVisible ? <InViewport onVisible={props.onVisible}>{content}</InViewport> : content;
+  const wrappedContent = props.onVisible ? (
+    <InViewport allowBiggerThanViewport onVisible={props.onVisible}>
+      {content}
+    </InViewport>
+  ) : (
+    content
+  );
   return (
     <div
       className={`message ${isMarked ? 'message-marked' : ''}`}
--- a/src/script/components/utils/InViewport.tsx
+++ b/src/script/components/utils/InViewport.tsx
@@ -24,11 +24,18 @@ import {viewportObserver} from '../../ui/viewportObserver';
 interface InViewportParams {
   children?: React.ReactNode;
   fullyInView?: boolean;
+  allowBiggerThanViewport?: boolean;
   onVisible: () => void;
   style?: React.CSSProperties;
 }
 
-const InViewport: React.FC<InViewportParams> = ({children, style, onVisible, fullyInView = true}) => {
+const InViewport: React.FC<InViewportParams> = ({
+  children,
+  style,
+  onVisible,
+  fullyInView = true,
+  allowBiggerThanViewport = false,
+}) => {
   const domNode = useRef<HTMLDivElement>();
 
   useEffect(() => {
@@ -58,6 +65,7 @@ const InViewport: React.FC<InViewportParams> = ({children, style, onVisible, ful
         triggerCallbackIfVisible();
       },
       fullyInView,
+      allowBiggerThanViewport,
       element.parentElement,
     );
     overlayedObserver.trackElement(element, isVisible => {
--- a/src/script/components/utils/InViewport.tsx
+++ b/src/script/components/utils/InViewport.tsx
@@ -22,10 +22,10 @@ import {overlayedObserver} from '../../ui/overlayedObserver';
 import {viewportObserver} from '../../ui/viewportObserver';
 
 interface InViewportParams {
-  children?: React.ReactNode;
-  fullyInView?: boolean;
-  allowBiggerThanViewport?: boolean;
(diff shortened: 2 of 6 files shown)


src/script/components/MessagesList/Message/index.tsx
```diff
@@ -106,7 +106,13 @@ const Message: React.FC<
   };
 
   const content = <MessageWrapper {...props} hasMarker={markerType !== MessageMarkerType.NONE} />;
-  const wrappedContent = props.onVisible ? <InViewport onVisible={props.onVisible}>{content}</InViewport> : content;
+  const wrappedContent = props.onVisible ? (
```
przemvs: Please spread onVisible property.

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -106,7 +106,13 @@ const Message: React.FC<
...
+    <InViewport allowBiggerThanViewport onVisible={props.onVisible}>
```
przemvs: Please spread onVisible property.

src/script/components/utils/InViewport.tsx
```diff
@@ -24,11 +24,18 @@ import {viewportObserver} from '../../ui/viewportObserver';
 interface InViewportParams {
   children?: React.ReactNode;
   fullyInView?: boolean;
+  allowBiggerThanViewport?: boolean;
```
przemvs, line 28: Move non optional props to top of interface

src/script/components/utils/InViewport.tsx
```diff
@@ -58,6 +65,7 @@ const InViewport: React.FC<InViewportParams> = ({children, style, onVisible, ful
         triggerCallbackIfVisible();
       },
       fullyInView,
+      allowBiggerThanViewport,
```
przemvs, line 69: I think it should be passed as useEffect dependency ?

src/script/ui/viewportObserver.ts
```diff
@@ -23,16 +23,25 @@ const observedElements = new Map();
...
     const isFullyInView = () => {
       if (container) {
         const minHeight = Math.min(container.clientHeight, element.clientHeight) * tolerance;
         return intersectionRect.height >= minHeight;
       }
       return intersectionRatio >= tolerance;
     };
-    const isVisible = isIntersecting && (!fullyInView || isFullyInView());
+
+    const isBiggerThanRoot = () => {
```
przemvs, line 37: I think you can memoize this function :)

src/script/ui/viewportObserver.ts
```diff
@@ -23,16 +23,25 @@ const observedElements = new Map();
...
+      return (
+        allowBiggerThanViewport &&
+        !!rootBounds &&
+        (element.clientHeight > rootBounds.height || element.clientWidth > rootBounds.width)
+      );
+    };
+
+    const isVisible = isIntersecting && (!fullyInView || isFullyInView() || isBiggerThanRoot());
```
przemvs: !fullyInView || isFullyInView() - hmm.. this naming of **isFullyInView** and **fullyInView** looks name the same :D Maybe rename **isFullyInView** to **checkIsFullyInView** ? And then move `(!fullyInView || checkIsFullyInView())` to variable isFullyInView ?

src/script/ui/viewportObserver.ts
```diff
@@ -73,11 +84,18 @@ const onElementInViewport = (
  * @param element the element to observe
  * @param onChange the callback to call when the element intersects or not
  * @param fullyInView should the element be fully in view
+ * @param allowBiggerThanViewport should fire when element is bigger than viewport
  * @param container the element containing the element
  */
-const trackElement = (element: HTMLElement, onChange: Function, fullyInView: boolean, container: HTMLElement): void => {
+const trackElement = (
```
przemvs, line 91: Hmm.. I think the fullyInView, allowBiggerThanViewport can be optional, and have default value as false :)

przemvs (changes requested):

src/script/ui/viewportObserver.ts
PatrykBuniX: Its just a regular function, not a hook or component, can't use `useMemo` in this case (rules of hooks).

src/script/ui/viewportObserver.ts
PatrykBuniX: I don't want to group these two into a variable, they are two separate conditions, I'll change `fullyInView` name to `requireFullyInView` though to make it more obvious what it's for :D .

src/script/ui/viewportObserver.ts
przemvs: Ahh.. i doesn't look at that. Sorry :D

src/script/components/utils/InViewport.tsx
tlebon: maybe we can turn on a lint rule for this also

src/script/ui/viewportObserver.ts
```diff
@@ -72,12 +84,19 @@ const onElementInViewport = (
...
+  element: HTMLElement,
+  onChange: Function,
+  requireFullyInView = false,
+  allowBiggerThanViewport = false,
+  container?: HTMLElement,
```
przemvs: Maybe move container above requireFullyInView and then in src/script/view_model/bindings/CommonBindings.ts you can pass to trackElement
```
viewportObserver.trackElement(
      element,
      (isInViewport: boolean) => {
        inViewport = isInViewport;
        triggerCallbackIfVisible();
      },
      container,
      true,
    );
```

What do you think? :)

przemvs (approved):
Code LGTM, 1 thing :)

src/script/components/utils/InViewport.tsx
przemvs: Yeah, it will be the best!

src/script/ui/viewportObserver.ts
PatrykBuniX: seems cool, so the boolean modifiers are always at the last place 👍

tlebon (approved):

atomrc:
Should also address https://github.com/wireapp/wire-desktop/issues/5961

PR #13346, opened 2022-07-15
chore: Fix betterer results

thisisamir98:
Fix betterer results


tlebon (approved):
if it passes it merges

PR #13367, opened 2022-07-18
fix(entropy): Improve UX of entropy creation process (SQSERVICES-1674)

comawill:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1674" title="SQSERVICES-1674" target="_blank">BugSQSERVICES-1674</a>  Entropy creation UI is misleading and indicates that the App might generate "bad" entropy
  </td></table>
  <br />
 

----

- The **PR Title**
----

# What's new in this PR?

This PR is a draft to discuss a possible solution to improve the UX of the entropy creation process.


src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -35,35 +35,36 @@ const EntropyCanvas = (props: CanvasProps) => {
...
-      pauseInterval();
-    }
-    if (frames > 300) {
-      startInterval();
-    }
-    if (frames >= 100) {
+    if (percent >= MIN_FRAMES) {
       clearInterval();
     }
   }, [percent, percent]);
```
tlebon: hmm i guess i wrote this but now I'm wondering why the dependency array has two copies of this

src/script/auth/page/EntropyContainer.tsx
```diff
@@ -88,7 +88,7 @@ const EntropyContainer = ({onSetEntropy}: Props) => {
           </Text>
         </>
       )}
-      {percent === 95 && frames < 300 && <Muted center>{_(setEntropyStrings.moreEntropyNeeded)}</Muted>}
+      {percent >= 95 && percent < 100 && <Muted center>{_(setEntropyStrings.moreEntropyNeeded)}</Muted>}
```
tlebon: i think this should just be removed, doesn't make sense to me anymore

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -35,35 +35,36 @@ const EntropyCanvas = (props: CanvasProps) => {
...
 
   const onMouseLeave = () => {
     onProgress(entropy, percent, true);
     pauseInterval();
-
-    setEntropy([...entropy, null]);
```
tlebon, line 66: this should still be in here, creates a UI bug now if you leave the box but come back on a different side.

src/script/auth/component/ProgressBar.tsx
```diff
@@ -46,7 +46,6 @@ export const ProgressBar = ({width, percent, error, style}: ProgressProps) => {
           backgroundColor: error ? 'red' : 'blue',
           borderRadius: '5px',
           height: 'inherit',
-          transition: '0.5s ease-in',
```
tlebon, line 49: i think this makes it appear nicer also, id probably leave this

src/script/auth/component/EntropyCanvas.tsx
tlebon:
```suggestion
  }, [clearInterval, percent]);
```

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -35,35 +35,36 @@ const EntropyCanvas = (props: CanvasProps) => {
...
   const [entropy, setEntropy] = useState<[number, number][]>([]);
   const frames = entropy.filter(Boolean).length;
+  // minimum number of frames
+  const MIN_FRAMES = 300;
+  // minimum duration in seconds
+  const MIN_DURATION = 30;
 
   const {clearInterval, startInterval, pauseInterval} = usePausableInterval(() => {
     // This prevents automatic batching of state updates from react 18. https://github.com/reactwg/react-18/discussions/21
     flushSync(() => {
```
tlebon: as far as i can tell, flushSync is no longer necessary. should be removed
```suggestion

```

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -35,35 +35,36 @@ const EntropyCanvas = (props: CanvasProps) => {
...
-      setPercent(percent => percent + 1);
+      if (lastFramesCount != frames) {
+        setTimePercent(timedPercent => timedPercent + 1);
+        setLastFramesCount(frames);
+        setPercent(Math.ceil(Math.min(timedPercent, (100 * frames) / MIN_FRAMES)));
+      }
     });
```
tlebon: the closing of flushSync should also be removed
```suggestion

```

src/script/auth/component/ProgressBar.tsx
comawill: With the transition, the change is horribly delayed, and the progress bar still moves for some time after the mouse movement was stopped. This is probably the fault of the many small incremental updates. Maybe there is a better solution, so I'm definitely open to suggestions.

src/script/auth/component/EntropyCanvas.tsx
comawill: ah... I understand, then this needs to be some how recognized when counting the frames, as this increments the size of the array without adding data

src/script/auth/component/ProgressBar.tsx
tlebon: maybe ill play around with this a bit and see if there's some better solution

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -19,89 +19,90 @@
...
+  // minimum number of frames
+  min_frames: number;
+  // minimum bits of overall estimated entropy
+  min_entropy_bits: number;
 }
 
 const EntropyCanvas = (props: CanvasProps) => {
-  const {sizeX, sizeY, onSetEntropy, onProgress, css, ...rest} = props;
-  const canvasRef = useRef(null);
+  const {sizeX, sizeY, onProgress, css, min_entropy_bits, min_frames, ...rest} = props;
```
tlebon: can we change these to camelCase
```suggestion
  const {sizeX, sizeY, onProgress, css, minEntropyBits, minFrames, ...rest} = props;
```

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -19,89 +19,90 @@
...
+  const [entropy] = useState<EntropyData>(new EntropyData());
+  const [previousPoint, setPreviousPoint] = useState<EntropyFrame | null>(null);
+  const [lastPoint, setLastPoint] = useState<EntropyFrame | null>(null);
 
   const {clearInterval, startInterval, pauseInterval} = usePausableInterval(() => {
-    // This prevents automatic batching of state updates from react 18. https://github.com/reactwg/react-18/discussions/21
-    flushSync(() => {
-      setPercent(percent => percent + 1);
-    });
+    setPercent(Math.floor(100 * Math.min(entropy.entropyBits / min_entropy_bits, entropy.length / min_frames)));
```
tlebon:
```suggestion
    setPercent(Math.floor(100 * Math.min(entropy.entropyBits / minEntropyBits, entropy.length / minFrames)));
```

src/script/auth/page/EntropyContainer.tsx
```diff
@@ -77,18 +77,18 @@ const EntropyContainer = ({onSetEntropy}: Props) => {
...
             css={{border: pause ? 'red 2px solid' : 'black 2px solid'}}
             data-uie-name="element-entropy-canvas"
-            sizeX={256}
-            sizeY={256}
-            onSetEntropy={onSetEntropy}
+            sizeX={512}
+            sizeY={512}
             onProgress={onProgress}
+            min_entropy_bits={1024}
+            min_frames={300}
```
tlebon:
```suggestion
            minEntropyBits={1024}
            minFrames={300}
```

src/script/strings.ts
```diff
@@ -292,17 +292,17 @@ export const setEmailStrings = defineMessages({
...
     id: 'setEntropy.headline',
   },
-  moreEntropyNeeded: {
-    defaultMessage: 'Please continue moving your mouse within the window to create enough entropy.',
-    id: 'setEntropy.moreEntropyNeeded',
-  },
   subheadline: {
     defaultMessage:
-      'Move your mouse as randomly as possible within the window below until the progress bar is 100% filled. This creates a random number which helps to protect your account and increases the overall security.',
+      'Move your mouse as randomly as possible within the window below until the progress bar is 100% filled. This increases the quality of the random numbers which are used to create your cryptographic long term secrets and therefore increases the overall security of your Wire client.',
```
tlebon, line 305: this text was previously cleared with content, thus new message may not be as understandable. @Astrid-Wire 
(keep in mind the en-us.json value will still appear over this)

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -19,89 +19,90 @@
...
   const onMouseMove = (event: MouseEvent<HTMLCanvasElement>) => {
     startInterval();
-    const newEntropy: [number, number] = [
-      event.pageX - event.currentTarget?.getBoundingClientRect()?.x,
-      event.pageY - event.currentTarget?.getBoundingClientRect()?.y,
-    ];
-    setEntropy(entropy => [...entropy, newEntropy]);
+    const boundingRect = event.currentTarget?.getBoundingClientRect();
+    const drawPoint: EntropyFrame = {
+      t: 0,
```
tlebon: is there a reason for t to be included in here besides its inclusion in the type? i suggest to make it an optional parameter in the type or use a different type perhaps.

src/script/util/Entropy.test.ts
```diff
@@ -0,0 +1,98 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {EntropyData, shannonEntropy, calculateDeltaValues} from './Entropy';
+
+//import {joaatHash} from './Crypto';
```
tlebon: maybe remove comment

src/script/util/Entropy.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (prevValue != null) {
+      result.push(Math.abs(value - prevValue));
+    }
+    prev[i] = value;
+  });
+  return new Uint8Array(result);
+}
+
+export class EntropyData {
+  readonly frames: EntropyFrame[];
```
tlebon, line 63: are we assuming that this will always be our entropy type? perhaps this may also have a generic option?

src/script/strings.ts
comawill: A proposal to change the translated text (en and de) was made on crowdin as well

src/script/util/Entropy.ts
comawill: At the moment yes. EntropyData and EntropyFrame are directly coupled especially for the assumption how to calculate the estimated entropy, so making this generic would require a different data model

src/script/auth/component/EntropyCanvas.tsx
comawill: it was mainly because of the type, made it optional now

src/i18n/en-US.json
```diff
@@ -1081,9 +1081,10 @@
   "setEmail.emailPlaceholder": "Email",
   "setEmail.headline": "Set email",
   "setEmail.tryAgain": "Try again",
+  "setEntropy.continue": "Continue",
   "setEntropy.headline": "Create Entropy",
   "setEntropy.moreEntropyNeeded": "Please continue moving your mouse within the window to create enough entropy.",
-  "setEntropy.subheadline": "Move your mouse as randomly as possible within the window below until the progress bar is 100% filled. This creates a random number which helps to protect your account and increases the overall security.",
+  "setEntropy.subheadline": "Move your mouse as randomly as possible within the window below until the progress bar is 100% filled. This increases the quality of the random numbers which are used to create your cryptographic long term secrets and therefore increases the overall security of your Wire client.",
```
tlebon: I'd still like @Astrid-Wire s comments on these

src/script/Config.ts
```diff
@@ -43,7 +43,8 @@ export class Configuration {
   readonly ENVIRONMENT = env.ENVIRONMENT || 'production';
   readonly FEATURE = {
     ...env.FEATURE,
-    ENABLE_EXTRA_CLIENT_ENTROPY: env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY && Runtime.isWindows(),
+    ENABLE_EXTRA_CLIENT_ENTROPY:
+      env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY && (Runtime.isWindows() || env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY_FORCE),
```
tlebon: Can we rename this? Enable and force together make less sense to me. 
```suggestion
      env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY && (Runtime.isWindows() || env.FEATURE.FORCE_EXTRA_CLIENT_ENTROPY),
```

src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -19,89 +19,89 @@
...
+  // minimum number of frames
+  minFrames: number;
+  // minimum bits of overall estimated entropy
+  minEntropyBits: number;
 }
 
 const EntropyCanvas = (props: CanvasProps) => {
-  const {sizeX, sizeY, onSetEntropy, onProgress, css, ...rest} = props;
-  const canvasRef = useRef(null);
+  const {sizeX, sizeY, onProgress, css, minEntropyBits, minFrames, ...rest} = props;
```
tlebon, line 40: I'm wondering if it makes sense to include a suggested/default value or JsDoc for any laymen who may want to use this in the future but has less understanding of bits or Frames

tlebon (approved):
Lgtm

src/script/auth/component/EntropyCanvas.tsx
tlebon: This may be covered in your comment on the interface

src/i18n/en-US.json
Astrid-Wire:
```suggestion
  "setEntropy.subheadline": "Move your mouse as randomly as possible within the window below until the progress bar is 100% filled. This increases the quality of the random numbers which are used to create your cryptographic long term secrets and therefore increases the overall security of this device.",
```

PR #13372, opened 2022-07-20
feat: improve tsconfig, betterer workflow to reduce betterer problems

aweiss-dev:
improve ts workflow


src/script/auth/module/action/AuthAction.ts
```diff
@@ -107,11 +107,10 @@ export class AuthAction {
...
         let entropyData: Uint8Array | undefined = undefined;
         if (getEntropy) {
+          entropyData = await getEntropy();
           try {
-            await core.loadAndValidateLocalClient();
-          } catch (e) {
-            entropyData = await getEntropy();
-          }
+            await core.loadAndValidateLocalClient(entropyData);
+          } catch (e) {}
```
tlebon: i dont think you should change this, it is set up like this for a reason

src/script/main/globals.ts
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2019 Wire Swiss GmbH
+ * Copyright (C) 2022 Wire Swiss GmbH
```
tlebon: staaahpp

src/types/window.d.ts
```diff
@@ -0,0 +1,143 @@
+/*
+ * Wire
+ * Copyright (C) 2019 Wire Swiss GmbH
```
tlebon: lol

src/types/window.d.ts
```diff
@@ -0,0 +1,143 @@
+/*
```
tlebon, line 1: please move all of these changes to a separate PR, they are unrelated to the topic at hand

.betterer.ts
```diff
@@ -20,8 +20,5 @@
 import {typescript} from '@betterer/typescript';
 
 export default {
-  'stricter compilation': () =>
-    typescript('./tsconfig.json', {
-      strict: true,
-    }).include('./src/**/*.{ts,tsx}'),
+  'stricter compilation': () => typescript('./tsconfig.json').include('./src/**/*.{ts,tsx}'),
```
tlebon, line 23: the things related to ts build i think should also be in a separate PR

src/script/auth/module/action/AuthAction.ts
```diff
@@ -385,7 +385,10 @@ export class AuthAction {
     return async (dispatch, getState, {core}) => {
       dispatch(AuthActionCreator.startValidateLocalClient());
       try {
-        await core.loadAndValidateLocalClient();
+        const {
+          authState: {entropy},
+        } = getState();
```
tlebon: i dont think you should assume the entropy has landed here, in several cases it is not added to the store currently. better to pass it into doInit/ or verify it has been added to the redux store in the Login and conversationJoin cases

src/script/auth/module/action/AuthAction.ts
aweiss-dev: i need the entropy to pass to the core function now. So I need to fetch it before the try statement

src/script/main/globals.ts
aweiss-dev: i am not doing this. its eslint :(

src/script/auth/module/action/AuthAction.ts
atomrc: I don't think you need an entropy source when you load the local client, it already has generated it's identity.  Only when you create a new one ou should could need custom entropy (just like it's done right now with Proteus devices). Or are there specific measures when loading created client ?

aweiss-dev: I just assumed that if you generated the client with the entropy, you will need to load it again with it to get the same identifier in the end.

atomrc: Yes they need to stay stable over time once created (to confirm with MLS specialists, but it should be the case)
The core API should probably not change and there should not be changes on the webapp side to support custom entropy

aweiss-dev: I appreciate your holiday response. But I don't understand :(. I need to change the webapp because entropy gets generated there (on windows through drawing) to pass it into the core.

It seems that I got the loadAndValidateLocalClient() function wrong and I don't need to pass entropy here.

atomrc: The webapp part is already implemented and the entropy is already given to the core, so the API is already there and existing (also the condition to display the entropy drawing field on windows). 
So here the change should only be passing thos entropy (that the webapp already gives to the core) to core-crypto and that should do the trick :)

.husky/pre-commit
```diff
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no-install lint-staged
-yarn betterer precommit --cache
+yarn betterer
```
tlebon, line 5: it may be the --cache flag only that is the issue idk.

package.json
```diff
@@ -235,12 +235,12 @@
...
     "test:app": "jest --forceExit",
     "test:server": "cd server && yarn test",
-    "test:types": "tsc --noEmit && cd server && tsc --noEmit",
+    "test:types": "tsc --project tsconfig.build.json --noEmit && cd server && tsc --noEmit",
     "translate:extract": "i18next-scanner 'src/{page,script}/**/*.{js,html,htm}'",
     "translate:merge": "formatjs extract --format './bin/translations_extract_formatter.js' --out-file './src/i18n/en-US.json' './src/script/strings.ts'",
     "translate:upload": "yarn translate:merge && ts-node ./bin/translations_upload.ts",
     "translate:download": "ts-node ./bin/translations_download.ts",
-    "betterer": "betterer",
+    "betterer": "betterer precommit",
```
tlebon, line 243: is there a use case for just betterer? i imagine for testing you don't want to run the precommit version

src/script/main/globals.ts
```diff
@@ -18,129 +18,7 @@
...
-import 'Components/Modals/GroupCreation/GroupCreationModal';
-import '../page/AppLock';
-import 'Components/modal';
-import 'Components/panel/ServiceDetails';
-import 'Components/panel/UserDetails';
-import 'Components/ServiceList';
-import 'Components/SearchInput';
-import 'Components/calling/CallingOverlayContainer';
-
-import 'Util/LocalizerUtil';
```
tlebon: i think maybe all these imports dont make as much sense in a d.ts file

package.json
aweiss-dev: i don't see one. Betterer without writing the file is kind of redundant. I would rather run yarn tsc --noEmit in that case

src/script/main/globals.ts
aweiss-dev: Now i get what you meant... I am on it

src/script/Config.ts
```diff
@@ -43,7 +43,8 @@ export class Configuration {
   readonly ENVIRONMENT = env.ENVIRONMENT || 'production';
   readonly FEATURE = {
     ...env.FEATURE,
-    ENABLE_EXTRA_CLIENT_ENTROPY: env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY && Runtime.isWindows(),
+    ENABLE_EXTRA_CLIENT_ENTROPY:
+      env.FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY && (this.ENVIRONMENT !== 'production' || Runtime.isWindows()),
```
atomrc: This is unrelated change (and the condition looks off)

tsconfig.build.json
```diff
@@ -0,0 +1,42 @@
+{
```
atomrc, line 1: Let's avoid having to maintain 2 separate ts config. 
This will not scale in the future

We should have a base config and, eventually, an override. Let's avoid duplication

aweiss-dev: good idea

src/script/Config.ts
aweiss-dev: Talked this one through with @tlebon. 

The change happens because the windows platform should only be a blocker when the app runs in production mode. While in development, I want to see the entropy generator as soon as I enable the flag, even when on a mac (our workstations).

atomrc: Maybe we could make it a faire flag for dev. This is going to be very annoying for us developers of we have to go through the process for every new device!!

Also, still this is unrelated to this PR. Shouldn't be here

aweiss-dev: gonna revert it for now

tlebon (approved):
seems to be working now

src/script/Config.ts
tlebon: > Maybe we could make it a faire flag for dev. This is going to be very annoying for us developers of we have to go through the process for every new device!!
> 
> Also, still this is unrelated to this PR. Shouldn't be here

I'm not clear why we would have to go thru the process for every new device? only if we have the flag enabled?

atomrc: Right, correct, there is already a flag, my mistake 😅🤦‍♂️
You can ignore my comment, gentlemen

PR #13394, opened 2022-07-27
runfix: Correctly save group id to database after group creation

thisisamir98:
Correctly save group id to database after group creation


src/script/entity/Conversation.ts
```diff
@@ -185,6 +186,13 @@ export class Conversation {
     this.team_id = undefined;
     this.type = ko.observable();
 
+    /**
+     * If a conversation has the groupId property it means that it
+     * is MLS protocol based as this property is for MLS conversations only.
+     * @returns boolean
+     */
+    this.isUsingMLSProtocol = !!this.groupId;
```
tlebon: just curious where we are using this?

thisisamir98: changed, we're gonna use protocol instead but it's gonna be used for example in calling to differentiate how we send calling messages or adding people into conversations

tlebon (approved):

PR #13422, opened 2022-08-01
runfix: Avoid duplicated system messages for MLS groups

thisisamir98:
Avoid duplicated system messages for MLS groups


src/script/components/MessagesList/index.tsx
```diff
@@ -59,6 +60,29 @@ interface MessagesListParams {
   showUserDetails: (user: User) => void;
 }
 
+const filterSpecificDuplicatedMemberMessages = (messages: MessageEntity[]) => {
+  const typesToFilter = ['conversation.member-join', 'conversation.group-creation'];
+  return messages.reduce<MessageEntity[]>((acc, message) => {
+    if (isMemberMessage(message)) {
+      const existingMemberMessages = acc.filter(isMemberMessage);
```
tlebon, line 67: some of this reduce feels a bit confusing to me. for instance arent these two lines checking the same thing?

src/script/components/MessagesList/index.tsx
```diff
@@ -59,6 +60,29 @@ interface MessagesListParams {
...
+      if (typesToFilter.includes(message.type) && existingMemberMessages.length) {
+        switch (message.type) {
+          case 'conversation.group-creation':
```
tlebon, line 70: is a group-creation a memberMessage? seems it wouldn't be.

src/script/components/MessagesList/index.tsx
```diff
@@ -59,6 +60,29 @@ interface MessagesListParams {
...
+            return acc;
+          case 'conversation.member-join':
+            return existingMemberMessages.some(m => m.htmlCaption() === message.htmlCaption())
```
tlebon, line 73: dont we have a lint rule against short vars?

src/script/components/MessagesList/index.tsx
aweiss-dev: they arent.

We only want to execute code if the currently processed message is from type MemberMessage (the messages array contains different Message Types eg Member, Content...)

The second line filters MemberMessages from the accumulated messages array => since different types are not relevant for further code

src/script/components/MessagesList/index.tsx
aweiss-dev: It is. I am also super confused about the type MemberMessage, since it's more used like a SystemMessage. But we don't have a subtype SystemMessages :?

src/script/components/MessagesList/index.tsx
aweiss-dev: Doesn't seem like :D. To be fair this is code that github copilot generated.
could also call it existingMemberMessage but since its a fairly short statement I thought m is fine

src/script/components/MessagesList/index.tsx
```diff
@@ -59,6 +60,29 @@ interface MessagesListParams {
...
+              ? acc
+              : [...acc, message];
+          default:
```
tlebon, line 76: this line seems unnecessary?

tlebon (approved):
fine for me

PR #13425, opened 2022-08-02
refactor: rename hook and fix a couple of typing errors

aweiss-dev:
rename hook and fix a couple of typing errors


src/script/conversation/ConversationRepository.ts
```diff
@@ -616,27 +617,36 @@ export class ConversationRepository {
...
+    ): message is MemberMessage => message.super_type === SuperType.MEMBER;
+
     const mappedMessageEntities = await this.addEventsToConversation(events, conversationEntity);
     conversationEntity.hasAdditionalMessages(hasAdditionalMessages);
     if (!hasAdditionalMessages) {
-      const firstMessage = conversationEntity.getFirstMessage() as MemberMessage;
-      const checkCreationMessage = firstMessage?.isMember() && firstMessage.isCreation();
-      if (checkCreationMessage) {
-        const groupCreationMessageIn1to1 = conversationEntity.is1to1() && firstMessage.isGroupCreation();
-        const one2oneConnectionMessageInGroup = conversationEntity.isGroup() && firstMessage.isConnection();
```
aweiss-dev, line 626: This needed a change because we couldnt be sure that the first message is always defined / a member message. isCreation() has been undefined for some reloads which caused the if statements to break -> unwanted side effects

src/script/conversation/ConversationRepository.ts
```diff
@@ -616,27 +617,36 @@ export class ConversationRepository {
...
-
-        if (wrongMessageTypeForConversation) {
-          this.messageRepository.deleteMessage(conversationEntity, firstMessage);
-          conversationEntity.hasCreationMessage = false;
-        } else {
-          conversationEntity.hasCreationMessage = true;
-        }
-      }
+      const allMessages = conversationEntity.getAllMessages();
+      if (!!allMessages.length) {
```
przemvs: I think you don't need a cast !!allMessages.length, because allMessages always be an array :) And if check if it's variable is truthy/falsy value.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -80,8 +80,8 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
     if (isFileSharingReceivingEnabled) {
       setDisplaySmall(false);
 
-      if (videoSrc) {
-        videoElement.play();
+      if (!!videoSrc.length) {
```
przemvs: You don't need to cast !! for it, because it always be a string.

src/script/conversation/ConversationRepository.ts
```diff
@@ -616,27 +617,36 @@ export class ConversationRepository {
   ): Promise<ContentMessage[]> {
     const hasAdditionalMessages = events.length === Config.getConfig().MESSAGES_FETCH_LIMIT;
 
+    const isMemberMessage = (
```
przemvs: I think you can move it outside the class.

przemvs (changes requested):

src/script/conversation/ConversationRepository.ts
aweiss-dev: thats a typeguard that is only used inside this function.

Do we have a place where typegards can live? If not we may need to create one (a folder) because we need to implement typeguards more often in the future.

src/script/conversation/ConversationRepository.ts
aweiss-dev: Gonna remove it. I am just used to cast length checks because of react (jsx) :D

src/script/components/MessagesList/Message/ContentMessage/asset/AudioAsset.tsx
```diff
@@ -57,19 +56,22 @@ const AudioAsset: React.FC<AudioAssetProps> = ({
...
+  const audio = audioElement.current;
+
+  const onTimeupdate = () => setAudioTime(audio?.currentTime ?? 0);
+  const showLoudnessPreview = !!asset.meta?.loudness?.length;
+  const onPauseButtonClicked = () => audioElement.current?.pause();
 
   const onPlayButtonClicked = async () => {
     if (audioSrc) {
-      audioElement?.play();
+      audioElement.current?.play();
```
przemvs: Hmm.. upper you have variable with audio, so maybe use here also :)

src/script/components/MessagesList/Message/ContentMessage/asset/AudioAsset.tsx
```diff
@@ -57,19 +56,22 @@ const AudioAsset: React.FC<AudioAssetProps> = ({
...
   const [audioSrc, setAudioSrc] = useState<string>();
-  const onTimeupdate = () => setAudioTime(audioElement.currentTime);
-  const showLoudnessPreview = !!(asset.meta?.loudness?.length > 0);
-  const onPauseButtonClicked = () => audioElement?.pause();
+
+  const audio = audioElement.current;
+
+  const onTimeupdate = () => setAudioTime(audio?.currentTime ?? 0);
+  const showLoudnessPreview = !!asset.meta?.loudness?.length;
+  const onPauseButtonClicked = () => audioElement.current?.pause();
```
przemvs: Hmm.. upper you have variable with audio, so maybe use here also :)

src/script/components/MessagesList/Message/ContentMessage/asset/AudioAsset.tsx
```diff
@@ -84,7 +86,7 @@ const AudioAsset: React.FC<AudioAssetProps> = ({
 
   useEffect(() => {
     if (audioSrc && audioElement) {
-      const playPromise = audioElement.play();
+      const playPromise = audioElement.current?.play();
```
przemvs, line 89: Hmm.. upper you have variable with audio, so maybe use here also :)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -141,66 +142,82 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
+                onTimeUpdate={() => {
+                  if (videoElement.current) {
+                    setVideoTimeRest(videoElement.current?.duration - videoElement.current?.currentTime);
+                  }
+                }}
+                onLoadedMetadata={() => {
+                  if (videoElement.current) {
+                    setVideoTimeRest(videoElement.current?.duration - videoElement.current?.currentTime);
+                  }
+                }}
```
przemvs: You can prepare function for it, you using it multiple times

src/script/components/MessagesList/index.tsx
```diff
@@ -232,7 +233,7 @@ const MessagesList: React.FC<MessagesListParams> = ({
   });
 
   return (
-    <div ref={setContainer} className={`messages ${verticallyCenterMessage() ? 'flex-center' : ''}`}>
+    <div ref={messagesContainerElement} className={`messages ${verticallyCenterMessage() ? 'flex-center' : ''}`}>
```
przemvs: Please use classnames for multiple classes.

src/script/conversation/ConversationRepository.ts
przemvs: Let's go with it :)

przemvs (changes requested):

src/script/components/MessagesList/Message/ContentMessage/asset/AudioAsset.tsx
aweiss-dev: its done on purpose here since it is in a useEffect and audio get newly generated each render while audioelement is stateful

src/script/components/MessagesList/index.tsx
aweiss-dev: not a big fan of classnames because libs like cslx are more versatile and if I just need a class string, string templates are a very good option.

But since we have the lib in use anyways....

src/script/view_model/panel/TimedMessagesPanel.tsx
```diff
@@ -78,10 +77,9 @@ const TimedMessagesPanel: React.FC<TimedMessagesPanelProps> = ({onClose, onGoBac
   }, [globalMessageTimer]);
 
   const timedMessageChange = (value: number): void => {
-    if (activeConversation) {
-      const finalTimer = value === 0 ? null : value;
-      activeConversation.globalMessageTimer(finalTimer);
-      repositories.conversation.updateConversationMessageTimer(activeConversation, finalTimer);
+    if (activeConversation && value !== 0) {
```
przemvs, line 80:
```suggestion
    if (activeConversation && value) {
```

Can be simplified :)

przemvs (changes requested):
1 thing left, lgtm :)

src/script/view_model/panel/TimedMessagesPanel.tsx
aweiss-dev: let's keep it specific here for clarity

PR #13373, opened 2022-07-20
refactor: migrate input-bar to react (#ACC-202)

przemvs:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-202" title="ACC-202" target="_blank">TaskACC-202</a>  [Web] Migrate input-bar component to React
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

_Briefly describe the issue you have solved or implemented with this pull request. If the PR contains multiple issues, use a bullet list._

### Causes (Optional)

_Briefly describe the causes behind the issues. This could be helpful to understand the adopted solutions behind some nasty bugs or complex issues._

### Solutions

_Briefly describe the solutions you have implemented for the issues explained above._

### Dependencies (Optional)

_If there are some other pull requests related to this one (e.g. new releases of frameworks), specify them here._

Needs releases with:

- [ ] GitHub link to other pull request

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Notes (Optional)

_Specify here any other facts that you think are important for this issue._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [ ] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #13445, opened 2022-08-09
feat: change mls boolean switcher to protocol select (FS-700) 

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-700" title="FS-700" target="_blank">TaskFS-700</a>  [Web] New Design for protocol selection for new Group 
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Change mls boolean switcher to protocol select

----

----

----
##### References


src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -70,17 +72,37 @@ enum GroupCreationModalState {
...
 }) => {
+  const protocolOptions: ProtocolOption[] = [ConversationProtocol.PROTEUS, ConversationProtocol.MLS].map(protocol => ({
+    label: t(`modalCreateGroupProtocolSelect.${protocol}`),
+    value: protocol,
+  }));
+
+  const initialProtocol: ProtocolOption = {
+    label: t('modalCreateGroupProtocolSelect.proteus'),
+    value: ConversationProtocol.PROTEUS,
+  };
```
przemvs: Use protocolOptions.find(protocol => protocol.value === ConversationProtocol.PROTEUS) :)

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -225,187 +247,205 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
+              placeholder={t('groupCreationParticipantsPlaceholder')}
+              enter={clickOnCreate}
+            />
+          )}
+
+          {stateIsParticipants && (
+            <div className="group-creation__list" ref={setScrollbarRef}>
+              {getContacts().length > 0 && (
+                <UserSearchableList
+                  users={getContacts()}
```
przemvs: Move getContacts() to variable ;)

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -70,17 +72,37 @@ enum GroupCreationModalState {
 
 const logger = getLogger('GroupCreationModal');
 
+interface ProtocolOption {
+  label: string;
+  value: ConversationProtocol;
+}
+
+function isProtocolOption(option?: Option | null): option is ProtocolOption {
+  const protocols = Object.values(ConversationProtocol) as string[];
```
przemvs: I think you can move it outside isProtocolOption. And move this typeguard to scripts/guards/<e.g: Protocol.ts>.

przemvs (changes requested):
Few changes ;)

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -53,6 +54,7 @@ import {
 import useEffectRef from 'Util/useEffectRef';
 import {useFadingScrollbar} from '../../../ui/fadingScrollbar';
 import {Config} from '../../../Config';
+import {theme} from '../../../page/MainContent/MainContent';
```
V-Gira: I'm making a note of this. The theme is going to live in the ui-kit soon. I'll make sure to update this

PatrykBuniX: Sounds good, thanks!

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: used `useMemo`

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: does it makes sense? It's being used only once in one file?

przemvs: Maybe you will use it later, good to have all typeguards in one place. Check there structure:
https://github.com/wireapp/wire-webapp/pull/13373/files#diff-ff5e3819c325e9820395a1d0f449dfbe21a346b505bad93db411c7ad9926ccc6

src/script/components/ModalComponent.tsx
```diff
@@ -150,7 +152,7 @@ const ModalComponent: React.FC<ModalComponentProps> = ({
           role="button"
           tabIndex={-1}
           onKeyDown={noop}
-          css={hasVisibleClass ? ModalContentVisibleStyles : ModalContentStyles}
+          css={{...(hasVisibleClass ? ModalContentVisibleStyles : ModalContentStyles), ...wrapperCSS}}
```
przemvs, line 155: I'm thinking about it, you replace by wrapperCSS previous css styles. It should be before or after this modalcontent styles ?

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: ok

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -17,11 +17,12 @@
  *
  */
 
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useState, useMemo} from 'react';
 import {container} from 'tsyringe';
 import cx from 'classnames';
 import {amplify} from 'amplify';
 import {WebAppEvents} from '@wireapp/webapp-events';
+import {Select, StyledApp} from '@wireapp/react-ui-kit';
```
V-Gira:
```suggestion
import {Select, StyledApp, THEME_ID} from '@wireapp/react-ui-kit';
```
The change to the ui-kit is live on dev

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -53,6 +54,8 @@ import {
 import useEffectRef from 'Util/useEffectRef';
 import {useFadingScrollbar} from '../../../ui/fadingScrollbar';
 import {Config} from '../../../Config';
+import {theme} from '../../../page/MainContent/MainContent';
```
V-Gira:
```suggestion
```

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -225,187 +235,205 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
-                users={getContacts()}
-                filter={participantsInput}
-                selected={selectedContacts}
-                onUpdateSelectedUsers={setSelectedContacts}
-                searchRepository={searchRepository}
-                teamRepository={teamRepository}
-                conversationRepository={conversationRepository}
-                noUnderline
-              />
+    <StyledApp theme={theme}>
```
V-Gira:
```suggestion
    <StyledApp themeId={THEME_ID.DEFAULT}>
```

V-Gira (commented):
I made a couple of suggestion so the component is in line with the Theme update on dev

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: Shouldn't it be `themes[THEME_ID.DEFAULT]` ? `themes` imported from ui-kit also

V-Gira: you could do `theme={themes[THEME_ID.DEFAULT]}`
but we export the `THEME_ID` enum and we have a `themeId` prop

PatrykBuniX: right, didn't spot the propname change here 😄

src/script/page/MainContent/MainContent.tsx
```diff
@@ -51,6 +51,47 @@ const Animated: React.FC<{children: React.ReactNode}> = ({children, ...rest}) =>
...
+    disabledColor: 'var(--text-input-placeholder)',
+  },
+  general: {
+    backgroundColor: 'var(--app-bg)',
+    color: 'var(--main-color)',
+    dangerColor: 'var(--res-500)',
+    primaryColor: 'var(--accent-color)',
+  },
+};
+
```
V-Gira: I somehow missed that massive green blob :D. The theme lives in ui-kit now, you can remove it from MainContent

PatrykBuniX: ok, removed it

V-Gira (approved):

tlebon (approved):

src/page/template/panel/group-participant-user.htm
```diff
@@ -15,16 +15,22 @@
...
       <!-- /ko -->
       <!-- ko if: canChangeRole() -->
-        <div class="panel__action-item conversation-details__admin" data-uie-name="toggle-admin">
-          <group-admin-icon aria-hidden="true" class="panel__action-item__icon"></group-admin-icon>
-          <base-toggle
+        <div class="conversation-details__admin">
+          <button class="panel__action-item" 
+          data-bind="click: onToggleAdmin, attr: {'aria-label': t('accessibility.conversationDetailsActionGroupAdminLabel')}" 
+          data-uie-name="toggle-admin" type="button"
+          aria-pressed={isAdmin()}>
```
przemvs, line 31: It shouldn't be in data-bind ?

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -225,187 +234,205 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
+                  onBlur={event => {
+                    const {value} = event.target as HTMLInputElement;
+                    const trimmedName = value.trim();
+                    setGroupName(trimmedName);
+                  }}
+                  onKeyDown={(event: React.KeyboardEvent<HTMLInputElement>) => {
+                    handleEnterDown(event, clickOnNext);
+                  }}
+                  value={groupName}
+                  isError={nameError.length > 0}
```
przemvs: Move nameError.length > 0 to variable and replace other places where you're using.

src/script/guards/ProtocolOption.ts
```diff
@@ -0,0 +1,31 @@
+/*
```
przemvs: I'm thinking about name. Maybe should be Protocol instead of ProtocolOption, because we will store all Protocol guards there.

src/script/guards/ProtocolOption.ts
```diff
@@ -0,0 +1,31 @@
...
+import {Option} from '@wireapp/react-ui-kit';
+
+export interface ProtocolOption {
+  label: string;
+  value: ConversationProtocol;
+}
+
+export function isProtocolOption(option?: Option | null): option is ProtocolOption {
+  const protocols = Object.values(ConversationProtocol) as string[];
+  return !!option && typeof option.value === 'string' && protocols.includes(option.value);
```
przemvs: It doesn't work?

```suggestion
  return typeof option?.value === 'string' && protocols.includes(option.value);
```

src/script/guards/ProtocolOption.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import {ConversationProtocol} from '@wireapp/api-client/src/conversation/NewConversation';
+import {Option} from '@wireapp/react-ui-kit';
+
+export interface ProtocolOption {
+  label: string;
+  value: ConversationProtocol;
+}
+
+export function isProtocolOption(option?: Option | null): option is ProtocolOption {
```
przemvs: I'll go with arrow fn, in other guards we use arrow fn. Also parameter should be any | undefined | null :) Like in other cases.

przemvs (changes requested):
Few changes. :)

src/script/components/ModalComponent.tsx
PatrykBuniX: It comes after the default styles so we can overwrite it if we want to.

src/page/template/panel/group-participant-user.htm
PatrykBuniX: i don't think so, this is not a change I made, probably came from dev merge

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: its being used 2 times only, isn't it more readable this way? What should the variable be called? 🤔

przemvs: For >= 2 usages is better to have variable. E.g. hasNameError :)

src/script/guards/ProtocolOption.ts
PatrykBuniX: Screenshot 2022-08-22 at 15 44 15
but `any | undefined | null` is equal to just `any` 🤔

PatrykBuniX: `undefined` comes from `?` idk if we really want to allows to pass anything (`any`) here... Its being used in select options so it has to always be at least `Option | null`, WDYT?

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
PatrykBuniX: ok

src/script/guards/ProtocolOption.ts
przemvs: Hmm... what if you will use it in different case ? And it get not an Option ?

PatrykBuniX: so I guess just `option: any` is enough?

przemvs (approved):

PR #13449, opened 2022-08-10
feat: Integrate MLS with AVS library

thisisamir98:
Integrate MLS with AVS library


src/script/calling/enum/CallMessageType.ts
```diff
@@ -19,7 +19,8 @@
 
 export enum CALL_MESSAGE_TYPE {
   CANCEL = 'CANCEL',
-  CONF_START = 'CONFSTART',
+  CONF_START = 'CONFEND',
+  CONF_END = 'CONFSTART',
```
V-Gira, line 23: Looks like a deliberate change, just want to make sure it's not something that got mixed up

tlebon: hm this is strange. id guess typo

src/script/calling/CallingRepository.test.ts
```diff
@@ -178,7 +201,7 @@ describe('CallingRepository', () => {
       const call = new Call(
         {domain: '', id: ''},
         createConversationId(),
-        undefined,
+        CONV_TYPE.CONFERENCE,
```
tlebon, line 204: just curious why all of these are changing to conference? any insight from your side

test/helper/TestFactory.js
```diff
@@ -333,6 +333,8 @@ export class TestFactory {
   async exposeCallingActors() {
     await this.exposeConversationActors();
     this.calling_repository = new CallingRepository(
+      // @ts-ignore
```
tlebon, line 336: why is this necessary?

src/script/calling/CallingRepository.test.ts
thisisamir98: fixes strict mode errors

src/script/calling/enum/CallMessageType.ts
thisisamir98: yes, a typo, thanks for catching, fixed in https://github.com/wireapp/wire-webapp/pull/13449/commits/1d2e697c422bc6d209e1cc50d765821556eab373

test/helper/TestFactory.js
thisisamir98: strict mode error, already was there because of this.message_repository, now that the order is changed betterer thinks there is a new error added but it is an already existing old one

PR #13486, opened 2022-08-15
feat: Integrate MLS with AVS library

thisisamir98:
feat: Integrate MLS with AVS library


src/script/conversation/EventBuilder.ts
```diff
@@ -54,7 +54,21 @@ export interface ConversationEvent<T> extends BaseEvent {
 }
 
 export interface CallingEvent {
-  content: CallEntity;
+  /**
+   * content is an object that comes from avs
+   */
+  content: {
```
atomrc, line 60: Nice to have clean typing for this 🙏

src/script/calling/CallingRepository.ts
```diff
@@ -546,18 +546,32 @@ export class CallingRepository {
...
+      sender: clientId,
+      time = new Date().toISOString(),
+    } = event;
     const isFederated = this.core.backendFeatures.isFederated && qualified_conversation && qualified_from;
     const userId = isFederated ? qualified_from : {domain: '', id: from};
     const conversationId = isFederated ? qualified_conversation : {domain: '', id: conversation};
     const currentTimestamp = this.serverTimeHandler.toServerTimestamp();
     const toSecond = (timestamp: number) => Math.floor(timestamp / 1000);
     const contentStr = JSON.stringify(content);
+    const conversationEntity = this.conversationState.findConversation(conversationId);
```
atomrc, line 567: Let's get rid of the `<stuff>Entity` convention. Let's just call it `conversation`
```suggestion
    const conversation = this.conversationState.findConversation(conversationId);
```

src/script/calling/CallingRepository.ts
```diff
@@ -571,11 +585,12 @@ export class CallingRepository {
...
       case CALL_MESSAGE_TYPE.CONFKEY: {
         if (source !== EventRepository.SOURCE.STREAM) {
-          const {id, domain} = conversationId;
-          const allClients = await this.core.service!.conversation.getAllParticipantsClients(id, domain);
-          // We warn the message repository that a mismatch has happened outside of its lifecycle (eventually triggering a conversation degradation)
+          const allClients = await this.clientRepository.getQualifiedClientsByUserIds(
+            conversationEntity.participating_user_ids(),
+            false,
+          );
           const shouldContinue = await this.messageRepository.updateMissingClients(
```
atomrc, line 595:
```suggestion
          // We warn the message repository that a mismatch has happened outside of its lifecycle (eventually triggering a conversation degradation)
          const shouldContinue = await this.messageRepository.updateMissingClients(
```

please don't remove useful comments

src/script/calling/CallingRepository.ts
```diff
@@ -299,28 +297,30 @@ export class CallingRepository {
...
+    const conversation = this.conversationState.findConversation(call.conversationId);
+    if (!conversation) {
+      this.logger.warn(`Unable to find a conversation with id of ${call.conversationId}`);
+      return false;
+    }
+    const allClients = await this.clientRepository.getQualifiedClientsByUserIds(
+      conversation.participating_user_ids(),
+      false,
+    );
+    const qualifiedClients = flattenQualifiedUserClients(allClients);
```
atomrc: Can you provide explanation to why we need to change this. 
I'd rather have the logic improved on the `core` side if needed rather than reimplementing methods in the webapp. 

What are the exact constraints here?

atomrc: Maybe fixing the `core` method to work with both MLS and proteus would be nice

atomrc (changes requested):

src/script/calling/CallingRepository.ts
thisisamir98: image

atomrc (approved):

PR #13548, opened 2022-08-25
fix: address hover and focus state of sidebar call-ui acc-80

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-80" title="ACC-80" target="_blank">SubtaskACC-80</a>  Calling UI conversation view
  </td></table>
  <br />
 

- add correct hover and focus state to call-ui sidebar
- remove focus from dropdown arrow
- open context menu when targeting the sidebar item (mouse or keyboard.)

image

### DIFF
--- a/src/script/components/list/ConversationListCallingCell.tsx
+++ b/src/script/components/list/ConversationListCallingCell.tsx
@@ -454,7 +454,7 @@ const ConversationListCallingCell: React.FC<CallingCellProps> = ({
                       .slice()
                       .sort((participantA, participantB) => sortUsersByPriority(participantA.user, participantB.user))
                       .map(participant => (
-                        <li key={participant.clientId}>
+                        <li key={participant.clientId} className="call-ui__participant-list__participant">
                           <ParticipantItem
                             key={participant.clientId}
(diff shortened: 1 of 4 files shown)


src/script/components/list/ConversationListCallingCell.tsx
```diff
@@ -148,7 +148,7 @@ const ConversationListCallingCell: React.FC<CallingCellProps> = ({
   const [showParticipants, setShowParticipants] = useState(false);
   const isModerator = roles[selfUser?.id] === DefaultConversationRoleName.WIRE_ADMIN;
 
-  const getParticipantContext = (event: React.MouseEvent<HTMLDivElement>, participant: Participant) => {
+  const getParticipantContext = (event: React.MouseEvent<HTMLDivElement> | MouseEvent, participant: Participant) => {
```
PatrykBuniX: is it necessary?

V-Gira: it isn't good catch!

PatrykBuniX (approved):

PR #13549, opened 2022-08-26
feat:  user leaves mls group (FS-683)

aweiss-dev:
feat:  user leaves mls group (FS-683)


atomrc:
superseeded by https://github.com/wireapp/wire-webapp/pull/13692/files

PR #13551, opened 2022-08-27
chore: Fix Message components type errors

atomrc:
This will fix typescript error around the Message rendering components (mesages and assets)

Bring the total number of strict type errors from 2062 to 2039


src/script/components/MessagesList/Message/ContentMessage/MessageQuote.tsx
```diff
@@ -162,7 +164,7 @@ const QuotedMessage: React.FC<QuotedMessageProps> = ({
   ]);
   const [canShowMore, setCanShowMore] = useState(false);
   const [showFullText, setShowFullText] = useState(false);
-  const [textQuoteElement, setTextQuoteElement] = useEffectRef();
+  const [textQuoteElement, setTextQuoteElement] = useEffectRef<HTMLDivElement>();
```
przemvs, line 167: Maybe move to useRef instead of useEffectRef ;)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +128,102 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
-                setHideControls(false);
+  return !isObfuscated ? (
+    <div className="video-asset" data-uie-name="video-asset" data-uie-value={asset.file_name}>
+      {!isFileSharingReceivingEnabled ? (
+        <RestrictedVideo />
+      ) : (
+        <>
+          <div
+            className={cx('video-asset__container', {'video-asset__container--small': displaySmall})}
+            onPointerLeave={() => startTimeout()}
```
przemvs:
```suggestion
            onPointerLeave={startTimeout}
```

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +128,102 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
-                            data-uie-name="status-video-time"
-                          >
-                            {formatSeconds(videoTimeRest)}
-                          </span>
-                        </div>
+              onPlaying={onVideoPlaying}
+              onTimeUpdate={() => videoElement && setVideoTimeRest(videoElement.duration - videoElement.currentTime)}
+              onLoadedMetadata={() =>
+                videoElement && setVideoTimeRest(videoElement.duration - videoElement.currentTime)
+              }
```
przemvs: Move this to function, you have duplicated it :)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +128,102 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
+              style={{
+                backgroundColor: videoPreview ? '#000' : '',
+                visibility: transferState === AssetTransferState.UPLOADING ? 'hidden' : undefined,
```
przemvs: Do smth like that :)
```suggestion
                {...(transferState === AssetTransferState.UPLOADING && { visibility: 'hidden'}),
```

przemvs: Also maybe move this transferState === AssetTransferState.UPLOADING to variable, you have used it multiple times :)

src/script/components/MessagesList/Message/ContentMessage/asset/index.tsx
```diff
@@ -113,11 +113,16 @@ const ContentAsset = ({
     case AssetType.LOCATION:
       return <LocationAssetComponent asset={asset as Location} />;
     case AssetType.BUTTON:
+      if (!(message instanceof CompositeMessage && asset instanceof Button && asset.id)) {
+        return null;
+      }
+      const assetId = asset.id;
```
przemvs: Maybe use this variable above if statement :) You can use it also in this if variable :)

src/script/ui/viewportObserver.ts
```diff
@@ -114,9 +114,12 @@ export const viewportObserver = {
   trackElement,
 };
 
-export const useViewPortObserver = (elementRef: HTMLElement, defaultIsVisible: boolean = false): boolean => {
+export const useViewPortObserver = (elementRef?: HTMLElement, defaultIsVisible: boolean = false): boolean => {
   const [isInViewport, setIsInViewport] = useState(defaultIsVisible);
   useEffect(() => {
+    if (!elementRef) {
+      return () => {};
```
przemvs, line 121: Only return doesn't work ? :)

src/script/ui/viewportObserver.ts
```diff
@@ -114,9 +114,12 @@ export const viewportObserver = {
   trackElement,
 };
 
-export const useViewPortObserver = (elementRef: HTMLElement, defaultIsVisible: boolean = false): boolean => {
+export const useViewPortObserver = (elementRef?: HTMLElement, defaultIsVisible: boolean = false): boolean => {
```
przemvs, line 117: Hmm.. defaultIsVisible shouldn't be ?: boolean ? 🤔

przemvs (changes requested):
Few changes :)

thisisamir98 (approved):

src/script/components/MessagesList/Message/ContentMessage/MessageQuote.tsx
atomrc: We need effect ref in this case, since we are listening to the object change and trigger an effect when it changes.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
atomrc: True that, good call 👍

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
atomrc: What is the benefit of this method, I don't see any benefit, I personally find it harder to read 🤔

src/script/ui/viewportObserver.ts
atomrc: It has a default value, so the `?` is not needed

src/script/ui/viewportObserver.ts
atomrc: Doesn't typecheck (cause the function has mixed return types then I guess). Let me try to see what I can do

atomrc: actually it does typecheck, good call, thanks :)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
przemvs: If you use it you don't need to pass undefined, and for me it's better than passing undefined :)

atomrc: Fixed with `className` :)

PatrykBuniX (approved):

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +129,100 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
-          <RestrictedVideo />
-        ) : (
-          <>
-            <div
-              className={cx('video-asset__container', {'video-asset__container--small': displaySmall})}
-              onPointerLeave={() => startTimeout()}
-              onPointerEnter={() => {
-                removeTimeout();
-                setHideControls(false);
+  const syncVideoTimeRest = () => videoElement && setVideoTimeRest(videoElement.duration - videoElement.currentTime);
```
przemvs: Please make if condition. It will be more readable.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +129,100 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
-                            className="video-asset__controls__bottom__time label-xs"
-                            data-uie-name="status-video-time"
-                          >
-                            {formatSeconds(videoTimeRest)}
-                          </span>
-                        </div>
+              onPlaying={onVideoPlaying}
+              onTimeUpdate={syncVideoTimeRest}
+              onLoadedMetadata={syncVideoTimeRest}
+              className={clx({hidden: transferState === AssetTransferState.UPLOADING})}
```
przemvs: Move transferState === AssetTransferState.UPLOADING to variable, you using this multiple times.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +129,100 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
+                        />
+                      ) : (
+                        <MediaButton
+                          mediaElement={videoElement}
+                          large
+                          asset={asset}
+                          play={onPlayButtonClicked}
+                          pause={onPauseButtonClicked}
+                          cancel={() =>
+                            transferState === AssetTransferState.UPLOADING ? cancelUpload() : asset.cancelDownload()
```
przemvs: Move transferState === AssetTransferState.UPLOADING to variable, you using this multiple times.

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -77,21 +78,24 @@ const Message: React.FC<
...
-  const markerType = getMessageMarkerType(message, previousMessage, lastReadTimestamp);
+  const markerType = getMessageMarkerType(message, lastReadTimestamp, previousMessage);
 
   useLayoutEffect(() => {
+    if (!messageElementRef.current) {
+      return;
+    }
     if (isMarked) {
-      props.scrollTo({center: true, element: messageElementRef.current});
+      props.scrollTo?.({center: true, element: messageElementRef.current});
```
przemvs: Why not to spread scrollTo property ? In this component you have spreaded different props.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -116,102 +129,100 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
+                          }
+                          transferState={transferState}
+                          uploadProgress={uploadProgress}
+                        />
```
przemvs: Maybe prepare commonProps for this both MediaButtons ? Or you can use only one compontn <MediaButton /> by passing large={!displaySmall} ?

przemvs (changes requested):
Few more comments :)

src/script/components/MessagesList/Message/index.tsx
atomrc: fair point :)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
atomrc: oh and I realized that the `useAssetTransfer` hook already contains those variables, good catch ;)

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
atomrc: makes sense, actually I can only switch the `large` property true/false and it works as expected, no need for this condition at all, good catch

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
atomrc: done

przemvs (approved):
LGTM! :)

PR #13573, opened 2022-08-29
runfix: don't try to fetch user without id

PatrykBuniX:
Same as title - no need to try to call fetch user by id function when no id is provided (fixes bug with connection request messages entities)


V-Gira (approved):
Straightforward

przemvs (approved):

thisisamir98 (approved):

PR #13576, opened 2022-08-30
refactor: Wrap entire application in React

atomrc:
This will completely wrap the messenger application into a react wrapper. 

This will allow us to quickly:
- join leave components that are already migrated to React to the root react wrapper
- start using context available to the entire app (not just isolated react trees) for repositories and UIKit config
- start using a state handler at the root level of the app
- start using react-router for routes
- start uniting login page with the rest of the app


src/style/foundation/main.less
```diff
@@ -49,7 +49,6 @@
 
 @import 'base';
 @import 'framework';
-@import 'loading-screen';
```
atomrc, line 52: We have the styles.ts file for this one now

src/script/view_model/ListViewModel.ts
```diff
@@ -144,8 +142,6 @@ export class ListViewModel {
     });
 
     this._initSubscriptions();
-
-    ko.applyBindings(this, document.getElementById(this.elementId));
```
atomrc, line 148: no more querying the DOM from the view model, the template takes care of binding things together (but this will be deleted soonish anyways ;) )

src/script/main/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+import {SIGN_OUT_REASON} from '../auth/SignOutReason';
+import {ClientType} from '@wireapp/api-client/src/client/';
+import {Config} from '../Config';
+import {APIClient} from '../service/APIClientSingleton';
+import {Core} from '../service/CoreSingleton';
+import {StorageKey} from '../storage';
+import {App, doRedirect} from './app';
+import {AppContainer} from './components/AppContainer';
+
+$(async () => {
```
atomrc: This is now the index file that does the job of doing all the init side effects.

atomrc:
Closed for now. Will be done later once the rest of the application is migrated to React

PR #13582, opened 2022-09-02
runfix: correct focus states of toggle dots acc-81

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-81" title="ACC-81" target="_blank">SubtaskACC-81</a>  Calling UI fullscreen view
  </td></table>
  <br />
 

- assign the correct style for focus state of pagination and device select toggle dots
- refactor device toggle button to allow proper keyboard navigation

### DIFF
--- a/src/script/components/calling/DeviceToggleButton.tsx
+++ b/src/script/components/calling/DeviceToggleButton.tsx
@@ -42,60 +42,60 @@ const DeviceToggleButton: React.FC<DeviceToggleButtonProps> = ({currentDevice, d
       data-uie-name="device-toggle-button"
       css={css`
         align-items: center;
-        flex-direction: column;
         color: #fff;
         cursor: pointer;
         display: inline-flex;
         ${styles};
       `}
     >
-      <div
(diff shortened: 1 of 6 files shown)


src/script/components/calling/DeviceToggleButton.tsx
```diff
@@ -42,60 +42,60 @@ const DeviceToggleButton: React.FC<DeviceToggleButtonProps> = ({currentDevice, d
...
-                height: 10,
-                width: 10,
-              }}
-            />
-          );
-        })}
-      </div>
+        return (
+          <div
+            role="button"
```
PatrykBuniX: probably not related to this PR, but why can't we use native `<button>` element here? :D

V-Gira: I don't see why not, I just assumed it was done that way for a reason previously. I can try it out

PatrykBuniX (approved):
looks good 👍

PR #13604, opened 2022-09-06
test: Conversation details unit tests

przemvs:
Add tests to the conversation details component


src/script/page/RightSidebar/ConversationDetails/utils/getConversationActions.ts
```diff
@@ -47,7 +47,7 @@ const getConversationActions = (
   const isSingleUser = is1to1Action || conversationEntity.isRequest();
   const userEntity = conversationEntity.firstUserEntity();
 
-  const nextConversationEntity = conversationRepository.getNextConversation(conversationEntity);
+  const getNextConversation = () => conversationRepository.getNextConversation(conversationEntity);
```
atomrc, line 50: This makes this trigger lazy (so it's loaded only if we really need the next conversation)

PatrykBuniX (approved):

PR #13614, opened 2022-09-06
refactor: migrate AddParticipants to React

przemvs:
Panel switch functionality will be prepared on end of migration all panel elements, because we need to write additional functionality to be distinguished migrated and not migrated elements.

image
image


src/script/page/RightSidebar/AddParticipants/AddParticipants.tsx
```diff
@@ -0,0 +1,334 @@
...
+    const userEntities = getUserEntities();
+    setContacts(userEntities);
+  }, []);
+
+  return (
+    <div id="add-participants" className="add-participants panel__page panel__page--visible">
+      <PanelHeader
+        showBackArrow
+        goBackUie="go-back-add-participants"
+        onGoBack={onBack}
```
arjita-mitra, line 189: Please talk to Astrid and confirm if you need to add these props for PanelHeader goBackTitle, title, closeBtnTitle

arjita-mitra (changes requested):

src/script/page/RightSidebar/AddParticipants/AddParticipants.tsx
przemvs: Don't worry about it. It's prepared :D 

image
image

src/script/page/RightSidebar/AddParticipants/AddParticipants.tsx
```diff
@@ -0,0 +1,334 @@
...
+    await addUsers();
+    onBack();
+  };
+
+  const onSearchInput = async (value: string) => {
+    await searchServices(value);
+    setSearchInput(value);
+  };
+
+  const getUserEntities = useCallback(() => {
```
PatrykBuniX: Is this function necessary here? We could just use `useMemo`, so there is no need to call this function on initial render (line 187) - we lose 1 rerender.

przemvs: Ahh.. sure, you're right. I'll fix it :)

przemvs: done.

PatrykBuniX (approved):
Looks fine 👍

PR #13616, opened 2022-09-07
feat: use core logout when logging out

atomrc:
### Description

Will use the `logout` method of the core (with the `clearData` param) that will also trigger the core-specific databases to be deleted in case `clearData` is given.

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -815,6 +815,7 @@ class App {
         }
       }
 
+      await this.core.logout(clearData);
       return _redirectToLogin();
     };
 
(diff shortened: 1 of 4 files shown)


thisisamir98 (approved):

PatrykBuniX (approved):

src/script/main/app.ts
```diff
@@ -815,6 +815,7 @@ class App {
         }
       }
 
+      await this.core.logout(clearData);
```
tlebon, line 818: should this be conditional to MLS?

atomrc: The core is the one knowing what to do on logout (deleting database for example). But the webapp needs to call `logout` everytime someone logs out (on a proteus as well)

tlebon (approved):

src/script/main/app.ts
tlebon: ah sorry didnt read the other PR yet, seems fine

PR #13617, opened 2022-09-07
refactor: Avoid rerendering when using fadingScrollbars

atomrc:
The `useEffectRef` is creating a re-render and show be avoided as much as possible. 

Instead we could use callback ref (see https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) and just give a function that will init the DOM specific triggers that we want without re-rendering. 

The same can be done everywhere we use `useEffectRef` and we should migrate those in the near future and totally get rid of the `useEffectRef`

### DIFF
--- a/src/script/page/LeftSidebar/panels/ListWrapper.tsx
+++ b/src/script/page/LeftSidebar/panels/ListWrapper.tsx
@@ -17,15 +17,14 @@
  *
  */
 
-import React, {ReactElement, useEffect} from 'react';
+import React, {ReactElement} from 'react';
 
 import {css} from '@emotion/react';
 import {throttle} from 'underscore';
 import {isScrollable, isScrolledBottom, isScrolledTop} from 'Util/scroll-helpers';
 import Icon from 'Components/Icon';
 import {t} from 'Util/LocalizerUtil';
-import useEffectRef from 'Util/useEffectRef';
-import {useFadingScrollbar} from '../../../ui/fadingScrollbar';
+import {initFadingScrollbar} from '../../../ui/fadingScrollbar';
 
 type LeftListWrapperProps = {
   /** A react element that will be inserted after the header but before the list */
@@ -64,9 +63,6 @@ const ListWrapper: React.FC<LeftListWrapperProps> = ({
   before,
   headerUieName,
 }) => {
-  const [scrollbarRef, setScrollbarRef] = useEffectRef<HTMLDivElement>();
-  useFadingScrollbar(scrollbarRef);
-
   const calculateBorders = throttle((element: HTMLElement) => {
     window.requestAnimationFrame(() => {
       if (element.offsetHeight <= 0 || !isScrollable(element)) {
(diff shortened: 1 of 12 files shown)


src/script/page/LeftSidebar/panels/ListWrapper.tsx
```diff
@@ -79,16 +75,13 @@ const ListWrapper: React.FC<LeftListWrapperProps> = ({
...
+      return;
     }
-    const onScroll = (event: Event) => calculateBorders(event.target as HTMLElement);
-    calculateBorders(scrollbarRef);
-    scrollbarRef.addEventListener('scroll', onScroll);
-
-    return () => scrollbarRef.removeEventListener('scroll', onScroll);
-  }, [scrollbarRef]);
+    calculateBorders(element);
+    element.addEventListener('scroll', () => calculateBorders(element));
```
przemvs, line 83: removeEventListener not needed ? 🤔

src/script/ui/fadingScrollbar.ts
```diff
@@ -17,83 +17,77 @@
...
-  useLayoutEffect(() => {
-    if (!element) {
-      return undefined;
-    }
-    const animationSpeed = 0.05;
+export const initFadingScrollbar = (element: HTMLElement | null): void => {
+  if (!element) {
+    return;
+  }
+  const animationSpeed = 0.05;
```
przemvs: Maybe move this variable above element, or pass as props with default value :)

src/script/ui/fadingScrollbar.ts
```diff
@@ -17,83 +17,77 @@
...
-
-    const events = {
-      mouseenter: fadeIn,
-      mouseleave: fadeOut,
-      mousemove: fadeInIdle,
-      scroll: fadeInIdle,
-    };
+  const fadeStep = (delta: number) => {
+    const initialAlpha = initialColor[3];
+    const currentAlpha = currentColor[3];
```
przemvs: Move magic numbers to enum :)

src/script/ui/fadingScrollbar.ts
```diff
@@ -17,83 +17,77 @@
...
+    const hasAppeared = delta > 0 && currentAlpha >= initialAlpha;
+    const hasDisappeared = delta < 0 && currentAlpha <= 0;
+    if (hasAppeared || hasDisappeared) {
+      return setAnimationState('idle');
+    }
+    currentColor[3] += delta;
```
przemvs: Magic number 🤔

src/script/ui/fadingScrollbar.ts
```diff
@@ -17,83 +17,77 @@
...
+    element.style.setProperty('--scrollbar-color', `rgba(${currentColor})`);
+  };
+  const fadeIn = () => setAnimationState('fadein');
+  const fadeOut = () => setAnimationState('fadeout');
+  const debouncedFadeOut = debounce(fadeOut, 1000);
```
przemvs: Move delay to variable :)

przemvs (changes requested):
Few comments :)

src/script/page/LeftSidebar/panels/ListWrapper.tsx
atomrc: The DOM element will be removed from the DOM completely and all its listeners cleared as well. We do not need this anymore :)

src/script/ui/fadingScrollbar.ts
atomrc: We don't need it above the condition on element, why moving it there?

src/script/ui/fadingScrollbar.ts
atomrc: I don't really see the point of that, we never use it anywhere else 🤔 ?

src/script/ui/fadingScrollbar.ts
atomrc: I just needed a single variable, an enum would be overkill (since we never want to access red, green and blue channel in the code

src/script/ui/fadingScrollbar.ts
atomrc: fixed

przemvs (approved):
LGTM! :)

PR #13623, opened 2022-09-08
runfix: Broader error handling for registration errors

atomrc:
#13621  was not enough to fix all the errors

### DIFF
--- a/src/script/auth/component/AccountForm.tsx
+++ b/src/script/auth/component/AccountForm.tsx
@@ -106,27 +106,26 @@ const AccountForm = ({account, ...props}: Props & ConnectedProps & DispatchProps
       await props.doSendActivationCode(registrationData.email);
       return props.onSubmit(event);
     } catch (error) {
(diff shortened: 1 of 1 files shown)


src/script/auth/component/AccountForm.tsx
```diff
@@ -106,27 +106,26 @@ const AccountForm = ({account, ...props}: Props & ConnectedProps & DispatchProps
       await props.doSendActivationCode(registrationData.email);
       return props.onSubmit(event);
     } catch (error) {
-      if (error && (error instanceof ValidationError || error instanceof BackendError) && error.label) {
-        switch (error.label) {
+      const label = (error as BackendError)?.label;
```
arjita-mitra, line 109: error instanceof ValidationError why this check is removed? @atomrc

atomrc: It created runtime errors, because we are receiving more than just Validation and Backend errors, we would need to find out **all** the different classes that go there, but for now I'd like to go on the safe side and handle anything that has a `label`

V-Gira (approved):

PR #13629, opened 2022-09-09
refactor: Avoid using effectRef when possible

atomrc:
`useEffectRef` can be a little harmful to the application since it triggers a re-render everytime the ref changes. 

Instead of using `useEffectRef` we can use a `disposableRef`. That a ref that is put directly on the element and can provide a dispose callback to be run whenever the dom element is removed from the DOM. 

In general, this PR tries to remove the use of hooks for manipulating DOM elements. `Components` are made for this and should be used when possible. 
Thie PR tries to do things the React way.


src/script/ui/viewportObserver.ts
```diff
@@ -113,23 +111,3 @@ export const viewportObserver = {
   removeElement,
   trackElement,
 };
-
-export const useViewPortObserver = (elementRef?: HTMLElement, defaultIsVisible: boolean = false): boolean => {
```
atomrc, line 117: This hook didn't really make sense. It only manipulate DOM elements. React component are tailored to manipulate DOM elements.

Instead the `InViewport` Component has been used

src/script/hooks/useHideElement.ts
```diff
@@ -1,85 +0,0 @@
...
- *
- */
-
-import {useEffect} from 'react';
-import {KEY} from 'Util/KeyboardUtil';
-import useEffectRef from 'Util/useEffectRef';
-
-const hideControlsClass = 'hide-controls';
-
-const useHideElement = (timeout: number, skipClass?: string) => {
```
atomrc, line 26: This hook was used in the previous CallingUI (where controls would appear when you move the mouse and disappear when the mouse is inactive). 

With the redesign of the calling UI, this mechanism is not needed anymore and can be removed.

@V-Gira can you confirm that there is nothing that relies on the mouse behavior with the new call UI?

src/script/components/utils/InViewport.tsx
```diff
@@ -23,17 +23,15 @@ import {viewportObserver} from '../../ui/viewportObserver';
...
-  style?: React.CSSProperties;
 }
 
-const InViewport: React.FC<InViewportParams> = ({
+const InViewport: React.FC<InViewportParams & React.HTMLProps<HTMLDivElement>> = ({
   children,
   style,
   onVisible,
-  requireFullyInView = true,
+  requireFullyInView = false,
```
atomrc, line 34: Seems like a more sensible default, since only 1 instance of this component needs a `true` value

src/script/components/list/ConversationListCell.tsx
```diff
@@ -119,7 +110,7 @@ const ConversationListCell: React.FC<ConversationListCellProps> = ({
   };
 
   return (
-    <li ref={setViewportElementRef}>
+    <li onContextMenu={openContextMenu}>
```
atomrc, line 113: No idea why this was made this way initially, but new way is the react way of things :D

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -138,7 +132,6 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
     [videoInput],
   );
   const showSwitchCamera = selfSharesCamera && availableCameras.length > 1;
-  const setWrapper = useHideElement(FullscreenVideoCallConfig.HIDE_CONTROLS_TIMEOUT, 'video-controls__button');
```
atomrc, line 141: see https://github.com/wireapp/wire-webapp/pull/13629/files#r967428507

src/script/components/MessagesList/Message/MemberMessage.tsx
```diff
@@ -77,20 +77,17 @@ const MemberMessage: React.FC<MemberMessageProps> = ({
...
-    if (messageHeaderLabelRef) {
-      const link = messageHeaderLabelRef.querySelector('.message-header-show-more');
-      if (link) {
-        const listener = () => onClickParticipants(highlightedUsers);
-        link.addEventListener('click', listener);
-        return () => {
-          link.removeEventListener('click', listener);
-        };
-      }
+  const initShowMore = useDisposableRef(element => {
```
atomrc, line 80: Sadly we still need this `disposableRef` here because the DOM we are manipulating doesn't come from React. It's one of those cases that needs this custom `disposableRef`

atomrc:
Let's wait before next release to merge

PatrykBuniX (approved):

src/script/hooks/useHideElement.ts
atomrc: talked with @V-Gira he confirmed this is not needed anymore

PR #13647, opened 2022-09-11
feat: Always send proteus messages as protobuf

atomrc:
attempts #2 at #13626


atomrc:
Let's wait for next release before merging

atomrc:
Will be implemented on the `core` side in the future

PR #13692, opened 2022-09-19
feat: leave mls conversation (FS-683)

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-683" title="FS-683" target="_blank">TaskFS-683</a>  [Web] Wire user leaves an MLS Group
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

[Use case: leaving a conversation (MLS)](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/617382226/Use+case+leaving+a+conversation+MLS)

Based on linked use-case doc - leaving from conversation works the same way as with `proteus` protocol. We're just sending delete conversation request (`DELETE: /conversations/{cnv_domain}/{cnv}/members/{usr_domain}/{usr}`)

One extra thing we have to do for `mls` conversations: on each user's clients, after we catch `conversation.member-leave` event we need to call `core-crypto`'s `wipeConversation` method.

----

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References


PatrykBuniX (commented):
❓ Because of [the way we're handling conversation events](https://github.com/wireapp/wire-webapp/blob/dev/src/script/conversation/ConversationRepository.ts#L1972), `wipeMLSConversation` method won't get called for case when user was offline (logged out / client closed) during leaving conversation from other client. The expression (linked above - fetching user by id) will result in 403 error - access denied. So we don't have access to the conversation we left (we don't know `groupId` and what protocol it was using). What should we do in this case?

//Update:
We'll use `conversationId` -> `groupId` mappings from indexedDB to specify `groupId` for convo we're trying to wipe.

src/script/conversation/ConversationRepository.ts
```diff
@@ -1927,6 +1940,14 @@ export class ConversationRepository {
     return this.pushToReceivingQueue(eventJson, eventSource);
   };
 
+  private async handleWipeMLSConversation(conversationId: QualifiedId) {
+    try {
+      this.core.service!.conversation.wipeMLSConversation(conversationId);
```
atomrc: So what we could do here is pass on a `groupId` instead of a `conversationId` to the core and avoid mapping convId to GroupId in the core. 

Here you could try to find the conversation in the database (this will give you the groupId). 
If the conversation doesn't exist in the database it's highly likely that it's also not on core and doesn't need wiping. 

WDYT?

atomrc: the gist of it would be 

```js
const conv = this.findConversation(qualifiedId);
if (conv) {
  core.wipe(conv.groupId);
}
```

PatrykBuniX: Good idea, done it ;)

src/script/conversation/ConversationRepository.ts
```diff
@@ -1927,6 +1940,18 @@ export class ConversationRepository {
     return this.pushToReceivingQueue(eventJson, eventSource);
   };
 
+  private async handleWipeMLSConversation(conversationId: string) {
+    try {
+      const conversation = await this.conversationService.loadConversation<ConversationRecord>(conversationId);
```
atomrc: ~I think this should be looked in the local state (I think you can try with `this.conversationState.findConversation`~

actually if you follow this suggestion, you won't even have to fetch to convo yourself

src/script/conversation/ConversationRepository.ts
```diff
@@ -1968,6 +1993,10 @@ export class ConversationRepository {
       }
     }
 
+    if (type === CONVERSATION_EVENT.MEMBER_LEAVE) {
+      this.handleWipeMLSConversation(conversationId.id);
+    }
+
```
atomrc: I think this code belongs in the `reactToConversationsEvent` and there you won't event have to fetch the conversation, the code before it will have done it for you

src/script/conversation/ConversationRepository.ts
PatrykBuniX: Unfortunately it doesn't work in case when my client was logged off, left convo from other client and I login back again. I get `undefined`...

src/script/conversation/ConversationRepository.ts
PatrykBuniX: The thing is this function won't even get called (we get 403 access denied error)

src/script/conversation/ConversationRepository.ts
```diff
@@ -872,17 +872,27 @@ export class ConversationRepository {
...
+      return localStateConversation;
+    }
+
+    if (searchInLocalDB) {
+      const localDBConversation = await this.conversationService.loadConversation<BackendConversation>(
+        conversation_id.id,
+      );
+      if (localDBConversation) {
+        return this.mapConversations([localDBConversation])[0];
+      }
```
PatrykBuniX, line 893: If the conversation was filtered out and not included in the local state, we try to load it directly from indexedDB, so the incoming event regarding it can be executed.

src/script/conversation/ConversationRepository.ts
```diff
@@ -872,17 +872,27 @@ export class ConversationRepository {
    * Check for conversation locally and fetch it from the server otherwise.
    * TODO(Federation): Remove "optional" from "domain"
    */
-  async getConversationById(conversation_id: QualifiedId): Promise<Conversation> {
+  async getConversationById(conversation_id: QualifiedId, searchInLocalDB = false): Promise<Conversation> {
```
PatrykBuniX, line 875: direct search in indexedDB is hidden under optional param, so we don't do it for all the cases

src/script/conversation/ConversationRepository.ts
```diff
@@ -1522,16 +1532,30 @@ export class ConversationRepository {
...
+    delete roles[userQualifiedId.id];
+    conversationEntity.roles(roles);
+
+    const currentTimestamp = this.serverTimeHandler.toServerTimestamp();
+    const userLeaveEvent = await this.core.service!.conversation.leaveConversation(conversationEntity.qualifiedId);
+    const event =
+      userLeaveEvent || EventBuilder.buildMemberLeave(conversationEntity, userQualifiedId, true, currentTimestamp);
+
+    await this.eventRepository.injectEvent(event, EventRepository.SOURCE.BACKEND_RESPONSE);
+    return event;
```
atomrc: It's not clear to me why this change was made. The body is almost exactly similar to `removeMemberFromProteusConversation` why not just using it?

PatrykBuniX: Wanted to split it, because: `leaveConversation` is used for leaving both mls and proteus, so its generic function for leaving. Maybe I will extract duplicated code to separate, generic function, wdyt? :D

atomrc: are you sure? Both, `removeUserFromProteusConversation` and `leaveConversation` do the exact same API call 🤔  I don't see any difference between the 2 (except that `leaveConversation` uses the connected user Id)

PatrykBuniX: Oh, you're right, we have to name it more generic then and use one core method for 3 things: 
1. removing other user from proteus convo
2. leaving proteus convo
3. leaving mls convo
Does it sound good to you? :D

atomrc: That sounds perfect 👌

atomrc (approved):

PR #13705, opened 2022-09-21
runfix: client list errors

PatrykBuniX:
Fix some ts errors in client list component and fix client removal reducer

### DIFF
--- a/src/script/auth/component/ClientItem.tsx
+++ b/src/script/auth/component/ClientItem.tsx
@@ -41,7 +41,7 @@ import {parseError, parseValidationErrors} from '../util/errorUtil';
 
 export interface Props extends React.HTMLProps<HTMLDivElement> {
   client: RegisteredClient;
-  clientError: Error;
+  clientError?: Error;
   onClick: (event: React.MouseEvent<HTMLDivElement>) => void;
   onClientRemoval: (password?: string) => void;
   requirePassword: boolean;
@@ -50,7 +50,7 @@ export interface Props extends React.HTMLProps<HTMLDivElement> {
 
 const ClientItem = ({selected, onClientRemoval, onClick, client, clientError, requirePassword}: Props) => {
   const {formatMessage: _} = useIntl();
-  const passwordInput = React.useRef<HTMLInputElement>();
+  const passwordInput = React.useRef<HTMLInputElement>(null);
 
   const CONFIG = {
     animationSteps: 8,
@@ -147,13 +147,17 @@ const ClientItem = ({selected, onClientRemoval, onClick, client, clientError, re
   const handleSubmit = (event: React.FormEvent): Promise<void> => {
     event.preventDefault();
     let localValidationError = null;
-    if (!passwordInput.current.checkValidity()) {
-      localValidationError = ValidationError.handleValidationState(
-        passwordInput.current.name,
-        passwordInput.current.validity,
-      );
+
+    if (passwordInput.current) {
+      if (!passwordInput.current.checkValidity()) {
+        localValidationError = ValidationError.handleValidationState(
+          passwordInput.current.name,
+          passwordInput.current.validity,
+        );
+      }
+      setIsValidPassword(passwordInput.current.validity.valid);
     }
-    setIsValidPassword(passwordInput.current.validity.valid);
+
     setValidationError(localValidationError);
     return Promise.resolve(localValidationError)
       .then(error => {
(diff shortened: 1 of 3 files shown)


src/script/auth/module/reducer/clientReducer.ts
```diff
@@ -78,7 +78,7 @@ export function clientReducer(state: ClientState = initialClientState, action: A
     case CLIENT_ACTION.CLIENT_REMOVE_SUCCESS: {
       return {
         ...state,
-        clients: [...state.clients.filter(client => client.id === action.payload)],
+        clients: [...state.clients.filter(client => client.id !== action.payload)],
```
PatrykBuniX, line 81: I think what we want is to filter removed client out of the list, not to make the list include only removed client

PatrykBuniX: It's normally not visible for the user as we automatically proceed to login flow.

src/script/auth/component/ClientItem.tsx
```diff
@@ -41,7 +41,7 @@ import {parseError, parseValidationErrors} from '../util/errorUtil';
 
 export interface Props extends React.HTMLProps<HTMLDivElement> {
   client: RegisteredClient;
-  clientError: Error;
+  clientError?: Error;
```
przemvs: Move optional types to end of interface

src/script/auth/component/ClientItem.tsx
```diff
@@ -147,13 +147,17 @@ const ClientItem = ({selected, onClientRemoval, onClick, client, clientError, re
...
-      );
+
+    if (passwordInput.current) {
+      if (!passwordInput.current.checkValidity()) {
+        localValidationError = ValidationError.handleValidationState(
+          passwordInput.current.name,
+          passwordInput.current.validity,
+        );
+      }
+      setIsValidPassword(passwordInput.current.validity.valid);
```
przemvs, line 158: I'll change it to if (!passwordInput.current) { return }

and then if (!passwordInput.current.checkValidity()) {} ... etc :)

src/script/auth/component/ClientList.tsx
```diff
@@ -91,7 +92,7 @@ const ClientList = ({
       {permanentClients.map(client => (
         <ClientItem
           client={client}
-          clientError={isSelectedClient(client.id) && clientError}
+          clientError={isSelectedClient(client.id) ? clientError : undefined}
```
przemvs, line 95: Please do smth like that instead of passing undefined :) 

```suggestion
          {...(isSelectedClient(client.id) && { clientError })}
```

src/script/auth/module/reducer/clientReducer.ts
przemvs: Good catch :D Or it takes removed clients array :D Clarify it with @atomrc

src/script/auth/component/ClientList.tsx
```diff
@@ -54,8 +54,9 @@ const ClientList = ({
 
   const setSelectedClient = (clientId: string) => {
     const isSelectedClient = currentlySelectedClient === clientId;
-    clientId = isSelectedClient ? null : clientId;
-    setCurrentlySelectedClient(clientId);
+
+    const selectedClientId = isSelectedClient ? null : clientId;
+    setCurrentlySelectedClient(selectedClientId);
```
przemvs, line 59: If you don't need change currently selected client to null please do smth like that: 

```
if (!selectedClient) {
  setCurrentlySelectedClient(clientId)
}
```

przemvs (changes requested):
Few comments :)

atomrc (approved):

src/script/auth/component/ClientItem.tsx
PatrykBuniX: We can't return here as we want the function to proceed

src/script/auth/component/ClientList.tsx
PatrykBuniX: its less readable for me but does the same thing... 🤔

src/script/auth/component/ClientList.tsx
PatrykBuniX: we actually want to change it to `null`

src/script/auth/component/ClientItem.tsx
przemvs: Okay!

src/script/auth/component/ClientList.tsx
przemvs: image
This is a difference :)

PatrykBuniX: Screenshot 2022-09-21 at 12 54 43

PatrykBuniX: but the result is the same, we're passing it as props object

przemvs: Yeah sure, I prefer spread it instead of passing undefined :)

przemvs (approved):

PR #13708, opened 2022-09-22
fix: Show expiration message when logout is detected [SQCORE-1272]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1272" title="SQCORE-1272" target="_blank">BugSQCORE-1272</a>  [Web] Desktop: No reason shown on welcome page when logout reason is a password reset
  </td></table>
  <br />
 

Regression Introduced by https://github.com/wireapp/wire-webapp/pull/8738

this `logout` event from the apiClient is triggered when the access token is not valid anymore. So the reason should have been `EXPIRED` and not `NOT_SIGNED_IN` in the original PR

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -190,7 +190,7 @@ class App {
     private readonly core: Core,
     private readonly apiClient: APIClient,
   ) {
-    this.apiClient.on(APIClient.TOPIC.ON_LOGOUT, () => this.logout(SIGN_OUT_REASON.NOT_SIGNED_IN, false));
+    this.apiClient.on(APIClient.TOPIC.ON_LOGOUT, () => this.logout(SIGN_OUT_REASON.SESSION_EXPIRED, false));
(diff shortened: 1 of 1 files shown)


src/script/main/app.ts
```diff
@@ -190,7 +190,7 @@ class App {
     private readonly core: Core,
     private readonly apiClient: APIClient,
   ) {
-    this.apiClient.on(APIClient.TOPIC.ON_LOGOUT, () => this.logout(SIGN_OUT_REASON.NOT_SIGNED_IN, false));
+    this.apiClient.on(APIClient.TOPIC.ON_LOGOUT, () => this.logout(SIGN_OUT_REASON.SESSION_EXPIRED, false));
```
atomrc, line 193: If the apiClient detects a 'logout' is means the token is expired (not that the user has logged out manually)

PatrykBuniX (approved):

PR #13713, opened 2022-09-22
fix: Delete ephemeral messages even when connectivity is down [SEC-198]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SEC-198" title="SEC-198" target="_blank">TaskSEC-198</a>  [webapp] Expired messages are not scrambled when there is no internet connection at time of expiry
  </td></table>
  <br />
 

### The problem

In the initial approach, we would delete the message locally **after** the http call to send the message has actually gone through. This is nice for when the user actually deletes a message (we actually want a visual confirmation that the message has been deleted), but for ephemeral messages, this is a problem since it means you will see the message for an indefinite time if the internet is down

### The solution

We can use the `injection` system that we have build it, it processes any event coming in optimistically. 
If the event is a `delete` event then the code will trigger the removal before the message is actually sent.

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -945,7 +945,7 @@ export class MessageRepository {
       targetedUsers?: QualifiedId[];
       /** will not wait for backend confirmation before actually deleting the message locally */
       optimiticRemoval?: boolean;
-    },
+    } = {},
   ): Promise<void> {
     const conversationId = conversation.id;
     const messageId = message.id;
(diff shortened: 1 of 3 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -941,7 +941,11 @@ export class MessageRepository {
   public async deleteMessageForEveryone(
     conversation: Conversation,
     message: Message,
-    targetedUsers?: QualifiedId[],
+    options: {
+      targetedUsers?: QualifiedId[];
+      /** will not wait for backend confirmation before actually deleting the message locally */
+      optimiticRemoval?: boolean;
```
V-Gira: Is that a typo?
`optimiticRemoval` > `optimisticRemoval`

atomrc: Nice catch 👍

V-Gira (approved):

PR #13716, opened 2022-09-23
feat: add new failure states for decryption failure cases (SQCORE-653)

tlebon:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-653" title="SQCORE-653" target="_blank">TaskSQCORE-653</a>  [Web] Show error message in UI when transfered files have been manipulated (wrong hash or wrong used algorithm)
  </td></table>
  <br />
 

Webapp should show error states when a file cannot be downloaded because of an incorrect hash or decryption error. 
This is especially useful/ important as it could show that some bad actor has tried to manipulate a file. 

Screenshot 2022-09-23 at 15 12 31

### DIFF
--- a/src/script/assets/AssetRepository.ts
+++ b/src/script/assets/AssetRepository.ts
@@ -104,6 +104,7 @@ export class AssetRepository {
       const isServerError = errorMessage.endsWith(HTTP_STATUS.INTERNAL_SERVER_ERROR.toString());
 
       const isExpectedError = isAssetNotFound || isServerError;
+
       if (!isExpectedError) {
         throw error;
       }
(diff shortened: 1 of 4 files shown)


src/script/assets/AssetRepository.ts
```diff
@@ -166,7 +167,12 @@ export class AssetRepository {
       asset.status(AssetTransferState.UPLOADED);
       return downloadBlob(blob, asset.file_name);
     } catch (error) {
-      asset.status(AssetTransferState.UPLOADED);
+      const errorMessage = (error as Error)?.message || '';
```
PatrykBuniX: You can wrap it with type guard - no need for assertion 😄 
```js
if(error instanceof Error){
//...
}
```

PatrykBuniX (approved):
Looks good 👍

V-Gira (approved):

PR #13742, opened 2022-09-27
runfix: ts strict errors and migrate some tests to testing-library

PatrykBuniX:
- fixed some strict typescript errors
- refactored some tests from jasmine to `@testing-library/react`


src/script/components/avatar/UserAvatar.test.tsx
```diff
@@ -91,7 +91,7 @@ describe('UserAvatar', () => {
   });
 
   it('shows avatar badge for connection request', async () => {
-    const participant = new User('id', null);
+    const participant = new User('id', undefined);
```
atomrc: you could probably do `new User('id')` ;) (also valid for all the other ts fixes)

PatrykBuniX: u're right, no need to pass undefined explicitly 👍 addressed

atomrc (approved):
Super nice PR. Thanks for taking the time to migrate those to testing lib ❤️

PR #13746, opened 2022-09-28
runfix: Avoid handling MLS protocol messages [FS-1003]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1003" title="FS-1003" target="_blank">BugFS-1003</a>  [WEB] Added again user after had left MLS convo, can't send MLS messages
  </td></table>
  <br />
 

MLS protocol messages are forwarded to core-crypto and should only be handled by this library. 
At the conversation level we should only handle content messages

This created a problem with people leaving conversation. 
Leaving a conversation generates a lot of MLS protocol messages that, from the webapp point of view, are sent from a participant that is no longer in the conversation. 
This triggered the webapp to update the list of participants and eventually re-add the person that just left in the local conversation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "@emotion/react": "11.9.0",
     "@wireapp/antiscroll-2": "1.3.1",
     "@wireapp/avs": "8.2.8",
-    "@wireapp/core": "31.3.3",
+    "@wireapp/core": "31.3.4",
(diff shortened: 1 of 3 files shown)


PatrykBuniX (approved):

przemvs (approved):
Code LGTM! :)

tlebon:
failing test but looks fine besides that

PR #13748, opened 2022-09-28
refactor: Migrate WarningsViewModel to React

thisisamir98:
telegram-cloud-photo-size-4-5882238633254172878-y


src/script/calling/CallingRepository.ts
```diff
@@ -500,7 +500,7 @@ export class CallingRepository {
         this.leaveCall(activeCall.conversationId, LEAVE_CALL_REASON.USER_TURNED_UNVERIFIED);
         amplify.publish(
           WebAppEvents.WARNING.MODAL,
-          ModalsViewModel.TYPE.ACKNOWLEDGE,
+          WarningModalType.ACKNOWLEDGE,
```
przemvs: I think amplify.publish should be changed for render modal function, and pass it to the DOM - like a Util/renderElement :)

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+    if (typeof action === 'function') {
+      if (content?.currentType === WarningModalType.OPTION) {
+        return action(optionChecked);
+      }
+      if (content?.currentType === WarningModalType.INPUT) {
+        return action(inputValue);
+      }
+      if (content?.currentType === WarningModalType.PASSWORD) {
+        return action(passwordValue);
+      }
```
przemvs: I'll prepare object and check by content?.currentType if exist, eg.

```
const actions = {
  [WarningModalType.OPTION]: action(optionChecked),
  ...moreAcitons
}

content.currentType ? actions[content.currentType] : action()
```

WDYT ?

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+    }
+  };
+
+  return (
+    <div id="modals">
+      <ModalComponent
+        isShown={isModalVisible}
+        onClosed={onModalHidden}
+        aria-describedby="modal-description"
+        onBgClick={() => onBgClick()}
```
przemvs:
```suggestion
        onBgClick={onBgClick}
```

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                data-uie-name="do-close"
+              >
+                <Icon.Close className="modal__header__icon" aria-hidden="true" />
+              </button>
+            </div>
+            <div className="modal__body" ref={initFadingScrollbar}>
+              {(messageHtml || messageText) && (
+                <div className="modal__text" data-uie-name="status-modal-text">
+                  {messageHtml && <div id="modal-description" dangerouslySetInnerHTML={{__html: messageHtml}} />}
+                  {messageText && <div id="modal-description">{messageText}</div>}
```
przemvs: What if you will have messageHTML and messageText? Then you'll have duplicated id's. :)

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                </div>
+              )}
+              {hasPassword && (
+                <form onSubmit={() => doAction(confirm, !!closeOnConfirm)}>
+                  <input
+                    className="modal__input"
+                    type="password"
+                    value={passwordValue}
+                    placeholder={inputPlaceholder ?? ''}
```
przemvs: You can define empty string in inputPlaceholder as default prop

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  />
+                </form>
+              )}
+              {hasInput && (
+                <form onSubmit={() => doAction(confirm, !!closeOnConfirm)}>
+                  <input
+                    maxLength={64}
+                    className="modal__input"
+                    value={inputValue}
+                    placeholder={inputPlaceholder ?? ''}
```
przemvs: You can define empty string in inputPlaceholder as default prop

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  />
+                </form>
+              )}
+              {errorMessage && <div className="modal__input__error">{errorMessage}</div>}
+              {hasOption && (
+                <div className="modal-option">
+                  <div className="checkbox accent-text">
```
przemvs: You can't join this 2 divs into 1 ? :)

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                      data-uie-name="modal-option-checkbox"
+                    />
+                    <label className="label-xs" htmlFor="clear-data-checkbox">
+                      <span className="modal-option-text text-background">{checkboxLabel}</span>
+                    </label>
+                  </div>
+                </div>
+              )}
+              <div className={cx('modal__buttons', {'modal__buttons--column': hasMultipleSecondary})}>
+                {(Array.isArray(secondaryAction) ? secondaryAction : [secondaryAction]).map(action => (
```
przemvs: Please move this checking to variable, it will be more readable :)

src/script/components/Modals/WarningModal/WarningModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+                  <button
+                    key={`${action?.text}-${action?.uieName}`}
```
przemvs: If it doesn't containt action?.text and action?.uieName it will be 'undefined-undefined' :(

src/script/components/Modals/WarningModal/WarningModalState.ts
```diff
@@ -0,0 +1,226 @@
...
+  titleText: '',
+};
+
+const logger = getLogger('WarningModalState');
+
+const useWarningModalState = create<WarningModalState>((set, get) => ({
+  addToQueue: (modalItem: ModalItem) =>
+    set(state => {
+      return {...state, queue: [...state.queue, modalItem]};
+    }),
```
przemvs: You can simplify it :)

src/script/components/Modals/WarningModal/WarningModalState.ts
```diff
@@ -0,0 +1,226 @@
...
+  currentModalContent: defaultContent,
+  currentModalId: null,
+  errorMessage: null,
+  existsInQueue: (modalItem: ModalItem): boolean =>
+    get().queue.findIndex(queueItem => queueItem.id === modalItem.id) !== -1,
+  queue: [],
+  removeFirstItemInQueue: () =>
+    set(state => {
+      return {...state, queue: [...state.queue.slice(1)]};
+    }),
```
przemvs: You can simplify it :)

src/script/components/Modals/WarningModal/WarningModalState.ts
```diff
@@ -0,0 +1,226 @@
...
+  replaceInQueue: (modalItem: ModalItem) =>
+    set(state => {
+      const updatedQueue = [...state.queue];
+      const index = updatedQueue.findIndex(queueItem => queueItem.id === modalItem.id);
+      updatedQueue[index] = modalItem;
```
przemvs: What do you think to use this (?) : 

```suggestion
      const updatedQueue = state.queue.map(queueItem => queueItem.id === modalItem.id ? modalItem : queueItem);
```

src/script/components/Modals/WarningModal/WarningModalState.ts
```diff
@@ -0,0 +1,226 @@
...
+    set(state => ({...state, currentModalContent: nextCurrentModaContent})),
+  updateCurrentModalId: (nextCurrentModalId: string | null) =>
+    set(state => ({...state, currentModalId: nextCurrentModalId})),
+  updateErrorMessage: nextErrorMessage => set(state => ({...state, errorMessage: nextErrorMessage})),
+}));
+
+const onIncmoingModalEvent = (type: WarningModalType, options: ModalOptions, modalId = createRandomUuid()) => {
+  const {currentModalId, existsInQueue, addToQueue, replaceInQueue} = useWarningModalState.getState();
+
+  const alreadyOpen = modalId && modalId === currentModalId;
```
przemvs: It cannot be `const alreadyOpen = modalId === currentModalId;` ?

src/script/components/Modals/WarningModal/WarningModalState.ts
```diff
@@ -0,0 +1,226 @@
...
+    onBgClick: preventClose ? noop : removeCurrentModal,
+    primaryAction: primaryAction ?? null,
+    secondaryAction: secondaryAction ?? null,
+    showClose,
+    titleText: text.title ?? null,
+  };
+
+  switch (type) {
+    case WarningModalType.ACCOUNT_NEW_DEVICES: {
+      content.titleText = t('modalAccountNewDevicesHeadline');
```
przemvs: I'll go with if && return instead of switch..case :)

src/script/components/Modals/WarningModal/WarningModalTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+  preventClose?: boolean;
+  /** Called when action in modal is triggered */
+  primaryAction?: Action;
+  /** Called when secondary action in modal is triggered */
+  secondaryAction?: Action[] | Action;
+  showClose?: boolean;
+  text?: Text;
+}
+
+export enum WarningModalType {
```
przemvs: And i think WarningModal it's bad name, because it's not only WarningModal ( if i good think ) :)

przemvs (changes requested):
Try to remove all ampliy.publish/amplify.subscribe. Instead of this use renderElement maybe. I think we wanna remove all amplify functionality ?

src/script/calling/CallingRepository.ts
atomrc: 👍 +1 
Yes we should try to move out from `amplify` which is making a lot of interactions hard to follow and not typed

src/script/components/Modals/WarningModal/WarningModal.tsx
thisisamir98: css styles break

src/script/components/Modals/WarningModal/WarningModalState.ts
thisisamir98: function has more things to do so can not return and also it has cases handling same logic for different types which in if statement would be a log of || like if (type === WarningModalType.INPUT || type === ....)

src/script/components/Modals/WarningModal/WarningModal.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/13748/commits/f209919f86a342b01b7bed87a75996e3b6e7069d

src/script/components/Modals/WarningModal/WarningModalTypes.ts
thisisamir98: changed to PrimaryModal

src/script/calling/CallingRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/13748/commits/342f8a0afe9ebd679f413dcf01969fa0c5773b22

V-Gira (commented):
Can't really contribute anything but I really like the move to zustand too, feels like I can understand the state management without massive headaches for a change :)

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+            </div>
+          </>
+        )}
+      </ModalComponent>
+    </div>
+  );
+};
+
+const PrimaryModal = {
+  add: addNewModalToQueue,
```
atomrc:
```suggestion
  show: addNewModalToQueue,
```

I think `show` makes a little more sense when we talk about modals. WDYT?

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import React, {useState} from 'react';
```
atomrc, line 20: Let's create a `index.ts` that exports the public API to the modal component. 
So that we can use it by doing `import ... from 'path/to/PrimaryModal` instead of `path/to/PrimaryModal/PrimaryModal`

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+import {Action, PrimaryModalType} from './PrimaryModalTypes';
+import {
+  usePrimaryModalState,
+  showNextModalInQueue,
+  defaultContent,
+  removeCurrentModal,
+  addNewModalToQueue,
+} from './PrimaryModalState';
+
+export interface PrimaryModalProps {}
```
atomrc: probably not needed :D

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+  );
+};
+
+const PrimaryModal = {
+  add: addNewModalToQueue,
+  init: () => {
+    renderElement<PrimaryModalProps>(PrimaryModalComponent)({});
+    showNextModalInQueue();
+  },
+  type: PrimaryModalType,
```
atomrc: we could export this in its own export and in the `index.ts` file

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+    inputPlaceholder,
+    messageHtml,
+    messageText,
+    modalUie,
+    onBgClick,
+    primaryAction,
+    secondaryAction,
+    titleText,
+    closeBtnTitle,
+  } = content || defaultContent;
```
atomrc: The default value should live only inside the state manager (the value it returns should be the default value and never be undefined if we don't expect it)

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                  {messageText && <div id="modal-description-text">{messageText}</div>}
+                </div>
+              )}
+              {hasPassword && (
+                <form onSubmit={() => doAction(confirm, !!closeOnConfirm)}>
+                  <input
+                    className="modal__input"
+                    type="password"
+                    value={passwordValue}
+                    placeholder={inputPlaceholder ?? ''}
```
przemvs: inputPlaceholder doesn't have default value (empty string) ?

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                  />
+                </form>
+              )}
+              {hasInput && (
+                <form onSubmit={() => doAction(confirm, !!closeOnConfirm)}>
+                  <input
+                    maxLength={64}
+                    className="modal__input"
+                    value={inputValue}
+                    placeholder={inputPlaceholder ?? ''}
```
przemvs: inputPlaceholder doesn't have default value (empty string) ?

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                    />
+                    <label className="label-xs" htmlFor="clear-data-checkbox">
+                      <span className="modal-option-text text-background">{checkboxLabel}</span>
+                    </label>
+                  </div>
+                </div>
+              )}
+              <div className={cx('modal__buttons', {'modal__buttons--column': hasMultipleSecondary})}>
+                {secondaryActions
+                  .filter((action): action is Action => action !== null)
```
przemvs: Maybe move this to guards/Modal.ts (isAction) or smth like that :)

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                  .map(action => (
+                    <button
+                      key={`${action?.text}-${action?.uieName}`}
```
przemvs: Key can be 'undefined-undefined' - i think it's wrong.

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                      type="button"
+                      onClick={() => action?.action && doAction(action?.action, true, true)}
```
przemvs:
```suggestion
                      onClick={() => action?.action && doAction(action.action, true, true)}
```

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+                      data-uie-name={action?.uieName}
+                      className={cx('modal__button modal__button--secondary', {
+                        'modal__button--full': hasMultipleSecondary,
+                      })}
+                    >
+                      {action?.text}
```
przemvs: Maybe add custom text if action didn't have any text.

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -0,0 +1,222 @@
...
+const PrimaryModal = {
+  add: addNewModalToQueue,
+  init: () => {
+    renderElement<PrimaryModalProps>(PrimaryModalComponent)({});
+    showNextModalInQueue();
+  },
+  type: PrimaryModalType,
+};
+
+export default PrimaryModal;
```
przemvs: I'm thinking about PrimaryModal name. for me it can be Modal - because it can be reused, PrimaryModal sounds like we have multiple Modal styles, or smth like that.

src/script/components/Modals/PrimaryModal/PrimaryModalState.ts
```diff
@@ -0,0 +1,214 @@
...
+    showClose = false,
+    text = {} as Text,
+  } = options;
+  const content = {
+    checkboxLabel: text.option ?? null,
+    closeBtnTitle: text.closeBtnLabel,
+    closeFn: close,
+    closeOnConfirm,
+    currentType: type,
+    inputPlaceholder: text.input ?? null,
```
przemvs: I think it can be empty string instead of null :)

src/script/components/Modals/PrimaryModal/PrimaryModalState.ts
```diff
@@ -0,0 +1,214 @@
...
+    data,
+    preventClose = false,
+    primaryAction,
+    secondaryAction,
+    hideSecondary,
+    showClose = false,
+    text = {} as Text,
+  } = options;
+  const content = {
+    checkboxLabel: text.option ?? null,
```
przemvs: I'll go with string types instead of null for string values :)

src/script/components/Modals/PrimaryModal/PrimaryModalTypes.ts
```diff
@@ -0,0 +1,83 @@
...
+  currentType: string | null;
+  inputPlaceholder: string | null;
+  messageHtml: string | null;
+  messageText: string | null;
+  modalUie: string;
+  onBgClick: () => void;
+  primaryAction: Action | null;
+  secondaryAction: Action[] | Action | null;
+  titleText: string | null;
+  closeBtnTitle?: string;
```
przemvs: I'll go for string types as optional and without null. We can pass empty string for string :)

przemvs (changes requested):
Good job with replacing amplify!

Few comments more :)

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
thisisamir98: this is a bit too much just for a null check

src/script/components/Modals/PrimaryModal/PrimaryModal.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+class PrimaryModalPage extends TestPage<{}> {
+  constructor() {
+    super(PrimaryModalComponent);
+  }
+
+  getWrapperElement = () => this.get('div#modals');
+  getPrimaryActionButton = () => this.get('[data-uie-name="do-action"]');
+  getSecondaryActionButton = () => this.get('[data-uie-name="do-secondary"]');
+}
```
atomrc: Let's use `@testing-library` directly and all it's super nice primitives (like `getByTestId`). 
You can inspire from what @PatrykBuniX has been doing lately migrating a few tests to testling-lib (cf https://github.com/wireapp/wire-webapp/pull/13751/files#diff-aa9cafe9a5b4c4728e53a5935e604adaf6c345a38a55778469861301792662e0R46)

przemvs (approved):
Code LGTM :)

atomrc (approved):

PatrykBuniX (approved):
💪

PR #13756, opened 2022-09-29
feat: add unmute self when in call by pressing spacebar

aweiss-dev:
- This feature adds an event listener for SPACEBAR. 
- When being in an active call and muted (while being allowed to unmute), pressing and holding the SPACEBAR will unmute the user.
- Releasing the SPACEBAR will mute the user again.

This is a working draft and needs probably more information for our users like:
- Option in the Call Settings to deactivate the feature?
- Explanation for the user or a notification that the feature exists?
- Graphical response for the user while pressing the spacebar to make it more obvious they are unmuted?


aweiss-dev:
outdated

PR #13781, opened 2022-10-03
runfix: refactor tests and fix ts strict errors

PatrykBuniX:
Migrate some tests to testing-library and fix some ts errors along the way.


src/script/components/MessagesList/Message/CallMessage.tsx
```diff
@@ -41,11 +41,11 @@ const CallMessage: React.FC<CallMessageProps> = ({message}) => {
     <div className="message-header">
       <div className="message-header-icon message-header-icon--svg">
         {isCompleted ? (
-          <div className="svg-green">
+          <div data-uie-name="pickup-icon" className="svg-green">
```
atomrc: Let's not use `uie-names` only for our unit test. They have a meaning for QA, so we should try to have only qa related ones in the code. 

Instead, I'd suggest you mock the `Icon` and render a specific text (instead of the actual icon) and use `getByText`. You can inspire from this (https://github.com/wireapp/wire-webapp/blob/dev/src/script/components/MessagesList/Message/CallMessage.test.tsx#L25) (but avoiding using the `class` here, rather some text). 


```js
jest.mock('Components/Icon', () => ({
  Hangup: function HangupIcon() {
    return <span>hangupIcon</span>;
  },
  Pickup: function PickupIcon() {
    return <span>pickupIcon</span>;
  },
}));
```

And later:

```js
getByText('hangupIcon')
```

atomrc (changes requested):
Just a single suggestion. The rest looks really nice 😎

src/script/components/MessagesList/Message/CallMessage.tsx
PatrykBuniX: Sure!

atomrc (approved):

PR #13784, opened 2022-10-04
runfix: Do not use usePrimaryModalState hook in a function

thisisamir98:
we should not use hooks in a function outside of a react component!


src/script/view_model/ActionsViewModel.ts
```diff
@@ -178,8 +178,7 @@ export class ActionsViewModel {
                   resolve();
                 } catch (error) {
                   if (error instanceof BackendError) {
-                    // eslint-disable-next-line react-hooks/rules-of-hooks
-                    const updateErrorMessage = usePrimaryModalState(state => state.updateErrorMessage);
+                    const {updateErrorMessage} = usePrimaryModalState.getState();
```
atomrc, line 181: Does this work out of the box?

atomrc (approved):

src/script/view_model/ActionsViewModel.ts
thisisamir98: yes it does.

PR #13790, opened 2022-10-04
refactor: Move WarningsContainer to Zustand

thisisamir98:
WarningsContainer is now using Zustand instead of amplify ✅ 
WarningsContainer is being tested now ✅

### DIFF
--- a/src/script/view_model/WarningsContainer/WarningsContainer.tsx
+++ b/src/script/view_model/WarningsContainer/WarningsContainer.tsx
@@ -21,29 +21,23 @@ import {WebAppEvents} from '@wireapp/webapp-events';
 import {amplify} from 'amplify';
 import cx from 'classnames';
 
-import {getLogger} from 'Util/Logger';
 import {t} from 'Util/LocalizerUtil';
-import {safeWindowOpen} from 'Util/SanitizationUtil';
 import {afterRender} from 'Util/util';
 
-import {Config} from '../Config';
-import {PermissionState} from '../notification/PermissionState';
+import {Config} from '../../Config';
 import {Runtime} from '@wireapp/commons';
 
-import React, {useEffect, useState} from 'react';
+import React, {useEffect} from 'react';
 import Icon from 'Components/Icon';
 import {registerReactComponent} from 'Util/ComponentUtil';
-import PrimaryModal from 'Components/Modals/PrimaryModal';
+import {CONFIG, TYPE} from './WarningsTypes';
+import {closeWarning, useWarningsState} from './WarningsState';
 
 const WarningsContainer: React.FC = () => {
-  const logger = getLogger('WarningsViewModel');
-
+  const name = useWarningsState(state => state.name);
+  const warnings = useWarningsState(state => state.warnings);
   const type = TYPE;
-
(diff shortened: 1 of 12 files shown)


yarn.lock
```diff
@@ -6406,11 +6406,6 @@ generate-changelog@1.8.0:
     commander "^2.9.0"
     github-url-from-git "^1.4.0"
 
-gensequence@^3.1.1:
-  version "3.1.1"
-  resolved "https://registry.yarnpkg.com/gensequence/-/gensequence-3.1.1.tgz#95c1afc7c0680f92942c17f2d6f83f3d26ea97af"
-  integrity sha512-ys3h0hiteRwmY6BsvSttPmkhC0vEQHPJduANBRtH/dlDPZ0UBIb/dXy80IcckXyuQ6LKg+PloRqvGER9IS7F7g==
-
```
atomrc, line 6413: humm this is weird, I don't see any changes in the package.json 🤔

src/script/view_model/WarningsContainer/WarningsContainer.tsx
```diff
@@ -154,7 +75,7 @@ const WarningsContainer: React.FC = () => {
   return (
     <div className={cx('warning', {'warning-dimmed': warningDimmed})}>
       {visibleWarning === type.REQUEST_CAMERA && (
-        <div className="warning-bar warning-bar-feature">
+        <div data-uie-name="request-camera" className="warning-bar warning-bar-feature">
```
atomrc, line 78: how about using `visibleWarning` everywhere in the `data-uie-name` so that we don't have to come up with further names when new warnings appear?

atomrc: we could even have a generic `<div>` wrapper that would set the uieName and the class according to the type of warning:

```jsx
const classnames: Record<type, string> = {
 [type.REQUEST_CAMERA]: 'warning-bar warning-bar-feature',
 [type.NO_INTERNET]: '...'
 //...
}

<div data-uie-name={visibleWarning} className={classnames[visibleWarning]}>
  { visibleWarning === type.REQUEST_CAMERA }
  { visibleWarning === type.NO_INTERNET }
  ...
</div>
```

atomrc: And now that you are backed by tests, you will know that you are not breaking the current feature ;)

atomrc: Ah shit, you might also need a mapping from `warningType => uieName` since we already have some uie-names in the wild that we need to restore :/ 

Nevermind ignore my previous comment, let's keep it the way you made it 👍

yarn.lock
thisisamir98: i'm not sure, I just did yarn install 🤔

atomrc (approved):
Looks neat man 😎 💪

yarn.lock
atomrc: 🤷  maybe try to re-run `yarn` locally. 
If there are no other changes let's keep it this way then I guess

PR #13797, opened 2022-10-05
runfix: refactor tests and fix ts strict errors

PatrykBuniX:
Next portion of tests refactoring to `@testing-library/react` and fixing some ts errors.


src/script/components/Modals/UserModal/UserModal.test.tsx
```diff
@@ -19,43 +19,23 @@
...
-    super(UserModal, props);
-  }
-
-  getWrapperElement = () => this.get('div.user-modal');
-  getStatusModalText = () => this.findByTestId('status-modal-text');
-}
+import {render, waitFor} from '@testing-library/react';
 
 describe('UserModal', () => {
-  it('does not render when user id is not passed', async () => {
```
PatrykBuniX, line 41: This seems not to be a case anymore since we're always passing qualified id object as `userId` prop. Typescript does not allow to pass `null` there.

atomrc (approved):
sweeeet 👍

PR #13803, opened 2022-10-05
runfix: Correctly stack modals and fix memory issue in renderElement

thisisamir98:
The order of how we render modals matters! some times for example in the flow of searching a user and trying to unblock them we show stack modals (user-modal & confirm-modal from primary modals) 
By calling `renderElement` we just append the modal to the wire-main in an unordered way so to fix it we need separate roots for these modals so they would be rendered in their own containers with an order. 

Also, the `renderElement` utility turns out to be using the same variables `elementContainer` & `reactRoot` for every call which is a memory bug actually, fixed by changing it to a Map!

### DIFF
--- a/src/page/template/wire-main.htm
+++ b/src/page/template/wire-main.htm
@@ -18,9 +18,6 @@
           #include('modal/legal-hold.htm')
         </div>
     </div>
-
-    #include('modals.htm')
-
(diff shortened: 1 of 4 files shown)


atomrc (approved):
💪  Nice investigation man

PR #13815, opened 2022-10-07
fix: issue with ui during 2fa too many clients flow

tlebon:
Fixes three issues with the 2 factor too many clients flow:
1. wrong password shows a broken open frame after rerender
Screenshot 2022-10-07 at 18 02 29

2. wrong password also is clearing the 2fa password since we were removing it in the finally, should be removed before the catch. 

3. clicking on another client after a wrong password error also shows the error on the other client. 
Screenshot 2022-10-07 at 18 02 35

### DIFF
--- a/src/script/auth/component/ClientItem.tsx
+++ b/src/script/auth/component/ClientItem.tsx
@@ -62,18 +62,24 @@ const ClientItem = ({selected, onClientRemoval, onClick, client, clientError, re
   const [password, setPassword] = useState('');
   const [isValidPassword, setIsValidPassword] = useState(true);
   const [validationError, setValidationError] = useState<ValidationError | null>(null);
+  const [isOpen, setIsOpen] = useState(requirePassword && (isSelected || isAnimating));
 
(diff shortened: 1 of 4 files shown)


src/script/auth/component/ClientItem.tsx
```diff
@@ -62,18 +62,24 @@ const ClientItem = ({selected, onClientRemoval, onClick, client, clientError, re
   const [password, setPassword] = useState('');
   const [isValidPassword, setIsValidPassword] = useState(true);
   const [validationError, setValidationError] = useState<ValidationError | null>(null);
+  const [isOpen, setIsOpen] = useState(requirePassword && (isSelected || isAnimating));
```
V-Gira, line 65: Looks so obvious when I see your solution. Couldn't figure one out when I looked at it, props!

V-Gira (approved):

atomrc (approved):

PatrykBuniX (approved):

PR #13838, opened 2022-10-11
refactor: Streamline message sending process

atomrc:
With the latest core, many message have been moved to the combination of `MessageBuilder.build...Message` and `send`. 
We need to adapt to this. 
This is also an opportunity for relying on the core to handle blocking message sending.


src/script/conversation/MessageRepository.ts
```diff
@@ -692,90 +667,94 @@ export class MessageRepository {
...
+          conversationId: conversation.qualifiedId,
+          nativePush,
+          onClientMismatch: mismatch => this.onClientMismatch?.(mismatch, conversation, silentDegradationWarning),
+          payload,
+          protocol: ConversationProtocol.PROTEUS,
+          targetMode,
+          userIds: this.generateRecipients(conversation, recipients, skipSelf),
+        };
 
-    return this.messageSender.queueMessage(() =>
```
atomrc, line 768: The message queuing is not completely hidden inside of the Core :)

src/script/conversation/MessageRepository.test.ts
```diff
@@ -104,7 +97,9 @@ async function buildMessageRepository(): Promise<[MessageRepository, MessageRepo
   };
   /* eslint-disable sort-keys-fix/sort-keys-fix */
   const deps = Object.values(dependencies) as ConstructorParameters<typeof MessageRepository>;
-  return [new MessageRepository(...deps), dependencies];
+  const messageRepository = new MessageRepository(...deps);
+  jest.spyOn(messageRepository as any, 'updateMessageAsSent').mockReturnValue(undefined);
```
przemvs, line 101: any? :D

src/script/conversation/MessageRepository.ts
```diff
@@ -692,90 +667,94 @@ export class MessageRepository {
...
     } = {
       playPingAudio: false,
       syncTimestamp: true,
     },
   ) {
     const {groupId} = conversation;
 
-    const injectOptimisticEvent: MessageSendingCallbacks['onStart'] = async genericMessage => {
+    const messageTimer = conversation.messageTimer();
+    const payload = enableEphemeral && messageTimer ? MessageBuilder.wrapInEphemeral(message, messageTimer) : message;
```
przemvs, line 701: What if messageTimer will be 0? Then we not wrapping it ?

src/script/conversation/MessageRepository.ts
```diff
@@ -692,90 +667,94 @@ export class MessageRepository {
...
     };
 
-    const updateOptimisticEvent: MessageSendingCallbacks['onSuccess'] = (genericMessage, sentTime) => {
-      this.updateMessageAsSent(conversation, genericMessage.messageId, syncTimestamp ? sentTime : undefined);
-    };
-
-    const handleSuccess = async (genericMessage: GenericMessage, sentTime?: string) => {
-      const preMessageTimestamp = new Date(new Date(sentTime).getTime() - 10).toISOString();
+    const handleSuccess = async (sentAt: string) => {
+      const preMessageTimestamp = new Date(new Date(sentAt).getTime() - 10).toISOString();
```
przemvs: I'll move new Date(sentAt).getTime() to other variable, and use it there, to more readaiblity. Also what 10 means?

src/script/conversation/MessageRepository.ts
```diff
@@ -692,90 +667,94 @@ export class MessageRepository {
...
-        ...commonProperties,
-        conversationDomain: conversation.domain || undefined,
-        nativePush,
-        onClientMismatch: mismatch => this.onClientMismatch?.(mismatch, conversation, silentDegradationWarning),
-        protocol: ConversationProtocol.PROTEUS,
-        targetMode,
-        userIds: this.generateRecipients(conversation, recipients, skipSelf),
-      }),
-    );
+    if ((await injectOptimisticEvent()) === false) {
```
przemvs: WDYT ?

```suggestion
    const shouldInjectOptimisticEvent = await injectOptimisticEvent();
    
    if (shouldInjectOptimisticEvent) {
```

przemvs (commented):
Few comments :)

src/script/conversation/MessageRepository.ts
atomrc: 0 would mean that there are no timer involved, so we don't want to wrap it in that case

src/script/conversation/MessageRepository.ts
atomrc: The `=== false` is important here, but make sense to extract to a variable 👍

src/script/conversation/MessageRepository.test.ts
atomrc: yeah because I'm mocking private method for simplicity

src/script/conversation/MessageRepository.ts
atomrc: done

src/script/conversation/MessageRepository.ts
atomrc: done

przemvs (approved):
LGTM :)

PR #13842, opened 2022-10-11
feat: Catch and log failures to sft server calls

thisisamir98:
As requested by the avs team to try to have a better understanding of call-dropping reasons...

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -86,6 +86,8 @@ import {PayloadBundleState} from '@wireapp/core/src/main/conversation';
 import {Core} from '../service/CoreSingleton';
 import {LEAVE_CALL_REASON} from './enum/LeaveCallReason';
 
+const avsLogger = getLogger('avs');
+
 interface MediaStreamQuery {
   audio?: boolean;
   camera?: boolean;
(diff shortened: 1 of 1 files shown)


src/script/calling/CallingRepository.ts
```diff
@@ -1159,11 +1160,15 @@ export class CallingRepository {
...
+        const response = await axios.post(url, data);
 
-      const {status, data: axiosData} = response;
-      const jsonData = JSON.stringify(axiosData);
-      this.wCall?.sftResp(this.wUser!, status, jsonData, jsonData.length, context);
+        const {status, data: axiosData} = response;
+        const jsonData = JSON.stringify(axiosData);
+        this.wCall?.sftResp(this.wUser!, status, jsonData, jsonData.length, context);
+      } catch (error) {
+        avsLogger.warn(`Request to sft server failed with error`, error);
```
atomrc, line 1173:
```suggestion
        const message = error instanceof Error ? error.message : error;
        avsLogger.error(`Request to sft server failed with error: ${message}`, error);
```
WDYT?

atomrc: Cause the object is not going to print in the exported text logs

atomrc (approved):
😎

PR #13843, opened 2022-10-11
chore: refine wire configuration - type

aweiss-dev:
make the wire config type more accessible and easier to read/understand.

### DIFF
--- a/src/types/Wire.types.ts
+++ b/src/types/Wire.types.ts
@@ -1,0 +1,97 @@
+/*
+ * Wire
+ * Copyright (C) 2022 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


src/types/Wire.types.ts
```diff
@@ -0,0 +1,97 @@
...
+  FEATURE: FeatureVariables;
+  MAX_GROUP_PARTICIPANTS: number;
+  MAX_VIDEO_PARTICIPANTS: number;
+  NEW_PASSWORD_MINIMUM_LENGTH: number;
+  URL: URLVariables;
+  VERSION: string;
+  WEBSITE_LABEL: string;
+};
+
+export interface WireConfiguration {
```
atomrc: The naming feels a little weird, this is not really a `Configuration`. The `env` is but not the `app` part 🤔

atomrc (approved):

PR #13846, opened 2022-10-12
runfix: Use primary modals directly in jsx

thisisamir98:
we should not use render element now that main app is in react

### DIFF
--- a/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
+++ b/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
@@ -97,8 +97,6 @@ export const PrimaryModalComponent: React.FC = () => {
 
   const secondaryActions = Array.isArray(secondaryAction) ? secondaryAction : [secondaryAction];
 
-  console.info('bardia', {content, currentId});
-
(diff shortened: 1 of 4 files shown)


src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -97,6 +97,8 @@ export const PrimaryModalComponent: React.FC = () => {
 
   const secondaryActions = Array.isArray(secondaryAction) ? secondaryAction : [secondaryAction];
 
+  console.info('bardia', {content, currentId});
```
przemvs: This is not needed i think :)

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -97,6 +97,8 @@ export const PrimaryModalComponent: React.FC = () => {
...
+
```
przemvs:
```suggestion
```

przemvs (approved):
LGTM, 1 console log to remove ;D

PR #13852, opened 2022-10-13
chore: work on linter warnings

aweiss-dev:
chore: work on linter warnings

- removed default exports -> replaced with named exports
- turn eslint rules 'no-magic-numbers', 'id-length' and '@typescript-eslint/explicit-module-boundary-types' off.
- exclude javascript files from eslint parsing


.eslintrc.json
```diff
@@ -44,8 +57,8 @@
     }
   ],
   "rules": {
-    "id-length": "warn",
-    "no-magic-numbers": "warn",
+    "id-length": "off",
+    "no-magic-numbers": ["off", {"ignoreArrayIndexes": true}],
```
PatrykBuniX, line 61: Do we need this config here when the rule is `off`? (`{"ignoreArrayIndexes": true}`)

V-Gira: :upside_down_face: 

the rule did make sense before removing the rule though

aweiss-dev: its there in case we want to turn the rule on again. this setting doesn't hurt and we don't forget it then :D

PR #13833, opened 2022-10-10
refactor: migrate tests to testing-library

PatrykBuniX:
- migrate last portion of tests to testing library,
- fix some of the ts strict errors along the way
- delete TestPage util 😄

PR #6308, opened 2019-04-08
refactor: Integrate AVS API

atomrc:
TODOs:
- [x] restore screensharing for the wrapper (with electron's desktopCapturer)
- [x] restore routing the audio from a call to the selected output device
- [x] handle upgrading from audio to video
- [x] Handle media stream change on the UI side (when the user chooses another input device while in a call)
- [x] Show video preview when an incoming video call is ringing 
- [x] Show video preview when outgoing video **group** call is started 
- [x] Check that group > 4 ppl cannot do video call
- [x] Listen to mute state changes in avs
- [x] Save call start/call end events in the database
- [x] Find a way to get the media streams from the avs lib (in order to display the video element on screen)
- [x] adapt the video grid handler with the new data format
- [x] Get the connected participants from avs lib
- [x] ~~Ability to remove a participant while in a call (should also be removed from avs)~~
- [x] Close ongoing call when the app is reloaded
- [x] Handle group calls

TODO (before merge):
- [x] restore the unit test in `package.json`;
- [x] remove deployment script to `-qa` environment.
- [x] Restore the build script to mode prod

PR #13855, opened 2022-10-13
chore: cleanup deprecated dependencies

atomrc:
Enzyme is not needed since https://github.com/wireapp/wire-webapp/pull/13833
sdp-transform is not needed since https://github.com/wireapp/wire-webapp/pull/6308


PatrykBuniX (approved):
🚀 💪

PR #13881, opened 2022-10-17
feat: scaled down view for call-ui [ACC-65]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-65" title="ACC-65" target="_blank">SubtaskACC-65</a>  Video Call View
  </td></table>
  <br />
 

Peek 2022-10-17 16-06

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -343,7 +343,7 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
               css={{
                 display: 'flex',
                 justifyContent: horizontalXsBreakpoint ? 'space-around' : 'space-between',
-                width: horizontalXsBreakpoint ? '100%' : '',
+                ...(horizontalXsBreakpoint && {width: '100%'} ),
               }}
             >
               <li className="video-controls__item">
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -290,13 +290,15 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
               <Icon.ArrowNext css={{position: 'relative', right: 4, transform: 'rotate(180deg)'}} />
             </button>
           )}
-          <div css={{bottom: 108, display: 'flex', justifyContent: 'center', position: 'absolute', width: '100%'}}>
-            <Pagination
-              totalPages={totalPages}
-              currentPage={currentPage}
-              onChangePage={newPage => changePage(newPage, call)}
-            />
-          </div>
+          {!verticalBreakpoint && (
+            <div css={{bottom: 108, display: 'flex', justifyContent: 'center', position: 'absolute', width: '100%'}}>
+              <Pagination
+                totalPages={totalPages}
+                currentPage={currentPage}
+                onChangePage={newPage => changePage(newPage, call)}
+              />
+            </div>
+          )}
         </>
       )}
       {!isChoosingScreen && (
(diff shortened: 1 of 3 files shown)


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -182,7 +188,19 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
 
   return (
     <div id="video-calling" className="video-calling">
-      <div id="video-title" className="video-title">
+      <div
+        id="video-title"
+        className="video-title"
+        css={{height: verticalBreakpoint ? '30px' : '', lineHeight: verticalBreakpoint ? '12px' : ''}}
```
tlebon: maybe conditional object properties instead of ternaries

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -266,43 +290,60 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
...
+                  </span>
+                </button>
+              </li>
+            )}
+            <div
+              className="video-controls__centered-items"
+              css={{
+                display: 'flex',
+                justifyContent: horizontalXsBreakpoint ? 'space-around' : 'space-between',
+                width: horizontalXsBreakpoint ? '100%' : '',
```
tlebon: same, example: 
```suggestion
                ...(horizontalXsBreakpoint && {width: '100%'} ),
```

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -422,19 +463,23 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
...
-                    setMaximizedParticipant(call, null);
-                  }}
-                  currentItem={activeCallViewTab}
-                  textSubstitute={participants.length.toString()}
-                />
-              )}
-            </div>
+            {!horizontalXsBreakpoint && (
+              <div
+                css={{display: 'flex', justifyContent: 'flex-end', minWidth: !horizontalSmBreakpoint ? '157px' : ''}}
```
tlebon: same

src/script/components/calling/FullscreenVideoCall.tsx
V-Gira: Oh yeah, that's much better, thx!

src/script/components/calling/FullscreenVideoCall.tsx
atomrc: Why not using css-only media definitions here? Doesn't really need to live in the js realm from what I can see

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -266,43 +290,60 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
...
-                  {t('videoCallOverlayConversations')}
-                </span>
-              </button>
-            </li>
-            <div className="video-controls__centered-items" css={{display: 'flex'}}>
+          <ul
+            className="video-controls__wrapper"
+            css={{
+              justifyContent: horizontalSmBreakpoint ? 'center' : '',
+              padding: verticalBreakpoint ? '10px' : horizontalXsBreakpoint ? '32px 10px' : '32px 40px 0 40px',
```
przemvs, line 312: Please don't use nested ternary operators :)

src/script/components/calling/FullscreenVideoCall.tsx
V-Gira: true, that's probably the better idea since I haven't refactored anything from the less file

atomrc (approved):

PR #13889, opened 2022-10-18
runfix: prevent sidebar header to overlap with button

V-Gira:
----

### Issue

- sidebar header is overlapping with preferences button when legal hold is active

Screenshot from 2022-10-18 11-29-12

### Solution

- remove the use of absolute positioning for buttons in the sidebar header and use css grid instead

image

### DIFF
--- a/src/style/list/list.less
+++ b/src/style/list/list.less
@@ -49,7 +49,6 @@
   display: grid;
   height: @content-title-bar-height;
   flex-shrink: 0;
-  justify-content: space-between;
   border-bottom: 1px solid var(--sidebar-border-color);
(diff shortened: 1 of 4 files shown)


src/style/list/list.less
```diff
@@ -45,13 +45,13 @@
...
 
 .left-list-header {
-  .flex-center;
   position: relative;
+  display: grid;
   height: @content-title-bar-height;
   flex-shrink: 0;
-  justify-content: center;
-  padding: 0 12px;
+  justify-content: space-between;
```
V-Gira:
```suggestion
```

przemvs (approved):
Code LGTM! :)

PR #13893, opened 2022-10-18
chore: Cleanup unused dependencies

atomrc:
Since https://github.com/wireapp/wire-webapp/pull/5202 the deployment of the webapp is handled by github CI job


PatrykBuniX (approved):
😎

PR #13898, opened 2022-10-19
feat: activate mls for client via mls feature config [FS-1088]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1088" title="FS-1088" target="_blank">TaskFS-1088</a>  [Web] Activate MLS for client via MLS feature config
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We want to add backward compatibility for the clients that were broken after mistakenly enabling FEATURE_ENABLE_MLS flag on their environment. We want to disable MLS for these clients.

Electron's secretsCrypto global config has changed from `secretsCrypto` to `systemCrypto`. Before passing MLS config config object into the core, we check if a current client is using desktop wrapper and whether the new global is defined and enable MLS only in this case.

----

----

----
##### References


atomrc (changes requested):
Code-wise, it's all good. But let's not merge for now as we need a plan to be able to work with the current deployed version of the wrapper

package.json
```diff
@@ -3,7 +3,7 @@
     "@emotion/react": "11.10.4",
     "@wireapp/antiscroll-2": "1.3.1",
     "@wireapp/avs": "8.2.16",
-    "@wireapp/core": "32.1.1",
+    "@wireapp/core": "^33.0.0",
```
atomrc: Should be an exact match, no `^`
```suggestion
    "@wireapp/core": "33.0.0",
```

PatrykBuniX: oops 👍

src/script/main/app.ts
```diff
@@ -31,7 +31,7 @@ import {Runtime} from '@wireapp/commons';
 
 import {getLogger, Logger} from 'Util/Logger';
 import {t} from 'Util/LocalizerUtil';
-import {arrayToBase64, checkIndexedDb, createRandomUuid} from 'Util/util';
+import {arrayToBase64, checkIndexedDb, createRandomUuid, isMLSDisabledOnElectron} from 'Util/util';
```
atomrc: as much as possible, we try not to mention electron in the webapp. It doesn't need to be aware of it. 
Let's call it `supportsMLS`

PatrykBuniX: sure 👍

src/script/util/util.ts
```diff
@@ -398,3 +398,6 @@ export const getSelectionPosition = (element: HTMLTextAreaElement, currentMentio
 
   return {newEnd, newStart};
 };
+
+// temporary hack that disables mls for old 'broken' desktop clients, see https://github.com/wireapp/wire-desktop/pull/6094
+export const isMLSDisabledOnElectron = Runtime.isDesktopApp() && !window.systemCrypto;
```
atomrc: I think this condition will disable mls for webapp not in electron

```suggestion
export const supportsMLS = !Runtime.isDesktopApp() || window.systemCrypto;
```

atomrc (approved):

atomrc (approved):

PR #13905, opened 2022-10-19
refactor: Move cryptobox hooks into @wireapp/core

atomrc:
Will let the core handle session creation and prekey generation hooks.
needs https://github.com/wireapp/wire-web-packages/pull/4535

### DIFF
--- a/src/script/cryptography/CryptographyRepository.ts
+++ b/src/script/cryptography/CryptographyRepository.ts
@@ -17,11 +17,8 @@
  *
  */
 
-import ko from 'knockout';
-import {amplify} from 'amplify';
 import type {QualifiedId} from '@wireapp/api-client/src/user/';
-import {Cryptobox, CryptoboxSession} from '@wireapp/cryptobox';
-import {WebAppEvents} from '@wireapp/webapp-events';
+import {CryptoboxSession} from '@wireapp/cryptobox';
 import type {PreKey as BackendPreKey} from '@wireapp/api-client/src/auth/';
 import {StatusCodes as HTTP_STATUS} from 'http-status-codes';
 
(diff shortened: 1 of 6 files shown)


aweiss-dev (approved):

src/script/cryptography/CryptographyRepository.ts
```diff
@@ -47,21 +43,9 @@ export interface ClientKeys {
...
   logger: Logger;
 
-  static get CONFIG() {
-    return {
-      UNKNOWN_DECRYPTION_ERROR_CODE: 999,
-    };
-  }
-
-  static get REMOTE_ENCRYPTION_FAILURE(): string {
-    return '💣';
```
PatrykBuniX, line 62: wow 💣

src/script/cryptography/CryptographyRepository.ts
```diff
@@ -175,10 +142,11 @@ export class CryptographyRepository {
         );
         const sessionId = this.core.service!.cryptography.constructSessionId({domain, id: userId}, clientId);
         const preKeyArray = base64ToArray(preKey.key);
-        return await this.cryptobox!.session_from_prekey(sessionId, preKeyArray.buffer);
+        return await this.cryptobox.session_from_prekey(sessionId, preKeyArray.buffer);
```
przemvs, line 145: await there is not needed :)

atomrc: true that :)

src/script/main/app.ts
```diff
@@ -152,8 +152,8 @@ class App {
     storage: StorageService;
   };
   repository: ViewModelRepositories = {} as ViewModelRepositories;
-  debug: DebugUtil;
-  util: {debug: DebugUtil};
+  debug?: DebugUtil;
+  util?: {debug: DebugUtil};
```
przemvs, line 156: Shouldn't be moved after required types?

przemvs (approved):
Few things, rest LGTM :)

src/script/cryptography/CryptographyRepository.ts
atomrc: actually I do need it 
image

src/script/main/app.ts
atomrc: It's an object, so the order is not important

PatrykBuniX (approved):

PR #13914, opened 2022-10-20
runfix: tweak auth page visuals

V-Gira:
### Issues

- QA tests are failing because of the position of "Forgot Password"
- Spinner isn't rendering correctly
- Some missing margins in mobile view
- The "Phone Login Situation" (see below)

### Solutions

- Revert position of "forgot password to the previous one"
- Add missing margins

#### The Phone Login Situation

The phone login input uses the `InputSubmitCombo` component from the ui-kit and is pretty broken.
That `InputSubmitCombo` has been removed from the rest of the auth page, but there's no replacement design here.

The "fix" is dirty and uses a nested ternary operator, but any more time spend on it feels like it would be wasted (unpopular feature + design needs to change)

before:
Peek 2022-10-20 15-11
after:
Peek 2022-10-20 15-04

### DIFF
--- a/src/script/auth/component/PhoneLoginForm.tsx
+++ b/src/script/auth/component/PhoneLoginForm.tsx
@@ -20,7 +20,17 @@
 import React, {useRef, useState} from 'react';
 
 import {LoginData} from '@wireapp/api-client/src/auth';
-import {ArrowIcon, Input, InputBlock, InputSubmitCombo, Loading, RoundIconButton, Select} from '@wireapp/react-ui-kit';
+import {
+  ArrowIcon,
+  Input,
+  InputBlock,
+  InputSubmitCombo,
+  Loading,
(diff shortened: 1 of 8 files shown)


src/script/auth/component/PhoneLoginForm.tsx
```diff
@@ -101,7 +114,7 @@ const PhoneLoginForm = ({isFetching, onSubmit}: LoginFormProps) => {
             id="enter-phone"
             name="phone-login"
             onChange={onPhoneNumberChange}
-            style={{width: 260}}
+            style={{width: isTiny ? 'calc(100vw - 200px)' : isMobile ? 'calc(100vw - 250px)' : '280px'}}
```
V-Gira, line 117: See description

atomrc (approved):

PR #13921, opened 2022-10-21
runfix: show expand option for collections over limit [FS-1025]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1025" title="FS-1025" target="_blank">BugFS-1025</a>  [Web] File collection shows 'Show all' button before file limit is reached
  </td></table>
  <br />
 

Display `Show all` button only in the case of preexisting limit. I assume this `true ||` was a leftover after react migration.

### DIFF
--- a/src/script/page/MainContent/panels/Collection/Collection.test.tsx
+++ b/src/script/page/MainContent/panels/Collection/Collection.test.tsx
@@ -101,6 +101,9 @@ describe('Collection', () => {
   });
 
   it('displays collection details when a section is selected', async () => {
+    const imageMessages = new Array(14).fill(null).map(() => createImageMessage());
+    mockConversationRepository.getEventsForCategory.mockResolvedValueOnce(imageMessages);
+
     const {getAllByText, getByText} = render(
       <Collection conversation={conversation} conversationRepository={mockConversationRepository as any} />,
     );
@@ -114,13 +117,13 @@ describe('Collection', () => {
 
   it('should display search results when term is typed', async () => {
     jest.useFakeTimers();
-    const {getAllByText, queryByText, container} = render(
+    const {getAllByText, queryByText, getByTestId} = render(
       <Collection conversation={conversation} conversationRepository={mockConversationRepository as any} />,
     );
 
     await waitFor(() => getAllByText('CollectionItem'));
     await act(async () => {
-      const input: HTMLInputElement = container.querySelector('[data-uie-name=full-search-header-input]');
+      const input = getByTestId('full-search-header-input');
       fireEvent.change(input, {target: {value: 'term'}});
       jest.advanceTimersByTime(500);
       await waitFor(() => expect(mockConversationRepository.searchInConversation).toHaveBeenCalled());
--- a/src/script/page/MainContent/panels/Collection/Collection.test.tsx
+++ b/src/script/page/MainContent/panels/Collection/Collection.test.tsx
@@ -101,7 +101,8 @@ describe('Collection', () => {
   });
 
   it('displays collection details when a section is selected', async () => {
-    const imageMessages = new Array(14).fill(null).map(() => createImageMessage());
+    const IMAGE_COLLECTION_LENGTH = 13;
+    const imageMessages = new Array(IMAGE_COLLECTION_LENGTH).fill(null).map(createImageMessage);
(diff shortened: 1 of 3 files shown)


src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+import {render, fireEvent} from '@testing-library/react';
+
+import CollectionSection from './CollectionSection';
+import {ContentMessage} from '../../../../entity/message/ContentMessage';
+import {createRandomUuid} from 'Util/util';
+
+const NUMBER_OF_ASSETS = 5;
+
+const messages = new Array(NUMBER_OF_ASSETS).fill(null).map(() => new ContentMessage(createRandomUuid()));
```
przemvs, line 28: It doesn't work? 🤔 

```suggestion
const messages = new Array(NUMBER_OF_ASSETS).fill(new ContentMessage(createRandomUuid()));
```

src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const getDefaultProps = ({limit}: {limit: number}) => ({
```
przemvs: WDYT? You have only 1 param there :)

```suggestion
const getDefaultProps = (limit: number) => ({
```

src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+      </CollectionSection>,
+    );
+
+    props.limit = NUMBER_OF_ASSETS;
+    expect(queryByText('collectionShowAll')).toBeNull();
+
+    rerender(
+      <CollectionSection {...props}>
+        <span className={`collection-header-icon icon-library`}></span>
+      </CollectionSection>,
```
przemvs: Please create mock for it, and pass eg. <Collection {...props} />, you have 3 same Collection sections, only have different limit prop :)

src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
PatrykBuniX: nope, I need different instance each time, it would fill the array with the same object (with the same id)

src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
PatrykBuniX: I've ended up using just empty `<span />` its not important what's passed as a children here 👍

src/script/page/MainContent/panels/Collection/Collection.test.tsx
```diff
@@ -101,6 +101,9 @@ describe('Collection', () => {
   });
 
   it('displays collection details when a section is selected', async () => {
+    const imageMessages = new Array(14).fill(null).map(() => createImageMessage());
```
przemvs: Move 14 to enum/const :)

```suggestion
    const imageMessages = new Array(14).fill(null).map(createImageMessage);
```

przemvs: Also maybe create util for creating array with items :)
And have usage fn(length, callback) :)

src/script/page/MainContent/panels/Collection/CollectionSection.test.tsx
przemvs: So if children not needed so use <CollectionSection {...props} /> :D

PatrykBuniX: I did not say it's not needed... It's required, but we can pass anything we want here (this test does not test what's inside)

src/script/page/MainContent/panels/Collection/Collection.test.tsx
PatrykBuniX: Added variable, simplified callback passed to map. We're creating image messages this way only once, creating a util is too much for a single test.

przemvs (approved):
lgtm

PR #13922, opened 2022-10-21
runfix: Conversation read message and right-sidebar opening

przemvs:
- Fix for opening Right Sidebar on clikc invite pepole.
- Add animations while right sidebar is opened - right sidebar overlapping title-bar and input-bar
- Propably fix for readed messages, more details in comment in there (Conversation.tsx)

### DIFF
--- a/src/script/components/Conversation/Conversation.tsx
+++ b/src/script/components/Conversation/Conversation.tsx
@@ -292,7 +292,7 @@ const ConversationList: FC<ConversationListProps> = ({
   const getInViewportCallback = (
     conversationEntity: ConversationEntity,
     messageEntity: Message,
-  ): (() => void) | undefined => {
+  ): (() => void) | null => {
     const messageTimestamp = messageEntity.timestamp();
     const callbacks: Function[] = [];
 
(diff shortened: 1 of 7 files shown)


src/script/components/Conversation/Conversation.tsx
```diff
@@ -325,7 +325,7 @@ const ConversationList: FC<ConversationListProps> = ({
     const isUnreadMessage = messageTimestamp > conversationEntity.last_read_timestamp();
     const isNotOwnMessage = !messageEntity.user().isMe;
 
-    let shouldSendReadReceipt = true;
+    let shouldSendReadReceipt = false;
```
przemvs, line 328: Restore previous default value from dev branch: https://github.com/wireapp/wire-webapp/blob/3dc266c3c1fb90597a6472aaef776b52248777fc/src/script/view_model/content/MessageListViewModel.ts#L271

Propably i forgot to change this value while testing.

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -73,7 +74,7 @@ export interface MessageParams extends MessageActions {
     deleteMessageEveryone: (conversation: Conversation, message: BaseMessage) => void;
   };
   messageRepository: MessageRepository;
-  onVisible?: () => void;
+  onVisible?: (() => void) | null;
```
PatrykBuniX: Do we really need both `null` and `undefined`? Can't we stick to one falsy value? 😄

przemvs: done :D

PatrykBuniX (approved):
lgtm

PR #13956, opened 2022-10-24
runfix: Automatically focus on primary action button of primary modal

thisisamir98:
Needed to fix automation tests

### DIFF
--- a/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
+++ b/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {FC, FormEvent, MouseEvent, useState} from 'react';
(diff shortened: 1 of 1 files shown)


V-Gira (approved):

PatrykBuniX (approved):

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -165,7 +173,7 @@ export const PrimaryModalComponent: FC = () => {
                       id="clear-data-checkbox"
                       checked={optionChecked}
                       data-uie-name="modal-option-checkbox"
-                      onChange={event => updateOptionChecked(event.target.checked)}
+                      onChange={onOptionChange}
```
atomrc, line 176: hmmmm this seems very suscpicious. Do we really want to close the modal when a checkbox is clicked? That's sounds like a weird UX

atomrc (changes requested):

PR #13957, opened 2022-10-24
feat: show call buttons in scaled down conversation view [ACC-33]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-33" title="ACC-33" target="_blank">StoryACC-33</a>  Responsive app design (1.4.4 / 1.4.10)
  </td></table>
  <br />
 

----
### Issue
- users only receive call notification in the sidebar view in scaled down view

### Solution
- create a call notification component for the conversation view based on the sidebar calling cell.

Peek 2022-10-25 11-13

#### addition

- This PR also addresses [these code reviews](https://github.com/wireapp/wire-webapp/pull/13881) (move the media query logic from React to plain css)


src/script/components/calling/ConversationCallingCell.tsx
```diff
@@ -0,0 +1,200 @@
...
+import {generateConversationUrl} from '../../router/routeGenerator';
+import {createNavigate, createNavigateKeyboard} from '../../router/routerBindings';
+import {CallActions} from '../../view_model/CallingViewModel';
+
+export interface CallingCellProps {
+  call: Call;
+  callActions: CallActions;
+  callingRepository: Pick<CallingRepository, 'supportsScreenSharing' | 'sendModeratorMute'>;
+  callState?: CallState;
+  classifiedDomains?: string[];
```
przemvs: Optional types should be after required :)

src/script/components/calling/ConversationCallingCell.tsx
```diff
@@ -0,0 +1,200 @@
...
+              className="conversation-list-cell conversation-list-cell-button"
+              onClick={createNavigate(conversationUrl)}
+              onKeyDown={createNavigateKeyboard(conversationUrl)}
+              tabIndex={0}
+              role="button"
+              aria-label={t('accessibility.openConversation', conversationName)}
+            >
+              {!temporaryUserStyle && (
+                <div className="conversation-list-cell-left">
+                  {isGroup && <GroupAvatar users={conversationParticipants} isLight={true} />}
```
przemvs:
```suggestion
                  {isGroup && <GroupAvatar users={conversationParticipants} isLight />}
```

src/script/components/calling/ConversationCallingCell.tsx
```diff
@@ -0,0 +1,200 @@
...
+                {isOutgoing && (
+                  <span className="conversation-list-cell-description" data-uie-name="call-label-outgoing">
+                    {t('callStateOutgoing')}
+                  </span>
+                )}
+                {isConnecting && (
+                  <span className="conversation-list-cell-description" data-uie-name="call-label-connecting">
+                    {t('callStateConnecting')}
+                  </span>
+                )}
```
przemvs: Maybe do this. this way:

```
const callState = {
  [CALL_STATE.INCOMING]: {
    dataUieName: 'call-label-incoming',
    text: t('callStateIncoming')
  },
  ...other
}
```

```
const currentCallState = callState?.[state];

{currentCallState && (
<span className="conversation-list-cell-description" data-uie-name={currentCallState.dataUieName}>
                   {currentCallState.text} 
                  </span>
)}
```

src/script/components/calling/ConversationCallingCell.tsx
```diff
@@ -0,0 +1,200 @@
...
+              </div>
+            </div>
+
+            <div className="conversation-list-cell-right">
+              {isConnecting ||
+                (isOngoing && (
```
przemvs: It's okay? Or it should be `{(isConnecting || isOngoing) && ...}`

src/script/components/calling/ConversationCallingCell.tsx
V-Gira: Wouldn't this require to have a use case for all 8 states? 
image

We're only using a conditional for 3 of them.

(This code is directly ported from `ConversationListCallingCell`)

przemvs: You can use Partial, and if we should prepare next state it will be easier to define it to the object :)

atomrc: Oh it's actually a good idea!!
You will just have to handle those 3 cases (the rest will return `undefined`), and the code for the rendering is going to be much simpler 👌

V-Gira: Couldn't figure that out yesterday, thanks!

src/script/components/calling/ConversationCallingCell.tsx
```diff
@@ -0,0 +1,208 @@
...
+      dataUieName: 'call-label-incoming',
+      text: t('callStateIncoming'),
+    },
+    [CALL_STATE.ANSWERED]: {
+      dataUieName: 'call-label-connecting',
+      text: t('callStateConnecting'),
+    },
+  };
+
+  const currentCallStatus = callStatus[state];
```
przemvs: state will always truthy value ?

V-Gira: Something to do with Partial? I removed ?. on @aweiss-dev 's suggestion

src/script/components/calling/CallingCell.tsx
```diff
@@ -51,26 +51,31 @@ import {ContextMenuEntry, showContextMenu} from '../../ui/ContextMenu';
 import {initFadingScrollbar} from '../../ui/fadingScrollbar';
 import {CallActions, CallViewTab} from '../../view_model/CallingViewModel';
 
-export interface CallingCellProps {
+interface VideoCallProps {
+  multitasking: Multitasking;
+  hasAccessToCamera?: boolean;
+  isSelfVerified?: boolean;
+  teamState?: TeamState;
+}
```
V-Gira, line 59: Those are the props that are not used by the simpler answering controls component. Is there a better way than making them conditional?

src/script/components/Conversation/Conversation.tsx
```diff
@@ -374,6 +382,27 @@ const ConversationList: FC<ConversationListProps> = ({
...
+                    classifiedDomains={undefined}
+                    call={call}
+                    callActions={callingViewModel.callActions}
+                    callingRepository={callingRepository}
+                    conversation={conversation}
+                    multitasking={callingViewModel.multitasking}
+                  />
+                </div>
+              )
+            );
```
przemvs: It will be more readable ;)

```suggestion
            if (!conversation && !smBreakpoint) {
                return null;
            }
            
            return (
                <div className="calling-cell" key={conversation.id}>
                  <CallingCell
                    classifiedDomains={undefined}
                    call={call}
                    callActions={callingViewModel.callActions}
                    callingRepository={callingRepository}
                    conversation={conversation}
                    multitasking={callingViewModel.multitasking}
                  />
                </div>
            );
```

src/script/components/Conversation/Conversation.tsx
```diff
@@ -374,6 +382,27 @@ const ConversationList: FC<ConversationListProps> = ({
...
+          {activeCalls.map(call => {
+            const conversation = conversationState.findConversation(call.conversationId);
+            const callingViewModel = mainViewModel.calling;
+            const callingRepository = callingViewModel.callingRepository;
+            return (
+              conversation &&
+              smBreakpoint && (
+                <div className="calling-cell" key={conversation.id}>
+                  <CallingCell
+                    classifiedDomains={undefined}
```
przemvs: I think you don't need pass it because it's optional :)

src/script/components/calling/ConversationCallingCell.tsx
PatrykBuniX: Yeah, it looks good to me, state is always defined, it's initial value is `CALL_STATE.UNKNOWN` 👍

przemvs (commented):
1 thing :)

src/script/components/Conversation/Conversation.tsx
PatrykBuniX: You mean `or` @przemvs? 😄 
```js
if (!conversation || !smBreakpoint) {
     return null;
}
```

src/script/components/Conversation/Conversation.tsx
PatrykBuniX: yep, no need to pass `undefined` explicitly 👍

src/script/components/calling/CallingCell.tsx
```diff
@@ -51,26 +51,31 @@ import {ContextMenuEntry, showContextMenu} from '../../ui/ContextMenu';
...
+interface AnsweringControlsProps {
   call: Call;
   callActions: CallActions;
   callingRepository: Pick<CallingRepository, 'supportsScreenSharing' | 'sendModeratorMute'>;
+  conversation: Conversation;
+  fullUi?: boolean;
```
PatrykBuniX: I'd rather go with more booleanish naming, what about `isFullUI`? 😄

src/script/components/calling/CallingCell.tsx
```diff
@@ -118,6 +123,25 @@ const ConversationListCallingCell: React.FC<CallingCellProps> = ({
   const isConnecting = state === CALL_STATE.ANSWERED;
   const isOngoing = state === CALL_STATE.MEDIA_ESTAB;
 
+  type labels = {dataUieName: string; text: string};
```
PatrykBuniX: move type definition above component 😄

src/script/components/Conversation/Conversation.tsx
przemvs: He have in the code && :D

src/script/components/Conversation/Conversation.tsx
V-Gira: woops ! That''s an oversight actually, I should pass the actual ko observable here

src/script/components/Conversation/Conversation.tsx
PatrykBuniX: yeah, but to keep the logic you need to use `||` instead of `&&` when using negation

PatrykBuniX: we want both values to be truthy to display calling cell fragment right?

V-Gira: It's or indeed, I corrected it

src/script/components/Conversation/Conversation.tsx
```diff
@@ -374,6 +383,29 @@ const ConversationList: FC<ConversationListProps> = ({
...
+            const conversation = conversationState.findConversation(call.conversationId);
+            const callingViewModel = mainViewModel.calling;
+            const callingRepository = callingViewModel.callingRepository;
+
+            if (!conversation || !smBreakpoint) {
+              return null;
+            }
+
+            return (
+              <div className="calling-cell" key={conversation.id}>
```
przemvs, line 396: Clean! 🚀

przemvs (approved):
🚀

PatrykBuniX (approved):

PR #13965, opened 2022-10-24
chore: replace babel with swc

atomrc:
Should speedup:
- build time (expected 2 times faster)
- test time


.swcrc
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://json.schemastore.org/swcrc",
```
atomrc, line 2: Seems to be a problem with the browserslist integration :/ 
https://github.com/swc-project/swc-loader/issues/37

PR #13966, opened 2022-10-25
runfix: enable sending the same asset multiple times in a row

PatrykBuniX:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

User was not able to send the same asset two (or more) times in a row. 

The reason behind the issue was that the file input we are using to upload assets was not resetting after uploading a file, so when trying to upload the same file twice, `onChange` event was not triggered (input's `value` field was not changing).

The solution is to wrap the component with `<form>` element, and use it's native `.reset()` method. 

Before:
asset-before

After:
asset-after

----

----

----
##### References

### DIFF
--- a/src/script/page/message-list/AssetUploadButton.tsx
+++ b/src/script/page/message-list/AssetUploadButton.tsx
@@ -1,0 +1,69 @@
+/*
+ * Wire
+ * Copyright (C) 2022 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {useState, useRef} from 'react';
+import {Config} from '../../Config';
+import {t} from 'Util/LocalizerUtil';
+import Icon from 'Components/Icon';
+
+interface AssetUploadButtonProps {
+  onSelectFiles: (files: File[]) => void;
+}
+
+export const AssetUploadButton = ({onSelectFiles}: AssetUploadButtonProps) => {
+  const [inputKey, setInputKey] = useState(Date.now());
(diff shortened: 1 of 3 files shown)


src/script/page/message-list/InputBarControls/ControlButtons.tsx
```diff
@@ -61,6 +61,15 @@ const ControlButtons: React.FC<ControlButtonsProps> = ({
 
   const pingTooltip = t('tooltipConversationPing');
 
+  const handleFileChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const {files} = event.target;
+    if (!files) {
+      return;
+    }
+    onSelectFiles(Array.from(files));
```
przemvs: Maybe do it with spread operator ? :) 🤔
onSelectFiles([...files]);

src/script/page/message-list/AssetUploadButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+      type="button"
+      aria-label={t('tooltipConversationFile')}
+      title={t('tooltipConversationFile')}
+      className="conversation-button controls-right-button no-radius file-button"
+      onClick={() => fileRef.current?.click()}
+      data-uie-name="do-share-file"
+    >
+      <Icon.Attachment />
+      <input
+        key={inputKey}
```
PatrykBuniX: After uploading a file, we change the `key` attribute of file input, what forces react to re-render this element. Couldn't find a better and safe way to reset the value. Solution below also works, but testing environment complains that we shouldn't overwrite read-only properties.
```js
event.target.value = '';
//or
fileRef.current.value = '';
```
I'm open to any suggestions you guys have 😄

src/script/page/message-list/InputBarControls/ControlButtons.tsx
PatrykBuniX: It gives the same result...

src/script/page/message-list/InputBarControls/ControlButtons.tsx
```diff
@@ -119,25 +118,7 @@ const ControlButtons: React.FC<ControlButtonsProps> = ({
             </li>
 
             <li>
-              <button
```
PatrykBuniX, line 122: moved this part to separate component `AssetUploadButton.tsx`

src/script/page/message-list/AssetUploadButton.tsx
PatrykBuniX: The ideal solution would be to call `form.reset()`, but we don't have a form in this case...

PatrykBuniX: I've ended up wrapping the button and input with `<form>` element and using `.reset()` method.

arjita-mitra (approved):

tlebon (approved):
cool stuff 👍

PR #13971, opened 2022-10-26
runfix: MessageList read messages

przemvs:
Propably fix for reading messages, wrong element was focused to get informations about element

### DIFF
--- a/src/script/ui/onHitTopOrBottom.ts
+++ b/src/script/ui/onHitTopOrBottom.ts
@@ -18,10 +18,6 @@
  */
 
 const onHitTopOrBottom = (scrollingElement: HTMLElement | null, onHitTop: () => void, onHitBottom: () => void) => {
-  if (!scrollingElement) {
-    return;
-  }
-
   const onScroll = (event: Event) => {
     const element = event.target as HTMLElement;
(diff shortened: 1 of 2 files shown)


src/script/ui/onHitTopOrBottom.ts
```diff
@@ -17,12 +17,13 @@
...
-const onHitTopOrBottom = (element: HTMLElement | null, onHitTop: () => void, onHitBottom: () => void) => {
-  if (!element) {
+const onHitTopOrBottom = (scrollingElement: HTMLElement | null, onHitTop: () => void, onHitBottom: () => void) => {
+  if (!scrollingElement) {
     return;
   }
 
-  const onScroll = () => {
+  const onScroll = (event: Event) => {
+    const element = event.target as HTMLElement;
```
atomrc, line 22: Can you explain what is the difference between this event target and the initial `scrollingElement`? I don't think I understand why it fixes it?

przemvs: You're right, it not have differences, i'm changed this to previous funcitonality

atomrc (approved):

PR #13978, opened 2022-10-26
runfix: reset audioElement media after stopping [FS-1116]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1116" title="FS-1116" target="_blank">BugFS-1116</a>  [Web] User is able to play calling sound after ending a call.
  </td></table>
  <br />
 

## Issue (funny one):

User was able to play calling audio sound after the call was ended (and probably some other sounds also) by using keyboard play shortcut or menu top-bar.

https://user-images.githubusercontent.com/45733298/198037606-0588d8bc-49c3-4b24-8c3b-aeaa8e02879c.mp4

## Cause
The reason behind the issue was the stop method we are using for stopping sound effects. The stop method uses `audio.pause()` method, so macOS want to do wire user a favour and let them play paused audio.

## Solution
The solution is to simply call `audio.load()` method after `audio.pause()`. Regarding to [MDN doc](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/load) this methods resets the media element to the initial state (not playing) and removes audio from macOS's quick access pause-play functionality.
Screenshot 2022-10-26 at 15 16 35

### DIFF
--- a/src/script/audio/AudioRepository.ts
+++ b/src/script/audio/AudioRepository.ts
@@ -206,6 +206,7 @@ export class AudioRepository {
     if (!audioElement?.paused) {
       this.logger.info(`Stopping sound '${audioId}'`);
       audioElement.pause();
+      audioElement.load();
     }
   };
 


V-Gira (approved):

src/script/audio/AudioRepository.ts
```diff
@@ -206,6 +206,7 @@ export class AudioRepository {
     if (!audioElement?.paused) {
       this.logger.info(`Stopping sound '${audioId}'`);
       audioElement.pause();
```
tlebon, line 208: should pause still be here also?

atomrc (approved):

src/script/audio/AudioRepository.ts
PatrykBuniX: Just tested and it seems to be working fine without `.pause()`, but I'm not sure if `.load()` will always pause playing so it's better to leave it there in my opinion

PR #13980, opened 2022-10-26
runfix: Closing right sidebar

przemvs:
Fix for toggle right sidebar when user click on user details.

Previous:
User click on different user when have open user - it closing right sidebar.

Now:
User click on different user when have open user - right sidebar changing details for clicked user.


V-Gira (approved):
Description to the point and really helpful :+1:

PR #14017, opened 2022-10-31
refactor: change px values to rem values for font variables [ACC-68]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-68" title="ACC-68" target="_blank">SubtaskACC-68</a>  Use em as a font size in the web app
  </td></table>
  <br />
 

----

### Issue

- introduction of a font size slider needs a refactor of font sizes and line heights px values to rem

#### Changes

- changes font size and line height variables from px to rem
- apply variables when possible
- refactor other px values for font size and line height to rem


src/style/content/conversation/input-bar.less
```diff
@@ -164,7 +164,7 @@
     color: var(--white);
 
     > svg {
-      margin-left: 0.3em;
+      margin-left: 4px;
```
V-Gira, line 169: looks counter intuitive to be introducing px values in this PR but this shouldn't be responsive with font-size changes

src/style/content/conversation/input-bar.less
```diff
@@ -52,8 +52,8 @@
   min-height: @conversation-input-line-height;
   max-height: @conversation-input-max-height;
   flex: 1 1;
-  margin-top: (@conversation-input-min-height - @conversation-input-line-height) / 2;
-  margin-bottom: (@conversation-input-min-height - @conversation-input-line-height) / 2;
+  margin-top: calc((@conversation-input-min-height - @conversation-input-line-height) / 2);
+  margin-bottom: calc((@conversation-input-min-height - @conversation-input-line-height) / 2);
```
atomrc, line 56: why do we need to use `calc` now? Is that a dynamic computation now?

src/style/common/variables.less
```diff
@@ -93,27 +93,27 @@
...
+@font-size-lg: 1.5rem;
+@font-size-xl: 2rem;
+@font-size-xxl: 3.5rem;
+
+@font-size-xsmall: 0.6875rem;
+@font-size-small: 0.75rem;
+@font-size-medium: 0.875rem;
+@font-size-base: 1rem;
+@font-size-large: 1.25rem;
+@font-size-xlarge: 1.5rem;
```
atomrc, line 109: do you think we could merge those 2 sets into a single one? 
I don't think I understand the difference between the 2

atomrc (approved):

PatrykBuniX (approved):

src/style/content/conversation/input-bar.less
V-Gira: I'd say so, it wouldn't work because of the em values

atomrc: OOhh right, the `conversation-input-line-height` is now in `em`, right 👌 

It's impossible to see on the PR cause this variable is affected to another variable value, thus my question

src/style/common/variables.less
V-Gira: Top ones are legacy and bottom ones date from the first go at a11y changes 7 months ago.

I kinda assumed there was a reason the legacy one didn't get refactored but there's probably none.

I'll see what a refactor would look like now

PR #14020, opened 2022-10-31
runfix: add convo protocol details back

PatrykBuniX:
Add ConversationProtocolDetails component back to the right sidebar after it was lost during knockout -> react migration.

### DIFF
--- a/src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.test.tsx
+++ b/src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.test.tsx
@@ -31,7 +31,7 @@ describe('ConversationProtocolDetails', () => {
 
     const {queryByText} = render(<ConversationProtocolDetails {...props} />);
 
-    expect(queryByText('mls')).not.toBeNull();
+    expect(queryByText('MLS')).not.toBeNull();
     expect(queryByText('MLS_128_DHKEMP256_AES128GCM_SHA256_P256')).not.toBeNull();
   });
 
@@ -42,6 +42,6 @@ describe('ConversationProtocolDetails', () => {
 
     const {queryByText} = render(<ConversationProtocolDetails {...props} />);
 
-    expect(queryByText('proteus')).not.toBeNull();
+    expect(queryByText('PROTEUS')).not.toBeNull();
   });
 });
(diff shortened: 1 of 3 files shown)


src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.tsx
```diff
@@ -81,14 +81,14 @@ const wrapperStyles: CSSObject = {
...
+    <div>
       <div className="conversation-details__list-head">{t('conversationDetailsProtocolDetails')}</div>
       <div css={wrapperStyles}>
         <div css={titleStyles}>Protocol</div>
         <div css={subTitleStyles} data-uie-name="protocol-name">
-          {protocol}
+          {protocol.toUpperCase()}
         </div>
-        {protocol.toLocaleLowerCase() === ConversationProtocol.MLS && cipherSuite && (
+        {protocol === ConversationProtocol.MLS && cipherSuite && (
```
PatrykBuniX, line 91: We were passing already uppercased version before. Knockout-rendered components do not check prop type validity. We have to pass `ConversationProtocol` now (which is lowercase) and render it uppercased.

src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.tsx
```diff
@@ -81,14 +81,14 @@ const wrapperStyles: CSSObject = {
 
 const ConversationProtocolDetails: React.FC<ConversationProtocolDetailsProps> = ({protocol, cipherSuite}) => {
   return (
-    <>
+    <div>
```
PatrykBuniX, line 84: We need to wrap all of its content with block element - knockout components seems to be rendered as block elements - taking all horizontal space. Styles were broken when rendered in react without the `div`

src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('ConversationProtocolDetails', () => {
 
     const {queryByText} = render(<ConversationProtocolDetails {...props} />);
 
-    expect(queryByText('mls')).not.toBeNull();
+    expect(queryByText('MLS')).not.toBeNull();
```
PatrykBuniX, line 34: From now, we're passing lowercase as a prop but displaying uppercase, see [this comment](https://github.com/wireapp/wire-webapp/pull/14020/files#r1009467528)

tlebon (approved):

V-Gira (approved):

PR #14021, opened 2022-11-02
refactor: Full migration from grunt to webpack

thisisamir98:
Let's bundle everything with webpack 📦


package.json
```diff
@@ -92,10 +92,10 @@
     "@wireapp/eslint-config": "2.0.0",
     "@wireapp/prettier-config": "0.5.0",
     "adm-zip": "0.5.9",
+    "archiver": "^5.3.1",
```
atomrc, line 95: Since we already have `zipjs` in the repo, can we use it for archive generation?

przemvs (approved):
Code LGTM! :)

src/page/index.ejs
```diff
@@ -0,0 +1,46 @@
...
+    <link href="/image/favicon.ico?<%= VERSION %>" type="image/x-icon" rel="shortcut icon" sizes="48x48" />
+    <link href="/image/meta/wire-icon-256.png?<%= VERSION %>" rel="icon" sizes="256x256" type="image/png" />
+    <link href="/image/meta/wire-icon-128.png?<%= VERSION %>" rel="icon" sizes="128x128" type="image/png" />
+    <link href="/image/meta/wire-icon-64.png?<%= VERSION %>" rel="icon" sizes="64x64" type="image/png" />
+
+    <meta property="og:title" content="<%= OPEN_GRAPH_TITLE %>" />
+    <meta property="og:description" name="description" content="<%= OPEN_GRAPH_DESCRIPTION %>" />
+    <meta property="og:type" content="website" />
+    <meta property="og:url" content="<%= APP_BASE %>" />
+    <meta property="og:image" content="<%= OPEN_GRAPH_IMAGE_URL %>" />
```
atomrc, line 21: Let's extract those in a common template (since it's copy pasted now)

PatrykBuniX (approved):

server/package.json
```diff
@@ -37,9 +37,10 @@
     "typescript": "4.8.4"
   },
   "scripts": {
-    "build": "yarn clean && tsc && yarn copy-assets && yarn create-commit-sha",
+    "build": "yarn clean && tsc && yarn copy-assets && yarn create-commit-sha && yarn create-version-file",
     "copy-assets": "node ./bin/copy_server_assets.js",
     "create-commit-sha": "node ./bin/create_commit_sha_file.js",
+    "create-version-file": "node ./bin/create_version_file.js",
```
atomrc, line 43: super nice that this is hidden in the server now 👍

V-Gira (approved):

server/bin/create_version_file.js
```diff
@@ -17,10 +17,24 @@
...
+}
+
+function formatDate(date) {
+  return `${[date.getFullYear(), padTo2Digits(date.getMonth() + 1), padTo2Digits(date.getDate())].join('.')}.${[
+    padTo2Digits(date.getHours()),
+    padTo2Digits(date.getMinutes()),
+  ].join('.')}`;
+}
+
+const version = formatDate(new Date());
```
atomrc: Why not using the previous technique?

```suggestion
const {format} = require('date-fns');
const version = format(new Date(), 'yyyy.MM.dd.HH.mm');
```

src/page/index.ejs
thisisamir98: I already tried to do so but unfortunately, webpack html plugin does not support partials 😢

atomrc: Alright no biggies :)

atomrc (approved):

PR #14024, opened 2022-11-02
feat: Send external commits to join conversation from new device

atomrc:
External proposals was the initial way we dealt with joining conversation with a new device. 
Now that external commits are implemented on backend, we could directly join with an external commit


przemvs (approved):
code LGTM

PatrykBuniX (approved):

PR #14037, opened 2022-11-03
runfix: Address emoji picker styles

atomrc:
This PR fixes the display of the emoji picker than was broken since https://github.com/wireapp/wire-webapp/pull/13493

It also modernizes the styles and migrate them to css-in-js

Before 
image

After 
image

### DIFF
--- a/src/script/components/Emoji/EmojiItem.tsx
+++ b/src/script/components/Emoji/EmojiItem.tsx
@@ -39,7 +39,6 @@ const EmojiItem: FC<EmojiItemProps> = ({emoji, onClick, onMouseEnter, selectedEm
     onMouseEnter={onMouseEnter}
     onClick={onClick}
     aria-label={emoji.name}
-    tabIndex={0}
   >
     <span css={symbolStyle}>{emoji.icon}</span>
     <span css={nameStyle}>{emoji.name}</span>
--- a/src/script/components/Emoji/EmojiItem.tsx
+++ b/src/script/components/Emoji/EmojiItem.tsx
@@ -21,6 +21,7 @@ import {FC} from 'react';
 
 import cx from 'classnames';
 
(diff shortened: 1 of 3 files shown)


src/script/components/Emoji/EmojiItem.tsx
```diff
@@ -33,15 +34,15 @@ interface EmojiItemProps {
 const EmojiItem: FC<EmojiItemProps> = ({emoji, onClick, onMouseEnter, selectedEmoji = false}) => (
   <button
     type="button"
```
atomrc, line 36: @arjita-mitra I think this `type=button` is not necessary here. Is it?

atomrc: I believe it's only needed for `input`, am I correct?

V-Gira (approved):
I'll take that as an endorsment for team `CSSObject`

src/script/components/Emoji/EmojiItem.tsx
atomrc: Oh actually I found my answer and it is needed :)

atomrc:
> I'll take that as an endorsment for team CSSObject

I'm team css-in-js yes :) (also because it's the solution we already have baked in)

PatrykBuniX (approved):

src/script/components/Emoji/EmojiItem.tsx
V-Gira: Same for `tabIndex={0}`, looking like leftovers from changing the component from `div` to `button`

atomrc: Oh correct the tabindex is not needed, good catch 👍

aweiss-dev (approved):

PR #14040, opened 2022-11-04
feat: add api v3 support for notifications endpoint [FS-1139]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1139" title="FS-1139" target="_blank">TaskFS-1139</a>  [Web] Implement API V3
  </td></table>
  <br />
 

Adds api v3 support to the client.
Fixes storing id of missed notification.


src/script/event/EventRepository.ts
```diff
@@ -260,7 +258,7 @@ export class EventRepository {
     const shouldUpdatePersistedId = missedNotificationId !== notificationId;
     if (shouldUpdatePersistedId) {
       amplify.publish(WebAppEvents.CONVERSATION.MISSED_EVENTS);
-      this.notificationService.saveMissedIdToDb(this.lastNotificationId());
+      this.notificationService.saveMissedIdToDb(missedNotificationId);
```
PatrykBuniX, line 261: The value of `this.lastNotificationId()` is not being ever set, what results in always storing `undefined` under  `'z.storage.StorageKey.NOTIFICATION.MISSED'` key (amplify storage). Seems to be a leftover after these changes: https://github.com/wireapp/wire-webapp/pull/13087/files#diff-8424911ba55dc95ddd7ee35e77e651a8894faa0c33cd485cbbf62f6537fe3d24L296

PatrykBuniX: It was the reason error message was shown on every render instead of showing a message when notification's id has changed (what believe is the expected behaviour) 
Screenshot 2022-11-04 at 18 00 44

przemvs (approved):
LGTM :)

atomrc (approved):

atomrc (approved):

PR #14064, opened 2022-11-07
runfix: Init apiClient version before initializing the app

atomrc:
We need to make sure the api version is set before we do anything to bootstrap the app.

### DIFF
--- a/src/script/main/index.tsx
+++ b/src/script/main/index.tsx
@@ -22,6 +22,7 @@ import 'core-js/full/reflect';
 import {ClientType} from '@wireapp/api-client/lib/client/';
 import {Runtime} from '@wireapp/commons';
 import {createRoot} from 'react-dom/client';
+import {container} from 'tsyringe';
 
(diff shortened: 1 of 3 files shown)


src/script/main/index.tsx
```diff
@@ -32,10 +33,14 @@ import {doRedirect} from './app';
 
 import {SIGN_OUT_REASON} from '../auth/SignOutReason';
 import {Config} from '../Config';
+import {APIClient} from '../service/APIClientSingleton';
 import {StorageKey} from '../storage';
 
 document.addEventListener('DOMContentLoaded', async () => {
+  const apiClient = container.resolve(APIClient);
   const config = Config.getConfig();
+  await apiClient.useVersion(config.SUPPORTED_API_VERSIONS);
```
atomrc, line 42: This is the important fix, we need to set the version before initiating the core

przemvs (approved):

PatrykBuniX (approved):

PR #14076, opened 2022-11-09
fix: load polyfills on top

aweiss-dev:
Eslint import/order suggests putting React and possibly other packages on top of the imports. 
We need to take care manually that polyfills and needed unassigned imports are included before our assigned imports.
Eslint will not autofix unassigned imports, so we can put them where we would like to have them and ignore the imort/order rule in the file.

### DIFF
--- a/src/script/auth/main.tsx
+++ b/src/script/auth/main.tsx
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2022 Wire Swiss GmbH
+ * Copyright (C) 2018 Wire Swiss GmbH
  *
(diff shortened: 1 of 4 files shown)


src/script/auth/main.tsx
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2018 Wire Swiss GmbH
+ * Copyright (C) 2022 Wire Swiss GmbH
```
atomrc:
```suggestion
 * Copyright (C) 2018 Wire Swiss GmbH
```

atomrc (approved):

PR #14082, opened 2022-11-09
refactor: cleanup the conversationState

atomrc:
This cleans the conversationState and improve the `selfConversation` detection. 

Previously the selfConversation was found using the userId. Now we use the type of the conversation to know which one is the selfConversation. This is way better semantic and the id of the conversation will probably change with MLS

PR #14085, opened 2022-11-10
runfix: Always use sorted/filtered conversations for conversation list

atomrc:
regression introduced by #14082

### DIFF
--- a/src/script/view_model/ListViewModel.ts
+++ b/src/script/view_model/ListViewModel.ts
@@ -123,7 +123,7 @@ export class ListViewModel {
       const hasConnectRequests = !!this.userState.connectRequests().length;
       const states: (string | Conversation)[] = hasConnectRequests ? [ContentState.CONNECTION_REQUESTS] : [];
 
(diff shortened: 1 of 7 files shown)


PatrykBuniX (approved):

src/script/view_model/ListViewModel.ts
```diff
@@ -123,7 +123,7 @@ export class ListViewModel {
       const hasConnectRequests = !!this.userState.connectRequests().length;
       const states: (string | Conversation)[] = hasConnectRequests ? [ContentState.CONNECTION_REQUESTS] : [];
 
-      return states.concat(this.conversationState.unarchivedConversations());
+      return states.concat(this.conversationState.visibleConversations());
```
przemvs, line 126:
```suggestion
      return [...states, this.conversationState.visibleConversations()];
```

WDYT ?

atomrc: In theory it's a little less performant since you need to duplicate (by spreading) the initial array and then create a new one. 
`.concat` will only create a single new array

PR #14087, opened 2022-11-10
runfix: cleanup releasing the loaded conversation

atomrc:
This cleans up the process of releasing the current loaded conversation when switching content. 

It basically leverage the React `useEffect` feature to detect a change in the current conversation and releases the previous conversation when unmounting or switching conversation inside of the component

### DIFF
--- a/src/script/components/MessagesList/MessageList.tsx
+++ b/src/script/components/MessagesList/MessageList.tsx
@@ -250,6 +250,7 @@ const MessagesList: FC<MessagesListParams> = ({
         onLoading(false);
       }, 10);
     });
+    return () => conversation.release();
   }, [conversation, initialMessage]);
 
   useLayoutEffect(() => {
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -107,6 +107,13 @@ export class ConversationState {
   }
 
   /**
+   * returns true if the conversation should be visible to the user
+   * @param conversation the conversation to check visibility
+   */
+  isVisible(conversation: Conversation): boolean {
+    return this.visibleConversations().some(conv => matchQualifiedIds(conv.qualifiedId, conversation.qualifiedId));
+  }
+  /**
    * Get unarchived conversation with the most recent event.
    * @param allConversations Search all conversations
    * @returns Most recent conversation
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -40,7 +40,7 @@ export class ConversationState {
   /**
    * current conversation that is being viewed
    */
-  public readonly activeConversation = ko.observable<Conversation | null>(null);
+  public readonly activeConversation = ko.observable<Conversation | undefined>();
   /**
    * ordered list of conversations that are actives. This is basically the conversations we want to show to the user
    */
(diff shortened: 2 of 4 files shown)


src/script/components/MessagesList/MessageList.tsx
```diff
@@ -250,6 +250,7 @@ const MessagesList: FC<MessagesListParams> = ({
         onLoading(false);
       }, 10);
     });
+    return () => conversation.release();
```
atomrc, line 253: We just need to release the conversation when the message list is being unloaded or the conversation changes :)

src/script/page/useAppState.ts
```diff
@@ -77,21 +76,19 @@ const useAppState = create<AppState>((set, get) => ({
...
-  setContentState: (contentState: ContentState) =>
+  setContentState: (contentState: ContentState) => {
+    const previousContentState = get().contentState;
     set(state => ({
       ...state,
       contentState,
-    })),
+      previousContentState,
+    }));
+  },
```
atomrc, line 86: Now setting the previous content is automatically done when setting a new state. 
There was an invisible bug in the implementation that occured when switching to the startUI (that keeps the conversation content on the right side)

src/script/view_model/ContentViewModel.ts
```diff
@@ -141,14 +140,6 @@ export class ContentViewModel {
...
 
   changeConversation = (conversationEntity: Conversation, messageEntity?: Message) => {
-    // Clean up old conversation
-    const conversation = this.conversationState.activeConversation();
-
-    if (conversation) {
-      conversation.release();
-    }
-
-    // Update new conversation
```
atomrc, line 151: This is covered by the hook now

src/script/view_model/ContentViewModel.ts
```diff
@@ -301,8 +288,7 @@ export class ContentViewModel {
   };
 
   private readonly releaseContent = (newContentState: ContentState) => {
-    const {contentState, previousContentState, setPreviousContentState} = useAppState.getState();
-    setPreviousContentState(contentState);
```
atomrc, line 305: Now when setting a content, it will keep track of the previous one

src/script/conversation/ConversationState.ts
```diff
@@ -106,6 +106,13 @@ export class ConversationState {
     });
   }
 
+  /**
+   * returns true if the conversation should be visible to the user
+   * @param conversation the conversation to check visibility
+   */
+  isVisible(conversation: Conversation): boolean {
```
przemvs: Maybe use conversation as optional and return false if conversation will be undefined :)

src/script/view_model/ContentViewModel.ts
```diff
@@ -276,11 +264,7 @@ export class ContentViewModel {
         this.switchContent(ContentState.CONNECTION_REQUESTS);
       }
 
-      const repoHasConversation = this.conversationState
-        .conversations()
-        .some(conversation => this.previousConversation && matchQualifiedIds(conversation, this.previousConversation));
-
-      if (this.previousConversation && repoHasConversation && !this.previousConversation.is_archived()) {
+      if (this.previousConversation && this.conversationState.isVisible(this.previousConversation)) {
```
przemvs: If you add optional prop in isVIsible you will don't need to check if conversation exist before :)

atomrc: Fair point, will do 👍

atomrc: done :)

PatrykBuniX (approved):

PR #14101, opened 2022-11-14
runfix: Avoid releasing active conversation for conversation states

atomrc:
This condition was wrongly translated into the Conversation react hooks. 
It should not release the active conversation when loading the `collection` state

### DIFF
--- a/src/script/page/MainContent/MainContent.tsx
+++ b/src/script/page/MainContent/MainContent.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {FC, ReactNode, useContext, useState} from 'react';
+import {FC, ReactNode, useContext, useEffect, useState} from 'react';
 
 import cx from 'classnames';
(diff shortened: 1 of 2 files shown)


src/script/page/MainContent/MainContent.tsx
```diff
@@ -70,6 +70,14 @@ const MainContent: FC<MainContentProps> = ({
 
   const {contentState} = useAppState();
 
+  useEffect(() => {
+    const stateWithConversation = [ContentState.WATERMARK, ContentState.COLLECTION, ContentState.CONVERSATION];
```
przemvs: We need this in useEffect? We can't implement it above the component declarations ? 🤔

przemvs (commented):
1 thing :)

PatrykBuniX (approved):

V-Gira (approved):

przemvs (approved):
🚀

PR #14103, opened 2022-11-14
runfix: redirect to conversations when group creation fails

atomrc:
Makes sure we display the conversations list even if group creation fails

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -35,6 +35,7 @@ import {TextInput} from 'Components/TextInput';
 import {BaseToggle} from 'Components/toggle/BaseToggle';
 import {InfoToggle} from 'Components/toggle/InfoToggle';
 import {UserSearchableList} from 'Components/UserSearchableList';
+import {ListState, useAppState} from 'src/script/page/useAppState';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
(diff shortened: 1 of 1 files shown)


src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -216,7 +217,12 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
           },
         );
         setIsShown(false);
-        amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity, {});
+        if (conversationEntity) {
+          amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity, {});
+        } else {
+          // If an error occured, we go back the the conversations list
```
PatrykBuniX:
```suggestion
          // If an error occured, we go back to the conversations list
```

PatrykBuniX (approved):

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -216,7 +217,12 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
+          useAppState.getState().setListState(ListState.CONVERSATIONS);
```
przemvs, line 224: You can use hook const { setListState} = useAppState() instead of this way :)

atomrc: I'd argue that this one is shorter and just a single line

PR #14106, opened 2022-11-14
runfix: Improve list/content state sync

atomrc:


PatrykBuniX (approved):

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -217,13 +216,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
           },
         );
         setIsShown(false);
-        if (conversationEntity) {
```
tlebon, line 220: so what happens if the conversation isnt found?

PR #14112, opened 2022-11-15
feat: Focus on first focusable element when opening right menu (ACC-296)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-296" title="ACC-296" target="_blank">SubtaskACC-296</a>  Focus switch on opening User Profile/Group Information
  </td></table>
  <br />
 

Focus on first focusable element when opening right menu (ACC-296)

### DIFF
--- a/src/script/page/RightSidebar/RightSidebar.tsx
+++ b/src/script/page/RightSidebar/RightSidebar.tsx
@@ -26,6 +26,7 @@ import {container} from 'tsyringe';
 import {WebAppEvents} from '@wireapp/webapp-events';
 
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
+import {focusableElements} from 'Util/util';
 
 import {AddParticipants} from './AddParticipants';
 import {ConversationDetails} from './ConversationDetails';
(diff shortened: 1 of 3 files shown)


src/script/page/RightSidebar/RightSidebar.tsx
```diff
@@ -166,6 +166,18 @@ const RightSidebar: FC<RightSidebarProps> = ({
     });
   }, []);
 
+  const containerRef = useCallback(
+    (element: HTMLDivElement | null) => {
+      const focussableElements =
+        'a:not([disabled]), button:not([disabled]), input[type=text]:not([disabled]), [tabindex]:not([disabled]):not([tabindex="-1"])';
```
PatrykBuniX: Maybe we should move this string to some const? We're using it in two places now (check out `util.ts` file)

PatrykBuniX: + maybe also create a custom hook for this functionality e.g `useFocusFirstElement` that will return ref callback 🤔, wdyt? seems pretty reusable to me

src/script/page/RightSidebar/RightSidebar.tsx
```diff
@@ -166,6 +166,18 @@ const RightSidebar: FC<RightSidebarProps> = ({
...
+      const nextElementToFocus = element?.querySelectorAll(focussableElements)[0] as HTMLElement;
```
V-Gira: Damn, would have taken forever to figure out :D :+1:

atomrc:
```suggestion
      const nextElementToFocus = element?.querySelector(focussableElements) as HTMLElement;
```

thisisamir98: Found it on https://stackoverflow.com/a/35173443/7091035

PatrykBuniX (approved):

PR #14113, opened 2022-11-15
runfix: Clear Registration Form after Create account

przemvs:
Clear Registration form after Create account.

Previous:
- User filled form, and click register button
- User navigated to Verify Email Page
- User click on navigation back
- User click create acc
- User has filled form with previous values

Now:
- User filled form, and click register button
- User navigated to Verify Email Page
- User click on navigation back
- User click create acc
- Form has empty values

### DIFF
--- a/src/script/auth/component/AccountForm.tsx
+++ b/src/script/auth/component/AccountForm.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import React from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 
 import {FormattedMessage, useIntl} from 'react-intl';
 import {connect} from 'react-redux';
@@ -27,6 +27,7 @@ import {ValidationUtil} from '@wireapp/commons';
 import {Button, Checkbox, CheckboxLabel, Form, Input, Small} from '@wireapp/react-ui-kit';
 
 import {KEY} from 'Util/KeyboardUtil';
+import {getLogger} from 'Util/Logger';
 
 import {Exception} from './Exception';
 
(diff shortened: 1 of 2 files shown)


thisisamir98 (approved):

src/script/auth/component/AccountForm.tsx
```diff
@@ -39,77 +40,89 @@ import {RootState, bindActionCreators} from '../module/reducer';
...
+  const [registrationData, setRegistrationData] = useState({
     accent_id: AccentColor.random().id,
-    email: account.email || '',
-    name: account.name || '',
-    password: account.password || '',
-    termsAccepted: account.termsAccepted || false,
+    email: '',
+    name: '',
+    password: '',
+    termsAccepted: false,
```
atomrc, line 59: I think there are scenarios where the user data is already filled in, we should probably not change this

atomrc (changes requested):

src/script/auth/component/AccountForm.tsx
przemvs: We have code in useEffect which update it:
```
useEffect(() => {
    setRegistrationData(prevState => ({
      ...prevState,
      email: account?.email || '',
      name: account?.name || '',
      password: account?.password || '',
      termsAccepted: !!account?.termsAccepted,
    }));
  }, [account]);
```

atomrc: oohh right

atomrc (approved):

PR #14114, opened 2022-11-15
runfix: Disable typing indicator for federated environments

thisisamir98:
Currently typing endpoint is not implemented on backend for federated environments

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1711,10 +1711,26 @@ export class ConversationRepository {
   }
 
   public async sendTypingStart(conversationEntity: Conversation) {
+    /*
+      Currently typing endpoint is not implemented on backend for federated environments
+      @todo: remove this condition when backend is ready and api-client in packages is updated to support domain in typing endpoint
+    */
+    const isFederated = this.core.backendFeatures?.isFederated;
+    if (isFederated) {
+      return;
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -1711,10 +1711,26 @@ export class ConversationRepository {
   }
 
   public async sendTypingStart(conversationEntity: Conversation) {
+    /*
+      Currently typing endpoint is not implemented on backend for federated environments
+      @todo: remove this condition when backend is ready and api-client in packages is updated to support domain in typing endpoint
+    */
+    const isFederated = this.core.backendFeatures?.isFederated;
```
przemvs, line 1718: Maybe declare this after constructor ?

```suggestion
    this.isFederated = this.core.backendFeatures?.isFederated;
```

przemvs (approved):
LGTM, i know it's a temporary solution so you can merge it without those changes ;)

src/script/conversation/ConversationRepository.ts
thisisamir98: image
we're in a method and it has nothing to do with the main class so why this.?

przemvs: I commented after review it :)

atomrc (approved):

PR #14116, opened 2022-11-15
feat: use streamlined Decryption Error from core

aweiss-dev:
Just update to the newest core version and fix typing related to the core


przemvs (approved):

V-Gira (commented):
Bit of a nitpick but since it's a feature, maybe put the feature in the title? Like `feat: use streamlined Decryption Error from core`

V-Gira (approved):

PR #14119, opened 2022-11-15
runfix: Show legal hold request modal when view mounts

atomrc:
Right now we try to trigger the legal hold request modal too early (before the team is loaded). 
We need to trigger it once the team is loaded and the app ready to be view. 

Triggering it when the main content loads is the perfect moment to do so

### DIFF
--- a/src/script/page/MainContent/MainContent.tsx
+++ b/src/script/page/MainContent/MainContent.tsx
@@ -28,6 +28,7 @@ import {Conversation} from 'Components/Conversation';
 import {HistoryExport} from 'Components/HistoryExport';
 import {HistoryImport} from 'Components/HistoryImport';
 import {Icon} from 'Components/Icon';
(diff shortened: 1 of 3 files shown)


src/script/page/MainContent/MainContent.tsx
```diff
@@ -77,6 +81,13 @@ const MainContent: FC<MainContentProps> = ({
     }
   }, [contentState, conversationState]);
 
+  useEffect(() => {
+    if (teamState.supportsLegalHold()) {
+      const {showRequestModal} = useLegalHoldModalState.getState();
```
przemvs: Maybe move it to the line 76, we have 1 zustand hook there, so make it consistent ;)

przemvs (commented):
1 thing

PatrykBuniX (approved):

V-Gira (approved):

PR #14128, opened 2022-11-16
runfix: Correctly send stop typing on blur

thisisamir98:
Correctly send stop typing on blur

### DIFF
--- a/src/script/components/InputBar/InputBar.tsx
+++ b/src/script/components/InputBar/InputBar.tsx
@@ -403,7 +403,14 @@ const InputBar = ({
     return () => {
       window.clearTimeout(timerId);
     };
-  }, [isTyping, inputValue, conversationRepository, conversationEntity]); // we want to send is typing based on isTyping & changes to input
+    /*
+      we need inputValue as a dependency because every time a character is typed
+      we want to delay/postpone the timer,
+      basically adding more seconds to it's cancellation due time
+      so if the user is still typing they would
+      stay in the typing indicator
+    */
+  }, [isTyping, inputValue, conversationRepository, conversationEntity]);
 
   const replyMessage = (messageEntity: ContentMessage): void => {
     if (messageEntity?.isReplyable() && messageEntity !== replyMessageEntity) {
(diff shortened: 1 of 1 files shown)


src/script/components/InputBar/InputBar.tsx
```diff
@@ -392,18 +391,19 @@ const InputBar = ({
...
+      }, CONFIG.IS_TYPING_TIMEOUT);
+    } else {
       conversationRepository.sendTypingStop(conversationEntity);
-      isTypingTimerIdRef.current = null;
-      setIsTyping(false);
-    }, CONFIG.IS_TYPING_TIMEOUT);
-  }, [isTyping]); // we want to send is typing based on isTyping
+    }
+    return () => {
+      clearTimeout(timerId);
```
atomrc: I guess we should also send the typingStop message in this case, isn't it?

src/script/components/InputBar/InputBar.tsx
```diff
@@ -392,18 +391,19 @@ const InputBar = ({
...
+    };
+  }, [isTyping, inputValue, conversationRepository, conversationEntity]); // we want to send is typing based on isTyping & changes to input
```
atomrc: Do you really need the `inputValue` there? 🤔

thisisamir98: yes, because every time a character is typed we want to delay/postpone the timer, basically adding more seconds to it's cancellation due time so if the user is still typing they would stay in the typing indicator

src/script/components/InputBar/InputBar.tsx
```diff
@@ -498,7 +505,11 @@ const InputBar = ({
 
     const {value: currentValue} = event.currentTarget;
     setInputValue(currentValue);
-    setIsTyping(true);
+    if (currentValue.length > 0) {
+      setIsTyping(true);
+    } else {
+      setIsTyping(false);
+    }
```
przemvs:
```suggestion
  setIsTyping(currentValue.length > 0);
```
Hmmm ?

atomrc (approved):

src/script/components/InputBar/InputBar.tsx
thisisamir98: image
it has more logic now, this looks more readable to me honestly

PR #14129, opened 2022-11-16
runfix: Do not send is typing event on conversation load

thisisamir98:
Do not send is typing event on conversation load


src/script/components/InputBar/InputBar.tsx
```diff
@@ -399,6 +402,13 @@ const InputBar = ({
   }, [isTyping, conversationRepository, conversationEntity]);
 
   useEffect(() => {
+    setIsTyping(undefined);
+  }, [conversationEntity]);
+
+  useEffect(() => {
+    if (isTyping == undefined) {
```
atomrc: This looks suspicious

atomrc (approved):

PR #14130, opened 2022-11-16
feat: Register Self and Team MLS conversations on app load [FS-1007 FS-1006]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1007" title="FS-1007" target="_blank">TaskFS-1007</a>  [web] join the MLS TeamGroup conversation
  </td></table>
  <br />
 

The `self` and `team` conversations have a special case when they are at epoch 0: they should be registered against coreCrypto. 

Because those conversations are created by backend, there are no commit that will be received by the current device, so we need to manually create those commits and send them to backend

### DIFF
--- a/src/script/mls/MLSConversations.test.ts
+++ b/src/script/mls/MLSConversations.test.ts
@@ -89,5 +89,27 @@ describe('MLSConversations', () => {
 
       expect(core.service!.mls.registerConversation).toHaveBeenCalledTimes(2);
     });
+
+    it('does not register self and team conversation that have epoch > 0', async () => {
+      const core = mockCore();
+      const nbProteusConversations = 5 + Math.ceil(Math.random() * 10);
+      const nbMLSConversations = 5 + Math.ceil(Math.random() * 10);
+
+      const proteusConversations = createConversations(nbProteusConversations, ConversationProtocol.PROTEUS);
+      const selfConversation = createConversation(ConversationProtocol.MLS);
+      selfConversation.epoch = 1;
+      selfConversation.type(CONVERSATION_TYPE.SELF);
+
+      const teamConversation = createConversation(ConversationProtocol.MLS);
+      teamConversation.epoch = 2;
+      teamConversation.type(CONVERSATION_TYPE.GLOBAL_TEAM);
+
+      const mlsConversations = createConversations(nbMLSConversations);
+      const conversations = [...proteusConversations, teamConversation, ...mlsConversations, selfConversation];
+
+      await initMLSConversations(conversations, new User(), core, {} as any);
+
+      expect(core.service!.mls.registerConversation).toHaveBeenCalledTimes(0);
+    });
   });
 });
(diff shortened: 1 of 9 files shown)


src/script/mls/MLSConversations.test.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+describe('MLSConversations', () => {
+  beforeEach(() => {
+    jest.spyOn(mlsConversationState.getState(), 'sendExternalToPendingJoin').mockReturnValue(undefined);
+  });
+
+  describe('initMLSConversations', () => {
+    it('joins all the MLS conversations', async () => {
+      const nbProteusConversations = 5 + Math.ceil(Math.random() * 10);
```
przemvs, line 66: Move 5 to variable, to know what it means :D Because idk what this 5 means :/

przemvs: And replace other 5's :D

src/script/mls/MLSConversations.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    it('register all uninitiated conversations', async () => {
+      const core = mockCore();
+      const nbProteusConversations = 5 + Math.ceil(Math.random() * 10);
+      const nbMLSConversations = 5 + Math.ceil(Math.random() * 10);
+
+      const proteusConversations = createConversations(nbProteusConversations, ConversationProtocol.PROTEUS);
+      const selfConversation = createConversation(ConversationProtocol.MLS);
+      selfConversation.epoch = 0;
+      selfConversation.type(CONVERSATION_TYPE.SELF);
```
przemvs: Maybe move this to createConversation? To have it more dynamic ? 🤔

src/script/mls/MLSConversations.ts
```diff
@@ -57,4 +61,52 @@ export async function initMLSConversations(
...
+export async function registerUninitializedConversations(
+  conversations: Conversation[],
+  selfUser: User,
+  selfClientId: string,
+  core: Account,
+): Promise<void> {
+  const uninitializedConversations = conversations.filter(
+    (conversation): conversation is MLSConversation =>
+      isMLSConversation(conversation) &&
+      conversation.epoch === 0 &&
```
przemvs, line 100: Create enum with epoch values to avoid magic numbers :)

przemvs (commented):
Few changes ;)

src/script/mls/MLSConversations.ts
atomrc: I think `0` is explicit enough and we have a special case for accepting 0 as a 'magic' number. 
It's not an enum here, it's just a value that is increased

src/script/mls/MLSConversations.test.ts
atomrc: Since the variable says `nbProteusConversations` I believe it's explicit enough that it's the number of conversations that I want to create. Don't you think?

src/script/mls/MLSConversations.test.ts
atomrc: Done

przemvs (approved):

PR #14137, opened 2022-11-17
runfix: Force React to reset MessageList when conversation are switched

atomrc:
Since the `MessageList` has quite a lot of internal state that is dependent on the current conversation UI state, it's safer, and probably will make things easier to maintain, to ask React to completely re-mount the component from scratch

### DIFF
--- a/src/script/components/Conversation/Conversation.tsx
+++ b/src/script/components/Conversation/Conversation.tsx
@@ -415,6 +415,7 @@ export const Conversation: FC<ConversationProps> = ({
           })}
 
           <MessagesList
+            key={activeConversation.id}
             conversation={activeConversation}
             selfUser={selfUser}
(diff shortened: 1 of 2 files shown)


przemvs (approved):

V-Gira:
@atomrc Copy pasting your comment here for my future reference: 

> Just for the record, and in case those fixes are not so clear:
> 
> React's key is a way to tell react that this component should stay fully mounted until this key hasn't changed. Which means that if the component is still rendered but suddenly the key changes, then React will completely trash the old component and re-renders a new one with a fresh state.
> 
> This makes sense in this context since switching conversation needs to reset a lot of different states (namely the scroll position, for example, or the conversation it was tied to for the file upload)
> 
> without a key then react will keep the same DOM elements for the entire time your <Component> is displayed in the view

PR #14138, opened 2022-11-17
runfix: Open right sidebar on conversation title enter keydown

thisisamir98:
Open right sidebar on conversation title enter keydown


przemvs:
I think we should add to TitleBar - onKeyDown method. 🤔

przemvs:
I spot the problem there:
src/script/page/RightSidebar/PanelHeader/PanelHeader.tsx
image

PR #14155, opened 2022-11-21
feat(entropy): improve entropy estimation (SQSERVICES-1799)

comawill:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1799" title="SQSERVICES-1799" target="_blank">BugSQSERVICES-1799</a>  Calculated Entropy Value may be calculated wrong in crafted cases
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Certain constructed input values to `EntropyData` cause unexpected high entropy estimations.

### Causes (Optional)

Values are evenly distributed, but don't provide actual variance

### Solutions

Add check of 2nd and 3rd derivation of data.

#### Test Coverage (Optional)

- [x] I have added automated test to this contribution

#### How to Test

Automated integration tests, will test this functionality and might fail with the newest changes. An increase of provided mouse movements, will fix this problem.

----

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


src/script/auth/component/EntropyCanvas.tsx
```diff
@@ -98,6 +98,7 @@ const EntropyCanvas = (props: CanvasProps) => {
     const drawPoint: EntropyFrame = {
       x: event.clientX - boundingRect.x,
       y: event.clientY - boundingRect.y,
+      t: 0,
```
tlebon: this seems like an unnecessary addition to satisfy the type. i think maybe we should use a partial instead.

tlebon: or set Date.now() here.

src/script/util/Entropy.test.ts
```diff
@@ -92,5 +131,348 @@ describe('Entropy', () => {
     it('returns zero for empty entropyData', () => {
       expect(new EntropyData().entropyBits).toEqual(0);
     });
+
+    it('makes sense with no movement', () => {
+      const data = new EntropyData();
+      [...Array(300)].forEach((_, i) => {
+        data.addFrame({t: 0, x: 0, y: 0});
```
tlebon: i have a general question about `t` in **all of the subsequent tests**: 

shouldn't it increment sequentially for testing appropriately? 

seems it is throwing off our final entropy values if we set it to be zero when this *should* never happen in a real life use case. 

```suggestion
      [...Array(300)].forEach((_, i) => {
        data.addFrame({t: i, x: 0, y: 0});
```

comawill: The idea here is to provide the worst possible values for `t` and that is the same value for all entries. This then shows that the resulting entropy value is only based on `x` and `y` but not `t`

src/script/auth/component/EntropyCanvas.tsx
comawill: Changed it to a separate type, because `EntropyFrame` is actually not what we have here. What do you think?

tlebon: I personally prefer a partial because it wouldn't add an extra type and it would should a relation between the two types. Either way is fine though

tlebon (approved):
Interesting stuff 👍

PR #14161, opened 2022-11-23
refactor: Improve ConversationState API for selfConversation

atomrc:
This adds 2 new methods to the conversation state that improves how self conversations are accessed

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1965,8 +1965,7 @@ export class ConversationRepository {
       `Handling event '${type}' in conversation '${conversationId.id}/${conversationId.domain}' (Source: ${eventSource})`,
     );
 
-    const selfConversation = this.conversationState.selfConversation();
-    const inSelfConversation = selfConversation && matchQualifiedIds(conversationId, selfConversation.qualifiedId);
+    const inSelfConversation = this.conversationState.isSelfConversation(conversationId);
     if (inSelfConversation) {
       const typesInSelfConversation = [
         CONVERSATION_EVENT.MEMBER_UPDATE,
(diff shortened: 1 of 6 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2589,8 +2588,7 @@ export class ConversationRepository {
     }
 
     const isBackendEvent = eventData.otr_archived_ref || eventData.otr_muted_ref;
-    const selfConversation = this.conversationState.selfConversation();
-    const inSelfConversation = selfConversation && matchQualifiedIds(selfConversation, conversationId);
+    const inSelfConversation = this.conversationState.isSelfConversation(conversationId);
```
arjita-mitra, line 2591: Can we rename it? its a little confusing between inSelfConversation and isSelfConversation

przemvs (approved):

PatrykBuniX (approved):

src/script/conversation/ConversationRepository.ts
```diff
@@ -2722,11 +2720,11 @@ export class ConversationRepository {
    * @returns Resolves when the event was handled
    */
   private async onMessageHidden(eventJson: MessageHiddenEvent) {
-    const {conversation: conversationId, data: eventData, from} = eventJson;
+    const {conversation, qualified_conversation, data: eventData, from} = eventJson;
```
arjita-mitra, line 2723: can we use camelCase here for the qualified_conversation to maintain consistency

V-Gira (approved):

arjita-mitra (changes requested):

src/script/conversation/ConversationRepository.ts
atomrc: We want to know if the message received is, actually, _in_ the self conversation (thus the name of the variable). 
What would you suggest as another name?

src/script/conversation/ConversationRepository.ts
atomrc: This is something that comes from the backend, we cannot really change it. 
If you look at the next line, it's used to create a camelCase variable. Does it seem fine to you?

arjita-mitra: we could do qualified_conversation: qualifiedConversation, but its okay as we are assigning it to a variable afterwards

src/script/conversation/ConversationRepository.ts
arjita-mitra: sounds good then

arjita-mitra (approved):

PR #14165, opened 2022-11-24
fix: Discard typing events after 1 minute

thisisamir98:
User A might start typing and after less than 10 seconds they might go offline for whatever reason hence they won't send any stop typing message anymore and other people in the conversation will end up seeing they are typing forever, this PR fixes this issue by discarding the event after one minute.

### DIFF
--- a/src/script/components/InputBar/TypingIndicator/index.ts
+++ b/src/script/components/InputBar/TypingIndicator/index.ts
@@ -23,4 +23,4 @@ export * from './TypingIndicator';
 
 export {useTypingIndicatorState} from './TypingIndicator.state';
 
-export const IS_TYPING_TIMEOUT = TIME_IN_MILLIS.SECOND * 10;
+export const TYPING_TIMEOUT = TIME_IN_MILLIS.SECOND * 10;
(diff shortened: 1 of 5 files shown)


arjita-mitra (approved):

src/script/components/InputBar/TypingIndicator/index.ts
```diff
@@ -17,6 +17,10 @@
  *
  */
 
+import {TIME_IN_MILLIS} from 'Util/TimeUtil';
+
 export * from './TypingIndicator';
 
 export {useTypingIndicatorState} from './TypingIndicator.state';
+
+export const IS_TYPING_TIMEOUT = TIME_IN_MILLIS.SECOND * 10;
```
atomrc: Why `IS`? I don't understand the naming here? 🤔

thisisamir98: agreed, fixed in https://github.com/wireapp/wire-webapp/pull/14165/commits/f0275432c211fda89aa4ea2af8a62da24db21e34

atomrc (approved):

PR #14182, opened 2022-11-29
runfix: Avoid sending legal hold update to self conversations

atomrc:
Since the `selfMlsConversation` is created by backend, it's considered as a regular conversation by the front-end. 
We need to filter some behavior that should not occur on special conversations such as `self`


tlebon (approved):
🚀

V-Gira (approved):

PR #14183, opened 2022-11-29
feat: add font size selection to the preferences [ACC-302]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-302" title="ACC-302" target="_blank">StoryACC-302</a>  Font Scaling in Preferences (Part of BITV 11.7)
  </td></table>
  <br />


src/script/hooks/useRootFontSize.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function setFontSizeToRoot(currentRootFontSize: RootFontSize) {
+  const root = document.documentElement;
+  root.style.fontSize = currentRootFontSize.toString();
+}
+
+export const useRootFontSize = (
+  rootFontSize: RootFontSize = RootFontSize.M,
+): [RootFontSize, React.Dispatch<React.SetStateAction<RootFontSize>>] => {
+  const [storedRootFontSize, setStoredRootFontSize] = useLocalStorage<RootFontSize>('root-font-size', rootFontSize);
```
przemvs: Move 'root-font-size' key to const. You can export it in this hook. e.g.
export const ROOT_FONT_KEY = '...';

src/script/page/MainContent/panels/preferences/OptionPreferences.tsx
```diff
@@ -161,7 +183,15 @@ const OptionPreferences: React.FC<OptionPreferencesProps> = ({
 
           <hr className="preferences-separator" />
 
-          <PreferencesSection title={t('preferencesOptionsPopular')}>
+          <PreferencesSection title={t('preferencesOptionsAppearance')}>
+            <div css={{marginBottom: '1.5rem', width: '100%'}}>
```
arjita-mitra: why are we using inline CSS?

src/script/page/RightSidebar/GroupParticipantUser/GroupParticipantUser.tsx
```diff
@@ -201,7 +201,7 @@ const GroupParticipantUser: FC<GroupParticipantUserProps> = ({
               </div>
             </div>
 
-            <div className="panel__action-item panel__info-text panel__item-offset" tabIndex={0}>
+            <div className="panel__info-text panel__item-offset" css={{padding: '16px'}} tabIndex={0}>
```
arjita-mitra, line 204: why are we using inline CSS?

src/style/foundation/video-calling.less
```diff
@@ -175,7 +175,7 @@
       left: 50%;
       margin: 0;
       color: var(--main-color);
-      font-size: @font-size-xsmall;
+      font-size: 11px;
```
arjita-mitra, line 178: shouldn't we use em here?

src/script/hooks/useRootFontSize.ts
```diff
@@ -0,0 +1,60 @@
...
+export enum RootFontSize {
+  XXS = '10px',
+  XS = '12px',
+  S = '14px',
+  M = '16px',
+  L = '18px',
+  XL = '20px',
+  XXL = '24px',
+  // XXXL = '30px',
+  // XXXXL = '36px',
```
arjita-mitra: remove commented code

arjita-mitra (changes requested):

src/style/foundation/video-calling.less
V-Gira: We decided to keep a px value for some elements for now, mainly the call ui and the sidebar button

src/script/page/MainContent/panels/preferences/OptionPreferences.tsx
V-Gira: I should probably raise that on the web chat, I'm under the impression we're moving toward inline CSS for oneoff rules like this

src/script/page/AppMain.tsx
```diff
@@ -74,6 +75,19 @@ const AppMain: FC<AppMainProps> = ({
...
+  useEffect(() => {
+    PrimaryModal.init();
+    showInitialModal(userAvailability);
+    // userAvailability not needed for dependency
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  useLayoutEffect(() => {
+    initializeApp();
+  }, []);
```
atomrc: I think this look a little weird, we consume `initializeApp` before it was defined. 
Let's re-move this back to where is was initially

src/script/page/MainContent/panels/preferences/OptionPreferences.tsx
atomrc: I'm personally, not a big fan of inline css. 
I'm not formally against it either, but when possible it's be nice to extract those into a `Component.styles.ts`

PR #14190, opened 2022-11-30
feat: trigger corecrypto migration script [FS-1095]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1095" title="FS-1095" target="_blank">TaskFS-1095</a>  [Web] Trigger CoreCrypto DB migration code when webapp is loaded with CoreCrypto as the proteus handler
  </td></table>
  <br />
 

Trigger `cryptobox` -> `corecrypto` migration script.
- when storage version upgrade is detected (`v19` -> `v20`) store, `"migrationState"` in localstorage.
- after app (core and storage) is initialised, check `"migrationState"` entry.
- if `"migrationState"` was defined, trigger migration script with stored db name
- after migration was successful, delete `"migrationState"` entry from the local storage.

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -402,18 +402,7 @@ export class App {
       }
       const selfUser = await this.initiateSelfUser();
 
-      const dbMigrationState = dbMigrationStateStore.getDBMigrationState();
-      if (dbMigrationState) {
-        const {storeName} = dbMigrationState;
-        try {
-          this.logger.log(`Migrating data from cryptobox store (${storeName}) to corecrypto.`);
-          await this.core.proteusCryptoboxMigrate(storeName);
-          this.logger.log(`Successfully migrated from cryptobox store (${storeName}) to corecrypto.`);
-          dbMigrationStateStore.deleteDBMigrationState();
-        } catch (error) {
-          this.logger.error('Client was not able to perform DB migration:', error);
-        }
-      }
+      await this.triggerDatabaseMigration();
 
       if (this.apiClient.backendFeatures.isFederated && this.repository.storage.storageService.db && context.domain) {
         // Migrate all existing session to fully qualified ids (if need be)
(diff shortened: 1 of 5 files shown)


src/script/main/app.ts
```diff
@@ -399,9 +400,21 @@ export class App {
...
       const selfUser = await this.initiateSelfUser();
 
+      const dbMigrationState = dbMigrationStateStore.getDBMigrationState();
+      if (dbMigrationState) {
+        const {storeName} = dbMigrationState;
+        try {
+          this.logger.log(`Migrating data from cryptobox store (${storeName}) to corecrypto.`);
+          await this.core.proteusCryptoboxMigrate(storeName);
+          this.logger.log(`Successfully migrated from cryptobox store (${storeName}) to corecrypto.`);
+          dbMigrationStateStore.deleteDBMigrationState();
```
atomrc: It would also be nice to delete all the tables that are deprecated (namely `keys`, `prekeys` and `sessions`) now :)

src/script/util/dbMigrationStateStore/dbMigrationStateStore.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const deleteDBMigrationState = () => {
+  localStorage.removeItem(storageKey);
+};
+
+export const dbMigrationStateStore = {
+  deleteDBMigrationState,
+  getDBMigrationState,
+  setDBMigrationState,
+};
```
atomrc: Let's try to come up with a slightly more semantic API. 
Here the `get`, `set` and `delete` are very generic and do not convey so much info. 

How about we try:
- `isCoreDBMigrationNeeded: () => boolean`
- `setNeedsCoreDBMigration: () => void` (no need for inputs, we just flag it)
- `markCoreDBMigrationDone: () => void`

src/script/main/app.ts
PatrykBuniX: Unfortunately we can only clear all the data from these tables (we can't delete whole tables - we would need to modify `schema` which is not allowed when database is running). Not sure what to do about it 🤔

atomrc: oohh I see, then maybe we will need to do it in a later `version` transaction.

atomrc: Here is the strategy I'd suggest:
- let's, for now, clear the content of the tables
- let's prepare a further migration script that will delete the tables **if** they are empty :)

atomrc (approved):
Seems solid 💪  nice one man

atomrc (approved):

PR #14234, opened 2022-11-30
feat: Handle reset session messages

atomrc:


src/script/cryptography/CryptographyMapper.ts
```diff
@@ -242,7 +242,7 @@ export class CryptographyMapper {
       default: {
         const logMessage = `Skipped event '${genericMessage.messageId}' of unhandled type '${genericMessage.content}'`;
         this.logger.debug(logMessage, {event, generic_message: genericMessage});
-        throw new CryptographyError(CryptographyError.TYPE.UNHANDLED_TYPE, logMessage);
+        return undefined;
```
aweiss-dev, line 245: the 'return undefined' can be removed. It should return undefined by default with no return statement

aweiss-dev (approved):

PatrykBuniX (approved):

PR #14236, opened 2022-12-01
refactor: Cleanup usage of core

atomrc:
Prepare for https://github.com/wireapp/wire-web-packages/pull/4650


src/script/client/ClientEntity.ts
```diff
@@ -75,17 +75,6 @@ export class ClientEntity {
...
-  /**
-   * Splits an ID into user ID, client ID & domain (if any).
-   */
-  static dismantleUserClientId(id: string): {clientId: string; domain?: string; userId: string} {
-    // see https://regex101.com/r/c8FtCw/1
-    const regex = /((?<domain>.+)@)?(?<userId>.+)@(?<clientId>.+)$/g;
-    const match = regex.exec(id);
-    const {domain, userId, clientId} = match?.groups || {};
-    return {clientId, domain, userId};
-  }
```
atomrc, line 87: This has been moved to a util module

src/__mocks__/@wireapp/core.ts
```diff
@@ -0,0 +1,40 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export class Account {
```
przemvs, line 20: Maybe use name for mocks /Class/Mock or Mocked/Class/? WDYT?

atomrc: It needs to be the same name as the class exported in the real package, actually.

przemvs: Or it should have same name as original class ?

przemvs (approved):

PR #14253, opened 2022-12-05
feat: api v3 access_role change [FS-1192]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1192" title="FS-1192" target="_blank">TaskFS-1192</a>  [Web] Parsing of conversation.access_update for API V3
  </td></table>
  <br />
 

Apply api v3 changes of `access_role` field(s).

Before API v3 behaviour:
fields:
- `access_role` (string value) - deprecated field 
- `access_role_v2` (optional array of string values)

Check for `access_role_v2` field, if exists -> parse its values, if it's not defined, parse `access_role` value.

since v3 behaviour:
fields:
- `access_role` (array of string values) - contains the values stored previously in `access_role_v2`
- `access_role_v2` (optional value) - contains the same value as `access_role`

Check for `access_role_v2` field, if exists -> parse its values. If it's not defined, check if `access_role` is array type, if it is it mean we're dealing with new version of api and we can parse it the same way as `access_role_v2`. If it's a string type, parse it "the old way" (legacy).

Needs https://github.com/wireapp/wire-web-packages/pull/4682

### DIFF
--- a/src/script/conversation/ConversationMapper.test.ts
+++ b/src/script/conversation/ConversationMapper.test.ts
@@ -18,10 +18,10 @@
  */
 
 import {
-  ACCESS_ROLE_V2,
+  CONVERSATION_ACCESS_ROLE,
   Conversation as ConversationBackendData,
   CONVERSATION_ACCESS,
-  CONVERSATION_ACCESS_ROLE,
+  CONVERSATION_LEGACY_ACCESS_ROLE,
   CONVERSATION_TYPE,
   Member as MemberBackendData,
   OtherMember as OtherMemberBackendData,
@@ -345,7 +345,7 @@ describe('ConversationMapper', () => {
 
       const remoteData: Partial<ConversationDatabaseData> = {
         access: [CONVERSATION_ACCESS.INVITE, CONVERSATION_ACCESS.CODE],
-        access_role: CONVERSATION_ACCESS_ROLE.NON_ACTIVATED,
+        access_role: CONVERSATION_LEGACY_ACCESS_ROLE.NON_ACTIVATED,
         creator: conversationCreatorId,
         id: conversationId,
         last_event_timestamp: new Date('1970-01-01T00:00:00.000Z').getTime(),
@@ -631,14 +631,18 @@ describe('ConversationMapper', () => {
       const accessModes = [CONVERSATION_ACCESS.INVITE, CONVERSATION_ACCESS.CODE];
       // ACCESS_STATE.TEAM.GUESTS_SERVICES
       const accessRole = [
-        ACCESS_ROLE_V2.GUEST,
-        ACCESS_ROLE_V2.NON_TEAM_MEMBER,
-        ACCESS_ROLE_V2.TEAM_MEMBER,
-        ACCESS_ROLE_V2.SERVICE,
+        CONVERSATION_ACCESS_ROLE.GUEST,
+        CONVERSATION_ACCESS_ROLE.NON_TEAM_MEMBER,
+        CONVERSATION_ACCESS_ROLE.TEAM_MEMBER,
+        CONVERSATION_ACCESS_ROLE.SERVICE,
       ];
 
       // ACCESS_STATE.TEAM.GUEST_ROOM
-      const accessRoleV2 = [ACCESS_ROLE_V2.GUEST, ACCESS_ROLE_V2.NON_TEAM_MEMBER, ACCESS_ROLE_V2.TEAM_MEMBER];
+      const accessRoleV2 = [
+        CONVERSATION_ACCESS_ROLE.GUEST,
+        CONVERSATION_ACCESS_ROLE.NON_TEAM_MEMBER,
+        CONVERSATION_ACCESS_ROLE.TEAM_MEMBER,
+      ];
 
       const conversationEntity = new Conversation('conversation-id', 'domain');
       conversationEntity.team_id = 'team_id';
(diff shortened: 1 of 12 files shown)


src/script/conversation/ConversationMapper.ts
```diff
@@ -489,4 +469,48 @@ export class ConversationMapper {
       : ACCESS_STATE.PERSONAL.ONE2ONE;
     return conversationEntity.accessState(personalAccessState);
   }
+
+  private static mapAccessStateV2(conversationEntity: Conversation, accessRole: ACCESS_ROLE_V2[]) {
```
atomrc: Maybe those 2 functions `mapAccessStaetV2` and `mapLegacyAccessState` could be pure and just return an `ACCESS_STATE`.

Then we would have a single call to `conversation.accessState(newAccessState)` in the parent function? 
You would also not need to inject `conversationEntity` into those functions. WDYT?

PatrykBuniX: Makes sense :D

src/script/conversation/ConversationMapper.test.ts
```diff
@@ -621,4 +625,170 @@ describe('ConversationMapper', () => {
...
+  describe('mapAccessState', () => {
+    it('maps "access_state_v2" first (if exists)', () => {
+      const accessModes = [CONVERSATION_ACCESS.INVITE, CONVERSATION_ACCESS.CODE];
+      // ACCESS_STATE.TEAM.GUESTS_SERVICES
+      const accessRole = [
+        ACCESS_ROLE_V2.GUEST,
+        ACCESS_ROLE_V2.NON_TEAM_MEMBER,
+        ACCESS_ROLE_V2.TEAM_MEMBER,
+        ACCESS_ROLE_V2.SERVICE,
+      ];
```
przemvs: I think you can move it above functions, and you using it in multiple places :) And I think this name should be smth like teamGuestServicesAccessRole

atomrc (approved):
Seem solid, good job man 🚀

src/script/conversation/ConversationMapper.ts
```diff
@@ -69,7 +69,8 @@ export interface SelfStatusUpdateDatabaseData {
 export type ConversationDatabaseData = ConversationRecord &
   Partial<ConversationBackendData> & {
     accessModes?: CONVERSATION_ACCESS[];
-    accessRole?: CONVERSATION_ACCESS_ROLE;
+    //CONVERSATION_ACCESS_ROLE for api <= v2, ACCESS_ROLE_V2[] since api v3
+    accessRole?: CONVERSATION_ACCESS_ROLE | ACCESS_ROLE_V2[];
```
przemvs: I think it can like that:
```suggestion
    accessRole?: CONVERSATION_ACCESS_ROLE | ACCESS_ROLE[];
```

src/script/conversation/ConversationMapper.ts
```diff
@@ -435,49 +437,34 @@ export class ConversationMapper {
   static mapAccessState(
     conversationEntity: Conversation,
     accessModes: CONVERSATION_ACCESS[],
-    accessRole?: CONVERSATION_ACCESS_ROLE,
-    accessRoleV2: ACCESS_ROLE_V2[] = [],
+    accessRole: CONVERSATION_ACCESS_ROLE | ACCESS_ROLE_V2[],
+    accessRoleV2?: ACCESS_ROLE_V2[],
```
przemvs: WDYT ?
```suggestion
    accessRole: CONVERSATION_ACCESS_ROLE | ACCESS_ROLE[],
    accessRoleV2?: ACCESS_ROLE[],
```

src/script/conversation/ConversationMapper.test.ts
PatrykBuniX: This way it's more readable and easier to see what case we're actually testing

src/script/conversation/ConversationMapper.ts
```diff
@@ -487,6 +474,51 @@ export class ConversationMapper {
...
+    const includesInviteMode = accessModes.includes(CONVERSATION_ACCESS.INVITE);
+    const isInviteModeOnly = includesInviteMode && accessModes.length === 1;
+
+    const isTeamOnlyMode = isTeamRole && isInviteModeOnly;
+    if (isTeamOnlyMode) {
+      return ACCESS_STATE.TEAM.TEAM_ONLY;
+    }
+
+    const isVerifiedRole = accessRole === CONVERSATION_ACCESS_ROLE.ACTIVATED;
+    if (isVerifiedRole) {
```
przemvs: isActivated ? 🤔

src/script/conversation/ConversationMapper.ts
```diff
@@ -487,6 +474,51 @@ export class ConversationMapper {
...
+      return ACCESS_STATE.TEAM.GUEST_ROOM;
+    }
+    const isNonVerifiedRole = accessRole === CONVERSATION_ACCESS_ROLE.NON_ACTIVATED;
```
przemvs: isInactivated maybe should better variable name? :D

src/script/conversation/ConversationMapper.ts
przemvs: Also you can pass this directly, because you using it 1 time ;)

src/script/conversation/ConversationMapper.ts
```diff
@@ -487,6 +474,51 @@ export class ConversationMapper {
...
+    accessModes: CONVERSATION_ACCESS[],
+    accessRole: CONVERSATION_ACCESS_ROLE,
+  ): ACCESS_STATE {
+    const isTeamRole = accessRole === CONVERSATION_ACCESS_ROLE.TEAM;
+
+    const includesInviteMode = accessModes.includes(CONVERSATION_ACCESS.INVITE);
+    const isInviteModeOnly = includesInviteMode && accessModes.length === 1;
+
+    const isTeamOnlyMode = isTeamRole && isInviteModeOnly;
+    if (isTeamOnlyMode) {
```
przemvs, line 510:
```suggestion
    if (isTeamRole && isInviteModeOnly) {
```

atomrc (approved):

przemvs (approved):

PR #14256, opened 2022-12-05
fix: reloading an active conversation is jumping onto wrong conversation(ACC-268)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-268" title="ACC-268" target="_blank">SubtaskACC-268</a>  9.2.4.3 Focus order for chat content area
  </td></table>
  <br />
 

----

- The **PR Title**
  - reloading an active conversation is jumping onto wrong conversation and the conversation id in the url is wrong
  
- The **PR Description**
  -  reloading an active conversation is jumping onto wrong conversation and the conversation id in the url is wrong
  - this is caused by a resetting history param logic which is suppose to reset the last message focus on app load
----

# What's new in this PR?

### Issues

- reloading an active conversation is jumping onto wrong conversation and the conversation id in the url is wrong

### Solutions

- instead of running the history state reset for every conversation, run it on app initialisation only once

#### How to Test

- open a conversation, now reload the app. The conversation id in the URL should't change.

### DIFF
--- a/src/script/page/AppMain.tsx
+++ b/src/script/page/AppMain.tsx
@@ -145,7 +145,7 @@ const AppMain: FC<AppMainProps> = ({
       }
     };
 
-    // on conversation/app load reset last message focus to ensure last message is focused
+    // on app load reset last message focus to ensure last message is focused
     // only when user enters a new conversation using keyboard(press enter)
     const historyState = window.history.state;
     if (historyState && !!historyState.eventKey) {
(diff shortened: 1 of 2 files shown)


src/script/page/AppMain.tsx
```diff
@@ -145,6 +145,14 @@ const AppMain: FC<AppMainProps> = ({
       }
     };
 
+    // on conversation/app load reset last message focus to ensure last message is focused
+    // only when user enters a new conversation using keyboard(press enter)
```
atomrc:
```suggestion
    // on app load reset last message focus to ensure last message is focused
    // only when user enters a new conversation using keyboard(press enter)
```

atomrc (approved):

atomrc (approved):

PR #14265, opened 2022-12-07
fix: click on emoji closes edit or reply

aweiss-dev:
Clicking outside the text area while editing something or writing a reply will close the current draft and revert the input back to the default input.

Since the emoji picker was outside the text area, clicking on an emoji closed the current edit/response draft.

This got fixed by adding the emoji container to the "outside textarea click - ignore list".

While doing this, I recognized that the ignore list had several other components' class names hardcoded. This is highly dangerous since that functionality will break if the class names of the other components get changed. This lead me to do some small refactoring to class names for that specific cases.

In the future, we should avoid using hard-coded class names and instead rely on variables. This still is dangerous since our less files which hold the same identifiers need to have them hard coded, but with further migration to CSS modules or other CSS-in-JS solutions, this can be solved.


src/script/components/InputBar/InputBar.tsx
```diff
@@ -898,8 +901,8 @@ const InputBar = ({
 
   return (
     <div
-      id="conversation-input-bar"
-      className={cx('conversation-input-bar', {'is-right-panel-open': isRightSidebarOpen})}
+      id={conversationInputBarClassName}
+      className={cx(`${conversationInputBarClassName}`, {'is-right-panel-open': isRightSidebarOpen})}
```
przemvs: I think it can be like that :)

```suggestion
      className={cx(conversationInputBarClassName, {'is-right-panel-open': isRightSidebarOpen})}
```

aweiss-dev: 👍

atomrc:
I think we should try the other way around and a a `data-*` for on the elements that need to block the outsideclick and avoid exporting classnames. 
What do you think?

aweiss-dev:
> I think we should try the other way around and a a `data-*` for on the elements that need to block the outsideclick and avoid exporting classnames. What do you think?

I changed this now, and I like it better already. 
I am a bit sad that we don't use styled-components because it is easier to implement generic wrappers through the ```as prop``` handling it has included (https://emotion.sh/docs/styled#as-prop). 
In this case the wrapper is not generic and just implemented as a div.

src/script/components/InputBar/util/clickHandlers.tsx
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import {HTMLProps, ReactNode} from 'react';
+
+interface IgnoreClickWrapperProps extends HTMLProps<HTMLDivElement> {
+  children: ReactNode;
+}
+const IgnoreClickWrapper = ({children, ...rest}: IgnoreClickWrapperProps) => {
+  return (
+    <div {...rest} data-ignore-click>
```
atomrc: Let's make it obvious that it's related to the outside click handler.

```suggestion
    <div {...rest} data-outside-click-ignore>
```

src/script/components/InputBar/util/clickHandlers.tsx
```diff
@@ -0,0 +1,41 @@
...
+      {children}
+    </div>
+  );
+};
+
+const handleClickOutsideOfInputBar = (event: Event, callback: () => void): void => {
+  const ignoredParent = (event.target as HTMLElement).closest('div[data-ignore-click]') !== null;
```
atomrc:
```suggestion
  const ignoredParent = (event.target as HTMLElement).closest('div[data-outside-click-ignore]') !== null;
```

src/script/components/InputBar/util/clickHandlers.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+const IgnoreClickWrapper = ({children, ...rest}: IgnoreClickWrapperProps) => {
+  return (
+    <div {...rest} data-ignore-click>
+      {children}
+    </div>
+  );
+};
+
+const handleClickOutsideOfInputBar = (event: Event, callback: () => void): void => {
```
atomrc, line 35:
```suggestion
const handleOutsideClick = (event: Event, callback: () => void): void => {
```

I don't think we need to mention the input bar at this point. This handler is more generic than this (and could potentially be used in many places)

src/script/components/InputBar/util/clickHandlers.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {HTMLProps, ReactNode} from 'react';
+
+interface IgnoreClickWrapperProps extends HTMLProps<HTMLDivElement> {
+  children: ReactNode;
+}
+const IgnoreClickWrapper = ({children, ...rest}: IgnoreClickWrapperProps) => {
```
atomrc:
```suggestion
const IgnoreOutsideClickWrapper = ({children, ...rest}: IgnoreClickWrapperProps) => {
```

atomrc (approved):

src/script/ui/ContextMenu.tsx
```diff
@@ -146,48 +154,60 @@ const ContextMenu: React.FC<ContextMenuProps> = ({entries, defaultIdentifier = '
...
+                  : {
+                      onClick: event => {
+                        event.preventDefault();
+                        cleanUp();
+                        entry.click?.(event.nativeEvent);
+                      },
+                      onMouseEnter: () => {
+                        setSelected(undefined);
+                      },
+                    })}
```
przemvs, line 195: Hmm..

Why not: 
`{...(entry.isDisabled && ({ props })}`

PatrykBuniX (approved):

src/script/ui/ContextMenu.tsx
aweiss-dev: I'm not too fond of this approach. Hiding whats going on behind the value of a prop has no benefit in my eyes, its more complex to read and find out whats going on.
Doing important stuff visible and spreading the ...rest we don't care about, on the other hand, is ok.

PR #14270, opened 2022-12-08
refactor: Cleanup file dropping/pasting in InputBar

atomrc:


src/script/components/InputBar/InputBar.tsx
```diff
@@ -65,6 +52,11 @@ import {formatLocale, TIME_IN_MILLIS} from 'Util/TimeUtil';
 import {formatBytes, getSelectionPosition} from 'Util/util';
 
 import {getRichTextInput} from './getRichTextInput';
+import {useDropFiles} from './hooks/useDropFiles';
+import {useFilePaste} from './hooks/useFilePaste';
+import {useResizeTarget} from './hooks/useResizeTarget';
+import {useScrollSync} from './hooks/useScrollSync';
+import {useTextAreaFocus} from './hooks/useTextAreaFocus';
```
atomrc, line 59: Since those hooks are only used here, they can be put right next to the input component

PatrykBuniX (approved):
Looks good

src/script/components/InputBar/hooks/useFilePaste.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (event.clipboardData?.types.includes('text/plain')) {
+        return;
+      }
+      // Avoid copying the filename into the input field
+      event.preventDefault();
+      const files = event.clipboardData?.files;
+      if (files) {
+        onFilePasted(files);
+      }
+    };
```
przemvs, line 34: WDYT to move it outside useEffect? 🤔

atomrc: This was to avoid re-creating it on each render (and not needing to put it in the dependencies)

I could have used a use callback as well, but it's just more code and duplicated dependencies

PR #14276, opened 2022-12-08
refactor: file upload

PatrykBuniX:
Refactor of file upload code in `Conversation` and `InputBar` components.


src/script/page/RootProvider.tsx
```diff
@@ -28,8 +28,15 @@ interface RootProviderProps {
...
 
-export {RootProvider};
+export const useMainViewModel = () => {
+  const context = useContext(RootContext);
+  if (!context) {
+    throw new Error('MainViewModel was not initialised');
+  }
+
+  return context;
+};
```
PatrykBuniX, line 42: We're always passing `MainViewModel` to `RootProvider` so this should be totally fine and typesafe (no need to check for possible null inside `Conversation` component)

src/script/components/InputBar/hooks/useDropFiles.ts
```diff
@@ -40,16 +45,14 @@ export const useDropFiles = (selector: string, onFileDropped: (files: File[]) =>
...
-    }
+    return () => {
+      elementRef.removeEventListener('drop', handleDrop);
+      elementRef.removeEventListener('dragover', onDragOver);
+    };
+  }, [elementRef, onFileDropped]);
 
-    return () => undefined;
-  }, [onFileDropped, selector]);
+  return {ref: setElementRef};
```
PatrykBuniX, line 57: We return ready to use ref setter - no need to pass any info about element we want to use it on.

PatrykBuniX: inspiration: https://usehooks-ts.com/react-hook/use-element-size

src/script/components/InputBar/InputBar.tsx
```diff
@@ -663,25 +576,14 @@ const InputBar = ({
     }
   };
 
-  /**
-   * higher order function to check if file sharing is enabled.
-   * If not enabled, it will show a warning modal else will return the given callback
-   *
-   * @param callback - function to be called if file sharing is enabled
-   */
-  function checkFileSharingPermission<T extends (...args: any[]) => void>(callback: T): T | (() => void) {
```
PatrykBuniX, line 675: I don't think this util makes sense now since we've moved a function that wrapped with it to Conversation component. We'd need to pass this function down here, what does not make sense since we got access to `isFileSharingSendingEnabled`. WDYT?

atomrc: humm the point of this was to deduplicate this condition and showing the error. Now it's duplicated again sadly (you have the same condition in the `Conversation` component)

src/script/components/InputBar/hooks/useDropFiles.ts
atomrc: dddaaammmnnn, this is good api. I think we could use that same idea for fadingscrollbar as well 😱

przemvs (approved):

src/script/components/Conversation/utils/checkFileSharingPermission.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * higher order function to check if file sharing is enabled.
+ * If not enabled, it will show a warning modal else will return the given callback
+ *
+ * @param callback - function to be called if file sharing is enabled
+ */
+
+export const checkFileSharingPermission = <T extends (...args: any[]) => void>(callback: T): T | (() => void) => {
+  const teamState = container.resolve(TeamState);
```
PatrykBuniX, line 34: since it's a singleton I think it can live in separate file

src/script/components/Conversation/Conversation.tsx
```diff
@@ -119,12 +135,104 @@ export const Conversation: FC<ConversationProps> = ({
     }
   }, [readMessagesBuffer.length]);
 
-  if (!mainViewModel) {
-    return null;
-  }
+  const uploadImages = useCallback(
+    (images: File[]) => {
+      if (!activeConversation) {
```
arjita-mitra: can we combine them in one if with || ?

arjita-mitra: if (!activeConversation || isHittingUploadLimit(images, repositories.asset)) {
    return;
}

V-Gira (approved):

src/script/components/Conversation/Conversation.tsx
```diff
@@ -119,12 +135,104 @@ export const Conversation: FC<ConversationProps> = ({
...
+      }
 
-  const {content: contentViewModel} = mainViewModel;
-  const {conversationRepository, repositories} = contentViewModel;
+      for (const image of Array.from(images)) {
+        const isImageTooLarge = image.size > CONFIG.MAXIMUM_IMAGE_FILE_SIZE;
+
+        if (isImageTooLarge) {
+          const isGif = image.type === 'image/gif';
+          const maxSize = CONFIG.MAXIMUM_IMAGE_FILE_SIZE / 1024 / 1024;
```
arjita-mitra: you can add a const for 1024

src/script/components/Conversation/Conversation.tsx
```diff
@@ -119,12 +135,104 @@ export const Conversation: FC<ConversationProps> = ({
...
+
+          showWarningModal(
+            t(isGif ? 'modalGifTooLargeHeadline' : 'modalPictureTooLargeHeadline'),
+            t(isGif ? 'modalGifTooLargeMessage' : 'modalPictureTooLargeMessage', maxSize),
+          );
+
```
arjita-mitra: do we need a new line here?

src/script/components/Conversation/Conversation.tsx
```diff
@@ -119,12 +135,104 @@ export const Conversation: FC<ConversationProps> = ({
...
+          return;
+        }
+      }
+
+      repositories.message.uploadImages(activeConversation, images);
```
arjita-mitra, line 159: do we need a new line here too? it looks too spacious to me.

src/script/components/Conversation/utils/checkFileSharingPermission.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (teamState.isFileSharingSendingEnabled()) {
+    return callback;
```
przemvs, line 37: This callback works ? Without running it ? 🤔

PatrykBuniX: Yeah, this is the assumption of HOF - Higher Order Function - this is a function that returns other function (we can call or pass it on later) 😄

arjita-mitra (changes requested):

src/script/components/Conversation/utils/checkFileSharingPermission.ts
przemvs: Ahh.. you're right :D

src/script/components/Conversation/Conversation.tsx
PatrykBuniX: sure 😄

src/script/components/Conversation/Conversation.tsx
PatrykBuniX: I find it more readable, we know its separate functionality

arjita-mitra (approved):

PR #14302, opened 2022-12-12
refactor: use disposable ref hook

PatrykBuniX:
Removes `useDisposableRef` hook completely.


src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -164,6 +167,7 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
           containerRef.current = element;
         }
       }}
+      key={`${editedTimestamp}:${text}`}
```
PatrykBuniX, line 162: This is important: `key` prop change => react knows it's dealing with new element, so the ref changes and it can recalculate message dimentions. (no need to add extra dependencies to deps array)

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -70,9 +71,11 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
       setCanShowMore?.(isWider || isHigher);
       return () => {};
```
atomrc: you probably don't need this return anymore :)

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
atomrc: Brilliant 👌

src/script/components/MessagesList/Message/MemberMessage.tsx
```diff
@@ -82,7 +82,7 @@ const MemberMessage: React.FC<MemberMessageProps> = ({
   const isMemberLeave = message.isMemberLeave();
   const isMemberChange = message.isMemberChange();
 
-  const initShowMore = useDisposableRef(element => {
+  const {ref: initShowMore} = useDisposableRef(element => {
```
atomrc: Actually, I'm wondering if we should not get rid of the disposableRef at all and just add a `click` listener on the parent React element and only react to the event if the target has the right class. WDYT?

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
PatrykBuniX: From now we'll also store element's reference in state (it changes only with `key` attribute change). Adding reference to deps array now makes it do it's all recalculations whenever element updates (whenever it's edited). WDYT?

atomrc: makes a lot of sense 👌

atomrc (approved):

PR #14306, opened 2022-12-12
refactor: Use delegate event handlers for message clicks

atomrc:
Instead of adding (and removing) event handlers directly on the message elements that need handling, this PR will register a global event handler that will detect on which type of element we clicked. 

This way, we let React take care of all the `(add|remove)EventListener` calls and we can focus on the business logic

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.test.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.test.tsx
@@ -103,10 +103,10 @@ describe('TextMessageRenderer', () => {
     expect(linkElem).not.toBe(null);
 
     fireEvent.click(linkElem);
-    expect(onClickElement).toHaveBeenCalled();
+    expect(onClickElement).toHaveBeenCalledTimes(1);
 
-    fireEvent.keyDown(linkElem);
-    expect(onClickElement).toHaveBeenCalled();
+    fireEvent.keyDown(linkElem, {key: 'Enter'});
+    expect(onClickElement).toHaveBeenCalledTimes(2);
   });
 
   it('should not trigger a key event if the message is not focused', () => {
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
@@ -20,6 +20,8 @@
 import {useEffect, FC, useState} from 'react';
 
 import {Text} from 'src/script/entity/message/Text';
+import {isKeyDownEvent} from 'src/script/guards/Event';
+import {isMouseEvent} from 'src/script/guards/Mouse';
 import {getAllFocusableElements, setElementsTabIndex} from 'Util/focusUtil';
 import {handleKeyDown} from 'Util/KeyboardUtil';
 
@@ -68,12 +70,11 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
   }, [isQuoteMsg, setCanShowMore, containerRef]);
 
   useEffect(() => {
-    const element = containerRef;
-    if (!element) {
+    if (!containerRef) {
       return;
     }
 
-    const interactiveMsgElements = getAllFocusableElements(element);
+    const interactiveMsgElements = getAllFocusableElements(containerRef);
     setElementsTabIndex(interactiveMsgElements, isCurrentConversationFocused);
   }, [isCurrentConversationFocused, containerRef]);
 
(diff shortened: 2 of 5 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -76,90 +70,64 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
...
+    // We will register the click event on the paragraph element and determine the type of the element clicked.
+    // This is because the paragraph element is fed with raw HTML and we cannot register the click event on the clickabled elements directly.
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
     <p
       ref={setContainerRef}
       key={`${editedTimestamp}:${text}`}
+      onClick={handleClick}
+      onAuxClick={handleClick}
+      onKeyDown={handleClick}
+      onKeyUp={handleClick}
```
PatrykBuniX: I think we should come up with some better naming for `handleClick` since it handles more events that just `click`. What about `handleInteraction` or sth like that? 😄

atomrc: Good call 👌

PatrykBuniX (approved):
Great! It simplifies things a lot! 🚀

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -76,90 +70,68 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
...
+    event: KeyboardEvent | MouseEvent,
+    elementType: ElementType,
+    messageDetails: MessageDetails,
+  ) => {
+    if (event.type === 'keydown' && isCurrentConversationFocused) {
+      handleKeyDown(event as KeyboardEvent, () => {
+        event.preventDefault();
+        onMessageClick(event, elementType, messageDetails);
       });
+    } else if (event.type === 'click') {
```
przemvs: We have isMousEvent TypeGuard ;)

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -76,90 +70,68 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
...
-      elements?.forEach(element => {
-        events.forEach(eventName => {
-          element.addEventListener(eventName, handleMsgEvent);
-        });
+  const forwardEvent = (
+    event: KeyboardEvent | MouseEvent,
+    elementType: ElementType,
+    messageDetails: MessageDetails,
+  ) => {
+    if (event.type === 'keydown' && isCurrentConversationFocused) {
```
przemvs: Add isKeyboardEvent TypeGuard maybe, then you will pass event to handleKeyDown without as KeyboardEvent :)

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
atomrc: Oh yeah, nice one

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
atomrc: Good call 👍

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
```diff
@@ -74,92 +70,69 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
   }, [isQuoteMsg, setCanShowMore, containerRef]);
 
   useEffect(() => {
-    const element = containerRef;
-    if (!element) {
-      return undefined;
+    if (!containerRef) {
+      return;
```
arjita-mitra, line 74: is this working? I was getting some error, no implicit return.

przemvs (approved):

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.test.tsx
```diff
@@ -103,10 +103,10 @@ describe('TextMessageRenderer', () => {
     expect(linkElem).not.toBe(null);
 
     fireEvent.click(linkElem);
-    expect(onClickElement).toHaveBeenCalled();
+    expect(onClickElement).toHaveBeenCalledTimes(1);
 
-    fireEvent.keyDown(linkElem);
-    expect(onClickElement).toHaveBeenCalled();
+    fireEvent.keyDown(linkElem, {key: 'Enter'});
+    expect(onClickElement).toHaveBeenCalledTimes(2);
```
arjita-mitra, line 109: why expected to called twice? could you please explain?

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.tsx
atomrc: Previously, you were returning an arrow function to get rid of the event listeners. But now that this is gone, this function returns nothing and typescript is happy because this function is never returning anything :)

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer.test.tsx
PatrykBuniX: He calls it twice in this test: once with click event, 2nd time with keydown (he does not reset the number of calls between these events) 😉

atomrc: So, yeah, there was one gap with this test. Here it simulates a `click` **and**  a keyboard event. 
So the callback should have been called twice (one for the click, one for the `enter` keydown)

arjita-mitra (approved):

PR #14307, opened 2022-12-12
refactor: Improve quote long text detection

atomrc:
A few improvements in this PR:
- long text detection is now completely isolated inside the TextRenderer components (based on a new `collapse` props)
- the `ShowMoreButton` is also isolated in the `TextRender` component and not visible to the `Quote` component (it just passes a `collapse` props to it)
- limits re-renders of the `Quote` component (as the children is now responsible for collapsing)

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer/TextMessageRenderer.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer/TextMessageRenderer.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {useEffect, FC, useState, HTMLProps} from 'react';
+import {useEffect, FC, useState, HTMLProps, useRef} from 'react';
 
 import {isKeyDownEvent} from 'src/script/guards/Event';
 import {isMouseEvent} from 'src/script/guards/Mouse';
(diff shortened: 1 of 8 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer/TextMessageRenderer.tsx
```diff
@@ -121,22 +121,36 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
...
   };
+  const extraClasses = showFullText ? 'message-quote__text--full' : '';
 
   return (
-    // We will register the click event on the paragraph element and determine the type of the element clicked.
-    // This is because the paragraph element is fed with raw HTML and we cannot register the click event on the clickabled elements directly.
-    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
-    <p
-      ref={setContainerRef}
-      key={`${editedTimestamp}:${text}`}
```
atomrc, line 131: @PatrykBuniX I removed the need for a `timestamp` here (since we want the ref to be updated when the text changes, so only a hash of the text is enough for us :) )

src/script/components/MessagesList/Message/ContentMessage/asset/TextMessageRenderer/TextMessageRenderer.tsx
```diff
@@ -121,22 +119,36 @@ export const TextMessageRenderer: FC<TextMessageRendererProps> = ({
...
+        onKeyDown={handleInteraction}
+        onKeyUp={handleInteraction}
+        dangerouslySetInnerHTML={{__html: text}}
+        dir="auto"
+        className={`${className} ${extraClasses}`}
+        {...props}
+      />
+      {canShowMore && (
+        <ShowMoreButton
+          onClick={() => setShowFullText(!showFullText)}
```
PatrykBuniX: I'd rather move this function above render and use previous value when toggling state (state setters are not synchronous operations and might behave unexpectedly 😅 ):
```ts
const toggleShowFullText = () => {
  setShowFullText(prev => !prev)
}
```

PatrykBuniX (approved):
Great improvements! 🚀

PR #14312, opened 2022-12-13
fix: Open sidebar when pressing enter on conversation title

przemvs:
@thisisamir98 please update description :)


thisisamir98:
When focus is on the title of the conversation and `Enter` key is pressed the right side bar should be opened and focus should be shifted to the first element of the right side bar, focus shifting was working fine but `Enter` key behavior was broken and it only worked with mouse click, this PR fixes the issue by adding a `tabIndex` to the conversation details.

PatrykBuniX (approved):

PR #14313, opened 2022-12-13
feat: Update HTML tags in Modals, AppLock and DeviceID

przemvs:
Replace HTML Tags to 'p' Tag in Modals, AppLock.
Replace HTML Tags to 'strong' in DeviceID.

Updated classNames for better readability :)

### DIFF
--- a/src/script/page/MainContent/panels/preferences/devices/components/FormattedId.tsx
+++ b/src/script/page/MainContent/panels/preferences/devices/components/FormattedId.tsx
@@ -24,7 +24,7 @@ interface FormattedIdProps {
 export const FormattedId = ({idSlices}: FormattedIdProps) => (
   <>
     {idSlices.map((slice, index) => {
-      const Component = index % 2 === 0 ? 'span' : 'strong';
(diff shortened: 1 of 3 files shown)


src/script/page/MainContent/panels/preferences/devices/components/FormattedId.tsx
```diff
@@ -17,16 +17,20 @@
...
-          {index % 2 === 0 ? <strong>{slice}</strong> : slice}
-        </span>
-      ))}
-    </>
-  );
-};
+export const FormattedId = ({idSlices}: FormattedIdProps) => (
+  <>
+    {idSlices.map((slice, index) => {
+      const Component = index % 2 === 0 ? 'span' : 'strong';
```
PatrykBuniX: Shouldn't `<strong>` go first (as it was before)? 🤔

przemvs: Sure, good catch! :)

PatrykBuniX (approved):

PR #14317, opened 2022-12-14
refactor: create drop file area component

PatrykBuniX:
Removes `useDropFile` hook and replaces it with `DropFileArea` component - no need of any state (or ref) management. We're using native react's event handlers that are being (un)subscribed automatically.
- much simpler reusable api
- no extra rerendering (no state)

### DIFF
--- a/src/script/components/DropFileArea/DropFileArea.test.tsx
+++ b/src/script/components/DropFileArea/DropFileArea.test.tsx
@@ -1,0 +1,79 @@
+/*
+ * Wire
+ * Copyright (C) 2022 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/script/components/DropFileArea/DropFileArea.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+ *
+ */
+
+import {fireEvent, render} from '@testing-library/react';
+
+import {DropFileArea} from './DropFileArea';
+
+const getDefaultProps = () => ({
+  onFileDropped: jest.fn(),
+});
```
przemvs, line 26: You need this as function? 🤔

PatrykBuniX: I want new reference for `onFileDropped` for each test, so I prefer to do it this way.

atomrc (approved):
😎🚀
Love it

src/script/components/DropFileArea/DropFileArea.test.tsx
przemvs: Sure :)

przemvs (approved):

thisisamir98 (approved):

PR #14319, opened 2022-12-14
fix: Firefox chat area accessibility issues, stop infinite loop for chat messages using arrow keys(last-first and vice versa), focus on non content messages interactive elements(ACC-268)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-268" title="ACC-268" target="_blank">SubtaskACC-268</a>  9.2.4.3 Focus order for chat content area
  </td></table>
  <br />
 

----

- The **PR Title**
- Firefox chat area accessibility issues, stop infinite loop for chat messages, focus on non content messages interactive elements(ACC-268)](fix: Firefox chat area accessibility issues, stop infinite loop for chat messages using arrow keys(last-first and vice versa), focus on non content messages interactive elements(ACC-268)

- The **PR Description**
  - Firefox chat area accessibility issues, stop infinite loop for chat messages, focus on non content messages interactive elements(ACC-268)](fix: Firefox chat area accessibility issues, stop infinite loop for chat messages using arrow keys(last-first and vice versa), focus on non content messages interactive elements
----
##### References
1. https://wearezeta.atlassian.net/browse/ACC-268


src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset.tsx
```diff
@@ -43,9 +43,16 @@ export interface ImageAssetProps {
   message: ContentMessage;
   onClick: (message: ContentMessage, event: React.MouseEvent | React.KeyboardEvent) => void;
   teamState?: TeamState;
+  isCurrentConversationFocused?: boolean;
```
przemvs, line 46: Use isFocused, because we don't need to know if is conversation focused, it should be more generic name :)

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -257,13 +257,18 @@ const MessagesList: FC<MessagesListParams> = ({
   }, [loaded]);
 
   const defaultFocus = -1;
-  const {currentFocus, handleKeyDown, setCurrentFocus} = useRoveFocus(filteredMessagesLength, defaultFocus);
+  const isMsgListInfinite = false;
```
przemvs, line 260: It's necessary there ? You can't pass false directly ? 🤔

src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset.tsx
arjita-mitra: I am not changing this name now, this name is being used everywhere and it has a relation with the conversation is focused or not.

src/script/components/MessagesList/MessageList.tsx
arjita-mitra: I feel its necessary. just passing a false value doesn't really say what is false thats why I have added a variable here. In future I would like to pass it as an Object and there we can use the variable name directly in the param list.

przemvs: So please change it to Object, it will be more readable :)

przemvs: Because later it will be hard to refactor.

PatrykBuniX (approved):

PR #14322, opened 2022-12-14
feat: replace numbers with enum for tabIndex 

aweiss-dev:
Replace tabIndex with enums

### DIFF
--- a/src/script/components/MessagesList/Message/util/index.ts
+++ b/src/script/components/MessagesList/Message/util/index.ts
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2021 Wire Swiss GmbH
+ * Copyright (C) 2022 Wire Swiss GmbH
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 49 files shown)


V-Gira (approved):

PatrykBuniX (approved):

src/script/components/MessagesList/Message/util/index.ts
```diff
@@ -0,0 +1,30 @@
+/*
```
tlebon, line 1: what about titling this file the name of the hook?

tlebon (approved):

przemvs (approved):

src/script/components/MessagesList/Message/util/index.ts
przemvs: You're right! :)

aweiss-dev: I didn't do that on purpose. The functionality is not big enough to be placed into its own file imho. Index is fine in this case and makes the import shorter :)

PR #14327, opened 2022-12-15
refactor: Replace functions binding to arrow functions

przemvs:
Replaced functions binding with arrow functions.


src/script/components/MessagesList/Message/ContentMessage/asset/FileAssetComponent.tsx
```diff
@@ -94,14 +97,14 @@ const FileAsset: React.FC<FileAssetProps> = ({
           role="button"
           tabIndex={messageFocusedTabIndex}
           aria-label={`${t('conversationContextMenuDownload')} ${fileName}.${fileExtension}`}
-          onClick={() => {
+          onClick={async () => {
             if (isUploaded) {
-              downloadAsset(asset);
+              await onDownloadAsset();
```
PatrykBuniX: What about moving this if statement to the `onDownloadAsset` function? We could just pass the function name to `onClick`

PatrykBuniX (approved):

PR #14329, opened 2022-12-15
feat: Send REJECT message through the self MLS conversation [FS-1327]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1327" title="FS-1327" target="_blank">TaskFS-1327</a>  [web] Send REJECT in the selfConversation when call if picked up from once device
  </td></table>
  <br />
 

This will allow warning the other of the user's devices that a call has been rejected or picked up from another device (and thus allow the current device to stop ringing). 

Since this message is sent through the self MLS conversation, we need to add a new payload that indicates which conversation is actually ringing (it is not the self conversation that is ringing)

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -559,9 +559,8 @@ export class CallingRepository {
   }
 
   private extractTargetedConversationId(event: CallingEvent): QualifiedId {
-    const {content, conversation, qualified_conversation} = event;
-    const targetedConversationId =
-      'qualified_conversation' in content ? content.qualified_conversation : qualified_conversation;
+    const {targetConversation, conversation, qualified_conversation} = event;
+    const targetedConversationId = targetConversation || qualified_conversation;
     const conversationId = targetedConversationId ?? {domain: '', id: conversation};
 
     return conversationId;
@@ -1120,7 +1119,7 @@ export class CallingRepository {
       const parsedPayload = JSON.parse(payload);
       const messageType = parsedPayload.type as CALL_MESSAGE_TYPE;
       if (messageType === CALL_MESSAGE_TYPE.REJECT) {
-        return void this.messageRepository.sendSelfCallingMessage(payload);
+        return void this.messageRepository.sendSelfCallingMessage(payload, conversation.qualifiedId);
       }
     }
 
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -558,6 +558,11 @@ export class CallingRepository {
     );
   }
 
+  /**
+   * Will extract the conversation that is referred to by the calling event.
+   * It could happen that messages relative to a particular conversation was sent in a different conversation (most likely the self conversation).
+   * We need to find the correct conversation in order for AVS to know which call is concerned
+   */
   private extractTargetedConversationId(event: CallingEvent): QualifiedId {
     const {targetConversation, conversation, qualified_conversation} = event;
     const targetedConversationId = targetConversation || qualified_conversation;
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -558,13 +558,21 @@ export class CallingRepository {
     );
   }
 
+  private extractTargetedConversationId(event: CallingEvent): QualifiedId {
+    const {content, conversation, qualified_conversation} = event;
+    const targetedConversationId =
+      'qualified_conversation' in content ? content.qualified_conversation : qualified_conversation;
+    const conversationId = targetedConversationId ?? {domain: '', id: conversation};
+
+    return conversationId;
+  }
+
   /**
    * Handle incoming calling events from backend.
    */
   onCallEvent = async (event: CallingEvent, source: string): Promise<void> => {
     const {
       content,
-      conversation,
       qualified_conversation,
       from,
       qualified_from,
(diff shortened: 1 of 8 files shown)


aweiss-dev (approved):

src/script/calling/CallingRepository.ts
```diff
@@ -558,13 +558,25 @@ export class CallingRepository {
     );
   }
 
+  /**
+   * Will extract the conversation that is referred to by the calling event.
+   * It could happen that messages relative to a particular conversation was sent in a different conversation (most likely the self conversation).
+   * We need to find the correct conversation in order for AVS to know which call is concerned
+   */
+  private extractTargetedConversationId(event: CallingEvent): QualifiedId {
+    const {targetConversation, conversation, qualified_conversation} = event;
```
arjita-mitra, line 567: I think its better to rename it to qualifiedConversation to maintain the naming convention. in the next line we are using 2 different convention otherwise

src/script/calling/CallingRepository.ts
```diff
@@ -574,27 +586,26 @@ export class CallingRepository {
...
-    const conversationId = isFederated ? qualified_conversation : {domain: '', id: conversation};
     const currentTimestamp = this.serverTimeHandler.toServerTimestamp();
     const toSecond = (timestamp: number) => Math.floor(timestamp / 1000);
     const contentStr = JSON.stringify(content);
-    const conversationEntity = this.conversationState.findConversation(conversationId);
+    const conversationId = this.extractTargetedConversationId(event);
+    const conversation = this.conversationState.findConversation(conversationId);
 
-    if (!conversationEntity) {
+    if (!conversation) {
```
arjita-mitra, line 595: Conversation can never be undefined?

src/script/calling/CallingRepository.ts
```diff
@@ -1112,7 +1124,7 @@ export class CallingRepository {
       const parsedPayload = JSON.parse(payload);
       const messageType = parsedPayload.type as CALL_MESSAGE_TYPE;
       if (messageType === CALL_MESSAGE_TYPE.REJECT) {
-        return;
+        return void this.messageRepository.sendSelfCallingMessage(payload, conversation.qualifiedId);
```
arjita-mitra, line 1127: Haven't seen this return void statement. Could you explain what's happening here?

src/script/conversation/ConversationState.ts
```diff
@@ -137,6 +137,14 @@ export class ConversationState {
     return proteusConversation;
   }
 
+  getSelfMLSConversation(): Conversation {
+    const proteusConversation = this.selfMLSConversation();
+    if (!proteusConversation) {
```
arjita-mitra: proteusConversation won't be undefined right?

src/script/conversation/MessageRepository.ts
```diff
@@ -1259,15 +1259,31 @@ export class MessageRepository {
...
    *
    * @param eventInfoEntity Event info to be send
    * @param conversationId id of the conversation to send call message to
    * @returns Resolves when the confirmation was sent
    */
-  public sendCallingMessage(conversation: Conversation, payload: string, options: MessageSendingOptions = {}) {
-    const message = MessageBuilder.buildCallMessage({content: payload});
+  public sendCallingMessage(
+    conversation: Conversation,
+    payload: string | {content: string; qualifiedConversationId: QualifiedId},
```
arjita-mitra, line 1282: sendSelfCallingMessage(payload: string)

But here we are passing string or object, why?

PatrykBuniX (approved):

src/script/calling/CallingRepository.ts
atomrc: Well it can if the ID in the event payload doesn't correspond to a conversation we have locally

src/script/calling/CallingRepository.ts
atomrc: It's the javascript `void` operator here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

It's to be absolutely sure we don't return the value returned by `sendSelfCallingMessage`. 

Another way to do it would be
```js
this.messageRepository.sendSelfCallingMessage(payload, conversation.qualifiedId);
return;
```

In this case I thing `void` is pretty nice :)

src/script/conversation/ConversationState.ts
PatrykBuniX: We should rename it `proteusConversation` to `mlsConversation` I guess it's copy-paste mistake here 😅

atomrc: It can if we are in a non-mls environment

src/script/conversation/MessageRepository.ts
atomrc: See [here](https://github.com/wireapp/wire-webapp/pull/14329/files#diff-8a277356da1eb1435094433268c1e8cf7392c374f617174bd0a6e60ea52ea4ebR1268), now we need a more complex object under certain circumstances (namely when in an MLS conversation).

src/script/conversation/ConversationState.ts
atomrc: @PatrykBuniX oh damn 😅

arjita-mitra: 😆

src/script/calling/CallingRepository.ts
atomrc: Yeah, this thing with encoded message payloads is that we have those 2 convention. 
But, TBH, it feels like a waste of resources to rename at this point since it's super located and that this variable is renamed to a camel cased variable on the next line. 
WDYT?

src/script/conversation/ConversationState.ts
arjita-mitra: in that case it should be !! mlsConversation?

src/script/calling/CallingRepository.ts
arjita-mitra: got it, I am familiar with the second syntax.

PR #14080, opened 2022-11-09
feat: Add typing indicator (ACC-57)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-57" title="ACC-57" target="_blank">StoryACC-57</a>  Typing indicator (6.2.1.1)
  </td></table>
  <br />
 

When 1 user is typing:
telegram-cloud-photo-size-4-6005718513517443730-y

When 2 users are typing:
telegram-cloud-photo-size-4-6005718513517443732-y

When more than 2 users are typing:
telegram-cloud-photo-size-4-6005718513517443731-y

Typing Animation:

PR #14330, opened 2022-12-15
fix: make sidebar radios keyboard accessible [ACC-332]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-332" title="ACC-332" target="_blank">BugACC-332</a>  [Web] Can't reach 'Self-deleting messages' time option via keyboard
  </td></table>
  <br />
 

### Issue

- radios for self deleting message options are not keyboard accessible

### Solution
- use the former `PreferenceRadio` component for all radios

### Change
- since the component is used in different places, refactor `PreferenceRadio` to just `Radio`
- move css from less files to cs in js (taking inspiration from https://github.com/wireapp/wire-webapp/pull/14080)
- remove unused css from less files


src/script/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <div data-uie-name={uieName}>
+      {options.map(({value, label, detailLabel, isDisabled = false}) => {
+        const currentId = id + value;
+
+        return (
+          <div
+            key={value}
+            css={radioOptionStyles}
+            role="radiogroup"
```
przemvs: Hmm... This role should be on parent element, here for e.g. on line 49, also aria-labelledby :)

But for better accessibility you should use there: fieldset :)

src/script/components/Radio/Radio.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const Radio = <T extends string | number>({
+  ariaLabelledBy,
+  name,
+  options,
+  onChange,
+  uieName = name,
+  selectedValue,
+}: RadioProps<T>) => {
+  const {current: id} = useRef(Math.random().toString(36).slice(2));
```
thisisamir98: let's use react's internal useId here.
https://reactjs.org/docs/hooks-reference.html#useid

thisisamir98 (approved):
well done, it would be great if you could add some tests for Radio.tsx
(good example is ButtonGroup.test.tsx)

src/script/components/Radio/RadioGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <input
+              css={radioInputStyles(isDisabled)}
+              disabled={isDisabled}
+              tabIndex={TabIndex.FOCUSABLE}
+              type="radio"
+              id={currentId}
+              name={name}
+              value={value}
+              onChange={() => onChange(value)}
+              checked={selectedValue === value}
```
przemvs: Move this selectedValue === value to variable, you have used it in 2 places :)

przemvs (approved):

przemvs (approved):

PR #14332, opened 2022-12-16
refactor: remove lru-cache lib in favour of Map

PatrykBuniX:
Use native `Map` object instead of `lru-cache` lib.

### DIFF
--- a/src/script/assets/AssetURLCache.ts
+++ b/src/script/assets/AssetURLCache.ts
@@ -30,7 +30,5 @@ export const setAssetUrl = (identifier: string, url: string) => {
   }
 
   cache.set(identifier, url);
-  window.URL.revokeObjectURL(url);
-
   return url;
 };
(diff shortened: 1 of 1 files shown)


src/script/assets/AssetURLCache.ts
```diff
@@ -31,11 +29,8 @@ export const setAssetUrl = (identifier: string, url: string) => {
     return isExistingUrl;
   }
 
-  const outdatedUrl = cache.set(identifier, url);
-
-  if (outdatedUrl != null) {
-    window.URL.revokeObjectURL(outdatedUrl);
-  }
+  cache.set(identifier, url);
+  window.URL.revokeObjectURL(url);
```
atomrc: That sounds wrong. Revoking the URL after setting it, that seems weird 🙃

PatrykBuniX: Isn't it what we've been doing before also? 🤔

atomrc: It's the old url that should be revoked

atomrc (approved):

PR #14333, opened 2022-12-16
fix: Hint about all conversations being archived

przemvs:
Before:
image

After:
image

### DIFF
--- a/src/style/list/conversations.less
+++ b/src/style/list/conversations.less
@@ -152,7 +152,6 @@
   .flex-center;
   .label-xs;
   .text-center;
-  .text-white;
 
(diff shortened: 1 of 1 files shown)


src/style/list/conversations.less
```diff
@@ -152,7 +152,6 @@
   .flex-center;
   .label-xs;
   .text-center;
-  .text-white;
```
atomrc, line 155: Does it still work in dark mode now?

przemvs: Tested and works. It's inherited from from parent element where we have defined variable :)

image

atomrc (approved):

PR #14354, opened 2022-12-19
refactor: Move singleInstance to react Hook

atomrc:
And remove Redux handler for auth page


src/script/hooks/useSingleInstance.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Returns `true` if the instance has been registered successfully.
+   * Returns `false` if the app is already running in another instance.
+   *
+   * Side effects: will also start the interval check if a callback was provided in the constructor
+   *
+   * @returns Was the app registered successfully.
+   */
+  registerInstance() {
+    this.instanceId = createRandomUuid();
+    const cookieName = CONFIG.COOKIE_NAME;
```
przemvs: Hmm.. Why not defined as variable in class ? If you define it in class you will use everywhere cookieName instead of CONFIG.COOKIE_NAME. WDYT?

przemvs:
Great Job @atomrc !

przemvs (approved):

PR #14361, opened 2022-12-20
fix: close right sidebar when changing view [ACC-352]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-352" title="ACC-352" target="_blank">BugACC-352</a>  [Web] Conversation details stay open when going back to conversation list in collapsed view
  </td></table>
  <br />
 

### Issue
- Right sidebar stays open when changing views in responsive view
Peek 2022-12-20 13-02

### Solution
- Close the right sidebar when clicking the back button
Peek 2022-12-20 13-49

### DIFF
--- a/src/script/components/TitleBar/TitleBar.tsx
+++ b/src/script/components/TitleBar/TitleBar.tsx
@@ -143,10 +143,7 @@ export const TitleBar: React.FC<TitleBarProps> = ({
 
   const {setCurrentView: setView} = useAppMainState(state => state.responsiveView);
 
-  const setLeftSidebar = () => {
-    setView(ViewType.LEFT_SIDEBAR);
-  };
-
+  const setLeftSidebar = () => setView(ViewType.LEFT_SIDEBAR);
   const {close: closeRightSidebar} = useAppMainState(state => state.rightSidebar);
 
   const onClickBack = () => {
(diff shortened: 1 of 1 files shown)


przemvs (approved):

przemvs:
You test it on desktop if it works properly ?

PatrykBuniX (approved):

src/script/components/TitleBar/TitleBar.tsx
```diff
@@ -144,6 +144,12 @@ export const TitleBar: React.FC<TitleBarProps> = ({
   const {setCurrentView: setView} = useAppMainState(state => state.responsiveView);
 
   const setLeftSidebar = () => setView(ViewType.LEFT_SIDEBAR);
```
V-Gira: `setLeftSidebar`is also used when clicking the call button and shouldn't close the right sidebar in that case so an ` onClickBack`function is added

atomrc (approved):

src/script/components/TitleBar/TitleBar.tsx
```diff
@@ -144,6 +144,12 @@ export const TitleBar: React.FC<TitleBarProps> = ({
...
+  const {close: closeRightSidebar} = useAppMainState(state => state.rightSidebar);
+
+  const onClickBack = () => {
```
atomrc: As a general rule of thumb, functions that you declare should have an imperative name to explain what it's doing. 

The rationale for this is that you are implementing the function that you will pass as a callback to some event. But at this point the `event` is not really relevant to the function. What is interesting is explaining what the function is actually doing:

```js
onClick={onClick} => this doesn't carry any information
onClick={closePanel} => this line carries the information of what is going to happen when a click is done
```

PR #14367, opened 2022-12-21
fix: Change font size of seach placeholder (ACC-327)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-327" title="ACC-327" target="_blank">BugACC-327</a>  [Web] User search reminder text is cut-off on federated env
  </td></table>
  <br />
 

Before:
telegram-cloud-photo-size-4-5843923659503286854-x

After:
image

### DIFF
--- a/src/style/components/search-input.less
+++ b/src/style/components/search-input.less
@@ -98,7 +98,7 @@
 
   &:placeholder-shown {
     .ellipsis-nowrap;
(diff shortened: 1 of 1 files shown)


V-Gira (approved):
That'll work :smile: It's cleared with design?

PR #14369, opened 2022-12-21
runfix: hash entropy to 32 bytes

atomrc:
Previously, Cryptobox was taking care of hashing the entropy data generated. 
But since the migration to coreCrypto, we need to give entropy that has the right amount of data

### DIFF
--- a/src/script/auth/page/EntropyContainer.tsx
+++ b/src/script/auth/page/EntropyContainer.tsx
@@ -47,7 +47,7 @@ const EntropyContainer = ({onSetEntropy, containerSize = 400}: Props) => {
     setPercent(percentage);
   };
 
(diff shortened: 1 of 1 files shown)


thisisamir98 (approved):

src/script/auth/page/EntropyContainer.tsx
```diff
@@ -47,6 +47,12 @@ const EntropyContainer = ({onSetEntropy, containerSize = 400}: Props) => {
     setPercent(percentage);
   };
 
+  const fowardEntropy = async (entropy: Uint8Array) => {
```
PatrykBuniX: shouldn't be `forward` with `r`?

PatrykBuniX (approved):

PR #14373, opened 2022-12-22
feat: add debug util to break lastNotificationId

PatrykBuniX:
Adds debug util that replaces currently stored `lastNotificationId` with random `uuid`.

### DIFF
--- a/src/script/util/util.ts
+++ b/src/script/util/util.ts
@@ -255,7 +255,8 @@ export const downloadFile = (url: string, fileName: string, mimeType?: string): 
   }, 100);
 };
 
-export const createRandomUuid = (): string => UUID.genV4().toString();
+export const createRandomUuid = (version: 4 | 1 = 4): string =>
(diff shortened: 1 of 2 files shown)


src/script/util/util.ts
```diff
@@ -255,7 +255,8 @@ export const downloadFile = (url: string, fileName: string, mimeType?: string):
   }, 100);
 };
 
-export const createRandomUuid = (): string => UUID.genV4().toString();
+export const createRandomUuid = (version: 4 | 1 = 4): string =>
+  (version === 4 ? UUID.genV4() : UUID.genV1()).toString();
```
atomrc, line 259: why do we need v1 for this? I guess v4 would do nicely as well, isn't it?

PatrykBuniX: Backend does return 500 error when v4 is used (v4 does not contain timestamps anymore what I believe is the case).

atomrc (approved):

PR #14375, opened 2022-12-22
refactor: Remove deprecated properties of mainViewModel

atomrc:


src/script/components/AppContainer/AppContainer.tsx
```diff
@@ -56,6 +56,12 @@ export const AppContainer: FC<AppProps> = ({config, clientType}) => {
     window.addEventListener('beforeunload', killInstance);
   }, []);
 
+  useEffect(() => {
+    // Prevent Chrome (and Electron) from pushing the content out of the
+    // viewport when using form elements (e.g. in the preferences)
+    document.addEventListener('scroll', () => window.scrollTo(0, 0));
```
PatrykBuniX: What about clearing this event?

atomrc: Very good call 👌

PatrykBuniX (approved):
👌

PR #14395, opened 2022-12-28
fix: Migrate to coreCrypto on login and app load

atomrc:
Right now when logging in with an already existing device will generate a new identity. 
Since https://github.com/wireapp/wire-web-packages/pull/4734 the core will also migrate cryptobox before any call to `proteus.init` ensuring that data have been migrated regardless of where it was init from the webapp


src/script/auth/module/action/AuthAction.ts
```diff
@@ -337,7 +337,7 @@ export class AuthAction {
         }
         const clientType = persist ? ClientType.PERMANENT : ClientType.TEMPORARY;
 
-        await core.init(clientType, {initClient: false});
+        await core.init(clientType);
```
atomrc, line 336: This `init` method now never init the client, it has to be called in a second step (which is a more natural way of using the api)

src/script/auth/module/action/creator/ClientActionCreator.ts
```diff
@@ -22,8 +22,6 @@ import type {RegisteredClient} from '@wireapp/api-client/lib/client/';
 import type {AppAction} from './';
 
 export enum CLIENT_ACTION {
-  CLIENT_INIT_FAILED = 'CLIENT_INIT_FAILED',
-  CLIENT_INIT_START = 'CLIENT_INIT_START',
   CLIENT_INIT_SUCCESS = 'CLIENT_INIT_SUCCESS',
```
atomrc, line 25: Thos events are never consumed

src/script/main/app.ts
```diff
@@ -362,18 +360,12 @@ export class App {
...
-        }
-        await this.core.runCryptoboxMigration();
-        await this.core.initClient({clientType});
-      } catch (error) {
+      await this.core.init(clientType);
+      const locaClient = await this.core.loadClient();
+      if (!locaClient) {
         throw new ClientError(CLIENT_ERROR_TYPE.NO_VALID_CLIENT, 'Client has been deleted on backend');
       }
+      await this.core.initClient(locaClient);
```
atomrc: The entire migration is now done by the core internally

src/script/auth/module/action/ClientAction.ts
```diff
@@ -62,20 +62,21 @@ export class ClientAction {
...
-      } catch (error) {
-        dispatch(ClientActionCreator.failedInitializeClient(error));
-        throw error;
-      }
+      const localClient = await core.initClient();
+      const creationStatus = localClient
+        ? {isNew: false, client: localClient}
+        : {
+            isNew: true,
+            client: await core.registerClient(
```
tlebon, line 70: what happens if registerClient throws an error?

atomrc: We are going to have a nasty error in the console, and that's fine there are not exception that we should handle at this point (it would mean that the backend is having hard time I guess)

atomrc: Actually, let me try the "to many devices" scenario now 🤔 
You might have a point

atomrc: Oh, it's handled correctly as well. We just don't have the `failedInitializeClient` call (that was useless anyway)
But we still throw the backend error

atomrc: Yep, confirmed with a test 👌 
Thanks for pointing it out

src/script/main/app.ts
```diff
@@ -362,16 +360,9 @@ export class App {
...
-      try {
-        context = await this.core.init(clientType, {initClient: false});
-        if (context.domain) {
-          // Migrate all existing session to fully qualified ids (if need be)
-          await migrateToQualifiedSessionIds(this.core.storage.db.sessions, context.domain);
-        }
-        await this.core.runCryptoboxMigration();
-        await this.core.initClient({clientType});
-      } catch (error) {
+      await this.core.init(clientType);
```
tlebon, line 363: just noticing the init options arent passed here, are they used sometime else?

tlebon: confirmed in ETS

src/script/auth/module/action/creator/AuthActionCreator.ts
```diff
@@ -70,9 +65,6 @@ export enum AUTH_ACTION {
   SEND_TWO_FACTOR_CODE_SUCCESS = 'SEND_TWO_FACTOR_CODE_SUCCESS',
   SILENT_LOGOUT_FAILED = 'SILENT_LOGOUT_FAILED',
   SILENT_LOGOUT_SUCCESS = 'SILENT_LOGOUT_SUCCESS',
-  VALIDATE_LOCAL_CLIENT_FAILED = 'VALIDATE_LOCAL_CLIENT_FAILED',
-  VALIDATE_LOCAL_CLIENT_START = 'VALIDATE_LOCAL_CLIENT_START',
-  VALIDATE_LOCAL_CLIENT_SUCCESS = 'VALIDATE_LOCAL_CLIENT_SUCCESS',
```
atomrc, line 75: Those are never fired nor consumed

src/script/main/app.ts
```diff
@@ -362,16 +360,9 @@ export class App {
...
+      const locaClient = await this.core.initClient();
+      if (!locaClient) {
```
tlebon: seems like a typo 
```suggestion
      const localClient = await this.core.initClient();
      if (!localClient) {
```

tlebon (approved):
update the typo if you like but otherwise its good

src/script/main/app.ts
atomrc: Woups 😅  nice catch

PR #14398, opened 2022-12-29
refactor: Only load client list where needed

atomrc:
This will:

- deprecate the history actions that are not needed since https://github.com/wireapp/wire-webapp/pull/7610/files#diff-57a49554ed92ac15e6745a24211bfd0c9b6017f3af59a56282ea7aa6bf176694L80
- avoid loading the clients list everytime we initialize the local device. It's only needed in the device manager and history info, so let's only load them here


src/script/auth/module/action/ClientAction.ts
```diff
@@ -74,7 +74,6 @@ export class ClientAction {
             ),
           };
 
-      await dispatch(clientAction.doGetAllClients());
```
atomrc, line 77: This was way overkill and was called too many times while the data was not consumed

src/script/auth/module/action/NotificationAction.ts
```diff
@@ -17,30 +17,9 @@
...
-        throw error;
-      }
-    };
-  };
-
-  resetHistoryCheck = (): ThunkAction => {
-    return async dispatch => {
-      dispatch(NotificationActionCreator.resetHistoryCheck());
-    };
-  };
```
atomrc, line 42: deprecated since https://github.com/wireapp/wire-webapp/pull/7610/files#diff-57a49554ed92ac15e6745a24211bfd0c9b6017f3af59a56282ea7aa6bf176694L80

src/script/auth/module/action/ClientAction.ts
atomrc: #SingleReponsibilityFailure 🙃

src/script/auth/module/reducer/clientReducer.ts
```diff
@@ -19,24 +19,22 @@
...
+  currentClient: RegisteredClient | null;
+  error: Error | null;
   fetching: boolean;
-  hasHistory: boolean;
   isNewClient: boolean;
 }
 
 export const initialClientState: ClientState = {
-  clients: [],
+  clients: null,
```
atomrc, line 33: This default value is to be able to detect if clients have been loaded at all or if it's really an empty list of devices

src/script/auth/page/HistoryInfo.tsx
```diff
@@ -70,10 +76,10 @@ const HistoryInfoComponent = ({
   return (
     <Page>
       <ContainerXS centerText verticalCenter style={{width: '100%'}}>
-        <H1 center>{_(headline, {brandName: Config.getConfig().BRAND_NAME})}</H1>
+        <H1 center>{_(historyInfoStrings.noHistoryHeadline, {brandName: Config.getConfig().BRAND_NAME})}</H1>
         <Paragraph center style={{marginBottom: 56}}>
           <FormattedMessage
-            {...infoText}
+            {...historyInfoStrings.noHistoryInfo}
```
tlebon, line 82: im a bit confused by this change. if you have history shouldnt it show a different text?

atomrc: The thing is: it's now impossible to see this screen and have an actual history. This is due to past limitations that were deprecated since https://github.com/wireapp/wire-webapp/pull/7610

tlebon: sharing links from pre-covid 😂. approved then.

tlebon (approved):

src/script/auth/page/HistoryInfo.tsx
atomrc: Yeah, I'm digging into some old stuff in there 😆

PR #14404, opened 2022-12-29
fix: Improve user-facing decryption errors

atomrc:
This will handle 3 new types of decryption errors:

- Invalid Signatures errors
image

- invalid message errors
image

- remote identity changed error
image

Needs:
- [ ] https://github.com/wireapp/wire-web-packages/pull/4735

### DIFF
--- a/src/script/components/MessagesList/Message/DecryptErrorMessage.tsx
+++ b/src/script/components/MessagesList/Message/DecryptErrorMessage.tsx
@@ -17,12 +17,12 @@
  *
  */
 
-import React from 'react';
+import React, {useState} from 'react';
 
 import {DeviceId} from 'Components/DeviceId';
 import {Icon} from 'Components/Icon';
 import {getWebsiteUrl, URL_PATH} from 'src/script/externalRoute';
-import {useKoSubscribableChildren} from 'Util/ComponentUtil';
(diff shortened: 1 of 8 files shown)


src/script/components/MessagesList/Message/DecryptErrorMessage.tsx
```diff
@@ -33,17 +33,18 @@ export interface DecryptErrorMessageProps {
...
-    '/highlight': '</span>',
-    highlight: '<span class="label-bold-xs">',
-  });
+  const caption = message.isIdentityChanged
+    ? t('conversationUnableToDecrypt2', message.user().name(), {
+        '/highlight': '</span>',
+        highlight: '<span class="label-bold-xs">',
+      })
+    : t('conversationUnableToDecrypt1', message.user().name(), {
+        '/highlight': '</span>',
```
tlebon, line 45: i know this is legacy but is there a reason this needs to have a backslash in the name instead of something like `closingHighlight`?

atomrc: This is how it is in the string copy. Changing this would discard all the translations we have on crowdin. 

The copy looks like this

```text
[highlight]{{user}}[/highlight]´s device identity changed. Undelivered message.
```

PR #14421, opened 2023-01-02
runfix: Handle Auth error at app load

atomrc:
Was broken since a little while, this initial `try/catch` condition was a little to generic
https://github.com/wireapp/wire-webapp/pull/14395/files#

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -59,7 +59,7 @@ import {ConversationRepository} from '../conversation/ConversationRepository';
 import {ConversationService} from '../conversation/ConversationService';
 import {MessageRepository} from '../conversation/MessageRepository';
 import {CryptographyRepository} from '../cryptography/CryptographyRepository';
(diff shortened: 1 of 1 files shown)


src/script/main/app.ts
```diff
@@ -476,66 +478,47 @@ export class App {
...
-      const isInvalidClient = type === ClientError.TYPE.NO_VALID_CLIENT;
-
-      if (isInvalidClient) {
-        return this.redirectToLogin(SIGN_OUT_REASON.SESSION_EXPIRED);
-      }
-
-      if (isAccessTokenError) {
-        this.logger.warn('Connectivity issues. Trigger reload.', error);
-        return window.location.reload();
-      }
```
atomrc, line 517: The only difference when it's a reload is that the reason for redirecting to `/auth` is different. 
This doesn't really make sense to me right now, but well as I don't really know what is the intent, I prefer to transcribe as is

PatrykBuniX (approved):

przemvs (approved):

V-Gira (approved):

PR #14425, opened 2023-01-02
feat: Call tiles focus [#ACC-336]

przemvs:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-336" title="ACC-336" target="_blank">BugACC-336</a>  [Web] Call tiles in group calls are not reachable with keyboard
  </td></table>
  <br />
 

User tiles in call should be focusable when user have a maximized call screen.

### DIFF
--- a/src/script/components/calling/GroupVideoGridTile.tsx
+++ b/src/script/components/calling/GroupVideoGridTile.tsx
@@ -20,12 +20,14 @@
 import React from 'react';
 
 import {QualifiedId} from '@wireapp/api-client/lib/user';
+import {TabIndex} from '@wireapp/react-ui-kit/lib/types/enums';
 
 import {VIDEO_STATE} from '@wireapp/avs';
 
 import {Avatar, AVATAR_SIZE} from 'Components/Avatar';
 import {Icon} from 'Components/Icon';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
+import {isEnterKey} from 'Util/KeyboardUtil';
 import {t} from 'Util/LocalizerUtil';
 
 import {Video} from './Video';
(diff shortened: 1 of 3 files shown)


src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -64,13 +66,24 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
   const activelySpeakingBoxShadow = `inset 0px 0px 0px 1px var(--group-video-bg), inset 0px 0px 0px 4px var(--accent-color), inset 0px 0px 0px 7px var(--app-bg-secondary)`;
   const groupVideoBoxShadow = participantCount > 1 ? 'inset 0px 0px 0px 2px var(--group-video-bg)' : 'initial';
 
+  const onTileClick = () => onParticipantDoubleClick(participant?.user.qualifiedId, participant?.clientId);
```
arjita-mitra: Can you give a more meaningful name? onTileClick is not very clear

arjita-mitra (changes requested):

src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -64,13 +66,24 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
...
+
+  const onTileEnterClick = (keyboardEvent: React.KeyboardEvent) => {
+    if (!isEnterKey(keyboardEvent)) {
+      return;
+    }
+
+    onParticipantDoubleClick(participant?.user.qualifiedId, participant?.clientId);
```
atomrc: You could use the `onTitleClick` here (since it's already bound to the correct data). 
And let's maybe rename it `handleTitleClick` to name it like an action

V-Gira (approved):

atomrc (approved):

PR #14428, opened 2023-01-02
refactor: login page state

PatrykBuniX:
- separates redux state types with `LoginData` type on Login related pages.
- fixes some TS strict errors

### DIFF
--- a/src/script/auth/module/reducer/authReducer.ts
+++ b/src/script/auth/module/reducer/authReducer.ts
@@ -18,6 +18,7 @@
  */
 
 import type {SSOSettings} from '@wireapp/api-client/lib/account/SSOSettings';
+import {LoginData} from '@wireapp/api-client/lib/auth';
 import {ClientType} from '@wireapp/api-client/lib/client/';
 import type {TeamData} from '@wireapp/api-client/lib/team/';
 import type {UserAsset} from '@wireapp/api-client/lib/user/';
@@ -42,10 +43,7 @@ export interface RegistrationDataState {
   termsAccepted: boolean;
 }
 
-export interface LoginDataState {
-  clientType: ClientType;
-  phone?: string;
-}
+export type LoginDataState = Pick<LoginData, 'clientType' | 'phone'>;
 
 export type AuthState = {
   readonly account: RegistrationDataState;
--- a/src/script/auth/module/reducer/authReducer.ts
+++ b/src/script/auth/module/reducer/authReducer.ts
@@ -18,7 +18,6 @@
  */
 
 import type {SSOSettings} from '@wireapp/api-client/lib/account/SSOSettings';
-import type {LoginData} from '@wireapp/api-client/lib/auth/';
 import {ClientType} from '@wireapp/api-client/lib/client/';
 import type {TeamData} from '@wireapp/api-client/lib/team/';
 import type {UserAsset} from '@wireapp/api-client/lib/user/';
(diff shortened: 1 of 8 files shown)


src/script/auth/page/Login.tsx
```diff
@@ -223,11 +224,12 @@ const LoginComponent = ({
             break;
           }
           default: {
+            const backendError = error;
```
PatrykBuniX, line 228: We have to assign value of error to a const to make sure it will be `BackendError` type at this point. `error` from `catch` is modifiable variable, so it could change to `unknown` before function passed to `.some` array method is called.

przemvs (approved):

src/script/auth/page/Login.test.tsx
```diff
@@ -79,7 +80,13 @@ describe('Login', () => {
     const password = 'password';
 
     spyOn(actionRoot.authAction, 'doLogin').and.returnValue(() =>
-      Promise.reject({label: BackendError.LABEL.TOO_MANY_CLIENTS}),
+      Promise.reject(
+        new BackendError({
+          label: BackendError.LABEL.TOO_MANY_CLIENTS,
+          code: StatusCodes.NOT_FOUND,
+          message: 'Too may clients',
```
przemvs: Missing letter :D

PatrykBuniX: Nice catch 👍

src/script/auth/page/Login.tsx
```diff
@@ -392,7 +398,7 @@ const mapStateToProps = (state: RootState) => ({
   isFetching: AuthSelector.isFetching(state),
   isSendingTwoFactorCode: AuthSelector.isSendingTwoFactorCode(state),
   loginData: AuthSelector.getLoginData(state),
-  loginError: AuthSelector.getError(state),
+  authError: AuthSelector.getError(state),
```
PatrykBuniX, line 401: This is not `loginError` it's more generic authError that can occur in some different places:

Screenshot 2023-01-03 at 09 55 28

src/script/auth/page/Login.tsx
```diff
@@ -192,9 +193,8 @@ const LoginComponent = ({
 
       return navigate(ROUTE.HISTORY_INFO);
     } catch (error) {
-      if ((error as BackendError).label) {
-        const backendError = error as BackendError;
-        switch (backendError.label) {
+      if (isBackendError(error)) {
```
PatrykBuniX, line 196: ended up using isBackendError predicate, this way we're sure we are dealing with error that contains label

src/script/auth/page/Login.tsx
atomrc: Ohhh that makes sense. That's a funny one, nice catch 👌

src/script/auth/module/reducer/authReducer.ts
```diff
@@ -43,6 +42,11 @@ export interface RegistrationDataState {
   termsAccepted: boolean;
 }
 
+export interface LoginDataState {
+  clientType: ClientType;
+  phone?: string;
+}
```
atomrc:
```suggestion
export type LoginDataState = Pick<LoginData, 'clientType' | 'phone'>.
```

Would this work? Or are they slightly different?

atomrc (approved):

src/script/auth/module/reducer/authReducer.ts
PatrykBuniX: Works fine, I like this approach 🚀

PR #14429, opened 2023-01-03
test: debounce hook test

arjita-mitra:
----

- The **PR Title**
  - Unit test for debounce hook

- The **PR Description**
  - Unit test for debounce hook
----

### DIFF
--- a/src/script/hooks/useDebounce.test.tsx
+++ b/src/script/hooks/useDebounce.test.tsx
@@ -22,8 +22,15 @@ import {renderHook, act} from '@testing-library/react';
 import {useDebounce} from './useDebounce';
 
 describe('useDebounce', () => {
-  it('should debounce the callback', () => {
+  beforeAll(() => {
     jest.useFakeTimers();
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
+
+  it('should debounce the callback', () => {
     const callback = jest.fn();
     const time = 1000;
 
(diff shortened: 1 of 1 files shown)


src/script/hooks/useDebounce.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should not call the callback if it is unmounted before the debounce time', () => {
+    const callback = jest.fn();
+    const time = 1000;
+
+    const {unmount} = renderHook(() => useDebounce(callback, time));
+
+    act(() => {
+      unmount();
+      jest.advanceTimersByTime(500);
```
PatrykBuniX: Shouldn't this time be more than 1000 here? To make sure callback would run if component wasn't unmounted? This tests will pass even when your remove `unmount()` from here.

arjita-mitra: I got your point but this test is - should not call the callback if it is unmounted before the debounce time

we need to make it - should not call the callback if it is unmounted after the debounce time

arjita-mitra: actually before/after time doesn't matter. if the component is unmounted before you advanced the timer to something > 1000 / <1000 the callback won't be called. so the important thing here is the component is unmounted before the timer could finish

PatrykBuniX: `should not call the callback if it is unmounted before the debounce time` callback will never be called in this case, doesn't matter if we unmount or not. So this is exactly the same thing you're testing in the first case - that callback should not run before debounce time passes

arjita-mitra: I think it makes more sense to advance the timer above 1000 to make the purpose of the test more clear

PatrykBuniX (approved):
Nice!

PR #14439, opened 2023-01-04
fix: Stop handling team events when team is deleted [FS-1406]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1406" title="FS-1406" target="_blank">BugFS-1406</a>  [Web] Database is accessed after is has been closed when team is deleted
  </td></table>
  <br />
 

It is possible for backend to send team events even after the team has been deleted.
On the front end it might end-up in calls that will try to access the DB after it has been closed or deleted. 

The solution is to stop handling team events once the team has been locally marked as deleted

### DIFF
--- a/src/script/team/TeamRepository.ts
+++ b/src/script/team/TeamRepository.ts
@@ -216,10 +216,15 @@ export class TeamRepository {
   }
 
   readonly onTeamEvent = (eventJson: any, source: EventSource): void => {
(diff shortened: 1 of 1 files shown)


PatrykBuniX (approved):
👍

V-Gira (approved):

PR #14443, opened 2023-01-04
feat: mls conferencing [FS-1282]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1282" title="FS-1282" target="_blank">TaskFS-1282</a>  [Web] Implement MLS conferencing
  </td></table>
  <br />
 

MLS conferencing works a little differently than proteus conferencing. The gist of it is the following

## Proteus strategy (current)

when someone joins a conference call, a targeted message is sent to this device with the encryption key of the conference.
This is possible because proteus allows targeted messages

## MLS strategy
Since MLS doesn't provide targeted messages inside a conversation, the encryption key is derived from a sub-conversation  key material. This sub-conversation is created dynamically once the call is started.
Every time someone joins the call (and thus the sub-conversation) and new epoch is triggered and a new secret is used for encryption.
We then need to subscribe to epoch change in order to derive the new conference key once someone new joins the call. 

Needs
- https://github.com/wireapp/wire-web-packages/pull/4811


src/script/calling/callingSubscriptionsHandler.ts
```diff
@@ -0,0 +1,43 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {QualifiedId} from '@wireapp/api-client/lib/user';
+
+const serializeQualifiedId = ({id, domain}: QualifiedId) => `${id}@${domain}`;
```
thisisamir98, line 22: calling repository has the exact same function. maybe we could extract/merge them to a single location or turn that method to static and use it from there?
https://github.com/wireapp/wire-webapp/blob/dev/src/script/calling/CallingRepository.ts#L742

atomrc: This doesn't need to match the ids that the calling repository is generating. This is internal behavior for the subscription handler and there is no need to leak the implementation between those 2 identities, they can be completely standalone

src/script/calling/CallingRepository.test.ts
```diff
@@ -458,7 +488,7 @@ describe.skip('E2E audio call', () => {
 
   it('calls and connect with the remote user', done => {
     onCallClosed = done;
-    const conversationId = createConversationId();
+    const conversationId = createConversation();
```
przemvs: Hmm.. Is it still conversationId?

src/script/calling/CallingRepository.test.ts
```diff
@@ -476,12 +506,12 @@ describe.skip('E2E audio call', () => {
...
         .catch(done.fail);
     };
-    client.startCall(conversationId, CONV_TYPE.ONEONONE, CALL_TYPE.NORMAL).catch(done.fail);
+    client.startCall(conversationId, CALL_TYPE.NORMAL).catch(done.fail);
   });
 
   it('answers an incoming call and connect with the remote peer', done => {
     onCallClosed = done;
-    const conversationId = createConversationId();
+    const conversationId = createConversation();
```
przemvs: Is it still conversationId?

src/script/calling/CallingRepository.test.ts
atomrc: ah true, we should get the `qualifiedId` from it, good catch

src/script/calling/callingSubscriptionsHandler.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const store = new Map<string, (() => void)[]>();
+
+/**
+ * will keep track of all the subscriptions that are made for a given call
+ * @param conversationId the conversation in which the call is happening
+ * @param unsubscribe the function to call to teardown the subscription once the call is terminated
+ */
+const addCall = (conversationId: QualifiedId, unsubscribe: () => void) => {
+  const serialisedId = serializeQualifiedId(conversationId);
```
przemvs: I think there should be serializedId instead of serialisedId ?

PatrykBuniX: I believe its 🇺🇸 vs 🇬🇧 thingy haha

src/script/calling/mlsConference.ts
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+    return {
+      userid: `${parentMember.userId}@${parentMember.domain}`,
+      clientid: parentMember.clientId,
+      in_subconv: isSubconversationMember,
+    };
+  });
+
+  return members;
```
przemvs: WDYT?

```suggestion
  return memberIds.map(parentMember => {
    const isSubconversationMember = !!subconversation.members.find(
      ({user_id, client_id}) => user_id === parentMember.userId && client_id === parentMember.clientId,
    );

    return {
      userid: `${parentMember.userId}@${parentMember.domain}`,
      clientid: parentMember.clientId,
      in_subconv: isSubconversationMember,
    };
  });
```

src/script/calling/mlsConference.ts
```diff
@@ -0,0 +1,101 @@
...
+  members: SubconversationEpochInfoMember[];
+  epoch: number;
+  secretKey: string;
+  keyLength: number;
+}> => {
+  const members = await generateSubconversationMembers({mlsService}, conversationId, subconversation);
+
+  const epoch = Number(await mlsService.getEpoch(subconversation.group_id));
+
+  const keyLength = 32;
```
przemvs: It must be in this function? Can't be in constants/above this fn ?

src/script/calling/callingSubscriptionsHandler.ts
atomrc: ahaha it's british english VS us english 😛 
But you're right we should harmonize this :)

src/script/calling/mlsConference.ts
atomrc: good call 👍

src/script/view_model/CallingViewModel.ts
```diff
@@ -163,13 +207,39 @@ export class CallingViewModel {
       return !!this.callState.joinedCall();
     };
 
-    this.callingRepository.onIncomingCall((call: Call) => {
+    const updateEpochInfo = async (conversationId: QualifiedId) => {
+      const subconversation = await this.mlsService.getConferenceSubconversation(conversationId);
```
przemvs, line 211: Maybe move it after !isMLSConversationEstablished ?

przemvs: We need to do another call for it if we don't reuse it before checking isMLSConversationEstablished?

PatrykBuniX: We're using it in the actual check 
```ts
const isMLSConversationEstablished = await this.mlsService.conversationExists(subconversation.group_id);
```

atomrc: Actually I wonder if we could cache this groupId somewhere instead of requesting it. I believe we have a mechanism for this already, let me see

atomrc: Ok, I was digging into it a little, I think this is a topic for a further PR about optimization. 
Let's have this one out already and make it faster later on :)

src/script/calling/CallingRepository.ts
```diff
@@ -1088,7 +1151,7 @@ export class CallingRepository {
       };
     }
 
-    this.sendCallingMessage(conversationId, payload, options).catch(error => {
+    this.sendCallingMessage(conversationId, payload, options, myClientsOnly === 1).catch(error => {
```
przemvs, line 1154: What this 1 means ? Maybe move it to enum ? 🤔

przemvs (approved):
LGTM, 1 comment to investigate ;)

src/script/calling/CallingRepository.ts
atomrc: 1 is a boolean here. It's boolean sent from web assembly. 1 mean `true` and 0 `off`. I don't think we need an enum for this

PR #14444, opened 2023-01-04
fix: 2fa entropy bug [SQSERVICES-1709]

tlebon:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Fix an issue where immediate login causes the webapp to crash if user is required to enter 2fa and entropy.

### Causes (Optional)
as 2fa is only known upon login, attempting an immediate login (ie. without password/email) breaks the flow when also adding entropy as it thinks the user is already logged in and doesn't try to catch an error. 

### Solutions

try to initialize the client before adding entropy to know if it is able to be logged in or not. 

[SQPIT-764]: https://wearezeta.atlassian.net/browse/SQPIT-764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/script/auth/page/Login.tsx
+++ b/src/script/auth/page/Login.tsx
@@ -161,6 +161,8 @@ const Login = ({
   const immediateLogin = async () => {
     try {
       await doInit({isImmediateLogin: true, shouldValidateLocalClient: false});
+      // Try to initialize to check if the client needs 2fa before checking for entropyData
+      await doInitializeClient(ClientType.PERMANENT, undefined, undefined, undefined);
       const entropyData = await getEntropy?.();
       await doInitializeClient(ClientType.PERMANENT, undefined, undefined, entropyData);
       return history.push(ROUTE.HISTORY_INFO);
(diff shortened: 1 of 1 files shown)


src/script/auth/page/Login.tsx
```diff
@@ -161,8 +161,12 @@ const Login = ({
   const immediateLogin = async () => {
     try {
       await doInit({isImmediateLogin: true, shouldValidateLocalClient: false});
+      // Try to initialize to check if the client needs 2fa before checking for entropyData
+      await doInitializeClient(ClientType.PERMANENT, undefined, undefined, undefined);
```
atomrc, line 165: Hummm this feels wrong 🤔  I guess you could end-up creating a device that has no-custom entropy (while it's needed)

tlebon: could you? it should recreate it if its needed no?

tlebon:
this doesn't seem to be solvable feasibly. I am closing this PR for now.

PR #14446, opened 2023-01-05
feat: answer mls conference call [FS-1285]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1285" title="FS-1285" target="_blank">TaskFS-1285</a>  [Web] Join an already ongoing conference
  </td></table>
  <br />
 

Spec doc: https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/641171479/Conferencing+in+MLS+Subconversations#Join-a-conference-call

- joins mls subconversation when answering mls conference call (`wcall_answer`). 
- `requestClients` avs callback
- `requestNewEpoch` avs callback
- listening for epoch change (`mlsService.on('newEpoch')`)

Needs: 
- https://github.com/wireapp/wire-web-packages/pull/4748
- https://github.com/wireapp/wire-web-packages/pull/4811


src/script/view_model/ListViewModel.ts
```diff
@@ -160,7 +160,7 @@ export class ListViewModel {
         },
       });
     } else {
-      this.callingRepository.answerCall(call);
+      this.callingViewModel.callActions.answer(call);
```
PatrykBuniX, line 163: make sure we go through mls-related flow

src/script/view_model/CallingViewModel.ts
```diff
@@ -150,6 +157,44 @@ export class CallingViewModel {
       });
     };
 
+    const joinOngoingCall = async (call: Call) => {
```
PatrykBuniX: Was also considering this kind of api:
```ts
const joinOngoingCall = async (call: Call, onJoin?: (call: Call) => void) => {
  //do subconversation related stuff and update epoch info (only for mls conference)
  
  onJoin?.(call)
}
```

We could then call this function like: `joinOngoingCall(call, this.callingRepository.answerCall)`, not sure how important the order is, because in case of incoming call, the spec says we should call `wcall_incoming` first before joining subconversation.

src/script/view_model/CallingViewModel.ts
```diff
@@ -139,9 +139,16 @@ export class CallingViewModel {
...
-      const keyData = conversation.isUsingMLSProtocol
+      const subconversationData = conversation.isUsingMLSProtocol
         ? await core.service!.mls.joinConferenceSubconversation(conversation)
         : undefined;
+
+      const keyData = subconversationData && {
+        epoch: subconversationData.subconversation.epoch,
+        keyLength: subconversationData.keyLength,
+        secretKey: subconversationData.secretKey,
+      };
```
atomrc: 🤔  are you sure about this change. The `joinConferenceSubconversation` should return this exact same payload, doesn't it? 
image

src/script/view_model/ListViewModel.ts
atomrc: I guess it's kind of stupid that we still call this class `viewModel` now that it's not tied to a view anymore 😅

src/script/view_model/CallingViewModel.ts
PatrykBuniX: I've updated it on my branch: https://github.com/wireapp/wire-web-packages/pull/4748/files

src/script/view_model/CallingViewModel.ts
```diff
@@ -148,23 +156,138 @@ export class CallingViewModel {
...
+      const forwardNewEpoch = async ({groupId, epoch}: {groupId: string; epoch: number}) => {
+        const subconversation = await mlsService.getConferenceSubconversation(conversationId);
+
+        if (groupId !== subconversation.group_id) {
+          return;
+        }
+
+        const {keyLength, secretKey, members} = await getSubconversationEpochInfo(conversationId, subconversation);
+        onEpochUpdate({epoch: Number(epoch), keyLength, secretKey, members});
+      };
```
PatrykBuniX: This is pretty important change - we need to make sure we fetch fresh subconversation every time we receive epoch change event to get the fresh list of its members. Previously we were using closed (closure) value of subconversation fetched above `forwardNewEpoch` function definition.

src/script/calling/CallingRepository.ts
```diff
@@ -858,6 +872,23 @@ export class CallingRepository {
     }
   }
 
+  setEpochInfo(conversationId: QualifiedId, subconversationData: SubconversationData, members: any[]) {
```
atomrc: Could we try to create a cute little type for `members` 👼

PatrykBuniX: Good catch, missed this one 👍

src/script/view_model/CallingViewModel.ts
```diff
@@ -148,23 +156,138 @@ export class CallingViewModel {
...
+
+      const epoch = Number(await mlsService.getEpoch(subconversation.group_id));
+
+      const keyLength = 32;
+      const secretKey = await mlsService.exportSecretKey(subconversation.group_id, keyLength);
+
+      return {members, epoch, keyLength, secretKey};
+    };
+
+    const subscribeToEpochUpdates = async (
```
atomrc: Maybe we could extract all the mls-related functions to another module in order not to bloat the CallingViewModel with those details. WDYT?

PatrykBuniX: I like it, let me try excluding it 👌

PatrykBuniX: done 👌

atomrc (approved):

PR #14451, opened 2023-01-05
fix: set default font weight to 400 in dark mode [ACC-290]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-290" title="ACC-290" target="_blank">BugACC-290</a>  [Web] Font used in Dark Mode is incorrect (light weight)
  </td></table>
  <br />
 

### Change
- when no font weight is specified, the default should be 300 in light mode and 400 in dark mode

### DIFF
--- a/src/style/foundation/base.less
+++ b/src/style/foundation/base.less
@@ -48,10 +48,14 @@ body {
   color: var(--main-color);
   font-family: @font-family-base;
   font-size: @font-size-sm;
-  font-weight: 300;
+  font-weight: @font-weight-light;
   overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


PatrykBuniX (approved):

src/style/foundation/base.less
```diff
@@ -48,10 +48,14 @@ body {
   color: var(--main-color);
   font-family: @font-family-base;
   font-size: @font-size-sm;
-  font-weight: 300;
+  font-weight: @font-weight-light;
```
V-Gira, line 51: image

PR #14470, opened 2023-01-09
runfix: Add error log on core.init

atomrc:
Add error logs when `core.init` call fails. 
It would previously swallow the error and throw a new type of error, so it's not possible to go back to the original error.

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -361,6 +361,8 @@ export class App {
       try {
         await this.core.init(clientType);
       } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : error;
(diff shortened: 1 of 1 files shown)


src/script/main/app.ts
```diff
@@ -361,6 +361,8 @@ export class App {
       try {
         await this.core.init(clientType);
       } catch (error) {
+        const errorMessage = error instanceof Error ? error.message : error;
+        console.error(`Error when initializing core: "${errorMessage}"`, error);
```
PatrykBuniX: Shouldn't we go with `logger.error` instead?

atomrc: yes, we should 😅

PatrykBuniX (approved):

PR #14471, opened 2023-01-09
fix: Typing indicator announcement with assertive mode (ACC-304)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-304" title="ACC-304" target="_blank">StoryACC-304</a>  Screen Reader announcement for Typing Indicator
  </td></table>
  <br />


V-Gira (approved):
Let's hope it's not as annoying as it sounds it would be :D

przemvs (approved):

PR #14475, opened 2023-01-10
feat: Auto detect light & dark theme (ACC-304)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-304" title="ACC-304" target="_blank">StoryACC-304</a>  Screen Reader announcement for Typing Indicator
  </td></table>
  <br />
 

Solution until we fully implement Auto/light/dark in preferences


src/script/hooks/useAutomaticTheme.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const THEMES = {
+  DARK: 'dark',
+  DEFAULT: 'default',
+};
+
+export const useSystemTheme = () => {
+  const isSystemDarkMode = useMatchMedia('(prefers-color-scheme: dark)');
+  return isSystemDarkMode;
+};
```
przemvs: We need this hook? It's used only inside useAutomaticHook.

src/script/hooks/useAutomaticTheme.ts
```diff
@@ -0,0 +1,74 @@
...
+    amplify.subscribe(WebAppEvents.PROPERTIES.UPDATE.INTERFACE.THEME, setTheme);
+    amplify.subscribe(WebAppEvents.PROPERTIES.UPDATED, (properties: WebappProperties) =>
+      setTheme(properties.settings.interface.theme),
+    );
+  }, []);
+
+  const isSystemDarkMode = useSystemTheme();
+
+  useEffect(() => {
+    const nextTheme = (isSystemDarkMode ? THEMES.DARK : THEMES.DEFAULT) as ThemeType;
```
przemvs, line 66:
```suggestion
    const nextTheme: ThemeType = (isSystemDarkMode ? THEMES.DARK : THEMES.DEFAULT);
```

This one doesn';t work ? 🤔

src/script/hooks/useAutomaticTheme.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import {useMatchMedia} from '@wireapp/react-ui-kit';
+import {WebAppEvents} from '@wireapp/webapp-events';
+
+import {PropertiesRepository} from '../properties/PropertiesRepository';
+import {PROPERTIES_TYPE} from '../properties/PropertiesType';
+
+type ThemeType = 'dark' | 'default';
+
+export const updateTheme = (newTheme: ThemeType) => {
```
przemvs: We need to export this updateTheme ? 🤔

src/script/hooks/useAutomaticTheme.ts
V-Gira: Also, doesn't this line force the user to use system settings, leaving them stuck on light mode if the system doesn't define a preference?

thisisamir98: image

src/script/hooks/useAutomaticTheme.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const THEMES_CLASS_PREFIX = 'theme-';
+
+export const THEMES = {
+  DARK: 'dark',
+  DEFAULT: 'default',
+};
+
+const useAutomaticTheme = (propertiesRepository: PropertiesRepository): ThemeType => {
+  const [theme, setTheme] = useState<ThemeType>(propertiesRepository.getPreference(PROPERTIES_TYPE.INTERFACE.THEME));
```
przemvs, line 49: propertiesRepository.getPreference(PROPERTIES_TYPE.INTERFACE.THEME) - move this to variable, you have duplicated it :)

V-Gira (commented):
I think we should talk with design and insist to implement the radio instead of checkbox now. The way it ignores your user settings every time you start the app is going to be annoying fast

PR #14480, opened 2023-01-12
runfix: Restore identity resetting

atomrc:
Will leave a way for QA to break the identity in both Cryptobox and CoreCrypto env

### DIFF
--- a/src/script/util/DebugUtil.ts
+++ b/src/script/util/DebugUtil.ts
@@ -110,7 +110,8 @@ export class DebugUtil {
   }
 
   async resetIdentity(): Promise<void> {
-    this.core['coreCryptoClient'].wipe();
+    const proteusService = this.core.service!.proteus;
+    await proteusService['cryptoClient'].debugResetIdentity();
   }
 
   /** Used by QA test automation. */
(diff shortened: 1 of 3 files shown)


src/script/util/DebugUtil.ts
```diff
@@ -110,7 +110,8 @@ export class DebugUtil {
   }
 
   async resetIdentity(): Promise<void> {
-    this.core['coreCryptoClient'].wipe();
+    const proteusService = this.core.service!.proteus;
+    await proteusService['cryptoClient'].debugResetIdentity();
   }
```
tlebon, line 115: should there be a proteus and coreCrypto version of this function?

atomrc: That's the beauty of it, the `CryptoClient` is a generic interface that wrap both CoreCrypto and Cryptobox, no need for the webapp to know what is underneath :)

tlebon: ah but this reset identity function *is* only for proteus then. i suppose the MLS equivalent exists elsewhere.

atomrc: If coreCrypto is used (which is the case for MLS) MLS is handled as well in this method 👌

PR #14481, opened 2023-01-12
runfix: address shadow input scrolling issue

PatrykBuniX:
Fix for textarea and shadow input scroll positions not being in sync.

This is temporary hack before we migrate `InputBar` to custom content-editable div element.

### DIFF
--- a/src/script/components/InputBar/hooks/useResizeTarget.ts
+++ b/src/script/components/InputBar/hooks/useResizeTarget.ts
@@ -19,24 +19,34 @@
 
 import {useEffect, DependencyList} from 'react';
 
-const useResizeTarget = (element: HTMLElement | null, targetElement: HTMLElement | null, deps?: DependencyList) => {
+const alignScrollBars = (shadowInput: HTMLElement, textarea: HTMLElement) => {
+  const shouldForceScrollbar = textarea.clientHeight < textarea.scrollHeight;
+
+  shadowInput.style.overflowY = shouldForceScrollbar ? 'scroll' : 'hidden';
+  textarea.style.overflowY = shouldForceScrollbar ? 'scroll' : 'hidden';
+};
+
+const alignHeights = (shadowInput: HTMLElement, textarea: HTMLElement) => {
+  const {offsetHeight: shadowInputHeight, scrollHeight: shadowInputScrollHeight} = shadowInput;
+  const {offsetHeight: textAreaOffsetHeight} = textarea;
+
+  if (shadowInputHeight !== textAreaOffsetHeight) {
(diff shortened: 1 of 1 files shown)


src/script/components/InputBar/hooks/useResizeTarget.ts
```diff
@@ -19,19 +19,34 @@
 
 import {useEffect, DependencyList} from 'react';
 
-const useResizeTarget = (element: HTMLElement | null, targetElement: HTMLElement | null, deps?: DependencyList) => {
+const alignScrollBars = (shadowInput: HTMLElement, textarea: HTMLElement) => {
+  const shouldForceScrollbar = textarea.clientHeight < textarea.scrollHeight;
+
+  shadowInput.style.overflowY = shouldForceScrollbar ? 'scroll' : 'hidden';
+  textarea.style.overflowY = shouldForceScrollbar ? 'scroll' : 'hidden';
```
PatrykBuniX, line 31: According to [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight) `scrollHeight` always shows the height of an element's content, including content not visible on the screen due to overflow. Using `scroll` and `hidden` values, we're making sure that scroll behaves the same way on both elements (`overflow: auto` behaviour differs on `textarea` and `div`).

PatrykBuniX:
Tested on Firefox and Chrome, seems to be working fine.

przemvs (approved):

src/script/components/InputBar/hooks/useResizeTarget.ts
atomrc: Please add this as a comment in the code as well, that would help for future understanding :)

PatrykBuniX: added 👌

src/script/components/InputBar/hooks/useResizeTarget.ts
```diff
@@ -19,19 +19,39 @@
 
 import {useEffect, DependencyList} from 'react';
 
-const useResizeTarget = (element: HTMLElement | null, targetElement: HTMLElement | null, deps?: DependencyList) => {
+const alignScrollBars = (shadowInput: HTMLElement, textarea: HTMLElement) => {
+  /* According to documentation scrollHeight always shows the height of an element's content,
```
atomrc:
```suggestion
  /* According to documentation (https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight) scrollHeight always shows the height of an element's content,
```

atomrc (approved):

PR #14507, opened 2023-01-17
fix: Very long name causes misaligned system message

przemvs:
Before:
image
* Also another system messages get cut.

Now:
image
In another views also fixed. :)

### DIFF
--- a/src/script/components/MessagesList/Message/VerificationMessage.tsx
+++ b/src/script/components/MessagesList/Message/VerificationMessage.tsx
@@ -63,7 +63,7 @@ const VerificationMessage: React.FC<VerificationMessageProps> = ({message}) => {
         <VerifiedIcon isVerified={isTypeVerified} />
       </div>
       <div
-        className="message-header-label"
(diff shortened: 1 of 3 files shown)


src/script/components/MessagesList/Message/VerificationMessage.tsx
```diff
@@ -102,7 +102,7 @@ const VerificationMessage: React.FC<VerificationMessageProps> = ({message}) => {
         )}
         {isTypeNewMember && (
           <>
-            <span className="ellipsis">{t('conversationDeviceNewPeopleJoined')}</span>
```
atomrc, line 105: Is this class still used somewhere? If not can we remove it?

V-Gira: It's still used in quite a lot of places, it applies overflow, text-overflow and whitespace behavior

V-Gira (approved):

atomrc (approved):

PR #14514, opened 2023-01-17
fix: message not found on quotes of your own messages [SQSERVICES-1834]

tlebon:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1834" title="SQSERVICES-1834" target="_blank">BugSQSERVICES-1834</a>  "You cannot see this message." in reply to own message
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

fix an issue with quotes of your own (edited) messages not appearing

### Causes (Optional)

reference to message id is changing, using replacing_message_id. also an issue with race condition for hashed quotes.

### DIFF
--- a/src/script/event/EventService.ts
+++ b/src/script/event/EventService.ts
@@ -149,7 +149,7 @@ export class EventService {
         .filter(
           record =>
             (record.id === eventId && record.conversation === conversationId) ||
-            (record.id === eventId && record.data.replacing_message_id === eventId),
+            (record.conversation === conversationId && record.data.replacing_message_id === eventId),
         )
         .sort(compareEventsById)
         .shift();
--- a/src/script/event/EventService.ts
+++ b/src/script/event/EventService.ts
@@ -127,21 +127,21 @@ export class EventService {
 
     try {
       if (this.storageService.db) {
-        let entry = await this.storageService.db
+        const entry = await this.storageService.db
           .table(StorageSchemata.OBJECT_STORE.EVENTS)
           .where('id')
           .equals(eventId)
           .filter(record => record.conversation === conversationId)
           .first();
-        if (!entry) {
-          entry = await this.storageService.db
-            .table(StorageSchemata.OBJECT_STORE.EVENTS)
-            .where('conversation')
-            .equals(conversationId)
-            .filter(item => item.data.replacing_message_id === eventId)
-            .first();
+        if (entry) {
+          return entry;
         }
-        return entry;
+        return await this.storageService.db
+          .table(StorageSchemata.OBJECT_STORE.EVENTS)
+          .where('conversation')
+          .equals(conversationId)
+          .filter(item => item.data.replacing_message_id === eventId)
+          .first();
       }
 
       const records = (await this.storageService.getAll(StorageSchemata.OBJECT_STORE.EVENTS)) as EventRecord[];
--- a/src/script/event/EventService.ts
+++ b/src/script/event/EventService.ts
@@ -127,18 +127,30 @@ export class EventService {
 
     try {
       if (this.storageService.db) {
-        const entry = await this.storageService.db
+        let entry = await this.storageService.db
           .table(StorageSchemata.OBJECT_STORE.EVENTS)
           .where('id')
           .equals(eventId)
           .filter(record => record.conversation === conversationId)
           .first();
+        if (!entry) {
+          entry = await this.storageService.db
+            .table(StorageSchemata.OBJECT_STORE.EVENTS)
+            .where('conversation')
+            .equals(conversationId)
+            .filter(item => item.data.replacing_message_id === eventId)
+            .first();
+        }
         return entry;
       }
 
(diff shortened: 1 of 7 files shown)


src/script/event/EventService.ts
```diff
@@ -127,18 +127,30 @@ export class EventService {
...
       }
 
       const records = (await this.storageService.getAll(StorageSchemata.OBJECT_STORE.EVENTS)) as EventRecord[];
       return records
-        .filter(record => record.id === eventId && record.conversation === conversationId)
+        .filter(
+          record =>
+            (record.id === eventId && record.conversation === conversationId) ||
+            (record.id === eventId && record.data.replacing_message_id === eventId),
+        )
```
tlebon: not sure how important this is.

src/script/notification/NotificationRepository.ts
```diff
@@ -259,7 +259,9 @@ export class NotificationRepository {
             const messageInfo = messageId
```
tlebon, line 259: all the edits in this file are just improving some logger [object Object] issues

src/script/event/preprocessor/QuotedMessageMiddleware.ts
```diff
@@ -117,24 +116,10 @@ export class QuotedMessageMiddleware {
       return {...event, data: decoratedData};
     }
 
-    const quotedMessageHash = new Uint8Array(quote.quotedMessageSha256).buffer;
-
-    const isValid = await MessageHasher.validateHash(quotedMessage, quotedMessageHash);
-    let quoteData;
-
-    if (!isValid) {
```
tlebon, line 125: confirmed with @comawill this is fine not confirming hashes if the message is found in db

src/script/event/EventService.ts
```diff
@@ -127,18 +127,30 @@ export class EventService {
 
     try {
       if (this.storageService.db) {
```
atomrc: in which case do we not have a `db`? Did you find a reason why we have those 2 branches?

tlebon: i could never get that branch to go off, not sure tbh

atomrc: oohhh it those experiment where we have different DB engines under the hood 😠 
image

atomrc: I guess we need to handle both scenarios 😬

src/script/event/EventService.ts
```diff
@@ -127,18 +127,30 @@ export class EventService {
...
           .first();
+        if (!entry) {
+          entry = await this.storageService.db
+            .table(StorageSchemata.OBJECT_STORE.EVENTS)
+            .where('conversation')
+            .equals(conversationId)
+            .filter(item => item.data.replacing_message_id === eventId)
+            .first();
+        }
         return entry;
```
atomrc:
```suggestion
        if (entry) {
          return entry;
        }
        return await this.storageService.db
          .table(StorageSchemata.OBJECT_STORE.EVENTS)
          .where('conversation')
          .equals(conversationId)
          .filter(item => item.data.replacing_message_id === eventId)
          .first();
```

atomrc: This way you can avoid the `let` on L130. WDYT?

src/script/event/EventService.ts
atomrc: It's for when we don't have a dexie DB under the hood (not sure if we have valid scenarios where we have a different engine... 🤔 ). But let's keep it there 👍

test/unit_tests/event/preprocessor/QuotedMessageMiddlewareSpec.js
```diff
@@ -81,44 +81,6 @@ describe('QuotedMessageMiddleware', () => {
...
     });
 
-    it('adds an error if hashes do not match', async () => {
-      const expectedError = {
-        type: QuoteEntity.ERROR.INVALID_HASH,
-      };
-
-      const quotedMessage = {
-        data: {
-          content: 'pas salut',
```
V-Gira, line 91: 'salut'

Really wonder who wrote this test :thinking:

atomrc: ahaha guess who implemented those tests 😆

src/script/event/EventService.ts
tlebon: should we remove it?

tlebon: okay then

src/script/event/EventService.ts
atomrc: For reference, it was introduced here with temporary guest devices https://github.com/wireapp/wire-webapp/pull/7086/files#diff-6ab223d7041cd0190d856a1ee68a04f4c60175878ff0fcc77810a789569eeff7

atomrc (approved):

PR #14549, opened 2023-01-24
test: Fix React warning

atomrc:
Fixes this error 

image


src/script/auth/page/CheckPassword.tsx
```diff
@@ -153,7 +153,6 @@ const CheckPasswordComponent = ({
                   formNoValidate
                   onClick={handleLogin}
                   disabled={isFetching || !password}
-                  showLoading={isFetching}
```
atomrc, line 156: This will actually never render with the value `true` since it's already caught in the parent condition (see L147)

V-Gira (approved):

PR #14552, opened 2023-01-26
feat: unsubscribe from mls service events after call ends

PatrykBuniX:
Unsubscribe from `MLSService` events after call ends.

### DIFF
--- a/src/script/view_model/CallingViewModel.ts
+++ b/src/script/view_model/CallingViewModel.ts
@@ -91,6 +91,7 @@ declare global {
 
 export class CallingViewModel {
   readonly activeCalls: ko.PureComputed<Call[]>;
+  readonly callSubscriptions: (() => void)[] = [];
   readonly callActions: CallActions;
   readonly isSelfVerified: ko.Computed<boolean>;
   readonly activeCallViewTab: ko.Observable<string>;
@@ -181,7 +182,7 @@ export class CallingViewModel {
         secretKey: string;
         keyLength: number;
       }) => void,
-    ): Promise<void> => {
+    ): Promise<() => void> => {
       const mlsService = core.service?.mls;
       if (!mlsService) {
         throw new Error('mls service was not initialised');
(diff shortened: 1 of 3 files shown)


src/script/view_model/CallingViewModel.ts
```diff
@@ -251,23 +284,22 @@ export class CallingViewModel {
...
 
-      //const {epoch, keyLength, secretKey, members} = await subscribeToEpochUpdates(conversationId, subconversation);
-
-      //this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
+      const {epoch, keyLength, secretKey, members} = await getSubconversationEpochInfo(conversationId, subconversation);
+      this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
     });
 
+    //once we leave the call, we unsubscribe from all the events we've subscribed to during the call
+    this.callingRepository.onLeaveCall(() => this.callSubscriptions.forEach(unsubscribe => unsubscribe()));
```
atomrc: We would probably need to filter the subscription to the call that is actually left. 
We currently do not implement multi ongoing calls, but we should make sure that  this is not going to cause trouble in the future (in case mulit ongoing calls are implemented)

atomrc: Maybe a `Map` would be better suited for this. 

```js
this.callingRepository.onLeaveCall((conversationId) => {
  ongoingSubscription.get(convId)?.();
  ongoingSubscription.delete(convId)
});
```

WDYT?

src/script/view_model/CallingViewModel.ts
```diff
@@ -251,23 +284,22 @@ export class CallingViewModel {
...
+      const subconversation = await mlsService.getConferenceSubconversation(conversationId);
 
-      //const {epoch, keyLength, secretKey, members} = await subscribeToEpochUpdates(conversationId, subconversation);
-
-      //this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
+      const {epoch, keyLength, secretKey, members} = await getSubconversationEpochInfo(conversationId, subconversation);
+      this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
     });
 
+    //once we leave the call, we unsubscribe from all the events we've subscribed to during the call
```
atomrc:
```suggestion
    //once we leave a call, we unsubscribe from all the events we've subscribed to during this call
```

Since we are in a global context, all the calls will go through this same handler :)

atomrc (approved):
Looks pretty neat 😎

PR #14553, opened 2023-01-26
refactor: Make CallingRepository a little more stateless

atomrc:
This will move the view-related logic from the calling repository to the view model (where it should have belonged in the first place). 

This touches those 2 scenarios:

Answering a call that is ringing when a call is already ongoing in another conversation:

https://user-images.githubusercontent.com/1090716/214856074-47028a4d-f4ef-4ed6-b04c-091bfb513c10.mov

Starting a call while another one is already ongoing:

https://user-images.githubusercontent.com/1090716/214856102-41a222fa-176b-46d8-8440-346b893867a2.mov

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -849,18 +849,14 @@ export class CallingRepository {
     this.wCall?.reject(this.wUser, this.serializeQualifiedId(conversationId));
   }
 
-  changeCallPage(newPage: number, call: Call): void {
+  changeCallPage(call: Call, newPage: number): void {
     call.currentPage(newPage);
     if (!this.callState.isSpeakersViewActive()) {
-      this.requestCurrentPageVideoStreams();
+      this.requestCurrentPageVideoStreams(call);
     }
   }
 
-  requestCurrentPageVideoStreams(): void {
-    const call = this.callState.joinedCall();
-    if (!call) {
-      return;
-    }
+  requestCurrentPageVideoStreams(call: Call): void {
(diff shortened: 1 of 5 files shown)


src/script/calling/CallingRepository.ts
```diff
@@ -692,7 +692,6 @@ export class CallingRepository {
     const convId = this.serializeQualifiedId(conversationId);
     this.logger.log(`Starting a call of type "${callType}" in conversation ID "${convId}"...`);
     try {
-      await this.checkConcurrentJoinedCall(conversationId, CALL_STATE.OUTGOING);
```
atomrc, line 695: It's not the responsibility of the CallingRepo to check if there is a call ongoing. This has been moved to the view layer (see `CallingViewModel`)

src/script/calling/CallingRepository.ts
```diff
@@ -849,18 +847,14 @@ export class CallingRepository {
     this.wCall?.reject(this.wUser, this.serializeQualifiedId(conversationId));
   }
 
-  changeCallPage(newPage: number, call: Call): void {
+  changeCallPage(call: Call, newPage: number): void {
     call.currentPage(newPage);
     if (!this.callState.isSpeakersViewActive()) {
-      this.requestCurrentPageVideoStreams();
+      this.requestCurrentPageVideoStreams(call);
```
atomrc, line 853: We now inject the call that is concerned with the paging change in order for the CallingRepo not to have to check the state

thisisamir98 (approved):

PR #14584, opened 2023-01-30
refactor: exclude mls-related conference logic to separate module

PatrykBuniX:


src/script/calling/mlsConference.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ */
+
+import {Subconversation} from '@wireapp/api-client/lib/conversation/Subconversation';
+import {QualifiedId} from '@wireapp/api-client/lib/user';
+import {MLSService} from '@wireapp/core/lib/messagingProtocols/mls';
+
+import {SubconversationEpochInfoMember} from './CallingRepository';
+
+export const generateSubconversationMembers = async (
```
atomrc: I believe this one can stay private (no need to export it)

PatrykBuniX: correct

src/script/view_model/CallingViewModel.ts
```diff
@@ -304,7 +225,11 @@ export class CallingViewModel {
 
       const subconversation = await mlsService.getConferenceSubconversation(conversationId);
 
-      const {epoch, keyLength, secretKey, members} = await getSubconversationEpochInfo(conversationId, subconversation);
+      const {epoch, keyLength, secretKey, members} = await getSubconversationEpochInfo(
+        {mlsService},
+        conversationId,
+        subconversation,
+      );
       this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
```
atomrc, line 233: Isn't this logic also included in the `subscribeToEpochUpdates` call? I believe, when subscribed, it also get the current version, am I wrong?

atomrc (approved):
neat 😎

src/script/calling/mlsConference.ts
```diff
@@ -0,0 +1,101 @@
...
+    members: SubconversationEpochInfoMember[];
+    epoch: number;
+    secretKey: string;
+    keyLength: number;
+  }) => void,
+): Promise<() => void> => {
+  const initialSubconversation = await mlsService.joinConferenceSubconversation(conversationId);
+
+  const forwardNewEpoch = async ({groupId, epoch}: {groupId: string; epoch: number}) => {
+    const subconversation = await mlsService.getConferenceSubconversation(conversationId);
```
atomrc, line 82: We should probably move this after the `if` condition, in order not to fetch the conversation if not needed

PR #14586, opened 2023-01-30
fix: close sidebar instead of closing window when opening conversation [SQSERVICES-1882]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1882" title="SQSERVICES-1882" target="_blank">BugSQSERVICES-1882</a>  When coming from team settings, webapp crashes when opening a new group. 
  </td></table>
  <br />
 

[SQSERVICES-1882]

### Issue
- If a user comes from the team settings panel into the webapp (immediate-login), the webapp (the whole tab) will close when joining a group via search.
Peek 2023-01-30 13-21

### Solution
- replace `close` by `onClose` in the `openConversation` promise

#### What happened?
- Here's a quote from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/close) about the `Window.close()` method: 
> `The Window.close()` method closes the current window, or the window on which it was called.
This method can only be called on windows that were opened by a script using the [Window.open()](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) method.

- Opening the app from Team Setting's sidebar uses the `Window.open()` method
- The `Window.close()` method in `openConversation` would only result in an error if the app was opened normally, and close the tab if the app was opened through Team Setting (or any external link using the `Window.open()` method)

[SQSERVICES-1882]: https://wearezeta.atlassian.net/browse/SQSERVICES-1882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
+++ b/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
@@ -104,24 +104,22 @@ const StartUI: React.FC<StartUIProps> = ({
     if (peopleSearchResults.current) {
       const {contacts, groups} = peopleSearchResults.current;
       if (contacts.length > 0) {
-        await openContact(contacts[0]);
-        return;
+        return openContact(contacts[0]);
       }
       if (groups.length > 0) {
-        await openConversation(groups[0]);
+        return openConversation(groups[0]);
       }
     }
   };
 
   const openContact = async (user: User) => {
     const conversationEntity = await actions.getOrCreate1to1Conversation(user);
-    await actions.open1to1Conversation(conversationEntity);
+    return actions.open1to1Conversation(conversationEntity);
   };
 
   const openOther = async (user: User) => {
     if (user.isOutgoingRequest()) {
-      await openContact(user);
-      return;
+      return openContact(user);
     }
 
     showUserModal({domain: user.domain, id: user.id});
--- a/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
+++ b/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
@@ -117,12 +117,12 @@ const StartUI: React.FC<StartUIProps> = ({
     return actions.open1to1Conversation(conversationEntity);
   };
 
-  const openOther = async (user: User) => {
+  const openOther = (user: User) => {
     if (user.isOutgoingRequest()) {
       return openContact(user);
     }
 
-    showUserModal({domain: user.domain, id: user.id});
+    return showUserModal({domain: user.domain, id: user.id});
   };
 
   const openService = (service: ServiceEntity) => {
@@ -135,8 +135,9 @@ const StartUI: React.FC<StartUIProps> = ({
 
   const openInviteModal = () => showInviteModal({userState});
 
-  const openConversation = (conversation: Conversation): Promise<void> => {
-    return actions.openGroupConversation(conversation).then(onClose);
+  const openConversation = async (conversation: Conversation): Promise<void> => {
+    await actions.openGroupConversation(conversation);
+    close();
   };
 
   const before = (
--- a/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
+++ b/src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
@@ -100,27 +100,27 @@ const StartUI: React.FC<StartUIProps> = ({
 
   const peopleSearchResults = useRef<SearchResultsData | undefined>(undefined);
 
-  const openFirstConversation = (): void => {
+  const openFirstConversation = async (): Promise<void> => {
     if (peopleSearchResults.current) {
       const {contacts, groups} = peopleSearchResults.current;
       if (contacts.length > 0) {
-        openContact(contacts[0]);
+        await openContact(contacts[0]);
         return;
       }
       if (groups.length > 0) {
-        openConversation(groups[0]);
+        await openConversation(groups[0]);
       }
     }
   };
 
   const openContact = async (user: User) => {
     const conversationEntity = await actions.getOrCreate1to1Conversation(user);
-    actions.open1to1Conversation(conversationEntity);
+    await actions.open1to1Conversation(conversationEntity);
   };
 
-  const openOther = (user: User) => {
+  const openOther = async (user: User) => {
     if (user.isOutgoingRequest()) {
-      openContact(user);
+      await openContact(user);
       return;
     }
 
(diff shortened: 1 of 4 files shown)


src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
```diff
@@ -138,7 +138,7 @@ const StartUI: React.FC<StartUIProps> = ({
   const openInviteModal = () => showInviteModal({userState});
 
   const openConversation = (conversation: Conversation): Promise<void> => {
-    return actions.openGroupConversation(conversation).then(close);
+    return actions.openGroupConversation(conversation).then(onClose);
```
V-Gira: this alone solves the issue, the rest of the PR is some small refactoring

PatrykBuniX (approved):
Nice catch

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
```diff
@@ -100,27 +100,27 @@ const StartUI: React.FC<StartUIProps> = ({
...
   const peopleSearchResults = useRef<SearchResultsData | undefined>(undefined);
 
-  const openFirstConversation = (): void => {
+  const openFirstConversation = async (): Promise<void> => {
     if (peopleSearchResults.current) {
       const {contacts, groups} = peopleSearchResults.current;
       if (contacts.length > 0) {
-        openContact(contacts[0]);
+        await openContact(contacts[0]);
         return;
```
PatrykBuniX: you can just return an async function here -> `return openContact(contacts[0]);` (same thing with the ones below)

przemvs (approved):

src/script/page/LeftSidebar/panels/StartUI/components/groupList/GroupListItem.tsx
```diff
@@ -43,8 +43,9 @@ const GroupListItem: React.FC<GroupListItemProps> = ({click, group}) => {
   } = useKoSubscribableChildren(group, ['display_name', 'participating_user_ets', 'is1to1']);
 
   const onClick = () => {
+    const url = generateConversationUrl(group.qualifiedId);
     click(group);
-    navigate(generateConversationUrl(group.qualifiedId));
+    createNavigate(url);
```
atomrc: This change seems suspicious to me. You create something but never return it, this function is never consumed 🤔

src/script/view_model/ActionsViewModel.ts
```diff
@@ -110,8 +110,8 @@ export class ActionsViewModel {
     return new Promise(resolve => {
       PrimaryModal.show(PrimaryModal.type.CONFIRM, {
         primaryAction: {
-          action: () => {
-            this.connectionRepository.cancelRequest(userEntity, hideConversation, nextConversationEntity);
+          action: async () => {
+            return this.connectionRepository.cancelRequest(userEntity, hideConversation, nextConversationEntity);
             resolve();
```
atomrc: This is wrong, the `resolve` will never be called

src/script/view_model/ActionsViewModel.ts
```diff
@@ -203,8 +203,8 @@ export class ActionsViewModel {
       return new Promise(resolve => {
         PrimaryModal.show(PrimaryModal.type.CONFIRM, {
           primaryAction: {
-            action: () => {
-              this.messageRepository.deleteMessage(conversationEntity, messageEntity);
+            action: async () => {
+              return this.messageRepository.deleteMessage(conversationEntity, messageEntity);
               resolve();
```
atomrc: Same here, the `resolve` will not be called

atomrc (changes requested):
A few floating promises have been fixed... The wrong way :)

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
```diff
@@ -100,28 +100,26 @@ const StartUI: React.FC<StartUIProps> = ({
 
   const peopleSearchResults = useRef<SearchResultsData | undefined>(undefined);
 
-  const openFirstConversation = (): void => {
+  const openFirstConversation = async (): Promise<void> => {
```
atomrc, line 103: you actually don't need to mark this one async. 

The rule of thumb is: you only need to put `async` if you need `await` in the body of the function. 

Else a simple `: Promise<...>` signature is enough :) 

```suggestion
  const openFirstConversation = (): Promise<void> => {
```

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
```diff
@@ -100,28 +100,26 @@ const StartUI: React.FC<StartUIProps> = ({
...
   };
 
   const openContact = async (user: User) => {
     const conversationEntity = await actions.getOrCreate1to1Conversation(user);
-    actions.open1to1Conversation(conversationEntity);
+    return actions.open1to1Conversation(conversationEntity);
   };
 
-  const openOther = (user: User) => {
+  const openOther = async (user: User) => {
```
atomrc: see https://github.com/wireapp/wire-webapp/pull/14586/files#r1090618504

src/script/view_model/ActionsViewModel.ts
```diff
@@ -330,9 +330,9 @@ export class ActionsViewModel {
     return this.openConversation(conversationEntity);
   };
 
-  private readonly openConversation = (conversationEntity: Conversation): void => {
+  private readonly openConversation = async (conversationEntity: Conversation): Promise<void> => {
     if (conversationEntity.is_archived()) {
-      this.conversationRepository.unarchiveConversation(conversationEntity, true);
+      return this.conversationRepository.unarchiveConversation(conversationEntity, true);
```
atomrc: This changes the behavior (the amplify event will not be fired)

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
V-Gira: I think I'm missing something obvious :thinking: 
image

atomrc (changes requested):

src/script/page/LeftSidebar/panels/StartUI/StartUI.tsx
atomrc: I guess you can just remove the return type at all, it's just extra maintenance here :) 
TypeScript is going to infer it for you ;) 

```suggestion
  const openFirstConversation = () => {
```

atomrc (approved):

PR #14587, opened 2023-01-30
feat: respect my_clients_only param in sendh callback

PatrykBuniX:


atomrc (approved):

src/script/calling/CallingRepository.ts
```diff
@@ -1129,6 +1129,9 @@ export class CallingRepository {
     targets: string | null,
     _unused: string | null,
     payload: string,
+    _len: number,
+    _trans: number,
+    my_clients_only: number,
```
atomrc:
```suggestion
    myClientsOnly: number,
```

Why not?

src/script/conversation/MessageRepository.ts
```diff
@@ -107,6 +107,7 @@ export interface MessageSendingOptions {
   /** Send native push notification for message. Default is `true`. */
   nativePush?: boolean;
   recipients?: QualifiedId[] | QualifiedUserClients | UserClients;
+  myClientsOnly?: boolean;
```
atomrc: I don't think this belongs here, the `MessageRepository` doesn't really care about this value

atomrc: I think instead it should only be an extra param of the `sendCallingMessage` of the calling repo

PatrykBuniX: 👍

atomrc (approved):

PR #14592, opened 2023-01-31
feat: user leaves mls conference call

PatrykBuniX:
Users leaves mls conference call.

Needs
- https://github.com/wireapp/wire-web-packages/pull/4834


src/script/view_model/CallingViewModel.mocks.ts
```diff
@@ -28,17 +28,20 @@ import {CallingViewModel} from './CallingViewModel';
...
   startCall: jest.fn(),
   answerCall: jest.fn(),
-  leaveCall: jest.fn(),
   onIncomingCall: jest.fn(),
   onRequestClientsCallback: jest.fn(),
   onRequestNewEpochCallback: jest.fn(),
   onCallParticipantChangedCallback: jest.fn(),
-  onLeaveCall: jest.fn(),
+  onCallClosed: jest.fn().mockImplementation(callback => (callClosedCallback = callback)),
+  leaveCall: jest.fn().mockImplementation(conversationId => callClosedCallback(conversationId)),
```
PatrykBuniX, line 44: As we're mocking `CallingRepository` there's no way we could trigger `callClosed` callback from AVS. For the sake of test, we mock `leaveCall` to call our callback immediately, so we're able to test unsubscribing from `newEpoch` mls service event.

src/script/view_model/CallingViewModel.test.ts
```diff
@@ -203,6 +207,28 @@ describe('CallingViewModel', () => {
...
+        callingViewModel.callingRepository.leaveCall(conversationId, LEAVE_CALL_REASON.MANUAL_LEAVE_BY_UI_CLICK);
+      });
+
+      const anotherEpochNumber = 3;
+      callingViewModel.mlsService.emit('newEpoch', {
+        epoch: anotherEpochNumber,
+        groupId: mockSubGroupId,
+      });
+
+      await new Promise(r => setTimeout(r, 0));
```
PatrykBuniX, line 224: We want to wait for all the callback queue tasks to be executed so we know that the function **was not** called. Without waiting, test will succeed every time (even without unsubscribing to epoch changes) because the function was not called **yet**.

src/script/calling/CallingRepository.ts
```diff
@@ -1268,6 +1267,8 @@ export class CallingRepository {
       return;
     }
 
+    this.callClosedCallback(conversationId);
```
atomrc, line 1270: Oh this is called even when the user leaves on purpose. That's good refactoring then 👌

src/script/view_model/CallingViewModel.test.ts
atomrc: Can you add this as a comment in the code? That could be useful in the future :)

atomrc (approved):
One small suggestion, but looks pretty good. Nice job 😎

src/script/view_model/CallingViewModel.test.ts
PatrykBuniX: Sure, will do 😄

atomrc (approved):

PR #14596, opened 2023-02-01
chore: Improve app load logs

atomrc:
This removes some noisy logs when loading the app and fine tunes some others

### DIFF
--- a/src/script/telemetry/app_init/AppInitTimings.ts
+++ b/src/script/telemetry/app_init/AppInitTimings.ts
@@ -51,7 +51,7 @@ export class AppInitTimings {
     const appLoaded = this.timings[AppInitTimingsStep.APP_LOADED];
     const appLoadedInSeconds = appLoaded / TIME_IN_MILLIS.SECOND;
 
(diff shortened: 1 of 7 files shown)


V-Gira (approved):

PatrykBuniX (approved):

src/script/telemetry/app_init/AppInitTimings.ts
```diff
@@ -51,7 +51,7 @@ export class AppInitTimings {
     const appLoaded = this.timings[AppInitTimingsStep.APP_LOADED];
     const appLoadedInSeconds = appLoaded / TIME_IN_MILLIS.SECOND;
 
-    return (Math.floor(appLoadedInSeconds / AppInitTimings.CONFIG.BUCKET_SIZE) + 1) * AppInitTimings.CONFIG.BUCKET_SIZE;
```
V-Gira, line 54: I'm a bit curious about what was the big idea there

atomrc: The big idea was to over-engineer a piece of code that has very little value (it always say that the app loaded in less than 10s, we could just print the actual number of seconds)

PR #14597, opened 2023-02-01
refactor: add mls service getter

PatrykBuniX:


src/script/view_model/CallingViewModel.ts
```diff
@@ -103,7 +103,7 @@ export class CallingViewModel {
     private readonly conversationState = container.resolve(ConversationState),
     readonly callState = container.resolve(CallState),
     private readonly teamState = container.resolve(TeamState),
-    core = container.resolve(Core),
+    private readonly core = container.resolve(Core),
```
atomrc, line 106: humm I'm wondering why it was like this in the first place 🤔

atomrc (approved):

src/script/view_model/CallingViewModel.ts
PatrykBuniX: good question 💁‍♂️

PR #14600, opened 2023-02-01
fix: Make the call end in case of being unable to reach sft server (SQSERVICES-1884)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1884" title="SQSERVICES-1884" target="_blank">BugSQSERVICES-1884</a>  "Conecting..." message shows up when answering the call
  </td></table>
  <br />
 

when joining/starting a call we need to send a request to the SFT server, now if for whatever reason this request would fail we need to let the avs library know of that so it can try again or drop, to achieve this we need to call `wCall.sftResp` with status code of a `1000` to let avs know the request was failed.

https://github.com/wireapp/wire-webapp/pull/14600/files#diff-815c4a1834d94a723e8cd6eb18d5608e92dd62d3fb18809251711b9239baef96R1203-R1207


V-Gira (approved):

PatrykBuniX (approved):

src/script/calling/CallingRepository.ts
```diff
@@ -797,7 +797,7 @@ export class CallingRepository {
       if (newState === VIDEO_STATE.STOPPED) {
         selfParticipant.releaseVideoStream(true);
       } else {
-        this.warmupMediaStreams(call, false, true);
+        void this.warmupMediaStreams(call, false, true);
```
tlebon: why are we adding void now?

thisisamir98: image

without that we get this warning but there is no need to catch anything on this, all code inside warmupMediaStreams is in a try catch and has it's possible error handled 

image

tlebon: this feels slightly incorrect to me, as though we are missing the correct async/awaits on things and this will bite us later when we cant figure out why something isn't awaiting that should be

thisisamir98: well in other places that it’s called it is also not awaited and i'm not sure if it's a good idea to change that behavior 🤔 but I could remove the `void` so the warning would be around.

tlebon: id prefer to leave the warning. my sense is that it is a *correct* warning for the callingRepo and we should be doing something with all of these.

thisisamir98: removed void operators for now: https://github.com/wireapp/wire-webapp/pull/14600/commits/877a88c05f673c2c891cb49b72b174b77d1c3cd8

tlebon (approved):

PR #14630, opened 2023-02-06
fix: message not found

tlebon:
----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] contains a reference JIRA issue number like `SQPIT-764`
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [ ] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

This was reopened because of a bug found where some quotes were not appearing that caused the previous PR to be reverted. #14629 

The STR are:

user1 and user2 have a conversation
user1 has a created client but not loaded
user2 sends a message in a conversation with user1
user2 quotes the message they just sent
user1 load the webapp with the existing client
-> user1 cannot see the quote

### Causes (Optional)

TBD

### Solutions

TBD

PR #14632, opened 2023-02-06
feat: improve quote event loading [SQSERVICES-1834]

tlebon:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1834" title="SQSERVICES-1834" target="_blank">BugSQSERVICES-1834</a>  "You cannot see this message." in reply to own message
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Tries a different approach to solve the 'message not found' bug in #14630. That solution caused some unexpected side effects. 

### Causes (Optional)
As *loadEvent* is more generic than expected, it is causing some issues in various locations. 

### Solutions
The solution to this is to create other methods to be consumed in specific locations.

### DIFF
--- a/src/script/components/MessagesList/Message/MessageWrapper.tsx
+++ b/src/script/components/MessagesList/Message/MessageWrapper.tsx
@@ -78,7 +78,7 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
   isMsgElementsFocusable,
 }) => {
   const findMessage = (conversation: Conversation, messageId: string) => {
-    return messageRepository.getMessageInConversationById(conversation, messageId, true, true);
+    return messageRepository.getMessageInConversationByReplacementId(conversation, messageId);
   };
   const clickButton = (message: CompositeMessage, buttonId: string) => {
     if (message.selectedButtonId() !== buttonId && message.waitingButtonId() !== buttonId) {
--- a/src/script/components/MessagesList/Message/MessageWrapper.tsx
+++ b/src/script/components/MessagesList/Message/MessageWrapper.tsx
@@ -77,8 +77,11 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
   totalMessage,
   isMsgElementsFocusable,
 }) => {
-  const findMessage = (conversation: Conversation, messageId: string) => {
-    return messageRepository.getMessageInConversationByReplacementId(conversation, messageId);
+  const findMessage = async (conversation: Conversation, messageId: string) => {
+    const event =
+      (await messageRepository.getMessageInConversationById(conversation, messageId)) ||
+      (await messageRepository.getMessageInConversationByReplacementId(conversation, messageId));
+    return await messageRepository.ensureMessageSender(event);
   };
   const clickButton = (message: CompositeMessage, buttonId: string) => {
     if (message.selectedButtonId() !== buttonId && message.waitingButtonId() !== buttonId) {
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -921,7 +921,7 @@ export class MessageRepository {
       if (!message.user().isMe && !message.ephemeral_expires()) {
         throw new ConversationError(ConversationError.TYPE.WRONG_USER, ConversationError.MESSAGE.WRONG_USER);
       }
-      const userIds = options.targetedUsers || conversation.allUserEntities.map(user => user.qualifiedId);
+      const userIds = options.targetedUsers || conversation.allUserEntities.map(user => user!.qualifiedId);
       const payload = MessageBuilder.buildDeleteMessage({
         messageId: message.id,
       });
@@ -1173,19 +1173,10 @@ export class MessageRepository {
    *
    * @param conversation Conversation message belongs to
    * @param messageId ID of message
-   * @param skipConversationMessages Don't use message entity from conversation
-   * @param ensureUser Make sure message entity has a valid user
    * @returns Resolves with the message
    */
-  async getMessageInConversationById(
-    conversation: Conversation,
-    messageId: string,
-    skipConversationMessages = false,
-    ensureUser = false,
-  ): Promise<StoredContentMessage> {
-    const messageEntity =
-      !skipConversationMessages &&
-      (conversation.getMessageByReplacementId(messageId) || conversation.getMessage(messageId));
+  async getMessageInConversationById(conversation: Conversation, messageId: string): Promise<StoredContentMessage> {
+    const messageEntity = conversation.getMessage(messageId);
     const message =
       messageEntity ||
       (await this.eventService.loadEvent(conversation.id, messageId).then(event => {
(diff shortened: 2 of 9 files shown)


src/script/entity/Conversation.ts
```diff
@@ -968,6 +968,15 @@ export class Conversation {
...
   }
+  /**
+   * Get a message by its replacing message id. Useful if the message in question is an edit and has replaced the original message.
+   * Only lookup in the loaded message list which is a limited view of all the messages in DB.
+   *
+   * @param messageId ID of message to be retrieved
+   */
+  getMessageByReplacementId(messageId: string): Message | ContentMessage | MemberMessage | SystemMessage | undefined {
+    return this.messages().find(messageEntity => (messageEntity as ContentMessage)?.replacing_message_id === messageId);
+  }
```
przemvs: Maybe use **isContentMessage** instead of **as** ? 🤔

przemvs (approved):
Code LGTM, 1 comment ;)

src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -78,7 +78,7 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
   isMsgElementsFocusable,
 }) => {
   const findMessage = (conversation: Conversation, messageId: string) => {
-    return messageRepository.getMessageInConversationById(conversation, messageId, true, true);
+    return messageRepository.getMessageInConversationByReplacementId(conversation, messageId);
```
atomrc: Humm I think we also need to get the message by regular ID at this point. The replacing_id should only be a fallback in case we don't find the message by ID

src/script/conversation/MessageRepository.ts
```diff
@@ -1173,17 +1173,38 @@ export class MessageRepository {
    *
    * @param conversation Conversation message belongs to
    * @param messageId ID of message
-   * @param skipConversationMessages Don't use message entity from conversation
-   * @param ensureUser Make sure message entity has a valid user
    * @returns Resolves with the message
    */
-  async getMessageInConversationById(
+  async getMessageInConversationById(conversation: Conversation, messageId: string): Promise<StoredContentMessage> {
+    const messageEntity = conversation.getMessage(messageId);
```
atomrc, line 1179: I believe the default value of this method is **not** to look up the local state. This might be a regression here (since now it always lookup to local state)

src/script/conversation/MessageRepository.ts
```diff
@@ -1197,7 +1218,7 @@ export class MessageRepository {
       );
     }
 
-    if (ensureUser && message.from && !message.user().id) {
+    if (message.from && !message.user().id) {
```
atomrc, line 1219: Let's extract this logic to a different method and have a single responsibility for the `getMessageInConversationByReplacementId`. 
This way the consumer can do:

```js
const event = getMessageInConverstation(..) || getMessageInConversationByReplacementId()
const event = loadUsers(event);
```

src/script/event/EventService.ts
```diff
@@ -127,18 +127,52 @@ export class EventService {
...
           .equals(eventId)
           .filter(record => record.conversation === conversationId)
           .first();
-        return entry;
       }
 
       const records = (await this.storageService.getAll(StorageSchemata.OBJECT_STORE.EVENTS)) as EventRecord[];
       return records
-        .filter(record => record.id === eventId && record.conversation === conversationId)
+        .filter(record => record.conversation === conversationId && record.data?.replacing_message_id === eventId)
```
atomrc, line 175: Let's **only** lookup by replacing id at this point. The consumer is responsible for falling back to something else if it's not found

src/script/event/preprocessor/QuotedMessageMiddleware.ts
```diff
@@ -104,37 +103,27 @@ export class QuotedMessageMiddleware {
 
     const messageId = quote.quotedMessageId;
 
-    const quotedMessage = await this.eventService.loadEvent(event.conversation, messageId);
+    let quotedMessage = await this.eventService.loadEvent(event.conversation, messageId);
```
atomrc:
```suggestion
    let quotedMessage = await this.eventService.loadEvent(event.conversation, messageId) ?? await this.eventService.loadReplacingEvent(event.conversation, messageId);
```

How about this?

src/script/event/preprocessor/QuotedMessageMiddleware.ts
```diff
@@ -104,37 +103,27 @@ export class QuotedMessageMiddleware {
...
+
+        const decoratedData = {...event.data, quote: quoteData};
+        return {...event, data: decoratedData};
+      }
+      quotedMessage = replacedMessage;
     }
 
-    const quotedMessageHash = new Uint8Array(quote.quotedMessageSha256).buffer;
-
-    const isValid = await MessageHasher.validateHash(quotedMessage, quotedMessageHash);
```
atomrc, line 122: Can you check if the messageHasher could not be removed from the codebase at all?

atomrc (changes requested):
I think there are some parts that might have introduced behavior changes.

src/script/conversation/MessageRepository.ts
tlebon: the default value is false and this is negated. so the default case is always to look up.

src/script/components/MessagesList/Message/MessageWrapper.tsx
tlebon: i think the naming is confusing you here but it does indeed to either/or, not only replacingId

src/script/event/EventService.ts
tlebon: true

src/script/event/preprocessor/QuotedMessageMiddleware.ts
tlebon: i dont believe that is the case, this was something from the original testing, and already approved by security. in some cases the hash will simply not match. there's really no way around it if we want to solve the original 'message not found bug'

tlebon: if you send a hash of a different message it must fail, even if we see the replacing id being updated.

atomrc: Oh, it's still used for sending quotes, right?

src/script/event/EventService.ts
tlebon: in the other function on L140, we are also checking by conversation && then id. (This was also the previous implementation) Should we change that as well or leave both the same? 

I assume there was some logic to checking both, although the chances of an overlap here seem relatively astronomical to me. Still sort of unclear when this section of the function would be called anyway.

src/script/event/preprocessor/QuotedMessageMiddleware.ts
tlebon: yep

src/script/conversation/MessageRepository.ts
tlebon: i did something but i think based on your comment you would like something different

atomrc (approved):
Seems solid 😎

PR #14633, opened 2023-02-06
fix: Make sure app is fully loaded before showing the UI

atomrc:
Quite a few promises were not awaited when initially loading the application. 
This could make the app show the UI even before the message are fully loaded in memory. 

## Before

https://user-images.githubusercontent.com/1090716/216980913-f6be4bbb-bec4-4354-81fa-23442779029f.mov

## After

https://user-images.githubusercontent.com/1090716/216980956-3cb9e3f6-b12e-4ec9-9822-e1377afcb09d.mov

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2541,7 +2541,7 @@ export class ConversationRepository {
    * @param eventJson JSON data of 'conversation.member-update' event
    * @returns Resolves when the event was handled
    */
-  private onMemberUpdate(
+  private async onMemberUpdate(
     conversationEntity: Conversation,
     eventJson: Pick<ConversationMemberUpdateEvent, 'data' | 'from'> & {conversation?: string},
   ) {
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -54,7 +54,6 @@ import {getNextItem} from 'Util/ArrayUtil';
 import {allowsAllFiles, getFileExtensionOrName, isAllowedFile} from 'Util/FileTypeUtil';
 import {replaceLink, t} from 'Util/LocalizerUtil';
 import {getLogger, Logger} from 'Util/Logger';
-import {PromiseQueue} from 'Util/PromiseQueue';
 import {matchQualifiedIds} from 'Util/QualifiedId';
 import {
   compareTransliteration,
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -149,7 +148,6 @@ export class ConversationRepository {
   private readonly event_mapper: EventMapper;
   private readonly eventService: EventService;
   public leaveCall: (conversationId: QualifiedId, reason: LEAVE_CALL_REASON) => void;
-  private readonly receiving_queue: PromiseQueue;
```
atomrc, line 152: This receiving_queue didn't make sense for a while now (it's never paused anywhere, so there is no point queueing things here)

src/script/conversation/ConversationRepository.ts
```diff
@@ -802,7 +799,7 @@ export class ConversationRepository {
     const mapOfUserIds = conversationEntities.map(conversationEntity => conversationEntity.participating_user_ids());
     const userIds = flatten(mapOfUserIds);
     await this.userRepository.getUsersById(userIds);
-    conversationEntities.forEach(conversationEntity => this.fetchUsersAndEvents(conversationEntity));
+    await Promise.all(conversationEntities.map(conversationEntity => this.fetchUsersAndEvents(conversationEntity)));
```
atomrc, line 802: Here we need to await for all the conversation to be loaded before resolving (this was part of the bug)

src/script/conversation/ConversationRepository.ts
```diff
@@ -2978,10 +2965,10 @@ export class ConversationRepository {
    *
    * @param conversationEntity Conversation fetch events and users for
    */
-  private fetchUsersAndEvents(conversationEntity: Conversation) {
+  private async fetchUsersAndEvents(conversationEntity: Conversation) {
     if (!conversationEntity.is_loaded() && !conversationEntity.is_pending()) {
-      this.updateParticipatingUserEntities(conversationEntity);
-      this.getUnreadEvents(conversationEntity);
+      await this.updateParticipatingUserEntities(conversationEntity);
+      await this.getUnreadEvents(conversationEntity);
```
atomrc, line 2971: Here we load the user entities and the events from the DB into the in-memory conversation. But we do not wait for the operation to be done before returning (thus the glitch)

PatrykBuniX (approved):
Great investigation 💪

przemvs (approved):

tlebon (approved):

PR #14636, opened 2023-02-07
runfix: remove redundant set_epoch_info on req_clients callback

PatrykBuniX:
It appeared that updating epoch info (`set_epoch_info`) when avs requests new client list (`req_clients`) is a redundant call. With mls conferences we're sure that the client list is always up-to-date because we call `set_epoch_info` everytime membership (epoch number) changes. We always get a message when a new clients joins/leaves a conversation.

- [x] Waiting for a confirmation from avs team that we can safely ignore this callback for mls conferences


src/script/calling/CallingRepository.ts
```diff
@@ -332,24 +330,26 @@ export class CallingRepository {
...
-        data.map(clientid => ({clientid, userid: this.serializeQualifiedId(userId)})),
-      ),
-    );
+      const clients: Clients = flatten(
+        qualifiedClients.map(({data, userId}) =>
+          data.map(clientid => ({clientid, userid: this.serializeQualifiedId(userId)})),
+        ),
+      );
+
+      this.wCall?.setClientsForConv(
```
PatrykBuniX, line 346: This call is not needed for mls conference calls.

atomrc (approved):

PR #14637, opened 2023-02-07
refactor: Stricter types for EventRecord

atomrc:
This is an attempt to have stricter types of the payload we store in the DB. 
It's a first step towards unifying all the events we handle in the webapp and have strict checking on those


przemvs (approved):
Nice!

PatrykBuniX (approved):
Nice improvements 🚀

PR #14640, opened 2023-02-08
refactor: Narrow down event types

atomrc:
This will differentiate between:
- Incoming events (coming from backend or manually injected)
- Stored events (events that are in the DB that have been mapped previously)

### DIFF
--- a/src/script/event/EventRepository.test.ts
+++ b/src/script/event/EventRepository.test.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {CONVERSATION_EVENT, USER_EVENT} from '@wireapp/api-client/lib/event/';
+import {BackendEvent, CONVERSATION_EVENT, USER_EVENT} from '@wireapp/api-client/lib/event/';
 
 import {Asset as ProtobufAsset} from '@wireapp/protocol-messaging';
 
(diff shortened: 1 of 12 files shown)


przemvs (approved):

src/script/storage/record/EventRecord.ts
```diff
@@ -40,13 +42,22 @@ export interface AssetRecord {
 export type UserReactionMap = {[userId: string]: ReactionType};
 
 /** represents an event that was saved to the DB */
-type StoredEvent = {
+export type StoredEvent<T> = {
   primary_key: string;
   category: number;
   id: string;
+  /** if the message is ephemeral, that's the amount of time it should be displayed to the user */
+  ephemeral_expires?: number | boolean | string;
```
arjita-mitra, line 55: why ephemeral_expires can be 3 different types?

src/script/event/EventRepository.test.ts
```diff
@@ -575,7 +577,7 @@ describe('EventRepository', () => {
 
       return testFactory
         .event_repository!['processEvent'](updateStatusEvent, EventSource.NOTIFICATION_STREAM)
-        .then(updatedEvent => {
+        .then((updatedEvent: any) => {
```
arjita-mitra, line 580: why are we using any here?

atomrc: This is for a test, so it's not really important to get the types right here 👌

src/script/storage/record/EventRecord.ts
atomrc: This is legacy sadly. 
It can mean:
- it's a datestring (string)
- it's a timestamp (number)
- it has been consumed (boolean)

atomrc: Actually let me add this as a comment :) Good call

PR #14642, opened 2023-02-08
runfix: client mismatch

PatrykBuniX:
https://github.com/wireapp/wire-webapp/pull/14638 introduced some regressions in calling (clients verified checks). The difference in previous hacky api call was that it did not return selfClient (and the new endpoint we're using does). 

This PR enables us to filter self client out of the userMap payload we receive from backend.


src/script/conversation/ConversationRepository.ts
```diff
@@ -184,8 +187,14 @@ export class ConversationRepository {
     // we register a client mismatch handler agains the message repository so that we can react to missing members
     // FIXME this should be temporary. In the near future we want the core to handle clients/mismatch/verification. So the webapp won't need this logic at all
     this.messageRepository.setClientMismatchHandler(async (mismatch, conversation, silent, consentType) => {
+      //we filter out self client id to omit it in mismatch check
+      const {userId, clientId} = this.core;
+      const selfClient = {domain: conversation?.domain, userId, clientId};
+      const filteredMissing = mismatch.missing && removeClientFromUserClientMap(mismatch.missing, selfClient);
+      const filteredMismatch = {...mismatch, missing: filteredMissing} as ClientMismatch | MessageSendingStatus;
```
PatrykBuniX, line 196: For some reason TS is not happy, even though we spread original object there and `mismatch.missing` has exact same type as `filteredMissing` 🤔

src/script/conversation/ConversationRepository.ts
```diff
@@ -184,8 +187,14 @@ export class ConversationRepository {
     // we register a client mismatch handler agains the message repository so that we can react to missing members
     // FIXME this should be temporary. In the near future we want the core to handle clients/mismatch/verification. So the webapp won't need this logic at all
     this.messageRepository.setClientMismatchHandler(async (mismatch, conversation, silent, consentType) => {
+      //we filter out self client id to omit it in mismatch check
+      const {userId, clientId} = this.core;
+      const selfClient = {domain: conversation?.domain, userId, clientId};
```
atomrc: Oh the domain need to be the user domain, not the conversation's one

PatrykBuniX: oops, fixed 👌

src/script/conversation/ConversationRepository.ts
atomrc: interesting indeed 🤔

atomrc (approved):

PR #14647, opened 2023-02-09
chore: Improve lint job to use more CPUs

atomrc:
Now that the entire build is faster than just the linting part, let's speed up the linting job a little bit 

This bring down the job from 3min to 1min


PatrykBuniX (approved):
Speeeeeed 🚀

PR #14674, opened 2023-02-13
test: mls conference calls

PatrykBuniX:
Add test cases for:
- starting a MLS conference call updates epoch info
- answering a MLS conference call updates epoch info
- after doing one of the above we start to listen to `newEpoch` mlsService event - every time this event is emitted we're updating epoch info


atomrc (approved):
Solid 🪨 🚀

przemvs (approved):

V-Gira (approved):

PR #14677, opened 2023-02-14
fix: Add padding to start-ui search when typing (SQSERVICES-1678)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1678" title="SQSERVICES-1678" target="_blank">BugSQSERVICES-1678</a>  Connection search box text passes underneath "clear all" button.
  </td></table>
  <br />
 

Changed from

image

to

image

### DIFF
--- a/src/style/components/search-input.less
+++ b/src/style/components/search-input.less
@@ -166,5 +166,5 @@ body.theme-dark {
 }
 
 .search-input-padding {
-  padding: 0 40px 0 40px !important;
+  padding: 0 40px !important;
 }
--- a/src/style/components/search-input.less
+++ b/src/style/components/search-input.less
@@ -164,3 +164,7 @@ body.theme-dark {
     margin-bottom: 6px;
   }
 }
+
+.search-input-padding {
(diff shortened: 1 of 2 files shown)


src/style/components/search-input.less
```diff
@@ -164,3 +164,7 @@ body.theme-dark {
     margin-bottom: 6px;
   }
 }
+
+.search-input-padding {
+  padding: 0 40px 0 40px !important;
```
przemvs: Maybe use **padding-inline: 40px** instead of **padding: 0 40px 0 40px** :)

thisisamir98: https://github.com/wireapp/wire-webapp/pull/14677/commits/13b0cee46c4fd378e103b2659006b7601ed9e543

how about this @przemvs 
even more simple!

przemvs: this also okay :)

przemvs (approved):

PR #14689, opened 2023-02-16
chore: Refactor deployment to dev/prod environments

atomrc:
The code for deploying to dev or prod env is a copy/paste for every single env. 
This PR refactors this and deduce the env to deploy to depending on the branch or the tag and avoids code duplication

### DIFF
--- a/.github/workflows/test_build_deploy.yml
+++ b/.github/workflows/test_build_deploy.yml
@@ -87,6 +87,7 @@ jobs:
 
       - uses: kanga333/variable-mapper@master
         with:
+          # We try to map a branch to a dev environment
           key: '${{github.ref}}'
           map: |
             {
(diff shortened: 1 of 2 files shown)


.github/workflows/test_build_deploy.yml
```diff
@@ -309,6 +223,7 @@ jobs:
         with:
           tag_name: ${{env.TAG}}
           name: ${{env.TAG}}
+          body_path: ./CHANGELOG.md
```
atomrc, line 228: This will add the changelog automatically to the github release

tlebon: is this not happening on web? it happens on tm but maybe its handled elsewhere I'm not sure now.

tlebon: yes this is included on TM, just checked.

tlebon (approved):
Seems like a nice change, ill have to incorporate this into TM as well.

.github/workflows/test_build_deploy.yml
atomrc: It was happening for production releases, but not staging for some reasons ... 🙃

PR #14728, opened 2023-02-21
fix: Add error boundaries (SQSERVICES-1920)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1920" title="SQSERVICES-1920" target="_blank">BugSQSERVICES-1920</a>  [Web] calls could crash with a white screen
  </td></table>
  <br />
 

Introduce error boundaries


package.json
```diff
@@ -32,6 +32,7 @@
     "platform": "1.3.6",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "react-error-boundary": "^3.1.4",
```
PatrykBuniX: Lets use exact version (without `^`)

src/script/components/panel/UserDetails.tsx
```diff
@@ -147,4 +149,12 @@ const UserDetails: React.FC<UserDetailsProps> = ({
   );
 };
 
+const UserDetails: React.FC<UserDetailsProps> = props => {
+  return (
+    <ErrorBoundary FallbackComponent={ErrorFallback}>
```
atomrc, line 151: In case this adds something to the DOM,cCan you check that this doesn't break the CSS flow?

src/script/components/panel/UserDetails.tsx
```diff
@@ -62,8 +64,8 @@ const UserDetails: React.FC<UserDetailsProps> = ({
     amplify.publish(WebAppEvents.USER.UPDATE, participant.qualifiedId);
   }, [participant]);
 
-  const isFederated = participant.isFederated;
-  const isGuest = !isFederated && participant.isGuest();
+  const isFederated = user.isFederated;
```
atomrc: Did you test this code? This is breaking since `user` doesn't have the `isFederated` property

src/script/components/panel/UserDetails.tsx
```diff
@@ -62,8 +64,8 @@ const UserDetails: React.FC<UserDetailsProps> = ({
...
+  const isGuest = !isFederated && user.isGuest;
```
atomrc, line 70: Did you find where the `isGuest` is transformed from the participant object?

atomrc (changes requested):

src/script/components/ErrorFallback.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+import {render} from '@testing-library/react';
+import {ErrorBoundary} from 'react-error-boundary';
+
+import {ErrorFallback} from './ErrorFallback';
+
+const SimpleError: React.FC = () => {
+  throw new Error('failed to render');
```
tlebon, line 27: this is a strange FC

tlebon:
do we have designs for the error boundary?

src/script/components/panel/UserDetails.tsx
atomrc: This PR should detect those mistakes automatically https://github.com/wireapp/wire-webapp/pull/14729

src/script/components/panel/UserDetails.tsx
thisisamir98: DOM is all fine, no changes.

image

thisisamir98:
@tlebon we will use modals and someone from design team will have a final look.

atomrc (approved):

PR #14752, opened 2023-02-27
runfix: Do not udpate conversation roles when self user leaves a conversation

atomrc:
Fixes a bug introduce by #14736


PatrykBuniX (approved):
Makes sense 👌

V-Gira (approved):

PR #14753, opened 2023-02-27
fix: Shortcut for incoming call not work properly

przemvs:
https://wearezeta.atlassian.net/browse/ACC-437?focusedCommentId=111099

Fix for answer or deciline call by shortcut.
Prev: Shortcut capturing another input actions.
Now: Shortcut capturing only answer / deciline call. Normal usage of keyboard works properly.

### DIFF
--- a/src/script/components/calling/CallingCell.tsx
+++ b/src/script/components/calling/CallingCell.tsx
@@ -259,6 +259,9 @@ const CallingCell: React.FC<CallingCellProps> = ({
 
   useEffect(() => {
     if (isIncoming) {
+      // Capture will be dispatched to registered element before being dispatched to any EventTarget beneath it in the DOM Tree.
+      // It's needed because when someone is calling we need to change order of shortcuts to the top of keyboard usage.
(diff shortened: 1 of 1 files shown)


src/script/components/calling/CallingCell.tsx
```diff
@@ -251,10 +259,10 @@ const CallingCell: React.FC<CallingCellProps> = ({
 
   useEffect(() => {
     if (isIncoming) {
-      window.addEventListener('keydown', answerOrRejectCall);
+      document.addEventListener('keydown', answerOrRejectCall, {capture: true});
```
atomrc, line 265: Can you just add a comment to why you add a `capture` here, would be nice :)

atomrc: Since this is rarely used, I think it would be nice to have a trace of the explanation in the code

V-Gira: I'd appreciate it :)

atomrc (approved):

V-Gira (approved):

PR #14759, opened 2023-02-28
feat: Add server flag to enable CoreCrypto [FS-1594]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1594" title="FS-1594" target="_blank">TaskFS-1594</a>  [web] Avoid releasing CoreCrypto on production before next release
  </td></table>
  <br />
 

Will make sure we don't enable coreCrypto on production just yet. 

Enabling coreCrypto will mean we will have to blacklist some wrapper versions (which should be developed properly)


server/config.ts
```diff
@@ -154,6 +154,7 @@ const config: ServerConfig = {
         Number(process.env.FEATURE_MLS_CONFIG_KEYING_MATERIAL_UPDATE_THRESHOLD),
       PERSIST_TEMPORARY_CLIENTS: process.env.FEATURE_PERSIST_TEMPORARY_CLIENTS != 'false',
       SHOW_LOADING_INFORMATION: process.env.FEATURE_SHOW_LOADING_INFORMATION == 'true',
+      USE_CORE_CRYPTO: process.env.FEATURE_USE_CORE_CRYPTO == 'true',
     },
     MAX_GROUP_PARTICIPANTS: (process.env.MAX_GROUP_PARTICIPANTS && Number(process.env.MAX_GROUP_PARTICIPANTS)) || 500,
```
przemvs, line 159: Maybe you can fix this magic numbers in this file ? :D

przemvs (approved):
Code LGTM, 1 thing ;)

server/config.ts
atomrc: This should not be considered a magic number, since it's assigned to an object properties (so it has a name). 
This magic number error should not be raised

PatrykBuniX (approved):

V-Gira (approved):

PR #14760, opened 2023-02-28
refactor: Kill all non-domain-aware apis

atomrc:
see https://github.com/wireapp/wire-web-packages/pull/4927

### DIFF
--- a/src/script/conversation/ConversationService.ts
+++ b/src/script/conversation/ConversationService.ts
@@ -19,11 +19,9 @@
 
 import type {
   CONVERSATION_ACCESS_ROLE,
-  ClientMismatch,
   Conversation as BackendConversation,
   ConversationCode,
   CONVERSATION_ACCESS,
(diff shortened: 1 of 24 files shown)


src/script/conversation/ConversationService.ts
```diff
@@ -280,43 +278,6 @@ export class ConversationService {
...
-   *   "<user-id>": {
-   *     "<client-id>": "<base64-encoded-encrypted-content>"
-   *   }
-   * }
-   *
-   * @param conversationId ID of conversation to send message in
-   * @param payload Payload to be posted
-   * @returns Promise that resolves when the message was sent
-   */
-  postEncryptedMessage(
```
atomrc, line 304: This was not used since a long time (now the core handles encryption)

V-Gira (approved):
Only qualified Ids shall remain :+1:

PR #14764, opened 2023-03-01
fix: Keyboard navigation for conversations list

przemvs:
Fix for switching conversation list when we get out from conversation list.
Prev: If we select context menu and click outside or use tab || shift + tab we going back to previous selected conversation.
Now: If we select context menu and click outside or use tab || shift + tab we selecting first conversation ( on top )


src/script/components/list/ConversationListCell.tsx
```diff
@@ -55,6 +55,7 @@ export interface ConversationListCellProps {
   showJoinButton: boolean;
   handleArrowKeyDown: (e: React.KeyboardEvent) => void;
   isFocused?: boolean;
+  setDefaultFocus: () => void;
```
atomrc: can you document what this function is supposed to do. It's not quite clear for the code itself

atomrc (approved):

PR #14772, opened 2023-03-03
feat: clear mls group on conversation delete [FS-1602]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1602" title="FS-1602" target="_blank">TaskFS-1602</a>  [Web] Wipe local MLS group when a conversation gets deleted 
  </td></table>
  <br />
 

Clearing local MLS group after **deleting** a conversation was never implemented (It was handled `on-leave` only). This PR makes sure that after deleting conversation locally from the store we also wipe MLS group (`corecrypto.wipeConversation`)

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -838,7 +838,7 @@ export class ConversationRepository {
       });
   }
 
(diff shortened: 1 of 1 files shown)


przemvs (approved):
🔥

V-Gira (approved):

PR #14788, opened 2023-03-06
feat: check if call epoch change event was fired by parent conversation

PatrykBuniX:
This will make parent conversation membership change affect subconversation.


src/script/calling/mlsConference.ts
```diff
@@ -104,11 +104,24 @@ export const subscribeToEpochUpdates = async (
 
   const forwardNewEpoch = async ({groupId, epoch}: {groupId: string; epoch: number}) => {
     if (groupId !== subconversationGroupId) {
-      return;
+      const parentConversationId = await mlsService.findConversationIdByGroupId?.(groupId);
```
PatrykBuniX: I'm not sure this is not too heavy as `forwardNewEpoch` will get called every time any of the group will introduce new epoch... `findConversationIdByGroupId` does a `.find()` on the list of stored conversations. Other approach would be to store `groupId` -> `conversationId` map.

atomrc: I think it's fine for now. Adding a new stored value could lead of sync problems later on. 
Epoch changes doesn't happen on a `ms` frequency basis. It's fine to do a `find` on everyone of them 👍

atomrc (approved):
LGTM

PR #14791, opened 2023-03-07
feat: Integrate datadog logging

przemvs:
You can check logs in DataDog: https://app.datadoghq.eu/ .
For access I think you need to contact to Mr. T. 
To watch logs login in DataDog ( link above ) => click on Logs => Service => Click on web-internal.


src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+      trackUserInteractions: true,
+      trackInteractions: true,
+      trackResources: true,
+      trackLongTasks: true,
+      defaultPrivacyLevel: 'mask-user-input',
+    });
+
+    datadogRum.startSessionReplayRecording();
+
+    const regex = /([a-z\d]{8})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{12})/gim;
```
atomrc: This naming is not explicit, it should convey what this regexp matches (uuid?)

src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const initializeDataDog = (config: Configuration, domain?: string) => {
+  const applicationId = config.dataDog?.applicationId;
+  const clientToken = config.dataDog?.clientToken;
+
+  if (!applicationId || !clientToken) {
+    return;
+  }
+
+  import('@datadog/browser-rum').then(({datadogRum}) => {
```
atomrc: Why not ?

```js
const datadogRum = await import('...');
```

src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+    datadogRum.startSessionReplayRecording();
+
+    const regex = /([a-z\d]{8})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{12})/gim;
+
+    const replacer = (_match: string, p1: string) => `${p1}***`;
+    const truncateDomain = (value: string) => `${value.substring(0, 3)}***`;
+    const replaceAllStrings = (string: string) => string.replaceAll(regex, replacer);
+    const replaceDomains = (string: string) => (domain ? string.replaceAll(domain, truncateDomain(domain)) : string);
+
+    import('@datadog/browser-logs').then(({datadogLogs}) => {
```
atomrc: Why not?

```js
const {datadogLogs} = await import('..')
```

src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+      trackLongTasks: true,
+      defaultPrivacyLevel: 'mask-user-input',
+    });
+
+    datadogRum.startSessionReplayRecording();
+
+    const regex = /([a-z\d]{8})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{12})/gim;
+
+    const replacer = (_match: string, p1: string) => `${p1}***`;
+    const truncateDomain = (value: string) => `${value.substring(0, 3)}***`;
```
atomrc: How do you detect domains??? It feels like you replace any string with just the 3 first chars?!

src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Configuration} from '../Config';
+
+export const initializeDataDog = (config: Configuration, domain?: string) => {
```
atomrc:
```suggestion
export function initializeDataDog(config: Configuration, domain?: string) {
```

atomrc: shorter syntax

src/script/util/DataDog.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  import('@datadog/browser-rum').then(({datadogRum}) => {
+    datadogRum.init({
+      applicationId,
+      clientToken,
+      site: 'datadoghq.eu',
+      service: 'web-internal',
+      env: config.ENVIRONMENT,
+      // Specify a version number to identify the deployed version of your application in Datadog
+      // version: '1.0.0',
```
atomrc:
```suggestion
    version: config.VERSION,
```

src/script/util/DataDog.ts
przemvs: The domain comes as prop, so we know whats name have domain. This only function to have more readable code ;)
On ss we have replacer for domain. 
image

atomrc: But you are only replacing the domain of the current user. Which means on federated environments we won't filter other domains. We will need to find a solution for this as well 🤔

przemvs: Ahh.. sure :/

src/script/conversation/ConversationService.ts
```diff
@@ -366,8 +366,8 @@ export class ConversationService {
 
     return this.storageService
       .save(StorageSchemata.OBJECT_STORE.CONVERSATIONS, conversation_et.id, conversationData)
-      .then(primary_key => {
-        this.logger.info(`State of conversation '${primary_key}' was stored`, conversationData);
+      .then(_primary_key => {
```
atomrc:
```suggestion
      .then(() => {
```

src/script/util/DebugUtil.ts
```diff
@@ -262,8 +262,8 @@ export class DebugUtil {
 
     const logMessage = `Hey ${this.userState.self().name()}, this is for you:`;
     this.logger.warn(logMessage, debugInformation);
-    this.logger.warn(`Conversation: ${debugInformation.conversation.name()}`, debugInformation.conversation);
-    this.logger.warn(`From: ${debugInformation.user.name()}`, debugInformation.user);
+    this.logger.warn(`Conversation: ${debugInformation.conversation.name()}`);
+    this.logger.warn(`From: ${debugInformation.user.name()}`);
```
atomrc: Oh actually, this is for debugging, we could leave the initial payload (it's okay since it's used by QA)

src/script/auth/main.tsx
```diff
@@ -85,4 +86,5 @@ async function runApp() {
 }
 
 enableLogging(Config.getConfig().FEATURE.ENABLE_DEBUG);
+initializeDataDog(Config.getConfig());
```
atomrc: How about moving this call inside of `enableLogging` (so that you don't have to put it in both `main.tsx` and `app.ts`)?

atomrc: and we can pass the full config to the `enableLogging` method

atomrc (approved):
nice job

PR #14794, opened 2023-03-07
fix: Do not try to load deleted users on app load [FS-1607]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1607" title="FS-1607" target="_blank">BugFS-1607</a>  [web] Webapp is not removing deleted users for the `users` database
  </td></table>
  <br />
 

This changes the strategy we use to load the user availabilities and put them on the user enties. 

What we did before was:
- load all the user availabilities in DB
- map those to locally loaded users and load missing users from the BE
-> this would lead to querying deleted users from the server

The new strategy is:
- load all the users from backend 
- for all those loaded users, find the availabilities in DB
-> this way some availabilities might not be mapped, but that means they belonged to deleted users

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -409,7 +409,7 @@ export class App {
 
       await conversationRepository.conversationRoleRepository.loadTeamRoles();
 
-      await userRepository.loadUsers();
+      await userRepository.loadTeamUserAvailabilities();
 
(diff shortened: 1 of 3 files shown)


przemvs (approved):
Code LGTM :)

PR #14795, opened 2023-03-07
fix: leave call on conversation delete [FS-1608]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1608" title="FS-1608" target="_blank">BugFS-1608</a>  [Web] client does not leave call when conversation gets deleted
  </td></table>
  <br />
 

Call was never ended after conversation was deleted, so end users were still able to here each other even though there was no conversation nor call visible on the UI.

This PR should fix this bug for both Proteus and MLS calls + `@wireapp/core` package version bump.

### DIFF
--- a/src/script/view_model/CallingViewModel.ts
+++ b/src/script/view_model/CallingViewModel.ts
@@ -240,10 +240,9 @@ export class CallingViewModel {
       this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
     };
 
-    const leaveCall = async (conversationId: QualifiedId) => {
-      const conversation = this.getConversationById(conversationId);
-
-      if (!conversation?.isUsingMLSProtocol) {
+    const closeCall = async (conversationId: QualifiedId, conversationType: CONV_TYPE) => {
+      // There's nothing we need to do for non-mls calls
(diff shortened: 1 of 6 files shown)


src/script/view_model/CallingViewModel.ts
```diff
@@ -240,10 +240,9 @@ export class CallingViewModel {
       this.callingRepository.setEpochInfo(conversationId, {epoch, keyLength, secretKey}, members);
     };
 
-    const leaveCall = async (conversationId: QualifiedId) => {
-      const conversation = this.getConversationById(conversationId);
-
-      if (!conversation?.isUsingMLSProtocol) {
+    const closeCall = async (conversationId: QualifiedId, conversationType: CONV_TYPE) => {
+      // There's nothing we need to do for non-mls calls
+      if (conversationType !== CONV_TYPE.CONFERENCE_MLS) {
```
PatrykBuniX, line 245: In case of closing a call after deleting a conversation, the conversation might not be found with `getConversationById` as call close works asynchronously (and is fired by avs). So it's better to relay on a call conversation type.

thisisamir98 (approved):

atomrc (approved):

PR #14800, opened 2023-03-10
fix: Improve message sending performances

### LINKED ISSUE #6486: latest update - long delay after pressing enter/send  before outgoing message appears
Wire version: 3.30.4368
Wire for web version: 2023.03.07.08.22
Operating system: 
Which antivirus software do you have installed: 

What steps will reproduce the problem?
enter message hit send 

What is the expected result?
you quickly see the message you just sent (on the senders side)

What is the actual result?
enter message hit send then nothing happens - eventually your sent message appears approximately 6 to 8 seconds after you pressed enter/send - this was confirmed by multiple other uses and secondary installs/deployments 

Please provide any additional information below. Attach a screenshot if possible.
 upon further testing it appears it's likely related to message history size as sending a message to contacts with little message history does not entail a long delay after pressing enter/send

atomrc:
Since #14632 we never query individual events by their direct index (the `id`). 
Instead we load **all** the events and go through them one by one until we find the matching ID (or replacing ID). 

This is very suboptimal and gets slower and slower as more messages are added to the conversation. 

We will still need to optimize loading quotes. But this should fix many performances problems

https://user-images.githubusercontent.com/1090716/224311015-dd80ccf0-63bb-4eda-954d-0496fb25c62c.mov

https://user-images.githubusercontent.com/1090716/224311037-32669af1-2faa-4d77-9c7a-551c4ff71fe3.mov

Should fix https://github.com/wireapp/wire-desktop/issues/6486

### DIFF
--- a/src/script/event/EventService.ts
+++ b/src/script/event/EventService.ts
@@ -127,17 +127,13 @@ export class EventService {
 
     try {
       if (this.storageService.db) {
+        const eventStore = this.storageService.db.table(StorageSchemata.OBJECT_STORE.EVENTS);
         // First lookup the event by its direct id (using the index)
-        const event = this.storageService.db
-          .table(StorageSchemata.OBJECT_STORE.EVENTS)
-          .where('id')
-          .equals(eventId)
-          .first();
+        const event = eventStore.where('id').equals(eventId).first();
         return (
           event ||
           // If the event was not found, fallback to filtering all the events and check if a `replacing` message is found
-          this.storageService.db
-            .table(StorageSchemata.OBJECT_STORE.EVENTS)
+          eventStore
             .where('conversation')
             .equals(conversationId)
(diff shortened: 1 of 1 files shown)


PatrykBuniX (approved):

src/script/event/EventService.ts
```diff
@@ -127,12 +127,22 @@ export class EventService {
...
-          .equals(conversationId)
-          .filter(item => item.data?.replacing_message_id === eventId || item.id === eventId)
+          .where('id')
+          .equals(eventId)
           .first();
+        return (
+          event ||
+          // If the event was not found, fallback to filtering all the events and check if a `replacing` message is found
+          this.storageService.db
+            .table(StorageSchemata.OBJECT_STORE.EVENTS)
```
przemvs: Move this to variable maybe ? You getting this table 2 times :)

przemvs (approved):

PR #14832, opened 2023-03-16
feat: user joins mls group with guest link

PatrykBuniX:
When user receives `member-join` event, and it's also a creator of this event, it means that user has joined by himself, therefore we need to add its other clients to mls group if conversation is using mls protocol


src/script/mls/MLSConversations.ts
```diff
@@ -110,3 +113,38 @@ export async function registerUninitializedConversations(
...
+ */
+export async function addOtherSelfClientsToMLSConversation(
+  conversationId: QualifiedId,
+  selfUserId: QualifiedId,
+  clientId: string,
+  conversationRepository: MLSConversationRepository,
+  core: Account,
+) {
+  const conversation = await conversationRepository.getConversationById(conversationId);
+  const {groupId, isUsingMLSProtocol} = conversation;
```
przemvs:
```suggestion
  const {groupId, isUsingMLSProtocol}  = await conversationRepository.getConversationById(conversationId);
```

src/script/conversation/ConversationRepository.ts
```diff
@@ -2460,28 +2460,32 @@ export class ConversationRepository {
...
+
     // Self user joins again
-    const selfUserRejoins = eventData.user_ids.includes(this.userState.self().id);
-    if (selfUserRejoins) {
+    const selfUserJoins =
+      eventData.user_ids.includes(this.userState.self().id) ||
+      !!eventData.users?.some(
+        ({qualified_id: qualifiedId}) =>
+          qualifiedId && matchQualifiedIds(qualifiedId, this.userState.self().qualifiedId),
+      );
```
przemvs: Maybe split it to 2 vars? It will be more readable.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2493,6 +2497,47 @@ export class ConversationRepository {
...
+
+    const isMLSConversationEstablished = await this.core.service!.conversation.isMLSConversationEstablished(
+      conversation.groupId,
+    );
+
+    if (!isMLSConversationEstablished) {
+      return;
+    }
+
+    const isMLSConversationMarkedAsEstablished = useMLSConversationState.getState().isEstablished(conversation.groupId);
```
przemvs: Create **useMLSConversationState.getState()** as variable, you're using it there, multiple times.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2493,6 +2497,47 @@ export class ConversationRepository {
...
+
+    if (!isMLSConversationMarkedAsEstablished) {
+      // If the conversation was not previously marked as established and the core if aware of this conversation, we can mark is as established
+      useMLSConversationState.getState().markAsEstablished(conversation.groupId);
```
przemvs: Create **useMLSConversationState.getState()** as variable, you're using it there, multiple times.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2493,6 +2497,47 @@ export class ConversationRepository {
...
+   * @param conversation Conversation member joined to
+   * @param isSelfJoin whether user has joined by itself, if so we need to add other self clients to mls group
+   */
+  private async handleMLSConversationMemberJoin(conversation: Conversation, isSelfJoin: boolean) {
+    if (!conversation.groupId) {
+      throw new Error(`groupId not found for MLS conversation ${conversation.id}`);
+    }
+
+    const isMLSConversationEstablished = await this.core.service!.conversation.isMLSConversationEstablished(
+      conversation.groupId,
```
przemvs: Maybe spread groupId from conversation object, above this const, and reuse it in another places :)

src/script/conversation/ConversationRepository.ts
```diff
@@ -2493,6 +2497,47 @@ export class ConversationRepository {
...
+    }
+
+    if (isSelfJoin) {
+      // if user has joined and was also event creator (eg. joined via guest link) we need to add its other clients to mls group
+      try {
+        await addOtherSelfClientsToMLSConversation(
+          conversation,
+          this.userState.self().qualifiedId,
+          this.core.clientId,
+          this.core,
```
przemvs, line 2535: Hmm.. We can't pass only this.core to this function? Because core is also required prop, so you can get clientId from this.core ;)

przemvs: Or this clientId can be different? So maybe create clientId as optional prop ;)

PatrykBuniX: I'd rather keep it this way to make this function more generic, we might not always want to use clientId from core

przemvs: Hmm.. So maybe use clientId as optional property, and if we pass clientId use clientId if not use core id, wdyt ?

PatrykBuniX: I'd rather not use any value implicitly, having required param we're making obvious what value will be used without looking at function's body ;)

przemvs (approved):

src/script/conversation/ConversationRepository.test.ts
```diff
@@ -746,6 +752,46 @@ describe('ConversationRepository', () => {
         });
       });
 
+      it('should add other self clients to mls group if user was event creator', () => {
```
atomrc, line 755: 🤩

src/script/conversation/ConversationRepository.ts
```diff
@@ -2460,28 +2460,34 @@ export class ConversationRepository {
       });
     }
 
+    // Self user is a creator of the event
+    const isFromSelf = eventJson.from === this.userState.self().id;
+
     // Self user joins again
-    const selfUserRejoins = eventData.user_ids.includes(this.userState.self().id);
-    if (selfUserRejoins) {
+
```
atomrc, line 2467:
```suggestion
```

src/script/conversation/ConversationRepository.ts
```diff
@@ -2460,28 +2460,34 @@ export class ConversationRepository {
       });
     }
 
+    // Self user is a creator of the event
+    const isFromSelf = eventJson.from === this.userState.self().id;
+
     // Self user joins again
```
atomrc: Is this still relevant?

atomrc (approved):
Looks solid. Great job 😎

src/script/conversation/ConversationRepository.ts
PatrykBuniX: nope, removed it 👌

PR #14839, opened 2023-03-19
feat: add ability to react on messages(SQSERVICES-1734)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1734" title="SQSERVICES-1734" target="_blank">StorySQSERVICES-1734</a>  [Web] Reactions on Messages
  </td></table>
  <br />
 

----

- The **PR Title**
  - add reaction on messages

- The **PR Description**
  - As a Wire user, I would like to have a wide range of reactions (emojis) at my disposal. So that I can use them to express myself in messages and reactions to messages.
---

#### How to Test

- added unit testing

##### References
1. https://wearezeta.atlassian.net/browse/SQSERVICES-1734
2. https://www.npmjs.com/package/emoji-picker-react


src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -62,14 +64,15 @@ export interface ContentMessageProps extends Omit<MessageActions, 'onClickResetS
   quotedMessage?: ContentMessage;
   selfId: QualifiedId;
   isMsgElementsFocusable: boolean;
+  messageRepository: MessageRepository;
```
atomrc: Humm the content message component should not know about the `messageRepository`. It's doesn't belong in this layer. 
The `MessageWrapper` on the other hand already knows about the `MessageRepository` and should be the one calling the method. 

Instead this component should be given a  callback `onClickReaction` and the parent (`MessageWrapper`) should make the call :)

src/script/conversation/MessageRepository.ts
```diff
@@ -786,6 +786,12 @@ export class MessageRepository {
     }
   }
 
+  public onReactionClick(conversationEntity: Conversation, message_et: ContentMessage, reaction: string): void {
+    if (!conversationEntity.removed_from_conversation()) {
+      window.setTimeout(() => this.sendReaction(conversationEntity, message_et, reaction), 100);
+    }
+  }
+
```
atomrc, line 794: Why not making the `sendReaction` a public method? Why the `timeout`?

atomrc: Why not making the sendReaction a public method? Why the timeout?

src/script/conversation/MessageRepository.ts
```diff
@@ -892,8 +898,9 @@ export class MessageRepository {
    * @param reactionType Reaction
    * @returns Resolves after sending the reaction
    */
-  private async sendReaction(conversation: Conversation, messageEntity: Message, reactionType: ReactionType) {
-    const reaction = MessageBuilder.buildReactionMessage({originalMessageId: messageEntity.id, type: reactionType});
+
+  private async sendReaction(conversation: Conversation, messageEntity: Message, reactions: ReactionType) {
```
atomrc: The method name `reaction` (singular) but it takes a `reactions` (plural). That seems inconsistent

yalc.lock
```diff
@@ -0,0 +1,10 @@
+{
+  "version": "v1",
+  "packages": {
+    "@wireapp/core": {
+      "signature": "ce501479df24a358c5bc9af495d7b8f4",
+      "file": true,
+      "replaced": "39.1.3"
+    }
+  }
+}
```
atomrc, line 10: This should not be committed

src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -215,6 +215,7 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
         onRetry={onRetry}
         isMessageFocused={isMessageFocused}
         isMsgElementsFocusable={isMsgElementsFocusable}
+        messageRepository={messageRepository}
```
atomrc:
```suggestion
        onClickReaction={() => messageRepository.sendReaction(...)}
```

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactionsList.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {render, fireEvent, within} from '@testing-library/react';
+
+import {MessageReactionsList} from './MessageReactionsList';
+
+describe('MessageReactionsList', () => {
```
atomrc, line 24: 🤩

atomrc (changes requested):

src/script/conversation/MessageRepository.ts
arjita-mitra: @atomrc I am not sure about that, its the same function that we used for toggleLike I used here. I had to change the reaction type to be string. since it was an existing working code so i didn't change it

arjita-mitra: /**
   * Toggle like status of message.
   *
   * @param conversationEntity Conversation entity
   * @param message_et Message to react to
   */
  public toggleLike(conversationEntity: Conversation, message_et: ContentMessage): void {
    if (!conversationEntity.removed_from_conversation()) {
      const reaction = message_et.is_liked() ? ReactionType.NONE : ReactionType.LIKE;
      message_et.is_liked(!message_et.is_liked());

      window.setTimeout(() => this.sendReaction(conversationEntity, message_et, reaction), 100);
    }
  }

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactionsList.test.tsx
arjita-mitra: is there anything to fix here or you are just happy 🤔

atomrc: Just happy :)

src/script/conversation/MessageRepository.ts
atomrc: I'd say the timeout def doesn't belong here. I'd like to know why we need it and if needed we need to document this. 
This is not explicit why it's needed 🤔

src/script/conversation/MessageRepository.ts
arjita-mitra: reaction can be single or multiple comma separated. what do you suggest in this case?

.env.localhost
```diff
@@ -17,7 +17,7 @@ ENABLE_DEV_BACKEND_API="true"
 #FEATURE_APPLOCK_UNFOCUS_TIMEOUT="30"
 #FEATURE_APPLOCK_SCHEDULED_TIMEOUT="30"
 #FEATURE_ENABLE_MLS="true"
-#FEATURE_ENABLE_REACTION="false"
+#FEATURE_ENABLE_REACTION="true"
```
atomrc, line 20: Is this consumed anywhere? 
Maybe this can be removed altogether now that we are developing the feature entirely in this branch (we don't need to keep the simple `like` around if we go this strategy)

yalc.lock
arjita-mitra: yes, i was planing to remove it after ui-kit version bump

src/script/components/MessagesList/Message/MessageWrapper.tsx
arjita-mitra: Not all message have reaction, only content message can have reaction. so it made more sense to me to have it inside contentmessage instead of message wrapper.

src/script/conversation/MessageRepository.ts
atomrc: Can we send multiple reactions at once? 
If that's the case, we should name it `sendReactions`. 

How is it possible to send multiple reactions at once? or does it overwrite each time we send?

src/script/components/MessagesList/Message/MessageWrapper.tsx
atomrc: The problem is, this injects too much power in the `ContentMessage`. It should never have full access to the `MessageRepository` this is not going to scale. 
Only a handful of components should have access to it. In this case it's the `MessageWrapper`

.env.localhost
arjita-mitra: yes, it is being used inside ContentMessage.tsx  
const isReactionFeatureEnabled = Config.getConfig().FEATURE.ENABLE_REACTION;

we need the feature flag to disable the feature from being displayed on dev, after we merge.

src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -189,6 +190,15 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
...
 
+  const handleReactionClick = useCallback(
+    (reaction: ReactionType) => {
+      if (!message.isContent()) {
+        return;
+      }
+      return void messageRepository.toggleReaction(conversation, message, reaction, selfId.id);
+    },
+    [conversation, message, messageRepository, selfId.id],
+  );
```
atomrc: I'd recommend not using `useCallback` and `useMemo` as default but rather when it's really needed. 
In this case it's a pure overhead

```suggestion
  const handleReactionClick = (reaction: ReactionType) => {
      if (!message.isContent()) {
        return;
      }
      return void messageRepository.toggleReaction(conversation, message, reaction, selfId.id);
    };
```

atomrc (approved):

PR #14867, opened 2023-03-21
chore: bump core version [FS-1562]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1562" title="FS-1562" target="_blank">BugFS-1562</a>  [Web][MLS] When the current device is deleted via another device, coreCrypto db wipe could fail
  </td></table>
  <br />
 

Should fix issue where client is not logged out properly after its device was removed using some other client.
For more details see https://github.com/wireapp/wire-web-packages/pull/4972

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "@emotion/react": "11.10.6",
     "@types/eslint": "8.4.10",
     "@wireapp/avs": "9.1.13",
-    "@wireapp/core": "39.2.7",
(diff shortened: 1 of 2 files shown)


V-Gira (approved):

PatrykBuniX:
Closing as version was already bumped. See https://github.com/wireapp/wire-webapp/pull/14868/

PR #14868, opened 2023-03-21
fix: Avoid duplicated current client in Database

atomrc:
See https://github.com/wireapp/wire-web-packages/pull/4992

This will avoid having an entry for the current client in the database when we load the application. 
This could cause the webapp to try to send an encrypted message to this particular device (which could cause a session not to be found)


V-Gira (approved):
Just gonna leave this here:
Also moves the `ReactionType` Enum from  webpackages to the webapp

PR #14876, opened 2023-03-23
fix: Never consider a Service as a User

atomrc:


atomrc:
Idea abandoned. Services needs some deeper changes

PR #14877, opened 2023-03-23
feat: split user list item

przemvs:
Splitted ParticipantItem into UserListItem.
Removed ParticipantItem - is not used anymore.


src/script/components/UserList/components/UserListItem/UserListItem.tsx
```diff
@@ -26,149 +26,115 @@ import {Checkbox, CheckboxLabel} from '@wireapp/react-ui-kit';
...
-import {noop, setContextMenuPosition} from 'Util/util';
 
-import {Participant} from '../../calling/Participant';
-import {User} from '../../entity/User';
-import {ServiceEntity} from '../../integration/ServiceEntity';
+import {User} from '../../../../entity/User';
 
-export interface ParticipantItemProps<UserType> extends Omit<React.HTMLProps<HTMLDivElement>, 'onClick' | 'onKeyDown'> {
+export interface UserListItemProps {
   badge?: boolean;
```
PatrykBuniX: Let's rename this prop to sth like `hasBadge` since it's `boolean` type.

PatrykBuniX: Or actually seems like it was storing the same value as `highlighted` prop. Let's name it `isHighlighted`.

src/script/components/UserList/components/UserListItem/UserListItem.tsx
```diff
@@ -225,65 +180,36 @@ const ParticipantItem = <UserType extends User | ServiceEntity>(
...
 
   const dataUieValues = {
-    'data-uie-name': isUser ? 'item-user' : 'item-service',
-    'data-uie-value': participantName,
+    'data-uie-name': 'item-user',
+    'data-uie-value': userName,
   };
 
   const commonClassName = cx('participant-item-wrapper', {
-    highlighted,
```
PatrykBuniX: we should not remove this one, it's `highlighted` classname that won't be added now. Should be `highlighted: isHighlighted` now 👌

src/script/components/UserList/UserList.tsx
```diff
@@ -23,7 +23,7 @@ import cx from 'classnames';
 import {container} from 'tsyringe';
 
 import {Icon} from 'Components/Icon';
-import {ParticipantItem} from 'Components/list/ParticipantItem';
+import {UserListItem} from 'Components/UserList/components/UserListItem';
```
atomrc:
```suggestion
import {UserListItem} from './components/UserListItem';
```

src/script/components/UserList/components/UserListItem/UserListItem.tsx
```diff
@@ -298,12 +224,13 @@ const ParticipantItem = <UserType extends User | ServiceEntity>(
...
-          onKeyDown={noInteraction ? handleContextKeyDown : event => onKeyDown(participant, event.nativeEvent)}
-          {...dataUieValues}
-          aria-label={t('accessibility.openConversation', participantName)}
+          aria-label={t('accessibility.openConversation', userName)}
           className={commonClassName}
+          {...(!noInteraction && {
+            onClick: event => onClick(user, event.nativeEvent),
+            onKeyDown: event => onKeyDown(user, event.nativeEvent),
+          })}
+          {...dataUieValues}
```
PatrykBuniX, line 234: Let's maybe move them inline since they are constant values now. 
```
data-uie-name="item-user"
data-uie-value="userName"
```

src/script/components/UserList/components/UserListItem/UserListItem.tsx
```diff
@@ -26,149 +26,115 @@ import {Checkbox, CheckboxLabel} from '@wireapp/react-ui-kit';
...
                       availability={availability}
                       className="participant-item__content__availability participant-item__content__name"
                       dataUieName="status-name"
-                      label={participantName}
+                      label={userName}
                     />
                   )}
 
-                  {(isService || !selfInTeam) && (
+                  {!selfInTeam && (
```
PatrykBuniX: Let's use ternary here 😅 `selfInTeam ? () : ()`

src/script/components/UserList/components/UserListItem/UserListItem.tsx
```diff
@@ -26,149 +26,115 @@ import {Checkbox, CheckboxLabel} from '@wireapp/react-ui-kit';
...
-  const {sharesCamera, sharesScreen, isActivelySpeaking, isMuted} = useKoSubscribableChildren(callParticipant, [
-    'sharesCamera',
-    'sharesScreen',
-    'isActivelySpeaking',
-    'isMuted',
-  ]);
+  const selfString = `(${capitalizeFirstChar(t('conversationYouNominative'))})`;
 
-  const contentInfoText = (() => {
+  const contentInfoText = useMemo(() => {
```
atomrc: Please avoid using `memo` for those. This is cognitive overload and an overhead

atomrc (changes requested):

src/script/components/UserList/components/UserListItem/UserListItem.tsx
PatrykBuniX: Is `useMemo` needed here? They are not too heavy operations and we basically have all the values in deps array, so it will be called every time any of those values change anyway 🤔

src/script/components/UserList/UserList.tsx
```diff
@@ -123,19 +123,19 @@ export const UserList = ({
...
         <li key={user.id}>
-          <ParticipantItem
+          <UserListItem
             noInteraction={noSelfInteraction && user.isMe}
-            participant={user}
+            user={user}
             noUnderline={isLastItem || noUnderline}
-            highlighted={highlightedUserIds.includes(user.id)}
-            customInfo={infos && infos[user.id]}
+            badge={highlightedUserIds.includes(user.id)}
```
atomrc: ⚠️ badge seems to mean something else, I think it should stay `hightlighted`

src/script/components/UserList/components/UserListItem/UserListItem.tsx
przemvs: It's used multiple times.

src/script/components/UserList/UserList.tsx
przemvs: Yup, probably it's missclick with cmd + v.

atomrc (approved):

PatrykBuniX (approved):

PR #14878, opened 2023-03-23
refactor: Make UserRepository less stateful

atomrc:
The `UserRepository` is tighlty coupled to the `UserState`. This class should be stateless and instead the method calls should be done injecting the piece of state that it needs.

### DIFF
--- a/src/script/user/UserService.ts
+++ b/src/script/user/UserService.ts
@@ -124,6 +124,9 @@ export class UserService {
    * @example ['0bb84213-8cc2-4bb1-9e0b-b8dd522396d5', '15ede065-72b3-433a-9917-252f076ed031']
    */
   async getUsers(userIds: QualifiedId[]) {
+    if (userIds.length === 0) {
+      return [];
+    }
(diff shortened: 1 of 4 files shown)


PatrykBuniX (approved):

src/script/user/UserService.ts
```diff
@@ -125,10 +125,9 @@ export class UserService {
    */
   async getUsers(userIds: QualifiedId[]) {
     if (userIds.length === 0) {
-      return Promise.resolve([]);
+      return [];
     }
-    const {found} = await this.apiClient.api.user.postListUsers({qualified_ids: userIds});
-    return found;
+    return (await this.apiClient.api.user.postListUsers({qualified_ids: userIds})).found;
```
przemvs, line 130: I think it looks better (more readable) if you have it in variable :)

atomrc: In a follow up PR, I will actually return the entire result. this is a temporary state for now (I realized it was too much work for this PR alone) 

Next it's going to be 

```js
return this.apiClient.api.user.postListUsers({qualified_ids: userIds});
```

PR #14880, opened 2023-03-23
feat: split ParticipantItem into ServiceListItem

przemvs:
Refactor for ParticipantItem. Now we have single ServiceListItem instead of one huuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu🪵uuuuuuge component :)

### DIFF
--- a/src/script/components/ServiceList/ServiceList.tsx
+++ b/src/script/components/ServiceList/ServiceList.tsx
@@ -33,14 +33,14 @@ export enum MODE {
 
 export interface ServiceListProps {
   services: ServiceEntity[];
-  onClick: (serviceEntity: ServiceEntity) => void;
+  onServiceClick: (serviceEntity: ServiceEntity) => void;
   isSearching?: boolean;
   mode?: MODE;
   dataUieName?: string;
 }
 
 export const ServiceList: React.FC<ServiceListProps> = ({
-  onClick,
+  onServiceClick,
   isSearching = false,
   mode = MODE.DEFAULT,
   services,
@@ -54,7 +54,7 @@ export const ServiceList: React.FC<ServiceListProps> = ({
       {services.map(service => (
         <li key={service.id}>
           <div className="search-list-button" data-uie-name={`service-list-service-${service.id}`}>
-            <ServiceListItem service={service} onClick={onClick} />
+            <ServiceListItem service={service} onClick={onServiceClick} />
           </div>
         </li>
       ))}
--- a/src/script/components/ServiceList/ServiceList.tsx
+++ b/src/script/components/ServiceList/ServiceList.tsx
@@ -17,72 +17,53 @@
  *
  */
 
-import React, {Fragment} from 'react';
+import React from 'react';
 
 import cx from 'classnames';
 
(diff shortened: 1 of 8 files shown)


src/script/components/ServiceList/components/ServiceListItem/ServiceListItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const ServiceListItem = ({service, onClick}: ServiceListItemProps) => {
+  const {name: serviceName} = useKoSubscribableChildren(service, ['name']);
+  const serviceShortDescription = service.summary;
+
+  return (
+    <div
+      tabIndex={TabIndex.FOCUSABLE}
+      role="button"
+      onClick={() => onClick(service)}
+      onKeyDown={event => handleKeyDown(event, () => onClick(service))}
```
atomrc, line 44: I think this behaviour is different than the previous one. since `handlKeyDown` doesn't handle the `SPACE` key (while the previous code did)

przemvs: image
Same :)

przemvs: image

src/script/components/ServiceList/ServiceList.tsx
```diff
@@ -17,72 +17,53 @@
...
+  COMPACT = 'ServiceList.MODE.COMPACT',
+  DEFAULT = 'ServiceList.MODE.DEFAULT',
+}
+
 export interface ServiceListProps {
   services: ServiceEntity[];
-  click: (serviceEntity: ServiceEntity) => void;
-  arrow: boolean;
-  noUnderline: boolean;
+  onClick: (serviceEntity: ServiceEntity) => void;
```
PatrykBuniX: I'd rename it to onServiceClick as it applies to a single service.

src/style/components/list/participant-item.less
```diff
@@ -120,7 +118,6 @@
       &__text {
         display: flex;
         min-width: 0; // this will ensure that ellipses is working
-        max-width: @participant-max-width;
         min-height: @avatar-diameter-m;
         flex-direction: column;
         flex-grow: 1;
```
PatrykBuniX, line 123: Are we sure these style changes don't affect other "types" of participant items we're creating?

src/script/components/ServiceList/ServiceList.tsx
przemvs: Sure, you're right!

src/style/components/list/participant-item.less
przemvs: Yup :) Tested!

src/script/components/ServiceList/components/ServiceListItem/ServiceListItem.tsx
atomrc: Oh right, I looked at another function 👌

atomrc (approved):

PatrykBuniX (approved):

PR #14882, opened 2023-03-23
chore: Create alias to access entities from anywhere

atomrc:
Will allow importing entities with `Entities/...`. 
Allows replacing `src/script/entity` and `../entity` with absolute `Entities/..`


atomrc:
Was a tryout. Might be done later, maybe

PR #14884, opened 2023-03-24
feat: create component for calling actions

PatrykBuniX:


src/script/components/calling/CallParticipantsListItem/CallParticipantStatusIcons.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const {sharesCamera, sharesScreen, isActivelySpeaking, isMuted} = useKoSubscribableChildren(callParticipant, [
+    'sharesCamera',
+    'sharesScreen',
+    'isActivelySpeaking',
+    'isMuted',
+  ]);
+
+  return (
+    <>
+      {sharesScreen && <Icon.Screenshare className="screenshare-icon" data-uie-name="status-screenshare" />}
```
arjita-mitra, line 40: Icons are hidden from screen readers(I believe we have aria-hidden=true} set in the component, what's the alternative you are providing for screen readers?

arjita-mitra: you can turn on screen reader and focus on the icon to listen the announcement

przemvs (approved):

src/script/components/calling/CallParticipantsListItem/CallParticipantStatusIcons.tsx
PatrykBuniX: This is refactoring only PR, we don't want to introduce new features with it. But it's definitely good point.

V-Gira (approved):

PR #14887, opened 2023-03-24
test: call participants list item

PatrykBuniX:


src/script/components/calling/CallingCell.tsx
```diff
@@ -596,7 +596,6 @@ const CallingCell: React.FC<CallingCellProps> = ({
                               callParticipant={participant}
                               selfInTeam={selfUser?.inTeam()}
                               isSelfVerified={isSelfVerified}
-                              external={teamState.isExternal(participant.user.id)}
```
PatrykBuniX, line 599: This one we can get directly from `User` entity (`callParticipant.user`).

przemvs (approved):
LGTM!

PR #14757, opened 2023-02-27
feat: implement retry option on message failure to send [FS-1571]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1571" title="FS-1571" target="_blank">Sub-taskFS-1571</a>  [web] implement retry mechanism for messages that have completely failed to send
  </td></table>
  <br />
 

### Feature

- implement an option to retry sending a message if sending failed
- implementation with this PR only works with text messages, asset messages display a discard button instead
Peek 2023-03-07 12-18

PR #14890, opened 2023-03-24
feat: implement retry mechanism for asset messages [FS-1634]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1634" title="FS-1634" target="_blank">Sub-taskFS-1634</a>  [web] implement retry mechanism for asset messages that have completely failed to send
  </td></table>
  <br />
 

### Feature

- We expand the retry mechanism for text messages https://github.com/wireapp/wire-webapp/pull/14757 to the rest of asset messages (files, audio, video, images)

Peek 2023-03-24 18-23

#### Changes

- We store the asset blob in an  observable in db
- The blob is stored when a message fails to be injected
- When retrying to send the asset message, the upload file method is called with the stored blob and the original message id is conserved to replace the message in db


src/script/entity/message/ContentMessage.ts
```diff
@@ -51,6 +51,7 @@ export class ContentMessage extends Message {
   public readonly other_likes: ko.PureComputed<User[]>;
   public readonly failedToSend: ko.Observable<{queued?: QualifiedUserClients; failed?: QualifiedId[]} | undefined> =
     ko.observable();
+  public readonly storedBlob: ko.Observable<Blob | undefined> = ko.observable();
```
atomrc:
```suggestion
  /** raw content of a file that was supposed to be sent but failed. Is undefined if the message has been successfully sent  */
  public readonly fileData: ko.Observable<Blob | undefined> = ko.observable();
```

atomrc: `Blob` is not really a good name. It's as if you named a field `boolean`, is doesn't hold any application meaning :)

src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -110,6 +111,8 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
         incomingQuote && isOutgoingQuote(incomingQuote) ? (incomingQuote as OutgoingQuote) : undefined;
 
       await messageRepository.sendTextWithLinkPreview(conversation, messageText, mentions, quote, messageId);
+    } else if (file) {
+      await messageRepository.retryUploadFile(conversation, file, firstAsset.isImage(), message.id);
```
atomrc, line 115: Can you confirm that this call will actually remove the `blob` from the database if successful?

src/script/event/EventRepository.ts
```diff
@@ -524,15 +524,28 @@ export class EventRepository {
...
+      return this.eventService.replaceEvent(updatedEvent);
+    };
 
     switch (previewStatus) {
       case ASSET_PREVIEW:
+
+      case RETRY_EVENT: {
+        return updateEvent();
+      }
+
```
atomrc: you could also do like this ;) 

```suggestion
      case RETRY_EVENT: 
```

up to you what you prefer

src/script/storage/record/EventRecord.ts
```diff
@@ -48,6 +48,7 @@ export type UserReactionMap = {[userId: string]: ReactionType};
 type SentEvent = {
   /** sending status of the event*/
   status: StatusType;
+  storedBlob?: Blob;
```
atomrc:
```suggestion
  /** raw content of a file that was supposed to be sent but failed. Is undefined if the message has been successfully sent  */
  fileData?: Blob;
```

src/script/components/MessagesList/Message/MessageWrapper.tsx
V-Gira: It doesn't, I adress it in this commit:  https://github.com/wireapp/wire-webapp/pull/14890/commits/8d84b4d2ef8dbe9c498fb0f8cb1fb5d3f4cb07af

atomrc (approved):

src/script/conversation/MessageRepository.ts
```diff
@@ -490,11 +503,52 @@ export class MessageRepository {
...
+   * @param conversation Conversation to post the file
+   * @param file File object to be stored in db
+   * @param messageId Id of the messsage in db to update
+   * @returns Resolves when database was updated
+   */
+  private async storeFileInDb(conversation: Conversation, messageId: string, file: Blob) {
+    try {
+      const messageEntity = await this.getMessageInConversationById(conversation, messageId);
+      messageEntity.fileData(file);
+      return await this.eventService.updateEvent(messageEntity.primary_key, {
```
przemvs: await is needed here ? :)

src/script/conversation/MessageRepository.ts
```diff
@@ -490,11 +503,52 @@ export class MessageRepository {
...
+        fileData: file,
+      });
+    } catch (error) {
+      if ((error as any).type !== ConversationError.TYPE.MESSAGE_NOT_FOUND) {
```
przemvs, line 545: I think we should handle error better, instead of use error as any :) And we should change it everywhere :)

PR #14916, opened 2023-03-27
fix: Parse conversation roles directly from 'conversation.create' event [FS-1659]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1659" title="FS-1659" target="_blank">BugFS-1659</a>  [Web] Inconsistent display of conversation admin
  </td></table>
  <br />
 

Currently the flow for creating a conversation locally when it was create by someone else if:
- we parse the content of the `conversation.create` event
- we map it to a conversation entity (without the roles)
- we re-fetch the entire conversation metadata
- we extract the roles for that fresh fetch

But, the `conversation.create` event actually contains everything we need to also compute the roles directly. 

This PR removes the extra fetch request and computes the roles directly from the payload the `conversation.create` event

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2394,14 +2394,6 @@ export class ConversationRepository {
 
     const creatorIsParticipant = createdByParticipant || createdBySelfUser;
 
-    const data = await this.conversationService.getConversationById(conversationEntity);
-    const allMembers = [...data.members.others, data.members.self];
-    const conversationRoles = allMembers.reduce<Record<string, string>>((roles, member) => {
-      roles[member.id] = member.conversation_role;
-      return roles;
-    }, {});
-    conversationEntity.roles(conversationRoles);
-
     if (!creatorIsParticipant) {
       (messageEntity as MemberMessage).memberMessageType = SystemMessageType.CONVERSATION_RESUME;
     }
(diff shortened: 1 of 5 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2394,14 +2394,6 @@ export class ConversationRepository {
...
     const creatorIsParticipant = createdByParticipant || createdBySelfUser;
 
-    const data = await this.conversationService.getConversationById(conversationEntity);
-    const allMembers = [...data.members.others, data.members.self];
-    const conversationRoles = allMembers.reduce<Record<string, string>>((roles, member) => {
-      roles[member.id] = member.conversation_role;
-      return roles;
-    }, {});
-    conversationEntity.roles(conversationRoles);
-
```
atomrc, line 2404: This is where we fetched the conversation for a second time and then mapped the roles (this is now done in the `ConversationMapper`)

src/script/conversation/ConversationRoleRepository.ts
```diff
@@ -89,10 +87,6 @@ export class ConversationRoleRepository {
     }
   };
 
-  readonly setConversationRoles = (conversation: Conversation, newRoles: ConversationRole[]): void => {
```
atomrc, line 92: This was never used

przemvs (approved):

V-Gira (approved):

PR #14874, opened 2023-03-22
refactor: Split ParticipantItem Component to separate components

przemvs:

PR #14918, opened 2023-03-27
fix: Correct size for verified icon in call ui (SQSERVICES-1982)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1982" title="SQSERVICES-1982" target="_blank">BugSQSERVICES-1982</a>  [Web] Verified shield in call participant list is very small
  </td></table>
  <br />
 

telegram-cloud-photo-size-4-5839268821192523324-x


przemvs:
This also fixed there: https://github.com/wireapp/wire-webapp/pull/14874

PR #14921, opened 2023-03-27
fix: let the classified-bar scroll with content [FS-1742]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1742" title="FS-1742" target="_blank">BugFS-1742</a>  [Web] Connection request profile photo overlaps on classified/unclassified banner when scrolling down
  </td></table>
  <br />
 

Problematic behavior:

bug

Fix:

fix


tlebon (approved):
still not sure what was wrong with the old one 🙃

thisisamir98 (approved):

src/script/components/ModalComponent.tsx
```diff
@@ -72,7 +72,7 @@ const ModalContentStyles: CSSObject = {
   maxHeight: 615,
   overflow: 'hidden',
   overflowY: 'auto',
-  padding: 16,
+  padding: 0,
```
atomrc: Are you confident this didn't change other modal's styles 😬 ?

src/style/modal/user-modal.less
```diff
@@ -27,4 +27,17 @@
...
+  }
+
+  .modal__body {
+    padding: 16px;
+
+    .classified-bar {
+      width: calc(100% + 32px);
+      margin-top: 12px;
+    }
+  }
```
atomrc: Is there anyway we could colocate those styles next to the targeted component in a `Component.styles.ts` file?

src/script/components/ModalComponent.tsx
aweiss-dev: Good question. I need to check 🥶

atomrc (approved):

PR #14922, opened 2023-03-28
feat: replace LESS with emotion css

przemvs:
Refactor LESS style to Emotion CSS.

### DIFF
--- a/src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
+++ b/src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
@@ -20,12 +20,24 @@
 import React from 'react';
 
 import {TabIndex} from '@wireapp/react-ui-kit/lib/types/enums';
-import cx from 'classnames';
 
 import {Availability} from '@wireapp/protocol-messaging';
 
 import {AvailabilityState} from 'Components/AvailabilityState';
 import {Icon} from 'Components/Icon';
+
+import {
+  contentInfoWrapper,
+  contentInfoText,
+  selfIndicator,
+  userName,
+  userAvailability,
+  ellipsis,
+  nameWrapper,
+  chevronIcon,
+  contentText,
+  wrapper,
+} from './ParticipantItemContent.styles';
 
 export interface ParticipantItemContentProps {
   name: string;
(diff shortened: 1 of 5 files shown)


src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
```diff
@@ -101,7 +112,7 @@ export const ParticipantItemContent = ({
         </button>
       )}
 
-      {showArrow && <Icon.ChevronRight className="chevron-right-icon participant-item__content__chevron" />}
```
przemvs, line 104: chevron-right-icon not exist in the code and is not necessary.

src/style/content/preferences/account.less
```diff
@@ -177,7 +177,7 @@
 .preferences-account-username-wrapper {
   position: relative;
   width: 256px;
-  min-height: @line-height-xl * 2;
+  min-height: calc(@line-height-xl * 2);
```
atomrc, line 180: why do you need a `calc` here? Doesn't seem to be a dynamic value 🤔

V-Gira: need to use calc with the css variable is my guess

`@line-height-xl` is no longer a rem value but `var(--line-height-xl)`

atomrc: But there is no css variable in there, only `less` variable (which are computed at compile time)

atomrc: ooohhhh the `less` variable contains a `css`  variable. 
Humm then we should probably use the css variable here, and avoid having the `less` proxy. WDYT?

src/style/components/list/participant-item.less
```diff
@@ -100,97 +96,6 @@
...
-      &__name-wrapper {
-        display: flex;
-        overflow: hidden;
-        width: 100%;
-        padding-right: 8px;
-        color: var(--main-color);
-      }
-
-      &__name {
-        .ellipsis;
```
atomrc, line 160: Are you absolutely sure those selectors are not used anywhere else?

src/style/content/preferences/account.less
V-Gira: yes, this makes sense

src/style/components/list/participant-item.less
V-Gira: after looking at `ParticipantItemContent.tsx` , looks good to me

atomrc: But that's the thing, you need to look at all the places that are **not** `ParticipantItemContent.tsx` and check that those classes do not exist somewhere else before we can safely remove them from the CSS file. 

That's the main problem with CSS, it's global and never scoped so removing declarations requires some rigorous checks

V-Gira: You're right, I'm forgetting that you could pass say `.participant-item__content__chevron` outside of  `ParticipantItemContent.tsx` if you wanted to

src/style/content/preferences/account.less
przemvs: Sure, we should avoid proxies with started with '@' and replace to css variables :) @atomrc

src/style/components/list/participant-item.less
przemvs: All fine there. :) In another PR the participant-item.less is completely removed. :)

atomrc (approved):

PR #14923, opened 2023-03-28
feat: Detect users that could not be listed on federated backends

atomrc:
This will add some `PlaceholderUser` in the live user list that will account for users that could not be fetched on backend (because of federation backend failure). 

This has no UI changes yet, a follow up PR will start showing those `PlaceholderUser`


src/script/entity/User/PlaceholderUser.ts
```diff
@@ -0,0 +1,31 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {QualifiedId} from '@wireapp/api-client/lib/user';
+
+import {User} from './User';
+
+export class PlaceholderUser extends User {
```
atomrc, line 24: This class is not consumed yet but will be used in the UI to switch rendering when dealing with a placeholder

V-Gira (approved):

PR #14927, opened 2023-03-29
feat: replace list participant styles with emotion

przemvs:
Replace whole list participant styles with emotion.

image

image

image

image

image


src/style/components/list/conversation-list-calling-cell.less
```diff
@@ -232,12 +203,12 @@ svg.small-icon {
...
+  .group-video-grid {
     height: 168px;
   }
 }
 
-conversation-list-calling-cell:first-of-type {
+.conversation-list-calling-cell:first-of-type {
   &::before {
     top: -40px;
   }
```
atomrc: Can those be removed? They probably didn't match anything anymore, right?

atomrc (approved):

V-Gira (approved):

PR #14928, opened 2023-03-29
feat: Show placeholder message header for unavailable users

atomrc:
Will show a placeholder header when a user's metadata could not be fetched due to some federated backend being offline

image


src/script/components/MessagesList/Message/ContentMessage/MessageHeader.tsx
```diff
@@ -18,82 +18,111 @@
...
+      {sender.isDirectGuest() && (
+        <span
+          className="message-header-icon-guest with-tooltip with-tooltip--external"
+          data-tooltip={t('conversationGuestIndicator')}
+          data-uie-name="sender-guest"
+        >
+          <Icon.Guest />
+        </span>
+      )}
+    </>
```
tlebon, line 80: seems like these could all be combined into one function that returned the correct values from an object with a key that is equivalent to the sender?

V-Gira: something similar has been done here https://github.com/wireapp/wire-webapp/pull/14883 but not merged to dev yet. I think the idea is to come back and see if we can use the same reusable component later

atomrc: Yes, my plan is to use this one once it's merged. I'll have to identify if the contraints are really the same and go for it :)

tlebon: even #14883 seems a bit less DRY than id expect. the span component is duplicated several times.

tlebon: anyway thats fine then.

przemvs (approved):

PR #14930, opened 2023-03-30
fix: Avoid initialising DataDog multiple times

przemvs:
Datadog initializing multiple times.

### DIFF
--- a/src/script/util/DataDog.ts
+++ b/src/script/util/DataDog.ts
@@ -21,7 +21,15 @@ import {Configuration} from '../Config';
 
 const uuidRegex = /([a-z\d]{8})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{4})-([a-z\d]{12})/gim;
 
+let isDataDogInitialized = false;
+
 export async function initializeDataDog(config: Configuration, domain?: string) {
+  if (isDataDogInitialized) {
+    return;
+  }
+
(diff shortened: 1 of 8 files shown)


src/script/util/DataDog.ts
```diff
@@ -65,6 +72,7 @@ export async function initializeDataDog(config: Configuration, domain?: string)
     forwardErrorsToLogs: true,
     forwardConsoleLogs: ['info', 'warn', 'error'], // For now those logs should be fine, we need to investigate if we need another logs in the future
     sessionSampleRate: 100,
+    silentMultipleInit: true,
```
atomrc: Let's remove this one, cause we would like to be warned in case we init it twice (this would be a mistake)
```suggestion
```

src/script/util/DataDog.ts
```diff
@@ -46,15 +54,14 @@ export async function initializeDataDog(config: Configuration, domain?: string)
     version: config.VERSION,
     sessionSampleRate: 100,
     sessionReplaySampleRate: 20,
+    silentMultipleInit: true,
```
atomrc: Let's remove this one, cause we would like to be warned in case we init it twice (this would be a mistake)
```suggestion
    silentMultipleInit: true,
```

atomrc (approved):

PR #14933, opened 2023-03-30
feat: Add tooltip to reaction buttons

thisisamir98:
image

image


src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactions.styles.ts
```diff
@@ -55,6 +55,15 @@ export const messageReactionButton: CSSObject = {
   verticalAlign: 'top',
 };
 
+export const messageReactionButtonTooltip: CSSObject = {display: 'flex', maxWidth: 130, whiteSpace: 'break-spaces'};
+export const messageReactionButtonTooltipImage: CSSObject = {marginRight: 8};
+export const messageReactionButtonTooltipText: CSSObject = {fontSize: '0.7rem'};
+export const messageReactionButtonTooltipTextLink: CSSObject = {
+  color: 'var(--blue-500)',
```
arjita-mitra, line 62: May be its accent colour not always blue. Please confirm.

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -277,7 +277,11 @@ const ContentMessageComponent: React.FC<ContentMessageProps> = ({
         )}
       </div>
 
-      <MessageReactionsList reactions={reactions} handleReactionClick={onClickReaction} />
+      <MessageReactionsList
+        reactions={reactions}
+        onUsersClick={() => onClickLikes(message)}
```
arjita-mitra: onUsersClick doesn't clearly describe the purpose of the method. onUserReactioncountClick/onTooltipReactionCount is more specific. I know we are already using onClickLikes in this component, but its kind of misleading what happening on click, can you rename it to showMessageDetails/showReactionDetails?

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactionsList.tsx
```diff
@@ -47,20 +56,51 @@ const MessageReactionsList: FC<MessageReactionsListProps> = ({reactions, handleR
...
+                    css={messageReactionButtonTooltipImage}
+                    emojiUrl={emojiUrl}
+                    emojiName={emojiName}
+                  />
+                  <p css={messageReactionButtonTooltipText}>
+                    <span
+                      onClick={onUsersClick}
+                      onKeyDown={event => {
+                        if (event.key === KEY.ENTER) {
+                          onUsersClick();
```
arjita-mitra: can be replaced with isEnterKey util

arjita-mitra (changes requested):

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactions.styles.ts
thisisamir98: it is always var blue, both in dark and light mode.

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/MessageReactionsList.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/14933/commits/adea797d32bf24dab3386072f4b67960cc5d0c16

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/14933/commits/f6a54bb9c9dfa6ed0c4887a8af70e3adb827d9a3

and 
https://github.com/wireapp/wire-webapp/pull/14933/commits/c814a56b837ae3977d5fb6bb176a57f3488e966e

atomrc (approved):

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
atomrc: Just a side note on this, cause this is actually important. It boils down to responsibility across components of the app. 

The rule of thumb is:

- a component that accepts a `callback`  should name it `on<UserAction>` (like `onClickAvatar`, `onKeyPress`...)
- the part where the callback is **given** should give an imperative name to the callack `onClickAvatar={showUser}``

This way we tell a little story when we read the intantiation of a particular component

```
<MyComponent onAvatarClick={showUser} />
```

I can understand what is happening just from reading that line of code. I know that I'm supposed to show the user when a click on the avatar is done. 

Which is not the case if I do `onAvatarClick={onAvatarClick}`. This piece of code doesn't convey any meaning a we cannot guess what `onAvatarClick` is doing

arjita-mitra (approved):

PR #14934, opened 2023-03-30
fix: display classified bar correctly [FS-1758]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1758" title="FS-1758" target="_blank">BugFS-1758</a>  [web] Enlarge font slider to 24px make classified banner not readable in call window 
  </td></table>
  <br />
 

#### Issues

- Classified banners disappears from the call ui at certain responsive breakpoints
- Pagination would overlay the banner leading to a broken UI
- Classified banner in the side bar call ui would not fit the text at larger font sizes

Peek 2023-03-30 11-24

#### Solutions

- Do not rely on absolute positioning for full screen call ui
Peek 2023-03-30 13-52

- Remove static height from classified banner for sidebar (Doesn't look good at the max font size, there's always limit on what you can fit inside  a box ;) )
Peek 2023-03-30 14-07

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -42,7 +42,8 @@ import {
   videoControlInActiveStyles,
   videoControlDisabledStyles,
   paginationButtonStyles,
-} from './FullscreenVideoCallStyles';
+  classifiedBarStyles,
+} from './FullscreenVideoCall.styles';
 import {GroupVideoGrid} from './GroupVideoGrid';
 import {Pagination} from './Pagination';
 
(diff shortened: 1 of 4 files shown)


atomrc (approved):

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -270,6 +254,16 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
...
           setMaximizedParticipant={participant => setMaximizedParticipant(call, participant)}
         />
+        {classifiedDomains && (
+          <ClassifiedBar
+            users={allUsers}
+            classifiedDomains={classifiedDomains}
+            style={{
+              lineHeight: '1.5em',
+              display: 'flex',
+            }}
```
przemvs: Move this styles to file with styles **FullscreenVideoCallStyles**, and maybe change it to FullscreenVideoCall.styles.ts :)

PR #14968, opened 2023-04-04
refactor: make call participant list item content separate component

PatrykBuniX:
Follow up to https://github.com/wireapp/wire-webapp/pull/14874.

Removes `ParticipantItemContent` from `CallParticipantsListItem`. I've created separate component that will take care of displaying content for `CallParticipantsListItem` only. Feature I'm working on would require adding another prop to generic `ParticipantItemContent` that is calling specific (audio establish state) and adding some calling related conditional rendering. We don't want this generic component to grow too big and contain not-generic props.

### DIFF
--- a/src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.styles.ts
+++ b/src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.styles.ts
@@ -1,0 +1,100 @@
+/*
+ * Wire
+ * Copyright (C) 2023 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CSSObject} from '@emotion/react';
+
+export const listItem = (noInteraction = false): CSSObject => ({
+  display: 'flex',
+  overflow: 'hidden',
+  height: '56px',
+  alignItems: 'center',
+  paddingRight: '16px',
+  margin: '0',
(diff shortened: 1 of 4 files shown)


src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
```diff
@@ -93,19 +90,6 @@ export const ParticipantItemContent = ({
         )}
       </div>
 
-      {onDropdownClick && (
```
PatrykBuniX, line 96: This one was used only for calling list item

src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+              dataUieName="status-name"
+              label={name}
+            />
+          ) : (
+            <div css={[userName, ellipsis]} data-uie-name="status-name">
+              {name}
+            </div>
+          )}
+
+          {selfString && <div css={selfIndicator}>{selfString}</div>}
```
atomrc, line 76:
```suggestion
          {isSelf && <div css={selfIndicator}>{selfString}</div>}
```

looks like a mistake :)

src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.styles.ts
atomrc, line 1: If those styles exists in a css files, could we delete them from there?

src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.tsx
PatrykBuniX: Yeah! Good catch

src/script/components/calling/CallParticipantsListItem/CallParticipantItemContent/CallParticipantItemContent.styles.ts
PatrykBuniX: They were copied from `ParticipantItemContent` component's `.styles.ts` file. They are the same for now, but I'd rather have them not shared between `ParticipantItemContent` and `CallParticipantItemContent`. Modifying styles of one could break the other one + `CallParticipantItemContent` will include some extra styles with established audio indication. WDYT?

atomrc: Fair enough 👌

atomrc (approved):

PR #14971, opened 2023-04-05
fix: Show Availability state for users

przemvs:
Prev: The Availability state is not shown for User list item,
Now: Proper display for UserListItem, where we should display AvailabilityState component instead of only userName .

### DIFF
--- a/src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
+++ b/src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
@@ -46,7 +46,7 @@ export interface ParticipantItemContentProps {
   hasUsernameInfo?: boolean;
   showArrow?: boolean;
   onDropdownClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
-  isUserContent?: boolean;
+  showAvailabilityState?: boolean;
 }
 
 export const ParticipantItemContent = ({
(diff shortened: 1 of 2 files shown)


src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
```diff
@@ -56,12 +57,13 @@ export const ParticipantItemContent = ({
...
   hasUsernameInfo = false,
   showArrow = false,
+  isUserContent = false,
 }: ParticipantItemContentProps) => {
   return (
     <div css={wrapper}>
       <div css={contentText}>
         <div css={nameWrapper}>
-          {selfInTeam && availability ? (
+          {isUserContent && selfInTeam ? (
```
przemvs: Before refactor we have it like that:

image

src/script/components/ParticipantItemContent/ParticipantItemContent.tsx
```diff
@@ -46,6 +46,7 @@ export interface ParticipantItemContentProps {
   hasUsernameInfo?: boolean;
   showArrow?: boolean;
   onDropdownClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  isUserContent?: boolean;
```
atomrc: I think we can drop this since now we know it's always going to be a `user` isn't it?

przemvs: Nah, this component is used in 2 places. So maybe i rename it for **showAvailabilityState** ?

atomrc: yes that would be better

atomrc (approved):

PR #14977, opened 2023-04-05
fix: Only allow remote user search on group creation and add participants [SQSERVICES-2001]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-2001" title="SQSERVICES-2001" target="_blank">BugSQSERVICES-2001</a>  [Web] Incorrect user display in Group search results
  </td></table>
  <br />
 

Currently the webapp is performing a remote search even on fixed user lists (like participant of a conversation). 

This should only happen if we want to add new people to a conversation. 

## Before 

(Here `Florian Cherie` doesn't belong in this conversation and should not appear in the search result)

https://user-images.githubusercontent.com/1090716/230125267-db10a481-1354-4933-ba75-e380ebcac1fd.mov

## After

`Florian Cherie` is not showing anymore in the search results

https://user-images.githubusercontent.com/1090716/230125396-3643103c-2bb6-46dd-b86b-80e80a918b45.mov

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -366,6 +366,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
                 teamRepository={teamRepository}
                 conversationRepository={conversationRepository}
                 noUnderline
+                allowRemoteSearch
               />
             )}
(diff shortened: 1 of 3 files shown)


aweiss-dev (approved):

arjita-mitra (approved):

tlebon (approved):
seems fine to me

PR #15002, opened 2023-04-12
feat: Store user's metadata in database [FS-1828]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1828" title="FS-1828" target="_blank">Sub-taskFS-1828</a>  [web] Store users in Database and only load unknown users
  </td></table>
  <br />
 

This will make sure that we download a user only once. 

- at first start, all the users will be downloaded and stored into the database
- at subsequent loads, only the users that are listed (in connections or team members) that we do not have locally will be downloaded
- when a user is updated (via a `websocket` `user.update` message), we update the in-memory user and the database entry

This will:
- speedup webapp loading time (since all connections and team members won't be downloaded anymore)
- allow better offline support


src/script/main/app.ts
```diff
@@ -392,30 +392,30 @@ export class App {
...
       telemetry.timeStep(AppInitTimingsStep.INITIALIZED_CRYPTOGRAPHY);
 
-      const team = await teamRepository.initTeam(selfUser.teamId);
+      const {members: teamMembers} = await teamRepository.initTeam(selfUser.teamId);
+      telemetry.timeStep(AppInitTimingsStep.RECEIVED_USER_DATA);
+
+      const connections = await connectionRepository.getConnections();
+      telemetry.addStatistic(AppInitStatisticsValue.CONNECTIONS, connections.length, 50);
+
+      await userRepository.loadUsers(connections, [...teamMembers, selfUser.qualifiedId]);
```
atomrc, line 401: This is the major change here. Now `getConnections` and `initTeam` do not fetch user's metadata anymore (doing less side-effects). 
Instead they return the userIds that needs to be loaded later on

The `userRepository.loadUsers` is responsible for loading the users from the DB and fetch the missing users on backend

src/script/connection/ConnectionRepository.test.ts
```diff
@@ -0,0 +1,121 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
thisisamir98:
```suggestion
 * Copyright (C) 2023 Wire Swiss GmbH
```

atomrc: humm technically this is not a new file (but git doesn't understand it 🙃 ). 
Will still update the date :)

src/script/user/UserRepository.ts
```diff
@@ -161,33 +168,45 @@ export class UserRepository {
...
   /**
-   * Will load the availability status to the team users and subscribe to changes.
+   * Will load all the users in memory (and save new users to the database).
    */
-  async loadTeamUserAvailabilities(teamMembers: User[]): Promise<void> {
-    if (!teamMembers.length) {
-      return;
-    }
+  async loadUsers(connections: ConnectionEntity[], extraUsers: QualifiedId[]): Promise<User[]> {
+    const users = connections.map(connectionEntity => connectionEntity.userId).concat(extraUsers);
```
przemvs, line 184: Why not:

```suggestion
    const users = [...connections.map(connectionEntity => connectionEntity.userId), ...extraUsers];
```

src/script/user/UserRepository.ts
```diff
@@ -482,22 +486,44 @@ export class UserRepository {
...
-    const resolveArray = await Promise.all(chunksOfUserIds.map(getUsers));
-    const newUserEntities = flatten(resolveArray);
+  }
+
+  private mapUserResponse(found: APIClientUser[], failed: QualifiedId[]): User[] {
+    const failedToLoad = failed.map(userId => {
+      /* When a federated backend is unreachable, we generate placeholder users locally with some default values
+       */
+      return new User(userId.id, userId.domain);
+    });
```
przemvs: If you remove return it won't be a 1 line code ? :)

src/script/user/UserRepository.ts
```diff
@@ -482,22 +486,44 @@ export class UserRepository {
...
+    const mappedUsers = this.userMapper.mapUsersFromJson(found).concat(failedToLoad);
```
przemvs, line 519:
```suggestion
    const mappedUsers = [...this.userMapper.mapUsersFromJson(found), ...failedToLoad];
```

src/script/user/UserGuards.ts
```diff
@@ -0,0 +1,24 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {Self as SelfAPIUser} from '@wireapp/api-client/lib/self';
+
+export function isSelfAPIUser(payload: any): payload is SelfAPIUser {
```
thisisamir98: isn't it better to change this payload: any to payload: StoredUser?

atomrc: Typeguards usually take anything. I could actually convert it to `unknown` :)

atomrc: Arf, actually it makes things more complicated with `unknown` 😓

thisisamir98: `unknown` makes more sense

src/script/user/UserRepository.ts
atomrc: fixed :) https://github.com/wireapp/wire-webapp/pull/15002/commits/1935ef696a460baed022f1b085ecb4cc36f1cb36

src/script/user/UserGuards.ts
atomrc: Alright, maybe you are right, it makes sense to narrow it down :)

PatrykBuniX (approved):

thisisamir98 (approved):

PR #15005, opened 2023-04-12
runfix: Migrate previously stored availability to new stored users

atomrc:
migrates, at load time, from DB entries from this (previous format that only has the `availability`)

image

to this (new format that has the entire user's metadata)

image

### DIFF
--- a/src/script/user/UserMapper.ts
+++ b/src/script/user/UserMapper.ts
@@ -82,7 +82,7 @@ export class UserMapper {
    */
   updateUserFromObject(userEntity: User, userData: Partial<StoredUser>): User {
     // We are trying to update non-matching users
-    const isUnexpectedId = userEntity.id !== '' && userData.id !== userEntity.id;
+    const isUnexpectedId = userEntity.id && userData.id && userData.id !== userEntity.id;
     if (isUnexpectedId) {
       throw new Error(`Updating wrong user entity. User '${userEntity.id}' does not match data '${userData.id}'.`);
     }
@@ -180,9 +180,9 @@ export class UserMapper {
 
     if (ssoId && Object.keys(ssoId).length) {
       userEntity.isSingleSignOn = true;
-    }
-    if (ssoId?.subject) {
-      userEntity.isNoPasswordSSO = true;
+      if (ssoId.subject) {
+        userEntity.isNoPasswordSSO = true;
+      }
     }
 
     if (teamId && !userEntity.isFederated) {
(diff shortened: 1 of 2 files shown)


src/script/user/UserMapper.ts
```diff
@@ -82,7 +82,7 @@ export class UserMapper {
    */
   updateUserFromObject(userEntity: User, userData: Partial<StoredUser>): User {
     // We are trying to update non-matching users
-    const isUnexpectedId = userEntity.id !== '' && userData.id !== userEntity.id;
+    const isUnexpectedId = userEntity.id && userData.id && userData.id !== userEntity.id;
```
przemvs, line 85:
```suggestion
    const isUnexpectedId = userData?.id !== userEntity?.id;
```

src/script/user/UserRepository.ts
```diff
@@ -196,16 +198,25 @@ export class UserRepository {
       await this.userService.removeUserFromDb(user.qualified_id);
     });
 
-    const missingUsers = users.filter(
-      user => !liveUsers.find(localUser => matchQualifiedIds(user, localUser.qualified_id)),
-    );
+    const missingUsers = users
+      .filter(user => !liveUsers.find(localUser => matchQualifiedIds(user, localUser.qualified_id)))
+      // we add the users that are locally incomplete
+      .concat(incompleteUsers.map((user: any) => ({id: user.id, domain: user.domain})));
```
przemvs, line 204: You can use you util to fill user id && domain ?

src/script/user/UserMapper.ts
atomrc: This is different behavior, I want to check that both the ID are defined before checking that they are different. 

Basically I never want to compare `undefined` to a string. 

If they one of them is falsy I want to boolean to be `false` :)

src/script/user/UserRepository.ts
atomrc: not sure it's worth it here, since there is not particular logic to it 🤔

src/script/user/UserRepository.ts
```diff
@@ -183,7 +183,9 @@ export class UserRepository {
   async loadUsers(connections: ConnectionEntity[], extraUsers: QualifiedId[]): Promise<User[]> {
     const users = connections.map(connectionEntity => connectionEntity.userId).concat(extraUsers);
     // TODO migrate old user entries that only have availability
-    const localUsers = await this.userService.loadUserFromDb();
+    const dbUsers = await this.userService.loadUserFromDb();
+    /* prio to April 2023, we were only storing the availability in the DB, we need to refetch those users */
```
V-Gira:
```suggestion
    /* prior to April 2023, we were only storing the availability in the DB, we need to refetch those users */
```

przemvs (approved):

src/script/user/UserRepository.ts
atomrc: woups good catch 👍

PR #15009, opened 2023-04-13
runfix: diameter avatar css variable

przemvs:
Fix for not centered user name in list

### DIFF
--- a/src/style/common/variables.less
+++ b/src/style/common/variables.less
@@ -281,7 +281,7 @@ body {
 // Avatar
 // ---------------------------------------------------------------------------
 body {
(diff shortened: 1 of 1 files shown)


PatrykBuniX (approved):

V-Gira (approved):

src/style/common/variables.less
```diff
@@ -281,7 +281,7 @@ body {
 // Avatar
 // ---------------------------------------------------------------------------
 body {
-  --avatar-diameter-m: '40px';
+  --avatar-diameter-m: 40px;
```
przemvs, line 284: Value can't be passed in ''. It breaks usage.

PR #15010, opened 2023-04-13
refactor: Streamline backup export process

atomrc:
This factorize a bunch of processing that we do on the exported entities and streamlines the process

### DIFF
--- a/src/script/backup/BackupRepository.ts
+++ b/src/script/backup/BackupRepository.ts
@@ -33,14 +33,13 @@ import {
   InvalidMetaDataError,
 } from './Error';
 
-import {ClientState} from '../client/ClientState';
 import {ConnectionState} from '../connection/ConnectionState';
 import type {ConversationRepository} from '../conversation/ConversationRepository';
 import type {Conversation} from '../entity/Conversation';
+import {User} from '../entity/User';
 import {ClientEvent} from '../event/Client';
 import {ConversationRecord} from '../storage/record/ConversationRecord';
 import {StorageSchemata} from '../storage/StorageSchemata';
-import {UserState} from '../user/UserState';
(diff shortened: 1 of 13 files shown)


src/script/backup/BackupRepository.ts
```diff
@@ -78,8 +77,6 @@ export class BackupRepository {
   constructor(
     backupService: BackupService,
     conversationRepository: ConversationRepository,
-    private readonly clientState = container.resolve(ClientState),
-    private readonly userState = container.resolve(UserState),
```
atomrc, line 82: The repository doesn't depend on state now, instead the client and user are injected in the methods that require them

src/script/backup/BackupService.ts
```diff
@@ -60,8 +58,8 @@ export class BackupService {
     return recordsPerTable.reduce((accumulator, recordCount) => accumulator + recordCount, 0);
   }
 
-  getTables(): Dexie.Table<any, string>[] {
-    return this.storageService.getTables(BackupService.CONFIG.SUPPORTED_TABLES);
+  getTables() {
+    return [this.storageService.db!.conversations, this.storageService.db!.events] as const;
```
atomrc, line 62: This greatly improve typings (as those tables are typed with the type of record stored in it)

test/helper/TestFactory.js
```diff
@@ -92,25 +90,6 @@ export class TestFactory {
     return this.storage_repository;
   }
 
-  /**
-   * @returns {Promise<BackupRepository>} The backup repository.
-   */
-  async exposeBackupActors() {
```
atomrc, line 98: Not used anymore (since I migrated the test to TS)

PatrykBuniX (approved):

PR #15012, opened 2023-04-13
refactor: Upgrade worker dependencies and migrate to TypeScript

atomrc:
This will allow the `zip` worker to be loaded only once and will migrate all the workers to typescript


package.json
```diff
@@ -21,6 +21,7 @@
     "dexie-batch": "0.4.3",
     "highlight.js": "11.7.0",
     "http-status-codes": "2.2.0",
+    "jimp": "0.22.7",
```
atomrc, line 24: This was hard copied in the repo, it's now loaded from npm and will benefit from automatic updates :)

babel.config.js
```diff
@@ -27,7 +27,7 @@ const presetEnvConfig = {
 module.exports = {
   env: {
     test: {
-      plugins: ['@emotion'],
+      plugins: ['@emotion', 'babel-plugin-transform-import-meta'],
```
atomrc, line 30: This will fix the problem with `import.meta.url` in the tests (see https://stackoverflow.com/a/70640363)

PatrykBuniX (approved):

arjita-mitra (approved):

przemvs (approved):

PR #15035, opened 2023-04-18
runfix: Make sure conversation users are loaded at app load time

atomrc:
When we load the app we need to load 3 types of users:
- direct connections
- team members
- participants of conversations we are part of (that are not team members nor direct connections)

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -740,7 +740,8 @@ export class ConversationRepository {
    */
   private async getUnreadEvents(conversationEntity: Conversation): Promise<void> {
     const first_message = conversationEntity.getOldestMessage();
-    const lower_bound = new Date(conversationEntity.last_read_timestamp());
+    // The lower bound should be right after the last read timestamp in order not to load the last read message again (thus the +1)
+    const lower_bound = new Date(conversationEntity.last_read_timestamp() + 1);
     const upper_bound = first_message
       ? new Date(first_message.timestamp())
       : new Date(conversationEntity.getLatestTimestamp(this.serverTimeHandler.toServerTimestamp()) + 1);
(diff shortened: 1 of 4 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -740,7 +740,8 @@ export class ConversationRepository {
    */
   private async getUnreadEvents(conversationEntity: Conversation): Promise<void> {
     const first_message = conversationEntity.getOldestMessage();
-    const lower_bound = new Date(conversationEntity.last_read_timestamp());
+    // The lower bound should be right after the last read timestamp in order not to load the last read message again (thus the +1)
+    const lower_bound = new Date(conversationEntity.last_read_timestamp() + 1);
```
atomrc, line 744: This is a major performance fix. 
At load time, right before showing the UI, the webapp was loading 1 message per conversation!!! That could mean a lot of messages if you are part of many discussions

src/script/conversation/ConversationRepository.ts
```diff
@@ -740,7 +740,8 @@ export class ConversationRepository {
    */
   private async getUnreadEvents(conversationEntity: Conversation): Promise<void> {
     const first_message = conversationEntity.getOldestMessage();
-    const lower_bound = new Date(conversationEntity.last_read_timestamp());
+    // The lower bound should be right after the last read timestamp in order not to load the last read message again (thus the +1)
```
przemvs, line 743: thus or this ? :)

atomrc: thus https://dictionary.cambridge.org/dictionary/english/thus

przemvs (approved):
LGTM! :)

src/script/conversation/ConversationRepository.ts
przemvs: never heard about this word :D

PR #15041, opened 2023-04-19
runfix: Avoid trying to load deleted users when initially loading conversations

atomrc:
At load time we update all the local conversation (adding `users` and `messages` entities to it). 
Thing is, some of those users will not be in the state (because they are `deleted` and won't be listed by backend). 

In order to avoid sending a storm of user requests (per each message that has a sender not in the app state), we can load the initial messages as `offline` and consider that a user that is not found locally as a `deleted` user

## Before 😬 
image

## After ✨ 
image


przemvs (approved):
code lgtm

PatrykBuniX (approved):

PR #15043, opened 2023-04-19
runfix: Make sure fetched users are added to the state

atomrc:
Regression introduced by #15041

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -615,6 +615,11 @@ export class UserRepository {
     return userData;
   }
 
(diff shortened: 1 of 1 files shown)


src/script/user/UserRepository.ts
```diff
@@ -629,7 +634,7 @@ export class UserRepository {
       return deletedUser;
     }
     try {
-      return this.userMapper.mapUserFromJson(await this.userService.getUser(userId));
+      return this.fetchUser(userId);
```
atomrc, line 637: This will make sure the user is saved once downloaded

PatrykBuniX (approved):

PR #15046, opened 2023-04-20
runfix: Fix sync between live self user and DB data

atomrc:
This fixes synching the current user's state with what is in the DB and will set the initial availability upon loading the app


src/script/user/UserRepository.ts
```diff
@@ -435,13 +444,17 @@ export class UserRepository {
...
+      return;
+    }
+    const hasAvailabilityChanged = availability !== selfUser.availability();
     const newAvailabilityValue = valueFromType(availability);
     if (hasAvailabilityChanged) {
-      const oldAvailabilityValue = valueFromType(this.userState.self().availability());
+      const oldAvailabilityValue = valueFromType(selfUser.availability());
       this.logger.log(`Availability was changed from '${oldAvailabilityValue}' to '${newAvailabilityValue}'`);
-      this.userState.self().availability(availability);
+      await this.updateUser(selfUser.qualifiedId, {availability});
```
atomrc, line 457: This is the main change, here we also save the value to the DB (previously it would not write to the DB)

src/script/view_model/ListViewModel.ts
```diff
@@ -98,7 +96,6 @@ export class ListViewModel {
 
     this.isActivatedAccount = this.userState.isActivatedAccount;
     this.isProAccount = this.teamState.isTeam;
-    this.selfUser = this.userState.self;
```
atomrc: This was never used in the class. Probably some artefacts of knockout bindings

src/script/user/UserRepository.ts
```diff
@@ -249,6 +252,12 @@ export class UserRepository {
       }
     });
 
+    // Map self user's availability status
+    const selfUserAvailability = dbUsers.find(user => user.id === selfUser.id);
```
PatrykBuniX: This one shouldn't be called availability, it seems to be user's db record, maybe just `selfUser` wdyt?

atomrc: Humm that's indeed good feedback. I named it this way because I only use the availability from it, but you're right, it's misleading 👌

PatrykBuniX (approved):

V-Gira (approved):

PR #15048, opened 2023-04-20
refactor: Make self user a reactive property

atomrc:
Currently the `selfUser` that we put in the `userState` is an impertive value and the `selfUser` is actually duplicated in the state (since it's also in the `users` list). 

Now the `self` is directly computed from the list of `users` we have in memory


src/script/user/UserRepository.ts
```diff
@@ -697,7 +698,6 @@ export class UserRepository {
     if (!user) {
       if (isMe) {
         userEntity.isMe = true;
-        this.userState.self(userEntity);
```
atomrc, line 700: No need to set the self user anymore, it's automatically found in the users list

PatrykBuniX (approved):

przemvs (approved):

PR #15069, opened 2023-04-24
runfix: Allow reseting user availabilities

atomrc:
Switching from a particular availability to a `0` value availability would prevent the user from being updated. 
We should consider falsy value as valid values for availabilities

### DIFF
--- a/src/script/user/UserMapper.ts
+++ b/src/script/user/UserMapper.ts
@@ -19,6 +19,8 @@
 
 import {container} from 'tsyringe';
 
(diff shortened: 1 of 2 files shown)


src/script/user/UserMapper.ts
```diff
@@ -125,9 +127,7 @@ export class UserMapper {
       userEntity.accent_id(accentId);
     }
 
-    if (availability) {
```
atomrc, line 128: If availability was `0` (which is a possible value), the user would not be updated

V-Gira (approved):

PR #15073, opened 2023-04-24
refactor: Remove dependency between UserMapper and UserState

atomrc:
This is an effort to reduce the impact of the `UserState` across the app. 
The ultimate goal is to have the UserState only consumed by the UI (not the repository or any deep layers of the app)


przemvs (approved):
fix type errors, overall look good :)

PatrykBuniX (approved):

PR #15074, opened 2023-04-25
runfix: Do not reset user availability when value is undefined

atomrc:
Do not let `undefined` reset the value for user's availability

### DIFF
--- a/src/script/user/UserMapper.ts
+++ b/src/script/user/UserMapper.ts
@@ -17,8 +17,6 @@
  *
  */
 
-import {Availability} from '@wireapp/protocol-messaging';
-
 import {joaatHash} from 'Util/Crypto';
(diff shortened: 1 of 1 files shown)


src/script/user/UserMapper.ts
```diff
@@ -119,7 +117,10 @@ export class UserMapper {
       userEntity.accent_id(accentId);
     }
 
-    userEntity.availability(availability);
+    if (availability !== undefined) {
+      // Availability should only change when it's a valid value (undefined should reset the availability)
```
PatrykBuniX: (undefined should reset the availability) -> (undefined **would** reset the availability)?

atomrc: Ahaha I forgot a `not` in the sentence 😆

PatrykBuniX (approved):

PR #15075, opened 2023-04-25
runfix: join mls conversations after processing messages

PatrykBuniX:
### Context
On app init there is a process of joining MLS groups with external commit user is a member of (from backend perspective). 

### Issue
The issue was it was joining the group before processing the message queue (slow/quick sync) that could include MLS welcome message, so user was joining the MLS group with external commit and then could receive welcome message what lead to some strange behaviour (the group was not established after decrypting the message that came after external join).

### Solution
The solution is to process message queue first and then join MLS groups user is a member of but haven't established yet. After user is added to a group, the welcome message is sent to all their clients, so the welcome message is the one that should initialise the conversation (no need to join with external commit sooner)

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -79,7 +79,7 @@ import {IntegrationRepository} from '../integration/IntegrationRepository';
 import {IntegrationService} from '../integration/IntegrationService';
 import {startNewVersionPolling} from '../lifecycle/newVersionHandler';
 import {MediaRepository} from '../media/MediaRepository';
-import {initMLSConversations, registerUninitializedConversations} from '../mls';
+import {initMLSConversations, registerUninitializedSelfAndTeamConversations} from '../mls';
 import {NotificationRepository} from '../notification/NotificationRepository';
 import {PreferenceNotificationRepository} from '../notification/PreferenceNotificationRepository';
 import {PermissionRepository} from '../permission/PermissionRepository';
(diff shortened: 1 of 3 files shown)


src/script/mls/MLSConversations.ts
```diff
@@ -54,9 +70,6 @@ export async function initMLSConversations(
     authorize: async () => true,
     userAuthorize: async () => true,
   });
-
-  const mlsConversations = conversations.filter(isMLSConversation);
-  await joinNewConversations(mlsConversations, core);
```
PatrykBuniX, line 59: We can split the responsibilities of joining mls groups and initialising mls callbacks for the core (this one should be called anytime before processing message queue)

src/script/main/app.ts
```diff
@@ -428,8 +429,13 @@ export class App {
       const notificationsCount = eventRepository.notificationsTotal;
 
       if (supportsMLS()) {
-        // Once all the messages have been processed and the message sending queue freed we can now add the potential `self` and `team` conversations
-        await registerUninitializedConversations(conversations, selfUser, clientEntity().id, this.core);
+        // Once all the messages have been processed and the message sending queue freed we can now:
+
+        //join all the mls groups we're member of and have not yet joined (eg. we were not send welcome message)
+        await initMLSConversations(conversations, this.core);
```
PatrykBuniX, line 435: This is the main change - initialising unestablished mls groups user is member of **after** processing message queue

przemvs (approved):

atomrc (approved):
seems solid to me 💪

PR #15076, opened 2023-04-25
runfix: Avoid having double user entries in DB

atomrc:
## The bug
When loading the app, you could be shown your previous availability status and get stuck to this

## The explanation
In the database there could be one entry for your user that is not fully qualified. This is the entry that was used to load your availability status from. 
**BUT** if you changed your status, then we would update the fully qualified entry. 

Next time you load the application, then only the non-qualified entry would be considered and you would end up with the previous status at each load.

## The solution

At load time, we deleted the non qualified users we have in DB and keep them in memory for later mapping (in case we don't find a qualified user in DB that correspond users in memory)

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -206,6 +206,11 @@ export class UserRepository {
       .concat(extraUsers);
     const users = uniq(allUserIds, false, (userId: QualifiedId) => userId.id);
 
+    // Remove all users that have non-qualified Ids in DB (there could be duplicated entries one qualified and one non-qualified)
+    // we want to get rid of the ambiguous entries
+    // the entries we get back will be used to feed the availabilities of those users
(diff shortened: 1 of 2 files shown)


src/script/user/UserRepository.ts
```diff
@@ -137,7 +137,6 @@ export class UserRepository {
     amplify.subscribe(WebAppEvents.CLIENT.UPDATE, this.updateClientsFromUser);
     amplify.subscribe(WebAppEvents.USER.SET_AVAILABILITY, this.setAvailability);
     amplify.subscribe(WebAppEvents.USER.EVENT_FROM_BACKEND, this.onUserEvent);
-    amplify.subscribe(WebAppEvents.USER.PERSIST, this.saveUserInDb);
```
atomrc, line 140: This event was never used anywhere

PatrykBuniX (approved):

PR #15081, opened 2023-04-26
runfix: apply correct design to unreachable user warnings

V-Gira:
### Issues

- A discard button was added to the failure to send warning without product approval (my bad)
- The warning for unreachable users would display even if there were none

image

image

### Solutions

- remove the discard button, it is still possible to delete the message in the context menu
- make sure the unreachable user warning is only sent if there are unreachable users

Screenshot from 2023-04-26 15-00-01
image

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/Warnings/PartialFailureToSend/PartialFailureToSend.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/Warnings/PartialFailureToSend/PartialFailureToSend.tsx
@@ -129,7 +129,7 @@ export const PartialFailureToSendWarning = ({failedToSend, knownUsers}: Props) =
 
               {/* maps through the unreachable users that will never receive the message:
               "3 participants from alpha.domain, 1 participant from beta.domain won't get your message" */}
-              {failed && (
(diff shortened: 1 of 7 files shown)


thisisamir98 (approved):

src/script/components/MessagesList/Message/ContentMessage/Warnings/PartialFailureToSend/PartialFailureToSend.tsx
```diff
@@ -129,7 +129,7 @@ export const PartialFailureToSendWarning = ({failedToSend, knownUsers}: Props) =
 
               {/* maps through the unreachable users that will never receive the message:
               "3 participants from alpha.domain, 1 participant from beta.domain won't get your message" */}
-              {failed && (
+              {unreachableUsers.length !== 0 && (
```
atomrc, line 132: How about adding a test case for this one ;) (check that this message doesn't display if there aren't any unreachable)?

PatrykBuniX (approved):

atomrc (approved):

atomrc:
serious business 💪 😎

PR #15117, opened 2023-05-05
feat: periodically check migration config [FS-1893]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1893" title="FS-1893" target="_blank">TaskFS-1893</a>  [Web] Periodically check if migration can be initialised
  </td></table>
  <br />
 

Periodically (after app is loaded and every 24h) check migration config to decide whether client should initialise / finalise the `MLS -> Proteus` migration process.

Note that MLS migration feature config response is mocked for now as it's not implemented on backend yet.


src/script/mls/mlsConversationState/mlsConversationState.ts
```diff
@@ -55,7 +57,10 @@ export const useMLSConversationState = create<StoreState>((set, get) => {
   return {
     established: initialState.established,
     filterEstablishedConversations: conversations =>
-      conversations.filter(conversation => !conversation.groupId || get().isEstablished(conversation.groupId)),
+      conversations.filter(conversation => {
+        const isMLS = isMLSConversation(conversation);
+        return !isMLS || get().isEstablished(conversation.groupId);
+      }),
```
PatrykBuniX, line 63: After mixed conversations (that also contain `groupId` field) were introduced, we want to hide only `"mls"` conversations that are not established. We're totally fine with showing `"mixed"` conversations that don't have MLS group established - when conversation is `"mixed"`, we still send messages via Proteus.

src/script/conversation/groupConversationsByProtocol.ts
```diff
@@ -0,0 +1,66 @@
...
+      groupedConversations[ConversationProtocol.MIXED].push(conversation);
+      continue;
+    }
+    if (isMLSConversation(conversation)) {
+      groupedConversations[ConversationProtocol.MLS].push(conversation);
+      continue;
+    }
+  }
+
+  return groupedConversations;
```
przemvs: Maybe use .reduce() method ? 🤔

src/script/mls/isMLSSupportedByEnvironment.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!isMLSEnabledOnBackend) {
+    return false;
+  }
+
+  let isBackendRemovalKeyPresent = false;
+  try {
+    const backendRemovalKey = (await apiClient.api.client.getPublicKeys()).removal;
+    if (backendRemovalKey) {
+      isBackendRemovalKeyPresent = true;
+    }
```
przemvs: Can't be simplified ?
```suggestion
      isBackendRemovalKeyPresent = !!backendRemovalKey;
```

przemvs (approved):

PR #14975, opened 2023-04-05
feat: Refresh users when conversation details is opened [FS-1690]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1690" title="FS-1690" target="_blank">Sub-taskFS-1690</a>  [web] Retry fetching missing user metadata when conversation details are opened
  </td></table>
  <br />
 

This will allow refreshing all the missing users of a conversation when the conversation details are opened 

https://user-images.githubusercontent.com/1090716/230091742-1b467266-8b78-4730-904f-3ae147bfdd9d.mov

PR #15158, opened 2023-05-11
feat: Retry fetching missing user/conversation metadata every 3h (FS-1691)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1691" title="FS-1691" target="_blank">Sub-taskFS-1691</a>  [web] Retry fetching missing user/conversation metadata every 3h
  </td></table>
  <br />
 

## Scenario:

- Backend A & B are federated.
- User X from A & User Y from B are in a conversation.
- Connection between A & B goes offline.
- User X does a fresh login with no previous data on its local database.

## Issue:

At this point user X is unable to fetch the meta data of user Y because the data lives on the B backend and connection is offline.

## Solution 
@atomrc Has already implemented a try to refetch that missing meta data when opening the user details menu here:
https://github.com/wireapp/wire-webapp/pull/14975

This PR will add another solution to this issue to try to refetch all missing data every 3 hours.

### DIFF
--- a/src/script/conversation/ConversationRepository.test.ts
+++ b/src/script/conversation/ConversationRepository.test.ts
@@ -1422,11 +1422,14 @@ describe('ConversationRepository', () => {
     it('should not call loadMissingConversations & refreshAllConversationsUnavailableParticipants for non federated envs', async () => {
       const conversationRepo = await testFactory.exposeConversationActors();
 
-      spyOn(testFactory.conversation_repository!, 'loadMissingConversations').and.callThrough();
-      spyOn(testFactory.conversation_repository!, 'refreshAllConversationsUnavailableParticipants').and.callThrough();
+      spyOn(conversationRepo, 'loadMissingConversations').and.callThrough();
+      spyOn(
+        conversationRepo,
+        'refreshAllConversationsUnavailableParticipants' as keyof ConversationRepository,
+      ).and.callThrough();
 
       expect(conversationRepo.loadMissingConversations).not.toHaveBeenCalled();
-      expect(conversationRepo.refreshAllConversationsUnavailableParticipants).not.toHaveBeenCalled();
+      expect(conversationRepo['refreshAllConversationsUnavailableParticipants']).not.toHaveBeenCalled();
     });
 
     it('should call loadMissingConversations & refreshAllConversationsUnavailableParticipants every 3 hours for federated envs', async () => {
@@ -1435,15 +1438,18 @@ describe('ConversationRepository', () => {
       });
       const conversationRepo = await testFactory.exposeConversationActors();
 
-      spyOn(testFactory.conversation_repository!, 'loadMissingConversations').and.callThrough();
-      spyOn(testFactory.conversation_repository!, 'refreshAllConversationsUnavailableParticipants').and.callThrough();
+      spyOn(conversationRepo, 'loadMissingConversations').and.callThrough();
+      spyOn(
+        conversationRepo,
+        'refreshAllConversationsUnavailableParticipants' as keyof ConversationRepository,
+      ).and.callThrough();
 
       jest.advanceTimersByTime(3600000 * 4);
 
       await Promise.resolve();
 
       expect(conversationRepo.loadMissingConversations).toHaveBeenCalled();
-      expect(conversationRepo.refreshAllConversationsUnavailableParticipants).toHaveBeenCalled();
+      expect(conversationRepo['refreshAllConversationsUnavailableParticipants']).toHaveBeenCalled();
     });
   });
 });
--- a/src/script/conversation/ConversationRepository.test.ts
+++ b/src/script/conversation/ConversationRepository.test.ts
@@ -100,8 +100,6 @@ describe('ConversationRepository', () => {
   };
 
   beforeAll(async () => {
-    jest.useFakeTimers();
-
     server = sinon.fakeServer.create();
     server.autoRespond = true;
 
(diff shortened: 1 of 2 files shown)


V-Gira (commented):
Looking good! :muscle:

src/script/conversation/ConversationRepository.ts
```diff
@@ -467,6 +468,32 @@ export class ConversationRepository {
     await this.userRepository.refreshUsers(unavailableUsers.map(user => user.qualifiedId));
   }
 
+  public async refreshAllConversationsUnavailableParticipants(): Promise<void> {
+    const allUnavailableUsers: User[] = [];
+    this.conversationState.conversations().forEach(conversation => {
+      allUnavailableUsers.push(...conversation.allUserEntities().filter(user => !user.isAvailable()));
+    });
```
przemvs: Why no .map()? :D

src/script/conversation/ConversationRepository.ts
```diff
@@ -291,6 +291,7 @@ export class ConversationRepository {
 
     this.conversationRoleRepository = new ConversationRoleRepository(this.teamRepository, this.conversationService);
     this.leaveCall = noop;
+    this.scheduleMissingUsersAndConversationsMetadataRefresh();
```
tlebon: if this is supposed to be federation only, why isn't it conditional at all?

src/script/conversation/ConversationRepository.test.ts
```diff
@@ -100,6 +100,8 @@ describe('ConversationRepository', () => {
   };
 
   beforeAll(async () => {
+    jest.useFakeTimers();
```
V-Gira: Is this somehow breaking some other test cases? The ci is breaking

tlebon:
please fill out the PR info

src/script/conversation/ConversationRepository.ts
thisisamir98: there is no missing data so we'll iterate over an empty array and no request will be fired, but still we could add a isFederated check to avoid that as well but performance wise that's micro optimization

src/script/conversation/ConversationRepository.test.ts
thisisamir98: maybe, i'm checking...

src/script/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15158/commits/9b2f6134b862c445d52393e491324510aa3df377
https://github.com/wireapp/wire-webapp/pull/15158/commits/2eaea15b8700d9042f5c5d068c6fdc399f945bfa

src/script/conversation/ConversationRepository.ts
tlebon: is there no way to tell we are in a federated environment? i dont see why we should run a bunch of async calls every 3 hours if we dont need to. (regardless of performance impact)

PatrykBuniX: maybe `this.core.backendFeatures.isFederated`? @thisisamir98

thisisamir98: https://github.com/wireapp/wire-webapp/pull/15158/commits/943806cda942c6839ec9f6ce0f8743c3f88e1a13

src/script/conversation/ConversationRepository.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15158/commits/5459a623a56bd63ccda7c72b8d3e168a32575835

src/script/conversation/ConversationRepository.ts
```diff
@@ -467,6 +468,35 @@ export class ConversationRepository {
...
+  }
+
+  /**
+   * Refresh missing conversations and unavailable users metadata every 3 hours
+   * @Note Federation only
+   */
+  private readonly scheduleMissingUsersAndConversationsMetadataRefresh = () => {
+    if (!this.core.backendFeatures.isFederated) {
+      return;
+    }
```
PatrykBuniX: if we do nothing in this case, maybe let's just not call this function at all, so wrap it with the condition:
```ts
if(this.core.backendFeatures.isFederated){
  scheduleMissingUsersAndConversationsMetadataRefresh()
}
```
wdyt?

src/script/conversation/ConversationRepository.ts
```diff
@@ -467,6 +468,35 @@ export class ConversationRepository {
     await this.userRepository.refreshUsers(unavailableUsers.map(user => user.qualifiedId));
   }
 
+  public async refreshAllConversationsUnavailableParticipants(): Promise<void> {
```
PatrykBuniX: do we need this one to be public since it's being used only in this class?

thisisamir98: the other one (refreshUnavailableParticipants) implemented previously by @atomrc was public so I made this public as well but we could make it private.

src/script/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15158/commits/6c6514c894005c97d7b11485a64f5759183e668c

tlebon (approved):

PatrykBuniX (approved):

PR #15162, opened 2023-05-12
refactor: exclude logic to merge single conversation

PatrykBuniX:
ConversationMapper refactor:
- rename `mergeConversation` to `mergeConversations` as it really merges multiple conversations
- create a method for merging single conversation and use it when merging multiple entities.


V-Gira (approved):
Doesn't seem to break anything :+1:

PR #15183, opened 2023-05-15
runfix: update url after creating conversation

PatrykBuniX:
### Issue:
After creating a conversation, user was not navigated to the url of the new conversation. After creating a group, it did show up as `"active"` conversation, but after refreshing the page it would load a conversation that was `"active"` before creating a new one.

### Solution:
- Use our `createNavigate` util that will redirect the user to proper url.
- I've also improved `createGroupConversation` method's implementation to always return the conversation and fail otherwise (previously it would return undefined what doesn't make sense).

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -39,7 +39,7 @@ import {UserSearchableList} from 'Components/UserSearchableList';
 import {generateConversationUrl} from 'src/script/router/routeGenerator';
 import {createNavigate, createNavigateKeyboard} from 'src/script/router/routerBindings';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
-import {handleEnterDown, offEscKey, onEscKey} from 'Util/KeyboardUtil';
+import {handleEnterDown, isKeyboardEvent, offEscKey, onEscKey} from 'Util/KeyboardUtil';
 import {t} from 'Util/LocalizerUtil';
 import {getLogger} from 'Util/Logger';
 import {sortUsersByPriority} from 'Util/StringUtil';
@@ -221,7 +221,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
         );
         setIsShown(false);
 
-        if ('key' in event) {
+        if (isKeyboardEvent(event)) {
           createNavigateKeyboard(generateConversationUrl(conversation.qualifiedId), true)(event);
         } else {
           createNavigate(generateConversationUrl(conversation.qualifiedId))(event);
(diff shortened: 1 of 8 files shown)


src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -216,7 +220,12 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
           },
         );
         setIsShown(false);
-        amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity, {});
+
+        if ('key' in event) {
+          createNavigateKeyboard(generateConversationUrl(conversation.qualifiedId), true)(event);
+        } else {
+          createNavigate(generateConversationUrl(conversation.qualifiedId))(event);
+        }
```
PatrykBuniX: Conversation show logic stays the same.

`ContentViewModel` was listening for `WebAppEvents.CONVERSATION.SHOW` event and was calling `showConversation` with passed arguments (see https://github.com/wireapp/wire-webapp/blob/dev/src/script/view_model/ContentViewModel.ts#L118). This is exactly what happens after changing the route, see: https://github.com/wireapp/wire-webapp/blob/dev/src/script/page/AppMain.tsx#L167

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -216,7 +220,12 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
           },
         );
         setIsShown(false);
-        amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity, {});
+
+        if ('key' in event) {
```
thisisamir98: I think you could use 
isKey from Util/KeyboardUtil 
🤔

PatrykBuniX: `isKey`'s job is to check if event was fired by specific keyboard key, we don't care about it here. `'key' in event` is just to differentiate between `React.MouseEvent` and `React.KeyboardEvent` so we can use `createNavigate` or `createNavigateKeyboard` accordingly, does it make sense?

thisisamir98: how about `isClickEvent` in `Mouse.ts`

thisisamir98: or maybe let's add a `isKeyboardEvent` to `KeyboardUtil`

PatrykBuniX: `isKeyboardEvent` sounds good to me, will create it and use it in all the places we do `('key' in event)` 🫡

thisisamir98 (approved):

PR #15187, opened 2023-05-15
feat: Add a system message for failed to add users (FS-1548)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1548" title="FS-1548" target="_blank">TaskFS-1548</a>  [Web] Implement UX on clients for conversation creation
  </td></table>
  <br />
 

When failed to add 1 person:
image

When failed to add more than 1 person:
image
image


src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  afterEach(() => {
+    userRepository['userState'].users.removeAll();
+  });
+
+  it('shows that 1 user could not be added', async () => {
+    const qualifiedId1 = {
+      domain: 'foma.wire.link',
+      id: 'oi123no4-4f9f-4b9e-8c51-124b1uy24i12',
+    };
```
przemvs: Maybe use this function generateQualifiedId :)

src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  it('shows that multiple users could not be added', async () => {
+    const qualifiedId1 = {
+      domain: 'foma.wire.link',
+      id: 'oi123no4-4f9f-4b9e-8c51-124b1uy24i12',
+    };
+    const qualifiedId2 = {
+      domain: 'foma.wire.link',
+      id: 'n1iuo42h-0c6e-4b6c-b973-12i4b12ib412',
+    };
```
przemvs: We have generator for this, but not exported: generateQualifiedId .

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const foundUser = allUsers.find(user => matchQualifiedIds(current, user.qualifiedId));
+      if (foundUser) {
+        return [...previous, foundUser];
+      }
+      return previous;
+    }, []);
+    const groupedUsers = groupBy(users, user => user.domain);
+    const total = users.length;
+    return [users, total, groupedUsers];
+  }, [allUsers, message.qualifiedIds]);
```
przemvs: We need this useMemo ? :)

thisisamir98: in my opinion it is heavy calculation, for each user (qualified id) we have to search all the users we have (are connected to at least)

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const learnMore = (
+    <>
+      {' '}
+      <Link
+        targetBlank
+        variant={LinkVariant.PRIMARY}
+        href={config.URL.SUPPORT.USER_ADD_FAILURE}
+        data-uie-name="go-contact-support"
+        style={{
+          color: COLOR_V2.BLUE_LIGHT_500,
```
V-Gira: Did you check if the color was blue or accent-color for the link? it's not obvious on the mock.

If it's blue, we can use `var(--blue-500)` , it should have the correct value in light and dark mode

src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -234,6 +235,9 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
   if (message.isCallTimeout()) {
     return <CallTimeoutMessage message={message} />;
   }
+  if (message.isFailedToAddUsersMessage()) {
+    return <FailedToAddUsersMessage message={message} />;
```
V-Gira, line 239: I think there's an issue with putting that warning in the message wrapper.

The mocks put the warning between the added users and the following system messages (read receipts etc)

putting it here would put here after every system message and add a timestamp.

`MemberMessage.tsx` seems to be where those system message live, wdyt?

thisisamir98: I just confirmed with design that order does not matter as going forward on mls we get that response some time later

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15187/commits/939ddea6778a66f5256c5134477a96a9349eb21f

przemvs (approved):

PatrykBuniX (approved):

PatrykBuniX (approved):

PR #15189, opened 2023-05-16
feat: Filter unavailable users in conversation creation (FS-1965)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1965" title="FS-1965" target="_blank">TaskFS-1965</a>  [Web] Filter list of unavailable users when creating a conversation
  </td></table>
  <br />
 

This should be filtered, we should not show unavailable users when creating a conversation:
image

Result:
image

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -71,8 +71,6 @@ enum GroupCreationModalState {
 
 const logger = getLogger('GroupCreationModal');
 
-const filterAvailableOnlyUsers = (user: User) => user.isAvailable();
-
 const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
   userState = container.resolve(UserState),
   teamState = container.resolve(TeamState),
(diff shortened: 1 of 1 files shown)


src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -149,14 +151,14 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
+        return userState.connectedUsers().filter(filterAvailableOnlyUsers);
       }
 
       if (isGuestEnabled) {
-        return teamState.teamUsers();
+        return teamState.teamUsers().filter(filterAvailableOnlyUsers);
       }
 
-      return teamState.teamMembers().sort(sortUsersByPriority);
+      return teamState.teamMembers().filter(filterAvailableOnlyUsers).sort(sortUsersByPriority);
```
przemvs: I'll remove this filters here, and You can filter contacts =>

`const filteredContacts = contacts.filter(user => user.isAvailable());`

thisisamir98: https://github.com/wireapp/wire-webapp/pull/15189/commits/f08c836743780bf38a19b14fc8e45e6a0f1c199d

PatrykBuniX (approved):

PR #15190, opened 2023-05-16
runfix: close create group modal on failure

PatrykBuniX:
Add breaks to switch case so it don't end up throwing error even if it was handled.

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -230,9 +230,11 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
           createNavigate(generateConversationUrl(conversation.qualifiedId))(event);
         }
       } catch (error) {
+        amplify.publish(WebAppEvents.CONVERSATION.SHOW, undefined, {});
         setIsCreatingConversation(false);
         logger.error(error);
       }
+
       setIsShown(false);
     }
   };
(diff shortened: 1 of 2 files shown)


V-Gira (approved):

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -223,17 +223,19 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
...
         );
-        setIsShown(false);
 
         if (isKeyboardEvent(event)) {
           createNavigateKeyboard(generateConversationUrl(conversation.qualifiedId), true)(event);
         } else {
           createNavigate(generateConversationUrl(conversation.qualifiedId))(event);
         }
       } catch (error) {
+        amplify.publish(WebAppEvents.CONVERSATION.SHOW, undefined, {});
```
PatrykBuniX, line 233: I've removed this here https://github.com/wireapp/wire-webapp/pull/15183/files#diff-374d17b2a9898c44a7be4d7d1e40a329f71a2d6b53c865e052296e2e4988646aL219, but it's needed in case of error (it will keep old url and navigate to connections request as before)

src/script/conversation/ConversationRepository.ts
```diff
@@ -455,20 +455,20 @@ export class ConversationRepository {
...
+            break;
+          case BackendErrorLabel.NOT_CONNECTED:
+            await this.handleUsersNotConnected(userEntities.map(user => user.qualifiedId));
+            break;
+          case BackendErrorLabel.LEGAL_HOLD_MISSING_CONSENT:
+            this.showLegalHoldConsentError();
+            break;
+        }
       }
+      throw error;
```
PatrykBuniX, line 471: we can always throw, error is handled by the modal

V-Gira (approved):

tlebon (approved):
seems fine

PR #15193, opened 2023-05-16
fix: showClose functionality of primaryModal component

aweiss-dev:


src/script/components/Modals/PrimaryModal/PrimaryModal.test.tsx
```diff
@@ -80,4 +83,51 @@ describe('PrimaryModal', () => {
...
+          action: () => {},
+          text: 'secondary-text',
+        },
+        text: {
+          message: 'test-message',
+          title: 'test-title',
+        },
+      });
+    });
+    const closeButton = container.querySelector('button[data-uie-name="do-close"]');
```
PatrykBuniX: you can use `getByTestId('do-close')` - we've `data-uie-name` configured as our `testid`

PatrykBuniX (approved):
looks good

PR #15197, opened 2023-05-17
fix: Fix group name change, focus edit icon

przemvs:
Fix for changing group name in Right Sidebar.

Previous: 
When user change cursor position and write smth, the cursor going to the end of the text.
We can't focus edit group name by keyboard.

Now: 
Changing group name work proper.
We can focus edit group name by keyboard.

### DIFF
--- a/src/script/page/RightSidebar/ConversationDetails/components/ConversationDetailsHeader/ConversationDetailsHeader.tsx
+++ b/src/script/page/RightSidebar/ConversationDetails/components/ConversationDetailsHeader/ConversationDetailsHeader.tsx
@@ -117,7 +117,6 @@ const ConversationDetailsHeader: FC<ConversationDetailsHeaderProps> = ({
           {!isEditingName ? (
             <div
               className="conversation-details__name"
-              title={t('tooltipConversationDetailsRename')}
               data-uie-name="status-name"
               {...(canRenameGroup && {
                 onClick: clickToEditGroupName,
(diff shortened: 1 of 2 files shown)


src/script/page/RightSidebar/ConversationDetails/components/ConversationDetailsHeader/ConversationDetailsHeader.tsx
```diff
@@ -120,7 +125,14 @@ const ConversationDetailsHeader: FC<ConversationDetailsHeaderProps> = ({
             >
               {displayName && <span className="conversation-details__name">{displayName}</span>}
 
-              {canRenameGroup && <Icon.Edit className="conversation-details__name__edit-icon" />}
+              {canRenameGroup && (
```
arjita-mitra, line 127: Remove title from the div otherwise screenreader is repeating the same thing twice

```
 <div
              className="conversation-details__name"
              title={t('tooltipConversationDetailsRename')}> 
```
              
Remove the title from the div since you have added it to the button which is correct.

arjita-mitra (changes requested):

src/script/page/RightSidebar/ConversationDetails/components/ConversationDetailsHeader/ConversationDetailsHeader.tsx
```diff
@@ -120,7 +125,14 @@ const ConversationDetailsHeader: FC<ConversationDetailsHeaderProps> = ({
...
+                <button
+                  className="conversation-details__name__edit-icon"
+                  aria-label={t('tooltipConversationDetailsRename')}
+                >
+                  <Icon.Edit aria-hidden="true" />
```
arjita-mitra: do we need to pass aria-hidden? i think by default aria-hidden is true for icons?

arjita-mitra (changes requested):

arjita-mitra (approved):

PR #15198, opened 2023-05-17
feat: start migration of proteus conversations [FS-1888]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1888" title="FS-1888" target="_blank">TaskFS-1888</a>  [Web] If migration time has arrived, start migrating conversations 
  </td></table>
  <br />
 

Start migration process of proteus conversations:
- update protocol to `"mixed"`
- update conversation locally
- establish mls group
- add other clients to the group (including other self-clients, but skipping current client)


src/i18n/en-US.json
```diff
@@ -436,6 +436,8 @@
   "conversationYouAccusative": "you",
   "conversationYouDative": "you",
   "conversationYouNominative": "you",
+  "conversationProtocolUpdatedToMixed": "Conversation has started migration to MLS protocol.",
```
PatrykBuniX, line 439: This will change, we don't have exact copy and ui for the "protocol has changed" system messages yet

src/script/conversation/MessageRepository.ts
```diff
@@ -777,9 +776,9 @@ export class MessageRepository {
     // Configure ephemeral messages
     conversationService.messageTimer.setConversationLevelTimer(conversation.id, conversation.messageTimer());
 
-    const sendOptions: Parameters<typeof conversationService.send>[0] = groupId
+    const sendOptions: Parameters<typeof conversationService.send>[0] = isMLSConversation(conversation)
       ? {
-          groupId,
+          groupId: conversation.groupId,
```
PatrykBuniX, line 781: We want to use MLS only when it's mls conversation, so:
- protocol is `"mls"`
- it has `"groupId"` field
Previously only MLS conversations had groupId, it has changed when `"mixed"` "protocol" was introduced. Mixed conversation also contains `"groupId"`, but we still want to send messages with proteus

src/script/conversation/ConversationRepository.ts
```diff
@@ -1697,6 +1698,53 @@ export class ConversationRepository {
...
+   *
+   * @param conversationId id of the conversation
+   * @param protocol new conversation protocol
+   * @returns Resolves with updated conversation entity
+   */
+  public async updateConversationProtocol(
+    conversation: Conversation,
+    protocol: ConversationProtocol.MIXED | ConversationProtocol.MLS,
+  ): Promise<Conversation> {
+    const response = await this.conversationService.updateConversationProtocol(conversation.qualifiedId, protocol);
```
przemvs: What do you think about rename response -> updateEvent, response is too generic

src/script/conversation/ConversationRepository.ts
```diff
@@ -1697,6 +1698,53 @@ export class ConversationRepository {
...
+   *
+   * @param conversationId id of the conversation
+   * @returns Resolves with updated conversation entity
+   */
+  private async refreshConversationProtocolProperties(conversation: Conversation) {
+    //refetch the conversation to get all new fields (groupId, ciphersuite, epoch and new protocol)
+    const remoteConversationData = await this.conversationService.getConversationById(conversation.qualifiedId);
+
+    //update fields that came after protocol update
+    const {cipher_suite: cipherSuite, epoch, group_id: newGroupId, protocol: newProtocol} = remoteConversationData;
```
przemvs, line 1739: There will be always conversationData ?

src/script/conversation/MessageRepository.ts
przemvs: Maybe add this comment to code ? 🤔

src/script/mls/MLSMigration/initialiseMigration/addMixedConversationMembersToMLSGroup/addMixedConversationMembersToMLSGroup.test.ts
```diff
@@ -0,0 +1,74 @@
...
+const selfUserId = {id: 'self-user-id', domain: 'local.wire.com'};
+
+describe('addMixedConversationMembersToMLSGroup', () => {
+  Object.defineProperty(mockCore, 'clientId', {value: selfUserClientId});
+
+  it('should add all mixed conversation members to MLS group', async () => {
+    const user1Id = {id: createUuid(), domain: 'local.wire.com'};
+    const user2Id = {id: createUuid(), domain: 'local.wire.com'};
+    const user3Id = {id: createUuid(), domain: 'local.wire.com'};
+    const conversationMembers = [user1Id, user2Id, user3Id];
```
przemvs: Maybe to smth like that:

`const conversationMembers = Array(userCount).fill(0).map(() => ({ id: createUuid(), domain: 'local.wire.com' }))`

WDYT ? :)

src/script/mls/MLSMigration/initialiseMigration/addMixedConversationMembersToMLSGroup/addMixedConversationMembersToMLSGroup.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    const mixedConversation = createMixedConversation(conversationMembers);
+
+    await addMixedConversationMembersToMLSGroup(mixedConversation, {core: mockCore, selfUserId});
+
+    expect(mockCore.service?.conversation.addUsersToMLSConversation).toHaveBeenCalledWith({
+      groupId: mixedConversation.groupId,
+      conversationId: mixedConversation.qualifiedId,
+      qualifiedUsers: [user1Id, user2Id, user3Id, {...selfUserId, skipOwnClientId: selfUserClientId}],
```
przemvs: Then there you can do smth like that:

`qualifiedUsers: [...conversationMembers, {...selfUserId, skipOwnClientId: selfUserClientId}],`

src/script/mls/MLSMigration/initialiseMigration/initialiseMigration.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      client: selfUserClientId,
+      user: selfUserId,
+    });
+    expect(mockCore.service!.mls!.wipeConversation).toHaveBeenCalledWith(mockGroupId);
+    expect(mockCore.service!.conversation!.addUsersToMLSConversation).not.toHaveBeenCalled();
+  });
+
+  it('Should initialise migration for proteus conversations and add other users to MLS group', async () => {
+    const user1Id = {id: 'user1', domain: 'local.wire.link'};
+    const user2Id = {id: 'user2', domain: 'local.wire.link'};
```
przemvs: There you can also use generate function which I commented earlier :)

src/script/conversation/MessageRepository.ts
PatrykBuniX: I don't think it's needed, it should be like this from the beginning - we should send with MLS if protocol is MLS, not if there's a group id

src/script/conversation/ConversationRepository.ts
PatrykBuniX: Yes, if it cannot find the conversation it should throw error (which should not happen, since conversation should exist as we've loaded it already on app start and we've just changed the protocol)

przemvs (approved):

arjita-mitra (approved):

PR #15298, opened 2023-06-05
runfix: periodic renew keying material

PatrykBuniX:
Bumps the core version
- new version improved periodic key packages update mechanism

### DIFF
--- a/src/script/mls/MLSConversations.ts
+++ b/src/script/mls/MLSConversations.ts
@@ -46,9 +46,15 @@ type MLSConversationRepository = Pick<
  * @param core - the instance of the core
  */
 export async function initMLSConversations(conversations: Conversation[], core: Account): Promise<void> {
+  const mlsService = core.service?.mls;
+  if (!mlsService) {
+    throw new Error('MLS service not available');
+  }
+
   const mlsConversations = conversations.filter(isMLSConversation);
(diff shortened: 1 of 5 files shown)


src/script/mls/MLSConversations.ts
```diff
@@ -45,9 +45,10 @@ type MLSConversationRepository = Pick<
  * @param conversations - all the conversations that the user is part of
  * @param core - the instance of the core
  */
-export function initMLSConversations(conversations: Conversation[], core: Account): Promise<void> {
+export async function initMLSConversations(conversations: Conversation[], core: Account): Promise<void> {
   const mlsConversations = conversations.filter(isMLSConversation);
-  return joinNewConversations(mlsConversations, core);
+  await joinNewConversations(mlsConversations, core);
+  return core.service?.mls?.schedulePeriodicKeyMaterialRenewals(mlsConversations.map(({groupId}) => groupId));
```
PatrykBuniX: This is the main change - consumer (webapp) is responsible for initiating the method by providing the list of group ids

aweiss-dev: I would love to see a type guard here or at least a log for the case that core.service?.mls?.schedulePeriodicKeyMaterialRenewals is not defined.

aweiss-dev (approved):

PR #15318, opened 2023-06-07
feat: maintain mls group list during migration [WPB-1115]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1115" title="WPB-1115" target="_blank">TaskWPB-1115</a>  [Web] Maintaining MLS group's clients list during migration
  </td></table>
  <br />
 

Maintain MLS group in `"mixed"` conversation. It convers:
- adding users (and being added to a group)
- removing a user (being removed)
- leaving a group
- receiving commits
- periodic key material update

Note that leaving a group and removing user from a group might not work before being deployed to draft-20.

- [x] add unit tests

### DIFF
--- a/src/script/conversation/ConversationRepository.test.ts
+++ b/src/script/conversation/ConversationRepository.test.ts
@@ -817,45 +817,41 @@ describe('ConversationRepository', () => {
         });
       });
 
-      it('should add other self clients to mls group if user was event creator', () => {
-        const mockDomain = 'example.com';
-        const mockSelfClientId = 'self-client-id';
-        const selfUser = generateUser({id: createUuid(), domain: mockDomain});
+      it.each([ConversationProtocol.MIXED, ConversationProtocol.MLS])(
+        'should add other self clients to mls/mixed conversation MLS group if user was event creator',
+        protocol => {
+          const mockDomain = 'example.com';
+          const mockSelfClientId = 'self-client-id';
+          const selfUser = generateUser({id: createUuid(), domain: mockDomain});
 
-        const conversationEntity = _generateConversation(
-          CONVERSATION_TYPE.REGULAR,
-          undefined,
-          ConversationProtocol.MLS,
-          mockDomain,
-        );
-        testFactory.conversation_repository['saveConversation'](conversationEntity);
+          const conversationEntity = _generateConversation(CONVERSATION_TYPE.REGULAR, undefined, protocol, mockDomain);
+          testFactory.conversation_repository['saveConversation'](conversationEntity);
 
-        const memberJoinEvent = {
-          conversation: conversationEntity.id,
-          data: {
-            user_ids: [selfUser.id],
-          },
-          from: selfUser.id,
-          time: '2015-04-27T11:42:31.475Z',
-          type: CONVERSATION_EVENT.MEMBER_JOIN,
-        } as ConversationMemberJoinEvent;
+          const memberJoinEvent = {
+            conversation: conversationEntity.id,
+            data: {
+              user_ids: [selfUser.id],
+            },
+            from: selfUser.id,
+            time: '2015-04-27T11:42:31.475Z',
+            type: CONVERSATION_EVENT.MEMBER_JOIN,
+          } as ConversationMemberJoinEvent;
 
-        spyOn(testFactory.conversation_repository['userState'], 'self').and.returnValue(selfUser);
+          spyOn(testFactory.conversation_repository['userState'], 'self').and.returnValue(selfUser);
 
-        Object.defineProperty(container.resolve(Core), 'clientId', {
-          get: jest.fn(() => mockSelfClientId),
-        });
+          container.resolve(Core).clientId = mockSelfClientId;
 
-        return testFactory.conversation_repository['handleConversationEvent'](memberJoinEvent).then(() => {
-          expect(testFactory.conversation_repository['onMemberJoin']).toHaveBeenCalled();
-          expect(testFactory.conversation_repository.updateParticipatingUserEntities).toHaveBeenCalled();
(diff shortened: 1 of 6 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -951,11 +959,9 @@ export class ConversationRepository {
     }
     this.deleteConversationFromRepository(conversationId);
     await this.conversationService.deleteConversationFromDb(conversationId);
-    if (conversationEntity.protocol === ConversationProtocol.MLS) {
+    if (isMLSCapableConversation(conversationEntity)) {
```
PatrykBuniX, line 961: If we're deleting `mixed` or `mls` conversation we need to wipe its MLS group.

src/script/conversation/ConversationRepository.ts
```diff
@@ -1455,25 +1461,25 @@ export class ConversationRepository {
...
+      if (isMLSCapableConversation(conversation)) {
+        const {events} = await this.core.service!.conversation.addUsersToMLSConversation({
+          conversationId: conversation.qualifiedId,
+          groupId: conversation.groupId,
           qualifiedUsers,
         });
-        if (conversationMemberJoinEvent) {
-          this.eventRepository.injectEvent(conversationMemberJoinEvent, EventRepository.SOURCE.BACKEND_RESPONSE);
+        if (!!events.length && isMLSConversation(conversation)) {
+          events.forEach(event => this.eventRepository.injectEvent(event));
```
PatrykBuniX, line 1481: When adding a new user we have to do it in 2 ways: "proteus" and "mls" way, because we might be adding user who potentially does not support MLS yet.

So if it's proteus or mixed conversation we add the user with proteus as usual and then also if it's mixed or mls conversation we add it with mls (adding user to MLS group and sending a commit bundle).

src/script/conversation/ConversationRepository.ts
```diff
@@ -2714,11 +2719,9 @@ export class ConversationRepository {
         eventJson.from = this.userState.self().id;
       }
 
-      if (conversationEntity.protocol === ConversationProtocol.MLS) {
+      if (isMLSCapableConversation(conversationEntity)) {
```
PatrykBuniX, line 2718: Whenever we leave (or get removed) from mixed / mls conversation we have to wipe its MLS group.

src/script/conversation/ConversationRepository.test.ts
```diff
@@ -85,8 +86,8 @@ const _generateConversation = (
   connectionEntity.status(connection_status);
   conversation.connection(connectionEntity);
 
-  if (conversationProtocol === ConversationProtocol.MLS) {
-    conversation.groupId = 'groupId';
+  if ([ConversationProtocol.MLS, ConversationProtocol.MIXED].includes(conversationProtocol)) {
```
przemvs: Create variable from [ConversationProtocol.MLS, ConversationProtocol.MIXED], and reuse it in next test cases :)

src/script/conversation/ConversationRepository.test.ts
```diff
@@ -1626,4 +1623,157 @@ describe('ConversationRepository', () => {
...
+
+  describe('addUsers', () => {
+    it('should add users to proteus conversation', async () => {
+      const conversation = _generateConversation();
+      const conversationRepository = await testFactory.exposeConversationActors();
+
+      const user1 = generateUser();
+      const user2 = generateUser();
+
+      const usersToAdd = [user1, user2];
```
przemvs: Hmm.. create fn which return for you users to add ;) So You can reuse it in test cases, instead of doing the sae variables, etc ;)

src/script/conversation/ConversationRepository.ts
```diff
@@ -2650,13 +2656,12 @@ export class ConversationRepository {
    * @param conversation Conversation member joined to
    * @param isSelfJoin whether user has joined by itself, if so we need to add other self clients to mls group
    */
-  private async handleMLSConversationMemberJoin(conversation: Conversation, isSelfJoin: boolean) {
+  private async handleMLSConversationMemberJoin(
+    conversation: MLSConversation | MixedConversation,
```
przemvs: Create new type which contains both (MLS and Mixed types).

eg.
`type <conversation_type_name> = MLSConversation | MixedConversation`
WDYT?

src/script/mls/mlsConversationState/mlsConversationState.ts
```diff
@@ -58,7 +58,7 @@ export const useMLSConversationState = create<StoreState>((set, get) => {
     established: initialState.established,
     filterEstablishedConversations: conversations =>
       conversations.filter(conversation => {
-        const isMLS = isMLSConversation(conversation);
+        const isMLS = isMLSCapableConversation(conversation);
```
przemvs: Just move it to the return statement, we don't need pass it to the variable. You using only once.

`return !isMLSCapableConversation(conversation) || get().isEstablished(conversation.groupId);`

src/script/conversation/ConversationRepository.test.ts
PatrykBuniX: I don't think creating a separate function makes sense, I've deleted `user1`, `user2` if they were not referenced and simply passed [`generateUser()`, `generateUser()`] array instead

src/script/conversation/ConversationRepository.ts
PatrykBuniX: I prefer to use both types explicitly to know what type of conversations we're dealing with

przemvs: Yee, but if you create this type and export You'll have types if you hover on it :) I see You using it in multiple places

PatrykBuniX: ok, added `MLSCapableConversation` type ;)

przemvs (approved):

PR #15320, opened 2023-06-08
runfix: allow message duplicate when retrying to send a message after a federation error [WBP-2010]

V-Gira:
- A new event status for federation errors was not added to valid cases for retry attempts

### DIFF
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -572,7 +572,7 @@ export class EventRepository {
     }
 
     const containsLinkPreview = newEventData.previews && !!newEventData.previews.length;
-    const isRetryAttempt = originalEvent.status === StatusType.FAILED || StatusType.FEDERATION_ERROR;
+    const isRetryAttempt = originalEvent.status === (StatusType.FAILED || StatusType.FEDERATION_ERROR);
 
     if (!containsLinkPreview && !isRetryAttempt) {
       const errorMessage =
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -572,7 +572,8 @@ export class EventRepository {
     }
 
     const containsLinkPreview = newEventData.previews && !!newEventData.previews.length;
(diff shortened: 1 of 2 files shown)


src/script/event/EventRepository.ts
```diff
@@ -572,7 +572,8 @@ export class EventRepository {
     }
 
     const containsLinkPreview = newEventData.previews && !!newEventData.previews.length;
-    const isRetryAttempt = originalEvent.status === StatusType.FAILED;
+    const isRetryAttempt =
+      originalEvent.status === StatusType.FAILED || originalEvent.status === StatusType.FEDERATION_ERROR;
```
aweiss-dev: This could be a typeguard function to be able to reuse this check aka
isFailed(event) and isFederationError(event)

src/script/event/EventRepository.ts
```diff
@@ -565,14 +565,22 @@ export class EventRepository {
   private handleMessageUpdate(originalEvent: EventRecord, newEvent: MessageAddEvent) {
     const newEventData = newEvent.data;
     const originalData = originalEvent.data;
+
+    const isFailed = (status: StatusType | undefined): status is StatusType.FAILED => {
+      return status === StatusType.FAILED;
+    };
```
aweiss-dev: was more thinking of this approach:


const isFailedEvent = (event: unknown): event is FailedEvent => event && event.status === StatusType.FAILED;

placed under the guards folder in the webapp

aweiss-dev (approved):

src/script/event/EventRepository.ts
przemvs: Maybe 
```
const isFailedEvent = (event: unknown): event is FailedEvent => 'status' in event && event.status ===  StatusType.FAILED;
```

przemvs: Because event will be unknown and we don't know if status exist in event :) @aweiss-dev

PatrykBuniX (approved):

PR #15344, opened 2023-06-15
runfix: subscribe to mls conversation state change in conversation state

PatrykBuniX:
When updating `established` / `pendingWelcome` state we have to do it reactive way, so overwrite the reference. Otherwise even when subscribing to store updates, just adding new item to the set won't notify the subscribers.

In conversation state we subscribe to store changes and notify subscribers of `sortedConversations` (including `visibleConversations`).

### DIFF
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -71,6 +71,12 @@ export class ConversationState {
     this.selfMLSConversation = ko.pureComputed(() =>
       this.conversations().find(conversation => isMLSConversation(conversation) && isSelfConversation(conversation)),
     );
+
+    //anytime mls conversation state changes, we update the sorted conversations that will recalculate visible conversations
+    useMLSConversationState.subscribe(() => {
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationState.ts
```diff
@@ -71,6 +71,12 @@ export class ConversationState {
     this.selfMLSConversation = ko.pureComputed(() =>
       this.conversations().find(conversation => isMLSConversation(conversation) && isSelfConversation(conversation)),
     );
+
+    //anytime mls conversation state changes, we update the sorted conversations that will recalculate visible conversations
+    useMLSConversationState.subscribe(() => {
+      this.sortedConversations.notifySubscribers();
+    });
```
PatrykBuniX, line 78: `sortedConversations` is used when calculating `visibleConversations`, so `visibleConversations` will recalculate every time mls conversations store gets updated.

przemvs (approved):

V-Gira (approved):

PR #15345, opened 2023-06-16
runfix: don't include conv domain when deleting conv form the store

PatrykBuniX:
This seems to be in the code forever, when deleting the conversation locally, we lookup for the domain field on a conversation an try to construct the key like `${id}@${domain}`, but when saving the conversation in the store, we never include the domain. We use plain `conversationEntity.id` as a `primaryKey` in the `indexedDB` store. 

So now, when deleting conversation locally, it is never being removed from the store. Since conversations are never deleted, clients load them on app init and might try to perform some actions (like periodic key material updates for mls-capable conversations) what results in some errors in the console (conversations were deleted on backend).

I've identified all the places where we add db conversation entry (`storageService.save(StorageSchemata.OBJECT_STORE.CONVERSATIONS, ...)`) and we always pass plain conversation id, see:

https://github.com/wireapp/wire-webapp/blob/d5ef1bbfe2b8d7a2d61c5952c7c908662381d3ae/src/script/conversation/ConversationService.ts#L355

https://github.com/wireapp/wire-webapp/blob/d5ef1bbfe2b8d7a2d61c5952c7c908662381d3ae/src/script/conversation/ConversationService.ts#L373

https://github.com/wireapp/wire-webapp/blob/d5ef1bbfe2b8d7a2d61c5952c7c908662381d3ae/src/script/conversation/ConversationService.ts#L304

I think we can simply start using plain conversation id when deleting conversations.

### DIFF
--- a/src/script/conversation/ConversationService.ts
+++ b/src/script/conversation/ConversationService.ts
@@ -295,9 +295,8 @@ export class ConversationService {
    * Deletes a conversation entity from the local database.
    * @returns Resolves when the entity was deleted
    */
-  async deleteConversationFromDb({id, domain}: QualifiedId): Promise<string> {
-    const key = domain ? `${id}@${domain}` : id;
-    const primaryKey = await this.storageService.delete(StorageSchemata.OBJECT_STORE.CONVERSATIONS, key);
+  async deleteConversationFromDb(conversationId: string): Promise<string> {
+    const primaryKey = await this.storageService.delete(StorageSchemata.OBJECT_STORE.CONVERSATIONS, conversationId);
     return primaryKey;
   }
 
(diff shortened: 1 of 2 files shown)


V-Gira (approved):

src/script/conversation/ConversationService.ts
```diff
@@ -295,9 +295,8 @@ export class ConversationService {
    * Deletes a conversation entity from the local database.
    * @returns Resolves when the entity was deleted
    */
-  async deleteConversationFromDb({id, domain}: QualifiedId): Promise<string> {
-    const key = domain ? `${id}@${domain}` : id;
-    const primaryKey = await this.storageService.delete(StorageSchemata.OBJECT_STORE.CONVERSATIONS, key);
+  async deleteConversationFromDb(conversationId: string): Promise<string> {
+    const primaryKey = await this.storageService.delete(StorageSchemata.OBJECT_STORE.CONVERSATIONS, conversationId);
```
przemvs: And you can remove `return primaryKey;`

```suggestion
    return this.storageService.delete(StorageSchemata.OBJECT_STORE.CONVERSATIONS, conversationId);
```

przemvs (approved):

PR #15367, opened 2023-06-19
runfix: delete conversation from the store after leaving with clear

PatrykBuniX:
### Problem
We were never clearing the conversation locally (from the local DB) after user has left the conversation with "clear conversation data" field checked. 

### Solution
After receiving member-leave event, if event creator is a self user and "clear data" checkbox was marked, we delete the conversation locally.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1547,7 +1547,7 @@ export class ConversationRepository {
    * @param conversationEntity Conversation to clear
    * @param leaveConversation Should we leave the conversation before clearing the content?
    */
-  public clearConversation(conversationEntity: Conversation, leaveConversation = false) {
+  public async clearConversation(conversationEntity: Conversation, leaveConversation = false) {
     const isActiveConversation = this.conversationState.isActiveConversation(conversationEntity);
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -1626,12 +1626,11 @@ export class ConversationRepository {
...
-    if (clearContent) {
-      this.clearConversation(conversationEntity, false);
-    }
-
     const userQualifiedId = this.userState.self().qualifiedId;
-    return this.removeMemberFromConversation(conversationEntity, userQualifiedId);
+
+    return clearContent
+      ? this.clearConversation(conversationEntity, true)
+      : this.removeMemberFromConversation(conversationEntity, userQualifiedId);
```
PatrykBuniX, line 1633: This is the main change - the `true` param in clearConversation method is `leaveConversation` - so when called with `true`, we clear the conversation locally and leave it. Previously it was called with false - conversation was not cleared and we were calling just `removeMemberFromConversation` - leaving a conversation without clearing the local store, even though user has marked the checkbox

V-Gira (approved):

przemvs (approved):

PR #15369, opened 2023-06-20
fix: handle failure to send mls messages (WPB-2300)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2300" title="WPB-2300" target="_blank">BugWPB-2300</a>  [Web] On sending a messages to a conversation owned by an offline b-e with MLS the option to retry does dot appear
  </td></table>
  <br />
 

### Issues

1. when sending a message to an mls conversation, the option to retry do not appear when the owning b-e is offline

2. same thing when sending a message while internet is offline using api v4

3. wrong styling in some cases

### Solutions

1. bump core to 40.5.4, see https://github.com/wireapp/wire-web-packages/pull/5254

2. handle `AxiosError` as failure to send

3. make sure error messages appear in red, and unsent messages gray

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -50,7 +50,6 @@ import {roundLogarithmic} from 'Util/NumberUtil';
 import {matchQualifiedIds} from 'Util/QualifiedId';
 import {capitalizeFirstChar} from 'Util/StringUtil';
 import {TIME_IN_MILLIS} from 'Util/TimeUtil';
-import {isBackendError} from 'Util/TypePredicateUtil';
 import {loadUrlBlob, supportsMLS} from 'Util/util';
 import {createUuid} from 'Util/uuid';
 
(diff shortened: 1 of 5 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -809,9 +808,7 @@ export class MessageRepository {
       }
       return result;
     } catch (error) {
-      if (isBackendError(error)) {
-        await this.updateMessageAsFailed(conversation, payload.messageId, error);
-      }
+      await this.updateMessageAsFailed(conversation, payload.messageId, error);
```
V-Gira, line 812: I could alternatively check for Axios errors with `isAxiosError(error)` but I think we would want to treat any error as a failure to send with the option to retry

src/script/conversation/MessageRepository.ts
```diff
@@ -1187,7 +1184,7 @@ export class MessageRepository {
     return undefined;
   }
 
-  private async updateMessageAsFailed(conversationEntity: Conversation, eventId: string, error: BackendError) {
+  private async updateMessageAsFailed(conversationEntity: Conversation, eventId: string, error: BackendError | any) {
```
PatrykBuniX: Can't we go with just `unknown` and check what type of error it is inside the method?

V-Gira: like this?
```
if (isBackendError(error) && error.label === BackendErrorLabel.FEDERATION_REMOTE_ERROR) 
```

PatrykBuniX: yes

PatrykBuniX (approved):

PR #15394, opened 2023-06-26
feat: bump core with core-crypto draft 20

PatrykBuniX:
Bumps core with corecrypto draft-20 support.


PatrykBuniX:
https://github.com/wireapp/wire-webapp/pull/15393

PR #15430, opened 2023-07-05
fix: display cannot find user warning instead of placeholder profile (WPB-2877)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2877" title="WPB-2877" target="_blank">BugWPB-2877</a>  [Web] Link to user profile with invalid userID shows default template
  </td></table>
  <br />
 

### Issue
- when trying to access a user profile that doesn't exist via url, a placeholder profile would appear
image

### Solution
- Expected behavior is to show a warning that a user could not be found
image

### DIFF
--- a/src/script/components/Modals/UserModal/UserModal.tsx
+++ b/src/script/components/Modals/UserModal/UserModal.tsx
@@ -156,7 +156,7 @@ const UserModal: React.FC<UserModalProps> = ({
       userRepository
         .getUserById(userId)
         .then(user => {
-          if (user.isDeleted) {
+          if (user.isDeleted || !user.isAvailable()) {
             setUserNotFound(true);
             return;
           }


src/script/components/Modals/UserModal/UserModal.tsx
```diff
@@ -156,7 +156,7 @@ const UserModal: React.FC<UserModalProps> = ({
       userRepository
         .getUserById(userId)
         .then(user => {
-          if (user.isDeleted) {
+          if (user.isDeleted || !user.isAvailable()) {
```
V-Gira, line 159: `user.isAvailable()` returns true if a user is successfully fetched

thisisamir98 (approved):

PR #15431, opened 2023-07-05
feat: join mls group after migration finalisation

PatrykBuniX:
Joins MLS group after migration finalisation.

Users who did not upgrade their clients on time will miss conversation protocol change events (both to `mixed` and to `mls`) and welcome messages. It's possible that there were some messages sent using MLS already, so we have to insert system message that user might have missed some messages due to proteus to MLS migration.

Todo:
- [x] unit tests


src/i18n/en-US.json
```diff
@@ -390,6 +390,7 @@
   "conversationMemberRemovedYou": "[bold]You[/bold] removed {{users}}",
   "conversationMessageDelivered": "Delivered",
   "conversationMissedMessages": "You haven’t used this device for a while. Some messages may not appear here.",
+  "conversationJoinedAfterMLSMigrationFinalisation": "You have joined this conversation after the MLS migration was completed. You might miss some messages.",
```
PatrykBuniX, line 393: This will change after we get the copy

src/script/main/app.ts
```diff
@@ -433,14 +443,24 @@ export class App {
...
+        await joinConversationsAfterMigrationFinalisation({
+          updatedConversations: conversations,
+          initialDatabaseConversations: initialLocalConversations,
+          core: this.core,
+          conversationRepository: conversationRepository,
+        });
+
         //join all the mls groups we're member of and have not yet joined (eg. we were not send welcome message)
-        await initMLSConversations(mlsConversations, this.core);
+        await joinNewMLSConversations(mlsConversations, this.core);
```
PatrykBuniX, line 463: - first we initialise already existing mls conversations (so schedule periodic key materials)
- then we register team and self conversations if they havent been yet established
- then we join the mls conversations that have been already migrated to mls (user was not supporting mls when protocol change events or welcome messages were sent)
- in the end we join all new mls conversations we are members of but have not yet joined on the current client

src/script/entity/message/JoinedAfterMLSMigrationFinalisationMessage.ts
```diff
@@ -0,0 +1,33 @@
...
+import {SystemMessageType} from 'src/script/message/SystemMessageType';
+import {t} from 'Util/LocalizerUtil';
+
+import {SystemMessage} from './SystemMessage';
+
+export class JoinedAfterMLSMigrationFinalisationMessage extends SystemMessage {
+  constructor() {
+    super();
+    this.system_message_type = SystemMessageType.JOINED_AFTER_MLS_MIGRATION_FINALISATION;
+    this.caption = ko.pureComputed(() => t('conversationJoinedAfterMLSMigrationFinalisation'));
```
thisisamir98, line 31: if this always ends up with t('conversationJoinedAfterMLSMigrationFinalisation')
why is it a pure computed? I don't see any reason to use that here

src/script/entity/message/JoinedAfterMLSMigrationFinalisationMessage.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
thisisamir98:
```suggestion
 * Copyright (C) 2023 Wire Swiss GmbH
```

src/script/entity/message/JoinedAfterMLSMigrationFinalisationMessage.ts
PatrykBuniX: caption field on SystemMessage is typed as `ko.PureComputed<string>`, it won't accept plain string, so we have to provide it this way

thisisamir98 (approved):

PR #15452, opened 2023-07-11
fix: import events without id field

PatrykBuniX:
Regression introduced in https://github.com/wireapp/wire-webapp/pull/15085

There are some types of events like `conversation.member-join` or `conversation.member-leave` that do not contain `"id"`. `"id"` field was typed as mandatory, while it was not true for mentioned types of events. 

`table.where('id').anyOf(ids).toArray()` was throwing error when the list of `ids` contained at least one `undefined` value. 

The fix is to type id as optional and filter out the events that do not contain id.

### DIFF
--- a/src/script/backup/BackupService.ts
+++ b/src/script/backup/BackupService.ts
@@ -119,7 +119,7 @@ export class BackupService {
       return entities.length;
     }
 
-    const ids = entities.map(generateId).filter((id): id is string => id !== undefined);
+    const ids = entities.map(generateId).filter((id): id is string => typeof id === 'string');
     const existingEntities = await table.where('id').anyOf(ids).toArray();
     const newEntities = entities.filter(
       entity => !existingEntities.some(existingEntity => generateId(existingEntity) === generateId(entity)),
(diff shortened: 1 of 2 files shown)


src/script/backup/BackupService.ts
```diff
@@ -112,14 +112,14 @@ export class BackupService {
...
-    generateId?: (entry: T) => string,
+    generateId?: (entry: T) => string | undefined,
   ): Promise<number> {
     if (!generateId) {
       await table.bulkAdd(entities);
       return entities.length;
     }
 
-    const ids = entities.map(generateId);
+    const ids = entities.map(generateId).filter((id): id is string => id !== undefined);
```
aweiss-dev: If you want to ensure that if is a string, why don't you check for it instead of undefined?

aweiss-dev (approved):

src/script/backup/BackupService.ts
PatrykBuniX: Yeah it's either `undefined` or `string` in this case so doesn't really matter, but I find your idea more readable, just updated it ;)

PR #15479, opened 2023-07-19
fix: Capitalise the first letter of the first word in the emoji name displayed on the right-side panel in the message details.

arjita-mitra:
----

- The **PR Title**
  - Capitalise the first letter of the first word in the emoji name displayed on the right-side panel in the message details.

- The **PR Description**
  - Capitalise the first letter of the first word in the emoji name displayed on the right-side panel in the message details.
----
##### References
1. https://wearezeta.atlassian.net/browse/WPB-3293

### DIFF
--- a/src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
+++ b/src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
@@ -229,6 +229,13 @@ const MessageDetails: FC<MessageDetailsProps> = ({
 
   const onParticipantClick = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
+  function capitalizeFirstLetter(emojiName: string) {
+    if (emojiName && emojiName.trim() !== '') {
+      return emojiName.charAt(0).toUpperCase() + emojiName.slice(1);
+    }
+    return '';
+  }
+
   return (
     <div id="message-details" className="panel__page message-details">
       <PanelHeader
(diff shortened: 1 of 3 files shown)


src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
```diff
@@ -229,6 +229,13 @@ const MessageDetails: FC<MessageDetailsProps> = ({
 
   const onParticipantClick = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
+  function capitalizeFirstLetter(emojiName: string) {
+    if (emojiName && emojiName.trim() !== '') {
+      return emojiName.charAt(0).toUpperCase() + emojiName.slice(1);
```
PatrykBuniX: We have `capitalizeFirstChar` method in our `StringUtil.ts` let's use it

arjita-mitra: I want to perform the sanity check.

if (emojiName && emojiName.trim() !== '') 

shall i add it to the capitalizeFirstChar method in StringUtil.ts?

PatrykBuniX: No, we can do just `return capitalizeFirstChar(emojiName)` after sanity check is valid - we will need to rename the original method then.

arjita-mitra: why not adding it to the capitalizeFirstChar method? this is a basic sanity check, it should be there in the method.

PatrykBuniX (approved):
👌

PR #15482, opened 2023-07-21
refactor: map connection conversation

PatrykBuniX:
Refactor of mapConnection method on conversation repository to use async await instead of `.then()`. Improves readability and it's also a prep work for 1:1 conversations with MLS.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1250,7 +1250,7 @@ export class ConversationRepository {
     }
 
     if (connectionEntity.isConnected() || connectionEntity.isOutgoingRequest()) {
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -1240,46 +1240,56 @@ export class ConversationRepository {
     }
   };
 
+  private readonly getConnectionConversation = (connectionEntity: ConnectionEntity) => {
+    const qualifiedId = connectionEntity.conversationId;
```
przemvs:
```suggestion
      const {conversationId: qualifiedId} = connectionEntity;
```

przemvs (approved):
LGTM! :) 1 small thing

PR #15497, opened 2023-07-24
feat: introduce variations to fail to add users (WPB-2299)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2299" title="WPB-2299" target="_blank">StoryWPB-2299</a>  [Web] Errors when adding users from non-fully connected graph
  </td></table>
  <br />
 

#### feature

- add an `errorMesssage` prop to the `failedToAddUser` component


src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -122,41 +142,29 @@ const FailedToAddUsersMessage: React.FC<FailedToAddUsersMessageProps> = ({
       <div className="message-body">
         {isOpen && (
           <>
-            {total <= 1 && (
```
V-Gira, line 125: condition `isOpen` is true when `total > 1` so this will never happen

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -49,6 +55,7 @@ const config = Config.getConfig();
 const FailedToAddUsersMessage: React.FC<FailedToAddUsersMessageProps> = ({
   isMessageFocused,
   message,
+  errorMessageType = ErrorMessageType.offlineBackEnd,
```
V-Gira, line 58: The different error messages are not part of the scope of the September release.
The only error message displayed stays the same as offline back-end

przemvs (approved):

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -101,15 +108,28 @@ const FailedToAddUsersMessage: React.FC<FailedToAddUsersMessageProps> = ({
...
+            </p>
+          )}
+          {total > 1 && (
+            <p
+              css={warning}
+              dangerouslySetInnerHTML={{
+                __html: t(`failedToAddParticipantsPlural`, {total: total.toString()}),
+              }}
+            />
+          )}
```
PatrykBuniX, line 132: Could be simplified to just `total <= 1 ? () : ()`, no?

V-Gira: It could, I kept it the way @thisisamir98  did it.
Probably easier to read?

PatrykBuniX: The result will be the same, I'm fine with it

PatrykBuniX (approved):

PR #15507, opened 2023-07-26
feat: Migrate input bar to Lexical

przemvs:
## Description

This is a major refactor of the InputBar to now use https://lexical.dev/
This will fix issues like dealing with international keyboard better and also improve the overall architecture of the input bar

## Note for the reviewers

This is a huge PR, no doubt about this. Note that:
- most of the logic is hidden inside of the `RichTextEditor` folder, there lives the core of the input bar logic. 

E2E test to run:

- [x] mentioning someone (with the @ symbol)
- [x] deleting a mention
- [x] reply to a message
- [x] switching conversation and check that text is still there (with mentions)
- [x] add emojis using the `:` symbol
- [x] inline replacement of emojis (eg. `;)` should be converted to 😉 )
- [x] typing indicator (start typing, stop typing, delete entire message, send message, switch conversation)
- [x] markdown editing

To Fix:

- [x] searching for a user in the current conversation switches focus back to the inputbar
- [x] pressing arrow up when in mention mode will load the last sent message in edit mode


src/style/content/conversation/input-bar.less
```diff
@@ -32,23 +32,34 @@
...
+.conversation-input-bar-text-placeholder::before {
+  padding-left: 8px;
+  color: var(--gray-70);
+  content: attr(data-placeholder);
+  .label-xs;
+
+  body.theme-dark & {
+    color: var(--gray-60);
+  }
+}
```
V-Gira:
```suggestion
.conversation-input-bar-text-placeholder::before {
  padding-left: 8px;
  color: var(--text-input-placeholder);
  content: attr(data-placeholder);
  .label-xs;
}
```

src/script/components/LexicalInput/LexicalInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+      amplify.subscribe(WebAppEvents.PROPERTIES.UPDATE.EMOJI.REPLACE_INLINE, setShouldReplaceEmoji);
+      amplify.subscribe(WebAppEvents.PROPERTIES.UPDATED, (properties: WebappProperties) => {
+        setShouldReplaceEmoji(properties.settings.emoji.replace_inline);
+      });
+    }, []);
+
+    return (
+      <LexicalComposer initialConfig={editorConfig}>
+        <div className="controls-center">
+          <div css={{width: '100%'}} className={cx('input-bar--wrapper')}>
```
arjita-mitra: why one inline css, when we are using a class?

src/script/components/LexicalInput/LexicalInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const [shouldReplaceEmoji, setShouldReplaceEmoji] = useState<boolean>(
+      propertiesRepository.getPreference(PROPERTIES_TYPE.EMOJI.REPLACE_INLINE),
+    );
+
+    const editorConfig: InitialConfigType = {
+      namespace: 'WireLexicalEditor',
+      theme,
+      onError(error: unknown) {
+        // eslint-disable-next-line no-console
+        console.log('[LexicalInput.tsx] przemvs error', error);
```
arjita-mitra: are we not logging the error in datadog? how about using logger.error?

src/script/components/LexicalInput/LexicalInput.tsx
```diff
@@ -0,0 +1,177 @@
...
+              }
+              placeholder={<Placeholder text={placeholder} hasLocalEphemeralTimer={hasLocalEphemeralTimer} />}
+              ErrorBoundary={LexicalErrorBoundary}
+            />
+
+            <BeautifulMentionsPlugin onSearch={queryMentions} />
+
+            <OnChangePlugin
+              onChange={(editorState, lexicalEditor) => {
+                lexicalEditor.registerTextContentListener(textContent => {
```
arjita-mitra: useCallback to prevent unnecessary function re-creations on each render.

src/script/components/LexicalInput/components/EditMessage.tsx
```diff
@@ -0,0 +1,48 @@
...
+    amplify.subscribe(WebAppEvents.CONVERSATION.MESSAGE.EDIT, (messageEntity: ContentMessage) => {
+      onMessageEdit(messageEntity, editor);
+    });
+
+    return () => {
+      amplify.unsubscribeAll(WebAppEvents.CONVERSATION.MESSAGE.EDIT);
+    };
+  }, [editor]);
+
+  return <></>;
```
arjita-mitra: we should return null here if we don't want to return anything.
return <></> when you need to return multiple elements without adding an extra parent node in the DOM.

arjita-mitra (changes requested):

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+  classNameFocused?: string;
+}
+
+export const Mention = (props: BeautifulMentionComponentProps) => {
+  const {mention, className = '', classNameFocused = '', nodeKey} = props;
+  const [editor] = useLexicalComposerContext();
+  const [isSelected, setSelected, clearSelection] = useLexicalNodeSelection(nodeKey);
+  const [selection, setSelection] = useState<RangeSelection | NodeSelection | GridSelection | null>(null);
+  const isFocused = $isNodeSelection(selection) && isSelected;
+  const ref = useRef(null);
```
arjita-mitra: specify the type of the ref here. useRef<HTMLSpanElement>(null);

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+    (payload: KeyboardEvent) => {
+      if (isSelected && $isNodeSelection($getSelection())) {
+        payload.preventDefault();
+        const node = $getNodeByKey(nodeKey);
+        if ($isBeautifulMentionNode(node)) {
+          node.remove();
+        }
+        setSelected(false);
+      }
+      return false;
```
arjita-mitra: instead of return false we can use event.preventDefault() to prevent default behaviour

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+    },
+    [isSelected, nodeKey, setSelected],
+  );
+
+  const onArrowLeftPress = useCallback(
```
arjita-mitra: adding some function description and some comment would be helpful for us to understand the purpose of the less intuitive parts of the code.

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+    (event: KeyboardEvent) => {
+      const node = $getNodeByKey(nodeKey);
+      if (!node || !node.isSelected()) {
+        return false;
+      }
+      let handled = false;
+      const nodeToSelect = getPreviousSibling(node);
+      if ($isElementNode(nodeToSelect)) {
+        nodeToSelect.selectEnd();
```
arjita-mitra: why are you using only if blocks? using else if will make it more readable. Alternatively you can use separate helper functions for each type of node selection to make the code more self-explanatory.

```
 const selectEnd = ($isElementNode(node: Node)): void => {
    node.selectEnd();
    handled = true;
  };

  const selectText = ($isTextNode(node: Node)): void => {
    node.select();
    handled = true;
  };
  
   if ($isElementNode(nodeToSelect)) {
    selectEnd(nodeToSelect);
  } else if ($isTextNode(nodeToSelect)) {
    selectText(nodeToSelect);
  } 
  // rest of the code
```

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+    (event: KeyboardEvent) => {
+      const node = $getNodeByKey(nodeKey);
+      if (!node || !node.isSelected()) {
+        return false;
+      }
+      let handled = false;
+      const nodeToSelect = getNextSibling(node);
+      if ($isElementNode(nodeToSelect)) {
+        nodeToSelect.selectStart();
+        handled = true;
```
arjita-mitra: same feedback goes here too

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+    if (isFocused) {
+      $setSelection(null);
+      return true;
+    }
+    return false;
+  }, [isFocused]);
+
+  // Make sure that the focus is removed when clicking next to the mention
+  const onSelectionChange = useCallback(() => {
+    // if (IS_IOS && isSelected) {
```
arjita-mitra: remove commented code

src/script/components/LexicalInput/components/SendMessageButton/SendMessageButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+const createMentionEntity = (user: Pick<User, 'id' | 'name' | 'domain'>, mentionPosition: number): MentionEntity => {
+  const userName = user.name();
+  const mentionLength = userName.length + 1;
+
+  return new MentionEntity(mentionPosition, mentionLength, user.id, user.domain);
+};
+
+interface SendMessageButtonProps {
+  textValue: string;
+  onSend: any;
```
arjita-mitra: why are we not passing specific type?

`type OnSendFunction = (text: string) => void;`

do you know why onSend function returns 2 different types?

 const onSend = (text: string): void | boolean

src/script/components/LexicalInput/components/SendMessageButton/SendMessageButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+        const isEmojisDropdownActive = document.querySelector('.emoji-menu');
+
+        if (isMentionsDropdownActive || isEmojisDropdownActive) {
+          return false;
+        }
+
+        if (event?.shiftKey) {
+          return true;
+        }
+        const mentionEntities = getMentionsToSend(editor, textValue, mentions);
```
arjita-mitra: this block of the code is same as the onclick handler. we should create a function and re-use it.

```
const handleSendMessage = (editor: LexicalEditor, textValue: string, mentions: User[]): void => {
  const mentionEntities = getMentionsToSend(editor, textValue, mentions);
  onSend(textValue, mentionEntities);
  editor.update(() => {
    $getRoot().clear();
  });
};
```

src/script/components/LexicalInput/hooks/useDebounce.ts
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2022 Wire Swiss GmbH
+ * Copyright (C) 2023 Wire Swiss GmbH
```
arjita-mitra: if this is an existing file then copyright should't change

src/script/components/LexicalInput/hooks/useIsFocused.ts
```diff
@@ -0,0 +1,52 @@
...
+      ),
+      editor.registerCommand(
+        BLUR_COMMAND,
+        () => {
+          setHasFocus(false);
+          return false;
+        },
+        COMMAND_PRIORITY_NORMAL,
+      ),
+    );
```
arjita-mitra: to avoid any potential memory leak we should unregister commands when the component unmounts.

```
const handleFocus = () => {
      setHasFocus(true);
    };
```

    const handleBlur = () => {
      setHasFocus(false);
    };
```
 return () => {
      editor.unregisterCommand(FOCUS_COMMAND, handleFocus);
      editor.unregisterCommand(BLUR_COMMAND, handleBlur);
    };
```

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+  );
+
+  const onArrowLeftPress = useCallback(
+    (event: KeyboardEvent) => {
+      const node = $getNodeByKey(nodeKey);
+      if (!node || !node.isSelected()) {
+        return false;
+      }
+      let handled = false;
+      const nodeToSelect = getPreviousSibling(node);
```
arjita-mitra: add proper checks for potential null or undefined values when calling functions like $getNodeByKey(nodeKey) and $isElementNode, $isTextNode, and $isDecoratorNode. 

```
if (nodeToSelect) {
   if ($isElementNode(nodeToSelect)) {
  // code here
  }
} else {
    node.selectPrevious();
    handled = true;
}
```

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,212 @@
...
+      }
+      if (handled) {
+        event.preventDefault();
+      }
+      return handled;
+    },
+    [nodeKey],
+  );
+
+  const onArrowRightPress = useCallback(
```
arjita-mitra: to avoid code duplication between left and right arrow key function we can pass an extra param called direction. and handle it based on it.
```

const handleArrowKeyEvent = useCallback(
  (event: KeyboardEvent, direction: 'left' | 'right') => {
  
  const node = $getNodeByKey(nodeKey);

    if (!node || !node.isSelected()) {
      return false;
    }

    let handled = false;
    let nodeToSelect: Node | null;

    if (direction === 'left') {
      nodeToSelect = getPreviousSibling(node);
    } else {
      nodeToSelect = getNextSibling(node);
    }

    if (nodeToSelect) {
      if ($isElementNode(nodeToSelect)) {
        if (direction === 'left') {
          nodeToSelect.selectEnd();
        } else {
          nodeToSelect.selectStart();
        }
        handled = true;
      } 
      // rest of the code
      }
}
```

arjita-mitra (changes requested):

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
```diff
@@ -0,0 +1,372 @@
...
+          }
+          return renamed;
+        },
+        COMMAND_PRIORITY_LOW,
+      ),
+      editor.registerCommand(
+        OPEN_MENTIONS_MENU_COMMAND,
+        ({trigger}) => insertMention(triggers, trigger),
+        COMMAND_PRIORITY_LOW,
+      ),
```
arjita-mitra: use cleanup function to un-register

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
```diff
@@ -0,0 +1,372 @@
...
+    }
+  }, []);
+
+  const rootElement = editor.getRootElement();
+
+  useEffect(() => {
+    return mergeRegister(
+      editor.registerCommand(
+        KEY_DOWN_COMMAND,
+        event => {
```
arjita-mitra: the useEffect hook is very large, making it harder to read. Can you break down the useEffect into smaller, more focused functions that handle specific aspects of the effect. This can make the code easier to understand and manage. For example-

 ```
// Handle arrow key down event
  useEffect(() => {
    const handleArrowKeyDown = (event: KeyboardEvent) => {
      // ... Arrow key handling logic ...
    };

    const arrowKeyDownListener = editor.registerCommand(KEY_DOWN_COMMAND, handleArrowKeyDown, COMMAND_PRIORITY_LOW);
    return () => {
      arrowKeyDownListener();
    };
  }, [editor]);


```

another useEffect for handleInsertMention and so on

File: src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx [hunk 20]
```diff
@@ -0,0 +1,372 @@
...
+
+  return (
+    <LexicalTypeaheadMenuPlugin
+      onQueryChange={setQueryString}
+      onSelectOption={handleSelectOption}
+      triggerFn={checkForMentionMatch}
+      options={options}
+      onClose={handleClose}
+      menuRenderFn={(anchorElementRef, {selectedIndex, selectOptionAndCleanUp, setHighlightedIndex}) => {
+        if (!anchorElementRef.current || !options.length) {
```
arjita-mitra: The menuRenderFn prop contains complex rendering logic.Extract this logic into a separate component or function to improve code readability and maintainability.

src/script/components/LexicalInput/plugins/DraftStatePlugin.tsx
```diff
@@ -0,0 +1,53 @@
...
+      const textValue = getTextValue(editor);
+      setInputValue(textValue);
+    }
+  }, [editor, loadDraftState, setInputValue]);
+
+  useEffect(() => {
+    void getDraftState();
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  return <></>;
```
arjita-mitra: we should return null

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
```diff
@@ -0,0 +1,372 @@
...
+                          setHighlightedIndex(index);
+                          selectOptionAndCleanUp(option);
+                        }}
+                        onMouseEnter={() => {
+                          setHighlightedIndex(index);
+                        }}
+                      />
+                    );
+                  })
+                  .reverse()}
```
arjita-mitra: why are you reversing the list here?

src/script/components/LexicalInput/components/EditMessage.tsx
przemvs: If we returning null, we getting error with null return.

src/script/components/LexicalInput/components/Mention/Mention.tsx
przemvs: registerCommand(COMMAND, listener, PRIORITY)

listener requires return boolean;

src/script/components/LexicalInput/plugins/EmojiPickerPlugin/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  const checkForTriggerMatch = useBasicTypeaheadTriggerMatch(':', {
+    minLength: 0,
+  });
+
+  const options: Array<EmojiOption> = useMemo(() => {
+    return emojiOptions
+      .filter((option: EmojiOption) => {
+        return queryString != null
+          ? new RegExp(queryString, 'gi').exec(option.title) || option.keywords != null
```
arjita-mitra: it's better to avoid unnecessary regular expression object creation on every render. You can define the regex object outside the filter method.

src/script/components/LexicalInput/components/Mention/Mention.tsx
przemvs: It comes from Mention plugin from Lexical.

src/script/components/LexicalInput/components/Mention/Mention.tsx
przemvs: It comes from Mention plugin from Lexical, and this if's is more readable ;)

src/script/components/LexicalInput/plugins/EmojiPickerPlugin/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const increaseUsageCount = (emojiName: string | null): void => {
+    if (emojiName) {
+      emojiUsageCount[emojiName] = getUsageCount(emojiName) + 1;
+      storeValue(StorageKey.CONVERSATION.EMOJI_USAGE_COUNT, emojiUsageCount);
+    }
+  };
+
+  useEffect(() => {
+    // @ts-ignore
+    import('../../utils/emoji-list.ts').then(file => setEmojis(file.default));
```
arjita-mitra: why are we using dynamic import. When using dynamic imports, like ../../utils/emoji-list.ts, there are potential issues related to static analysis and tree-shaking.

src/script/components/LexicalInput/components/SendMessageButton/SendMessageButton.tsx
przemvs: Missed type.

Probably forgot to change this

src/script/components/LexicalInput/components/SendMessageButton/SendMessageButton.tsx
przemvs: Good catch!

src/script/components/LexicalInput/plugins/EmojiPickerPlugin/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const rootElement = lexicalEditor.getRootElement();
+
+  return (
+    <LexicalTypeaheadMenuPlugin
+      onQueryChange={setQueryString}
+      onSelectOption={onSelectOption}
+      triggerFn={checkForTriggerMatch}
+      options={options}
+      menuRenderFn={(anchorElementRef, {selectedIndex, selectOptionAndCleanUp, setHighlightedIndex}) => {
+        if (!anchorElementRef.current || !options.length) {
```
arjita-mitra: complex logical operation, creating of new function on every render should be avoided. extract the menuRenderFn from here and place it outside render. 

we can use useMemo for getPosition

src/script/components/LexicalInput/hooks/useDebounce.ts
przemvs: It's not used, I remove hook.

src/script/components/LexicalInput/hooks/useIsFocused.ts
przemvs: It's clearing :) https://lexical.dev/docs/api/modules/lexical_utils#mergeregister

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+      ),
+    );
+  }, [editor, onShiftTab]);
+
+  return <></>;
+};
+
+// Example of usage global commands
+// editor.registerCommand(
+//   ON_SHIFT_TAB,
```
arjita-mitra: remove unused code

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+        () => {
+          onShiftTab();
+          return true;
+        },
+        COMMAND_PRIORITY_HIGH,
+      ),
+    );
+  }, [editor, onShiftTab]);
+
+  return <></>;
```
arjita-mitra: return null

src/script/components/LexicalInput/plugins/LexicalEditorRefPlugin.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  if (typeof editorRef === 'function') {
+    editorRef(editor);
+  } else if (typeof editorRef === 'object') {
+    if (editorRef) {
+      editorRef.current = editor;
+    }
+  }
+
+  return <></>;
```
arjita-mitra: return null

src/script/components/LexicalInput/utils/mention-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+    prevNode,
+    nextNode,
+    cursorAtStartOfNode,
+    cursorAtEndOfNode,
+    wordCharBeforeCursor,
+    wordCharAfterCursor,
+  };
+}
+
+export function insertMention(triggers: string[], trigger: string, value?: string) {
```
arjita-mitra: To make the insertMention function more concise and readable, we can rewrite it like this.

```
// Insert a mention node or a text node with the trigger to open the mention menu.
  const mentionNode = value ? $createBeautifulMentionNode(trigger, value) : $createTextNode(trigger);

  // Insert a mention with a leading space if the node at the cursor is not a text node.
  if (!isTextNodeAtCursor(node, offset)) {
    insertWithSpace(node, mentionNode, $createTextNode(' '));
    return true;
  }

  const nodes: LexicalNode[] = [];
  if (shouldInsertSpaceBefore(node, offset, prevNode)) {
    nodes.push($createTextNode(' '));
  }
  nodes.push(mentionNode);
  if (shouldInsertSpaceAfter(node, offset, nextNode)) {
    const spaceNode = $createTextNode(' ');
    nodes.push(spaceNode);
    spaceNode.selectPrevious();
  } 

  return true;

// write those functions here
```

src/script/components/LexicalInput/utils/mention-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  // Insert a mention with a leading space if the node at the cursor is not a text node.
+  if (!($isParagraphNode(node) && offset === 0) && !$isTextNode(node)) {
+    selection.insertNodes([$createTextNode(' '), mentionNode]);
+    return true;
+  }
+
+  let spaceNode: TextNode | null = null;
+  const nodes: LexicalNode[] = [];
+  if (wordCharBeforeCursor || (cursorAtStartOfNode && prevNode !== null && !$isTextNode(prevNode))) {
```
arjita-mitra: do we need to check wordCharBeforeCursor? since we are already checking cursorAtStartOfNode.

src/script/components/LexicalInput/utils/mention-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+    nodes.push($createTextNode(' '));
+  }
+  nodes.push(mentionNode);
+  if (wordCharAfterCursor || (cursorAtEndOfNode && nextNode !== null && !$isTextNode(nextNode))) {
```
arjita-mitra: do we need to check wordCharAfterCursor? since we are already checking cursorAtEndOfNode.

src/script/components/LexicalInput/utils/mention-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+    if ($isTextNode(mentionNode)) {
+      mentionNode.select();
+    } else if (spaceNode) {
+      spaceNode.selectPrevious();
+    }
+  }
+
+  return true;
+}
+
```
arjita-mitra: getNextSibling and getPreviousSibling are almost identical. we avoid code duplication follow DRY principle. we can rewrite it as.
```

// Generic function to get the next or previous sibling based on the direction
function getSibling(node: LexicalNode, direction: 'forward' | 'backward'): LexicalNode | null {
  let sibling: LexicalNode | null = direction === 'forward' ? node.getNextSibling() : node.getPreviousSibling();
  while (sibling !== null && sibling.getType() === 'zeroWidth') {
    sibling = direction === 'forward' ? sibling.getNextSibling() : sibling.getPreviousSibling();
  }
  return sibling;
}

// The original getNextSibling and getPreviousSibling functions can now be replaced by getSibling

export function getNextSibling(node: LexicalNode) {
  return getSibling(node, 'forward');
}

export function getPreviousSibling(node: LexicalNode) {
  return getSibling(node, 'backward');
}
```

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+  useLayoutEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
+      // "Shift" + "Tab"
+      if (event.shiftKey && isTabKey(event)) {
+        editor.dispatchCommand(ON_SHIFT_TAB, event);
+      }
+    };
+
+    return editor.registerRootListener((rootElement: HTMLElement | null, prevRootElement: HTMLElement | null) => {
+      if (prevRootElement !== null) {
```
arjita-mitra: shouldn't we unregister root listener too?

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+        prevRootElement.removeEventListener('keydown', onKeyDown);
+      }
+      if (rootElement !== null) {
+        rootElement.addEventListener('keydown', onKeyDown);
+      }
+    });
+  }, [editor]);
+
+  useEffect(() => {
+    return mergeRegister(
```
arjita-mitra: shouldn't we unregister command too?

arjita-mitra (changes requested):

src/script/components/LexicalInput/components/Mention/Mention.tsx
arjita-mitra: I don't think so, it was a large block of if's without any description. its definitely not more readable.

src/script/components/LexicalInput/components/EditMessage.tsx
arjita-mitra: what is the error? it is allowed to return null from react components

src/script/components/LexicalInput/components/Mention/Mention.tsx
arjita-mitra: what comes from the plugin? this is our component, we should have basic function description to understand the code without having to check the plugin manual every time.

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
przemvs: It also cleaning up

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
przemvs: I made it like was previous :) To have first mention on the bottom

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
przemvs: This is an example :)

src/script/components/LexicalInput/utils/mention-utils.ts
przemvs: I'll stay with current sign. It's easier to read.

src/script/components/LexicalInput/LexicalInput.tsx
arjita-mitra: why are we creating function inside render, we should create the function outside render  to prevent unnecessary function re-creations on each render.

src/script/components/LexicalInput/LexicalInput.tsx
przemvs: image

src/script/components/LexicalInput/LexicalInput.tsx
przemvs: image

src/script/components/LexicalInput/hooks/useDebounce.ts
arjita-mitra: this is an existing file, for existing file we shouldn't change the copyright year.

src/script/components/LexicalInput/LexicalInput.tsx
przemvs: Yes, we should log this error in datadog. We should know if users have some issues with Lexical.

src/script/components/LexicalInput/components/Mention/Mention.tsx
arjita-mitra: this has not been addressed.

src/script/components/LexicalInput/LexicalInput.tsx
przemvs: There is a lot of changes, so You can miss this one-line temporary changes. I moved it to input-bar--wrapper class, to have it consistent.

src/script/components/LexicalInput/plugins/BeautifulMentionsPlugin.tsx
arjita-mitra: not addressed

src/script/components/LexicalInput/components/Mention/Mention.tsx
przemvs: I specified this type by adding: useRef<HTMLSpanElement>(...);

image

src/script/components/LexicalInput/components/Mention/Mention.tsx
przemvs: It's change to one function named **onArrowPress**. So now we can reuse it for keys: **ArrowLeft** and **ArrowRight** and other keys if we want.  So I'll add for You comment for what element we handling.

src/script/components/LexicalInput/utils/mention-utils.ts
arjita-mitra: still the same
```

 if (wordCharBeforeCursor || (cursorAtStartOfNode && prevNode !== null && !$isTextNode(prevNode))) {
    nodes.push($createTextNode(' '));
  }
```

przemvs: We need to check it, it checking different things.

src/script/components/LexicalInput/utils/mention-utils.ts
przemvs:
```
const cursorAtStartOfNode = offset === 0;
  const cursorAtEndOfNode = textContent.length === offset;
  const wordCharBeforeCursor = isWordChar(charBeforeCursor, triggers);
  const wordCharAfterCursor = isWordChar(charAfterCursor, triggers);
```

src/script/components/LexicalInput/components/Mention/Mention.tsx
```diff
@@ -0,0 +1,208 @@
...
+  );
+
+  const onArrowPress = useCallback(
+    (event: KeyboardEvent) => {
+      const node = $getNodeByKey(nodeKey);
+      if (!node || !node.isSelected()) {
+        return false;
+      }
+
+      const isLeftDirection = event.key === 'ArrowLeft';
```
arjita-mitra: use KEY.ARROW_LEFT from keyboardutils.ts

arjita-mitra (changes requested):

src/script/components/LexicalInput/components/EditMessage.tsx
przemvs: image

src/script/components/LexicalInput/plugins/EmojiPickerPlugin/index.tsx
przemvs: We can't move it outside function, because we need to check by queryString

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
przemvs: No, check documentation: https://lexical.dev/docs/concepts/commands#editorregistercommand

src/script/components/LexicalInput/utils/mention-utils.ts
przemvs: I'll keep it, because there a lot of logic. We need to be careful when we changing it.

src/script/components/LexicalInput/plugins/GlobalEventsPlugin.tsx
przemvs: We don't need. You can check documentation: https://lexical.dev/docs/concepts/commands#editorregistercommand

src/script/components/LexicalInput/LexicalInput.tsx
```diff
@@ -0,0 +1,182 @@
...
+                  data-uie-name="input-message"
+                />
+              }
+              placeholder={<Placeholder text={placeholder} hasLocalEphemeralTimer={hasLocalEphemeralTimer} />}
+              ErrorBoundary={LexicalErrorBoundary}
+            />
+
+            <BeautifulMentionsPlugin onSearch={queryMentions} />
+
+            <OnChangePlugin onChange={onChange} />
```
arjita-mitra: can you give the onChange function a meaningful name which describes the action we are doing onchange?

arjita-mitra (changes requested):

src/script/util/DraftStateUtil.ts
```diff
@@ -22,73 +22,52 @@ import {generateConversationInputStorageKey} from 'Util/util';
...
-export interface Draft {
-  mentions: MentionEntity[];
-  reply: Reply;
-  text: string;
+export interface DraftState {
+  editorState: string | null;
+  messageReply?: Promise<ContentMessage>;
 }
 
 export const saveDraftState = async (
```
atomrc, line 33: Can you confirm that when loading this code with old draft (created with the previous version of the input bar), the draft will load in the input?

src/script/components/InputBar/InputBar.test.tsx.tmp
```diff
@@ -185,20 +190,20 @@ describe('InputBar', () => {
...
+  //
+  //   const mentionCandidate = getMentionCandidate([], selectionStart, selectionEnd, inputValue);
+  //
+  //   const userEntity = new User(createUuid());
+  //   userEntity.name(userName);
+  //
+  //   const mentionEntity = createMentionEntity(userEntity, mentionCandidate);
+  //
+  //   expect(mentionEntity?.startIndex).toBe(3);
+  //   expect(mentionEntity?.length).toBe(7);
```
atomrc: This file needs some love it seems

src/script/components/InputBar/InputBar.tsx
```diff
@@ -517,16 +346,26 @@ const InputBar = ({
...
+        cancelMessageReply();
+      });
+    }
+  };
+
+  const onSend = (messageText: string, mentions: MentionEntity[]): void => {
     if (pastedFile) {
-      return sendPastedFile();
+      sendPastedFile();
+      return;
```
atomrc: how about

```suggestion
      return void sendPastedFile;
```

atomrc (changes requested):

atomrc (changes requested):
Ignore this one, it seems like it was never implemented correctly

~Seems like drafts are never deleted in the local storage (entries keeps being created as I switch conversation).~

~Let's clean them when the message is sent or the input put to empty string~

atomrc (changes requested):
regression:

- cannot put 2 emojis right after one another 

https://github.com/wireapp/wire-webapp/assets/1090716/58ef5252-ff36-4cf2-abb3-28ffe1ad342e

atomrc (changes requested):
bug when switching conversation that has never been visited before

image

przemvs:
> Ignore this one, it seems like it was never implemented correctly
> 
> ~Seems like drafts are never deleted in the local storage (entries keeps being created as I switch conversation).~
> 
> ~Let's clean them when the message is sent or the input put to empty string~

So maybe we can create follow up task for this one ? 🤔

src/script/components/Conversation/Conversation.tsx
```diff
@@ -515,6 +515,7 @@ export const Conversation: FC<ConversationProps> = ({
           />
 
           <InputBar
+            key={activeConversation?.id}
```
atomrc, line 524: This will reset the input bar everytime we switch conversation

atomrc:
This branch has been deployed and can be tested on https://wire-webapp-qa.zinfra.io/

src/script/components/LexicalInput/LexicalInput.tsx
atomrc: Fixed :)

src/script/util/DraftStateUtil.ts
atomrc: Will actually not work. Need to discuss with product if it's okay if old drafts are not restored

src/script/components/LexicalInput/plugins/ReverseTypeaheadMenuPlugin.tsx
```diff
@@ -0,0 +1,738 @@
...
+ *
+ */
+
+/**
+ * This is a modified version of https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx
+ * It allows showing a list that is filled from the bottom up instead of the top down (the default selected item is at the bottom of the list).
+ * The only differences lie in:
+ * - the default index we select when the component is instanciated (the last one by default)
+ * - a protection to avoid re-rendering the entire component when the list updates (which makes the scroll flicker)
+ */
```
atomrc: reviewing this file can be ignored as it's a duplicate of the native `TypeaheadMenu` from Lexical https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalTypeaheadMenuPlugin.tsx

Only it handles reversed list (with the selection at the bottom of the list instead of the top)

src/script/components/RichTextEditor/plugins/EmojiPickerPlugin/EmojiPickerPlugin.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+import {
+  LexicalTypeaheadMenuPlugin,
+  MenuOption,
+  MenuRenderFn,
+  MenuTextMatch,
+  useBasicTypeaheadTriggerMatch,
+} from '@lexical/react/LexicalTypeaheadMenuPlugin';
+import emojiList from 'emoji-picker-react/src/data/emojis.json';
```
atomrc, line 31: Add comment to explain where the emoji.json comes from

src/script/components/RichTextEditor/plugins/ReverseTypeaheadMenuPlugin.tsx
atomrc, line 1: Reviewers: please bear in mind that this plugin was copy/pasted from the lexical repo (and modified to work in reverse order)

src/script/components/AppContainer/hooks/useTheme.ts
```diff
@@ -39,16 +36,6 @@ function setTheme(theme: Theme) {
...
-  updateTheme();
-
-  useEffect(() => {
-    listenedEvents.forEach(event => amplify.subscribe(event, updateTheme));
-    return () => {
-      listenedEvents.forEach(event => amplify.unsubscribe(event, updateTheme));
-    };
-  }, [updateTheme]);
+  setTheme(getTheme());
+  useUserPropertyChange(getTheme, WebAppEvents.PROPERTIES.UPDATE.INTERFACE.THEME, setTheme);
```
atomrc, line 40: Since I needed the same logic for the "replaceEmoji" user preference, I extracted that logic to a more generic hook

src/script/components/InputBar/InputBar.tsx
```diff
@@ -146,83 +138,75 @@ const InputBar = ({
...
-  const richTextInput = getRichTextInput(currentMentions, inputValue);
+  const hasLocalEphemeralTimer = isSelfDeletingMessagesEnabled && !!localMessageTimer && !hasGlobalMessageTimer;
 
   // To be changed when design chooses a breakpoint, the conditional can be integrated to the ui-kit directly
   const isScaledDown = useMatchMedia('max-width: 768px');
 
-  const config = {
-    GIPHY_TEXT_LENGTH: 256,
-  };
+  const showGiphyButton = messageContent.text.length > 0 && messageContent.text.length <= config.GIPHY_TEXT_LENGTH;
```
PatrykBuniX: question: can we use `textValue` variable here instead of `messageContent.text`?. It's declared on line `153`.

atomrc: Yes, absolutely 👌 https://github.com/wireapp/wire-webapp/pull/15507/commits/566c8d18dbfc14e800b94ad50719d061d794fd18

src/script/components/InputBar/InputBar.tsx
```diff
@@ -146,83 +138,75 @@ const InputBar = ({
...
-    updateSelectionState(updateSelection);
-    setTimeout(() => {
-      textareaRef.current?.setSelectionRange(endPosition, endPosition);
-      textareaRef.current?.focus();
-    }, 0);
-  };
+  useTypingIndicator({
+    isEnabled: isTypingIndicatorEnabled,
+    text: messageContent.text,
+    onTypingChange: useCallback(
```
PatrykBuniX, line 193: question: Does this `useCallback` optimises anything here? We don't store any reference to it, feels to me that it's recreated everytime the hook is called, but i could be wrong.

atomrc: Screenshot 2023-09-05 at 10 00 13

It's actually used in an effect in the hook. So we want to be careful about re-creating this function at each render (it would cause the effect to trigger on each render)

PatrykBuniX: Yeah, `useCallback` make sense in general, but what I mean is: shouldn't we store a function that is returned by `useCallback` in some variable in component's body? Not sure if the definition is still cached if we use it this way 🤔

atomrc: Oh, it is. That the purpose of `useCallback` it's going to return the same reference if the deps array doesn't change. 
It doesn't matter if you store it in a variable or not, it's the `useCallback` function that works like this

PatrykBuniX: Okay makes sense, all good then 😄

src/script/components/RichTextEditor/plugins/HistoryPlugin.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {registerHistory, createEmptyHistoryState} from '@lexical/history';
+import {useLexicalComposerContext} from '@lexical/react/LexicalComposerContext';
+
+/**
+ * Will automatically focus the input field when the user types anywhere in the document
+ */
+export function HistoryPlugin(): null {
+  const [editor] = useLexicalComposerContext();
+
+  useEffect(() => registerHistory(editor, createEmptyHistoryState(), 300), [editor]);
```
przemvs: Please move 300 to const, now we don't know what it mean :(

package.json
```diff
@@ -3,6 +3,8 @@
     "@datadog/browser-logs": "^4.50.0",
     "@datadog/browser-rum": "^4.50.0",
     "@emotion/react": "11.11.1",
+    "@lexical/history": "^0.11.3",
+    "@lexical/react": "^0.11.1",
```
PatrykBuniX: Don't we want to use exact versions here as well?

atomrc: good call 👌

atomrc (approved):

PatrykBuniX (approved):

PR #15508, opened 2023-07-27
feat: MLS 1:1 conversations [WPB-2183]

PatrykBuniX:
<!--do not remove this marker, its needed to replace info when ticket title is updated -->
<!--jira-description-action-hidden-marker-start-->

<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2183" title="WPB-2183" target="_blank"><img alt="Story" src="https://wearezeta.atlassian.net/rest/api/2/universal_avatar/view/type/issuetype/avatar/10815?size=medium" />WPB-2183</a>  [Web] [Clients] Select protocol for 1-1 conversation
  </td></table>
  <br />
 

<!--jira-description-action-hidden-marker-end-->
<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

<!--do not remove this marker, its needed to replace info when ticket title is updated -->

When MLS is supported and both of the users in a connection support MLS protocol (`supported_protocols` field on user entity includes `"mls"`) we want to:
- establish 1:1 MLS conversation
- add the other user into the MLS group
- move all the events (messages) into the new conversation
- hide old proteus conversation


src/script/components/Conversation/ReadOnlyConversationMessage.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <Icon.Info />
+        </div>
+      </div>
+      <p className="readonly-message-header-label" data-uie-name="element-readonly-conversation">
+        <span
+          dangerouslySetInnerHTML={{
+            __html: mlsCompatibilityMessage,
+          }}
+        />
+        {state === CONVERSATION_READONLY_STATE.READONLY_ONE_TO_ONE_SELF_UNSUPPORTED_MLS && (
```
przemvs, line 56: Move this to variable, You using it multiple times :)

src/script/components/Conversation/ReadOnlyConversationMessage.tsx
```diff
@@ -0,0 +1,77 @@
...
+  handleMLSUpdate: () => void;
+  displayName: string;
+}
+export const ReadOnlyConversationMessage: FC<ReadOnlyConversationMessageProps> = ({
+  state,
+  handleMLSUpdate,
+  displayName,
+}) => {
+  const mlsCompatibilityMessage =
+    state === CONVERSATION_READONLY_STATE.READONLY_ONE_TO_ONE_OTHER_UNSUPPORTED_MLS
```
przemvs, line 39: Move this to variable, You using it multiple times :)

src/script/components/Conversation/ReadOnlyConversationMessage.tsx
PatrykBuniX: Those two checks you've commented on are actually different. Once is `READONLY_ONE_TO_ONE_SELF_UNSUPPORTED_MLS`, the other one is `READONLY_ONE_TO_ONE_OTHER_UNSUPPORTED_MLS` (`SELF` vs `OTHER`)

src/script/components/Conversation/ReadOnlyConversationMessage.tsx
PatrykBuniX: https://github.com/wireapp/wire-webapp/pull/15508#discussion_r1350414740

przemvs (approved):

src/i18n/en-US.json
```diff
@@ -727,9 +727,14 @@
   "messageFailedToSendWillNotReceiveSingular": "won't get your message.",
   "messageFailedToSendWillReceivePlural": "will get your message later.",
   "messageFailedToSendWillReceiveSingular": "will get your message later.",
-  "mlsConversationRecovered": "You haven’t used this device for a while, or an issue has occurred. Some older messages may not appear here.",
+  "messageReactionDetails": "{{emojiCount}} reaction, react with {{emojiName}} emoji",
```
atomrc: hummm, this seems unrelated to the topic 🤔

src/script/conversation/ConversationMapper.ts
```diff
@@ -178,6 +179,10 @@ export class ConversationMapper {
       conversationEntity.verification_state(verification_state);
     }
 
+    if (archived_state !== undefined) {
+      conversationEntity.archivedState(archived_state);
+    }
+
```
atomrc: This seems somewhat unrelated 🤔

src/script/conversation/ConversationService.ts
```diff
@@ -430,4 +430,17 @@ export class ConversationService {
...
+   * Will check if mls group is established locally.
+   * Group is established after the first commit was sent in the group and epoch number is at least 1.
+   * @param groupId groupId of the conversation
+   */
+  async isMLSGroupEstablishedLocally(groupId: string): Promise<boolean> {
+    return this.coreConversationService.isMLSGroupEstablishedLocally(groupId);
+  }
+
+  async getMLS1to1Conversation(userId: QualifiedId) {
+    return this.apiClient.api.conversation.getMLS1to1Conversation(userId);
```
atomrc: don't we have a `core` method that would do that? As much as possible we should avoid using the `apiClient` directly in the webapp (in the long run the apiClient should be hidden to the webapp)

src/script/page/AppMain.tsx
```diff
@@ -225,6 +225,7 @@ const AppMain: FC<AppMainProps> = ({
                 selfUser={selfUser}
                 isRightSidebarOpen={!!currentState}
                 openRightSidebar={toggleRightSidebar}
+                onRefresh={app.refresh}
```
atomrc: suggestion: maybe the naming is not really explicit. I feel that the app will completely reload when something (I don't know what) is refreshed. 

maybe we could try to be more explicit `onUserSupportedProtocolUpdated` (I don't really know what is the trigger here, you would probably find a better name :D )

src/script/main/app.ts
```diff
@@ -413,10 +414,6 @@ export class App {
         conversationRepository.initMLSConversationRecoveredListener();
       }
 
-      if (connections.length) {
-        await Promise.allSettled(conversationRepository.mapConnections(connections));
-      }
-
```
atomrc, line 419: I guess this is done by the `init1To1Conversations` right? Maybe it would make sense to put it in another function or is this to tied to actual 1:1 conversations?

src/i18n/en-US.json
PatrykBuniX: No idea what it's doing here, good catch, will remove it.

src/script/conversation/ConversationService.ts
PatrykBuniX: We do not have it yet but i will add it 👌. I like the approach you're suggesting 🙏

src/script/page/AppMain.tsx
PatrykBuniX: This one should not start with `"on"` actually, I suggest naming it `reloadApp`, it's called when user clicks an update button in a warning message we display when they do not support mls and the other user (the one we try to communicate with via 1:1 conversation) does support only mls (does not support proteus). After app is reloaded, we will re-evaluate 1:1 conversations again. There's a chance (but not guaranty), that after the reload self user's team has started supporting mls, so conversation can be finally established. 

So it's not a listener (`"onSomething"`), but rather an action (`"doSomething"`). Makes sense to you?

src/script/main/app.ts
PatrykBuniX: Correct, this logic was moved to `init1To1Conversations`, I don't quite get what you mean, you've commented on removed code 🤔

atomrc: My question is: would it make sense to have a separate function that does the connection mapping (rather than doing it inside of the `init1to1conversations` method). In order to split reponsibilities. But maybe it doesn't make sense, I leave you judge of this :D

src/script/page/AppMain.tsx
atomrc: Sure, maybe this makes more sense 👍

src/script/main/app.ts
PatrykBuniX: Yeah, I'd leave it as it is now - we're initialising 1:1 conversation every time we map through connection (and find a 1:1 conversation for this connection), so i'd say it's tied to it. Right after that we need to map through team-owned 1:1 conversations (that don't have connections assigned) anyway, I prefer to have initApp method calling a single method that takes care of initialising all the 1:1 conversations.

atomrc (approved):

PR #15530, opened 2023-07-31
feat: emoji pills should be sorted by number of reactions in desc order

arjita-mitra:
----

- The **PR Title**
  - sort reactions by user count
- The **PR Description**
  - emoji pills should be sorted by number of reactions in desc order.
----

### DIFF
--- a/src/script/util/ReactionUtil.ts
+++ b/src/script/util/ReactionUtil.ts
@@ -44,6 +44,13 @@ export function getEmojiUrl(unicode: string) {
   return `/image/emojis/img-apple-64/${unicode}.png`;
 }
 
+/**
+ *
+ * @param reactionsList This is an array of tuples, each tuple consists of two elements a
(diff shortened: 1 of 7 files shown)


src/script/util/ReactionUtil.ts
```diff
@@ -43,3 +43,9 @@ export function groupByReactionUsers(reactions: Reactions): ReactionsGroupedByUs
 export function getEmojiUrl(unicode: string) {
   return `/image/emojis/img-apple-64/${unicode}.png`;
 }
+
+export function sortReactionsByUserCount(reactionsList: [string, string[]][]) {
+  return reactionsList.sort(
+    ([, reactionAUserList], [, reactionBUserList]) => reactionBUserList.length - reactionAUserList.length,
```
tlebon, line 56: maybe some comments are good to explain that the reaction list is an array and the first value is the name of the reaction and not used for sorting? Could be helpful for people in the future possibly.

tlebon (approved):

PR #15555, opened 2023-08-03
feat: handle connection request with no longer federating users (WPB-2221)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-2221" title="WPB-2221" target="_blank">TaskWPB-2221</a>  [Web] Handle connection requests to users on backends with which we don’t federate
  </td></table>
  <br />
 

### Feature

- display a modal when trying to connect to a user whom back-end we're no longer federated with
Screenshot from 2023-08-03 18-02-02

*note: I used the latest suggestion from the localization team that is not reflected on the design yet*

### DIFF
--- a/src/script/connection/ConnectionRepository.ts
+++ b/src/script/connection/ConnectionRepository.ts
@@ -191,29 +191,37 @@ export class ConnectionRepository {
       return true;
     } catch (error) {
       if (isBackendError(error)) {
-        if (error.label === BackendErrorLabel.LEGAL_HOLD_MISSING_CONSENT) {
-          const replaceLinkLegalHold = replaceLink(
-            Config.getConfig().URL.SUPPORT.LEGAL_HOLD_BLOCK,
-            '',
-            'read-more-legal-hold',
-          );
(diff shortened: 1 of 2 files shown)


src/script/connection/ConnectionRepository.ts
```diff
@@ -189,21 +190,32 @@ export class ConnectionRepository {
...
+        }
+        if (error.label === BackendErrorLabel.FEDERATION_NOT_ALLOWED) {
+          PrimaryModal.show(PrimaryModal.type.ACKNOWLEDGE, {
+            text: {
+              htmlMessage: t('modalUserCannotConnectNotFederatingMessage', userEntity.name()),
+              title: t('modalUserCannotConnectNotFederatingHeadline'),
+            },
+          });
+        }
+        this.logger.error(`Failed to send connection request to user '${userEntity.id}': ${error.message}`, error);
```
tlebon: idea: this could be a switch statement instead

tlebon (approved):
update to a switch if you like but that shouldnt block this merging before a playtest

tlebon (approved):
Looks good

przemvs (approved):

PR #15571, opened 2023-08-07
refactor: remove pending welcome mls state

PatrykBuniX:
Removes pending welcome state from mls conversations state since it's not needed anymore, see: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/563053166/Use+case+being+added+to+a+conversation+MLS

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -39,6 +39,7 @@ import {
   ConversationRenameEvent,
   ConversationTypingEvent,
   CONVERSATION_EVENT,
+  ConversationMLSWelcomeEvent,
 } from '@wireapp/api-client/lib/event';
 import {BackendErrorLabel} from '@wireapp/api-client/lib/http/';
 import type {BackendError} from '@wireapp/api-client/lib/http/';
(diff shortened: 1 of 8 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2068,21 +2069,13 @@ export class ConversationRepository {
...
+
     const inSelfConversation = this.conversationState.isSelfConversation(conversationId);
     if (inSelfConversation) {
-      const typesInSelfConversation = [
-        CONVERSATION_EVENT.MEMBER_UPDATE,
-        ClientEvent.CONVERSATION.MESSAGE_HIDDEN,
-        /**
-         * As of today (07/07/2022) the backend sends `WELCOME` message to the user's own
-         * conversation (not the actual conversation that the welcome should be part of)
-         */
```
atomrc, line 2079: is this not true anymore? Is it sent in the real conversation now?

PatrykBuniX: Yes, the `conversation` field on welcome message event now contains the id of the actual conversation and not the id of the self conversation as it was before.

Screenshot 2023-08-07 at 21 20 31

atomrc: great news!!! 👐

atomrc (approved):

src/script/conversation/ConversationRepository.ts
PatrykBuniX: @atomrc and then the next step is to fetch conversation if it was not known locally as usual, see https://github.com/wireapp/wire-webapp/blob/dev/src/script/conversation/ConversationRepository.ts#L2101. It was not possible previously so we had to wait for `conversation.member.join` event in case welcome arrived first. Right now it doesn't really matter which one comes first.

PR #15576, opened 2023-08-08
runfix: archived or abandoned conversation not visible

PatrykBuniX:
Filter out only those MLS conversations that are not established and self user was not removed from the conversation. Conversation should still be visible after leaving a conversation without clearing the content - even if mls group is not established anymore.

### DIFF
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -78,10 +78,7 @@ export class ConversationState {
     });
 
     this.visibleConversations = ko.pureComputed(() => {
-      const filteredMLSConversations = useMLSConversationState
-        .getState()
-        .filterEstablishedConversations(this.sortedConversations());
-      return filteredMLSConversations.filter(
+      return this.sortedConversations().filter(
         conversation =>
           !conversation.is_cleared() &&
           !conversation.is_archived() &&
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationState.ts
```diff
@@ -111,14 +103,28 @@ export class ConversationState {
...
-        return !(conversationEntity.is_cleared() && conversationEntity.removed_from_conversation());
+        if (isMLSConversation(conversationEntity)) {
+          const isMLSConversationEstablished = useMLSConversationState
+            .getState()
+            .isEstablished(conversationEntity.groupId);
+
+          //conversation doesn't need to be established if user was removed
+          if (!isRemoved && !isMLSConversationEstablished) {
+            return false;
+          }
```
PatrykBuniX, line 124: This is the main change, the rest is just refactoring of where things are computed.

src/script/mls/mlsConversationState/mlsConversationState.ts
```diff
@@ -52,8 +51,6 @@ type StoreState = MLSConversationState & {
 export const useMLSConversationState = create<StoreState>((set, get) => {
   return {
     established: initialState.established,
-    filterEstablishedConversations: conversations =>
-      conversations.filter(conversation => !conversation.groupId || get().isEstablished(conversation.groupId)),
```
PatrykBuniX, line 56: Not needed anymore since we use `.isEstablished()` directly.

atomrc (approved):

PR #15580, opened 2023-08-10
fix: unarchived conversation is missing after reload [WPB-3753]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3753" title="WPB-3753" target="_blank">BugWPB-3753</a>  [Web] Conversation content is gone after leaving a conversation and unarchiving
  </td></table>
  <br />
 

Conversation was missing in conversations list when user has left it without clearing the content, unarchived automatically archived conversation and reloaded the webapp. We should not hide conversations we've left (or were removed from) that are still known by the local db. This PR should fix it.

### DIFF
--- a/src/script/conversation/ConversationMapper.ts
+++ b/src/script/conversation/ConversationMapper.ts
@@ -311,27 +311,27 @@ export class ConversationMapper {
 
     const conversationsMap = new Map<string, ConversationDatabaseData>();
 
-    for (const conversation of localConversations) {
-      const conversationId = conversation.qualified_id?.id || conversation.id;
-      conversationsMap.set(conversationId, conversation);
+    for (const localConversation of localConversations) {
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationMapper.ts
```diff
@@ -296,40 +293,48 @@ export class ConversationMapper {
...
+
+      const localConversationData = (remoteConversation.qualified_id || {
+        id: conversationId,
+        domain: '',
+      }) as ConversationDatabaseData;
+
+      conversationsMap.set(conversationId, this.mergeSingleConversation(localConversationData, remoteConversation, i));
+    }
+
+    return Array.from(conversationsMap.values());
```
atomrc, line 337: Nice usage of `Map` 😎

atomrc (approved):

PR #15583, opened 2023-08-11
feat: Add stopping to federate (WPB-203)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-203" title="WPB-203" target="_blank">TaskWPB-203</a>  [Web] System message when stopping to federate
  </td></table>
  <br />
 

This pull request implements a set of client-side actions to handle events related to federated communication termination. The specification outlines the steps for two types of events: `federation.delete` and `federation.connectionRemoved`. The main focus is on cleaning up local state after receiving these events and inserting a system message to explain what happened to the user.

For the `federation.delete` event: (Backend A has stopped federating with us)

- receive the event from backend
- leave the conversations locally that are owned by the backend A which was deleted.
- remove the deleted backend A users locally from our own conversations.
- insert system message to the affected conversations about federation termination.

For the `federation.connectionRemoved` event: (Backend A & B stopped federating, user is on C)

- receive the event from backend
- Identify all conversations that are not owned from A or B domain and that contain users from A and B
  - remove users from A and B from those conversations
  - insert system message in those conversations about backend A and B stopping to federate
- identify all conversations owned by domain A that contains users from B 
  - remove users from B from those conversations 
  - insert system message in those conversations about backend A and B stopping to federate
- Identify all conversations owned by domain B that contains users from A 
  - remove users from A from those conversations
  - insert system message in those conversations about backend A and B stopping to federate


src/script/conversation/ConversationRepository.ts
```diff
@@ -310,13 +312,130 @@ export class ConversationRepository {
...
+    if (type === FEDERATION_EVENT.FEDERATION_DELETE) {
+      const {domain: deletedDomain} = data;
+
+      await this.onFederationDelete(deletedDomain);
+    }
+
+    if (type === FEDERATION_EVENT.FEDERATION_CONNECTION_REMOVED) {
+      const {domains: deletedDomains} = data;
+      await this.onFederationConnectionRemove(deletedDomains);
+    }
```
atomrc: let's convert that into a switch in case we will need to handle more events of this type in the future. 
Also we could early return in case the type has been matched

src/script/conversation/EventMapper.ts
```diff
@@ -295,6 +297,11 @@ export class EventMapper {
         break;
       }
 
+      case ClientEvent.CONVERSATION.FEDERATION_STOP: {
+        messageEntity = this._mapEventFederationStop(event);
+        break;
+      }
+
```
atomrc, line 304: don't we need to handle the connectionRemove event (and display it) as well?

src/script/event/EventRepository.ts
```diff
@@ -324,6 +324,9 @@ export class EventRepository {
       case EVENT_TYPE.CALL:
         amplify.publish(WebAppEvents.CALL.EVENT_FROM_BACKEND, event, source);
         break;
+      case EVENT_TYPE.FEDERATION:
+        amplify.publish(WebAppEvents.FEDERATION.EVENT_FROM_BACKEND, event, source);
+        break;
       case EVENT_TYPE.CONVERSATION:
         amplify.publish(WebAppEvents.CONVERSATION.EVENT_FROM_BACKEND, event, source);
         break;
```
atomrc, line 332: note for the future, I'd like the EventRepo to not really filter the type of events. 
It should just send all the events in a single event and all the repos subscribing to it will make the work of filtering depending on the event type. :)

src/script/conversation/EventMapper.ts
thisisamir98: this is the event that we are creating on the client side for the system message, connection removed is handled here:
https://github.com/wireapp/wire-webapp/pull/15583/files#diff-8424911ba55dc95ddd7ee35e77e651a8894faa0c33cd485cbbf62f6537fe3d24R327-R329

src/script/conversation/ConversationRepository.ts
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/15583/commits/76d1df00d310492f36e57e1f9338a2982bc423b7

src/script/conversation/ConversationRepository.ts
```diff
@@ -310,13 +312,154 @@ export class ConversationRepository {
...
+      - insert system message in those conversations about backend A and B stopping to federate
+   * @param domains The domains that stopped federating with each other
+   */
+  private readonly onFederationConnectionRemove = async (domains: string[]) => {
+    const selfUser = this.userState.self();
+    const [domainOne, domainTwo] = domains;
+    const allConversations = this.conversationState.conversations();
+
+    allConversations
+      .filter(conversation => conversation.domain === domainOne)
```
arjita-mitra, line 401: these two blocks are almost identical except domain type. We can create a function and call it twice with each domain.

```
const handleFederationUsers = async (domainToFilter: string, userDomainToDelete: string) => {
  allConversations
    .filter(conversation => conversation.domain === domainToFilter)
    .forEach(async conversation => {
      const usersToDelete = conversation.allUserEntities().filter(user => user.domain === userDomainToDelete);
      if (usersToDelete.length > 0) {
        await this.removeDeletedFederationUsers(conversation, usersToDelete);
        await this.insertFederationStopSystemMessage(conversation, [domainOne, domainTwo]);
      }
    });
};

await handleFederationUsers(domainOne, domainTwo);
await handleFederationUsers(domainTwo, domainOne);
```

src/script/conversation/ConversationRepository.ts
```diff
@@ -310,13 +312,154 @@ export class ConversationRepository {
...
+        const hasAnyUserFromDomainTwo = conversation
+          .allUserEntities()
+          .find(user => user.qualifiedId.domain === domainTwo);
+
+        return hasAnyUserFromDomainOne && hasAnyUserFromDomainTwo;
+      })
+      .forEach(async conversation => {
+        const usersToDelete = conversation
+          .allUserEntities()
+          .filter(user => user.domain === domainOne || user.domain === domainTwo);
```
arjita-mitra: Array.includes() here makes the intent clearer

.filter(user => [domainOne, domainTwo].includes(user.domain));

src/script/conversation/ConversationRepository.ts
```diff
@@ -310,13 +312,154 @@ export class ConversationRepository {
...
+
+    allConversations
+      .filter(conversation => {
+        if (conversation.domain !== selfUser.qualifiedId.domain) {
+          return false;
+        }
+
+        const hasAnyUserFromDomainOne = conversation
+          .allUserEntities()
+          .find(user => user.qualifiedId.domain === domainOne);
```
arjita-mitra: Instead of using .find() twice in the filter, we can convert the list of user domains in a conversation into a Set for faster lookups, and then check for the existence of domainOne and domainTwo in the Set.



```
const userDomains = new Set(
      conversation.allUserEntities().map(user => user.qualifiedId.domain)
    );
```

    return userDomains.has(domainOne) && userDomains.has(domainTwo);

arjita-mitra (dismissed):

src/script/conversation/ConversationRepository.ts
atomrc: that's is a good suggestion, indeed. But without unit tests to back it, it's also risky. 
I'd recommend writing tests first (I can help with that), make sure that we are compliant with the specs and then refactoring to a cleaner/more performant code.

src/script/conversation/EventMapper.ts
atomrc: but there is no system message for connectionRemoved event?

src/script/conversation/ConversationRepository.ts
thisisamir98: nice https://github.com/wireapp/wire-webapp/pull/15583/commits/fc35485500af1f1fc508cf1a7a29b10e46460b8a

src/script/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15583/commits/fc35485500af1f1fc508cf1a7a29b10e46460b8a

atomrc (approved):

przemvs (approved):

PR #15600, opened 2023-08-14
runfix: do not maintain mls conversation state

PatrykBuniX:
Removes MLS conversations state. We were storing information about what mls conversations are being established, this could have been useful in the past when welcome message was sent in the user's self conversation instead of actual conversation user was invited to. Now when welcome message is received, it contains `conversationId`, so we are sure conversation exists at this point.

I don't see any value in maintaining the state of whether the conversation is established or not, this should be all inside corecrypto. It was only used to hide a conversation if it was not established yet.

I think that even showing a unestablished conversation entity is much more valuable, since we could perform some lazy sync checks whether group is established / epoch is in sync etc, when e.g. user is navigating to unestablished conversation, or is trying to send a message, but corecrypto fails with `Conversation doesn't exist` error message. Just hiding such a conversation will not allow us to perform any recovery / rejoin mechanisms.


atomrc (approved):
damn, I like those PR removing code 🫶

PR #15608, opened 2023-08-16
feat: handle non-federating users errors (WPB-3461)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3461" title="WPB-3461" target="_blank">StoryWPB-3461</a>  [Web] Implement client protocol for errors when adding users from non-fully connected graph
  </td></table>
  <br />
 

- Handle the new federated error for conversation creation
image

- Modify the FailedToAddUsersMessageEvent to accomodate failure to add users from backends that do not federate with each other

image


src/script/conversation/ConversationRepository.ts
```diff
@@ -593,29 +593,27 @@ export class ConversationRepository {
...
-          break;
-        case BackendErrorLabel.NOT_CONNECTED:
-          await this.handleUsersNotConnected(userEntities.map(user => user.qualifiedId));
-          break;
-        case BackendErrorLabel.LEGAL_HOLD_MISSING_CONSENT:
-          this.showLegalHoldConsentError();
-          break;
-        default:
-          this.logger.error(error);
+      if (isBackendError(error)) {
```
thisisamir98, line 603: I actually like this early return patter:
      if (!isBackendError(error)) {
        this.logger.error(error);
        throw error;
        }
        switch ...

atomrc: let's not `logger.error` since we throw, the caller will most likely print the error anyways

src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
```diff
@@ -148,4 +156,35 @@ describe('FailedToAddUsersMessage', () => {
     expect(elementMessageFailedToAddDetails[0].getAttribute('data-uie-value')).toEqual(qualifiedId1.domain);
     expect(elementMessageFailedToAddDetails[1].getAttribute('data-uie-value')).toEqual(qualifiedId2.domain);
   });
+  it('shows details of failed to add users from non federating backends', async () => {
```
atomrc, line 160:
```suggestion

  it('shows details of failed to add users from non federating backends', async () => {
```

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -142,29 +172,20 @@ const FailedToAddUsersMessage: React.FC<FailedToAddUsersMessageProps> = ({
...
-                          : domainUsers.map(user => user.name()).join(', '),
-                      domain,
-                    }),
-                  }}
-                />
+            {message.reason === AddUsersFailureReasons.UNREACHABLE_BACKENDS && (
+              <>
+                {Object.entries(groupedUsers).map(([domain, domainUsers]) => (
+                  <MessageDetails key={domain} domain={domain} message={message} users={domainUsers}>
+                    {learnMore}
```
atomrc, line 179: I guess we should only show the `learnMore` once for all the backends (here it's going to show for every single backend line, I feel, right?)

src/script/conversation/ConversationRepository.ts
atomrc: The problem with early return in error handling is that it might filter too much and makes it a little harder to add new error types to be handled here. 

Another solution, that is easily extensible would be using a switch/case statement

```js
switch (true) {
  case isBackendError(error):
    // handler here

  default
   throw error;
}
```

atomrc (approved):
LGTM

PR #15609, opened 2023-08-16
feat: Handle adding users from unreachable backends

atomrc:
This will allow the UI to show error messages when federated users could not be added to a conversation. 

## case 1: when only some users could not be added

https://github.com/wireapp/wire-webapp/assets/1090716/58afddbe-6251-4519-84e3-4e1971078214

## case 2: when no user could be added

https://github.com/wireapp/wire-webapp/assets/1090716/f5db5c44-f91f-4149-bb34-062c336912f0

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "@datadog/browser-rum": "^4.46.0",
     "@emotion/react": "11.11.1",
     "@wireapp/avs": "9.3.7",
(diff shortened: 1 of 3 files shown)


V-Gira (approved):

V-Gira (approved):
Better :)

PR #15629, opened 2023-08-21
fix: Viewing an archived conversation should not automatically “unarchive” it

arjita-mitra:
----

- The **PR Title**
  - Viewing an archived conversation should not automatically “unarchive” it

- The **PR Description**
  - Viewing an archived conversation should not automatically “unarchive” it
  - Viewing an archived conversation should stay at archived view
  - refactor showConversation method
----
##### References
1.https://wearezeta.atlassian.net/browse/WPB-3771


src/script/notification/NotificationRepository.ts
```diff
@@ -672,6 +672,10 @@ export class NotificationRepository {
       };
     }
 
+    if (conversationEntity?.archivedState()) {
+      return () =>
+        amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity || conversationId, {archive: true});
+    }
     return () => amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity || conversationId, {});
```
atomrc: Let's simplify the boolean logic here

```suggestion
    return () => amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity || conversationId, {archive: conversationEntity?.archivedState()});
```

src/script/conversation/ConversationRepository.ts
```diff
@@ -2548,11 +2534,7 @@ export class ConversationRepository {
       if (previouslyArchived) {
         // Add to check for un-archiving at the end of stream handling
         if (eventFromStream) {
-          return this.conversationsWithNewEvents.set(conversationEntity.id, conversationEntity);
-        }
-
-        if (eventFromWebSocket && conversationEntity.shouldUnarchive()) {
-          return this.unarchiveConversation(conversationEntity, false, 'event from WebSocket');
+          this.conversationsWithNewEvents.set(conversationEntity.id, conversationEntity);
```
atomrc: Do you know what is this `conversationWithNewEvents` map for?

src/script/conversation/ConversationRepository.ts
```diff
@@ -2045,12 +2045,6 @@ export class ConversationRepository {
   }
 
   private checkChangedConversations() {
-    this.conversationsWithNewEvents.forEach(conversationEntity => {
-      if (conversationEntity.shouldUnarchive()) {
-        this.unarchiveConversation(conversationEntity, false, ConversationRepository.eventFromStreamMessage);
-      }
-    });
-
     this.conversationsWithNewEvents.clear();
```
PatrykBuniX: Since this is not the only method we're calling in this method, we can just remove it and call `.clear()` directly in places where it was called (seems to be called only in 1 place)

src/script/conversation/ConversationRepository.ts
```diff
@@ -2247,19 +2241,11 @@ export class ConversationRepository {
     const onEventPromise = isConversationCreate
       ? Promise.resolve(null)
       : this.getConversationById(conversationId, true);
-    let previouslyArchived = false;
+    const previouslyArchived = false;
```
PatrykBuniX: Since it's hardcoded to false now, we can get rid of this variable and remove a param from `handleConversationNotification` method as well

src/script/view_model/ContentViewModel.ts
```diff
@@ -158,84 +261,35 @@ export class ContentViewModel {
...
+      const conversationEntity = await this.getConversationEntity(conversation, domain);
 
       if (!conversationEntity) {
-        rightSidebar.close();
-
-        throw new ConversationError(
-          ConversationError.TYPE.CONVERSATION_NOT_FOUND,
-          ConversationError.MESSAGE.CONVERSATION_NOT_FOUND,
-        );
+        return this.closeRightSidebarAndThrowError();
```
atomrc: Not sure I understand why extracting this code to a function if it's not used anywhere else. 
The previous version was more explicit I feel

src/script/view_model/ContentViewModel.ts
```diff
@@ -133,10 +134,112 @@ export class ContentViewModel {
...
+    conversationEntity: Conversation,
+    openFirstSelfMention: boolean,
+    exposeMessageEntity: Message | undefined,
+  ): void {
+    const messageEntity = openFirstSelfMention ? conversationEntity.getFirstUnreadSelfMention() : exposeMessageEntity;
+    if (conversationEntity.is_cleared()) {
+      conversationEntity.cleared_timestamp(0);
+    }
+    this.changeConversation(conversationEntity, messageEntity);
+  }
```
atomrc: Doesn't really seem to be processing message, is it? I don't think I understand the name of the function 🤔

src/script/view_model/ContentViewModel.ts
```diff
@@ -158,84 +261,35 @@ export class ContentViewModel {
...
-              title: t('conversationNotFoundTitle', Config.getConfig().BRAND_NAME),
-            },
-          },
-          undefined,
-        );
+      this.processMessageEntity(conversationEntity, openFirstSelfMention, exposeMessageEntity);
+      this.showAndNavigate(conversationEntity, openNotificationSettings);
+    } catch (error: any) {
+      if (this.isConversationNotFoundError(error)) {
+        this.showErrorModal();
```
atomrc: The name is a little too generic, It's not showing any error modal, it's showing the conversation not found error modal. 
I think it was okay to inline that code (doesn't really need to be extracted to a function)

src/script/view_model/ContentViewModel.ts
arjita-mitra: It's determining the message entity to work with based on the 2 criteria. So I named it processMessageEntity. But you are right its not processing any message so a better name would be updateConversationStatusBasedOnMessage or handleConversationWithMessageEntity. what do you think?

src/script/conversation/ConversationRepository.ts
arjita-mitra: that is not totally clear to me. I checked all the occurrence of it, its used only in this file. we are creating a map and setting it here with conversationEntity.

atomrc: Please check but I think you can remove it now (since it's not used to unarchive convo)

src/script/view_model/ContentViewModel.ts
arjita-mitra: I am following the principle of one function one action. Since I was breaking down all the actions into their own functions irrespective of its re-use so i created one for this too. If reading a function name describes its action then we don't even have to see the description of it, I found this more readable but i am fine with bringing it back.

src/script/view_model/ContentViewModel.ts
arjita-mitra: I think its good to extract it, makes the catch block much more compact and readable. And we can reuse this conversation error modal in future. I can rename this to showConversationErrorModal. what do you think?

src/script/page/LeftSidebar/panels/Archive.tsx
```diff
@@ -56,9 +56,7 @@ const Archive = ({
   ]);
 
   const onClickConversation = async (conversation: Conversation) => {
-    await conversationRepository.unarchiveConversation(conversation, true, 'opened conversation from archive');
-    onClose();
-    amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversation, {});
+    amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversation, {archive: true});
```
PatrykBuniX: I don't think we need to pass `{archive: true}` here, we're reading `.archivedState()` on conversation entity inside `showConversation` in `ContentViewModel` method, no? Seems redundant to me.

src/script/notification/NotificationRepository.ts
```diff
@@ -672,7 +672,7 @@ export class NotificationRepository {
       };
     }
 
-    return () => amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity || conversationId, {});
+    return () => amplify.publish(WebAppEvents.CONVERSATION.SHOW, conversationEntity || conversationId, {archive: conversationEntity?.archivedState()});
```
PatrykBuniX: We can read `archivedState` in `CONVERSATION.SHOW` event listener (`showConversation` method) directly, no need to pass it here.

src/script/view_model/ContentViewModel.ts
atomrc: Then let's make it very explicit what it's doing and maybe name it `showConversationNotFoundErrorModal` in this case.

src/script/view_model/ContentViewModel.ts
atomrc: But here it does 2 actions, (closing **and** throwing). The rule of thumb is that naming a function with **and** is usually a sign that the single responsibility principle is broken. 

There are 2 types of function: 
- glue functions (that are called from the UI and that could perform a lot of actions in a single call)
- utility functions (that are focused on doing a single task). 

In this case `showConversation` is a glue function (that will perform a lot of tasks to get the conversation ready to be shown).

src/script/view_model/ContentViewModel.ts
atomrc: is it finding the message to focus when opening the conversation? I feel this is what it's doing. The name should reflect that I guess. 

Aslo it's not clear to me why this particular function is changing the `cleared_timestamp` I feel it's not its responsibility (should probably be done in the glue function `showConversation`)

src/script/conversation/ConversationRepository.ts
arjita-mitra: checked and removed

src/script/view_model/ContentViewModel.ts
arjita-mitra: i have removed this method, cleared_timestamp is not a responsibility of this function so put it back in the glue function. Now  this.changeConversation(conversationEntity, messageEntity); is already a method so putting it back to the glue function too.

src/script/view_model/ContentViewModel.ts
arjita-mitra: got it, resolved it.

src/script/conversation/ConversationRepository.ts
arjita-mitra: its not required anymore.

atomrc (approved):
Looking good. I like that we removed so much code 🤩

PatrykBuniX (approved):
Good job 👍

PR #15636, opened 2023-08-21
refactor: Optimize app re-rendering

atomrc:
Because the `useWindowTitle` was a hook, everytime there was a change in there, the entire app would re-render. 
Extracting it into its own component makes title update only updating that headless component and leaves the app alone.

## Before

While searching for users, the entire app would re-render when users are fetched from backend
(squares appearing on screen represent component re-renders)

https://github.com/wireapp/wire-webapp/assets/1090716/e1b8c6ba-5221-4cf6-9c42-8ee78425ff9a

## After

No more full app re-render when backend search is done
(squares appearing on screen represent component re-renders)

https://github.com/wireapp/wire-webapp/assets/1090716/f5f8447f-161a-437c-bfe5-6473bc499075

### DIFF
--- a/src/script/page/useWindowTitle.ts
+++ b/src/script/page/useWindowTitle.ts
@@ -40,7 +40,7 @@ const windowTitleLogger = getLogger('WindowTitlesViewModel');
 const MIN_UNREAD_COUNT = 0;
 const MIN_CONNECTION_REQUEST_COUNT = 1;
 
-export const useWindowTitle = () => {
+const useWindowTitle = () => {
   const userState = container.resolve(UserState);
   const conversationState = container.resolve(ConversationState);
 
@@ -162,3 +162,9 @@ export const useWindowTitle = () => {
     initiateTitleUpdates();
   }, [initiateTitleUpdates, updateNotificationState]);
 };
+
+export function WindowTitleUpdater() {
+  useWindowTitle();
+
+  return null;
+}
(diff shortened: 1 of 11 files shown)


src/script/user/UserState.ts
```diff
@@ -39,10 +39,9 @@ export class UserState {
   public readonly isActivatedAccount: ko.PureComputed<boolean>;
   public readonly isTemporaryGuest: ko.PureComputed<boolean>;
   public readonly numberOfContacts: ko.PureComputed<number>;
-  public readonly self: ko.PureComputed<User>;
+  public readonly self = ko.observable<User | undefined>();
```
atomrc, line 42: another small improvements, is that `self` was re-computed everytime the `users` list would change. 
Which means that the entire app would re-render. 

Manually setting the `selfUser` at app start and then never update it again fixes this

src/script/page/useWindowTitle.ts
```diff
@@ -162,3 +162,9 @@ export const useWindowTitle = () => {
     initiateTitleUpdates();
   }, [initiateTitleUpdates, updateNotificationState]);
 };
+
+export function WindowTitleUpdater() {
```
przemvs: Add types there :)

aweiss-dev (approved):

PatrykBuniX (approved):

PR #15642, opened 2023-08-24
refactor: system message

PatrykBuniX:
System message component should be a generic UI component used for any kind of the system message, at the same time not being aware of different types of system message. All it should do is render UI based on the props we provide (we should not do any conditional rendering based on different types of system message).


src/script/components/MessagesList/Message/MessageWrapper.tsx
```diff
@@ -233,6 +233,13 @@ export const MessageWrapper: React.FC<MessageParams & {hasMarker: boolean; isMes
     return <FailedToAddUsersMessage isMessageFocused={isMessageFocused} message={message} />;
   }
   if (message.isSystem()) {
+    if (message.isConversationRename()) {
+      return (
+        <SystemMessage message={message}>
+          <div className="message-body font-weight-bold">{message.name}</div>
+        </SystemMessage>
+      );
+    }
```
przemvs: WDYT to render it in one return ?

```
<SystemMessage message={message}>
  {message.isConversationRename() && <div className="message-body font-weight-bold">{message.name}</div>}
</SystemMessage>
```

src/script/entity/message/MessageTimerUpdateMessage.ts
```diff
@@ -49,5 +49,8 @@ export class MessageTimerUpdateMessage extends SystemMessage {
 
       return this.user().isMe ? t('conversationResetTimerYou') : t('conversationResetTimer');
     });
+
+    this.icon = SystemMessageIcon.TIMER;
+    this.includeSenderName = true;
```
przemvs: I'll move it before "functions" :) WDYT?
In another Message classes also

src/script/components/MessagesList/Message/SystemMessage.tsx
```diff
@@ -17,51 +17,57 @@
...
-
   return (
     <>
       <div className="message-header" data-uie-name="element-message-system">
-        <div className="message-header-icon message-header-icon--svg text-foreground">
-          {message.system_message_type === SystemMessageType.CONVERSATION_RENAME && <Icon.Edit />}
-          {message.system_message_type === SystemMessageType.CONVERSATION_MESSAGE_TIMER_UPDATE && <Icon.Timer />}
-          {message.system_message_type === SystemMessageType.CONVERSATION_RECEIPT_MODE_UPDATE && <Icon.Read />}
-        </div>
+        {message.icon && (
```
przemvs: You always will have icon property in message ? :)

src/script/entity/message/SystemMessage.ts
```diff
@@ -23,16 +23,20 @@ import {Message} from './Message';
...
 
 import {SuperType} from '../../message/SuperType';
-import {SystemMessageType} from '../../message/SystemMessageType';
+import {SystemMessageIcon, SystemMessageType} from '../../message/SystemMessageType';
 
 export class SystemMessage extends Message {
   public caption: ko.PureComputed<string>;
   public system_message_type: SystemMessageType;
+  public icon: SystemMessageIcon | null;
+  public includeSenderName: boolean;
```
atomrc: Although the idea of making the component as dumb as possible, I think this implementation has some layering problems. 

Here we are in a pure logic class that should not have a single mention of UI layer parts. 
`icon` and `includeSenderName` are intents that are only for the UI (to which this class should not care about). 

Here a `type` property would be the right semantic. 
For the particular display logic, depending on the `type` we could have sub-components of the `SystemMessage` that renders differently according to the `type`

WDYT?

atomrc (changes requested):

przemvs (approved):

atomrc (approved):
Looks solid 🤘

PR #15644, opened 2023-08-24
feat:  mls decrypt error recovery [WPB-523]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-523" title="WPB-523" target="_blank">StoryWPB-523</a>  [Web] MLS Client recover from "not able to decrypt a message"
  </td></table>
  <br />
 

After webapp is notified about conversation being recovered from the epoch mismatch, we should inject a system message

Screenshot 2023-08-24 at 14 34 14


src/i18n/en-US.json
```diff
@@ -727,6 +727,7 @@
   "messageReactionDetails": "{{emojiCount}} reaction, react with {{emojiName}} emoji",
   "mlsToggleInfo": "When this is on, conversation will use the new messaging layer security (MLS) protocol.",
   "mlsToggleName": "MLS",
+  "mlsConversationRecovered": "The MLS group key was updated without our knowledge. This could happen due to lost messages between backends, or a bug. We have automatically rejoined the conversation, but you may have lost messages. Please contact your administrator.",
```
PatrykBuniX, line 730: This is likely to change in the future - there's no final copy for this feature yet.

przemvs (approved):

src/script/entity/message/MLSConversationRecoveredMessage.ts
```diff
@@ -0,0 +1,33 @@
...
+import {SystemMessageType} from 'src/script/message/SystemMessageType';
+import {t} from 'Util/LocalizerUtil';
+
+import {SystemMessage} from './SystemMessage';
+
+export class MLSConversationRecoveredMessage extends SystemMessage {
+  constructor() {
+    super();
+    this.system_message_type = SystemMessageType.MLS_CONVERSATION_RECOVERED;
+    this.caption = ko.pureComputed(() => t('mlsConversationRecovered'));
```
atomrc, line 31: doesn't need to be an observable, it never going to change overtime. 
(BTW if it's an observable in the other classes, it can be switched to a regular string :) )

atomrc (approved):
Pretty neat, one slight improvement to consider :)

src/script/entity/message/MLSConversationRecoveredMessage.ts
PatrykBuniX: Valid point, will investigate and try to improve this in a separate pr 😄

PR #15646, opened 2023-08-24
runfix: Let the feature change notifier store the previous value

atomrc:
This will make sure that feature that have changes while the user was offline will be detected by the component. 

Previously the `TeamRepository` was storing the value and reading it and it would read and load the new value even before the component is mounted. The component didn't get a chance to detect the change


src/script/page/components/FeatureConfigChangeNotifier/FeatureConfigChangeNotifier.store.ts
```diff
@@ -0,0 +1,36 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {FeatureList} from '@wireapp/api-client/lib/team/feature/';
+
+const LOCAL_STORAGE_FEATURE_CONFIG_KEY = 'FEATURE_CONFIG_KEY';
```
PatrykBuniX, line 22: I am thinking whether we should not include userId somewhere in the key? Consider this scenario:
- I log out as a user `A` (without clearing the client)
- I log in as a user `B`
- user `B` reads feature config that was stored by user A previously

PatrykBuniX (approved):
🚀

src/script/page/components/FeatureConfigChangeNotifier/FeatureConfigChangeNotifier.store.ts
atomrc: It's deal't in the payload actually, the payload is a object `{ [userId] : featureConfig }`. 

But very well though, man 😎

PR #15650, opened 2023-08-24
runfix: add correct error message for adding non federating users (WPB-4283)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4283" title="WPB-4283" target="_blank">TaskWPB-4283</a>  [Web] wrong message when failing to add single non-federating user
  </td></table>
  <br />
 

### Issue
- the error message for adding users to a conv with a non federating backend refers to offline backend
image

### Solution
- A string was missing, introduce the new error message
image

### Add-On

- mobile platforms are not including the full spec and are using a MVP, I'm adding a flag to use that MVP if product change their mind
image


src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -50,6 +50,9 @@ const errorMessageType = {
   [AddUsersFailureReasons.UNREACHABLE_BACKENDS]: 'OfflineBackend',
 } as const;
 
+// Mobile platforms are not using the full specs yet, this flag allows to switch to MVP specs if product decides to
+const useMvpString = false;
```
atomrc: why adding a flag and not just commenting code out (with the explanation that you added here)?

atomrc (approved):

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
V-Gira: I started out with commenting out the code :upside_down_face: 
Since we need to remove the code anyway, I'll revert to that

atomrc (approved):

PatrykBuniX (approved):

PR #15651, opened 2023-08-24
refactor: use plain string instead of observable for system message

PatrykBuniX:
System message caption doesn't need to be a computed value - it does not change over time.

### DIFF
--- a/src/script/conversation/EventMapper.ts
+++ b/src/script/conversation/EventMapper.ts
@@ -653,9 +653,7 @@ export class EventMapper {
    * @returns Rename message entity
    */
   private _mapEventRename({data: eventData}: LegacyEventRecord) {
-    const messageEntity = new RenameMessage();
-    messageEntity.name = eventData.name;
-    return messageEntity;
+    return new RenameMessage(eventData.name);
   }
 
   /**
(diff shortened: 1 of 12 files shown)


src/script/entity/message/DeleteConversationMessage.ts
```diff
@@ -28,19 +27,14 @@ import {SystemMessageType} from '../../message/SystemMessageType';
 import type {Conversation} from '../Conversation';
 
 export class DeleteConversationMessage extends SystemMessage {
-  public readonly system_message_type: SystemMessageType;
-  public readonly caption: ko.PureComputed<string>;
```
PatrykBuniX, line 32: Those are defined on SystemMessage itself, no need to redeclare them here

src/script/conversation/EventMapper.ts
```diff
@@ -653,9 +653,7 @@ export class EventMapper {
    * @returns Rename message entity
    */
   private _mapEventRename({data: eventData}: LegacyEventRecord) {
-    const messageEntity = new RenameMessage();
-    messageEntity.name = eventData.name;
-    return messageEntity;
+    return new RenameMessage(eventData.name);
```
PatrykBuniX, line 656: We can pass a new name directly to a constructor - more convenient way

thisisamir98 (approved):

PR #15686, opened 2023-08-29
fix: Filter DeleteMessage when computing the oldest loaded message [WPB-3506]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3506" title="WPB-3506" target="_blank">BugWPB-3506</a>  [Web] Conversation list is not updated with new messages
  </td></table>
  <br />
 

## The bug

loading a conversation that contains a `DeleteMessage` that hasn't been seen yet by the user will not show messages after this deleted message

## The problem

When messages arrives for a conversation that is not currently viewed by the user, we stack those messages in the `messages` property of the conversation. 

When the user clicks on a conversation to view, these are the steps that are triggered to load the messages we want to display to them:
- we check what is the **oldest** message that we have in the `messages` list;
- starting from this oldest message, we load 30 messages from DB in reverse chronology (so messages that were sent **before** this oldest message that we have loaded)
- those loaded messages are added to the `messages` list 
- the conversation can finally be showed to the user

This logic works pretty fine, except for `DeleteMessage` which can have a timestamp in the past. 
Since delete messages timestamp  are actually the timestamp of the message that was deleted. 

## An example

let's say we have a conversation between user1 and user2 with those messages:

- message1 by user2 (sent at timestamp 0)
- message2 by user2 (sent at timestamp 2)
- message3 by user2(sent at timestamp 3)

user1, is not currently looking at this conversation (so no messages are loaded into memory).

- user2, now deletes `message1`. 
- user1 receives the `delete` event and store it in the `messages` list
- the oldest message that user1 has loaded is the delete message for `message1` which has a timestamp of `0`. 
- user1 tries to load the conversation
- we try to load all messages in the conversation from timestamp `0`
- > we end-up not loading `message2` and `message3`.

## The fix

We computing the `oldest` message, we should not consider `delete` messages as we cannot trust their timestamp

### DIFF
--- a/src/script/guards/Message.ts
+++ b/src/script/guards/Message.ts
@@ -20,6 +20,7 @@
 import {Draft} from 'Util/DraftStateUtil';
 
 import {ContentMessage} from '../entity/message/ContentMessage';
+import {DeleteMessage} from '../entity/message/DeleteMessage';
 import {MemberMessage} from '../entity/message/MemberMessage';
 import {SuperType} from '../message/SuperType';
 
@@ -30,6 +31,9 @@ export const isReadableMessage = (message: any): message is ContentMessage =>
 
 export const isContentMessage = (message: any): message is ContentMessage =>
   message && 'super_type' in message && message.super_type === SuperType.CONTENT;
+
+export const isDeleteMessage = (message: any): message is DeleteMessage =>
+  message && 'super_type' in message && message.super_type === SuperType.DELETE;
 
 export const isMemberMessage = (message: any | undefined | null): message is MemberMessage =>
   message && 'super_type' in message && message.super_type === SuperType.MEMBER;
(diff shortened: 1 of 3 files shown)


src/script/guards/Message.ts
```diff
@@ -31,6 +32,9 @@ export const isReadableMessage = (message: any): message is ContentMessage =>
 export const isContentMessage = (message: any): message is ContentMessage =>
   message && 'super_type' in message && message.super_type === SuperType.CONTENT;
 
+export const isDeleteMessage = (message: any): message is DeleteMessage =>
```
tlebon, line 35: (thought): is this a DeleteMessage or a DeletedMessage? 🤔

PatrykBuniX (approved):
Amazing 🚀 Very nice explained, what a deep dive it was!

tlebon (approved):
great that you left some info for anyone reading this in the future

src/script/guards/Message.ts
atomrc: I'm not sure I undertand the question? 🤔 

What we are dealing with is a `DeleteMessage` the message that was deleted doesn't really have a representation (it was just deleted and replaced with that `DeleteMessage` which is a system message). 

Does that make sense?

tlebon: yes it is a DeleteMessage

PR #15687, opened 2023-08-30
feat: mls default ciphersuite config [WPB-3098]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3098" title="WPB-3098" target="_blank">EpicWPB-3098</a>  PoC "Postquantum" support in Wire: use a PQ/T hybrid quantum-resistant Key Exchange Mechanism
  </td></table>
  <br />
 

Allows to specify a default ciphersuite to be used in mls service when uploading key packages or creating a new mls group.

Needs https://github.com/wireapp/wire-web-packages/pull/5446


src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.tsx
```diff
@@ -22,38 +22,9 @@ import React from 'react';
 import {CSSObject} from '@emotion/react';
 import {ConversationProtocol} from '@wireapp/api-client/lib/conversation/NewConversation';
 
-import {t} from 'Util/LocalizerUtil';
+import {Ciphersuite} from '@wireapp/core-crypto';
```
PatrykBuniX: I think it makes more sense to use the enum directly from CC, so we don't have to manually add more options to the enum if new ciphersuite is introduced (what has just happened).

przemvs (approved):

src/script/components/panel/ConversationProtocolDetails/ConversationProtocolDetails.tsx
atomrc: hummm this doesn't sound safe. CoreCrypto is lazy loaded and in some case it might not even be loaded. 

Also the webapp should not even be aware that coreCrypto exists. 
Maybe the `core` could expose such enum, WDYT?

PatrykBuniX: Ok, but then should we reexport corecrypto's enum from core, or redefine similar enum ourselfes? 🤔

atomrc (approved):

PR #15689, opened 2023-08-30
refactor: Optimize app state slice selection

atomrc:
## Description

As per [zustand documentation](https://docs.pmnd.rs/zustand/recipes/recipes#selecting-multiple-state-slices), it's more optimal to select just a slice of a state rather than the entire state. 

This will make the component re-renders **only** if the slice of the state has changed (it's performing a strict equal between the previous selected state and the new one). 

This PR will optimize a few re-rendering when messages are received


PatrykBuniX (approved):
Nice 🚀

PR #15743, opened 2023-09-05
chore: indicate if connection map was triggered by socket event

PatrykBuniX:
## Description

Adds an optional param to mapConnection method indicating if it was triggered after parsing an event that came through websocket. This will be useful for 1:1 conversations with MLS - we need to add a delay in this case to reduce the possibility of a race condition where two users try to establish a 1:1 MLS conversation at the same time.

## Checklist

### DIFF
--- a/src/script/connection/ConnectionRepository.ts
+++ b/src/script/connection/ConnectionRepository.ts
@@ -125,8 +125,7 @@ export class ConnectionRepository {
       // Get conversation related to connection and set its type to 1:1
       // This case is important when the 'user.connection' event arrives after the 'conversation.member-join' event: https://wearezeta.atlassian.net/browse/SQCORE-348
 
-      const isWebSocketEvent = source === EventRepository.SOURCE.WEB_SOCKET;
-      amplify.publish(WebAppEvents.CONVERSATION.MAP_CONNECTION, connectionEntity, isWebSocketEvent);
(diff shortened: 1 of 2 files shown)


src/script/connection/ConnectionRepository.ts
```diff
@@ -124,7 +124,9 @@ export class ConnectionRepository {
       await this.userRepository.refreshUser(connectionEntity.userId);
       // Get conversation related to connection and set its type to 1:1
       // This case is important when the 'user.connection' event arrives after the 'conversation.member-join' event: https://wearezeta.atlassian.net/browse/SQCORE-348
-      amplify.publish(WebAppEvents.CONVERSATION.MAP_CONNECTION, connectionEntity);
+
+      const isWebSocketEvent = source === EventRepository.SOURCE.WEB_SOCKET;
+      amplify.publish(WebAppEvents.CONVERSATION.MAP_CONNECTION, connectionEntity, isWebSocketEvent);
```
atomrc: how about passing the `source` there and have the logic of `=== SOURCE.WEB_SOCKET` in the `mapConnection` method?

PatrykBuniX: @atomrc it works for me https://github.com/wireapp/wire-webapp/pull/15743/commits/d5398be40d4fdfb884839dea83a85a4957660fb0 ;)

atomrc (approved):

PR #15744, opened 2023-09-05
chore: cleanup unused UserRepository methods

atomrc:
## Description

Those mehods, because they are public, could not be automatically removed if not used. 
Doing some manual check show that those are not used anymore. 

## Checklist


atomrc:
Closed as some of those methods will be used in the future

PR #15747, opened 2023-09-06
runfix: use correct accent color for amber-dark-800

V-Gira:
## Description

It was reported that the self mention highlight was wrong for amber accent color in dark mode

before:
image

After:
image

## Checklist

### DIFF
--- a/src/style/common/accent-color.less
+++ b/src/style/common/accent-color.less
@@ -179,7 +179,7 @@
 @amber-dark-500: #ffd426;
 @amber-dark-600: #ccaa1e;
 @amber-dark-700: #997f17;
-@amber-dark-800: #201204;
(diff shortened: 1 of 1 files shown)


src/style/common/accent-color.less
```diff
@@ -179,7 +179,7 @@
 @amber-dark-500: #ffd426;
 @amber-dark-600: #ccaa1e;
 @amber-dark-700: #997f17;
-@amber-dark-800: #201204;
+@amber-dark-800: #66550f;
```
V-Gira, line 182: That change was made by mistake when introducing `dark-900` accent colors, i'm reverting to the previous `amber-dark-800` value (none of the other `dark-800` values changed)

przemvs (approved):

PR #15793, opened 2023-09-12
refactor: Inject self user into component

atomrc:
## Description
This will remove the tight coupling that we have between the components and the `UserState`. The `selfUser` needs to be explicitely injected into the components that need it

## Checklist

### DIFF
--- a/src/script/page/RightSidebar/GroupParticipantUser/GroupParticipantUser.tsx
+++ b/src/script/page/RightSidebar/GroupParticipantUser/GroupParticipantUser.tsx
@@ -83,8 +83,11 @@ const GroupParticipantUser: FC<GroupParticipantUserProps> = ({
     'isTeam',
     'team',
   ]);
-  const {isActivatedAccount, self: selfUser} = useKoSubscribableChildren(userState, ['isActivatedAccount', 'self']);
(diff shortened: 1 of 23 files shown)


src/script/page/RightSidebar/GroupParticipantUser/GroupParticipantUser.tsx
```diff
@@ -83,8 +83,11 @@ const GroupParticipantUser: FC<GroupParticipantUserProps> = ({
     'isTeam',
     'team',
   ]);
-  const {isActivatedAccount, self: selfUser} = useKoSubscribableChildren(userState, ['isActivatedAccount', 'self']);
-  const {is_verified: isSelfVerified} = useKoSubscribableChildren(selfUser, ['is_verified']);
+  const {self: selfUser} = useKoSubscribableChildren(userState, ['self']);
+  const {is_verified: isSelfVerified, isActivatedAccount} = useKoSubscribableChildren(selfUser, [
```
przemvs: Can we pass there selfUser instead of userState ? 🤔  And define selfUser also in parent component? 🤔

przemvs (approved):

PR #15816, opened 2023-09-14
fix: Do not show failed messages as delivered

atomrc:
## Description

We used to have a check that was to permissive to know if a message is delivered. 
Now that we have errors in the `StatusType` enum that have a value higher than `StatusType.DELIVERED` we need to be explicit about the status that could be considered `delivered`. 

### Before 
image

### After
image

## Checklist

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -1215,12 +1215,12 @@ export class MessageRepository {
   private async updateMessageAsFailed(conversationEntity: Conversation, eventId: string, error: unknown) {
     try {
       const messageEntity = await this.getMessageInConversationById(conversationEntity, eventId);
-      if (isBackendError(error) && error.label === BackendErrorLabel.FEDERATION_REMOTE_ERROR) {
-        messageEntity.status(StatusType.FEDERATION_ERROR);
-        return this.eventService.updateEvent(messageEntity.primary_key, {status: StatusType.FEDERATION_ERROR});
-      }
-      messageEntity.status(StatusType.FAILED);
-      return this.eventService.updateEvent(messageEntity.primary_key, {status: StatusType.FAILED});
+      const errorStatus =
+        isBackendError(error) && error.label === BackendErrorLabel.FEDERATION_REMOTE_ERROR
+          ? StatusType.FEDERATION_ERROR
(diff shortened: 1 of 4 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -1215,12 +1215,12 @@ export class MessageRepository {
...
-        return this.eventService.updateEvent(messageEntity.primary_key, {status: StatusType.FEDERATION_ERROR});
-      }
-      messageEntity.status(StatusType.FAILED);
-      return this.eventService.updateEvent(messageEntity.primary_key, {status: StatusType.FAILED});
+      const errorStatus =
+        isBackendError(error) && error.label === BackendErrorLabel.FEDERATION_REMOTE_ERROR
+          ? StatusType.FEDERATION_ERROR
+          : StatusType.FAILED;
+      messageEntity.status(errorStatus);
+      return this.eventService.updateEvent(messageEntity.primary_key, {status: errorStatus});
```
atomrc, line 1223: This is a small refactoring to avoid code duplication

src/script/entity/Conversation.ts
```diff
@@ -964,12 +964,12 @@ export class Conversation {
...
    * Get the last delivered message.
    */
-  getLastDeliveredMessage(): Message | undefined {
+  private getLastDeliveredMessage(): Message | undefined {
     return this.messages()
       .slice()
       .reverse()
       .find(messageEntity => {
-        const isDelivered = messageEntity.status() >= StatusType.DELIVERED;
+        const isDelivered = [StatusType.DELIVERED, StatusType.SEEN].includes(messageEntity.status());
```
atomrc, line 972: This is the fix. 
For more explanation, here is what the enum looks like

```ts
export enum StatusType {
  DELIVERED = 3,
  FAILED = 0,
  FEDERATION_ERROR = 5,
  SEEN = 4,
  SENDING = 1,
  SENT = 2,
  UNSPECIFIED = -1,
}
```

as you can see `FEDERATION_ERROR` has a higher value than `DELIVERED` and thus caused the bug

V-Gira (approved):

PatrykBuniX (approved):

przemvs (approved):

PR #15841, opened 2023-09-18
runfix: identify mls capable client

PatrykBuniX:
## Description

Fixes the way we identify a client being MLS capable. We were mistakenly assuming that client with `mls_public_keys` field is an mls client (even if this field contained empty object, see https://github.com/wireapp/wire-web-packages/pull/5516). `mls_public_keys` is empty object by default, we need to check if it really contains any values (Public key is uploaded after mls device is registered).

## Checklist


src/script/client/ClientUtils.ts
```diff
@@ -30,3 +30,15 @@ export const wasClientActiveWithinLast4Weeks = ({last_active: lastActiveISODate}
...
+/**
+ * Check if client is MLS-capable device.
+ * Client is considered MLS capable if it was registered and has uploaded MLS public keys.
+ * @param client - client to check
+ * @returns {boolean}
+ */
+export const isClientMLSCapable = ({mls_public_keys: mlsPublicKeys}: RegisteredClient): boolean => {
+  const haveClientUploadedPublicKeys = Object.values(mlsPublicKeys).length > 0;
+
+  return haveClientUploadedPublicKeys;
```
przemvs: **Suggestion:** Simplify it :)

```suggestion
  return Object.values(mlsPublicKeys).length > 0;
```

PatrykBuniX: fixed https://github.com/wireapp/wire-webapp/pull/15841/commits/622104d337b8ef87d8b4d69a51b4bf7a4210886e

przemvs (approved):

PR #15888, opened 2023-09-25
feat: evaluate list of self supported protocols on team protocols update

PatrykBuniX:
## Description

Every time team (or backend) admin changes the list of supported protocols (in feature config) we want to also re-evaluate the list of self user's supported protocols. This PR adds this functionality. Every time we receive feature config update, and see that `"mls"` was the  feature that had been updated, we compare the local list of current supported protocols stored locally in team config with the list received in `"mls"` feature config update. If supported protocols were changed, we re-evaluate the list of self supported protocols and update it on backend.

## Checklist

### DIFF
--- a/src/script/team/TeamRepository.ts
+++ b/src/script/team/TeamRepository.ts
@@ -29,7 +29,7 @@ import type {
   TeamUpdateEvent,
 } from '@wireapp/api-client/lib/event';
 import {TEAM_EVENT} from '@wireapp/api-client/lib/event/TeamEvent';
-import {FeatureStatus} from '@wireapp/api-client/lib/team/feature/';
+import {FeatureStatus, FeatureList} from '@wireapp/api-client/lib/team/feature/';
 import type {TeamData} from '@wireapp/api-client/lib/team/team/TeamData';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
 import {amplify} from 'amplify';
@@ -74,7 +74,10 @@ export interface AccountInfo {
 }
 
 type Events = {
-  featureUpdated: Pick<TeamFeatureConfigurationUpdateEvent, 'name' | 'data'>;
+  featureUpdated: {
+    prevFeatureList?: FeatureList;
+    event: TeamFeatureConfigurationUpdateEvent;
+  };
 };
 
 export class TeamRepository extends TypedEventEmitter<Events> {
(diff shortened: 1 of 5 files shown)


przemvs (approved):

src/script/team/TeamRepository.ts
```diff
@@ -399,15 +414,17 @@ export class TeamRepository {
...
   ): Promise<void> => {
     if (source !== EventSource.WEBSOCKET) {
       // Ignore notification stream events
       return;
     }
+
     // When we receive a `feature-config.update` event, we will refetch the entire feature config
-    await this.updateFeatureConfig();
+    const {prevFeatureList} = await this.updateFeatureConfig();
+    this.emit('featureUpdated', {event, prevFeatureList});
```
atomrc, line 427: suggestion: maybe we could just send the previous value rather than the entire feature config, WDYT?

atomrc (commented):
question: Will this work when the entire feature list is updated (that happens, for example, once every 24h)?
We might need to account for this as well

src/script/team/TeamRepository.ts
PatrykBuniX: Well, I tried to do it but didn't manage to achieve it type-wise. When accessing `prevFeatureList[event.name]` at this point, the type will not be properly narrowed down to e.g. `FeatureMLS` (we don't know what's `event.name` at compile time, it's `FEATURE_KEY` type)

Event data is nicely tied to its name.
Screenshot 2023-09-27 at 11 25 17

But we cannot tie specific feature to it by FEATURE_KEY at this point. It's still union of different features.
Screenshot 2023-09-27 at 11 27 59

Maybe it's possible but i cannot find a way? Any ideas?

PatrykBuniX:
> question: Will this work when the entire feature list is updated (that happens, for example, once every 24h)? We might need to account for this as well

@atomrc 
How about emitting the event inside `updateFeatureConfig` method instead? This one is called in case of 24h timer as well.

EDIT: Nevermind, we don't have access to event there, I will just handle this case separately.

src/script/team/TeamRepository.ts
atomrc: ah nevermind we can probably change that later, no biggies here 👌

atomrc (approved):

PR #15889, opened 2023-09-25
feat: manage conversation verification state for mls [WPB-482]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-482" title="WPB-482" target="_blank">StoryWPB-482</a>  [Web] Create separate unverified/verified/degraded state machine for MLS/E2EI
  </td></table>
  <br />
 

We already have a state handler for proteus conversations.

The goal was to add a new state handler for the MLS protocol, based on the automated E2EI process.

We also have to consider a mixed conversation state, but we handle mixed the same as proteus for now.


src/script/conversation/ConversationVerificationStateHandler/proeteusStateHandler/proteusStateHandler.ts
atomrc, line 1: There is a typo in the name of the folder `proeteus` ;)

src/script/conversation/ConversationVerificationStateHandler/mlsStateHandler/index.ts
```diff
@@ -0,0 +1,20 @@
+/*
```
przemvs: I think file name should be uppercased. MLSStateHandler :)

src/script/conversation/ConversationVerificationStateHandler/mlsStateHandler/mlsStateHandler.ts
```diff
@@ -0,0 +1,200 @@
+/*
```
przemvs: I think file name should be uppercased: MLSStateHandler :)

src/script/conversation/ConversationRepository.ts
```diff
@@ -269,11 +270,16 @@ export class ConversationRepository {
...
-    this.verificationStateHandler = new ConversationVerificationStateHandler(
+
+    // we register and store a handler, that we can manually trigger for incoming events from proteus and mixed conversations
+    this.proteusVerificationStateHandler = new ProteusConversationVerificationStateHandler(
       this.eventRepository,
       this.userState,
       this.conversationState,
     );
+    // we register a handler that will handle MLS conversations on its own
+    new MLSConversationVerificationStateHandler(this.eventRepository, this.conversationState, this.core);
```
atomrc: suggestion: Why creating a class for this? Since it returns nothing that seems to be consumed, why not just exporting a function `registerMLSVerificationStateHandler`?

package.json
```diff
@@ -4,6 +4,7 @@
     "@datadog/browser-rum": "^4.49.0",
     "@emotion/react": "11.11.1",
     "@loadable/component": "^5.15.3",
+    "@peculiar/x509": "^1.9.5",
```
PatrykBuniX: Let's use exact package version

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  private async checkEpoch({groupId, epoch}: {groupId: string; epoch: number}): Promise<void> {
+    this.logger.log(`Epoch changed to ${epoch} for groupId ${groupId}`);
+    const conversationEntity = getConversationByGroupId({conversationState: this.conversationState, groupId});
+    if (!conversationEntity) {
+      this.logger.error(`Epoch changed but conversationEntity can't be found`);
+      return;
+    }
+    // We only want to process MLS conversations
+    if (!isMLSConversation(conversationEntity)) {
```
PatrykBuniX: Since only MLS conversations have groupId, let's have `getConversationByGroupId` return `MLSConversation | undefined` already, so there's no need to do extra check already 🤔

src/script/conversation/ConversationRepository.ts
tlebon: we dont want this to be saved to a method as well?

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -0,0 +1,210 @@
+/*
+ * Wire
+ * Copyright (C) 2018 Wire Swiss GmbH
```
tlebon:
```suggestion
 * Copyright (C) 2023 Wire Swiss GmbH
```

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -0,0 +1,137 @@
...
+          userIds.find(userId => matchQualifiedIds(userId, userIdInConversation)),
+        );
+
+        if (!!matchingUserIds.length) {
+          return {conversationEntity, userIds: matchingUserIds};
+        }
+      }
+      return undefined;
+    })
+    .filter(activeConversationInfo => !!activeConversationInfo) as {
```
tlebon: i always prefer this style but this is just a suggestion:

```suggestion
    .filter(Boolean) as {
```

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -0,0 +1,137 @@
...
+    conversationEntity: Conversation;
+    userIds: QualifiedId[];
+  }[];
+};
```
tlebon: can this just be as GetActiveConversationsWithUsersResult

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  return false;
+};
+
+/**
+ * Check whether to verify conversation and set corresponding state
+ *
+ * @param conversationEntity Conversation entity to evaluate
+ * @returns `true` if conversation state changed to verified
+ */
+interface WillChangeToVerifiedParams {
```
tlebon: can these two functions  (willChangeToDegraded, willChangeToVerified) reuse a base function and return differently based on a param? seems they are inverse of each other.

aweiss-dev: We would kill readability with that approach.
They do the opposite of each other, and in my head, this means they should be distinct functions.

src/script/conversation/ConversationRepository.ts
aweiss-dev: Its not needed i think. This class has no public accessible methods that would need to be called. Its self sufficient

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
aweiss-dev: Lets use a flatMap for this and get rid of "as"

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
atomrc: Well the logic is not entirely similar (since the degraded state could be either `DEGRADED` or `UNVERIFIED`) so it might, indeed, not make sense to try to factor those. It's sad that it's the only diff, otherwise it would actually have made a lot of sense to have a generic function that check if a conversation would go from one state to the other.

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  conversationEntity: Conversation;
+  logger: Logger;
+}
+export const willChangeToVerified = ({conversationEntity, logger}: WillChangeToVerifiedParams): boolean => {
```
atomrc: Naming is a little weird here `willChangeToVerified` seems to imply that it will just return a boolean. 
But it's actually also changing the verified state of the conversation. 

We have a single responsibility principe failure here. Not sure how this could be handled better though 🤔

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -0,0 +1,210 @@
...
+import {ConversationState} from '../../ConversationState';
+import {ConversationVerificationState} from '../../ConversationVerificationState';
+import {EventBuilder} from '../../EventBuilder';
+import {getActiveConversationsWithUsers, willChangeToDegraded, willChangeToVerified} from '../shared';
+
+export class ProteusConversationVerificationStateHandler {
+  private readonly logger: Logger;
+
+  constructor(
+    private readonly eventRepository: EventRepository,
```
atomrc: Not really comfortable with the idea of injecting the `eventRepository` here. 
Injecting the event does not belong in this layer of the app, we are to deep into specific logic at this point. 
It should not know anything about the event system. 

The consumer of this class should be the one creating and injecting those events

atomrc: Idea: Instead we devise a nice little callback API;

```ts
constructor(userState, conversationState, onConversationVerificationStateChange: (conversationId: QualifiedId, newVerificationState: VerificationState));

// ...

if (willDegrade(conversation)) {
  onConversationVerificationStateChange(conversation.qualifiedId, VerificationState.DEGRADED);
}
```

This way this layer doesn't really know what should happen when degradation happens and instead the consumer is responsible for providing those

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
aweiss-dev: Lets go with "attemptChangeToVerified" and "attemptChangeToDegraded"

atomrc: yeah, at least it conveys the right idea. I'm fine with it 👌

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -0,0 +1,173 @@
...
+    // We need to check if the core service is available
+    if (!this.core.service?.mls) {
+      this.logger.error('MLS service not available');
+      return;
+    }
+    // We need to check if the e2eIdentity service is available
+    if (!this.core.service?.e2eIdentity) {
+      this.logger.error('E2E identity service not available');
+      return;
+    }
```
przemvs, line 55: Maybe You can do there smth like that:
image

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+        this.checkChangeToDegraded(conversationEntity, userIds, VerificationMessageType.UNVERIFIED);
+      }
+    });
+  };
+
+  /**
+   * Self user or other participant added clients.
+   * @param userId ID of user that added client (can be self user ID)
+   */
+  readonly onClientAdded = (userId: QualifiedId, _clientEntity?: ClientEntity): void => {
```
przemvs: We need clientEntity there? Because it's not used

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  /**
+   * Change that could verify conversation.
+   *
+   * @param conversationEntity Changed conversation entity
+   * @returns `true` if state changed
+   */
+  private checkChangeToVerified(conversationEntity: Conversation): boolean {
+    // We want to process only Proteus and Mixed conversations
+    if (isProteusConversation(conversationEntity) || isMixedConversation(conversationEntity)) {
```
przemvs, line 155: 'We want to process only Proteus and Mixed conversations' - there we have or not and :D

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+   * @param type Type of degradation
+   * @returns `true` if state changed
+   */
+  private checkChangeToDegraded(
+    conversationEntity: Conversation,
+    userIds: QualifiedId[],
+    type: VerificationMessageType,
+  ): boolean {
+    // We want to process only Proteus and MLS conversations
+    if (isProteusConversation(conversationEntity) || isMixedConversation(conversationEntity)) {
```
przemvs, line 184: Maybe create fn to check if isProteus or Mixed conversation, e.g. isProteusOrMixedConversation, and You can reuse it

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
aweiss-dev: Why is this necessary if it is not shared but only used in one place? :)

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
aweiss-dev: No, a conversation is either mixed or proteus or MLS. It can't be more than one at a time.

src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
aweiss-dev: Sure, but I don't intend to reuse it. I like having it specific. :)

przemvs (approved):

PatrykBuniX (approved):

PR #15894, opened 2023-09-26
fix: Add pending icon to outgoing connection request (WPB-4849)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4849" title="WPB-4849" target="_blank">BugWPB-4849</a>  Missing pending icon on outgoing connection request
  </td></table>
  <br />
 

## Description
This commit addresses a backend issue where outgoing connection requests are incorrectly categorized as type 3 (1:1 conversations) instead of type 2 (connect). To work around this issue, we now utilize the "isRequest" property from the users[0] of the conversation users. This approach ensures that the conversation type is accurately represented as type 2 in the incoming list of conversations, aligning with the expected behavior.

Before:
image

After:
image

## Checklist

### DIFF
--- a/src/script/components/list/ConversationListCellStatusIcon.tsx
+++ b/src/script/components/list/ConversationListCellStatusIcon.tsx
@@ -37,18 +37,7 @@ const ConversationListCellStatusIcon = ({conversation}: ConversationListCellStat
     unreadState,
     mutedState,
     isRequest,
-  } = useKoSubscribableChildren(conversation, [
-    'isGroup',
-    'is1to1',
-    'selfUser',
-    'participating_user_ets',
-    'display_name',
-    'removed_from_conversation',
-    'availabilityOfUser',
-    'unreadState',
-    'mutedState',
-    'isRequest',
-  ]);
+  } = useKoSubscribableChildren(conversation, ['participating_user_ets', 'unreadState', 'mutedState', 'isRequest']);
 
   const cellState = useMemo(() => generateCellState(conversation), [unreadState, mutedState, isRequest]);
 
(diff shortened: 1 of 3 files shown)


src/script/components/list/ConversationListCellStatusIcon.tsx
```diff
@@ -0,0 +1,133 @@
...
+    'isGroup',
+    'is1to1',
+    'selfUser',
+    'participating_user_ets',
+    'display_name',
+    'removed_from_conversation',
+    'availabilityOfUser',
+    'unreadState',
+    'mutedState',
+    'isRequest',
```
przemvs: Some of this deps is not used :)

src/script/components/list/ConversationListCellStatusIcon.tsx
```diff
@@ -0,0 +1,133 @@
...
+  ]);
+
+  const cellState = useMemo(() => generateCellState(conversation), [unreadState, mutedState, isRequest]);
+
+  const {isRequest: isUserRequest} = useKoSubscribableChildren(users[0], ['isRequest']);
```
atomrc: How about fixing the `isRequest` of the `Conversation` class so that it does exactly this. 
We could then (and should) remove all traces of `CONVERSATION_TYPE.CONNECT` since it's not reliable anymore

src/script/components/list/ConversationListCellStatusIcon.tsx
thisisamir98: nice catch, fixed in: https://github.com/wireapp/wire-webapp/pull/15894/commits/f7815ce662e2a849deb8000efbf5a1c7f20bffe3

src/script/components/list/ConversationListCellStatusIcon.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/15894/commits/80675c242b6346006faceb76575551721973a1cb

atomrc (approved):
Pretty solid PR 💪 😎

PR #15897, opened 2023-09-27
feat: refresh self protocols on team refresh

PatrykBuniX:
## Description

Makes user refresh self supported protocols after team (and its whole feature list) was refreshed. This happens every 24h.

### DIFF
--- a/src/script/team/TeamRepository.ts
+++ b/src/script/team/TeamRepository.ts
@@ -78,6 +78,7 @@ type Events = {
     prevFeatureList?: FeatureList;
     event: TeamFeatureConfigurationUpdateEvent;
   };
+  teamRefreshed: void;
 };
 
 export class TeamRepository extends TypedEventEmitter<Events> {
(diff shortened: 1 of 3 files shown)


przemvs (approved):

src/script/team/TeamRepository.ts
```diff
@@ -156,6 +157,7 @@ export class TeamRepository extends TypedEventEmitter<Events> {
       try {
         await this.getTeam();
         await this.updateFeatureConfig();
+        this.emit('teamRefreshed');
```
atomrc, line 160: suggestion: maybe we could narrow down the scope of this event to just `featureConfigUpdated` and move it to the `updateFeatureConfig` function (this way we are sure this event will emit even if we just update the config and not the entire team). WDYT?

PatrykBuniX: `updateFeatureConfig` is called every time **any** of the feature is updated though, so we would refresh protocols too often in my opinion.

atomrc: Hummm I see. ok then, fine for now 👍

atomrc (approved):

PR #15898, opened 2023-09-27
fix: Avoid race condition when dealing with image and image placeholder [WPB-4890]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4890" title="WPB-4890" target="_blank">BugWPB-4890</a>  [Web] After back online previously sent images shown as grey icons
  </td></table>
  <br />
 

## Description

When incoming messages arrive, they are treated one-by-one but, because of the async nature of their processing, there could be 2 events that are being processed at the same time. 

That's not a big issue in general, but in the particular case of images, we can fall into this scenario:
- the placeholder of an image is sent (containing all the metadata of the image)
- processing starts for the placeholder
- the actual image arrives
- processing starts for the image
- when the image is processed we look, in the DB, for the placeholder in order to replace it
- since the placeholder is not yet save to DB, it's not found
-> as a result: we do not replace the placeholder with the actual image and save both events 
-> on the UI level, the placeholder shows forever, along with the actual image

### Before

the placeholder and the image are processed at the same. When receiving the event, we only display the placeholder (the other message is filtered). When switching conversation, we can see the actual image along with the placeholder

https://github.com/wireapp/wire-webapp/assets/1090716/4205390c-fcfa-44f0-8179-91c354d9af7b

### After

The placeholder is completely processed before the actual image and correctly replaced in DB

https://github.com/wireapp/wire-webapp/assets/1090716/40e9b703-1fc8-4603-8b51-ec2a74a7f069

## Checklist

### DIFF
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -167,6 +167,10 @@ export class EventRepository {
     }
   };
 
+  /**
+   * this function will process any incoming event. It is being queue in case 2 events arrive at the same time.
+   * Processing events should happen sequentially (thus the queue)
+   */
   private readonly handleIncomingEvent = queue(async (payload: HandledEventPayload, source: NotificationSource) => {
     try {
       await this.handleEvent(payload, source);
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -168,7 +168,7 @@ export class EventRepository {
   };
 
   /**
-   * this function will process any incoming event. It is being queue in case 2 events arrive at the same time.
+   * this function will process any incoming event. It is being queued in case 2 events arrive at the same time.
    * Processing events should happen sequentially (thus the queue)
    */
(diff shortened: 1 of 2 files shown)


aweiss-dev (approved):

tlebon (approved):

V-Gira (approved):

src/script/event/EventRepository.ts
```diff
@@ -166,7 +167,11 @@ export class EventRepository {
     }
   };
 
-  private readonly handleIncomingEvent = async (payload: HandledEventPayload, source: NotificationSource) => {
+  /**
+   * this function will process any incoming event. It is being queue in case 2 events arrive at the same time.
```
V-Gira:
```suggestion
   * this function will process any incoming event. It is being queued in case 2 events arrive at the same time.
```

V-Gira (approved):

PR #15925, opened 2023-10-02
runfix: Do not show connection requests on group conversations

atomrc:
## Description

Fix for #15894 . 
We want to show connection request state only for 1:1 conversations, not group conversations

## Checklist


PatrykBuniX (approved):
Good catch!

PR #15927, opened 2023-10-02
fix: do not create a placeholder 1:1 conversation for accepted conversations

PatrykBuniX:
## Description

Do not create a placeholder conversation after sending a POST request to /connections/{domain}/{uid} responds with a connection which status is other than `"sent"`.

## Checklist

### DIFF
--- a/src/script/components/panel/UserActions.tsx
+++ b/src/script/components/panel/UserActions.tsx
@@ -19,6 +19,7 @@
 
 import React from 'react';
 
+import {ConnectionStatus} from '@wireapp/api-client/lib/connection';
 import {CONVERSATION_TYPE} from '@wireapp/api-client/lib/conversation';
 import {amplify} from 'amplify';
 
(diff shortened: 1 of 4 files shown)


src/script/connection/ConnectionRepository.ts
```diff
@@ -229,7 +234,6 @@ export class ConnectionRepository {
             break;
           }
         }
-        return false;
       }
       throw error;
```
PatrykBuniX: Instead of returning false, we just throw an error we will ignore later

src/script/components/panel/UserActions.tsx
```diff
@@ -219,19 +220,24 @@ const UserActions: React.FC<UserActionsProps> = ({
...
-            if (!connectionIsSent) {
+            try {
+              const {connectionStatus, conversationId} = await actionsViewModel.sendConnectionRequest(user);
+
+              // If connection's state is SENT, we create a local 1:1 conversation that will act as a placeholder
+              // before the other user has accepted the request.
+              const connectionConversation =
+                connectionStatus === ConnectionStatus.SENT
+                  ? createPlaceholder1to1Conversation(user, selfUser)
+                  : await actionsViewModel.getConversationById(conversationId);
```
PatrykBuniX: This is the main change - we do not create a placeholder for connections that are already accepted (or are in some other state).

src/script/connection/ConnectionRepository.ts
atomrc: humm this actually changes the behavior. Previously we would throw unexpected errors and they would be visible in the console, but now they are going to be swallowed into the void

atomrc (changes requested):

src/script/connection/ConnectionRepository.ts
PatrykBuniX: If you follow the path starting from `UserActions.tsx`, we were actually never catching any (unexpected) errors seems like.

atomrc (approved):
Seems solid. Nice fix man 🙏

src/script/connection/ConnectionRepository.ts
PatrykBuniX: fixed, by returning `null` in this case

PR #15933, opened 2023-10-05
refactor: self supported protocols check

PatrykBuniX:
## Description

Some refactoring on updating self supported protocols:
- evaluating self supported protocols in now a separate module as it's business logic mostly
- wrapped a call to BE with try/catch in refreshSupportedProtocols (in case it runs on some older API version)
- periodic self supported protocols check is not hidden under the feature flag anymore (there's no harm in users updating it even if mls or mls migration features are not enabled)

## Checklist


przemvs (approved):

src/script/self/SelfRepository.ts
```diff
@@ -208,31 +128,23 @@ export class SelfRepository {
...
-    };
-    await refreshProtocolsTask();
+    await this.refreshSelfSupportedProtocols();
 
     await this.core.recurringTaskScheduler.registerTask({
       every: TIME_IN_MILLIS.DAY,
-      task: refreshProtocolsTask,
+      task: async () => {
+        await this.refreshSelfSupportedProtocols();
+      },
```
atomrc:
```suggestion
      task: () => this.refreshSelfSupportedProtocols()
```

maybe we could change the signature of a `Task` so that it allows `unknown | Promise<unknown>` rather than `void` to avoid those situations. WDYT?

atomrc (approved):

src/script/self/SelfRepository.ts
PatrykBuniX: Yeah, makes sense, will update it

PatrykBuniX: done with https://github.com/wireapp/wire-webapp/pull/15933/commits/37abf0972d48785b6c0d3b9f2ae419d9ed55583c

atomrc (approved):

PR #15950, opened 2023-10-09
runfix: Reset complete draft state when sending message

atomrc:
## Description

Fixes a bug where, when you send a reply to a message then switch conversation and go back to the original conversation, then the editor is still in reply mode. 

### After 

https://github.com/wireapp/wire-webapp/assets/1090716/006e4ea6-dbba-4873-8e3b-a4f63b45084c

### Before

https://github.com/wireapp/wire-webapp/assets/1090716/8a95b6eb-3863-448d-845b-0377f41e91b5

## Checklist

### DIFF
--- a/src/script/components/InputBar/InputBar.tsx
+++ b/src/script/components/InputBar/InputBar.tsx
@@ -202,12 +202,11 @@ export const InputBar = ({
     ),
   });
 
-  const resetDraftState = (resetInputValue = false) => {
-    if (resetInputValue) {
-      editorRef.current?.update(() => {
-        $getRoot().clear();
-      });
-    }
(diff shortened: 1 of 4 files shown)


src/script/components/InputBar/InputBar.tsx
```diff
@@ -345,7 +344,7 @@ export const InputBar = ({
     }
 
     editorRef.current?.focus();
-    editorRef.current?.update(() => $getRoot().clear());
+    resetDraftState();
```
atomrc, line 347: This is the actual fix

przemvs (approved):

PatrykBuniX (approved):

PR #15965, opened 2023-10-10
refactor: Streamline leaving and clearing conversation

atomrc:
## Description

Avoid mixing clearing and leaving conversation. Make the code much leaner and avoid confusion between those 2 actions

**previously**: `leaveConversation` and `clearConversation` could both do leaving AND clearing the conversation. 
**now**: `leaveConversation` only does leaving and `clearConversation` only does clearing

## Checklist


PatrykBuniX (approved):
Nice, I agree it was hard to follow the flow with the previous structure 👌

PR #15972, opened 2023-10-11
runfix: leave mls conversation

PatrykBuniX:
## Description

https://github.com/wireapp/wire-webapp/pull/15965 introduced a regression in leaving MLS conversations. Leaving a MLS conversation is different than removing other users (we cannot send remove proposal for the self client, remove proposal has to be created by backend). When leaving MLS conversation we should remove self user "the proteus way", so just send a request to backend. For more details see https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/157287189/Use+case+leaving+a+conversation.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1742,9 +1742,13 @@ export class ConversationRepository {
    */
   public async leaveConversation(conversation: Conversation, {localOnly = false} = {}) {
     const userQualifiedId = this.userState.self().qualifiedId;
-    return localOnly
-      ? this.removeMembersLocally(conversation, [userQualifiedId])
-      : this.removeMembers(conversation, [userQualifiedId]);
+
+    if (localOnly) {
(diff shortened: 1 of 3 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -1742,9 +1742,13 @@ export class ConversationRepository {
...
     const userQualifiedId = this.userState.self().qualifiedId;
-    return localOnly
-      ? this.removeMembersLocally(conversation, [userQualifiedId])
-      : this.removeMembers(conversation, [userQualifiedId]);
+
+    if (localOnly) {
+      return this.removeMembersLocally(conversation, [userQualifiedId]);
+    }
+
+    const events = await this.removeMembersFromConversation(conversation, [userQualifiedId]);
```
PatrykBuniX, line 1750: We cannot call `.removeMembers` method. We want to always delete the user "the proteus way", no need to check for conversation protocol.

atomrc (approved):

PR #15989, opened 2023-10-12
refactor: Extract federation event handling to an event processor

atomrc:
## Description

In an effort to remove amplify from the codebase, this will add a mechanism to register event processors to the `EventRepository`. 
This way the repository can make direct calls of those processors and `await` them before continuing while still being unaware what the job of a processor is. 

In the future, we could convert all those `amplify.publish` by a focused specialized eventProcessor

## Checklist

### DIFF
--- a/src/script/event/EventRepository.ts
+++ b/src/script/event/EventRepository.ts
@@ -23,7 +23,6 @@ import {
   ConversationOtrMessageAddEvent,
   ConversationMLSMessageAddEvent,
 } from '@wireapp/api-client/lib/event';
-import type {BackendEvent} from '@wireapp/api-client/lib/event';
 import {NotificationSource, HandledEventPayload} from '@wireapp/core/lib/notification';
 import {amplify} from 'amplify';
 import ko from 'knockout';
@@ -38,6 +37,7 @@ import {queue} from 'Util/PromiseQueue';
 import {TIME_IN_MILLIS} from 'Util/TimeUtil';
 
 import {ClientEvent} from './Client';
+import {EventProcessor, IncomingEvent} from './EventProcessor';
 import type {EventService} from './EventService';
 import {EventSource} from './EventSource';
 import {EVENT_TYPE} from './EventType';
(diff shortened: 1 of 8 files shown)


src/script/event/EventRepository.ts
```diff
@@ -332,15 +334,14 @@ export class EventRepository {
    * @param source Source of notification
    */
   private async distributeEvent(event: IncomingEvent, source: EventSource) {
+    await Promise.all(this.eventProcessors.map(processor => processor.processEvent(event, source)));
```
atomrc, line 337: Now all processors will be awaited before continuing (will remove the `hack` on line 358 in this file). 
In the long run this should make things way cleaner

src/script/event/processor/FederationEventProcessor/FederationEventProcessor.ts
```diff
@@ -0,0 +1,147 @@
...
+
+import {
+  getUsersToDeleteFromFederatedConversations,
+  getFederationDeleteEventUpdates,
+} from './ConversationFederationUtils';
+
+import {EventProcessor, IncomingEvent} from '../../EventProcessor';
+import {EventRepository} from '../../EventRepository';
+
+export class FederationEventProcessor implements EventProcessor {
```
atomrc, line 40: This file is basically a copy/paste of the code removed from the ConversationRepository

aweiss-dev (approved):

src/script/event/EventRepository.ts
PatrykBuniX: **question**: Wouldn't it be more safe to use `Promise.allSettled` instead to be sure all of the processors fire? With `.all` it will reject after one of the processors fail.

aweiss-dev: I assumed that we want all of them to succeed?

PatrykBuniX: Depends if we want to follow `all-or-nothing` or `at-least-something` approach 😅

PatrykBuniX: There's only one processor right now, there could be a lot of them in the future, I don't think one of them failing should abort all the rest (and then also the order matters, which is not good i'd say).

PatrykBuniX (approved):

src/script/event/EventRepository.ts
PatrykBuniX: Never mind, `.all` will not make all promises stop executing, it will just not await them (wait for their result), they should still be running in the background.

PR #15991, opened 2023-10-12
feat: refetch user on navigating to conversation

PatrykBuniX:
## Description

When navigating to a 1:1 conversation, we should refetch user's supported protocols to make sure we're up-to-date with their list of supported protocols and that the correct protocol will be chosen for 1:1 conversation.

## Checklist

### DIFF
--- a/src/script/view_model/ContentViewModel.ts
+++ b/src/script/view_model/ContentViewModel.ts
@@ -150,7 +150,7 @@ export class ContentViewModel {
       return conversationEntity;
     }
 
-    return this.conversationRepository.init1to1Conversation(conversationEntity);
+    return this.conversationRepository.init1to1Conversation(conversationEntity, true);
   };
 
   /**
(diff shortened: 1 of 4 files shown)


src/script/view_model/ContentViewModel.ts
```diff
@@ -150,7 +150,7 @@ export class ContentViewModel {
       return conversationEntity;
     }
 
-    return this.conversationRepository.init1to1Conversation(conversationEntity);
+    return this.conversationRepository.init1to1Conversation(conversationEntity, true);
```
atomrc, line 153: Maybe we could do that only if we are in an MLS env, would that make sense?

atomrc (approved):

src/script/view_model/ContentViewModel.ts
PatrykBuniX: This functionality should work fine even if MLS is disabled on backend, so there's no need to perform any mls support check here ;)

PR #15994, opened 2023-10-12
fix: dont run the MLSStateHandler without MLS enabled

aweiss-dev:
This PR fixes the problem that the MLSConversationVerificationStateHandler Class gets instantiated with disabled MLS.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -68,6 +68,7 @@ import {
 } from 'Util/StringUtil';
 import {TIME_IN_MILLIS} from 'Util/TimeUtil';
 import {isBackendError} from 'Util/TypePredicateUtil';
+import {supportsMLS} from 'Util/util';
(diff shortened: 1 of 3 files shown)


atomrc (approved):

src/script/conversation/ConversationRepository.ts
```diff
@@ -272,12 +272,15 @@ export class ConversationRepository {
...
       this.conversationState,
     );
-    // we register a handler that will handle MLS conversations on its own
-    registerMLSConversationVerificationStateHandler(
-      this.onConversationVerificationStateChange,
-      this.conversationState,
-      this.core,
-    );
+
+    if (Config.getConfig().FEATURE.ENABLE_MLS === true) {
```
PatrykBuniX: There's a `supportsMLS` function in `util.ts` file we can use instead.

PatrykBuniX (approved):

PR #16017, opened 2023-10-17
fix: missing degredation warning

aweiss-dev:
Fix for automation issue:

```
- 1x Missing system message when conversation gets unverified after unverifying one of my devices [C404406]
```

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
@@ -181,7 +181,7 @@ export class ProteusConversationVerificationStateHandler {
     userIds: QualifiedId[],
     type: VerificationMessageType,
   ): boolean {
-    // We want to process only Proteus and MLS conversations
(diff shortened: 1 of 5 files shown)


src/script/conversation/ConversationVerificationStateHandler/Proteus/ProteusStateHandler.ts
```diff
@@ -182,10 +183,8 @@ export class ProteusConversationVerificationStateHandler {
   ): boolean {
     // We want to process only Proteus and MLS conversations
     if (isProteusConversation(conversationEntity) || isMixedConversation(conversationEntity)) {
```
PatrykBuniX: **question**: Not relevant to changes, but the comment says we process only proteus and mls, but the check lets proteus and mixed conversations in, is that intentional? As a recap, mixed conversation is really a proteus conversation that is being migrated, it has a MLS group under the hood where proteus conversation participants are being added during migration (It will change the protocol to MLS once the migration is finalised)

PatrykBuniX (approved):

PR #16020, opened 2023-10-17
runfix: problem with existing database entries

aweiss-dev:
Reverts wireapp/wire-webapp#16017

### DIFF
--- a/src/script/conversation/ConversationMapper.test.ts
+++ b/src/script/conversation/ConversationMapper.test.ts
@@ -44,7 +44,6 @@ import {BaseError} from 'src/script/error/BaseError';
 import {createUuid} from 'Util/uuid';
 
 import {ACCESS_STATE} from './AccessState';
-import {ConversationVerificationState} from './ConversationVerificationState';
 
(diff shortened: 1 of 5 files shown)


src/script/conversation/ConversationMapper.test.ts
```diff
@@ -370,7 +369,7 @@ describe('ConversationMapper', () => {
         status: 0,
         team_id: teamId,
         type: 0,
-        verification_state: ConversationVerificationState.UNVERIFIED,
```
atomrc, line 373: I believe this was a nice change, why removing it? 🤔

src/script/conversation/ConversationMapper.ts
```diff
@@ -61,7 +60,7 @@ export interface SelfStatusUpdateDatabaseData {
   otr_muted_status: number;
   receipt_mode: number;
   status: number;
-  verification_state: ConversationVerificationState;
+  verification_state: number;
```
atomrc, line 63: This was also a valid change

PR #16030, opened 2023-10-18
fix: factor in the selfClient when it comes to the calculation of isMLSVerified

aweiss-dev:
Previously, I didn't consider the self-client when updating/reading the user's client's certificate info.
This caused the self-user to be falsely marked as validated for MLS conversations, even if the user didn't go through the E2EE enrollment process.

This is the fix for that problem.

Refactoring Part:
I took this task to do a minor refactoring of the self-client. The Observable was unnecessary since the local client doesn't change during the app's runtime. 
This fact should be more clear now.

### DIFF
--- a/src/script/client/ClientRepository.test.ts
+++ b/src/script/client/ClientRepository.test.ts
@@ -56,7 +56,7 @@ describe('ClientRepository', () => {
       const client = new ClientEntity(false, null);
       client.id = clientId;
 
-      testFactory.client_repository['clientState'].currentClient(client);
+      testFactory.client_repository['clientState'].currentClient = client;
 
       const clients = [
         {class: ClientClassification.DESKTOP, id: '706f64373b1bcf79'},
@@ -112,9 +112,9 @@ describe('ClientRepository', () => {
       spyOn(clientService, 'getClientById').and.returnValue(Promise.resolve(clientPayloadServer));
       spyOn(clientService, 'putClientCapabilities').and.returnValue(Promise.resolve());
 
-      return testFactory.client_repository.getValidLocalClient().then(clientObservable => {
-        expect(clientObservable).toBeDefined();
-        expect(clientObservable().id).toBe(clientId);
+      return testFactory.client_repository.getValidLocalClient().then(client => {
+        expect(client).toBeDefined();
+        expect(client.id).toBe(clientId);
       });
     });
 
@@ -168,7 +168,7 @@ describe('ClientRepository', () => {
     beforeEach(() => {
       (jasmine as any).getEnv().allowRespy(true);
       spyOn(Runtime, 'isDesktopApp').and.returnValue(false);
-      testFactory.client_repository['clientState'].currentClient(undefined);
+      testFactory.client_repository['clientState'].currentClient = undefined;
     });
 
     it('returns true on Electron', () => {
@@ -179,7 +179,7 @@ describe('ClientRepository', () => {
         type: ClientType.PERMANENT,
       };
       const clientEntity = ClientMapper.mapClient(clientPayload, true, null);
-      testFactory.client_repository['clientState'].currentClient(clientEntity);
+      testFactory.client_repository['clientState'].currentClient = clientEntity;
       spyOn(Runtime, 'isDesktopApp').and.returnValue(true);
       const isPermanent = testFactory.client_repository.isCurrentClientPermanent();
 
@@ -194,7 +194,7 @@ describe('ClientRepository', () => {
         type: ClientType.TEMPORARY,
       };
       const clientEntity = ClientMapper.mapClient(clientPayload, true, null);
-      testFactory.client_repository['clientState'].currentClient(clientEntity);
+      testFactory.client_repository['clientState'].currentClient = clientEntity;
       spyOn(Runtime, 'isDesktopApp').and.returnValue(true);
       const isPermanent = testFactory.client_repository.isCurrentClientPermanent();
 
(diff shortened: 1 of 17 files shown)


src/script/entity/User/User.ts
```diff
@@ -210,10 +212,14 @@ export class User {
       return this.devices().every(client_et => client_et.meta.isVerified?.());
     });
     this.isMLSVerified = ko.pureComputed(() => {
-      if (this.devices().length === 0 && !this.isMe) {
-        return false;
+      if (this.devices().length === 0) {
+        if (!this.isMe) {
+          return false;
+        }
+        const clientState = container.resolve(ClientState);
```
PatrykBuniX: Hmm it seems wrong to me to have ClientState resolved on User's entity level 🤔 How about just adding a field to user entity next to `devices` called e.g. `currentDevice` and set it once the self client is initialised?

aweiss-dev: I thought about that, but threw it in the bin since this code is only used in one place and the user doesn't take use of the selfClient in any other way.

We can change that if it is necessary.

PatrykBuniX: That would be nice, User entity should be as dumb as possible

atomrc: oh yeah, entities using the app state is definitely something that, architecture-wise, might create tough to debug situation. 
Explicitely setting the `selfDevice` at load time feels safer and cleaner in term of app-layering

aweiss-dev: Since the function returned an observable, the localClient in User is also an Observable. Doesn't hurt :)

src/script/entity/User/User.ts
```diff
@@ -54,6 +54,7 @@ export class User {
   public readonly connection: ko.Observable<ConnectionEntity>;
   /** does not include current client/device */
   public readonly devices: ko.ObservableArray<ClientEntity>;
+  public localClient: ko.Observable<ClientEntity> | undefined;
```
PatrykBuniX: It doesn't need to be observable. We're assigning it only once on app init and never update the reference.

PatrykBuniX (approved):

src/script/client/ClientRepository.test.ts
```diff
@@ -56,7 +56,7 @@ describe('ClientRepository', () => {
       const client = new ClientEntity(false, null);
```
aweiss-dev, line 56: Just changed to function calls to direct access

src/script/client/ClientRepository.ts
```diff
@@ -245,13 +245,16 @@ export class ClientRepository {
    * Get and validate the local client.
    * @returns Resolves with an observable containing the client if valid
    */
-  async getValidLocalClient(): Promise<ko.Observable<ClientEntity>> {
+  async getValidLocalClient(): Promise<ClientEntity> {
```
aweiss-dev, line 248: This function changed

src/script/components/HistoryExport/HistoryExport.tsx
```diff
@@ -179,15 +179,19 @@ const HistoryExport = ({switchContent, user, clientState = container.resolve(Cli
       setNumberOfRecords(numberOfRecords);
       setNumberOfProcessedRecords(0);
 
-      const archiveBlob = await backupRepository.generateHistory(
```
aweiss-dev, line 182: Just added an undefined check, this was a hidden "Problem" before since the currentClient cut have been undefined

src/script/conversation/MessageRepository.test.ts
```diff
@@ -76,7 +76,7 @@ async function buildMessageRepository(): Promise<[MessageRepository, MessageRepo
   const userState = new UserState();
```
aweiss-dev, line 76: Just changed to function calls to direct access

src/script/conversation/MessageRepository.ts
```diff
@@ -753,7 +753,7 @@ export class MessageRepository {
```
aweiss-dev, line 753: Just changed to function calls to direct access

src/script/event/EventRepository.ts
```diff
@@ -63,7 +62,6 @@ import {Warnings} from '../view_model/WarningsContainer';
```
aweiss-dev, line 62: Removed unused currentClient

src/script/page/AppLock/AppLock.tsx
```diff
@@ -211,7 +211,7 @@ const AppLock: React.FC<AppLockProps> = ({
     const target = event.target as HTMLFormElement & {password: HTMLInputElement};
```
aweiss-dev, line 211: Just changed to function calls to direct access

src/script/page/MainContent/panels/preferences/devices/DevicesPreferences.test.tsx
```diff
@@ -40,7 +40,7 @@ function createDevice(): ClientEntity {
 describe('DevicesPreferences', () => {
```
aweiss-dev, line 40: Just changed to function calls to direct access

src/script/util/DebugUtil.ts
```diff
@@ -334,7 +334,7 @@ export class DebugUtil {
     messageId: string,
     conversationId: string = this.conversationState.activeConversation().id,
```
aweiss-dev, line 335: Just changed to function calls to direct access

test/helper/TestFactory.js
```diff
@@ -123,7 +123,7 @@ export class TestFactory {
     currentClient.time = '2016-10-07T16:01:42.133Z';
```
aweiss-dev, line 123: Just changed to function calls to direct access

src/script/components/HistoryExport/HistoryExport.tsx
atomrc: In the future this should be injected, we should not consume the state so deep in the layers of the app (not a change for now :) )

atomrc (approved):

PR #16083, opened 2023-10-23
runfix: mls 1to1 establishment on user protocols update

PatrykBuniX:
## Description

When the other user update their supported protocols, we're receiving `user.update` event that could arrive before the welcome message for 1:1 conversation with the updated user. This PR adds a delay to make sure the welcome message is not wasted, in case the self client would react to `user.update` event and try to establish mls group before receiving the welcome (corecrypto would throw "conversation already exists error").

This PR also adds missing disabled attribute to call button on medium breakpoint for read-only conversation.

## Checklist

### DIFF
--- a/src/script/components/TitleBar/TitleBar.tsx
+++ b/src/script/components/TitleBar/TitleBar.tsx
@@ -342,6 +342,7 @@ export const TitleBar: React.FC<TitleBarProps> = ({
                 css={{marginBottom: 0}}
                 onClick={onClickStartAudio}
                 data-uie-name="do-call"
+                disabled={isReadOnlyConversation}
               >
                 <Icon.Pickup />
               </IconButton>
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1131,7 +1131,8 @@ export class ConversationRepository {
     const localMLSConversation = this.conversationState.findMLS1to1Conversation(otherUserId);
 
     if (protocol === ConversationProtocol.MLS || localMLSConversation) {
-      return this.initMLS1to1Conversation(otherUserId, isMLSSupportedByTheOtherUser, isLiveUpdate);
+      const shouldDelayMLSGroupEstablishment = isLiveUpdate && isMLSSupportedByTheOtherUser;
+      return this.initMLS1to1Conversation(otherUserId, isMLSSupportedByTheOtherUser, shouldDelayMLSGroupEstablishment);
     }
 
     const proteusConversation = await this.getOrCreateProteus1To1Conversation(userEntity);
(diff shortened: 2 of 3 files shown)


src/script/components/TitleBar/TitleBar.tsx
```diff
@@ -342,6 +342,7 @@ export const TitleBar: React.FC<TitleBarProps> = ({
                 css={{marginBottom: 0}}
                 onClick={onClickStartAudio}
                 data-uie-name="do-call"
+                disabled={isReadOnlyConversation}
```
PatrykBuniX, line 345: This one was already added there https://github.com/wireapp/wire-webapp/pull/16083/files#diff-bdc492056c4abd9bd96ba59134b8a002c8d11abc91d3b555dd1ec1967ae8b8f3R319, but it was missing on medium breakpoint

src/script/conversation/ConversationRepository.ts
```diff
@@ -1130,7 +1131,8 @@ export class ConversationRepository {
     const localMLSConversation = this.conversationState.findMLS1to1Conversation(otherUserId);
 
     if (protocol === ConversationProtocol.MLS || localMLSConversation) {
-      return this.initMLS1to1Conversation(otherUserId, isMLSSupportedByTheOtherUser);
+      const shouldDelayMLSGroupEstablishment = isLiveUpdate && isMLSSupportedByTheOtherUser;
```
PatrykBuniX, line 1138: When the other user update their supported protocols, we're receiving user.update event that could come before the welcome message for 1:1 conversation with them. We have to add a delay to make sure the welcome message is not wasted, in case the self client would establish mls group before receiving it.

src/script/conversation/ConversationRepository.ts
```diff
@@ -1500,16 +1518,6 @@ export class ConversationRepository {
...
 
-    // After connection request is accepted, both sides (users) of connection will react to conversation status update event.
-    // We want to reduce the possibility of two users trying to establish an MLS group at the same time.
-    // A user that has previously sent a connection request will wait for a short period of time before establishing an MLS group.
-    // It's very likely that this user will receive a welcome message after the user that has accepted a connection request, establishes an MLS group without any delay.
-    if (shouldDelayGroupEstablishment) {
-      await new Promise(resolve =>
-        setTimeout(resolve, ConversationRepository.CONFIG.ESTABLISH_MLS_GROUP_AFTER_CONNECTION_IS_ACCEPTED_DELAY),
-      );
-    }
```
PatrykBuniX, line 1511: This part was moved at the top of init mls conversation method, so we delay not only the establishment part but also the migration part - so there's no 3s delay between deleting a proteus 1:1 conversation and mls 1:1 conversation appearing back.

src/script/conversation/ConversationRepository.ts
```diff
@@ -1741,19 +1748,25 @@ export class ConversationRepository {
...
 
-    const isWebSocketEvent = source === EventSource.WEBSOCKET;
     const isConnectionAccepted = connectionEntity.isConnected();
 
     // Check what protocol should be used for 1:1 conversation
     const {protocol, isMLSSupportedByTheOtherUser, isProteusSupportedByTheOtherUser} =
       await this.getProtocolFor1to1Conversation(otherUserId);
 
+    const isWebSocketEvent = source === EventSource.WEBSOCKET;
+    const shouldDelayMLSGroupEstablishment = isWebSocketEvent && isMLSSupportedByTheOtherUser;
```
PatrykBuniX, line 1762: No need to delay anything if the other user does not support mls - we will not receive a welcome message in this case anyway.

src/script/conversation/ConversationRepository.ts
atomrc: This would be a precious "in-code" comment to add  there 😉

atomrc (approved):

PR #16086, opened 2023-10-23
runfix: Make sure event is stored in DB before processing quotes

atomrc:
## Description

When processing incoming event, the QuotedMessageMiddleware rely on the fact that the received event was already stored in DB. 
Previously we used a hack to do the DB update only *after* the event has been stored (with `setTimeout`), but now we can do that cleanly and run the `Storage` middleware *before* the Quoted middleware

## Checklist

### DIFF
--- a/src/script/event/preprocessor/QuotedMessageMiddleware.ts
+++ b/src/script/event/preprocessor/QuotedMessageMiddleware.ts
@@ -83,7 +83,6 @@ export class QuotedMessageMiddleware implements EventMiddleware {
       if (quote && typeof quote !== 'string' && 'message_id' in quote && 'id' in event) {
         quote.message_id = event.id as string;
       }
-      // we want to update the messages quoting the original message later, thus the timeout
       await this.eventService.replaceEvent(reply);
     });
     const decoratedData = {...event.data, quote: originalEvent.data.quote};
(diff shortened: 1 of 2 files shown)


src/script/main/app.ts
```diff
@@ -387,9 +387,9 @@ export class App {
 
       eventRepository.setEventProcessMiddlewares([
         serviceMiddleware,
-        quotedMessageMiddleware,
         readReceiptMiddleware,
         eventStorageMiddleware,
+        quotedMessageMiddleware,
```
atomrc, line 392: We want to make sure the event is stored before processing quote updates. It's now possible to do it cleanly with the order of the middlwares

src/script/event/preprocessor/QuotedMessageMiddleware.ts
```diff
@@ -78,13 +78,13 @@ export class QuotedMessageMiddleware implements EventMiddleware {
...
     this.logger.info(`Updating '${replies.length}' replies to updated message '${originalMessageId}'`);
-    replies.forEach(reply => {
+    replies.forEach(async reply => {
       const quote = reply.data.quote;
       if (quote && typeof quote !== 'string' && 'message_id' in quote && 'id' in event) {
         quote.message_id = event.id as string;
       }
       // we want to update the messages quoting the original message later, thus the timeout
-      setTimeout(() => this.eventService.replaceEvent(reply));
+      await this.eventService.replaceEvent(reply);
```
atomrc, line 86: The setTimeout can now be removed (we don't need to hackingly wait for the storage middleware to have done its job)

V-Gira (approved):

PatrykBuniX (approved):

PR #16101, opened 2023-10-24
feat: Sort user reactions chronologically [WPB-4253]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4253" title="WPB-4253" target="_blank">EpicWPB-4253</a>  Align order of given reaction pills
  </td></table>
  <br />
 

## Description

We were, previously, storing reactions as a map of `Record<userId, reactions>`. Which means that the order of the reactions received is lost. 

This PR changes the format in which the reactions are stored to an array of tuples `[reaction, userIds]`. This way, reactions are sorted chronologically. 

### Before

https://github.com/wireapp/wire-webapp/assets/1090716/4452dd39-cdd0-4b82-ab24-4a0221acb096

### After 

https://github.com/wireapp/wire-webapp/assets/1090716/e859d960-da19-404b-88bb-88abe641f281

## Checklist

### DIFF
--- a/src/script/conversation/EventMapper.ts
+++ b/src/script/conversation/EventMapper.ts
@@ -24,6 +24,7 @@ import {LinkPreview, Mention} from '@wireapp/protocol-messaging';
 
 import {t} from 'Util/LocalizerUtil';
 import {getLogger, Logger} from 'Util/Logger';
+import {userReactionMapToReactionMap} from 'Util/ReactionUtil';
 import {base64ToArray} from 'Util/util';
 
 import {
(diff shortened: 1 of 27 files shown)


src/script/conversation/EventMapper.ts
```diff
@@ -188,7 +189,7 @@ export class EventMapper {
     }
 
     if (event.reactions) {
-      originalEntity.reactions(event.reactions);
+      originalEntity.reactions(userReactionMapToReactionMap(event.reactions));
```
atomrc, line 192: mapping the reactions to the new format (making sure the old format is never lying around in the app state)

PatrykBuniX (approved):
Looks solid 😎

PR #16106, opened 2023-10-25
runfix: Return reaction event instead of targeted message event

atomrc:
## Description

When a reaction arrives we, mistakenly, returned the message targeted by the reaction instead of the reaction event itself. 
This was because we needed the message primary key to update the DB during the DBOperation execution. 

## Checklist

### DIFF
--- a/src/script/event/preprocessor/EventStorageMiddleware/eventHandlers/reactionEventHandler.ts
+++ b/src/script/event/preprocessor/EventStorageMiddleware/eventHandlers/reactionEventHandler.ts
@@ -35,6 +35,7 @@ function computeEventUpdates(target: StoredEvent<MessageAddEvent>, reactionEvent
   } = reactionEvent;
   const reactionMap = target.reactions ? userReactionMapToReactionMap(target.reactions) : [];
   return {
+    primary_key: target.primary_key,
     reactions: addReaction(reactionMap, reaction, qualified_from ?? {id: from, domain: ''}),
     version: version,
   };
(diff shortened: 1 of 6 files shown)


src/script/event/preprocessor/EventStorageMiddleware/eventHandlers/reactionEventHandler.ts
```diff
@@ -50,7 +51,7 @@ export const handleReactionEvent: EventHandler = async (event, {findEvent}) => {
   }
   return {
     type: 'sequential-update',
-    event: targetEvent,
+    event,
```
atomrc, line 54: This is this actual fix :)

src/script/event/preprocessor/EventStorageMiddleware/types.ts
```diff
@@ -27,7 +27,7 @@ import {IdentifiedUpdatePayload} from '../../EventService';
 export type HandledEvents = ClientConversationEvent | ConversationEvent;
 export type DBOperation =
   | {type: 'update'; event: HandledEvents; updates: IdentifiedUpdatePayload}
-  | {type: 'sequential-update'; event: EventRecord; updates: Partial<EventRecord>}
+  | {type: 'sequential-update'; event: HandledEvents; updates: IdentifiedUpdatePayload}
```
atomrc, line 30: I could have detected it just from the type level. It was weird to return a stored event at this point :)

PatrykBuniX (approved):

PR #16102, opened 2023-10-24
runfix: Update openState of emoji/mention picker when menu is closed

atomrc:
## Description

Will prevent blocking sending when a mention valid mention is being typed but the user manually closes the selection menu.

### Before

https://github.com/wireapp/wire-webapp/assets/1090716/c0e1941a-f599-42ac-852a-8c41c82f1e8c

### After

https://github.com/wireapp/wire-webapp/assets/1090716/327d069e-91e0-44e7-87c7-d3cbfee100ce

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

PR #16129, opened 2023-10-30
runfix: Do not try to match mention trigger on full text

atomrc:
## Description

Broken since https://github.com/wireapp/wire-webapp/pull/16102/files#diff-89a3f1b31103998681d89eb526fd0e7d793234a5d4a9e33eb4c05660d1b059abL104

## Checklist

### DIFF
--- a/src/script/components/RichTextEditor/plugins/EmojiPickerPlugin/EmojiPickerPlugin.tsx
+++ b/src/script/components/RichTextEditor/plugins/EmojiPickerPlugin/EmojiPickerPlugin.tsx
@@ -120,7 +120,7 @@ export function EmojiPickerPlugin({openStateRef}: Props) {
       return null;
     }
 
-    return checkForEmojis(info.textContent);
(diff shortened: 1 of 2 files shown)


src/script/components/RichTextEditor/plugins/EmojiPickerPlugin/EmojiPickerPlugin.tsx
```diff
@@ -120,7 +120,7 @@ export function EmojiPickerPlugin({openStateRef}: Props) {
       return null;
     }
 
-    return checkForEmojis(info.textContent);
+    return checkForEmojis(text);
```
atomrc, line 123: The different between the `text` passed to the function and the `info.textContent` is that:
- `text` contains the text up to the cursor (so removes what is after the cursor)
- `info.textContent` is the entire content of the input bar

PatrykBuniX (approved):

PR #16132, opened 2023-10-31
fix: backend sends gracePeriod in seconds instead of ms

aweiss-dev:
This PR fixes a bug with the E2EI GracePeriod. The backend sends the GracePeriod in Seconds, I did initially think it would be MS, they get converted now.

### DIFF
--- a/src/script/E2EIdentity/E2EIdentity.ts
+++ b/src/script/E2EIdentity/E2EIdentity.ts
@@ -42,7 +42,7 @@ export enum E2EIHandlerStep {
 
 interface E2EIHandlerParams {
   discoveryUrl: string;
-  gracePeriodInMS: number;
+  gracePeriodInSeconds: number;
 }
 
 class E2EIHandler {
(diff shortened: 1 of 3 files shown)


src/script/E2EIdentity/E2EIdentity.ts
```diff
@@ -54,12 +54,12 @@ class E2EIHandler {
   private gracePeriodInMS: number;
   private currentStep: E2EIHandlerStep | null = E2EIHandlerStep.UNINITIALIZED;
 
-  private constructor({discoveryUrl, gracePeriodInMS}: E2EIHandlerParams) {
+  private constructor({discoveryUrl, gracePeriodInSeconds}: E2EIHandlerParams) {
     // ToDo: Do these values need to te able to be updated? Should we use a singleton with update fn?
     this.discoveryUrl = discoveryUrl;
-    this.gracePeriodInMS = gracePeriodInMS;
+    this.gracePeriodInMS = gracePeriodInSeconds * 1000;
```
arjita-mitra: suggestion: good to use 1000 as const

atomrc: There is a util for that. Look for `TIME_IN_MILLIS.SECOND` ;)

atomrc (approved):

PR #16138, opened 2023-11-01
fix: Use qualified Ids when matching users for search filtering [WPB-4823]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4823" title="WPB-4823" target="_blank">BugWPB-4823</a>  [Web] [Bundestag] Team members not showing up in group search
  </td></table>
  <br />
 

## Description

This attempts to fix an issue where some local users could not be found in the search. 
The theory being: since we try to match users by comparing actual user entities (the entire object), if 2 different entities represent the same user, then the check will be false. 

```js
const userId = 'user1';
const userA = new User(userId);
const userB = new User(userId);

userA === userB; // false ❌
```

To fix that, we now match users only by their IDs (which means that 2 entities representing the same user would return `true`)

```js
const userId = 'user1';
const userA = new User(userId);
const userB = new User(userId);

userA.id === userB.id; // true ✅
```

## Checklist

### DIFF
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -196,7 +196,7 @@ export class ConversationState {
   }
 
   /**
-   * indicate whether the selfUser has a conversation with this other user
+   * indicate whether the selfUser has a conversation (1:1 or group conversation) with this other user
    * @param user the user to check
    */
   hasConversationWith(user: User) {
--- a/src/script/conversation/ConversationState.ts
+++ b/src/script/conversation/ConversationState.ts
@@ -195,6 +195,14 @@ export class ConversationState {
         });
   }
 
+  /**
+   * indicate whether the selfUser has a conversation with this other user
+   * @param user the user to check
+   */
+  hasConversationWith(user: User) {
+    return this.connectedUsers().some(connectedUser => matchQualifiedIds(connectedUser.qualifiedId, user.qualifiedId));
+  }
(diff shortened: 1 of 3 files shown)


src/script/conversation/ConversationState.ts
```diff
@@ -195,6 +195,14 @@ export class ConversationState {
         });
   }
 
+  /**
+   * indicate whether the selfUser has a conversation with this other user
```
V-Gira, line 199:
```suggestion
   * indicate whether the selfUser has a one one one conversation with this other user
```
Question: is this correct?
Just making sure if a clarification is needed to differentiate user known through group conversation but not connected

atomrc: no, it's actually any kind of conversation (including group conversations). We can have unconnected users there (as opposed to what the `connectedUsers` suggests 😬 )

V-Gira:
```suggestion
   * indicate whether the selfUser has a one on one conversation with this other user
```
Question: is this correct?
Just making sure if a clarification is needed to differentiate user known through group conversation but not connected

atomrc:
```suggestion
   * indicate whether the selfUser has a conversation (1:1 or group conversation) with this other user
```

V-Gira (approved):

PR #16149, opened 2023-11-02
fix: add missing undefined/null check to e2ei typeguards

aweiss-dev:


src/script/page/components/FeatureConfigChange/guards.ts
```diff
@@ -19,8 +19,16 @@
 
 import {FeatureMLS, FeatureMLSE2EId} from '@wireapp/api-client/lib/team';
 
-export const isFeatureMLSE2EI = (feature: any): feature is FeatureMLSE2EId =>
-  'config' in feature && 'verificationExpiration' in feature.config;
+const isDefinedObject = (value: unknown): value is {} => typeof value === 'object' && value !== null;
```
atomrc:
```suggestion
const isObject = (value: unknown): value is {} => typeof value === 'object' && value !== null;
```

I think this is sufficient

PatrykBuniX (approved):

PR #16159, opened 2023-11-06
refactor: Avoid temporal coupling when mapping user.inTeam property

atomrc:
## Description

In preparation for a change in how team members are dealt with, we need to clear how the `inTeam` property is computed. 
It should not be `set` but rather `computed` everytime we need it. 
This prevent temporal coupling (in case we are mapping users before the state is fully ready). 

## Checklist

### DIFF
--- a/src/script/components/panel/UserDetails.tsx
+++ b/src/script/components/panel/UserDetails.tsx
@@ -21,6 +21,7 @@ import React, {useEffect} from 'react';
 
 import {amplify} from 'amplify';
 import {ErrorBoundary} from 'react-error-boundary';
+import {container} from 'tsyringe';
 
 import {WebAppEvents} from '@wireapp/webapp-events';
 
@@ -30,6 +31,7 @@ import {ErrorFallback} from 'Components/ErrorFallback';
 import {Icon} from 'Components/Icon';
 import {UserClassifiedBar} from 'Components/input/ClassifiedBar';
 import {UserName} from 'Components/UserName';
+import {TeamState} from 'src/script/team/TeamState';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
 import {t} from 'Util/LocalizerUtil';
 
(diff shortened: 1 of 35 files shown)


src/script/components/panel/UserDetails.tsx
```diff
@@ -69,13 +72,10 @@ const UserDetailsComponent: React.FC<UserDetailsProps> = ({
     amplify.publish(WebAppEvents.USER.UPDATE, participant.qualifiedId);
   }, [participant]);
 
-  const isFederated = participant.isFederated;
-  const isGuest = !isFederated && user.isGuest;
```
atomrc, line 73: This was redundant check (`isGuest` already guarantees that the user is not federated)

src/script/user/UserMapper.ts
```diff
@@ -187,12 +181,8 @@ export class UserMapper {
       }
     }
 
-    const currentTeam = this.teamState.team()?.id;
     if (teamId) {
       userEntity.teamId = teamId;
-      if (!userEntity.isFederated && currentTeam && currentTeam === teamId) {
-        userEntity.inTeam(true);
-      }
```
atomrc, line 195: This is the meat of the PR. removing this temporal coupling. 

The problem with this is:
- if we create a user outside of the `UserMapper` and it has a valid teamId, it won't be considered a team member;
- if we map users before the `teamState` is correctly set, we also will miss team members

src/script/components/panel/UserDetails.tsx
```diff
@@ -69,13 +72,10 @@ const UserDetailsComponent: React.FC<UserDetailsProps> = ({
...
-
   return (
     <div className="panel-participant">
       <div className="panel-participant__head">
-        {user.inTeam ? (
+        {teamState.isInTeam(participant) ? (
```
arjita-mitra, line 78: suggestion: giving it a name will help to make its intention clear. isParticipantinTeam something like this, if i understood the meaning correctly.

atomrc: do you mean: assigning this to a variable or changing the name of the method?

arjita-mitra: Assigning it to a variable

arjita-mitra (approved):

PR #16160, opened 2023-11-06
feat: Do not load all team members on app load [WPB-4553 WPB-4548 WPB-1912]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4553" title="WPB-4553" target="_blank">StoryWPB-4553</a>  [Web] Stop maintaining a local list of all team members
  </td></table>
  <br />
 

## Description

This will prevent all team members from being loaded at app load time. 
The idea is to only load all the users we have direct connection with. This includes:
- users that are connected to the selfUser (via a connection request)
- users the selfUser have interacted with through a conversation (but which are not necessarily connected to me)

## Checklist

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -415,15 +415,18 @@ export class App {
       onProgress(10);
       telemetry.timeStep(AppInitTimingsStep.INITIALIZED_CRYPTOGRAPHY);
 
-      const {members: teamMembers} = await teamRepository.initTeam(selfUser.teamId);
+      const connections = await connectionRepository.getConnections();
+
       telemetry.timeStep(AppInitTimingsStep.RECEIVED_USER_DATA);
 
-      const connections = await connectionRepository.getConnections();
       telemetry.addStatistic(AppInitStatisticsValue.CONNECTIONS, connections.length, 50);
 
       const conversations = await conversationRepository.loadConversations();
 
-      await userRepository.loadUsers(selfUser, connections, conversations, teamMembers);
+      const contacts = await userRepository.loadUsers(selfUser, connections, conversations);
+      if (selfUser.teamId) {
+        await teamRepository.initTeam(selfUser.teamId, contacts);
+      }
 
       if (supportsMLS()) {
         //if mls is supported, we need to initialize the callbacks (they are used when decrypting messages)
--- a/src/script/search/SearchRepository.ts
+++ b/src/script/search/SearchRepository.ts
@@ -178,11 +178,11 @@ export class SearchRepository {
    * @note We skip a few results as connection changes need a while to reflect on the backend.
    *
    * @param query Search query
-   * @param isHandle Is query a user handle
+   * @param teamId Current team ID the selfUser is in (will help prioritize results)
    * @param maxResults Maximum number of results
    * @returns Resolves with the search results
    */
-  async searchByName(term: string, maxResults = CONFIG.MAX_SEARCH_RESULTS): Promise<User[]> {
+  async searchByName(term: string, teamId = '', maxResults = CONFIG.MAX_SEARCH_RESULTS): Promise<User[]> {
     const {query, isHandleQuery} = this.normalizeQuery(term);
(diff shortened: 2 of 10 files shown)


src/script/main/app.ts
```diff
@@ -415,15 +415,18 @@ export class App {
...
-      const connections = await connectionRepository.getConnections();
       telemetry.addStatistic(AppInitStatisticsValue.CONNECTIONS, connections.length, 50);
 
       const conversations = await conversationRepository.loadConversations();
 
-      await userRepository.loadUsers(selfUser, connections, conversations, teamMembers);
+      const contacts = await userRepository.loadUsers(selfUser, connections, conversations);
+      if (selfUser.teamId) {
+        await teamRepository.initTeam(selfUser.teamId, contacts);
+      }
```
atomrc, line 430: We have to change the order at which things happene. 
Before, we loaded the team users and the connected users at the same time. 

Now the idea is to load all the users we interacted (or are connected) with first and then sort out which ones are team members later on

src/script/search/SearchRepository.ts
```diff
@@ -178,11 +178,11 @@ export class SearchRepository {
    * @note We skip a few results as connection changes need a while to reflect on the backend.
    *
    * @param query Search query
-   * @param isHandle Is query a user handle
+   * @param teamId Current team ID the selfUser is in (will help prioritize results)
    * @param maxResults Maximum number of results
    * @returns Resolves with the search results
    */
-  async searchByName(term: string, maxResults = CONFIG.MAX_SEARCH_RESULTS): Promise<User[]> {
+  async searchByName(term: string, teamId = '', maxResults = CONFIG.MAX_SEARCH_RESULTS): Promise<User[]> {
```
atomrc, line 185: Now that we have less team members in the local state, we need to sort remote search results to have the current team members first

src/script/team/TeamRepository.ts
```diff
@@ -127,19 +126,25 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
     await this.updateFeatureConfig();
     if (!teamId) {
-      return {team: undefined, members: []};
+      return undefined;
     }
-    const members = await this.loadTeamMembers(team);
+    await this.updateTeamMembersByIds(
+      team,
+      contacts.filter(user => user.teamId === teamId).map(({id}) => id),
+    );
```
atomrc, line 144: We now sort out the team members from the other users from the list of contacts. We avoid downloading the entire team

src/script/team/TeamRepository.ts
```diff
@@ -127,19 +126,25 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
+   * @param teamId the Id of the team to init
+   * @param contacts all the contacts the self user has, team members will be deduced from it.
+   */
+  async initTeam(teamId: string, contacts: User[] = []): Promise<TeamEntity | undefined> {
     const team = await this.getTeam();
     // get the fresh feature config from backend
     await this.updateFeatureConfig();
     if (!teamId) {
-      return {team: undefined, members: []};
+      return undefined;
```
arjita-mitra, line 139: why are we changing the return type? is it because now we are only returning team not members?

src/script/team/TeamRepository.ts
```diff
@@ -127,19 +126,25 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
     }
-    const members = await this.loadTeamMembers(team);
+    await this.updateTeamMembersByIds(
```
atomrc, line 141: This will also fix https://wearezeta.atlassian.net/browse/WPB-1912

src/script/team/TeamRepository.ts
atomrc: Oh that's because we do not need the members anymore from the consuming side :)

arjita-mitra (approved):

PatrykBuniX (approved):
Looks good 👍

PR #16164, opened 2023-11-08
feat: Prepare core e2ei service method

przemvs:
## Description

Update E2EIdentity functionality and write tests.

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/E2EIdentity/E2EIdentity.test.ts
+++ b/src/script/E2EIdentity/E2EIdentity.test.ts
@@ -30,6 +30,19 @@ import {E2EIHandler, E2EIHandlerStep} from './E2EIdentity';
 import {getModalOptions, ModalType} from './Modals';
 import {OIDCService} from './OIDCService/OIDCService';
 
+const CoreSingletonMock = {
+  enrollE2EI: jest.fn(),
+  service: {
+    e2eIdentity: {
+      isEnrollmentInProgress: jest.fn(),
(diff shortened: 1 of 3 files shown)


src/script/E2EIdentity/E2EIdentity.test.ts
```diff
@@ -30,6 +30,19 @@ import {E2EIHandler, E2EIHandlerStep} from './E2EIdentity';
...
+  service: {
+    e2eIdentity: {
+      isEnrollmentInProgress: jest.fn(),
+      clearAllProgress: jest.fn(),
+      hasActiveCertificate: jest.fn(),
+      getCertificateData: jest.fn(),
+      getUserDeviceEntities: jest.fn(),
+    },
+  },
+};
```
PatrykBuniX: I don't think mocking this one here is necessary. We are mocking the core in the core mock file already https://github.com/wireapp/wire-webapp/pull/16164/files#diff-c742f60c718dc2a769d90aa8ff070b14e7d5dc80dce9b63584e60e1657ea4812. We should just spy on core methods instead e.g. `enrollE2EI`.

atomrc: I actually think most of those mocks are actually not needed. This test file can be cleaned up

atomrc (approved):

PR #16166, opened 2023-11-08
fix: automatically map user roles when team user list changes [WPB-1912]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-1912" title="WPB-1912" target="_blank">BugWPB-1912</a>  [web] "External" label is not consistently loaded
  </td></table>
  <br />
 

## Description

We previously had multiple sources of truth to know who are the team members:
- `team.members()`
- `teamState.teamMembers()` 

The `team.members` is actually just a manually maintained copy of the users that are part of the team. It doesn't make sense to maintain this list just in order to update users' roles when the list changes. 

Instead, this PR, makes the `teamState.teamMembers` the single source of truth to know all the users that are part of the same team as the self user. 
This way we can safely subscribe to the changes of this list and then update the roles of all the users in the list to get the always up-to-date info for the UI. 

### Before 

https://github.com/wireapp/wire-webapp/assets/1090716/27e77bfc-597a-401c-b741-6295c61140ea

### After

https://github.com/wireapp/wire-webapp/assets/1090716/70020e47-cb82-4d36-b09a-fbd1531718e7

## Checklist

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -415,7 +415,7 @@ export class App {
       onProgress(10);
       telemetry.timeStep(AppInitTimingsStep.INITIALIZED_CRYPTOGRAPHY);
 
-      const {members: teamMembers} = await teamRepository.initTeam(selfUser.teamId);
+      const teamMembers = await teamRepository.initTeam(selfUser.teamId);
       telemetry.timeStep(AppInitTimingsStep.RECEIVED_USER_DATA);
 
       const connections = await connectionRepository.getConnections();
--- a/src/script/team/TeamEntity.ts
+++ b/src/script/team/TeamEntity.ts
@@ -20,7 +20,6 @@
 import ko from 'knockout';
 
 import {AssetRemoteData} from '../assets/AssetRemoteData';
-import type {User} from '../entity/User';
 import {assetV3} from '../util/ValidationUtil';
 
 export class TeamEntity {
@@ -30,14 +29,12 @@ export class TeamEntity {
   /** Team icon (asset key) */
   iconKey?: string;
   id?: string;
-  members: ko.ObservableArray<User>;
   name: ko.Observable<string>;
 
   constructor(id?: string) {
     this.creator = undefined;
     this.icon = '';
     this.iconKey = undefined;
-    this.members = ko.observableArray<User>([]);
     this.id = id;
     this.name = ko.observable('');
   }
@@ -46,7 +43,7 @@ export class TeamEntity {
     let hasIcon = false;
 
     try {
-      hasIcon = this.icon && assetV3(this.icon);
+      hasIcon = !!this.icon && assetV3(this.icon);
     } catch (error) {}
 
     if (hasIcon) {
--- a/src/script/team/TeamRepository.ts
+++ b/src/script/team/TeamRepository.ts
@@ -109,16 +109,16 @@ export class TeamRepository extends TypedEventEmitter<Events> {
     amplify.subscribe(WebAppEvents.TEAM.UPDATE_INFO, this.sendAccountInfo.bind(this));
   }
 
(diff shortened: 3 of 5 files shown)


src/script/main/app.ts
```diff
@@ -415,7 +415,7 @@ export class App {
       onProgress(10);
       telemetry.timeStep(AppInitTimingsStep.INITIALIZED_CRYPTOGRAPHY);
 
-      const {members: teamMembers} = await teamRepository.initTeam(selfUser.teamId);
+      const teamMembers = await teamRepository.initTeam(selfUser.teamId);
```
atomrc, line 418: We never needed the team entity at this point

src/script/team/TeamEntity.ts
```diff
@@ -30,14 +29,12 @@ export class TeamEntity {
   /** Team icon (asset key) */
   iconKey?: string;
   id?: string;
-  members: ko.ObservableArray<User>;
```
atomrc, line 33: This was the other source of truth for "all the team members"

src/script/team/TeamRepository.ts
```diff
@@ -102,14 +103,6 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
 
     this.userRepository.getTeamMembersFromUsers = this.getTeamMembersFromUsers;
-    this.teamState.teamMembers.subscribe(() => this.userRepository.mapGuestStatus());
-
-    this.isSelfConnectedTo = userId => {
-      return (
-        this.teamState.memberRoles()[userId] !== ROLE.PARTNER ||
-        this.teamState.memberInviters()[userId] === this.userState.self().id
-      );
-    };
```
atomrc, line 106: This was duplicated, for some reasons 🤔

src/script/team/TeamRepository.ts
```diff
@@ -102,14 +103,6 @@ export class TeamRepository extends TypedEventEmitter<Events> {
     this.userRepository = userRepository;
 
     this.userRepository.getTeamMembersFromUsers = this.getTeamMembersFromUsers;
-    this.teamState.teamMembers.subscribe(() => this.userRepository.mapGuestStatus());
```
atomrc, line 105: The subscription is now done when the team is init (which means that public users will never have to subscribe)

src/script/team/TeamRepository.ts
```diff
@@ -127,19 +120,27 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
+    this.teamState.teamMembers.subscribe(members => {
+      // Subscribe to team members change and update the user role and guest status
+      this.userRepository.mapGuestStatus(members);
+      const roles = this.teamState.memberRoles();
+      members.forEach(user => {
+        if (roles[user.id]) {
+          user.teamRole(roles[user.id]);
+        }
+      });
+    });
```
atomrc, line 139: This is where the magic happens. Instead of manually re-setting the roles of each entity when events happen, we can just subscribe to changes to the user list and then update the roles accordingly. 
We don't need to hook anymore to events

src/script/team/TeamRepository.ts
```diff
@@ -337,39 +338,21 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
-    );
 
-    if (append) {
-      const knownUserIds = teamEntity.members().map(({id}) => id);
-      const newUserEntities = userEntities.filter(({id}) => !knownUserIds.includes(id));
-      teamEntity.members.push(...newUserEntities);
-    } else {
-      teamEntity.members(userEntities);
-    }
-    this.updateMemberRoles(teamEntity, mappedMembers);
```
atomrc, line 351: No need to maintain this list anymore

PatrykBuniX (approved):

V-Gira (approved):

PR #16171, opened 2023-11-10
feat: apply new design to connection modal (WPB-4802)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4802" title="WPB-4802" target="_blank">TaskWPB-4802</a>  [Web] Connection request modal redesign
  </td></table>
  <br />
 

## Description

Implement new design with a cancel button when a single action is shown in the user modal.
If multiple actions are available, a list is shown as before, even in a modal.
If a single action is available, a list is shown if the user is viewed from the right sidebar

Before:
image

After:
image

## Checklist


src/script/components/panel/UserActions.tsx
```diff
@@ -310,7 +313,11 @@ const UserActions: React.FC<UserActionsProps> = ({
     removeUserFromConversation,
   ].filter((item): item is MenuItem => !!item);
 
-  return <PanelActions items={items} />;
+  return items.length === 1 && isModal ? (
+    <SingleAction item={items[0]} onCancel={onAction} />
```
V-Gira, line 317: the `onAction` method passed here is the `hide` method from the `UserModal` component, it closes the modal

src/script/components/panel/SingleAction/SingleAction.tsx
```diff
@@ -0,0 +1,51 @@
...
+export interface MenuItem {
+  click: () => void;
+  icon: string;
+  identifier: string;
+  label: string;
+}
+
+export interface SingleActionProps {
+  item: MenuItem;
+  onCancel: (action: any) => void;
```
arjita-mitra, line 33: can't we be more specific with the action type?

atomrc: I think it doesn't even need arguments, isn't it @V-Gira ?

atomrc (approved):

src/script/components/panel/SingleAction/SingleAction.tsx
V-Gira: so we pass `() => void` to UserActions [here](https://github.com/wireapp/wire-webapp/blob/virgile/user-actions-modal/src/script/components/Modals/UserModal/UserModal.tsx#L221) from `UserModal`
Then `onAction` gets typed to `(actions: Actions) => void` [here](https://github.com/wireapp/wire-webapp/pull/16171/files#diff-f8053f2d1c4626a4daa1de18c4b8a9c3c9ba1fafa766a2de19d7cf26ad9e358eR75)

If I remove the argument from `SingleAction.tsx` it would complain that I can't pass `(actions: Actions) => void`  to `() => void` :upside_down_face: 

I could add `do-nothing` as an action type or pass `{onAction as any}` [here](https://github.com/wireapp/wire-webapp/pull/16171/files#diff-f8053f2d1c4626a4daa1de18c4b8a9c3c9ba1fafa766a2de19d7cf26ad9e358eR317) instead

src/script/components/panel/SingleAction/SingleAction.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const SingleAction: React.FC<SingleActionProps> = ({item, onCancel}) => {
```
przemvs: If I good remember we shouldn't use FC<T>, instead of this do this:

```suggestion
const SingleAction = ({item, onCancel}: SingleActionProps) => {
```

PR #16216, opened 2023-11-13
refactor: move all mls conference logic to calling repo

PatrykBuniX:
## Description

Moves all the mls-conference related logic from `CallingViewModel` (that should be responsible for view logic only and should not be aware of mls) to `CallingRepository`

## Checklist


atomrc (approved):
Makes a lot of sense 👌

V-Gira (approved):

PR #16221, opened 2023-11-16
feat: rely on CoreCrypto to get device full identity

atomrc:
## Description

This completely relies on the future format of `CoreCrypto` for a device identity. It will contain certificate validation + thumbprint information. 
No need to compute this on our end. 

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentity.ts
+++ b/src/script/E2EIdentity/E2EIdentity.ts
@@ -21,14 +21,16 @@ import {QualifiedId} from '@wireapp/api-client/lib/user';
 import {WireIdentity} from '@wireapp/core/lib/messagingProtocols/mls';
 import {container} from 'tsyringe';
 
+import {MLSStatuses} from 'Components/VerificationBadge';
 import {Core} from 'src/script/service/CoreSingleton';
+import {getCertificateState} from 'Util/certificateDetails';
 import {base64ToArray, supportsMLS} from 'Util/util';
 import {createUuid} from 'Util/uuid';
 
 import {Config} from '../Config';
 
 export type TMP_DecoratedWireIdentity = WireIdentity & {
-  state: 'verified' | 'unverified';
+  state: MLSStatuses;
   thumbprint: string;
 };
 
@@ -50,7 +52,14 @@ export async function getDeviceIdentity(
   deviceId: string,
 ): Promise<TMP_DecoratedWireIdentity | undefined> {
   const identities = await getE2EIdentityService().getUserDeviceEntities(groupId, {[deviceId]: userId});
-  return identities?.length ? {...identities[0], state: 'verified', thumbprint: createUuid()} : undefined;
+  if (identities?.length) {
+    const extraData = {
+      state: getCertificateState(identities[0].certificate),
+      thumbprint: createUuid(),
+    };
+    return {...identities[0], ...extraData};
+  }
+  return undefined;
 }
 
 export async function getConversationState(groupId: string) {
--- a/src/script/E2EIdentity/E2EIdentity.ts
+++ b/src/script/E2EIdentity/E2EIdentity.ts
@@ -18,47 +18,48 @@
  */
 
 import {QualifiedId} from '@wireapp/api-client/lib/user';
+import {WireIdentity} from '@wireapp/core/lib/messagingProtocols/mls';
 import {container} from 'tsyringe';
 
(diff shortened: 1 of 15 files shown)


src/script/E2EIdentity/E2EIdentity.ts
```diff
@@ -18,47 +18,48 @@
...
-import {base64ToArray} from 'Util/util';
+import {base64ToArray, supportsMLS} from 'Util/util';
+import {createUuid} from 'Util/uuid';
+
+import {Config} from '../Config';
+
+export type TMP_DecoratedWireIdentity = WireIdentity & {
+  state: 'verified' | 'unverified';
+  thumbprint: string;
+};
```
atomrc: This is a temporary type that, loosely, matches what CoreCrypto is going to return us in the future

src/script/E2EIdentity/E2EIdentity.ts
```diff
@@ -18,47 +18,48 @@
...
 
 export function getE2EIdentityService() {
-  return container.resolve(Core).service?.e2eIdentity;
+  const e2eIdentityService = container.resolve(Core).service?.e2eIdentity;
+  if (!e2eIdentityService) {
+    throw new Error('trying to query E2EIdentity data in an non-e2eidentity environment');
+  }
```
atomrc, line 41: We want to crash loud and clear when trying to query coreCrypto for e2ei info when not in a e2ei context

src/script/page/MainContent/panels/preferences/DevicesPreferences/DevicesPreference.tsx
```diff
@@ -73,33 +72,20 @@ export const DevicesPreferences: React.FC<DevicesPreferencesProps> = ({
...
+        }
+        return e2eIdentity.getDeviceIdentity(selfConversation.groupId, selfUser.qualifiedId, deviceId);
+      }
+    : undefined;
 
   if (selectedDevice) {
     return (
       <DeviceDetailsPreferences
-        getCertificate={getDeviceCertificate}
-        renderDeviceBadges={renderDeviceBadges}
```
atomrc, line 102: No need for this renderProp anymore. We now feed the `getDeviceIdentity` to the `<DeviceVerificationBadges>` component

src/script/E2EIdentity/E2EIdentity.ts
przemvs: We have similar function in E2EIdentityEnrollment now :) 
image

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -86,29 +85,22 @@ export const UserVerificationBadges = ({user, groupId}: {user: User; groupId?: s
...
     }
   });
 
-  return (
-    <VerificationBadges
-      isProteusVerified={device.meta?.isVerified?.() ?? false}
-      MLSStatus={isMLSVerified ? MLSStatuses.VALID : undefined}
-    />
-  );
+  return <VerificationBadges isProteusVerified={device.meta?.isVerified?.() ?? false} MLSStatus={MLSStatus} />;
```
przemvs: WDYT?

```suggestion
  return <VerificationBadges isProteusVerified={!!device.meta?.isVerified?.()} MLSStatus={MLSStatus} />;
```

src/script/util/certificateDetails.ts
```diff
@@ -33,35 +33,16 @@ const checkExpirationDate = (notAfter: Date) => {
   return currentDate > dateBeforeEnd;
 };
 
-export const getCertificateDetails = (certificate?: string) => {
+export const getCertificateState = (certificate?: string) => {
+  if (!certificate) {
+    return MLSStatuses.NOT_DOWNLOADED;
+  }
```
przemvs, line 39: You can fix also after early return checking certificate in functions, because after early return we will have all the time access to certificate ;)

atomrc: fixed https://github.com/wireapp/wire-webapp/pull/16221/commits/83061f0a74a1303a87a716e826f5f5d3b63706c5

src/script/E2EIdentity/E2EIdentity.ts
atomrc: yeah, true, but it would create a cyclic dependency if we injected one into the other. Also the error message is different and I think that makes sense

przemvs (approved):

PR #16225, opened 2023-11-16
runfix: tweak media queries in call ui

V-Gira:
## Description

The vertical breakpoints were intended for dealing with issues with large font sizes and high level of zoom.
They're not necessary if we don't show labels anymore

#### Before
Peek 2023-11-16 16-39

#### After
Peek 2023-11-16 16-32

### DIFF
--- a/src/style/foundation/video-calling.less
+++ b/src/style/foundation/video-calling.less
@@ -50,10 +50,6 @@
   // fade in animation
   animation: video-fade-in @animation-timing-slower @ease-out-quart forwards;
   opacity: 0;
-
-  @media (max-height: @screen-height-sm) {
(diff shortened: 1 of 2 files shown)


src/style/foundation/video-calling.less
```diff
@@ -180,7 +163,7 @@
     cursor: pointer;
     line-height: 0;
 
-    @media (max-height: @screen-height-sm) {
+    @media (max-width: @call-ui-xs) {
       margin: 0 5px;
     }
```
V-Gira, line 168: The vertical media-query wasn't doing much, but reducing the margin for the smallest width breakdown allow to reach the target 320px

atomrc (approved):

PR #16247, opened 2023-11-20
fix: do not hide conversation when its content is cleared

PatrykBuniX:
## Description

According to spec, conversation should not be gone from the conversations list after its content gets cleared. "Clear content" button should only clear the contents and do nothing more.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -731,19 +731,19 @@ export class ConversationRepository {
       }
       const addCreationMessage = !conversationEntity.hasCreationMessage;
       if (addCreationMessage) {
-        this.addCreationMessage(conversationEntity, this.userState.self().isTemporaryGuest());
+        await this.addCreationMessage(conversationEntity, this.userState.self().isTemporaryGuest());
       }
     }
 
     return mappedMessageEntities;
   }
 
-  private addCreationMessage(
+  private async addCreationMessage(
     conversationEntity: Conversation,
     isTemporaryGuest: boolean,
     timestamp?: number,
     eventSource?: EventSource,
-  ): void {
+  ): Promise<void> {
     conversationEntity.hasCreationMessage = true;
 
     if (conversationEntity.inTeam()) {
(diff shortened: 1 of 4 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2134,24 +2134,15 @@ export class ConversationRepository {
...
-
+  public async clearConversationContent(conversation: Conversation) {
     await this.messageRepository.updateClearedTimestamp(conversation);
-    await this._clearConversation(conversation);
-
-    if (isActiveConversation) {
-      amplify.publish(WebAppEvents.CONVERSATION.SHOW, nextConversationEntity, {});
-    }
+    await this.deleteMessages(conversation);
+    return this.addCreationMessage(conversation, !!this.userState.self()?.isTemporaryGuest());
```
PatrykBuniX: We still want the creation message to be in the conversation.

src/script/conversation/ConversationRepository.ts
```diff
@@ -3710,6 +3700,7 @@ export class ConversationRepository {
     conversationEntity.hasCreationMessage = false;
 
     const iso_date = timestamp ? new Date(timestamp).toISOString() : undefined;
+    conversationEntity.removeMessages();
```
PatrykBuniX, line 3700: Removing messages also from the local state (not only from the indexedDB)

src/script/conversation/ConversationState.ts
```diff
@@ -84,7 +84,6 @@ export class ConversationState {
     this.visibleConversations = ko.pureComputed(() => {
       return this.sortedConversations().filter(
         conversation =>
-          !conversation.is_cleared() &&
```
PatrykBuniX, line 87: From this point we don't want to hide cleared conversations.

V-Gira (approved):

src/script/conversation/ConversationRepository.ts
```diff
@@ -2463,7 +2453,7 @@ export class ConversationRepository {
   /**
    * Clears conversation content from view and the database.
    *
-   * @param conversationEntity Conversation entity to delete
+   * @param conversationEntity Conversation entity to clear
    * @param timestamp Optional timestamps for which messages to remove
    */
   private async _clearConversation(conversationEntity: Conversation, timestamp?: number) {
```
atomrc: Shouldn't this one also be adjusted as well?

PatrykBuniX: Good catch, I've ended up removing this method and creating `onConversationContentCleared` that is fired when we receive member update event in self-conversation with information that conversation was cleared, so all other self clients also do the "clear". https://github.com/wireapp/wire-webapp/pull/16247/commits/16cee22b6e3014c11bd048e59313065e4902b291

atomrc (approved):

PR #16248, opened 2023-11-20
fix: show confirm modal when calling a conversation from wrapper top bar menu (WPB-5445)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5445" title="WPB-5445" target="_blank">TaskWPB-5445</a>  [Web] Prevent accidental calls when chosen from top bar menu
  </td></table>
  <br />
 

## Description
clicking on "Call" in the top bar menu of the electron wrapper, while in a group with 4 or more members, will now be prompted with a confirmation modal. This aligns with the behavior already present in the web app when initiating calls within group conversations.

## Checklist


PatrykBuniX:
Hmm, seems like already have this logic defined in CallingViewModel, see https://github.com/wireapp/wire-webapp/blob/fix/clear-content-hides-conversation/src/script/view_model/CallingViewModel.ts#L186. Since we want to check the number of participants every time we try to start a call, maybe this logic could be moved to `CallingRepository.startCall`? WDYT?

src/script/calling/CallingRepository.ts
```diff
@@ -683,16 +685,39 @@ export class CallingRepository {
...
+            text: t('modalConfirmSecondary'),
+          },
+          text: {
+            htmlMessage: `<div class="modal-description">
+            ${t('groupCallConfirmationModalTitle', memberCount)}
+          </div>`,
+            closeBtnLabel: t('groupCallModalCloseBtnLabel'),
+          },
+        });
+      } else {
```
atomrc: This code must exist somewhere else. We should make sure that the calling button AND the desktop shortcut both go through the exact same code!

Also modals triggering should belong in the view layer (either react component or old view models), not in repositories

atomrc (changes requested):

src/script/calling/CallingRepository.ts
thisisamir98: @atomrc I changed the implementation to handle all of the logic in calling view model, could you have another look?

atomrc (approved):

src/script/view_model/CallingViewModel.ts
```diff
@@ -81,7 +83,7 @@ declare global {
     setSinkId?: (sinkId: string) => Promise<void>;
   }
 }
-const maxGroupSize = 4;
+const MAX_USERS_TO_CALL_WITHOUT_CONFIRM = 1;
```
V-Gira:
```suggestion
const MAX_USERS_TO_CALL_WITHOUT_CONFIRM = 4;
```

Not against having the modal show up every time, but 1 is probably a little low :D

PatrykBuniX (approved):

src/script/view_model/CallingViewModel.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/16248/commits/c45196d1e882235a49cc69987d8cf9ed2b5f422f

V-Gira (approved):

PR #16252, opened 2023-11-21
runfix: cleared conversation visible

PatrykBuniX:
## Description

We don't want cleared conversations to disappear or be deleted from the local store if the self user was removed from the group and/or the group was cleared.

Removing a conversation from the list (and from local database) should be an explicit action, clearing or leaving (or both) a conversation should not make it disappear.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2669,11 +2669,9 @@ export class ConversationRepository {
         conversationEntity =>
           this.reactToConversationEvent(conversationEntity, eventJson, eventSource) as Promise<EntityObject>,
       )
-      .then((entityObject = {} as EntityObject) => {
-        if (type !== CONVERSATION_EVENT.MEMBER_JOIN && type !== CONVERSATION_EVENT.MEMBER_LEAVE) {
-          this.handleConversationNotification(entityObject as EntityObject, eventSource);
-        }
-      })
+      .then((entityObject = {} as EntityObject) =>
+        this.handleConversationNotification(entityObject as EntityObject, eventSource, type),
+      )
       .catch((error: BaseError) => {
         const ignoredErrorTypes: string[] = [
           ConversationError.TYPE.MESSAGE_NOT_FOUND,
@@ -2908,10 +2906,23 @@ export class ConversationRepository {
    * @param eventSource Source of event
    * @returns Resolves when the conversation was updated
    */
-  private async handleConversationNotification(entityObject: EntityObject, eventSource: EventSource) {
+  private async handleConversationNotification(
+    entityObject: EntityObject,
+    eventSource: EventSource,
+    eventType: CLIENT_CONVERSATION_EVENT | CONVERSATION_EVENT,
+  ) {
     const {conversationEntity, messageEntity} = entityObject;
 
-    if (conversationEntity) {
+    if (!conversationEntity) {
+      return;
+    }
+
+    const eventsToSkip: (CLIENT_CONVERSATION_EVENT | CONVERSATION_EVENT)[] = [
+      CONVERSATION_EVENT.MEMBER_JOIN,
+      CONVERSATION_EVENT.MEMBER_LEAVE,
+    ];
+
+    if (!eventsToSkip.includes(eventType)) {
       const eventFromWebSocket = eventSource === EventRepository.SOURCE.WEB_SOCKET;
       const eventFromStream = eventSource === EventRepository.SOURCE.STREAM;
 
@@ -2925,11 +2936,11 @@ export class ConversationRepository {
         if (!eventFromStream) {
           amplify.publish(WebAppEvents.NOTIFICATION.NOTIFY, messageEntity, undefined, conversationEntity);
         }
-
-        if (conversationEntity.is_cleared()) {
-          conversationEntity.cleared_timestamp(0);
-        }
       }
+    }
+
+    if (conversationEntity.is_cleared()) {
+      conversationEntity.cleared_timestamp(0);
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -371,24 +371,6 @@ export class ConversationRepository {
...
-        conversationEntity.isGroup() &&
-        conversationEntity.is_cleared() &&
-        conversationEntity.removed_from_conversation()
-      ) {
-        this.conversationService.deleteConversationFromDb(conversationEntity.id);
-        this.deleteConversationFromRepository(conversationEntity);
-      }
-    });
-
-    this.cleanupEphemeralMessages();
```
PatrykBuniX, line 389: We don't want cleared conversations we're not a member of to be automatically deleted when app is loaded.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2152,9 +2134,10 @@ export class ConversationRepository {
    * @param conversation Conversation to clear content from
    * @param timestamp Timestamp of the event
    */
-  private async clearConversationContent(conversation: Conversation, timestamp?: number) {
+  private async clearConversationContent(conversation: Conversation, timestamp: number) {
     await this.deleteMessages(conversation, timestamp);
-    return this.addCreationMessage(conversation, !!this.userState.self()?.isTemporaryGuest(), timestamp);
+    await this.addCreationMessage(conversation, !!this.userState.self()?.isTemporaryGuest(), timestamp);
+    conversation.setTimestamp(timestamp, Conversation.TIMESTAMP_TYPE.CLEARED);
```
PatrykBuniX, line 2140: After inserting the creation message we update conversation cleared timestamp, so we know the conversation is cleared even though, the creation message is still there.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2924,12 +2918,12 @@ export class ConversationRepository {
...
-
-        if (conversationEntity.is_cleared()) {
-          conversationEntity.cleared_timestamp(0);
-        }
       }
     }
+
+    if (conversationEntity.is_cleared()) {
+      conversationEntity.cleared_timestamp(0);
+    }
```
PatrykBuniX, line 2926: Every we receive any conversation in the conversation we reset cleared_timestamp to 0 so we allow the end user to clear it again.

src/script/conversation/ConversationRepository.ts
```diff
@@ -3207,27 +3199,23 @@ export class ConversationRepository {
       await this.conversationRoleRepository.updateConversationRoles(conversationEntity);
     }
 
-    if (!selfLeavingClearedConversation) {
```
PatrykBuniX, line 3210: We do inject leaving a conversation event when user leaves a conversation after clearing the content beforehand (as two separate operations).

src/script/conversation/ConversationRepository.ts
```diff
@@ -3289,7 +3277,7 @@ export class ConversationRepository {
       await this.clearConversationContent(conversationEntity, conversationEntity.cleared_timestamp());
     }
 
-    if (isActiveConversation && (conversationEntity.is_archived() || conversationEntity.is_cleared())) {
+    if (isActiveConversation && conversationEntity.is_archived()) {
```
PatrykBuniX, line 3280: We still allow user to see cleared conversation, we do not redirect them to next conversation.

src/script/conversation/ConversationState.ts
```diff
@@ -108,17 +108,9 @@ export class ConversationState {
...
-        const isRemoved = conversationEntity.removed_from_conversation();
-
-        if (
-          isSelfConversation(conversationEntity) ||
-          states_to_filter.includes(conversationEntity.connection().status())
-        ) {
-          return false;
-        }
-
-        return !(isCleared && isRemoved);
```
PatrykBuniX, line 121: We don't filter out cleared conversations we're not member of anymore.

src/script/view_model/ContentViewModel.ts
```diff
@@ -271,9 +271,6 @@ export class ContentViewModel {
         this.conversationState.activeConversation(conversationEntity);
       }
 
-      if (conversationEntity.is_cleared()) {
-        conversationEntity.cleared_timestamp(0);
-      }
```
PatrykBuniX, line 276: We do not mark conversation as not cleared on conversation open event. Only event appearing in the conversation should do that.

src/script/conversation/ConversationRepository.ts
```diff
@@ -2140,9 +2122,9 @@ export class ConversationRepository {
    * @param conversation Conversation to clear content from
    * @param timestamp Timestamp of the event
    */
-  public async clearConversation(conversation: Conversation, timestamp?: number) {
+  public async clearConversation(conversation: Conversation) {
     await this.messageRepository.updateClearedTimestamp(conversation);
-    return this.clearConversationContent(conversation, timestamp);
+    return this.clearConversationContent(conversation, new Date().getTime());
```
atomrc, line 2127: actually, can you check something for me?

Since here we send the `clearConversation` event through the wire to all the devices, does the current device also receives it? In this case we might not even need to call `this.clearConversationContent` at this point (since the injected event will trigger the right function calls). 

Am I right?

src/script/conversation/ConversationRepository.ts
```diff
@@ -2668,11 +2651,9 @@ export class ConversationRepository {
         conversationEntity =>
           this.reactToConversationEvent(conversationEntity, eventJson, eventSource) as Promise<EntityObject>,
       )
-      .then((entityObject = {} as EntityObject) => {
-        if (type !== CONVERSATION_EVENT.MEMBER_JOIN && type !== CONVERSATION_EVENT.MEMBER_LEAVE) {
-          this.handleConversationNotification(entityObject as EntityObject, eventSource);
-        }
-      })
+      .then((entityObject = {} as EntityObject) =>
+        this.handleConversationNotification(entityObject as EntityObject, eventSource, type),
```
atomrc, line 2655: can you give some details on this change? I'm not sure I'm following

atomrc: Ah, nevermind, the logic was moved to the underlying function, good call :)

atomrc (approved):
LGTM 😎

src/script/conversation/ConversationRepository.ts
PatrykBuniX: Just double checked, no the self client (client that sends the message) does not receive the message. Self client is the one that sends the message in a self-conversation (proteus self or mls self or both), so it's not targeted as a recipient. We need to run the method manually for the self client.

PR #16253, opened 2023-11-21
runfix: Correctly show user/devices verification state

atomrc:
This is the last bit to having properly verification state computed for users and devices. 

It now uses CoreCrypto 1.0.0-rc19 which brings some corrections and new apis.

### DIFF
--- a/src/script/components/UserDevices.tsx
+++ b/src/script/components/UserDevices.tsx
@@ -20,7 +20,6 @@
 import React, {useEffect, useMemo, useState} from 'react';
 
 import {ClientClassification} from '@wireapp/api-client/lib/client/';
-import {QualifiedId} from '@wireapp/api-client/lib/user';
 
 import {partition} from 'Util/ArrayUtil';
 import {t} from 'Util/LocalizerUtil';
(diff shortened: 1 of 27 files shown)


src/script/components/UserDevices.tsx
```diff
@@ -81,14 +80,13 @@ const sortUserDevices = (devices: ClientEntity[]): ClientEntity[] => {
 
 interface UserDevicesProps {
   clientRepository: ClientRepository;
-  conversationState?: ConversationState;
   cryptographyRepository: CryptographyRepository;
   current: UserDevicesHistoryEntry;
   goTo: (state: UserDevicesState, headline: string) => void;
-  renderDeviceBadges?: (device: ClientEntity, userId: QualifiedId) => React.ReactNode;
```
atomrc, line 88: no more `render props` here, we pass on the `groupId` so that the `WireIdentity` can be fetched from `coreCrypto` directly.

PatrykBuniX (approved):
Looks good 👌

przemvs (approved):

PR #16280, opened 2023-11-27
chore: delete unused user client attributes (WPB-5648)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5648" title="WPB-5648" target="_blank">TaskWPB-5648</a>  [Web] Remove user client activation coordinates from project
  </td></table>
  <br />
 

## Description
This PR deletes some ClientEntity attributes that have not be used in the app for a long time.

## Checklist

### DIFF
--- a/src/script/client/ClientEntity.ts
+++ b/src/script/client/ClientEntity.ts
@@ -34,14 +34,11 @@ export class ClientEntity {
   address?: string;
   class: ClientClassification | '?';
   cookie?: string;
-  domain?: string;
+  domain?: string | null;
   id: string;
   isSelfClient: boolean;
   label?: string;
-  location?: {
-    lat?: number;
-    lon?: number;
-  };
+
   meta: {
     isVerified?: ko.Observable<boolean>;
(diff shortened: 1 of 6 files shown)


atomrc (approved):

src/script/client/ClientEntity.ts
```diff
@@ -34,14 +34,11 @@ export class ClientEntity {
   address?: string;
   class: ClientClassification | '?';
   cookie?: string;
-  domain?: string;
+  domain?: string | null;
```
PatrykBuniX, line 37: **question**: do we really expect it to be `undefined | null | string`?

PatrykBuniX (approved):

src/script/client/ClientEntity.ts
thisisamir98: yes, that's how it is being passed in the constructor

PatrykBuniX: would that make sense (and would it be easy enough to do) to simplify it to either `null` or `undefined`?

PR #16285, opened 2023-11-28
fix: prevent typing in conversation while in fullscreen call-ui (WPB-5044)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5044" title="WPB-5044" target="_blank">BugWPB-5044</a>  Typing while in fullscreen call remembers the last conversation you had opened
  </td></table>
  <br />
 

## Description

It was possible to type in the current conversation when in a fullscreen call.

The util that prevents focus outside off fullscreen does not call `preventDefault` when pressing a key other than tab.

Calling `preventDefault` before calling the `preventFocusOutside` util, excluding the enter key, prevents typing in fullscreen while allowing keyboard navigation (the `Tab` key is handled directly by `preventFocusOutside`)

## Checklist

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -31,7 +31,7 @@ import {ConversationClassifiedBar} from 'Components/input/ClassifiedBar';
 import {isMediaDevice} from 'src/script/guards/MediaDevice';
 import {MediaDeviceType} from 'src/script/media/MediaDeviceType';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
-import {KEY} from 'Util/KeyboardUtil';
+import {isEnterKey, KEY} from 'Util/KeyboardUtil';
 import {t} from 'Util/LocalizerUtil';
 import {preventFocusOutside} from 'Util/util';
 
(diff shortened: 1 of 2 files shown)


src/script/util/util.ts
```diff
@@ -267,6 +267,7 @@ const focusableElementsSelector =
 
 export const preventFocusOutside = (event: KeyboardEvent, parentId: string): void => {
   if (!isTabKey(event)) {
+    event.preventDefault();
     return;
   }
   event.preventDefault();
```
atomrc: how about moving the `event.preventDefault()` before the `if` and avoid this duplicate?

V-Gira: duh :upside_down_face:

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -265,6 +265,9 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
 
   useEffect(() => {
     const onKeyDown = (event: KeyboardEvent): void => {
+      if (!isEnterKey(event)) {
+        event.preventDefault();
+      }
```
V-Gira, line 270: Excluding the `Enter` key to allow selection of call actions

atomrc (approved):

PR #16292, opened 2023-11-29
runfix: Do not crash when self mls conversation does not exist

atomrc:


src/script/page/MainContent/panels/preferences/AccountPreferences.tsx
```diff
@@ -140,7 +140,7 @@ export const AccountPreferences = ({
             {name}
           </h3>
 
-          <UserVerificationBadges user={selfUser} groupId={conversationState.getSelfMLSConversation()?.groupId} />
+          <UserVerificationBadges user={selfUser} groupId={conversationState.selfMLSConversation()?.groupId} />
```
atomrc, line 143: The `getSelfMLSConversation` throws an error if not available. No need for such big gun for that :)

PatrykBuniX (approved):

PR #16293, opened 2023-11-29
runfix: join mls groups one by one

PatrykBuniX:
## Description

Some initial bits of preventing 420 errors (too many requests) from backend. Instead of trying to join all mls groups at once, we join the one by one. Next step will be to add exponential backoff algorithm to api requests that are being used when joining with external commit (`GET: /groupinfo` and `POST: /commit-bundles`).

## Checklist

### DIFF
--- a/src/script/mls/MLSConversations.ts
+++ b/src/script/mls/MLSConversations.ts
@@ -50,25 +50,24 @@ export async function initMLSConversations(
 
   const mlsConversations = conversations.filter(isMLSCapableConversation);
 
-  await Promise.allSettled(
-    mlsConversations.map(async mlsConversation => {
-      const {groupId, qualifiedId} = mlsConversation;
+  for (const mlsConversation of mlsConversations) {
+    const {groupId, qualifiedId} = mlsConversation;
 
-      const doesMLSGroupExist = await conversationService.mlsGroupExistsLocally(groupId);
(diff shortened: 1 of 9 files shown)


src/script/mls/MLSConversations.ts
```diff
@@ -50,25 +50,24 @@ export async function initMLSConversations(
 
   const mlsConversations = conversations.filter(isMLSCapableConversation);
 
-  await Promise.allSettled(
-    mlsConversations.map(async mlsConversation => {
-      const {groupId, qualifiedId} = mlsConversation;
+  for (const mlsConversation of mlsConversations) {
```
atomrc, line 60: I think that behavior-wise, this changes how the joining works. 

From what I can understand:

- previously we would try to join all conversations (some could fail, it wouldn't matter)

but now:

- we try one by one and stop at the first failing conversation. 

Is that correct? 
Is that by design?

PatrykBuniX: That's very valid point, i've added error handling (not to change the behaviour we just ignore error and log it into the console) https://github.com/wireapp/wire-webapp/pull/16293/commits/9b9027d0b0744f14fb7cd2a8bd6c7bf16de9832d

atomrc: nice `onError` callback ❤️

atomrc (approved):

PR #16295, opened 2023-11-29
runfix: allow keyboard navigation in full screen device select (WPB-6075)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6075" title="WPB-6075" target="_blank">BugWPB-6075</a>  [Web] Cannot use keyboard navigation to select devices in call ui
  </td></table>
  <br />
 

## Description

Arrow navigation doesn't work to select a device in the new select in full screen calls

## Checklist

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -546,9 +546,7 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                             // eslint-disable-next-line jsx-a11y/no-autofocus
                             autoFocus
                             value={selectedVideoOptions}
-                            onChange={selectedOption => {
-                              updateVideoOptions(String(selectedOption?.value));
-                            }}
+                            onChange={selectedOption => updateVideoOptions(String(selectedOption?.value))}
                             id="select-camera"
                             dataUieName="select-camera"
                             controlShouldRenderValue={false}
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -465,6 +465,9 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                     {audioOptionsOpen ? (
                       <>
                         <Select
+                          // eslint-disable-next-line jsx-a11y/no-autofocus
+                          autoFocus
+                          onBlur={() => setAudioOptionsOpen(false)}
                           value={selectedAudioOptions}
                           id="select-microphone"
                           dataUieName="select-microphone"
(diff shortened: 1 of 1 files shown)


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -465,6 +470,8 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                     {audioOptionsOpen ? (
                       <>
                         <Select
+                          // eslint-disable-next-line jsx-a11y/no-autofocus
+                          autoFocus
```
V-Gira, line 475: `autoFocus` is needed foe arrow navigation to function in the menu

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -453,8 +455,11 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                   <button
                     className="device-toggle-button"
                     css={audioOptionsOpen ? videoControlActiveStyles : videoControlInActiveStyles}
-                    onClick={() => {
-                      setAudioOptionsOpen(prev => !prev);
+                    onClick={event => {
+                      const target = event.target as Element;
+                      if (!target.closest('#select-microphone')) {
+                        setAudioOptionsOpen(prev => !prev);
+                      }
```
V-Gira: prevents the `onclick` event to trigger in the select menu
It would trigger when selecting an option right after setting the `optionsOpen` state of the menu to false, leaving the menu open

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -238,6 +239,7 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
     const camera = videoOptions[0].options.find(item => item.value === selectedOption) ?? selectedVideoOptions[0];
     setSelectedVideoOptions([camera]);
     switchCameraInput(camera.id);
+    setVideoOptionsOpen(false);
```
V-Gira, line 242: We set the `optionsOpen` state to `false` here in the `updateOptions` method instead of toggling it off on the click event

src/script/components/calling/FullscreenVideoCall.tsx
atomrc: so you cannot toggle the button anymore? How do you close it?

V-Gira: I'll add the code comment to make that clearer
The select menu (`#select-microphone`) is part of the button.
If it's not the target of the click event (when the actual dropdown arrow is clicked), the toggle works normally

V-Gira: does this clear things up?
`// We want to ensure to only toggling the menu open or closed when clicking the icon, not the select menu`

atomrc (approved):
LGTM

PR #16300, opened 2023-11-30
runfix: register mls verification state handler after fetching conversations

PatrykBuniX:
## Description

Verification state handler should be initialised after fetching the conversation list from backend but before processing messages queue. We should also not use `filteredConversations` when trying to find mls group by its groupId since it filters out self mls conversation. There was an unexpected error statement because of that:

Screenshot 2023-11-30 at 13 19 01

## Checklist

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -26,7 +26,7 @@ import {E2EIVerificationMessageType} from 'src/script/message/E2EIVerificationMe
 import {Core} from 'src/script/service/CoreSingleton';
 import {Logger, getLogger} from 'Util/Logger';
 
-import {MLSConversation} from '../../ConversationSelectors';
+import {isMLSConversation, MLSConversation} from '../../ConversationSelectors';
 import {ConversationState} from '../../ConversationState';
 import {ConversationVerificationState} from '../../ConversationVerificationState';
 import {getConversationByGroupId, OnConversationE2EIVerificationStateChange} from '../shared';
@@ -91,6 +91,10 @@ class MLSConversationVerificationStateHandler {
     const conversation = getConversationByGroupId({conversationState: this.conversationState, groupId});
     if (!conversation) {
       this.logger.error(`Epoch changed but conversationEntity can't be found`);
+      return;
+    }
+
+    if (!isMLSConversation(conversation)) {
       return;
     }
 
--- a/src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
@@ -65,7 +65,7 @@ export const getConversationByGroupId = ({
   conversationState,
   groupId,
 }: GetConversationByGroupIdParams): MLSConversation | undefined => {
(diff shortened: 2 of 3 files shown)


src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -64,9 +64,9 @@ interface GetConversationByGroupIdParams {
 export const getConversationByGroupId = ({
   conversationState,
   groupId,
-}: GetConversationByGroupIdParams): MLSConversation | undefined => {
-  const conversation = conversationState.filteredConversations().find(conversation => conversation.groupId === groupId);
-  return conversation && isMLSConversation(conversation) ? conversation : undefined;
+}: GetConversationByGroupIdParams): MLSCapableConversation | undefined => {
+  const conversation = conversationState.conversations().find(conversation => conversation.groupId === groupId);
```
PatrykBuniX, line 68: We don't want to filter self mls conversation out when searching for conversation by groupId (this is what `filteredConversations` does).

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
```diff
@@ -64,9 +64,9 @@ interface GetConversationByGroupIdParams {
...
+  return conversation && isMLSCapableConversation(conversation) ? conversation : undefined;
```
PatrykBuniX, line 69: "Mixed" conversations also have a groupId assigned, we should not ignore them at this point.

src/script/main/app.ts
```diff
@@ -436,6 +432,7 @@ export class App {
       if (supportsMLS()) {
         //if mls is supported, we need to initialize the callbacks (they are used when decrypting messages)
         conversationRepository.initMLSConversationRecoveredListener();
+        registerMLSConversationVerificationStateHandler(this.updateConversationE2EIVerificationState);
```
PatrykBuniX, line 435: Moved handler initialiser to the point when all the conversations are already fetched from backend.

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -94,6 +94,10 @@ class MLSConversationVerificationStateHandler {
       return;
     }
 
+    if (!isMLSConversation(conversation)) {
+      return;
+    }
```
PatrykBuniX, line 99: I don't think we do want to mls-verify mixed conversations, we care only about its proteus verification state until we finalise the migration (protocol changes to mls)

src/script/main/app.ts
aweiss-dev: I don't think moving this handler matters that much since it relies on events from the core to actually do anything.

PatrykBuniX: Don't we need to have the conversation in local state to be able to search it by group id? Now we're sure it will be there, there's no harm i believe

atomrc: It's always good not to subscribe to something if you know you don't need it. It's good to have this `if` (in case that's really what is being debated here)

src/script/conversation/ConversationVerificationStateHandler/shared/conversation/index.ts
atomrc: oh yeah, great catch 👌

atomrc (approved):

PR #16320, opened 2023-12-04
runfix: display error message for mls conversation creation with offline backend (WPB-3695)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3695" title="WPB-3695" target="_blank">StoryWPB-3695</a>  [Web] MLS Conversation creation when a participant is on a remote offline backend
  </td></table>
  <br />
 

## Description
We handle errors during MLS conversation creation correctly but we do not get the name of the unreachable backends in the response
Additionally, we do not display anything if the username is not in memory anymore

#### Before
Screenshot from 2023-12-01 18-05-57
Screenshot from 2023-12-04 15-07-00

#### After
image
image

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
+++ b/src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
@@ -150,7 +150,7 @@ describe('FailedToAddUsersMessage', () => {
     const message = createFailedToAddUsersMessage({
       users: [qualifiedId1, qualifiedId2],
       reason: AddUsersFailureReasons.UNREACHABLE_BACKENDS,
-      backends: ['test-2.domain'],
+      backends: ['test.domain', 'test-2.domain'],
     });
 
     const {getByText, getAllByText} = render(
@@ -171,7 +171,7 @@ describe('FailedToAddUsersMessage', () => {
     const details = getAllByText(
       (_, element) =>
         element?.textContent ===
-        'Arjita and Tom could not be added to the group as the backend of test-2.domain could not be reached.',
+        'Arjita and Tom could not be added to the group as the backend of test.domain, test-2.domain could not be reached.',
     );
 
     expect(details.length).toBeGreaterThanOrEqual(1);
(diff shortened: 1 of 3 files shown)


src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -174,7 +178,7 @@ const FailedToAddUsersMessage: React.FC<FailedToAddUsersMessageProps> = ({
       </div>
       <div className="message-body" css={{flexDirection: 'column'}}>
         {isOpen && (
-          <MessageDetails users={users} reason={message.reason} domains={message.backends}>
+          <MessageDetails users={users} reason={message.reason} domains={users.map(user => user.domain)}>
```
V-Gira, line 180: Getting the unreachable domains in the response only works in proteus, mapping the domain of the user we have in database works for both proteus and mls

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -61,7 +61,11 @@ const MessageDetails = ({users, children, reason, domains}: MessageDetailsProps)
   const baseTranslationKey =
     users.length === 1 ? 'failedToAddParticipantsSingularDetails' : 'failedToAddParticipantsPluralDetails';
 
-  const domainStr = domains.join(', ');
+  const getUsername = (name: string) => (name.length > 0 ? name : t('unavailableUser'));
```
V-Gira: displays "Name not available" instead of an empty string if we don't have the username in database

src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
```diff
@@ -171,7 +171,7 @@ describe('FailedToAddUsersMessage', () => {
     const details = getAllByText(
       (_, element) =>
         element?.textContent ===
-        'Arjita and Tom could not be added to the group as the backend of test-2.domain could not be reached.',
+        'Arjita and Tom could not be added to the group as the backend of test.domain, test-2.domain could not be reached.',
```
V-Gira, line 174: The previous copy wasn't correct :sweat_smile:

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
atomrc: There is a `useUserName` hook that does just that ;)

atomrc: humm maybe the hook is not the right tool there, but you could add a new function in the `UserName.tsx` file that return the name string for a user object (and the hook could use this function)

PatrykBuniX (approved):

PR #16326, opened 2023-12-05
feat: Degraded conversation - warnings

przemvs:
## Description
Show modals for sending message and initiate call while conversation is degraded.

image

image

## Checklist

- use (x) data
- be sure to look at ...


src/script/components/InputBar/InputBar.tsx
```diff
@@ -348,6 +350,29 @@ export const InputBar = ({
...
+        },
+        secondaryAction: {
+          action: () => {},
+          text: t('conversation.E2EICancel'),
+        },
+        text: {
+          message: t('conversation.E2EIDegradedNewMessage'),
+          title: t('conversation.E2EIConversationNoLongerVerified'),
+        },
+      });
```
atomrc: Are you sure this modal appears only once (and not on every message sending when the conversation is degraded)? 🤔

src/script/view_model/CallingViewModel.ts
```diff
@@ -203,25 +205,60 @@ export class CallingViewModel {
...
-            action: async () => await startCall(conversationEntity, callType),
-            text: t('groupCallModalPrimaryBtnName'),
-          },
-          secondaryAction: {
-            text: t('modalConfirmSecondary'),
+
+      PrimaryModal.show(PrimaryModal.type.CONFIRM, {
+        primaryAction: {
+          action: async () => {
+            if (memberCount > MAX_USERS_TO_CALL_WITHOUT_CONFIRM) {
```
atomrc, line 216: This logic is duplicated here https://github.com/wireapp/wire-webapp/pull/16326/files#diff-97b34e53266556ca5a4b5b9a75e7db3dff9492ae6ef96121764c303eec005a53R260

This needs refactoring, there should only be 1 place where the limitation modal should be triggered

przemvs: Will do it in some follow-up ticket with better modal handling for it.

src/script/components/InputBar/InputBar.tsx
przemvs: You're right :) It should display only once

https://wearezeta.atlassian.net/browse/WPB-767?focusedCommentId=119950

src/script/calling/CallingRepository.ts
```diff
@@ -192,6 +193,40 @@ export class CallingRepository {
...
+            },
+            text: {
+              message: t('conversation.E2EIGroupCallDisconnected'),
+              title: t('conversation.E2EIConversationNoLongerVerified'),
+            },
+          };
+
+          PrimaryModal.show(PrimaryModal.type.ACKNOWLEDGE, modalOptions, `degraded-${call.conversationId}`);
+        }
+      });
```
atomrc: outch, we are subscribing without ever unsuscribing, there is something wrong in the design here. 

We should: 
- `subscribe` when the call starts
- `unsubscribe` when the call ends

atomrc: actually, since you are in a `ko.computed` I don't think you need to `subscribe` and the function will re-run automatically when something changes

przemvs: Ahh... You're right.. 🤦🏼

atomrc (approved):

PR #16327, opened 2023-12-05
runfix: Update self identity verification when enrollment is successful

atomrc:
## Description

This will allow the selfUser's verification state to be recomputed once the enrolment process is done

https://github.com/wireapp/wire-webapp/assets/1090716/e033d6ae-5732-4a24-8972-0ee130b0a84d

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityEnrollment.ts
+++ b/src/script/E2EIdentity/E2EIdentityEnrollment.ts
@@ -90,7 +90,7 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
     E2EIHandler.instance = null;
   }
 
-  public initialize({discoveryUrl, gracePeriodInSeconds}: E2EIHandlerParams): void {
+  public initialize({discoveryUrl, gracePeriodInSeconds}: E2EIHandlerParams) {
     if (isE2EIEnabled()) {
       if (!hasActiveCertificate()) {
         const gracePeriodInMs = gracePeriodInSeconds * TIME_IN_MILLIS.SECOND;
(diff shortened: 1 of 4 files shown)


src/script/E2EIdentity/E2EIdentityEnrollment.ts
```diff
@@ -82,13 +78,8 @@ export class E2EIHandler {
...
    */
-  public static getInstance(params?: E2EIHandlerParams) {
-    if (!E2EIHandler.instance) {
-      if (!params) {
-        throw new Error('GracePeriodTimer is not initialized. Please call getInstance with params.');
-      }
-      E2EIHandler.instance = new E2EIHandler(params);
-    }
+  public static getInstance() {
+    E2EIHandler.instance = E2EIHandler.instance ?? new E2EIHandler();
```
atomrc, line 82: I needed to simplify the API, in order to be able to get the instance while not configure it. 

So now, to enroll, you need to call `getInstance().initialize()` before calling `enroll`

src/script/hooks/useDeviceIdentities.ts
```diff
@@ -17,24 +17,34 @@
...
+    const userIdentities = await getUsersIdentities(groupId, [userId]);
+    setDeviceIdentities(userIdentities.get(userId.id) ?? []);
   }, [userId.id, groupId]);
 
+  useEffect(() => {
+    void refreshDeviceIdentities();
+  }, [refreshDeviceIdentities]);
+
+  useEffect(() => {
+    E2EIHandler.getInstance().on('enrollmentSuccessful', refreshDeviceIdentities);
```
przemvs, line 42: Can we use better typing for 'enrollmentSuccesful' ? Use some enums?

atomrc: it's already typechecked since it extends `TypedEventEmitter` ;) No need for extra code there, we get it for free

przemvs (approved):

PR #16340, opened 2023-12-07
fix: Wait for conversation to be known before triggering verification checks

atomrc:
## Description

This will make sure when we receive an `epochChange` event from the core that the conversation does exist in the webapp as well (and be able to update the verification state of that conversation accordingly) 

## Checklist


PatrykBuniX (approved):
Very elegant solution, good job man!

przemvs (approved):
Yayy! Nice! This is what I thinking about it also :D

PR #16347, opened 2023-12-08
runfix: use proteus as default supported protocol if user'S backend is offline (WPB-5752)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5752" title="WPB-5752" target="_blank">BugWPB-5752</a>  [Web] MLS 1 on 1 conversations do not load if the other user's backend is unreachable
  </td></table>
  <br />
 

## Description

1 on 1 conversations with a user from an unreachable backend never load, even if those conversation already exist locally.
It's because we cannot get a supported protocol for the user and the error from `getUserSupportedProtocols` isn't handled

### Draft

If we catch the error, we can try to get a locally supported protocol for the user, and if that fails, default to proteus

In the worst case scenario of an MLS conv existing locally, no local supported protocol in database (launching the app and the top conv is from an federated user from a currently unreachable backend), the conv would still reinitialize as MLS, as we would recognize the conversation exists here https://github.com/wireapp/wire-webapp/blob/virgile/offline-mls-1on1/src/script/conversation/ConversationRepository.ts#L1701

## Checklist

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -720,12 +720,19 @@ export class UserRepository extends TypedEventEmitter<Events> {
         return localSupportedProtocols;
       }
     }
+    try {
+      const supportedProtocols = await this.userService.getUserSupportedProtocols(userId);
 
-    const supportedProtocols = await this.userService.getUserSupportedProtocols(userId);
-
-    //update local user entity with new supported protocols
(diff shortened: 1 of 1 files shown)


PatrykBuniX:
Closing in favour of https://github.com/wireapp/wire-webapp/pull/16382

PR #16370, opened 2023-12-11
ci: Publish to wire-builds

smatting:
This PR 

- adds a helm chart template at `charts/webapp` (copy of the chart wire-server)
- Updates the "CI" workflows (`test_build_deploy.yml`)
  - remove publishing of docker images and helm charts
  - remove unused step `Upload WebApp artifacts`
  - remove unused authentication step `Authenticate git clone`
- Creates a new "Publish Docker image, Helm chart and Wire build" workflow (`publish.yml`):
  - which only runs on specific branches (not on every PR)
  - builds Docker images and pushes tags with unique names. This happens on every tracked branch (was only on commit labelled "staging" or "production" before)
  - publishes a helm chart with patched image tags to a new helm repo `charts-webapp`
  - bumps the helm chart version https://github.com/wireapp/wire-builds at configurable target branches

Overall this PR should only add new processes, all existing processes (pushing docker images) should remain the same.
The new processes around https://github.com/wireapp/wire-builds are not set up, so this PR doesn't affect anything yet.

Tracked in https://wearezeta.atlassian.net/browse/WPB-5438


bin/semver
atomrc, line 1: Let's see if we could do that with a node package (there might be one already installed)

.github/workflows/test_build_deploy.yml
```diff
@@ -117,27 +138,8 @@ jobs:
             }
 
       - name: Build
-        if: env.prod_env || env.dev_env
```
atomrc, line 120: Let's not remove this condition but rather refine it (if we remove it it will run on every PR we make)

atomrc: I'd suggest, for now at least `env.prod_env || env.dev_env || env.wire_builds_target_branches`)

smatting: We decided to solve the problem by separating workflows

.github/workflows/publish.yml
```diff
@@ -0,0 +1,265 @@
+name: Publish Docker image, Helm chart and Wire build
+
+on:
+  push:
+    branches: [master, dev, edge, avs, mobile, acc, stefan]
```
atomrc:
```suggestion
    branches: [master, dev, edge, stefan]
```

Those are retired envs, now is a good time to sunset them and start fresh :)

bin/push_docker.js
```diff
@@ -36,33 +36,36 @@ require('dotenv').config();
...
+/** Key to  ./app-config/package.json, either
+ *
+ *  "wire-web-config-default-master"
+ *   or
+ *  "wire-web-config-default-staging"
+ * */
+const configurationEntry = process.argv[2];
+/** Version tag of webapp (e.g. "2023-11-09-staging.0", "dev") */
+const versionTag = process.argv[3];
+const uniqueTagOut = process.argv[4] || '';
```
atomrc: we also use this command in `create_docker_image` action. Let's make sure it's not going to break this workflow

bin/semver
atomrc, line 1: Let's add https://www.npmjs.com/package/semver to the repo and use this cli tool rather than copy pasting a script for that. 

You can run `yarn add -D semver` at the root of the repo and then use this script inside of your code :)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,261 @@
...
+      - name: Skip CI
+        if: |
+          contains(env.TITLE || env.PR_LAST_COMMIT_MESSAGE, 'skip ci') ||
+          contains(env.TITLE || env.PR_LAST_COMMIT_MESSAGE, '[ci skip]')
+        uses: andymckay/cancel-action@0.3
+
+      - name: Authenticate git clone
+        env:
+          GH_TOKEN: ${{secrets.OTTO_THE_BOT_GH_TOKEN}}
+        run: echo "machine github.com login ${GH_TOKEN}" > ~/.netrc
```
atomrc: question: do you need to be an authenticated user for this script to run properly? I don't think we need (and actually our other job might not even need it either 🤔 )

bin/push_docker.js
atomrc: hummm it might actually break this other action (since the new input is at the first place)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,261 @@
...
+        run: |
+          set -o pipefail
+          yarn test --coverage --coverage-reporters=lcov --detectOpenHandles=false 2>&1 | tee ./unit-tests.log
+
+      - name: Monitor coverage
+        uses: codecov/codecov-action@v3.1.4
+        with:
+          fail_ci_if_error: false
+          files: ./coverage/lcov.info
+          flags: unittests
```
atomrc:
```suggestion
```
We don't need to monitor the coverage for this action. This will be done in the CI job (for PRs)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,261 @@
...
+        run: |
+          echo "wire_builds_target_branches: $wire_builds_target_branches"
+          echo "wire_builds_target_branches=$wire_builds_target_branches" >> $GITHUB_OUTPUT
+
+      - name: Set TITLE and BUILD_DESKTOP
+        env:
+          PR_TITLE: ${{github.event.pull_request.title || env.PR_LAST_COMMIT_MESSAGE}}
+        run: |
+          echo "TITLE=${PR_TITLE}" >> $GITHUB_ENV
+          echo "BUILD_DESKTOP=${{contains(env.TAG, 'staging') || contains(env.TAG, 'production') || contains(env.PR_LAST_COMMIT_MESSAGE, '+Desktop')}}" >> $GITHUB_ENV
```
atomrc:
```suggestion
      - name: Set TITLE 
        env:
          PR_TITLE: ${{github.event.pull_request.title || env.PR_LAST_COMMIT_MESSAGE}}
        run: echo "TITLE=${PR_TITLE}" >> $GITHUB_ENV
```
You don't need this desktop build here :)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,261 @@
...
+
+      - name: Print environment variables
+        run: |
+          echo -e "BRANCH_NAME = ${BRANCH_NAME}"
+          echo -e "TAG = ${TAG}"
+          echo -e "TITLE = ${TITLE}"
+          echo -e "PR_LAST_COMMIT_MESSAGE = ${PR_LAST_COMMIT_MESSAGE}"
+          echo -e "COMMIT_URL = ${COMMIT_URL}"
+          echo -e "COMMITTER = ${COMMITTER}"
+          echo -e "BUILD_DESKTOP = ${BUILD_DESKTOP}"
```
atomrc:
```suggestion
```

.github/workflows/publish.yml
smatting: Done in f0128a2fc069e442e2fdb919acdbcc23abb5b3d0

bin/semver
smatting: I tried https://classic.yarnpkg.com/en/package/semver but it's missing a few features that we need, namely:
- getting and setting components of semver (the package only supports bumping)
- comparing
- validating

I also checked if there any other cli tools that we could use here https://classic.yarnpkg.com/en/packages?q=semver&p=1 but I couldn't find any. 

The bash script I've copied is from https://github.com/fsaintjacques/semver-tool which seems to be of good quality (has multiple contributers and tests). Let's consider keeping it?

bin/push_docker.js
smatting: Fixed in ef506e860fed7a4e33406c2f304a150563a86005

.github/workflows/publish.yml
smatting: Yeah, it wasn't needed. I removed it from both workflows in 119af02dcf04069e8f4cde240fd8e0872547834c and everything succeeded

.github/workflows/publish.yml
smatting: Removed in fd1d2c8c6daeb79399d18353699efc1d7b86b759

.github/workflows/publish.yml
smatting: fixed in 825fa975c08368600f09b615239f3b0a6f9a64d2

.github/workflows/publish.yml
```diff
@@ -0,0 +1,218 @@
+name: Publish Docker image, Helm chart and Wire build
+on:
+  push:
+    branches: [master, dev, edge, stefan]
```
tlebon: should we remove the stefan branch also if this is finalized? 
```suggestion
    branches: [master, dev, edge]
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,218 @@
...
+
+          if [[ "$version_tag" == *"production"* ]]; then
+            echo "FUTUREWORK: bump some production branch on wire-builds once it exists"
+          fi
+
+          if [[ "$version_tag" == *"staging"* ]]; then
+            echo "FUTUREWORK: bump some cloud staging branch on wire-builds once it exists"
+          fi
+
+          if [ "$version_tag" == "dev" ] || [ "$version_tag" == "stefan" ] ; then
```
tlebon:
```suggestion
          if [ "$version_tag" == "dev" ] ; then
```
same topic

bin/chart-next-version.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+#
```
tlebon, line 19:
```suggestion
/*
 * Wire
 * Copyright (C) 2023 Wire Swiss GmbH
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see http://www.gnu.org/licenses/.
 *
 */
 #
```
should we include legal info here?

tlebon (approved):
just a few minor topics but i think its fine. i wont merge today (unless you comment/merge the suggestions, then its fine). please squash if you merge :)

.github/workflows/publish.yml
smatting: I removed it

bin/chart-next-version.sh
smatting: Why not, I added it

PR #16371, opened 2023-12-12
fix: use updated version of @stablelib/utf8 [WPB-5844]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5844" title="WPB-5844" target="_blank">BugWPB-5844</a>  Encryption fails when unicode emojis or reactions are added
  </td></table>
  <br />
 

Currently, trying to send a reaction to a message with encrypted DB will yield the following error.

image-20231211-150443

The library update will fix this problem for us.

### DIFF
--- a/package.json
+++ b/package.json
@@ -217,7 +217,8 @@
   },
   "resolutions": {
     "libsodium": "0.7.10",
-    "xml2js": "0.5.0"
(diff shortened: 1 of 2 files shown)


tlebon (approved):
🆓 🥟

PR #16374, opened 2023-12-13
feat(E2EI): add proxy to circumvent CSP headers

aweiss-dev:
The IDP we must access to get the OAuth information during the E2EI process is unknown to us. Therefore we don't have the possibility to add the targetURL to our CSP-Settings. 
This combination prevents the browser scripts to directly access the IDP.

This PR adds a proxy on express side (node.js server) to circumvent this issue.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -14,7 +14,7 @@
     "geolite2": "1.3.0",
     "hbs": "4.2.0",
     "helmet": "7.1.0",
-    "http-proxy-middleware": "^2.0.6",
+    "http-proxy-middleware": "2.0.6",
     "http-status-codes": "2.3.0",
     "logdown": "3.3.1",
     "maxmind": "4.3.10",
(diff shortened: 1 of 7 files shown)


server/routes/E2EIProxy/OIDCProxyRoute.ts
```diff
@@ -0,0 +1,111 @@
...
+  const isValidUrl = (urlString: string) => {
+    try {
+      new URL(urlString);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  };
+
+  // @ts-ignore
```
aweiss-dev, line 38: This is fore some 'fake' ts error that not all code paths have a return statement.
Not needed with express

przemvs (approved):

server/package.json
```diff
@@ -14,6 +14,7 @@
     "geolite2": "1.3.0",
     "hbs": "4.2.0",
     "helmet": "7.1.0",
+    "http-proxy-middleware": "^2.0.6",
```
PatrykBuniX: Let's use exact versions

PR #16400, opened 2023-12-19
runfix: import backup from migrated db version [WPB-5906]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5906" title="WPB-5906" target="_blank">BugWPB-5906</a>  [Web] Backup from Prod cannot be restored on Edge
  </td></table>
  <br />
 

## Description

Previously it was not possible to import a backup that was made on some older version of wire if there were some DB schema migration in between that version and the current one. In that case we were just throwing an "incompatible backup version error". 

With this PR, importing a backup made from older version should be possible. After history is imported we will also run schema upgrade scripts to assure that all the upgrades that should have taken place during db migration while updating wire were made.

## Checklist


tlebon (approved):

atomrc (approved):
Nice 👏

PR #16405, opened 2023-12-20
fix: use local users if they cannot be loaded from BE [WPB-5959]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5959" title="WPB-5959" target="_blank">BugWPB-5959</a>  [Web] 1:1 MLS conversation disappears after page refresh when of user's backend goes offline
  </td></table>
  <br />
 

## Description

When initially loading a list of users, it can happen that one of the federated backends is offline (unreachable), therefore we will not receive fresh user data from the server. It's possible that we've already fetched this user before, thus we can try looking it up in local db (indexedDB) before just returning a placeholder for "unavailable user". 

## Checklist

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -220,13 +220,13 @@ export class UserRepository extends TypedEventEmitter<Events> {
     /* prior to April 2023, we were only storing the availability in the DB, we need to refetch those users */
     const [localUsers, incompleteUsers] = partition(dbUsers, user => !!user.qualified_id);
 
-    /** users we have in the DB that are not matching any loaded users */
+    // We can remove users that are not linked to any "known users" from the local database.
+    // Known users are the users that are part of the conversations or the connections (or some extra users)
     const orphanUsers = localUsers.filter(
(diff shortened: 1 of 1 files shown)


src/script/user/UserRepository.ts
```diff
@@ -592,12 +592,28 @@ export class UserRepository extends TypedEventEmitter<Events> {
...
+    const failedToLoad = failed.map(userId => {
+      // When a federated backend is unreachable, we try to load a user from the local database.
+      const dbUserRecord = dbUsers?.find(user => matchQualifiedIds(user.qualified_id, userId));
+
+      if (dbUserRecord && selfUser) {
+        return this.userMapper.mapUserFromJson(dbUserRecord, selfDomain);
+      }
+
+      // Otherwise, we generate placeholder users locally with some default values.
+      return new User(userId.id, userId.domain);
```
PatrykBuniX, line 613: Before returning a placeholder for "unavailable user" we try looking it up in local indexedDB because there's a chance that we've already saved this user before.

atomrc (approved):
ohhh that's a pretty nice solution 😎 🚀 
nice job man

PR #16413, opened 2023-12-21
runfix: revert htmlMessage removal from PrimaryModal

tlebon:
## Description
Reverting removal of dangerouslySetInnerHtml as a number of modals will do string interpolation from strings that have manually set links or other custom html in them. 

## Checklist


src/script/components/Modals/PrimaryModal/PrimaryModalTypes.ts
```diff
@@ -79,6 +79,7 @@ export interface ModalContent {
   currentType: string | PrimaryModalType;
   inputPlaceholder: string;
   message: React.ReactNode;
+  messageHtml?: string;
```
atomrc, line 83: why do we have both now? what is the diff between `htmlMessage` and `messageHtml`?

atomrc, line 83:
```suggestion
  /** @deprecated please use `message` instead */
  messageHtml?: string;
```

tlebon: we had both before as well. depends through which flow the modal is initiated.

atomrc: let's just use `htmlMessage` from now on (and deprecate it)

atomrc (approved):

PR #16434, opened 2023-12-28
runfix: establishing mls 1to1 between team members

PatrykBuniX:
## Description

According to spec, team-owned 1:1 MLS conversations should not be established automatically, it should happen only right before a user tries to send a message in a chat. We want to get rid of the behaviour where 1:1 conversation appeared at the top of the list if we were searched by somebody else.

To change this behaviour we're no longer establishing 1:1 mls conversations with two team members every time when 1:1 is initialised. Before every message is sent, we make sure mls 1:1 conversation is established. If it's not yet, we will first establish it and then send a message in a conversation.

Before:
before

After:
after

## Checklist


atomrc (approved):
LGTM 😎

PR #16467, opened 2024-01-03
feat: retreive current device certificate from CoreCrypto

aweiss-dev:
This PR changes the location where we get the current device certificate from localStorage to CoreCrypto. CC is our single point of truth regarding Identities and Certificates.

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityVerification.ts
+++ b/src/script/E2EIdentity/E2EIdentityVerification.ts
@@ -92,14 +92,15 @@ const fetchSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
 };
 
 export async function hasActiveCertificate(): Promise<boolean> {
-  return Boolean((await getActiveCertificate()).length);
+  const activeCertificate = await getActiveCertificate();
+  return typeof activeCertificate === 'string' && Boolean(activeCertificate.length);
 }
 
-export async function getActiveCertificate(): Promise<string> {
+export async function getActiveCertificate(): Promise<string | undefined> {
   const selfDeviceIdentity = await fetchSelfDeviceIdentity();
 
   if (!selfDeviceIdentity) {
-    return '';
+    return undefined;
   }
 
   return selfDeviceIdentity.certificate;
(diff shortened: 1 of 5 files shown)


src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -73,8 +74,35 @@ export async function getConversationVerificationState(groupId: string) {
...
+
+export async function hasActiveCertificate(): Promise<boolean> {
+  return Boolean((await getActiveCertificate()).length);
+}
+
+export async function getActiveCertificate(): Promise<string> {
+  const selfDeviceIdentity = await fetchSelfDeviceIdentity();
+
+  if (!selfDeviceIdentity) {
+    return '';
```
PatrykBuniX: **question:** Why do we need to return an empty string? Wouldn't it be more type-safe if we return `null` or `undefined` instead?

src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -73,8 +74,35 @@ export async function getConversationVerificationState(groupId: string) {
...
-  return getE2EIdentityService().hasActiveCertificate();
+const fetchSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
+  const conversationState = container.resolve(ConversationState);
+  const selfMLSConversation = conversationState.getSelfMLSConversation();
+  const userIdentities = await getUsersIdentities(
+    selfMLSConversation.groupId,
+    selfMLSConversation.allUserEntities().map(user => user.qualifiedId),
+  );
+  const currentClientId = selfMLSConversation.selfUser()?.localClient?.id;
+  const userId = selfMLSConversation.selfUser()?.id;
```
atomrc, line 85: Hummm really not ideal in term of architecture that such a deep component is tied to the application state. 
The self conversation should be injected by higher level components

src/script/E2EIdentity/E2EIdentityVerification.ts
aweiss-dev: Yeah, I can see that. Initially, I just didn't want to alter the current logic.

atomrc (approved):

PatrykBuniX (approved):

PR #16471, opened 2024-01-04
chore: Stop processing team events (WPB-4538)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4538" title="WPB-4538" target="_blank">EpicWPB-4538</a>  [Platform] Stop processing team events
  </td></table>
  <br />
 

The following team events are deprecated on backend and we need to skip processing them, expect the `team.member-leave` which we need to keep for backward compatibility.

team.member-join
team.member-leave
team.update
team.member-update


test/helper/EventGenerator.ts
```diff
@@ -68,6 +72,24 @@ export function createReactionEvent(targetMessageId: string, reaction: string =
...
 
+export function createMemberLeaveEvent(conversationId: string, userIds: string[]): MemberLeaveEvent {
+  const conversationQualifiedId = {id: conversationId, domain: ''};
+
+  return {
+    conversation: conversationId,
+    qualified_conversation: conversationQualifiedId,
+    data: {
+      qualified_user_ids: userIds.map(userId => ({id: userId, domain: ''})),
+      reason: MemberLeaveReason.USER_DELETED,
```
atomrc, line 83: I think we are missing the part that displays the `x was removed from the team` when this value is set on the event. Am I missing something?

thisisamir98: fixed in https://github.com/wireapp/wire-webapp/pull/16471/commits/3edd9bc30e76f711cfa6ee3274e035c0e8e0ef50

yarn.lock
```diff
@@ -8771,12 +8771,12 @@ __metadata:
   linkType: hard
 
 "follow-redirects@npm:^1.15.4":
-  version: 1.15.4
-  resolution: "follow-redirects@npm:1.15.4"
+  version: 1.15.5
```
atomrc, line 8819: This change is suspicious 🤔

src/script/event/preprocessor/EventStorageMiddleware/EventStorageMiddleware.ts
```diff
@@ -73,9 +80,37 @@ export class EventStorageMiddleware implements EventMiddleware {
   }
 
   private validateEvent(event: HandledEvents, duplicateEvent?: EventRecord) {
+    if (event.type === CONVERSATION_EVENT.MEMBER_LEAVE) {
```
atomrc, line 83: Let's add some comment for future reference

```suggestion
    if (event.type === CONVERSATION_EVENT.MEMBER_LEAVE) {
    /* When we receive a `member-leave` event, we should check that the user is actually still part of the conversation before forwarding the event. If the user is already not part of the conversation, then we can throw a validation error (that means the user was already removed by another member-leave event) */
    
```

atomrc (approved):

PR #16476, opened 2024-01-05
runfix: don't hide dropdown if clicked inside of it

PatrykBuniX:
## Description

It was not possible to change audio/video device using dropdown menu in fullscreen calling view. The reason behind that is `onBlur` event was fired before onChange event.

Solution for that is not to hide audio/video menu when clicked inside of the dropdown menu.

## Checklist

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -456,7 +456,11 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                     onClick={() => {
                       setAudioOptionsOpen(prev => !prev);
                     }}
-                    onBlur={() => setAudioOptionsOpen(false)}
+                    onBlur={event => {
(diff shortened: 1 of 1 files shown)


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -456,7 +456,11 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
                     onClick={() => {
                       setAudioOptionsOpen(prev => !prev);
                     }}
-                    onBlur={() => setAudioOptionsOpen(false)}
+                    onBlur={event => {
+                      if (!event.currentTarget.contains(event.relatedTarget)) {
```
PatrykBuniX, line 460: In case of blur event, the `relatedTarget` property is an element that gained focus after it was removed from some other element.

arjita-mitra (approved):

przemvs (approved):

PR #16490, opened 2024-01-08
fix: duplicated link preview message [WPB-5728]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5728" title="WPB-5728" target="_blank">BugWPB-5728</a>  [Web] Duplicate message with link preview
  </td></table>
  <br />
 

## Description

It was possible that a message containing a link preview would get sent twice if it was edited when still greyed out (in a sending state). Reason behind that is our current implementation of link preview messages. If we detect that a message contains a link and we are able to generate a preview for it, we send another message (this time with an actual preview) that will replace the initial message.

Why was "link preview" message sent twice?

1. send an initial message, with id `A` (its stuck in a sending state for a while)
2. we edit the message, so another message with id `B` is being sent, message `A` got replaced with message `B`
3. `A` message was sent successfully, we generate a preview message - `A1` for it and try to replace a message `A` with it
4. since message `A` was already replaced by `B`, it's not possible to replace `A` with `A1`, `A1` is sent as a new message
5. `B` message was sent successfully, we generate a preview message - `B1` and replace `B` with it.

So the actual issue is in point 4. above.

## Solution

The solution is to maintain a little state in a memory responsible for keeping track of handling sending a link preview message. We do not allow to generate a link preview for a message, which id exists in a state. When editing a message, we lock handling the link preview for a message we are editing (original message).

What happens now:
1. send message `A`
2. edit message `A`, it will be replaced with `A1`, lock handling a preview for message `A`
3. message `A` was sent, we try to handle link preview, but it was locked, we do not generate a preview for `A`
4. edit message was sent successfully, we generate a preview for `A1`, it was never locked, we create a preview

## Checklist

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -392,18 +392,23 @@ export class MessageRepository {
       );
     }
 
+    const originalMessageId = originalMessage.id;
     const messagePayload = {
       conversation,
       mentions,
       message: textMessage,
       messageId: createUuid(), // We set the id explicitely in order to be able to override the message if we generate a link preview
-      originalMessageId: originalMessage.id,
+      originalMessageId,
     };
 
-    cancelSendingLinkPreview(originalMessage.id);
-    const {state} = await this.sendEdit(messagePayload);
-    if (state !== MessageSendingState.CANCELED) {
-      await this.handleLinkPreview(messagePayload);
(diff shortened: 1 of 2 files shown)


przemvs (approved):

atomrc (approved):
Brilliant 👏

PR #16498, opened 2024-01-09
runfix: Enroll to E2EI before joining conversations

atomrc:
## Description

We currently have a problem in the enrollement flow:
- the user logs in
- a MLS device will be created with basic credentials keyPackages
- we load all the conversations
- we join all the conversation via external commits
- we finally trigger the enrollment flow

The problem is we already joined the conversation before we had the chance to upgrade the credentials of the user with certified ones. 

The new flow goes as follow:
- the user logs in
- a MLS device is create with basic credentials keyPackages
- we trigger the enrollement process (if needed)
- we wait for the user to have a valid certificate
- then we load all the conversations
- and we join the conversations

So now we directly join converstaions with valid x509 certificates

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityVerification.ts
+++ b/src/script/E2EIdentity/E2EIdentityVerification.ts
@@ -95,11 +95,7 @@ const fetchSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
 };
 
 export async function hasActiveCertificate(): Promise<boolean> {
-  const identity = await getActiveWireIdentity();
-  if (!identity?.certificate) {
-    return false;
-  }
-  return typeof identity.certificate === 'string' && Boolean(identity.certificate.length);
+  return getE2EIdentityService().isE2EIEnabled();
 }
 
 export async function getActiveWireIdentity(): Promise<WireIdentity | undefined> {
--- a/src/script/E2EIdentity/E2EIdentityVerification.ts
+++ b/src/script/E2EIdentity/E2EIdentityVerification.ts
@@ -53,9 +53,6 @@ export function isE2EIEnabled(): boolean {
 }
 
 export async function getUsersIdentities(groupId: string, userIds: QualifiedId[]) {
-  if (await isFreshMLSSelfClient()) {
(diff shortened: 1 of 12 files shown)


src/script/components/AppContainer/AppContainer.tsx
```diff
@@ -49,6 +51,7 @@ export const AppContainer: FC<AppProps> = ({config, clientType}) => {
   window.wire.app = app;
   const mainView = new MainViewModel(app.repository);
   useTheme(() => app.repository.properties.getPreference(PROPERTIES_TYPE.INTERFACE.THEME));
+  useAccentColor();
```
atomrc, line 54: This will set the correct class to the `body` so that the modals are styled with the user's accent color

src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -92,11 +92,7 @@ const fetchSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
 };
 
 export async function hasActiveCertificate(): Promise<boolean> {
-  const identity = await getActiveWireIdentity();
-  if (!identity?.certificate) {
-    return false;
-  }
-  return typeof identity.certificate === 'string' && Boolean(identity.certificate.length);
+  return getE2EIdentityService().isE2EIEnabled();
```
przemvs, line 96: Are You sure? hasaActiveCertificate and You returning isE2EIEnabled? So maybe rename it :)

atomrc: yes, this is the name of the method on CoreCrypto side that tells us if the user is enrolled. 
This logic is correct :)

przemvs: So maybe please add comment there about it :D

przemvs (approved):

src/script/E2EIdentity/E2EIdentityVerification.ts
atomrc: sure, makes sense 👌

PR #16503, opened 2024-01-10
runfix: stop counting e2ei system messages as unread [WPB-6081]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6081" title="WPB-6081" target="_blank">BugWPB-6081</a>  [Web] E2EI system message creates regular unread message notification
  </td></table>
  <br />
 

## Description

We don't want to include e2ei verification system messages in unread messages count.

## Checklist

### DIFF
--- a/src/script/entity/Conversation.ts
+++ b/src/script/entity/Conversation.ts
@@ -45,7 +45,6 @@ import {CallMessage} from './message/CallMessage';
 import type {ContentMessage} from './message/ContentMessage';
 import type {Message} from './message/Message';
 import {PingMessage} from './message/PingMessage';
-import {SystemMessage} from './message/SystemMessage';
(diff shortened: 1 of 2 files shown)


src/script/entity/Conversation.ts
```diff
@@ -72,7 +71,7 @@ interface UnreadState {
   pings: PingMessage[];
   selfMentions: ContentMessage[];
   selfReplies: ContentMessage[];
-  systemMessages: SystemMessage[];
```
atomrc, line 75: humm this is probably a mistake. The `E2EIVerificationMessage` should be a `SystemMessage`. Maybe this is the only change needed here 🤔

atomrc (approved):

PR #16505, opened 2024-01-10
feat: Encrypt enrollment oauth stored data [WPB-6034]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6034" title="WPB-6034" target="_blank">BugWPB-6034</a>  [web] use safe storage to store oauth data 
  </td></table>
  <br />
 

## Description

This will encrypt the data stored for oauth (like access token or refresh token). 

For encryption we use the key that is generate by the core that is given back to the webapp. 

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityEnrollment.ts
+++ b/src/script/E2EIdentity/E2EIdentityEnrollment.ts
@@ -72,7 +72,7 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
   private readonly userState = container.resolve(UserState);
   private config?: EnrollmentConfig;
   private currentStep: E2EIHandlerStep | null = E2EIHandlerStep.UNINITIALIZED;
-  private oidcService: OIDCService | null = null;
+  private oidcService?: OIDCService;
 
   private get coreE2EIService() {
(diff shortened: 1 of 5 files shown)


src/script/E2EIdentity/E2EIdentityEnrollment.ts
```diff
@@ -84,6 +84,14 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
     return e2eiService;
   }
 
+  private createOIDCService() {
```
aweiss-dev, line 87: Shouldn't we check if this.oidcService already exists and return that instance if available?

PatrykBuniX (approved):

aweiss-dev (approved):

src/script/E2EIdentity/E2EIdentityEnrollment.ts
atomrc: Well, I replaced the `new OIDC` to this new function. 
There might be some problems if we start reusing old instances when we used to create new ones. 

So I went to the safe side an re-created new instances everytime.

PR #16508, opened 2024-01-12
feat: Hide secondary button in soft lock modal

przemvs:
## Description

We don't want to display secondary buttons while we have a fresh device.

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityEnrollment.ts
+++ b/src/script/E2EIdentity/E2EIdentityEnrollment.ts
@@ -345,6 +345,7 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
       secondaryActionFn: () => {
         this.showE2EINotificationMessage();
       },
+      hideSecondary: !!this.config?.isFreshMLSSelfClient,
     });
 
     PrimaryModal.show(modalType, modalOptions);
--- a/src/script/E2EIdentity/Modals/Modals.ts
+++ b/src/script/E2EIdentity/Modals/Modals.ts
@@ -114,7 +114,7 @@ export const getModalOptions = ({
         },
       };
       modalType =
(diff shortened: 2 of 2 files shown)


src/script/E2EIdentity/E2EIdentityEnrollment.ts
```diff
@@ -345,6 +345,7 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
       secondaryActionFn: () => {
         this.showE2EINotificationMessage();
       },
+      hideSecondary: !!this.config?.isFreshMLSSelfClient,
```
aweiss-dev, line 348: This is not enough actually.

This button needs also be hidden for renewals, after the end of snooze period

Since we have that case quite often, a helper function that checks for both cases would be great :)

aweiss-dev (changes requested):

src/script/E2EIdentity/E2EIdentityEnrollment.ts
przemvs: Yes, but it will be done in Arjita PR

arjita-mitra: I haven't added that check yet, you can add it as part of this PR

aweiss-dev: Quick reminder that we have 5 cases for soft-lock

1. New Device without certificate, we ignore the snooze period
2. Existing Device without certificate, we soft lock after expired snooze period
3. Existing Device with certificate, we soft lock after expired snooze period for refresh
4. Existing Device with expired certificate, soft lock immediately
4. Revoked Device (not yet implemented)

PR #16536, opened 2024-01-17
refactor: streamline softLock logic

atomrc:
## Description

This harmonises the way we compute if we should enable soft lock or not. 

This changes the following parts:
- harmonising the `enrollmentSuccess` and `identityUpate` events to be a single event that always sends the `config` and `identity`
- add early return to the logic that computes the soft lock situation

## Checklist

### DIFF
--- a/src/script/hooks/useAppSoftLock.ts
+++ b/src/script/hooks/useAppSoftLock.ts
@@ -20,7 +20,7 @@
 import {useEffect, useState} from 'react';
 
 import {CallingRepository} from '../calling/CallingRepository';
-import {E2EIHandler, EnrollmentConfig, isE2EIEnabled, isFreshMLSSelfClient} from '../E2EIdentity';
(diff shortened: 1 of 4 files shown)


src/script/hooks/useAppSoftLock.ts
```diff
@@ -59,21 +59,5 @@ export function useAppSoftLock(callingRepository: CallingRepository, notificatio
...
-
-  useEffect(() => {
-    if (!freshMLSSelfClient) {
-      return () => {};
-    }
-    E2EIHandler.getInstance().on('enrollmentSuccessful', checkIfIsFreshMLSSelfClient);
-    return () => {
-      E2EIHandler.getInstance().off('enrollmentSuccessful', checkIfIsFreshMLSSelfClient);
-    };
-  }, [freshMLSSelfClient]);
```
atomrc, line 76: This logic has been factored in the handling of the `identityUpdate` event

przemvs (approved):

PR #16537, opened 2024-01-17
test: improve api of useAppSoftLock hook

atomrc:
## Checklist


PatrykBuniX (approved):

src/script/components/AppContainer/AppContainer.tsx
```diff
@@ -77,10 +77,7 @@ export const AppContainer: FC<AppProps> = ({config, clientType}) => {
 
   const {repository: repositories} = app;
 
-  const {isFreshMLSSelfClient, softLockLoaded = false} = useAppSoftLock(
-    repositories.calling,
-    repositories.notification,
-  );
+  const {softLockEnabled} = useAppSoftLock(repositories.calling, repositories.notification);
```
atomrc, line 80: We don't need the `loaded` info anymore since now enrollment and renewal happen at load time (which means the app is not loaded)

src/script/hooks/useAppSoflLock.test.ts
```diff
@@ -0,0 +1,98 @@
+/*
```
przemvs: useAppSoflLock => useAppSoftLock :)

przemvs (approved):

PR #16543, opened 2024-01-17
runfix: Re-enable showing enrollment flow when e2ei is activated

atomrc:
## Description

With the latest API changes to the enrollment flow, `initialize` doesn't automatically trigger the enrollment flow. 
We need to manual start the process when the e2ei is enabled on the team config level

## Checklist

### DIFF
--- a/src/script/E2EIdentity/DelayTimer/DelayTimer.ts
+++ b/src/script/E2EIdentity/DelayTimer/DelayTimer.ts
@@ -32,8 +32,7 @@ interface CreateGracePeriodTimerParams {
   delayPeriodExpiredCallback: () => void;
 }
 
-class DelayTimerService {
-  private static instance: DelayTimerService | null = null;
+export class DelayTimerService {
   private gracePeriodInMS: number;
   private gracePeriodExpiredCallback: () => void;
   private delayPeriodExpiredCallback: () => void;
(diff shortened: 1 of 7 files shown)


src/script/E2EIdentity/DelayTimer/DelayTimer.ts
```diff
@@ -32,42 +32,21 @@ interface CreateGracePeriodTimerParams {
...
-   */
-  public static getInstance(params?: CreateGracePeriodTimerParams) {
-    if (!DelayTimerService.instance) {
-      if (!params) {
-        throw new Error('DelayTimerService is not initialized. Please call getInstance with params.');
-      }
-      DelayTimerService.instance = new DelayTimerService(params);
-    }
-    return DelayTimerService.instance;
-  }
```
atomrc, line 69: It was never needed that this class is a singleton. 
It simplifies the API to make it a regular instantiable class

src/script/page/components/FeatureConfigChange/FeatureConfigChangeHandler/FeatureConfigChangeHandler.tsx
```diff
@@ -38,7 +38,7 @@ export function FeatureConfigChangeHandler({teamState}: Props): null {
   useEffect(() => {
     if (config) {
       // initialize feature handlers
-      configureE2EI(logger, config);
+      configureE2EI(logger, config)?.attemptEnrollment();
```
atomrc, line 41: This is the actual fix. When we detect the feature change, we trigger the `attemptEnrollment` right away

PatrykBuniX (approved):
Looks good

PR #16554, opened 2024-01-18
runfix: add e2ei-verification type to expected events

PatrykBuniX:
## Description

We were not expecting to receive conversation.e2ei-service to arrive in a self conversation.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -17,6 +17,7 @@
  *
  */
 
+import {CONVERSATION_TYPE} from '@wireapp/api-client/lib/conversation';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
 import {E2eiConversationState} from '@wireapp/core/lib/messagingProtocols/mls';
 import {container} from 'tsyringe';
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2806,7 +2806,11 @@ export class ConversationRepository {
 
     const inSelfConversation = this.conversationState.isSelfConversation(conversationId);
     if (inSelfConversation) {
-      const typesInSelfConversation = [CONVERSATION_EVENT.MEMBER_UPDATE, ClientEvent.CONVERSATION.MESSAGE_HIDDEN];
+      const typesInSelfConversation = [
+        CONVERSATION_EVENT.MEMBER_UPDATE,
+        ClientEvent.CONVERSATION.MESSAGE_HIDDEN,
+        CLIENT_CONVERSATION_EVENT.E2EI_VERIFICATION,
+      ];
```
atomrc: Question: would it be possible to not inject this event in the selfConversation (upstream)? 
It doesn't really make sense to send this event in the self conv

atomrc (approved):

src/script/conversation/ConversationRepository.ts
PatrykBuniX: Good idea, done with https://github.com/wireapp/wire-webapp/pull/16554/commits/b7fb6026f4e42bdad19e8f0290be5193903825de

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -99,7 +100,8 @@ class MLSConversationVerificationStateHandler {
       return this.logger.warn(`Epoch changed but conversation could not be found after waiting for 5 seconds`);
     }
 
-    if (!isMLSConversation(conversation)) {
+    const isSelfMLSConversation = isMLSConversation(conversation) && conversation.type() === CONVERSATION_TYPE.SELF;
```
atomrc: I don't even think you need to check for `isMLSConversation` at this point. checking for `self` is enough (proteus self conversation will already be filtered out on line 104)

atomrc (approved):

PR #16589, opened 2024-01-22
fix: don't load self convo's messages into memory [WPB-6166]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6166" title="WPB-6166" target="_blank">SpikeWPB-6166</a>  [Web] [4hrs] Investigate possible memory leakages due to backup restore
  </td></table>
  <br />
 

## Description

After importing a backup we are loading all the unread messages into the conversations. It could happen that a user has a lot  of events sent inside their self conversation (type `1`). 

Since self conversation is never appearing in the ui, there's no way for user to read a message in this conversation. We should never load messages of self conversation into the memory.

## Checklist


atomrc (approved):
Looks solid. Great catch man 👏

PR #16675, opened 2024-01-29
runfix: team feature config updates

PatrykBuniX:
## Description

Since we want to deprecate team feature config update events (`"feature-config.update"` type) and use pulling (every 30s) instead, we need a mechanism for detecting particular feature updates.

This PR introduces a `getTeamFeatureUpdate` that based on previous and new state return an information whether the feature:
- got enabled
- got disabled
- remained unchanged
- has changed its config (but remained enabled)

## Checklist


src/script/main/app.ts
```diff
@@ -398,6 +400,15 @@ export class App {
...
       }
 
+      teamRepository.on('featureConfigUpdated', async configUpdate => {
+        const {type} = getTeamFeatureUpdate(configUpdate, FEATURE_KEY.MLSE2EID);
+
+        if (type !== FeatureUpdateType.UNCHANGED) {
+          const client = await configureE2EI(this.logger, configUpdate.newFeatureList);
+          return client?.attemptEnrollment();
+        }
+      });
```
atomrc: Humm this would probably be better somewhere in a hook that will have an unmount cleanup. 

Also it doesn't really make sense to subscribe before the app is fully loaded. 

So I'd suggest isolating this to a hook that is initialize by the `AppMain` component. WDYT?

PatrykBuniX: Great idea 👌

src/script/team/TeamFeatureUpdater/TeamFeatureUpdater.ts
```diff
@@ -0,0 +1,126 @@
...
+  {prevFeatureList, newFeatureList}: {prevFeatureList?: FeatureList; newFeatureList?: FeatureList},
+  key: Key,
+): FeatureUpdate<Key> => {
+  const newFeature = newFeatureList?.[key];
+
+  if (!prevFeatureList) {
+    // Feature was added and is enabled
+    if (newFeature && newFeature.status === FeatureStatus.ENABLED) {
+      return {type: FeatureUpdateType.ENABLED, next: newFeature};
+    }
```
atomrc: aren't we missing the `disabled` case here?  (maybe we need some mapping?)

```suggestion
    if (newFeature) {
      return {type: newFeature.status, next: newFeature};
    }
```

atomrc: Oh maybe when `newFeature` is empty, that means disabled, right?

src/script/team/TeamRepository.ts
```diff
@@ -383,18 +383,14 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
 
-  private readonly onFeatureConfigUpdate = async (
-    event: TeamFeatureConfigurationUpdateEvent,
-    source: EventSource,
-  ): Promise<void> => {
+  private readonly onFeatureConfigUpdate = async (source: EventSource): Promise<void> => {
     if (source !== EventSource.WEBSOCKET) {
       // Ignore notification stream events
       return;
     }
```
atomrc: I think we can remove this piece of code now. No config update is going to come from the WS anymore 😢

src/script/team/TeamFeatureUpdater/TeamFeatureUpdater.ts
```diff
@@ -0,0 +1,126 @@
...
+  next?: FeatureList[Key];
+};
+
+type FeatureUpdate<Key extends FEATURE_KEY> =
+  | FeatureUpdateEnabled<Key>
+  | FeatureUpdateDisabled<Key>
+  | FeatureUpdateConfigChanged<Key>
+  | FeatureUpdateUnchanged<Key>;
+
+export const getTeamFeatureUpdate = <Key extends FEATURE_KEY>(
```
atomrc: That pretty clean and neat API 👏 
Love it

src/script/team/TeamFeatureUpdater/TeamFeatureUpdater.ts
PatrykBuniX: Correct, in this case, when the feature has never existed in the list, even if it appears, so `newFeature` is truthy, but the feature status is disabled, we still don't care about this feature.

src/script/main/app.ts
PatrykBuniX: Addressed with https://github.com/wireapp/wire-webapp/pull/16675/commits/637e9fdee3f27b64f1e7fdd042e12f5863cf73ca

src/script/team/TeamFeatureUpdater/TeamFeatureUpdater.ts
atomrc: suggestion: Since this class is mostly doing diffing between the previous config and the new one, I might suggest a name change: How about `TeamFeatureConfigChangeDetector` (or something along those lines).

src/script/team/TeamRepository.ts
PatrykBuniX: I don't think we need the handler at all now, removed it https://github.com/wireapp/wire-webapp/pull/16675/commits/c1570a44b15045831bf5ba06fc10d477cc6100f0 ;)

src/script/team/TeamFeatureUpdater/TeamFeatureUpdater.ts
PatrykBuniX: Addressed with [287ee04](https://github.com/wireapp/wire-webapp/pull/16675/commits/287ee04de9652be1da5aee66310d516dff6919c3) 😉

atomrc (approved):

PR #16684, opened 2024-01-29
runfix: open 1to1 offline [WPB-6306]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6306" title="WPB-6306" target="_blank">BugWPB-6306</a>  [Web] Unable to open 1:1 conversation when being offline
  </td></table>
  <br />
 

### Problem

It was not possible to open a 1:1 conversation with the other user when being offline. The reason behind that is before opening a 1:1 conversation we were fetching other user's supported protocols from backend and then based on that initialising proteus or mls 1:1 conversation. When being offline, backend was not reachable, we were stuck in retrying the request, 1:1 conversation could open after a long delay or not open at all.

### Solution

Instead of refetching user's supported protocols before opening the conversation, we first initialise it using local supported protocols (the ones saved on a locally stored user entity) and then refetch it from backend in the background. If they were updated, we will emit an event which will re-init the conversation.

## Checklist

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -720,28 +720,32 @@ export class UserRepository extends TypedEventEmitter<Events> {
    * @param userId - the user to fetch the supported protocols for
    */
   private async refreshUserSupportedProtocols(userId: QualifiedId): Promise<void> {
-    const localUser = this.findUserById(userId);
+    try {
+      const localUser = this.findUserById(userId);
 
-    if (!localUser) {
-      return;
+      if (!localUser) {
+        return;
+      }
+
+      const localSupportedProtocols = localUser.supportedProtocols();
+      const supportedProtocols = await this.userService.getUserSupportedProtocols(userId);
+
+      const haveSupportedProtocolsChanged =
+        !localSupportedProtocols ||
+        !(
+          localSupportedProtocols.length === supportedProtocols.length &&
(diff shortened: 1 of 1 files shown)


przemvs (approved):

src/script/user/UserRepository.ts
```diff
@@ -715,16 +715,56 @@ export class UserRepository extends TypedEventEmitter<Events> {
...
+  public async getUserSupportedProtocols(
+    userId: QualifiedId,
+    shouldRefreshUser = false,
+  ): Promise<ConversationProtocol[]> {
     const localSupportedProtocols = this.findUserById(userId)?.supportedProtocols();
 
-    if (!forceRefetch && localSupportedProtocols) {
+    // Refresh user supported protocols in the background
+    if (shouldRefreshUser) {
+      void this.refreshUserSupportedProtocols(userId);
```
atomrc: suggestion: you could pass the entire user entity here (and avoid re looking it up in the state in the `refreshUserSupportedProtocol` method)

atomrc (approved):

src/script/user/UserRepository.ts
atomrc:
```suggestion
      // trigger a refresh of the supported protocols in the background. No need to await for this one 
      void this.refreshUserSupportedProtocols(userId);
```

V-Gira (approved):

atomrc (approved):

PR #16695, opened 2024-01-30
refactor: create grouped data structure for message list [WPB-6282]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6282" title="WPB-6282" target="_blank">TaskWPB-6282</a>  [web] have a grouped data structure for message grouping
  </td></table>
  <br />
 

## Description

In order to have better control of how we render messages in the message list and in order to ease grouping strategies, we want to have a grouped data structure (rather than just a flat list of messages). 

So instead of `Message[]` data type to render messages, we now have a `(Marker|MessageGroup)[]` data type. 

`Marker` represents a time marker (messages from today, or from a particular hour or unread)
`MessageGroup` represents messages  from the same sender that should be grouped together because sent roughly at the same time

## Checklist


src/style/content/conversation/message-list.less
```diff
@@ -244,55 +244,6 @@
   height: 16px;
 }
 
-// MESSAGE - TIMESTAMP
-.message-timestamp {
```
atomrc, line 248: this has been migrated to the `Marker.styles.ts` file

src/script/components/MessagesList/Message/types.ts
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export type ScrollToElement = (elm: {center?: boolean; element: HTMLElement}, isUnread?: boolean) => void;
```
przemvs, line 20: EOL missing

atomrc: It's just Github not showing it, but it's there 
image

przemvs (approved):
🚀

PR #16718, opened 2024-02-01
refactor: Do not base message focus on elements indexes

atomrc:
## Description

This will be needed when we will group messages by timestamp. 

Previously the message list was a flat list of messages. So referring to a particular index in this array, gave the correct message element. 

With the changes from https://github.com/wireapp/wire-webapp/pull/16695 we need a stronger index lookup. 

Instead we will be basing our focus on message ids rather than indexes. 

## Associated bugfix

This also prevents focus from changing when the message list is updated

### Before 

https://github.com/wireapp/wire-webapp/assets/1090716/95a3b46a-d88a-49f9-afb2-f8fe4ac0256f

### After

https://github.com/wireapp/wire-webapp/assets/1090716/0112b5b5-0474-4d79-9ce5-c0225ef03fa5

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/index.tsx
+++ b/src/script/components/MessagesList/Message/index.tsx
@@ -76,10 +76,8 @@ export interface MessageParams extends MessageActions {
   selfId: QualifiedId;
   shouldShowInvitePeople: boolean;
   teamState?: TeamState;
-  totalMessage: number;
-  index: number;
   isMessageFocused: boolean;
-  handleFocus: (index: number) => void;
+  handleFocus: (id: string) => void;
   handleArrowKeyDown: (e: React.KeyboardEvent) => void;
   isMsgElementsFocusable: boolean;
   setMsgElementsFocusable: (isMsgElementsFocusable: boolean) => void;
@@ -95,11 +93,9 @@ const Message: React.FC<
     lastReadTimestamp,
     onVisible,
     scrollTo,
-    totalMessage,
     isMessageFocused,
     handleFocus,
     handleArrowKeyDown,
-    index,
     isMsgElementsFocusable,
     setMsgElementsFocusable,
   } = props;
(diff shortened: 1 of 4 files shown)


src/script/components/MessagesList/Message/index.tsx
```diff
@@ -145,21 +141,6 @@ const Message: React.FC<
...
-    if (history.state?.eventKey === 'Enter') {
-      handleFocus(totalMessage - 1);
-
-      // reset the eventKey to stop focusing on every new message user send/receive afterwards
-      // last message should be focused only when user enters a new conversation using keyboard(press enter)
-      history.state.eventKey = '';
-      window.history.replaceState(history.state, '', window.location.hash);
-    }
-  }, [totalMessage]);
-
```
atomrc, line 162: This logic has been moved to the parent container. The handling of focusing a message is now handled by the message list

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -123,7 +119,7 @@ const Message: React.FC<
       scrollTo?.({center: true, element: messageElementRef.current});
 
       // for reply message, focus on the original message when original message link is clicked for keyboard users
-      handleFocus(index);
+      handleFocus(message.id);
```
atomrc, line 122: We now base the focus on element ids rather than position in the array

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -95,11 +93,9 @@ const Message: React.FC<
     lastReadTimestamp,
     onVisible,
     scrollTo,
-    totalMessage,
```
atomrc, line 98: Those infos are not needed anymore to compute the focus index, only the messageId is necessary

src/script/hooks/useRoveFocus.ts
```diff
@@ -24,35 +24,40 @@ import type {KeyboardEvent as ReactKeyboardEvent} from 'react';
 
 import {isKey, isTabKey, KEY} from 'Util/KeyboardUtil';
 
-function useRoveFocus(size: number, defaultFocus = 0, infinite = true) {
```
atomrc, line 27: I have simplified the API, since `defaultFocus` and `infinite` were never set to different values

aweiss-dev (approved):

PR #16751, opened 2024-02-05
runfix: mls 1to1 migration system message

PatrykBuniX:
## Description

When loading the app for the first time on a new client, we should verify whether there are no proteus 1:1 conversations found on backend, with user we have established a MLS 1:1 conversation with. 

If such a conversation is found, we should blacklist it so it's never refetched from backend.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationSelectors.ts
+++ b/src/script/conversation/ConversationSelectors.ts
@@ -18,7 +18,11 @@
  */
 
 import {ConnectionStatus} from '@wireapp/api-client/lib/connection/';
-import {CONVERSATION_TYPE, ConversationProtocol} from '@wireapp/api-client/lib/conversation/';
+import {
+  CONVERSATION_TYPE,
+  ConversationProtocol,
+  Conversation as BackendConversation,
+} from '@wireapp/api-client/lib/conversation/';
 import {QualifiedId} from '@wireapp/api-client/lib/user/';
 
 import {matchQualifiedIds} from 'Util/QualifiedId';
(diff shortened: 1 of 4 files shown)


src/script/conversation/ConversationSelectors.ts
```diff
@@ -58,6 +62,38 @@ export function isTeamConversation(conversation: Conversation): boolean {
   return conversation.type() === CONVERSATION_TYPE.GLOBAL_TEAM;
 }
 
+export function isProteusTeam1to1Conversation({
+  name,
+  type,
+  inTeam,
+  otherMembersLength,
+}: {
+  name: string | undefined;
```
przemvs, line 71:
```suggestion
  name?: string;
```

src/script/conversation/ConversationSelectors.ts
```diff
@@ -58,6 +62,38 @@ export function isTeamConversation(conversation: Conversation): boolean {
...
+  otherMembersLength: number;
+}): boolean {
+  const isGroupConversation = type === CONVERSATION_TYPE.REGULAR;
+  const hasOneParticipant = otherMembersLength === 1;
+  return isGroupConversation && hasOneParticipant && inTeam && !name;
+}
+
+export function isBackendProteus1to1Conversation(conversation: BackendConversation): boolean {
+  const isProteus1to1 =
+    conversation.protocol === ConversationProtocol.PROTEUS && conversation.type === CONVERSATION_TYPE.ONE_TO_ONE;
```
przemvs, line 83: Why not to use isProteusConversation from ConversationSelectors?

PatrykBuniX: Because this one expects Conversation entity (knockout observable state) as a param, but we're verifying backend conversation (api client) here.

src/script/conversation/ConversationSelectors.ts
PatrykBuniX: We don't want this param to be optional, it always has to be passed - either a string value or `undefined` explicitly.

przemvs (approved):

atomrc (approved):

PR #16774, opened 2024-02-06
fix: allow end user to set all url paths (WPB-4676)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4676" title="WPB-4676" target="_blank">TaskWPB-4676</a>  URL_PATHS in webapp have inconsistencies and deprecations
  </td></table>
  <br />
 

## Description

- get rid of unnecessary logic building urls in `externatRoutes.ts`
- avoid using hard coded support urls and subpaths
- change the release note url from github to medium

## Checklist


src/script/externalRoute.ts
```diff
@@ -20,95 +20,46 @@
...
-  DECRYPT_ERROR_1: '/articles/207948115',
-  DECRYPT_ERROR_2: '/privacy/error-2/',
-  MANAGE_SERVICES: '/services/',
-  MANAGE_TEAM: '/login/',
-  PASSWORD_RESET: '/forgot/',
-  PRIVACY_HOW: '/privacy/how/',
-  PRIVACY_UNVERIFIED_USERS: '/articles/202857164',
-  PRIVACY_WHY: '/articles/207859815',
-  SUPPORT_USERNAME: '/support/username/',
-} as const;
```
V-Gira, line 51: - some of those are removed as deprecated
- support links still in use have been moved with existing ones as environment variables instead of being hardcoded
- url paths intended to be dynamically appended have been made into the `URL_SUBPATH` env variables as to be end user configurable

src/script/externalRoute.ts
```diff
@@ -20,95 +20,46 @@
...
-  if (!url) {
-    return undefined;
-  }
-  const language = currentLanguage().slice(0, 2);
-  const websiteLanguage = language == 'de' ? language : 'en-us';
-  return url.replace(
-    `${Config.getConfig().URL.SUPPORT.INDEX}`,
-    `${Config.getConfig().URL.SUPPORT.INDEX}/hc/${websiteLanguage}`,
-  );
-};
```
V-Gira: help center links do not need to be localized as they follow the browser locale setting

server/config/client.config.ts
```diff
@@ -81,27 +81,36 @@ export function generateConfig(params: ConfigGeneratorParams, env: Env) {
       MOBILE_BASE: env.URL_MOBILE_BASE,
       PRICING: env.URL_PRICING,
       PRIVACY_POLICY: env.URL_PRIVACY_POLICY,
+      SUBPATH: {
```
atomrc: I think `SUBPATH` is not the right wording for it. Maybe `URL_PATH` would do? WDYT?

PatrykBuniX (approved):

server/config/client.config.ts
V-Gira: Wasn't too sure about subpath tbh.
That would require updating the default config, and since we're at it, maybe we could get rid of the granularity and use a full path for all 4?
The granularity is not taken advantage of in the app and that sounds more straightforward to me for end user config, WD*Y*T?

atomrc: yes, let's flatten those URL, no more `BASE` URLs, I'd be up for that

src/script/externalRoute.ts
```diff
@@ -20,79 +20,41 @@
...
+const getPrivacyPolicyUrl = (): string | undefined => addLocaleToUrl(URL.PRIVACY_POLICY || undefined);
+const getTermsOfUsePersonalUrl = (): string | undefined => addLocaleToUrl(URL.TERMS_OF_USE_PERSONAL || undefined);
+const getTermsOfUseTeamUrl = (): string | undefined => addLocaleToUrl(URL.TERMS_OF_USE_TEAMS || undefined);
 
-export const getPrivacyPolicyUrl = (): string => addLocaleToUrl(URL.PRIVACY_POLICY || undefined);
-export const getTermsOfUsePersonalUrl = (): string => addLocaleToUrl(URL.TERMS_OF_USE_PERSONAL || undefined);
-export const getTermsOfUseTeamUrl = (): string => addLocaleToUrl(URL.TERMS_OF_USE_TEAMS || undefined);
+export const getManageServicesUrl = (utmSource?: string): string | undefined =>
+  getTeamSettingsUrl(URL.URL_PATH.MANAGE_SERVICES, utmSource);
```
V-Gira, line 56: hard to find doc on those UTM sources
we use 2 variations of those in the webapp:
- `client_landing` when navigating from the main portion of the app
- `client_settings` when navigating from the settings

Do we want to keep that granularity @atomrc ?
Should I had those to the `externalUrl` object and stop exporting those methods?

atomrc: ah!! I do not know about those utms, honestly. But, since we will be tracking team settings usages soonish, I'm guessing those will be used at some point (if they are not currently). 
We should probably keep a way to add utm to some urls I guess

V-Gira: I'll keep the export of the method and add some JSDoc to make the utmSource thing more transparent, how does that sound?

atomrc: That sounds good to me :)

src/script/externalRoute.ts
```diff
@@ -20,79 +20,56 @@
...
 
-export const getWebsiteUrl = (path: string = '', pkCampaign?: string): string | undefined => {
-  if (URL.WEBSITE) {
+const getWebsiteUrl = (path: string = '', pkCampaign?: string): string | undefined => {
+  if (URL.WEBSITE_BASE) {
     const query = pkCampaign ? `?pk_campaign=${pkCampaign}&pk_kwd=desktop` : '';
-    const websiteUrl = `${URL.WEBSITE}${path}${query}`;
-    return addLocaleToUrl(URL.WEBSITE ? websiteUrl : undefined);
+    const websiteUrl = `${URL.WEBSITE_BASE}${path}${query}`;
+    return addLocaleToUrl(URL.WEBSITE_BASE ? websiteUrl : undefined);
```
przemvs: This check is not necessary :D You will have this value, because You checked on line 32 ;)

przemvs:
```suggestion
    return addLocaleToUrl(websiteUrl);
```

src/script/externalRoute.ts
```diff
@@ -20,79 +20,56 @@
...
 
-const getHelpCenterUrl = (path: (typeof URL_PATH)[keyof typeof URL_PATH]) => {
-  if (URL.SUPPORT) {
-    const helpcenterUrl = `${URL.SUPPORT}${path}`;
-    return addLocaleToHelpCenterUrl(URL.SUPPORT ? helpcenterUrl : undefined);
-  }
-  return undefined;
+const getAccountPagesUrl = (path: string = ''): string | undefined => {
+  const accountPagesUrl = `${URL.ACCOUNT_BASE}${path}`;
+  return URL.ACCOUNT_BASE ? accountPagesUrl : undefined;
```
przemvs: You can simplify it :) 

```suggestion
  return URL.ACCOUNT_BASE ? `${URL.ACCOUNT_BASE}${path}` : undefined;
```

src/script/externalRoute.ts
```diff
@@ -20,79 +20,56 @@
...
-export const getCreateTeamUrl = (): string =>
-  Config.getConfig().FEATURE.ENABLE_ACCOUNT_REGISTRATION && `${Config.getConfig().URL.TEAMS_BASE}/register/email`;
-export const getDecryptErrorUrl = (): string => getHelpCenterUrl(URL_PATH.DECRYPT_ERROR_1);
-export const getPrivacyUnverifiedUsersUrl = (): string => getHelpCenterUrl(URL_PATH.PRIVACY_UNVERIFIED_USERS);
-export const getPrivacyWhyUrl = (): string => getHelpCenterUrl(URL_PATH.PRIVACY_WHY);
+const getCreateTeamUrl = (): string | undefined =>
+  Config.getConfig().FEATURE.ENABLE_ACCOUNT_REGISTRATION ? `${URL.TEAMS_BASE}${URL.URL_PATH.CREATE_TEAM}` : undefined;
 
-export const addLocaleToUrl = (url?: string): string => {
+export const addLocaleToUrl = (url?: string | undefined): string | undefined => {
```
przemvs: Why url?: string | undefined? :D You append undefined by ? after url :)

przemvs (approved):
LGTM! :)

atomrc (approved):
LGTM 😎

PR #16779, opened 2024-02-07
runfix: Initialize soft lock upon e2eHandler initialization

atomrc:
## Description

This will make sure we do initialize the softLock as soon as the e2e enrollmenet class is initialized with the team config. 

## Checklist

### DIFF
--- a/src/script/hooks/useAppSoftLock.ts
+++ b/src/script/hooks/useAppSoftLock.ts
@@ -40,12 +40,11 @@ export function useAppSoftLock(callingRepository: CallingRepository, notificatio
 
   useEffect(() => {
     const e2eiHandler = E2EIHandler.getInstance();
-    if (!e2eiHandler.isE2EIEnabled()) {
-      return () => {};
-    }
 
+    e2eiHandler.on('initialized', handleSoftLockActivation);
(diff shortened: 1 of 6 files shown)


src/script/hooks/useAppSoftLock.ts
```diff
@@ -40,12 +40,11 @@ export function useAppSoftLock(callingRepository: CallingRepository, notificatio
 
   useEffect(() => {
     const e2eiHandler = E2EIHandler.getInstance();
-    if (!e2eiHandler.isE2EIEnabled()) {
-      return () => {};
-    }
```
atomrc, line 45: This check was always "false" since the handler was not yet initialized when we first call this hook (and was never re-rendered).

src/script/hooks/useAppSoftLock.ts
```diff
@@ -40,12 +40,11 @@ export function useAppSoftLock(callingRepository: CallingRepository, notificatio
...
 
+    e2eiHandler.on('initialized', handleSoftLockActivation);
```
atomrc, line 44: Here we make sure that we compute the lock state as soon as the handler is initialized

PatrykBuniX (approved):

PR #16791, opened 2024-02-08
runfix: Avoid focusing on messages with undefined ids [WPB-6509] 

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6509" title="WPB-6509" target="_blank">BugWPB-6509</a>  [Web] Edge - Conversation scrolls down automatically when entering conversation
  </td></table>
  <br />
 

## Description

This caused a problem where loading a conversation with messages that have no ID would trigger a weird initial scroll. 

The problem is we didn't check for `undefined` focused message ID. 

### Before

https://github.com/wireapp/wire-webapp/assets/1090716/b2a22cd7-6057-47e7-b007-0c851570f72f

### After

https://github.com/wireapp/wire-webapp/assets/1090716/9860aeb8-44b6-4b80-b6e4-cc4fcf3c51ec

## Checklist

### DIFF
--- a/src/script/components/MessagesList/MessageList.tsx
+++ b/src/script/components/MessagesList/MessageList.tsx
@@ -118,7 +118,7 @@ export const MessagesList: FC<MessagesListParams> = ({
 
   const messageListRef = useRef<HTMLDivElement | null>(null);
   const [loaded, setLoaded] = useState(false);
-  const [focusedMessage, setFocusedMessage] = useState<string | undefined>(initialMessage?.id);
+  const [highlightedMessage, setHighlightedMessage] = useState<string | undefined>(initialMessage?.id);
   const conversationLastReadTimestamp = useRef(conversation.last_read_timestamp());
 
   const filteredMessages = filterMessages(allMessages);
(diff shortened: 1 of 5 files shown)


src/script/components/MessagesList/MessageList.tsx
```diff
@@ -277,13 +277,16 @@ export const MessagesList: FC<MessagesListParams> = ({
...
-          return messages.map((message, index) => {
+          return messages.map(message => {
             const isLastDeliveredMessage = lastDeliveredMessage?.id === message.id;
 
             const visibleCallback = getVisibleCallback(conversation, message);
 
             const key = `${message.id || 'message'}-${message.timestamp()}`;
 
+            const isHighlighted = !!highlightedMessage && highlightedMessage === message.id;
+            const isFocused = !!focusedId && focusedId === message.id;
```
atomrc, line 288: This is the actual bugfix. We need to check `!!focusedId` before doing the comparison between `focusedId` and `message.id`. 

In case `focusedId === undefined` then this condition would be true

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -320,7 +323,7 @@ export const MessagesList: FC<MessagesListParams> = ({
                 }}
                 selfId={selfUser.qualifiedId}
                 shouldShowInvitePeople={shouldShowInvitePeople}
-                isMessageFocused={focusedId === message.id}
```
atomrc, line 323: This was the wrong condition here

V-Gira (approved):
Good call on renaming `isMarked`, wasn't exactly clear :)

PR #16794, opened 2024-02-09
fix: dl path new device

tlebon:
## Description
fixes an edge case where you login to an account with a new device that has a dl path update

## Checklist

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -23,6 +23,7 @@ import 'core-js/full/reflect';
 
 import {Context} from '@wireapp/api-client/lib/auth';
 import {ClientClassification, ClientType} from '@wireapp/api-client/lib/client/';
+import {FEATURE_KEY} from '@wireapp/api-client/lib/team';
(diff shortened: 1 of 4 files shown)


aweiss-dev (approved):

src/script/main/app.ts
```diff
@@ -398,6 +406,15 @@ export class App {
...
 
+      if (dlPathNotifier) {
+        localStorage.setItem(
+          'enforcedDownloadLocation',
+          teamFeatures[FEATURE_KEY.ENFORCE_DOWNLOAD_PATH]?.config.enforcedDownloadLocation ?? '',
+        );
+        const {modalOptions, modalType} = getModalOptions({type: ModalType.DOWNLOAD_PATH_CHANGED});
+        PrimaryModal.show(modalType, modalOptions);
+      }
+
```
atomrc: Let's avoid adding inline code to the app.ts file. Let's extract that to a team config handler (like we did for e2ei) and have the code of that feature living there

tlebon: makes sense

PR #16815, opened 2024-02-12
fix: handle connection requests with deleted users [WPB-4356]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-4356" title="WPB-4356" target="_blank">BugWPB-4356</a>  Web/Desktop - Connection request from a deleted profile cannot be removed/ignored
  </td></table>
  <br />
 

## Description

- filter out connection requests with deleted users - connection request is a conversation of type 3 (CONNECT)
- delete outgoing connection request when trying to cancel it and it fails because of the other user being deleted.
- delete incoming connection request when trying to accept/ignore it if the other user is deleted

## Checklist

### DIFF
--- a/src/script/connection/ConnectionRepository.ts
+++ b/src/script/connection/ConnectionRepository.ts
@@ -52,6 +52,7 @@ export class ConnectionRepository {
   private readonly connectionService: ConnectionService;
   private readonly userRepository: UserRepository;
   private readonly logger: Logger;
+  private deleteConnectionRequestConversation?: (userId: QualifiedId) => Promise<void>;
 
   static get CONFIG(): Record<string, BackendEventType[]> {
     return {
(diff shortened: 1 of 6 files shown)


przemvs (approved):

src/script/connection/ConnectionRepository.ts
```diff
@@ -52,6 +52,7 @@ export class ConnectionRepository {
   private readonly connectionService: ConnectionService;
   private readonly userRepository: UserRepository;
   private readonly logger: Logger;
+  private deleteConnectionRequestConversation?: (userId: QualifiedId) => Promise<void>;
```
atomrc: We could probably find a naming that suggest it's a callback (here we feel it's going to trigger an imperative action)

atomrc (approved):

PR #16818, opened 2024-02-13
runfix: Remove E2EI feature flag [WPB-6467]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6467" title="WPB-6467" target="_blank">BugWPB-6467</a>  [Web] UI says "Something went wrong" but user gets a valid certificate
  </td></table>
  <br />
 

## Description

Now the team feature config becomes the single source of truth whether e2ei should be enabled or not. 

The core has also changed to allow this: 
- now the `initClient` will take a boolean to say whether e2ei enrollment will happen (to avoid uploading identity to the backend before the enrollment is done) (see https://github.com/wireapp/wire-web-packages/pull/5972)

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@peculiar/x509": "1.9.6",
     "@wireapp/avs": "9.6.9",
     "@wireapp/commons": "5.2.4",
-    "@wireapp/core": "43.14.3",
+    "@wireapp/core": "44.0.0",
     "@wireapp/react-ui-kit": "9.15.1",
     "@wireapp/store-engine-dexie": "2.1.7",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "@peculiar/x509": "1.9.6",
     "@wireapp/avs": "9.6.9",
     "@wireapp/commons": "5.2.4",
-    "@wireapp/core": "43.14.3",
+    "@wireapp/core": "44.0.0",
```
PatrykBuniX, line 11: 🚀

PatrykBuniX (approved):
Nice, less sources of truth!!! 🚀

PR #16845, opened 2024-02-14
fix: improve visibility of join button for temporary guests (WPB-3550)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-3550" title="WPB-3550" target="_blank">TaskWPB-3550</a>  [WEB] Improve visibility of "join call" button in a guest conversation
  </td></table>
  <br />
 

## Description

When a temporary guest joins a conversation where a call is already ongoing, the join button is easily missed

We decided to show the call ui in that case, even if the call is declined (a temporary guest only has access to 1 conversation, no other groups or 1on1)

Before:
image

After:
Screenshot from 2024-02-14 17-00-09

## Checklist

### DIFF
--- a/src/script/components/calling/CallingCell.tsx
+++ b/src/script/components/calling/CallingCell.tsx
@@ -123,7 +123,6 @@ const CallingCell: React.FC<CallingCellProps> = ({
   const {activeCallViewTab} = useKoSubscribableChildren(callState, ['activeCallViewTab']);
   const isMuted = muteState !== MuteState.NOT_MUTED;
 
-  const isStillOngoing = reason === CALL_REASON.STILL_ONGOING;
   const isDeclined = !!reason && [CALL_REASON.STILL_ONGOING, CALL_REASON.ANSWERED_ELSEWHERE].includes(reason);
 
   const isOutgoing = state === CALL_STATE.OUTGOING;
@@ -173,7 +172,6 @@ const CallingCell: React.FC<CallingCellProps> = ({
   const disableVideoButton = isOutgoingVideoCall || isVideoUnsupported;
   const disableScreenButton = !callingRepository.supportsScreenSharing;
 
-  const showJoinButton = conversation && isStillOngoing && temporaryUserStyle;
   const [showParticipants, setShowParticipants] = useState(false);
   const isModerator = selfUser && roles[selfUser.id] === DefaultConversationRoleName.WIRE_ADMIN;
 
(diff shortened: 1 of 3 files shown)


src/script/components/calling/CallingCell.tsx
```diff
@@ -304,19 +302,7 @@ const CallingCell: React.FC<CallingCellProps> = ({
...
-          className="call-ui__button call-ui__button--green call-ui__button--join"
-          style={{margin: '40px 16px 0px'}}
-          onClick={() => callActions.answer(call)}
-          type="button"
-          data-uie-name="do-call-controls-call-join"
-        >
-          {t('callJoin')}
-        </button>
-      )}
-
```
V-Gira, line 318: this is the old join button component, strictly for temporary guests

src/script/components/calling/CallingCell.tsx
```diff
@@ -304,19 +302,7 @@ const CallingCell: React.FC<CallingCellProps> = ({
...
-      {conversation && !isDeclined && (
+      {conversation && (!isDeclined || isTemporaryUser) && (
```
V-Gira, line 305: we always show the call ui to temp users, even if the call is declined

src/script/components/calling/CallingCell.tsx
```diff
@@ -448,163 +434,170 @@ const CallingCell: React.FC<CallingCellProps> = ({
             <ConversationClassifiedBar conversation={conversation} classifiedDomains={classifiedDomains} />
           )}
 
-          {!isDeclined && (
-            <>
```
V-Gira, line 452: doesn't appear this check was ever necessary as it's already performed on (previously) line 319

src/script/components/calling/CallingCell.tsx
```diff
@@ -448,163 +434,170 @@ const CallingCell: React.FC<CallingCellProps> = ({
...
+                  {isDeclined ? (
+                    <button
+                      className="call-ui__button call-ui__button--green call-ui__button--join call-ui__button--join--large "
+                      onClick={answerCall}
+                      type="button"
+                      data-uie-name="do-call-controls-call-join"
+                    >
+                      {t('callJoin')}
+                    </button>
+                  ) : (
```
V-Gira, line 557: `IsDeclined` can only be true at that stage if `ìsTemporaryUser` is also true.

In that case we show a larger join button and no decline button if a call is ongoing bit the temp user hasn't accepted it yet

atomrc: damn this logic is quite hard to follow 😅

atomrc (approved):

src/script/components/calling/CallingCell.tsx
V-Gira: I could add a few comments, maybe rename
```
const isDeclined = !!reason && [CALL_REASON.STILL_ONGOING, CALL_REASON.ANSWERED_ELSEWHERE].includes(reason);
```
to `isNotAccepted`?

atomrc: Nah, I think the problem lies mainly in the rendering part, things are a little too nested. But not sure there is an easy way to fix that now; :)

PR #16847, opened 2024-02-15
chore: Fix changelog generation

atomrc:
## Description

Currently the condition to generate a changelog and a github release is not correct

## Checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -348,10 +348,6 @@ jobs:
         target: ${{fromJson(needs.set_deployment_targets.outputs.deployment_targets)}}
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -381,10 +377,6 @@ jobs:
       release_url: ${{ steps.release.outputs.url }}
 
     steps:
-      - name: Set tag
-        run: |
-          echo "TAG=$(git tag --points-at ${{github.sha}})" >> $GITHUB_ENV
```
atomrc, line 386: instead of using the `tag` we use the release name that was generated in an earlier step

V-Gira (approved):

PR #16849, opened 2024-02-15
feat: add badges to new navigation sidebar (WPB-6651)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6651" title="WPB-6651" target="_blank">TaskWPB-6651</a>  [Web] Add focus and badges to sidebar items
  </td></table>
  <br />
 

image

### DIFF
--- a/src/style/common/label.less
+++ b/src/style/common/label.less
@@ -22,6 +22,11 @@
   font-weight: @font-weight-semibold;
 }
 
+.label-sm {
+  font-size: @font-size-small;
+  font-weight: @font-weight-semibold;
+}
+
 .label-xs {
   .label;
 
--- a/src/style/common/label.less
+++ b/src/style/common/label.less
@@ -22,11 +22,6 @@
   font-weight: @font-weight-semibold;
 }
 
-.label-sm {
-  font-size: @font-size-small;
-  font-weight: @font-weight-semibold;
-}
-
 .label-xs {
   .label;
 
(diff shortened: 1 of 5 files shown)


src/style/common/label.less
```diff
@@ -22,6 +22,11 @@
   font-weight: @font-weight-semibold;
 }
 
+.label-sm {
+  font-size: @font-size-small;
+  font-weight: @font-weight-semibold;
+}
+
```
V-Gira: I think that's `label-2` in `typing.less` , the new label templates @przemvs  was talking about the other day

thisisamir98: image
gray-70 from .label-2 does not work here

thisisamir98: https://github.com/wireapp/wire-webapp/pull/16849/commits/25cd5b5bce57313f0fdfbbdecea9351b32096ff8

src/style/list/conversations.less
```diff
@@ -89,13 +96,20 @@
...
+    background: var(--red-500);
+    color: var(--button-tertiary-bg);
+    .label-sm;
+  }
+
+  &.active {
+    background-color: var(--gray-95);
+  }
+  .conversations-sidebar-btn--text {
+    color: var(--gray-100);
```
V-Gira: That looks, like something from before your last PR, it's duplicated with the correct values below

V-Gira (approved):

V-Gira (commented):
Looking good, just one thing that looks like it should be snipped :scissors:

V-Gira (approved):

PR #16889, opened 2024-02-23
test: multiple add users failures

PatrykBuniX:
## Description

Added a test covering a add user failure system message with different reason for each user. No logic changes, just testing :)

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
+++ b/src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
@@ -19,7 +19,7 @@
 
 import {act, render} from '@testing-library/react';
 import {QualifiedId} from '@wireapp/api-client/lib/user';
-import {AddUsersFailureReasons} from '@wireapp/core/lib/conversation';
+import {AddUsersFailure, AddUsersFailureReasons} from '@wireapp/core/lib/conversation';
 
 import en from 'I18n/en-US.json';
 import {withTheme, generateQualifiedIds} from 'src/script/auth/util/test/TestUtil';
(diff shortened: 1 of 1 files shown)


src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
```diff
@@ -212,4 +220,75 @@ describe('FailedToAddUsersMessage', () => {
...
+
+    const mainMessage = getAllByText(
+      (_, element) => element?.textContent === '3 participants could not be added to the group.',
+    );
+    expect(mainMessage.length).toBeGreaterThanOrEqual(1);
+
+    const details1 = getAllByText(
+      (_, element) =>
+        element?.textContent ===
+        `${user1.name()} could not be added to the group as their backends do not federate with each other.`,
```
V-Gira, line 273: that's a weird copy.
Seems like there's unnecessary (until now) strings for `...SingularDetails...` , a single error wouldn't have the show details button as far as I remember.

Not related to this but I'll keep it in mind for an other look

V-Gira (approved):

src/script/components/MessagesList/Message/FailedToAddUsersMessage.test.tsx
PatrykBuniX: Hmm, but it's not a single error, it's one of the errors but with this particular reason just for this single user.

V-Gira: that's why I say "until now", I don't think that situation existed prior to the mls rework.
Doesn't seem to be a reason to have 2 "singular" strings

PR #16913, opened 2024-02-26
runfix: verify mls state on welcome message [WPB-6794]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6794" title="WPB-6794" target="_blank">BugWPB-6794</a>  [Web] MLS conversation with verified users is not marked as verified right after creation
  </td></table>
  <br />
 

## Description

Bumps a core with version that emits a 'newEpoch' event also when receiving a welcome message. With this update, we will re-evaluate mls verification state right after being added to a MLS group with a welcome message (what was not a case before).

It also includes some refactoring and updates some comments.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@peculiar/x509": "1.9.6",
     "@wireapp/avs": "9.6.12",
     "@wireapp/commons": "5.2.4",
-    "@wireapp/core": "45.0.1",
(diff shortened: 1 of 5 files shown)


atomrc (approved):
solid 💪

PR #16920, opened 2024-02-26
runfix: assign full width to message list

V-Gira:
## Description

The left gap seems to have been implemented twice, as padding, and as a restriction on full width.
Was spotted by the action menu shifting to the left

before:
image

after:
image

### DIFF
--- a/src/style/content/conversation/message-list.less
+++ b/src/style/content/conversation/message-list.less
@@ -274,7 +274,7 @@
 // MESSAGE - BODY
 .message-body {
   position: relative;
-  max-width: calc(100% - var(--conversation-message-sender-width));
+  max-width: 100%;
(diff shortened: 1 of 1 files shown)


src/style/content/conversation/message-list.less
```diff
@@ -274,7 +274,7 @@
 // MESSAGE - BODY
 .message-body {
   position: relative;
-  max-width: calc(100% - var(--conversation-message-sender-width));
+  max-width: 100%;
   padding-left: var(--conversation-message-sender-width);
```
V-Gira, line 278: ` var(--conversation-message-sender-width)` (the left gap for a user's avatar) is already taken care of with padding here

przemvs (approved):

PR #16923, opened 2024-02-26
feat: remove 3rd party certificate library

aweiss-dev:
Since Core-Crypto added all the information we need, regarding the E2E - certificate, we can remove the 3rd party dependency from our code base.


src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -40,6 +40,17 @@ export type WireIdentity = Omit<DeviceIdentity, 'status'> & {
   status: MLSStatuses;
 };
 
+export function getCertificateDetails(identity: WireIdentity) {
+  const currentDate = new Date();
+  const timeRemainingMS = new Date(Number(identity.notAfter)).getTime() - currentDate.getTime();
```
aweiss-dev, line 45: number conversion needed because we get a bigint :)

PR #16928, opened 2024-02-27
runfix: reset verification state on leave [WPB-6864]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6864" title="WPB-6864" target="_blank">BugWPB-6864</a>  [Web] Remove mls verification state after leaving/being removed from the group
  </td></table>
  <br />
 

### Before
Group was still marked as verified, errors in the console
before

### After
Group mls verification state is reset to default (verified), no errors
after

## Description

getUserIdentities can now return `undefined` in case mls group does not exist anymore. Before trying to degrade/verify conversation state we should check if mls conversation does exist. If it does not - we reset mls verification state to the initial value (unverified).

Also whenever we leave the group (or get removed from it), we also reset the state to the initial value.

Needs https://github.com/wireapp/wire-web-packages/pull/6018 to be merged first.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2464,7 +2464,8 @@ export class ConversationRepository {
    * @param mlsConversation mls conversation
    */
   async wipeMLSCapableConversation(conversation: MLSCapableConversation) {
-    return this.conversationService.wipeMLSCapableConversation(conversation);
+    await this.conversationService.wipeMLSCapableConversation(conversation);
+    conversation.mlsVerificationState(ConversationVerificationState.UNVERIFIED);
   }
 
   async leaveGuestRoom(): Promise<void> {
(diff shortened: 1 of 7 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2464,7 +2464,8 @@ export class ConversationRepository {
    * @param mlsConversation mls conversation
    */
   async wipeMLSCapableConversation(conversation: MLSCapableConversation) {
-    return this.conversationService.wipeMLSCapableConversation(conversation);
+    await this.conversationService.wipeMLSCapableConversation(conversation);
+    conversation.mlsVerificationState(ConversationVerificationState.UNVERIFIED);
```
PatrykBuniX, line 2468: If for whatever reason mls group is wiped, reset verification state to default value

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -143,6 +148,12 @@ class MLSConversationVerificationStateHandler {
       return;
     }
 
+    const conversationExists = await this.core.service?.mls?.conversationExists(conversation.groupId);
```
PatrykBuniX, line 151: If when verifying e2ei state, mls group doesn't exist anymore, reset to default (unverified)

atomrc (approved):

PR #16933, opened 2024-02-27
chore: set node version fix to 18.16 

aweiss-dev:
All Version from 18.17 on cause crypto error in our tests.


tlebon:
question is if this will kill our instances running in aws on node 16 🤔 
i guess no if we were already set to 18

aweiss-dev:
outdated

PR #16944, opened 2024-02-28
feat(e2ei): check all e2ei certs in a conversation against their users name and handle (WPB-6194)

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6194" title="WPB-6194" target="_blank">BugWPB-6194</a>  [Web] Clients should verify user identifiers inside certificate
  </td></table>
  <br />
 

CoreCrypto only provides us with the information if the stored certificate for a client is still valid and active. 

It can not tell us if the certificate still matches the current user information of the user.
This needs to be checked on app side, since the user information might have been changed since the certificate had ben issued.

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -141,7 +141,7 @@ class MLSConversationVerificationStateHandler {
   }> => {
     const userIdentities = await getAllGroupUsersIdentities(conversation.groupId);
     const processedUserIds: Set<string> = new Set();
-    let problemFound = false;
+    let userVerificationState = UserVerificationState.ALL_VALID;
 
     if (userIdentities) {
       for (const [userId, identities] of userIdentities.entries()) {
@@ -159,7 +159,7 @@ class MLSConversationVerificationStateHandler {
 
         if (!identity || !user) {
           this.logger.warn(`Could not find user or identity for userId: ${userId}`);
-          problemFound = true;
+          userVerificationState = UserVerificationState.SOME_INVALID;
           break;
         }
 
@@ -167,14 +167,14 @@ class MLSConversationVerificationStateHandler {
         const matchingHandle = this.checkUserHandle(identity, user);
         if (!matchingHandle || !matchingName) {
           this.logger.warn(`User identity and user entity do not match for userId: ${userId}`);
-          problemFound = true;
+          userVerificationState = UserVerificationState.SOME_INVALID;
           break;
         }
       }
     }
 
     return {
-      userVerificationState: problemFound ? UserVerificationState.SOME_INVALID : UserVerificationState.ALL_VALID,
+      userVerificationState,
       userIdentities,
     };
   };
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -71,10 +71,18 @@ class MLSConversationVerificationStateHandler {
    * Changes mls verification state to "degraded"
    * @param conversation
    */
-  private async degradeConversation(conversation: MLSConversation, userIdentities: Map<string, WireIdentity[]>) {
+  private async degradeConversation(
+    conversation: MLSConversation,
+    userIdentities: Map<string, WireIdentity[]> | undefined,
+  ) {
+    if (!userIdentities) {
+      return;
+    }
+
     const state = ConversationVerificationState.DEGRADED;
     conversation.mlsVerificationState(state);
     const degradedUsers: QualifiedId[] = [];
+
     for (const [, identities] of userIdentities.entries()) {
       if (identities.length > 0 && identities.some(identity => identity.status !== MLSStatuses.VALID)) {
         degradedUsers.push(identities[0].qualifiedUserId);
@@ -130,18 +138,17 @@ class MLSConversationVerificationStateHandler {
   ): Promise<{
     userVerificationState: UserVerificationState;
     userIdentities: Map<string, WireIdentity[]> | undefined;
-    problematicUserIds: Set<string>;
   }> => {
     const userIdentities = await getAllGroupUsersIdentities(conversation.groupId);
-    const processedUserIds: string[] = [];
-    const problematicUserIds: Set<string> = new Set();
+    const processedUserIds: Set<string> = new Set();
+    let problemFound = false;
 
     if (userIdentities) {
       for (const [userId, identities] of userIdentities.entries()) {
-        if (processedUserIds.includes(userId)) {
+        if (processedUserIds.has(userId)) {
           continue;
         }
-        processedUserIds.push(userId);
+        processedUserIds.add(userId);
 
         /**
          * We need to wait for the user entity to be available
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -120,6 +122,64 @@ class MLSConversationVerificationStateHandler {
...
+    const processedUserIds: string[] = [];
+    const problematicUserIds: Set<string> = new Set();
+
+    for (const [userId, identities] of userIdentities.entries()) {
+      if (processedUserIds.includes(userId)) {
+        continue;
+      }
+      processedUserIds.push(userId);
+
+      const user = await waitFor(() => conversation.allUserEntities().find(user => user.qualifiedId.id === userId));
```
aweiss-dev: We have a race condition here in certain circumstances. In some cases the users are locally not updated when we get the epochChange event.

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -120,6 +122,64 @@ class MLSConversationVerificationStateHandler {
     await this.checkAllConversationsVerificationState();
   };
 
+  private checkUserHandle = (identity: WireIdentity, user: User): boolean => {
+    // WireIdentity handle format is "{scheme}%40{username}@{domain}"
+    // Example: wireapp://%40hans.wurst@elna.wire.link
+    const {handle: identityHandle} = identity;
+    // We only want to check the username part of the handle
+    const regex = /.*%40([^@]+)@[^@]+/;
```
atomrc: This looks wrong having to hardcode that. I don't think we should make any assumption to what this format is going to be sadly

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -131,6 +191,7 @@ class MLSConversationVerificationStateHandler {
   private onEpochChanged = async ({groupId}: {groupId: string}): Promise<void> => {
     // There could be a race condition where we would receive an epoch update for a conversation that is not yet known by the webapp.
     // We just wait for it to be available and then check the verification state
+
```
atomrc:
```suggestion
```

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
atomrc: Please add a comment in the code explaining the race condition. This should not be left uncommented. 

Please explain the exact reason why there is a race condition

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -120,14 +122,70 @@ class MLSConversationVerificationStateHandler {
...
+          problematicUserIds.add(userId);
+        }
+      }
+    }
+
+    return {
+      userVerificationState:
+        problematicUserIds.size > 0 ? UserVerificationState.SOME_INVALID : UserVerificationState.ALL_VALID,
+      userIdentities,
+      problematicUserIds,
```
atomrc: This seems to be never consumed

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -120,14 +122,70 @@ class MLSConversationVerificationStateHandler {
...
+          problematicUserIds.add(userId);
+          continue;
+        }
+
+        const matchingName = identity.displayName === user.name();
+        const matchingHandle = this.checkUserHandle(identity, user);
+        if (!matchingHandle || !matchingName) {
+          this.logger.warn(`User identity and user entity do not match for userId: ${userId}`);
+          problematicUserIds.add(userId);
+        }
```
atomrc: Why not just returning at this point. No point in continuing checking other users, right?

aweiss-dev: I wanted to provide a complete list of problematic cases, in case we need to display that information at some point. But I am thinking too far ahead here. I'll follow your suggestion

przemvs (approved):

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -64,19 +71,19 @@ class MLSConversationVerificationStateHandler {
...
     conversation.mlsVerificationState(state);
 
     const degradedUsers: QualifiedId[] = [];
-    for (const [userId, identities] of userIdentities.entries()) {
-      if (identities.some(identity => identity.status !== MLSStatuses.VALID)) {
-        degradedUsers.push({id: userId, domain: ''});
+    if (userIdentities) {
+      for (const [userId, identities] of userIdentities.entries()) {
+        if (identities.some(identity => identity.status !== MLSStatuses.VALID)) {
+          degradedUsers.push({id: userId, domain: ''});
```
PatrykBuniX: Why the domain is empty?

aweiss-dev: I would guess because we dont have it at this point?

That code existed before, just added the if around it

PatrykBuniX: Isn't it a stringified qualified id actually? Something like `${user}@${domain}`?

aweiss-dev: so, we push a stringified qualified id into the id part of a qualified id? Sound wrong to me

aweiss-dev: Fixed by another PR and merged

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -115,14 +122,68 @@ class MLSConversationVerificationStateHandler {
...
+
+  private checkAllUserCredentialsInConversation = async (
+    conversation: MLSCapableConversation,
+  ): Promise<{
+    userVerificationState: UserVerificationState;
+    userIdentities: Map<string, WireIdentity[]> | undefined;
+  }> => {
+    const userIdentities = await getAllGroupUsersIdentities(conversation.groupId);
+    const processedUserIds: Set<string> = new Set();
+    let problemFound = false;
```
atomrc: why a boolean here and not ? 

```js
let userVerificationState = UserVerificationState.ALL_VALID;
```

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -115,14 +122,68 @@ class MLSConversationVerificationStateHandler {
...
+        /**
+         * We need to wait for the user entity to be available
+         * There is a race condition when adding a new user to a conversation, the host will receive the epoch update before the user entity is available
+         */
+        const user = await waitFor(() => conversation.allUserEntities().find(user => user.qualifiedId.id === userId));
+        const identity = identities.at(0);
+
+        if (!identity || !user) {
+          this.logger.warn(`Could not find user or identity for userId: ${userId}`);
+          problemFound = true;
```
atomrc:
```suggestion
          userVerificationState = UserVerificationState.SOME_INVALID;
```

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -115,14 +122,68 @@ class MLSConversationVerificationStateHandler {
...
+          break;
+        }
+
+        const matchingName = identity.displayName === user.name();
+        const matchingHandle = this.checkUserHandle(identity, user);
+        if (!matchingHandle || !matchingName) {
+          this.logger.warn(`User identity and user entity do not match for userId: ${userId}`);
+          problemFound = true;
```
atomrc:
```suggestion
          userVerificationState = UserVerificationState.SOME_INVALID;
```

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -115,14 +122,68 @@ class MLSConversationVerificationStateHandler {
...
+          break;
+        }
+      }
+    }
+
+    return {
+      userVerificationState: problemFound ? UserVerificationState.SOME_INVALID : UserVerificationState.ALL_VALID,
```
atomrc:
```suggestion
      userVerificationState,
```

atomrc (approved):

PR #16996, opened 2024-03-04
runfix: start grace period timers

PatrykBuniX:
## Description

After cert get expired, and when trying to renew it, app gets reloaded, in app init flow, we call `startTimers`. This method was trying to check self devices identity by loading the self mls conversation (its groupId) from the local state. It's possible that we call startTimers before all the conversations are loaded into the local store, what resulted in an error that self mls conversation was not found - see screenshot below.

Screenshot 2024-03-04 at 15 07 55

The fix is to fetch the self mls conversation from backend if it's not yet known locally, and get the conversation's groupId from there.

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityVerification.ts
+++ b/src/script/E2EIdentity/E2EIdentityVerification.ts
@@ -90,16 +90,13 @@ export async function getConversationVerificationState(groupId: string) {
  * Checks if E2EI has active certificate.
  */
 const getSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
-  let selfMLSConversationGroupId: string;
-  try {
-    // Try to get the self MLS conversation from the conversation state
-    const conversationState = container.resolve(ConversationState);
-    selfMLSConversationGroupId = conversationState.getSelfMLSConversation().groupId;
-  } catch {
-    // If the conversation state is not available, try to get the self MLS conversation from backend
-    const selfMLSConversation = await getCoreConversationService().getMLSSelfConversation();
-    selfMLSConversationGroupId = selfMLSConversation.group_id;
-  }
+  const conversationState = container.resolve(ConversationState);
+
+  // Try to get the self MLS conversation from the conversation state
+  // If the conversation state is not available, try to get the self MLS conversation from backend
+  const selfMLSConversationGroupId =
+    conversationState.selfMLSConversation()?.groupId ||
+    (await getCoreConversationService().getMLSSelfConversation()).group_id;
 
   const userIdentities = await getAllGroupUsersIdentities(selfMLSConversationGroupId);
 
--- a/src/script/E2EIdentity/E2EIdentityVerification.ts
+++ b/src/script/E2EIdentity/E2EIdentityVerification.ts
@@ -49,6 +49,14 @@ export function getE2EIdentityService() {
   return e2eIdentityService;
 }
 
+export function getCoreConversationService() {
+  const conversationService = container.resolve(Core).service?.conversation;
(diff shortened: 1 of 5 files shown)


src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -81,17 +89,28 @@ export async function getConversationVerificationState(groupId: string) {
...
-  const conversationState = container.resolve(ConversationState);
-  const selfMLSConversation = conversationState.getSelfMLSConversation();
-  const userIdentities = await getAllGroupUsersIdentities(selfMLSConversation.groupId);
+const getSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
+  let selfMLSConversationGroupId: string;
+  try {
+    // Try to get the self MLS conversation from the conversation state
+    const conversationState = container.resolve(ConversationState);
+    selfMLSConversationGroupId = conversationState.getSelfMLSConversation().groupId;
+  } catch {
```
atomrc: Instead of a try/catch, you can use `conversationState.selfMLSConversation` that will return `undefined` if it is not set

src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -81,17 +89,28 @@ export async function getConversationVerificationState(groupId: string) {
...
   if (!userIdentities) {
     return undefined;
   }
 
-  const currentClientId = selfMLSConversation.selfUser()?.localClient?.id;
-  const userId = selfMLSConversation.selfUser()?.qualifiedId;
+  const core = container.resolve(Core);
+
+  const currentClientId = core.clientId;
+  const userId = {id: core.userId, domain: core.backendFeatures.domain};
```
atomrc, line 110: We should probably change this to a fully qualified ID on core side :D

src/script/E2EIdentity/E2EIdentityVerification.ts
PatrykBuniX: Oh right, good idea!

PatrykBuniX: done https://github.com/wireapp/wire-webapp/pull/16996/commits/c1b7adec6235da82ca0f7cae79184bfaa51dcb95

src/script/E2EIdentity/E2EIdentityVerification.ts
PatrykBuniX: Agreed!

src/script/E2EIdentity/E2EIdentityVerification.ts
```diff
@@ -81,17 +89,25 @@ export async function getConversationVerificationState(groupId: string) {
...
+const getSelfDeviceIdentity = async (): Promise<WireIdentity | undefined> => {
   const conversationState = container.resolve(ConversationState);
-  const selfMLSConversation = conversationState.getSelfMLSConversation();
-  const userIdentities = await getAllGroupUsersIdentities(selfMLSConversation.groupId);
+
+  // Try to get the self MLS conversation from the conversation state
+  // If the conversation state is not available, try to get the self MLS conversation from backend
+  const selfMLSConversationGroupId =
+    conversationState.selfMLSConversation()?.groupId ||
+    (await getCoreConversationService().getMLSSelfConversation()).group_id;
```
atomrc:
```suggestion
    conversationState.selfMLSConversation()?.groupId ??
    (await getCoreConversationService().getMLSSelfConversation()).group_id;
```

Just to be on the safe side :)

atomrc (approved):

PR #17006, opened 2024-03-07
chore: Publish tagged commits to wire-builds repo

atomrc:
## Description

This will allow tags with the pattern `release-q1-2024` to be pushed to the `wire-builds` repo

## Checklist


.github/workflows/publish.yml
```diff
@@ -57,33 +58,6 @@ jobs:
...
-          if [ "$version_tag" == "dev" ]; then
-            wire_builds_target_branches='["dev"]'
-          fi
-
-          if [ "$version_tag" == "release/q1-2024" ]; then
-            wire_builds_target_branches='["q1-2024"]'
-          fi
-
-          echo "wire_builds_target_branches: $wire_builds_target_branches"
-          echo "wire_builds_target_branches=$wire_builds_target_branches" >> $GITHUB_OUTPUT
```
atomrc, line 85: This was simplified and moved later on in the job (see https://github.com/wireapp/wire-webapp/pull/17006/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R160)

.github/workflows/publish.yml
```diff
@@ -184,10 +157,36 @@ jobs:
...
+          # },
+          # "staging": {
+          #   "targets": "[\"TBD\"]"
+          # },
+          map: |
+            {
+              "dev": {
+                "targets": "[\"dev\"]"
+              },
+              "q1-2024": {
```
atomrc, line 179: This will match both a tag that has `q1-2024` or a release branch with that same pattern

.github/workflows/publish.yml
```diff
@@ -184,10 +157,36 @@ jobs:
...
+                "targets": "[\"q1-2024\"]"
+              }
+            }
+          export_to: log,output
+
   publish_wire_builds:
     name: Bump webapp chart in wire-builds
     runs-on: ubuntu-latest
     needs: build
+    if: ${{needs.build.outputs.wire_builds_target_branches}}
```
atomrc, line 189: This avoids erroring if no target branch is defined (currently erroring when this happens)

.github/workflows/publish.yml
```diff
@@ -381,10 +375,6 @@ jobs:
       release_url: ${{ steps.release.outputs.url }}
 
     steps:
-      - name: Set tag
-        run: |
-          echo "TAG=$(git tag --points-at ${{github.sha}})" >> $GITHUB_ENV
-
```
atomrc, line 387: The tag is removed here in favor of the `release_name` we computed in the first step (the first job is the single source of truth whether we should create a release or not)

aweiss-dev (approved):

PatrykBuniX (approved):
Looks fine

PR #17009, opened 2024-03-07
feat: name check on shields

aweiss-dev:
To ensure the identity of devices and users, we need to compare the users _name_ and _handle_ against the issued End-2-End-Identity Certificate. This needs to be done by the Webapp, since coreCrypto has no access to the current _name_ and _handle_ of a user.

### DIFF
--- a/src/script/components/VerificationBadge/VerificationBadges.tsx
+++ b/src/script/components/VerificationBadge/VerificationBadges.tsx
@@ -81,28 +81,37 @@ const useConversationVerificationState = (conversation: Conversation) => {
   return {MLS: mlsState, proteus: proteusVerificationState};
 };
 
-const useMLSStatus = (identity?: WireIdentity) => {
+const useMLSStatus = (identity?: WireIdentity, userEntity?: User) => {
   const [MLSStatus, setMLSStatus] = useState<MLSStatuses | undefined>(undefined);
+  const [user, setUser] = useState<User | undefined>(userEntity);
   const {current: userState} = useRef(container.resolve(UserState));
 
   useEffect(() => {
     if (!identity) {
-      setMLSStatus(undefined);
       return;
     }
-    void (async () => {
-      const user = await waitFor(() =>
-        userState
-          .users()
-          .find(user => stringifyQualifiedId(user.qualifiedId) === stringifyQualifiedId(identity.qualifiedUserId)),
-      );
-      if (!!user) {
-        const matchingName = identity.displayName === user.name();
-        const matchingHandle = checkUserHandle(identity, user);
-        setMLSStatus(matchingName && matchingHandle ? identity.status : undefined);
-      }
-    })();
-  }, [identity, userState]);
+
+    if (!user) {
+      void (async () => {
+        const user = await waitFor(() =>
+          userState
+            .users()
+            .find(user => stringifyQualifiedId(user.qualifiedId) === stringifyQualifiedId(identity.qualifiedUserId)),
+        );
+        setUser(user);
+      })();
+    }
+  }, [identity, userEntity]);
+
+  useEffect(() => {
+    if (!identity || !user) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


przemvs (approved):

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -76,6 +81,32 @@ const useConversationVerificationState = (conversation: Conversation) => {
...
+    if (!identity) {
+      setMLSStatus(undefined);
+      return;
+    }
+    void (async () => {
+      const user = await waitFor(() =>
+        userState
+          .users()
+          .find(user => stringifyQualifiedId(user.qualifiedId) === stringifyQualifiedId(identity.qualifiedUserId)),
+      );
```
atomrc: in which is the user not there?

atomrc: doesn't this create some blinking on the UI?

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -85,7 +116,10 @@ export const UserVerificationBadges = ({
   groupId?: string;
   isSelfUser?: boolean;
 }) => {
-  const {status: MLSStatus} = useUserIdentity(user.qualifiedId, groupId, isSelfUser);
+  const {deviceIdentities} = useUserIdentity(user.qualifiedId, groupId, isSelfUser);
+  const identity = deviceIdentities ? deviceIdentities[0] : undefined;
+  const {MLSStatus} = useMLSStatus(identity);
```
atomrc: you have access to the `User` at this point, why not giving it to `useMLSStatus` here and avoid one lookup later on?

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -98,7 +132,8 @@ export const DeviceVerificationBadges = ({
   device: ClientEntity;
   getIdentity?: (deviceId: string) => WireIdentity | undefined;
 }) => {
-  const MLSStatus = getIdentity?.(device.id)?.status;
+  const identity = getIdentity?.(device.id);
+  const {MLSStatus} = useMLSStatus(identity);
```
atomrc: maybe you can do the user lookup at this level (and avoid it in the `UserVerificationBadge`, since the user is already given there)

aweiss-dev: can do

src/script/components/VerificationBadge/VerificationBadges.tsx
aweiss-dev: i did change the hook

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -76,6 +81,41 @@ const useConversationVerificationState = (conversation: Conversation) => {
...
+    if (!user) {
+      void (async () => {
+        const user = await waitFor(() =>
+          userState
+            .users()
+            .find(user => stringifyQualifiedId(user.qualifiedId) === stringifyQualifiedId(identity.qualifiedUserId)),
+        );
+        setUser(user);
+      })();
+    }
```
atomrc: nah, this doesn't belong here, the caller should be the one retrieving the `user`
Let's keep this component as stupid as possible

aweiss-dev: sure. can do. 
I don't see the harm since everything that the logic needs is in one place.
But as stupid as it needs is a good way to see things as well.

src/script/components/VerificationBadge/VerificationBadges.tsx
aweiss-dev: I didn't experience that yet, but it theoretically could lead to that. We need to keep an eye on that

PR #17039, opened 2024-03-11
fix(e2ei): evaluate all devices of a user for the mlsStatus

aweiss-dev:
Introduced a bug which didnt check every device of a user when it came to the E2EI verification state.

This PR fixes that logic.

Additionally, we did show a user when one of hi devices had been revoked, expired etc. This was not according to design and gets fixed as well.

### DIFF
--- a/src/script/components/VerificationBadge/VerificationBadges.tsx
+++ b/src/script/components/VerificationBadge/VerificationBadges.tsx
@@ -81,7 +81,7 @@ const useConversationVerificationState = (conversation: Conversation) => {
   return {MLS: mlsState, proteus: proteusVerificationState};
 };
 
-const useMLSStatuses = ({identities, user}: {identities?: WireIdentity[]; user?: User}): MLSStatuses[] | undefined => {
+const getMLSStatuses = ({identities, user}: {identities?: WireIdentity[]; user?: User}): MLSStatuses[] | undefined => {
   if (!identities || !user) {
     return undefined;
   }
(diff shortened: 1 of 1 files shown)


tlebon (approved):

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -103,9 +109,18 @@ export const UserVerificationBadges = ({
...
-  return <VerificationBadges context="user" isProteusVerified={isProteusVerified} MLSStatus={MLSStatus} />;
+  const mlsStatuses = getMLSStatuses({
+    identities: deviceIdentities,
+    user,
+  });
+  const status = !mlsStatuses
+    ? undefined
+    : mlsStatuses.length > 0 && mlsStatuses.every(status => status === MLSStatuses.VALID)
+      ? MLSStatuses.VALID
+      : MLSStatuses.NOT_ACTIVATED;
```
atomrc: wait it can also be `expired` or `revoked` 🤔

atomrc: ah maybe at the user level we don't have such distinction, but then `not activated` seems wrong 🤔

aweiss-dev: Yeah, we don't narrow down on a user level that much. We could have multiple devices with different statuses. What should we display then? In a case where one device is revoked and another is expired.

This piece of code is how it worked before. I agree that NOT_ACTIVATED feels wrong as well.
I think it has been used as a culprit to show the expired icon anyways in that cases.
If something goes wrong with the name validation I dont show any icon

atomrc (approved):

PR #17044, opened 2024-03-11
runfix: handle a crash during change to mixed protocol

PatrykBuniX:
## Description

The flow of updating the conversation's protocol to mixed is:
- update the protocol to mixed
- backend will assign a MLS group_id to the conversation (and set the epoch number to 0)
- client initialises the group and tries to add as many other users and possible
- other users/clients can join a group with an external commit

### Problem
It's possible that the client just updates the protocol and crashes (or just reloads the app). In this case, the mls group was not yet established and the client, when loading the app again, will try joining it with an external commit. This will not work as the group was not established on backend (epoch 0) and it's not possible to fetch the groupInfo. Mixed group would get stuck on epoch 0 and no web client would try to establish it again.

### Solution

Before trying to join the mixed group with an external commit, check it's epoch number, if it's below 1 (eq to 0), instead of joining with ext commit, try establishing it.

## Checklist

### DIFF
--- a/src/script/mls/MLSConversations.ts
+++ b/src/script/mls/MLSConversations.ts
@@ -62,23 +62,49 @@ export async function initMLSGroupConversations(
   );
 
   for (const mlsConversation of mlsGroupConversations) {
-    try {
-      const {groupId, qualifiedId} = mlsConversation;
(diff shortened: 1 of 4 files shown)


atomrc (approved):
Looks good

PR #17048, opened 2024-03-11
runfix: update epoch number locally on epoch change

PatrykBuniX:
## Description

We were never updating the epoch number locally (on conversation entity) after it was advanced inside the core. With this PR we are sure local epoch is always in sync with that's inside core and core-crypto.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -182,7 +182,7 @@ class MLSConversationVerificationStateHandler {
     const conversations = this.conversationState.conversations();
     await Promise.all(conversations.map(conversation => this.checkConversationVerificationState(conversation)));
   };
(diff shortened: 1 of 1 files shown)


aweiss-dev (approved):

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -193,6 +193,7 @@ class MLSConversationVerificationStateHandler {
       return this.logger.warn(`Epoch changed but conversation could not be found after waiting for 5 seconds`);
     }
 
+    conversation.epoch = Number(newEpoch);
```
atomrc, line 196: do we actually consume this info somewhere?

PatrykBuniX: - previously only when initialising a 1:1 conversation
- now also before joining a mixed conversation with an external commit

atomrc (approved):

PR #17056, opened 2024-03-12
runfix: refresh mls conversation state

PatrykBuniX:
## Description

There's no trigger from core-crypto (or anywhere else) when a cert is getting expired. We were evaluating the mls conversation state every time an epoch has changed in a conversation. Since cert expiration does not trigger the epoch change, we would not refresh conversation's state until next epoch update.

This PR adds two triggers when we refresh conversation's mls verification state:
- every tame we navigate to a conversation
- every time when a message is sent

A refresh method is throttled with 3s of delay so we don't fire it too often when spamming messages or spam-clicking on a conversation list item.

When recording a gif below I've mocked the behaviour to always degrade the conversation.
Kapture 2024-03-12 at 14 08 40

## Checklist

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.test.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.test.ts
@@ -26,10 +26,27 @@ import {Core} from 'src/script/service/CoreSingleton';
 import {createUuid} from 'Util/uuid';
 import {waitFor} from 'Util/waitFor';
 
-import {registerMLSConversationVerificationStateHandler} from './MLSStateHandler';
+import {MLSConversationVerificationStateHandler} from './MLSStateHandler';
+
+const createMLSConversationVerificationStateHandler = (
+  domain: string,
+  onConversationVerificationStateChange: OnConversationE2EIVerificationStateChange,
+  onSelfClientCertificateRevoked: () => Promise<void>,
+  conversationState: ConversationState,
+  core: Core,
+) => {
+  return new MLSConversationVerificationStateHandler(
+    domain,
+    onConversationVerificationStateChange,
+    onSelfClientCertificateRevoked,
+    conversationState,
+    core,
+  );
+};
 
 import {ConversationState} from '../../ConversationState';
 import {ConversationVerificationState} from '../../ConversationVerificationState';
+import {OnConversationE2EIVerificationStateChange} from '../shared';
 
 describe('MLSConversationVerificationStateHandler', () => {
   const conversationState = new ConversationState();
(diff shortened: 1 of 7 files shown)


src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -237,19 +236,3 @@ export const checkUserHandle = (identity: WireIdentity, user: User): boolean =>
   const {username, domain} = user;
   return identityHandle.includes(`${username()}@${domain}`);
 };
-
-export const registerMLSConversationVerificationStateHandler = (
```
PatrykBuniX, line 241: This was moved to conversation repo.

src/script/conversation/MessageRepository.ts
```diff
@@ -824,6 +824,10 @@ export class MessageRepository {
       await this.conversationRepositoryProvider().makeSureMLS1to1ConversationIsEstablished(conversation);
     }
 
+    if (isMLS) {
+      void this.conversationRepositoryProvider().refreshMLSConversationVerificationState(conversation);
```
PatrykBuniX: Refresh conversation state before a message is sent in a MLS conversation.

src/script/view_model/ContentViewModel.ts
```diff
@@ -297,6 +298,9 @@ export class ContentViewModel {
         this.conversationState.activeConversation(conversationEntity);
       }
 
+      if (isMLSConversation(conversationEntity)) {
+        void this.conversationRepository.refreshMLSConversationVerificationState(conversationEntity);
+      }
```
PatrykBuniX: Refresh conversation state before navigating to a MLS conversation.

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.test.ts
```diff
@@ -26,10 +26,27 @@ import {Core} from 'src/script/service/CoreSingleton';
...
+  core: Core,
+) => {
+  return new MLSConversationVerificationStateHandler(
+    domain,
+    onConversationVerificationStateChange,
+    onSelfClientCertificateRevoked,
+    conversationState,
+    core,
+  );
+};
```
atomrc: I guess we could just remove this abstraction layer and just expose the class itself, I don't see the value of having the factory function that just wraps the constructor

src/script/conversation/MessageRepository.ts
atomrc: I think we should await that call actually (so that message sending is blocked in case we degrade)

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.test.ts
PatrykBuniX: You're right, done with https://github.com/wireapp/wire-webapp/pull/17056/commits/e36efcce84b0cb7504572becf14bfab4e9df701c

src/script/conversation/MessageRepository.ts
PatrykBuniX: What do you mean by blocked? The refresh method does not throw any errors... Or you mean to have a degraded system message appearing first before the actual message?

PatrykBuniX: Discussed on a different chat, addressed the comment with https://github.com/wireapp/wire-webapp/pull/17056/commits/37b6517a93f70312aa080195ea42dc2cc2740130

atomrc (approved):
looks solid

PR #16810, opened 2024-02-12
refactor: Extract common image functionnalities

atomrc:
## Description

This will harmonize how images are loaded across the app (using the same component (namely `Image`) to load images everywhere)

This should fix quite a few bugs related to images jumping in size when switching states (from loading to loaded for example). 
It should also quite improve scrolling up in a conversation and avoid random jumps

### Scrolling up in the conversation with images (no jump)

https://github.com/wireapp/wire-webapp/assets/1090716/ab8046cf-ff3b-4056-8320-e134e0f1c8fa

### Sending/loading and replying to images

https://github.com/wireapp/wire-webapp/assets/1090716/452bfb3a-2f72-4a8d-bfe2-2ce8f0f3cb4a

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

PR #17063, opened 2024-03-13
fix: side scrolling in message list

aweiss-dev:
## Description

We had an issue with the chat being able to side-scroll.
This is caused by setting a max-image width with a fixed pixel value, sometimes bigger than the viewport. This causes placeholder images of very big images to be rendered with too much width.

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
@@ -125,7 +125,7 @@ export const ImageAsset = ({
   const imageAsset: CSSObject = {
     aspectRatio: isFileSharingReceivingEnabled ? `${asset.ratio}` : undefined,
     maxHeight: '80vh',
-    maxWidth: imageWidth,
+    maxWidth: !imageUrl?.url ? '100%' : imageWidth,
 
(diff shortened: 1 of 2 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
```diff
@@ -120,7 +119,7 @@ export const ImageAsset = ({
   };
 
   const imageStyle: CSSObject = {
-    width: imageWidth,
+    width: 'max-content',
```
PatrykBuniX: **question:** Wouldn't it always make the image full parent's width? Even if we want to send a small image? E.g 50x50px will go full screen size?

atomrc: Very valid question +1

aweiss-dev: No, because it is not used as the width: on the element, instead it was used for max-width:
So smaller images would still display smaller.

But thinking about it, it would overflow for big images. fit-content is what should really be used here. Since it adjusts the image to fit the outer div instead of overflowing

atomrc:
Closed in favor of #16810

PR #17069, opened 2024-03-13
fix: jump on device details - page

aweiss-dev:
## Description

Trough the way we implemented accessibility in our application, by setting interactive elements manually to tabIndex="0", we force the browser after a page transition to focus the first new element. 

If the first element that has a tabIndex="0" is outside of the current viewport, the browser will "jump" there.

### Problem in action:
Problem

### Solution (only for MLS Device with activated E2EI):
**added the missing tabIndex and aria-attributes to the interactive elements (buttons) and verification shields.** 

Solution for E2Ei

### Soltion for Proteus devices:
**Added the missing tabIndex and aria-tooltips to the verification shields**

### DIFF
--- a/src/script/components/VerificationBadge/VerificationBadges.tsx
+++ b/src/script/components/VerificationBadge/VerificationBadges.tsx
@@ -22,6 +22,7 @@ import {CSSProperties, useEffect, useMemo, useRef, useState} from 'react';
 import {CSSObject} from '@emotion/react';
 import {ConversationProtocol} from '@wireapp/api-client/lib/conversation';
 import {stringifyQualifiedId} from '@wireapp/core/lib/util/qualifiedIdUtil';
+import {TabIndex} from '@wireapp/react-ui-kit/lib/types/enums';
 import {container} from 'tsyringe';
 
 import {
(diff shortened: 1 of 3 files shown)


przemvs (approved):

atomrc (approved):

PatrykBuniX (approved):

atomrc (approved):

src/script/page/MainContent/panels/preferences/DevicesPreferences/components/E2EICertificateDetails/E2EICertificateDetails.tsx
```diff
@@ -73,7 +73,7 @@ export const E2EICertificateDetails = ({identity, isCurrentDevice}: E2EICertific
             onClick={() => setIsCertificateDetailsModalOpen(true)}
             data-uie-name="show-certificate-details"
             aria-label={t('E2EI.showCertificateDetails')}
-            tabIndex={TabIndex.FOCUSABLE}
+            //tabIndex={TabIndex.FOCUSABLE}
```
PatrykBuniX: question: Is that intentional? If so, let's just remove it

src/script/components/VerificationBadge/VerificationBadges.tsx
```diff
@@ -243,6 +261,8 @@ export const VerificationBadges = ({
   if (!MLSStatus && !isProteusVerified) {
     return null;
   }
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const id = useRef(new Date().getTime());
```
PatrykBuniX: I believe it's about calling hooks conditionally? Can't we just move it above the first return statement and live in peace with react and eslint? :D

src/script/page/MainContent/panels/preferences/DevicesPreferences/components/E2EICertificateDetails/E2EICertificateDetails.tsx
aweiss-dev: nice find

PR #17071, opened 2024-03-14
fix: clicking on a link in link preview opens it twice [WPB-6817]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6817" title="WPB-6817" target="_blank">BugWPB-6817</a>  [Web - Edge] Clicking on a link preview opens link twice in browser
  </td></table>
  <br />
 

## Description

Clicking directly on a link in a link preview asset was opening the link twice in a new tab - once for link click (default browser behaviour) and once for the onClick event. To solve that we check whether the link element (`<a>`) was clicked directly, if so, no need to open a link manually.

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/LinkPreviewAssetComponent.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/LinkPreviewAssetComponent.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import React from 'react';
+import React, {UIEvent} from 'react';
 
 import cx from 'classnames';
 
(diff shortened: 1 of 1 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/LinkPreviewAssetComponent.tsx
```diff
@@ -123,6 +128,7 @@ const LinkPreviewAsset: React.FC<LinkPreviewAssetProps> = ({header = false, mess
                 rel="noopener noreferrer"
                 title={preview.url}
                 data-uie-name="link-preview-url"
+                ref={linkRef}
```
atomrc, line 131: Question: I'm wondering is a simple `onclick => event.stopPropagation()` would work?

PatrykBuniX: It would but then we need one for key down also

atomrc: Ah true, then fine like this 👌

atomrc (approved):

PR #17073, opened 2024-03-14
runfix: Warn user when applock is disabled on the team level

atomrc:
## Description

Will fix the flow of disabling appLock for an entire team. In this case, individual users should be warned that the lock has been removed. 

## Checklist


src/script/page/components/FeatureConfigChange/FeatureConfigChangeNotifier/FeatureConfigChangeNotifier.ts
```diff
@@ -44,15 +43,14 @@ import {loadFeatureConfig, saveFeatureConfig} from './FeatureConfigChangeNotifie
...
-      newConfig?: FeatureWithoutConfig | Feature<any> | undefined,
-    ) => undefined | {htmlMessage: string; title: StringIdentifer; primaryAction?: Action}
-  >
-> = {
+type FeatureMessageGenerator = {
+  [K in keyof FeatureList]: (
+    oldConfig?: FeatureList[K],
+    newConfig?: FeatureList[K],
+  ) => undefined | {htmlMessage: string; title: StringIdentifer; primaryAction?: Action};
+};
```
atomrc, line 51: This is an improvement of the type, now the `oldConfig` and `newConfig` will be correctly typed with the actual feature config

PatrykBuniX (approved):
Looks good 🚀

src/script/page/components/FeatureConfigChange/FeatureConfigChangeNotifier/FeatureConfigChangeNotifier.ts
```diff
@@ -79,6 +77,18 @@ const featureNotifications: Partial<
       title: 'featureConfigChangeModalAudioVideoHeadline',
     };
   },
+
+  [FEATURE_KEY.APPLOCK]: (oldConfig, newConfig) => {
+    const shouldWarn = oldConfig?.config.enforceAppLock === true && newConfig?.config.enforceAppLock === false;
```
przemvs, line 82: WDYT? :)

```suggestion
    const shouldWarn = !!oldConfig?.config.enforceAppLock && !newConfig?.config.enforceAppLock;
```

atomrc: That would not work, we want to check that they really are `true` or `false`. `undefined` should be excluded

przemvs (approved):

PR #17091, opened 2024-03-18
fix: autofocus modal degraded conversation modal primary button [WPB-7082]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7082" title="WPB-7082" target="_blank">BugWPB-7082</a>  [Web] Conversation no longer verified modal not focused
  </td></table>
  <br />
 

## Description

- replace primary action with secondary in conversation degraded modal when sending a message, so now the primary action is "cancel".
- add autofocus to Primary Modal's primary button, so the primary action can be performed by the user immediately with an entry key press.
- get rid of the behaviour where extra enter is inserted before `onSend` method is handled.

Before
before

After
after

## Checklist

### DIFF
--- a/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
+++ b/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
@@ -50,7 +50,7 @@ export const PrimaryModalComponent: FC = () => {
   const updateErrorMessage = usePrimaryModalState(state => state.updateErrorMessage);
   const updateCurrentModalContent = usePrimaryModalState(state => state.updateCurrentModalContent);
   const currentId = usePrimaryModalState(state => state.currentModalId);
-  const primaryActionButtonRef = useRef<HTMLButtonElement>(null);
+  const primaryActionButtonRef = useRef<HTMLButtonElement | null>(null);
   const isModalVisible = currentId !== null;
   const {
     checkboxLabel,
@@ -216,7 +216,10 @@ export const PrimaryModalComponent: FC = () => {
 
   const primaryButton = !!primaryAction?.text && (
     <button
-      ref={primaryActionButtonRef}
+      ref={ref => {
+        ref?.focus();
+        primaryActionButtonRef.current = ref;
+      }}
       type="button"
       onClick={doAction(confirm, !!closeOnConfirm)}
       disabled={isPrimaryActionDisabled()}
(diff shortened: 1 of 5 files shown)


src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -50,7 +50,7 @@ export const PrimaryModalComponent: FC = () => {
   const updateErrorMessage = usePrimaryModalState(state => state.updateErrorMessage);
   const updateCurrentModalContent = usePrimaryModalState(state => state.updateCurrentModalContent);
   const currentId = usePrimaryModalState(state => state.currentModalId);
-  const primaryActionButtonRef = useRef<HTMLButtonElement>(null);
+  const primaryActionButtonRef = useRef<HTMLButtonElement | null>(null);
```
PatrykBuniX, line 53: Needed to be able to overwrite the ref value.

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -216,7 +216,10 @@ export const PrimaryModalComponent: FC = () => {
 
   const primaryButton = !!primaryAction?.text && (
     <button
-      ref={primaryActionButtonRef}
+      ref={ref => {
+        ref?.focus();
```
PatrykBuniX, line 220: Whenever primary button is rendered, focus is first.

przemvs (approved):

src/script/components/RichTextEditor/plugins/SendPlugin.tsx
```diff
@@ -33,12 +33,20 @@ export function SendPlugin({onSend}: Props): null {
...
           return true;
         }
 
+        // When sending a message with "Enter", we want to prevent the default behavior (new line)
+        event.preventDefault();
         onSend();
-        return false;
+
+        // By returning true, we tell the editor that we've handled the event and it should stop propagation (only for the same command priority level)
+        return true;
```
atomrc, line 49: nice catch 👌

src/script/components/RichTextEditor/plugins/TypeaheadMenuPlugin.tsx
```diff
@@ -453,7 +453,7 @@ function LexicalPopoverMenu<TOption extends TypeaheadOption>({
           selectOptionAndCleanUp(options[selectedIndex]);
           return true;
         },
-        COMMAND_PRIORITY_LOW,
```
atomrc, line 456: doesn't this conflicts with the changes to the send plugin?

PatrykBuniX: Well it would indeed conflict if it was left with the same priority, but now since it was changed to NORMAL (so higher then the send plugin one), it will still be executed.

atomrc (approved):

atomrc (approved):

V-Gira (approved):

PR #17098, opened 2024-03-18
fix: Mitigate UI flickering when quotes are loaded

atomrc:
## Description

Since quoted messages are loaded async, we first display an empty DOM element and then the quote. 
This creates a quite huge bump in the UI when the quote finally loads. 

To mitigate this, we could render a fake quoted message first and then wait for the real message to come in. This will, at least, roughly take up the space that the quote should take in the end (not an exact height, but should at least mitigate flickering)

### Before 

https://github.com/wireapp/wire-webapp/assets/1090716/a40ab6e5-f517-420d-aa1c-fbd62fd41d04

### After

https://github.com/wireapp/wire-webapp/assets/1090716/29a239d9-1538-457f-9b25-c54dea9fe404

## Checklist


PatrykBuniX (approved):
Great improvement! 🚀

PR #17100, opened 2024-03-19
fix: Improve link preview image display

atomrc:
### Before
image

### After
image

## Checklist

### DIFF
--- a/src/style/components/asset/link-preview-asset.less
+++ b/src/style/components/asset/link-preview-asset.less
@@ -18,7 +18,7 @@
  */
 
 // variables
-@link-preview-height: 88px;
+@link-preview-height: 5.5rem;
 
 .link-preview-asset {
   .asset-container-style;
--- a/src/style/components/asset/link-preview-asset.less
+++ b/src/style/components/asset/link-preview-asset.less
@@ -32,7 +32,8 @@
 }
 
 .link-preview-image-container {
-  width: @link-preview-height;
+  min-width: @link-preview-height;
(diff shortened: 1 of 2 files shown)


src/style/components/asset/link-preview-asset.less
```diff
@@ -32,7 +32,8 @@
 }
 
 .link-preview-image-container {
-  width: @link-preview-height;
+  min-width: @link-preview-height;
+  height: @link-preview-height;
```
V-Gira, line 36: let's change the variable to rem, the reason height wasn't fixed was for larger font sizes.

80px would be 5rem, wdyt?

atomrc: Oh that's why we had this strange : `width: @...-height` :D

atomrc: Using rem sounds like a nice solution 👍

V-Gira (approved):
Let's see how that looks :)

src/style/components/asset/link-preview-asset.less
```diff
@@ -18,7 +18,7 @@
  */
 
 // variables
-@link-preview-height: 88px;
+@link-preview-height: 5.5rem;
```
V-Gira, line 21: Oh, I read 80px for some reason, good on ya for double checking :D

PatrykBuniX (approved):

V-Gira (approved):

PR #17107, opened 2024-03-19
fix: condition that decides if we delete local conversations

aweiss-dev:
The functionality expected a customer Error to be thrown, but since we call the apiClient directly and this instance finally throws the 404, we get the default BackendError instead.

I changed the condition to match the case.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2100,7 +2100,7 @@ export class ConversationRepository {
         // We need to check if the conversation exists on backend
         await this.conversationService.getConversationById(inccessibleConversation);
       } catch (error) {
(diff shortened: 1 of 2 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2100,7 +2100,7 @@ export class ConversationRepository {
         // We need to check if the conversation exists on backend
         await this.conversationService.getConversationById(inccessibleConversation);
       } catch (error) {
-        if (error instanceof ConversationError && error.type === ConversationError.TYPE.CONVERSATION_NOT_FOUND) {
+        if (isBackendError(error) && error.label === BackendErrorLabel.NO_CONVERSATION) {
```
aweiss-dev, line 2103: There is no guarantee that we will throw a custom error at this point. So, the solution is to check for a backend error and the desired label.

atomrc (approved):

PR #17109, opened 2024-03-19
runfix: ensure images are always in viewport

V-Gira:
## Description

Automation checks if a vertical image is shown correctly in viewport, it currently fails depending of the size of the window.
a max height of `80vh` will not be adequate if the ui elements with fixed pixed value exeed 20% of the viewport height, we can replace it with `calc(100vh - ammountOfPixels)`

Before:
image

After:
image

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
@@ -50,6 +50,9 @@ export const ImageAsset = ({asset, message, onClick}: ImageAssetProps) => {
 
   const imageContainerStyle: CSSObject = {
     maxWidth: 'var(--conversation-message-asset-width)',
+    // Images should be in the viewport regardless of the window's size
+    // To ensure this, we calculate the maximum height as:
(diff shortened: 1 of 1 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/ImageAsset/ImageAsset.tsx
```diff
@@ -50,7 +50,7 @@ export const ImageAsset = ({asset, message, onClick}: ImageAssetProps) => {
 
   const imageContainerStyle: CSSObject = {
     maxWidth: 'var(--conversation-message-asset-width)',
-    maxHeight: '80vh',
+    maxHeight: 'calc(100vh - 155px - 1rem)',
```
V-Gira, line 56: unfortunately, we don't use css variable for the height of the relevant components, they are:
- Title bar: 45px
- Typing indicator: 40px
- Input Bar: 56px
- Image padding: 12px
- Classified bar: 1rem

153px + 1rem, rounded up to 155px

atomrc: This should appear as a comment in the code ;)

atomrc (approved):

PR #17112, opened 2024-03-20
fix: locking the user out of the app permanentely with applock [WPB-7192, WPB-7193]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7192" title="WPB-7192" target="_blank">BugWPB-7192</a>  [Web] Cancelling setting up voluntary app lock can softlock the account
  </td></table>
  <br />
 

<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7193" title="WPB-7193" target="_blank">BugWPB-7193</a>  [Web] Reloading the app while setting app lock pw does not cancel the process
  </td></table>
  <br />

### DIFF
--- a/src/script/page/AppLock/AppLock.tsx
+++ b/src/script/page/AppLock/AppLock.tsx
@@ -149,12 +149,16 @@ const AppLock: React.FC<AppLockProps> = ({
   useEffect(() => {
     const hasBeenEnabled = !prevAppLockEnabled.current && isAppLockEnabled;
 
+    if (hasBeenEnabled) {
+      showAppLock(APPLOCK_STATE.SETUP);
+      return;
+    }
     if (isAppLockEnabled) {
       if (!appLockState.hasPassphrase()) {
         appLockRepository.setDisabled();
         return;
       }
-      showAppLock(hasBeenEnabled ? APPLOCK_STATE.SETUP : APPLOCK_STATE.LOCKED);
+      showAppLock(APPLOCK_STATE.LOCKED);
     }
     prevAppLockEnabled.current = isAppLockEnabled;
   }, [isAppLockEnabled]);
--- a/src/script/page/AppLock/AppLock.tsx
+++ b/src/script/page/AppLock/AppLock.tsx
@@ -87,7 +87,7 @@ const AppLock: React.FC<AppLockProps> = ({
 
   const MAX_RETRIES = 3;
   const signUserOut = (count?: number) => {
-    if (count && count <= MAX_RETRIES) {
+    if (!!count && count <= MAX_RETRIES) {
       return;
     }
     amplify.publish(WebAppEvents.LIFECYCLE.SIGN_OUT, SIGN_OUT_REASON.USER_REQUESTED);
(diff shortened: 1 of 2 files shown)


src/script/page/AppLock/AppLock.tsx
```diff
@@ -95,10 +104,17 @@ const AppLock: React.FC<AppLockProps> = ({
 
   const {current: modalObserver} = useRef(
     new MutationObserver(() => {
-      const modalInDOM = document.querySelector('[data-uie-name="applock-modal"]');
-      if (!modalInDOM) {
-        amplify.publish(WebAppEvents.LIFECYCLE.SIGN_OUT, SIGN_OUT_REASON.USER_REQUESTED);
-      }
+      let retries = 0;
+
+      const interval = setInterval(() => {
```
aweiss-dev, line 109: Here was a race condition, where the modal entered the view after the first check of mutationObserver. This lead to the user constantly being logged out.

src/script/page/AppLock/AppLock.tsx
```diff
@@ -131,11 +147,16 @@ const AppLock: React.FC<AppLockProps> = ({
...
   useEffect(() => {
+    const hasBeenEnabled = !prevAppLockEnabled.current && isAppLockEnabled;
+
     if (isAppLockEnabled) {
-      showAppLock();
-    } else if (appLockState.hasPassphrase()) {
-      appLockRepository.removeCode();
+      if (!appLockState.hasPassphrase()) {
+        appLockRepository.setDisabled();
+        return;
```
aweiss-dev, line 158: I have added the password check here to prevent the user from being shown the applock without us having a passphrase.

src/script/user/AppLockRepository.ts
```diff
@@ -99,7 +96,12 @@ export class AppLockRepository {
     }
   };
 
-  setCode = async (code: string): Promise<void> => {
+  public setDisabled = () => {
```
aweiss-dev, line 99: There was no way to disable the applock without showing additional modals, now there is

src/script/page/AppLock/AppLock.tsx
atomrc: can you explain a little more what is happening here? Why does the `style` changes after the modal was shown, I don't think I understand 🤔 
Why does it not happen all the time the applock is enabled?

I have the feeling there is a cleaner fix for that

src/script/page/AppLock/AppLock.tsx
```diff
@@ -78,27 +78,42 @@ const AppLock: React.FC<AppLockProps> = ({
...
     'isAppLockActivated',
     'isAppLockEnabled',
     'isAppLockEnforced',
   ]);
 
+  const MAX_RETRIES = 3;
+  const signUserOut = (count?: number) => {
+    if (!!count && count <= MAX_RETRIES) {
+      return;
+    }
```
atomrc, line 92: This check doesn't belong here, it should be done inside the running interval

atomrc (approved):
A few things to consider, but looks fine. 

Let's do a follow up PR that improves how we detect that the DOM has been tempered with

PR #17117, opened 2024-03-20
runfix: update conversation participant list on team member leave event

PatrykBuniX:
## Description

After receiving team member leave event, we were never removing removed users' ids from conversation entity's `participating_user_ids` field and `participating_user_ets` later on, what could result in multiple removal system messages being inserted.

We base the logic of whether to display the message or not on `participating_user_ets` containing the user we have received a remove event for. There are two checks:
- https://github.com/wireapp/wire-webapp/blob/runfix/removed-from-team-message/src/script/conversation/ConversationRepository.ts#L2750
- https://github.com/wireapp/wire-webapp/blob/runfix/removed-from-team-message/src/script/event/preprocessor/EventStorageMiddleware/EventStorageMiddleware.ts#L108

Conversation entity was only update in case of receiving a `conversation.member-leave` event, see https://github.com/wireapp/wire-webapp/blob/f863ba6f479bc191000fea255240a6618b31a110/src/script/conversation/ConversationRepository.ts#L3684

It was never done for `team.member-leave` event. So in case `team.member-leave` was received first, when `conversation.member-leave` was received later, the user was still apparently a conversation member (what was not true).

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2745,19 +2745,35 @@ export class ConversationRepository {
     const userEntity = await this.userRepository.getUserById(userId);
     const eventInjections = this.conversationState
       .conversations()
-      .filter(conversationEntity => {
-        const conversationInTeam = conversationEntity.teamId === teamId;
-        const userIsParticipant = UserFilter.isParticipant(conversationEntity, userId);
-        return conversationInTeam && userIsParticipant && !conversationEntity.removed_from_conversation();
+      .filter(conversation => {
+        const conversationInTeam = conversation.teamId === teamId;
+        const userIsParticipant = UserFilter.isParticipant(conversation, userId);
+        return conversationInTeam && userIsParticipant && !conversation.removed_from_conversation();
       })
-      .map(conversationEntity => {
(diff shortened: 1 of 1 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2745,19 +2745,36 @@ export class ConversationRepository {
...
+        const userIsParticipant = UserFilter.isParticipant(conversation, userId);
+        return conversationInTeam && userIsParticipant && !conversation.removed_from_conversation();
       })
-      .map(conversationEntity => {
-        const leaveEvent = EventBuilder.buildTeamMemberLeave(conversationEntity, userEntity, isoDate);
-        return this.eventRepository.injectEvent(leaveEvent);
+      .map(async conversation => {
+        const leaveEvent = EventBuilder.buildTeamMemberLeave(conversation, userEntity, isoDate);
+        await this.eventRepository.injectEvent(leaveEvent);
+        await this.clearUsersFromConversation(conversation, [userEntity]);
```
PatrykBuniX, line 2756: This is the actual fix, everything else is just refactoring/renaming.

aweiss-dev (approved):

atomrc (approved):
seems solid

PR #17158, opened 2024-03-25
fix: message search reload [WPB-7237]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7237" title="WPB-7237" target="_blank">BugWPB-7237</a>  [Web] Cannot read more messages after leaving the search mode
  </td></table>
  <br />
 

## Description

When being in a search mode, when new messages are sent, we don't load them to `.messages()` ko state directly (as they are messages that are being currently displayed to the user) but we load them to `.incomingMessages`. Also, always when navigation out of conversation we "release" it, meaning we delete all its messages from the memory (mentioned above `.messages()` state).

Previously we were never clearing conversation's messages if it contained any unread events (including messages from `.incomingMessages()`), what caused old searched messages still to be loaded when navigating back to the conversation.

The fix is to always clear conversation's messages when switching to a different conversation, and if it contained any incoming messages, move them to the regular messages (those displayed to the user).

This PR also removes conversation's:
- `is_loaded` state that was set initially to `false` and never updated
- `messages_visible` that was calculated and never consumed

Before
before 01 48 03

After
after 18 33 56

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/entity/Conversation.ts
+++ b/src/script/entity/Conversation.ts
@@ -149,6 +149,7 @@ export class Conversation {
   public readonly legalHoldStatus: ko.Observable<LegalHoldStatus>;
   public readonly localMessageTimer: ko.Observable<number>;
   public readonly messages_unordered: ko.ObservableArray<Message>;
+  // Sorted messages that are ready to be displayed in the conversation
   public readonly messages: ko.PureComputed<Message[]>;
   public readonly messageTimer: ko.PureComputed<number>;
   public readonly name: ko.Observable<string>;
--- a/src/script/entity/Conversation.ts
+++ b/src/script/entity/Conversation.ts
@@ -127,7 +127,7 @@ export class Conversation {
   public readonly lastDeliveredMessage: ko.PureComputed<Message | undefined>;
   public readonly is_archived: ko.Observable<boolean>;
   public readonly is_cleared: ko.PureComputed<boolean>;
-  //* Indicates if the conversation is currently loading messages into its state. */
+  /** Indicates if the conversation is currently loading messages into its state. */
   public readonly isLoadingMessages: ko.Observable<boolean>;
   public readonly is_verified: ko.PureComputed<boolean | undefined>;
   public readonly is1to1: ko.PureComputed<boolean>;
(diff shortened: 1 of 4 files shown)


src/script/entity/Conversation.ts
```diff
@@ -150,7 +149,7 @@ export class Conversation {
   public readonly legalHoldStatus: ko.Observable<LegalHoldStatus>;
   public readonly localMessageTimer: ko.Observable<number>;
   public readonly messages_unordered: ko.ObservableArray<Message>;
-  public readonly messages_visible: ko.PureComputed<Message[]>;
+  // Sorted messages that are ready to be displayed in the conversation
```
atomrc:
```suggestion
  /** Sorted messages that are ready to be displayed in the conversation */
```

This will allow this to be parsed as JS doc and shown in the editor ;)

src/script/entity/Conversation.ts
```diff
@@ -204,7 +203,6 @@ export class Conversation {
     this.teamId = undefined;
     this.type = ko.observable();
 
-    this.is_loaded = ko.observable(false);
     this.is_pending = ko.observable(false);
```
atomrc: If you've understood what `is_pending` is about, could you document it with JS doc (I'm currently not sure what this state is about)?

atomrc (approved):
nice one

src/script/entity/Conversation.ts
PatrykBuniX: Added a comment https://github.com/wireapp/wire-webapp/pull/17158/commits/cc00a27937577f5fdf672354e0af6f4c263fbe0c. We set it while messages are being loaded into the conversation' state. With that we later decide if we sometimes abort loading them again if the process is ongoing. I think this could be renamed to `isLoadingMessages`.

src/script/entity/Conversation.ts
```diff
@@ -127,7 +127,7 @@ export class Conversation {
   public readonly lastDeliveredMessage: ko.PureComputed<Message | undefined>;
   public readonly is_archived: ko.Observable<boolean>;
   public readonly is_cleared: ko.PureComputed<boolean>;
-  public readonly is_loaded: ko.Observable<boolean>;
+  //* Indicates if the conversation is currently loading messages into its state. */
```
atomrc: typo here that should be `/**` not `//*` :D

src/script/entity/Conversation.ts
atomrc: Yep, this name seems better indeed 👍

src/script/entity/Conversation.ts
PatrykBuniX: Good catch! https://github.com/wireapp/wire-webapp/pull/17158/commits/904d0055f1035332453a7468da3a2c263a560840

PR #17159, opened 2024-03-25
chore: Monorepo

atomrc:
This is a first step into having a monorepo structure for all our apps and packages. 

The long run idea is to have 
- https://github.com/wireapp/wire-account 
- https://github.com/wireapp/wire-desktop
- https://github.com/wireapp/wire-webapp

in the same repo along with the custom packages they all consume. 

### Scope

This first step will:
- use only `yarn` as the monorepo manager
- only move the `webapp` to the monorepo (libraries will be added in a follow up PR)
- make sure CI and deploy jobs work fine with the new structure

What is not included in this iteration:
- caching (a high level library like turborepo/lerna/nx/bit... will be included later)
- migrating libraries to the monorepo
- migrating other end-user apps to the monorepo

### TODO

- [x] restore deployment script
- [x] restore wire-builds repository update
- [ ] make sure dependabot config is still working with the new setup


.yarn/plugins/@yarnpkg/plugin-typescript.cjs
atomrc, line 1: this plugin is now included with yarn 4.0.1

apps/wire-webapp/.lintstagedrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "*": "prettier --ignore-unknown --write",
+  "*.less": ["stylelint --fix"],
+  "*.{js,jsx,ts,tsx}": ["eslint --fix"]
+}
```
atomrc, line 5: This is the way to go for a monorepo structure (see https://github.com/lint-staged/lint-staged?tab=readme-ov-file#how-to-use-lint-staged-in-a-multi-package-monorepo)

PR #17164, opened 2024-03-26
refactor: Use core to decrypt assets

atomrc:
## Description

The core already implements encryption and decryption of assets. We should use the methods inside of the core instead of the deprecated methods from the webapp. 

This will also unlock using node > 18 to run tests and CI

see https://github.com/wireapp/wire-web-packages/pull/6110

## Checklist


src/script/assets/AssetRepository.ts
```diff
@@ -146,11 +137,17 @@ export class AssetRepository {
   }
 
   private loadBuffer(asset: AssetRemoteData) {
-    const request = this.core.service!.asset.downloadAsset(asset.urlData, fraction => {
+    const isEncryptedAsset = !!asset.otrKey && !!asset.sha256;
+    const progressCallback = (fraction: number) => {
       asset.downloadProgress(fraction * 100);
-    });
-    asset.cancelDownload = request.cancel;
```
V-Gira, line 152: Is this still being handled?

PatrykBuniX (approved):

src/script/assets/AssetRepository.ts
atomrc: Good catch, I think that's an oversight on my end 👁️

atomrc: @V-Gira fixed with https://github.com/wireapp/wire-webapp/pull/17164/commits/fd0a464096143027d3dcac91da5b4218e9a2885d

V-Gira (approved):

src/script/assets/AssetRepository.ts
V-Gira: nice :)

PR #17168, opened 2024-03-26
chore: Improve eslint and lint-staged workflows

atomrc:
## Description

This is in preparation of the monorepo structure. 
- We need to make the `lint-staged` config on a per-project basis (which means that we need a config file per project)
- we also need the `eslintconfig` to be aware of the current project it's in (thus the `__dirname` in the config). 

One nice addition is that new aliases are automatically detected from the `tsconfig` and not manually added to the `eslint` config


.eslintrc.js
```diff
@@ -0,0 +1,64 @@
...
+    'prefer-promise-reject-errors': 'off',
+    'valid-jsdoc': 'off',
+    'jest/no-jasmine-globals': 'off',
+  },
+  settings: {
+    'import/resolver': {
+      typescript: {
+        project: [__dirname + '/tsconfig.build.json', __dirname + './server/tsconfig.json'],
+      },
+    },
```
atomrc, line 53: This the part that has changed from the original config

.eslintrc.js
```diff
@@ -0,0 +1,64 @@
...
+    'server/dist/',
+    'server/node_modules/',
+    'src/ext/',
+    'src/script/localization/**/webapp*.js',
+    'src/worker/',
+    '*.js',
+  ],
+  parserOptions: {
+    project: ['./tsconfig.build.json', './server/tsconfig.json'],
+    tsconfigRootDir: __dirname,
```
atomrc, line 27: This line was also added to the config

.eslintrc.js
atomrc, line 1: We needed to make this file a `js` file cause it needs to access the `__dirname` value

przemvs (approved):
LGTM! :)

PR #17179, opened 2024-03-28
runfix: address darkmode color values

V-Gira:
## Description

This addresses:
- toggle sidebar button
- folder background
- Unread message badge
- Selected list item background color
- Lack of contrast of the gray font color (also fixed in a few other part of the app that went unnoticed)

Before:
Screenshot from 2024-03-28 17-42-49
After:
image

### DIFF
--- a/src/style/list/conversations.less
+++ b/src/style/list/conversations.less
@@ -57,9 +57,9 @@
       height: 24px;
       align-items: center;
       justify-content: center;
-      border: 1px solid #dce0e3;
+      border: 1px solid var(--border-color);
       border-radius: 20px;
-      background-color: #fff;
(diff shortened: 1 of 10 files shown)


src/style/list/conversations.less
```diff
@@ -197,24 +197,14 @@
...
-
-  .conversations-sidebar-btn--badge {
-    padding: 1px 6px;
-    border: 1px solid var(--white);
-    border-radius: 5px;
-    background: @red-500;
-    color: var(--white);
-    font-size: @font-size-small;
-    font-weight: @font-weight-semibold;
-  }
```
V-Gira, line 217: This block was duplicated, I moved it at a higher level

atomrc (approved):

PR #17183, opened 2024-03-29
feat: push to talk mvp

PatrykBuniX:
## Description

This PR introduces the MVP version of push-to-talk functionality.

Whenever a call is established (whether we start it or answer an incoming one), we start subscribing to keyboard events on window. Whenever space key (currently it's just spacebar, likely to extend to custom config in the future) is pressed and held, we unmute the self participant, when the key is released, we go back to muted state.

## Checklist


src/script/util/KeyboardUtil.ts
```diff
@@ -78,7 +78,10 @@ export const isMetaKey = (keyboardEvent: KeyboardEvent): boolean =>
 export const isPasteAction = (keyboardEvent: KeyboardEvent): boolean =>
   isMetaKey(keyboardEvent) && isKey(keyboardEvent, KEY.KEY_V);
 
-export const isRemovalAction = (key: string): boolean => [KEY.BACKSPACE, KEY.DELETE].includes(key);
+const removalKeys: string[] = [KEY.BACKSPACE, KEY.DELETE];
+export const isRemovalAction = (key: string): boolean => removalKeys.includes(key);
+
+export const isSpaceOrEnterKey = (key: string): boolean => key === KEY.SPACE || key === KEY.ENTER;
```
przemvs, line 84:
```suggestion
export const isSpaceOrEnterKey = (key: string): boolean => [KEY.SPACE, KEY.ENTER].includes(key);
```

PatrykBuniX: I wanted to do this, but TS doesn't like it because

`Argument of type 'string' is not assignable to parameter of type '"Enter" | " "'.ts(2345)`, 

so I did it like this not to cast the type ;)

src/script/calling/CallingRepository.ts
```diff
@@ -268,6 +272,22 @@ export class CallingRepository {
...
+
+    const activeCall = this.callState.joinedCall();
+
+    if (!activeCall) {
+      return;
+    }
+
+    this.subscribeToPushToTalk(key, activeCall);
+  };
+
```
atomrc: Hummm I was thinking this feature could live only in the UI layer and not have any impact on the CallingRepository. Is there anything needed in the calling repo at all? 

My idea was:

In the calling UI component, when mounted with an ongoing call => activate a react hook that would toggle the mic on/off
When unmounted unsubscribe. 

Would that be possible? 

I'd like to have the least amount of changes in the callingRepo

src/script/calling/CallingRepository.ts
```diff
@@ -1430,6 +1481,10 @@ export class CallingRepository {
       await this.leaveMLSConference(conversationId);
     }
 
+    // Remove all the tasks related to the call
+    callingSubscriptions.removeCall(conversationId);
+    this.unsubscribeFromPushToTalk();
```
atomrc: This would be handled by React's unmount

src/script/hooks/usePushToTalk/usePushToTalk.ts
```diff
@@ -0,0 +1,84 @@
...
+import {handleKeyPress} from 'Util/KeyboardUtil';
+
+interface PushToTalk {
+  key: string | null;
+  toggleMute: (shouldMute: boolean) => void;
+  isMuted: () => boolean;
+}
+
+const subscribe = ({key, toggleMute, isMuted}: PushToTalk & {key: string}) => {
+  let wasUnmutedWithKeyPress = false;
```
atomrc: This should probably be a ref (in case the component re-renders in between a press and a release). 
Unlikely scenario, I admit, but still let cover it just in case

src/script/hooks/usePushToTalk/usePushToTalk.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import {handleKeyPress} from 'Util/KeyboardUtil';
+
+interface PushToTalk {
+  key: string | null;
+  toggleMute: (shouldMute: boolean) => void;
+  isMuted: () => boolean;
+}
+
+const subscribe = ({key, toggleMute, isMuted}: PushToTalk & {key: string}) => {
```
atomrc:
```suggestion
const subscribeToKeyPress = ({key, toggleMute, isMuted}: PushToTalk & {key: string}) => {
```

Let's be a tiny bit more specific here :)

src/script/hooks/usePushToTalk/usePushToTalk.ts
PatrykBuniX: Done with https://github.com/wireapp/wire-webapp/pull/17183/commits/5bea16393a3eb15b91c6ef59e4ab2ed467144282

atomrc (approved):

PR #17204, opened 2024-04-03
runfix: team feature config updates

PatrykBuniX:
## Description

We should update the local state with new team feature config before dispatching the feature config updated event in case there are some event listeners that do some calculations based on a recent feature config state (e.g. currently recalculating self supported protocols after team supported protocols were updated).

## Checklist

### DIFF
--- a/src/script/team/TeamRepository.test.ts
+++ b/src/script/team/TeamRepository.test.ts
@@ -111,6 +111,12 @@ describe('TeamRepository', () => {
 
       jest.spyOn(teamService, 'getAllTeamFeatures').mockResolvedValue(featuresFromBackend);
 
+      teamRepo.on('featureConfigUpdated', update => {
+        expect(update.prevFeatureList).toEqual(localFeatures);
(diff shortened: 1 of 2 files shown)


atomrc (approved):
Make sense, good catch 👌

src/script/team/TeamRepository.test.ts
```diff
@@ -86,6 +86,47 @@ describe('TeamRepository', () => {
...
+          status: FeatureStatus.ENABLED,
+        },
+      } as FeatureList;
+
+      jest.spyOn(teamService, 'getAllTeamFeatures').mockResolvedValue(featuresFromBackend);
+
+      teamRepo.on('featureConfigUpdated', update => {
+        expect(update.prevFeatureList).toEqual(localFeatures);
+        expect(update.newFeatureList).toEqual(featuresFromBackend);
+        expect(teamState.teamFeatures()).toEqual(featuresFromBackend);
```
PatrykBuniX, line 117: team feature state should be already updated at this point

PR #17208, opened 2024-04-03
feat: Ability to Zoom in/out full screen image MVP

przemvs:
## Description

In this PR zoom for image was prepared. We have possibility to zoom image to natural image width and height. If image is zoomable we will see **zoom-in** cursor button. You can zoom image while You in FullScreen image. If we want add possibility to zoom by scale we can prepare it in another MVP.

We need to check if we need some buttons to change zoom value.

https://github.com/wireapp/wire-webapp/assets/13432884/26e0f070-a592-44b9-b354-24d25e9d0b27

## Checklist

- use (x) data
- be sure to look at ...


src/script/components/Modals/DetailViewModal/index.tsx
```diff
@@ -234,16 +246,91 @@ const DetailViewModal: FC<DetailViewModalProps> = ({
...
     };
+
     document.addEventListener('keydown', onKeyDown);
+
     return () => {
       document.removeEventListener('keydown', onKeyDown);
     };
   }, []);
 
+  const onButtonClick = (event: React.MouseEvent<HTMLElement>) => {
```
atomrc: Let's try to extract that entire functionality into its own component that will wrap the `` tag, WDYT? 
This will leave the `DetailViewModal` component rather small and readable

przemvs: Done :)

src/script/components/ZoomableImage/ZoomableImage.tsx
```diff
@@ -0,0 +1,128 @@
...
+const ZOOM_IMAGE_SCALE = 3;
+
+interface ZoomableImageProps {
+  src: string;
+  id: string;
+  alt?: string;
+  className?: string;
+}
+
+export const ZoomableImage = ({src, id, alt = '', className = '', ...props}: ZoomableImageProps) => {
```
atomrc:
```suggestion
export const ZoomableImage = (props: ZoomableImageProps) => {
```

you don't need to destructure here, right?

src/script/components/ZoomableImage/ZoomableImage.tsx
```diff
@@ -0,0 +1,128 @@
...
+ *
+ */
+
+import React, {useEffect, useRef, useState} from 'react';
+
+import cx from 'classnames';
+
+const ZOOM_IMAGE_SCALE = 3;
+
+interface ZoomableImageProps {
```
atomrc:
```suggestion
interface ZoomableImageProps extends extends HTMLProps<HTMLImageElement> {
```

src/script/components/ZoomableImage/ZoomableImage.tsx
przemvs: image

I'm getting this while not pass alt to the image as prop

przemvs: Also I need id for a key, to rerender image to avoid zoom in next/prev image

atomrc: you can do `useId` from react, right? I don't think it's a good idea to use an html `id` as a key, they have 2 distinct roles they should not be confused 

Try with the suggestion here, it should fix the warning you see  https://github.com/wireapp/wire-webapp/pull/17208/files/757c10cbb24912667584a7ddcbca5b4ec5a2a138#r1549687274

src/script/components/ZoomableImage/ZoomableImage.tsx
```diff
@@ -0,0 +1,128 @@
...
+  src: string;
+  id: string;
+  alt?: string;
+  className?: string;
```
atomrc:
```suggestion
```

src/script/components/ZoomableImage/ZoomableImage.tsx
przemvs: Ahh.. You're right with useId.

This extends doesn't help

src/script/components/ZoomableImage/ZoomableImage.tsx
przemvs: What with this case? 
` className={cx(className, {zoomed: isZoomEnabled})}`
Change it to:
` className={cx(props.className, {zoomed: isZoomEnabled})}` ?

src/script/components/ZoomableImage/ZoomableImage.tsx
```diff
@@ -0,0 +1,116 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import React, {HTMLProps, useRef, useState} from 'react';
+
+import cx from 'classnames';
+
+const ZOOM_IMAGE_SCALE = 3;
```
atomrc: Wait, I'm not sure about this. We should zoom to 100% of the image width, not to some ratio of the container's width 🤔 
This means we could magnify images further than 100%, I don't think we want that

Small images should not be zoomable

atomrc (approved):
LGTM 😎

PR #17211, opened 2024-04-03
chore: Upgrade coreCrypto to rc55 [WPB-7400]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7400" title="WPB-7400" target="_blank">TaskWPB-7400</a>  [web] upgrade coreCrypto to rc55
  </td></table>
  <br />

### DIFF
--- a/src/script/properties/PropertiesRepository.ts
+++ b/src/script/properties/PropertiesRepository.ts
@@ -78,6 +78,7 @@ export class PropertiesRepository {
       enable_debugging: false,
       settings: {
         call: {
+          push_to_talk_key: '',
           enable_soundless_incoming_calls: false,
           enable_vbr_encoding: true,
         },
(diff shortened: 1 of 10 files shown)


src/script/properties/PropertiesRepository.ts
```diff
@@ -78,6 +78,7 @@ export class PropertiesRepository {
       enable_debugging: false,
       settings: {
         call: {
+          push_to_talk_key: '',
```
atomrc, line 81: in order to use the latest core version, I need to adapt the changes that happened since then (thus this unrelated change). 

In the near future this will not happen as we will version the core along with the webapp's code (monorepo)

thisisamir98 (approved):

V-Gira (approved):

PR #17273, opened 2024-04-15
feat: add literal string replacement support for react localizer util

PatrykBuniX:
## Description

Adds a support for replacing string literal values in `replaceReactComponents` util. This will make it easier for us to migrate from `dangerouslySetInnerHtml` in cases we needed it to inject dynamic values into copy templates. 

## Checklist


atomrc (approved):
solid PR 😍

PR #17278, opened 2024-04-16
runfix: Make sure newly created 1to1 conversation have updated users

atomrc:
## Description

This fixes a bug where, if you have multiple connection requests, then accepting one will not add the newly created conversation in the conversation list instantly.

### Before

https://github.com/wireapp/wire-webapp/assets/1090716/f5d7476a-7acb-4585-8285-7a81d2dd7321

### After

https://github.com/wireapp/wire-webapp/assets/1090716/4813a669-6af2-4b8e-97bc-57d5541e1940

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -1306,7 +1306,9 @@ export class ConversationRepository {
     const connection = user.connection();
     if (connection) {
       this.logger.log(`There's a connection with user ${userId.id}, getting a 1:1 conversation for the connection`);
-      return this.get1to1ConversationForConnection(connection, options);
+      const conversation = await this.get1to1ConversationForConnection(connection, options);
+      // In case we got a conversation back, we make sure the participating user entities are up to date
+      return conversation ? this.updateParticipatingUserEntities(conversation) : null;
     }
 
     const {protocol, isMLSSupportedByTheOtherUser, isProteusSupportedByTheOtherUser} =


V-Gira (approved):

src/script/conversation/ConversationRepository.ts
```diff
@@ -1306,7 +1306,9 @@ export class ConversationRepository {
     const connection = user.connection();
     if (connection) {
       this.logger.log(`There's a connection with user ${userId.id}, getting a 1:1 conversation for the connection`);
-      return this.get1to1ConversationForConnection(connection, options);
+      const conversation = await this.get1to1ConversationForConnection(connection, options);
+      // In case we got a conversation back, we make sure the participating user entities are up to date
+      return conversation ? this.updateParticipatingUserEntities(conversation) : null;
```
PatrykBuniX, line 1311: Could this be checked inside the method returning the conversation entity already?

atomrc: Well I wanted the side effect to be as high as possible in the call stack. As much as possible I try to isolate side effects to the highest level possible. 

This seemed like the right layer to do this. 
Also makes the `get1to1ConversationForConnection` more focused (it just gets the conversation and doesn't try to be too clever about it)

PatrykBuniX: Makes sense 👌

PatrykBuniX (approved):

PR #17279, opened 2024-04-16
refactor: use call state view mode instead of multitasking

PatrykBuniX:
## Description

Removes `multitasking` state and adds `viewMode` to `callState`. It simplifies the current calling view logic and makes it easier to extend the ui with more "modes" in the future.

## Checklist


atomrc (approved):
nice

PR #17283, opened 2024-04-17
refactor: conversation entity in call entity

PatrykBuniX:
## Description

This PR removes the call entity from conversation entity (as it was not used) and adds a replaces `conversationId` field with a `conversation` field in a call entity. 

Conversation without a call is possible.

Call without a conversation is not possible. This removes the need for querying a conversation in a local state based on its id.

## Checklist


atomrc (approved):
Makes a lot of sense

PR #17286, opened 2024-04-17
runfix: Properly send read receipts for 1to1 conversation asset messages

atomrc:
## Description

This will make sure when sending assets in a 1:1 conversation that we set the `expect_read_confirmation` message header properly. 
Also fixes a regression where read-receipt would be sent even though the sender did not request a read receipt.

### DIFF
--- a/src/script/conversation/MessageRepository.ts
+++ b/src/script/conversation/MessageRepository.ts
@@ -604,7 +604,6 @@ export class MessageRepository {
   private async sendAssetRemotedata(conversation: Conversation, file: Blob, messageId: string, asImage: boolean) {
     const retention = this.assetRepository.getAssetRetention(this.userState.self(), conversation);
     const options = {
-      expectsReadConfirmation: this.expectReadReceipt(conversation),
       legalHoldStatus: conversation.legalHoldStatus(),
       public: true,
       retention,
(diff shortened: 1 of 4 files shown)


src/script/conversation/MessageRepository.ts
```diff
@@ -604,7 +604,6 @@ export class MessageRepository {
   private async sendAssetRemotedata(conversation: Conversation, file: Blob, messageId: string, asImage: boolean) {
     const retention = this.assetRepository.getAssetRetention(this.userState.self(), conversation);
     const options = {
-      expectsReadConfirmation: this.expectReadReceipt(conversation),
```
atomrc, line 607: That was an old mistake, the header should not have been set on the upload options, but rather when creating the payload of the message

PatrykBuniX (approved):

przemvs (approved):
LGTM!

V-Gira (approved):

src/script/event/preprocessor/ReceiptsMiddleware.ts
```diff
@@ -49,10 +51,12 @@ export class ReceiptsMiddleware implements EventMiddleware {
       case ClientEvent.CONVERSATION.LOCATION:
       case ClientEvent.CONVERSATION.MESSAGE_ADD: {
         const qualifiedConversation = event.qualified_conversation || {domain: '', id: event.conversation};
-        return this.conversationRepository.getConversationById(qualifiedConversation).then(conversation => {
-          event.data.expects_read_confirmation = this.conversationRepository.expectReadReceipt(conversation);
-          return event;
-        });
+        const conversation = await this.conversationRepository.getConversationById(qualifiedConversation);
+        if (conversation?.isGroup()) {
```
V-Gira, line 55: just point ting that out, any reason not to go back to
```suggestion
        if (conversation && conversation?.isGroup()) {
```
see here https://github.com/wireapp/wire-webapp/pull/7288/files

V-Gira: nevermind was just a work around without the `conversation?` question mark

PR #17292, opened 2024-04-19
fix: handle error when creating a conversation with a bot

PatrykBuniX:
## Description

If we fail adding a service to newly created 1:1 conversation with bot, we should deleted the conversation so the empty group conversation does not appear in the UI.

## Checklist

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -2464,8 +2464,7 @@ export class ConversationRepository {
    */
   public async addServiceToExistingConversation(
     conversationEntity: Conversation,
-    providerId: string,
-    serviceId: string,
+    {providerId, serviceId}: {providerId: string; serviceId: string},
   ) {
     try {
       await this.addService(conversationEntity, {providerId, serviceId});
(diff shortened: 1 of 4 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -2387,6 +2387,49 @@ export class ConversationRepository {
...
+   * @param serviceId serviceId ID of the service
+   * @param providerId providerId ID of the provider
+   * @returns Resolves when conversation with the integration was created
+   */
+  async create1to1ConversationWithService({
+    providerId,
+    serviceId,
+  }: {
+    providerId: string;
+    serviceId: string;
```
przemvs: You can move this props to type, You reusing it also in addService method :)

przemvs (approved):
LGTM, 1 thing with same types :)

src/script/conversation/ConversationRepository.ts
PatrykBuniX: Addressed with https://github.com/wireapp/wire-webapp/pull/17292/commits/975eceeb6055352f16a8e1efc1a06866ecb4ea3c

przemvs (approved):

PR #17293, opened 2024-04-19
feat: picture in picture calling PoC [WPB-7221]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7221" title="WPB-7221" target="_blank">SpikeWPB-7221</a>  [Web] Investigate and Create POC for Picture-in-Picture Calling UI pop-out
  </td></table>
  <br />
 

## Description

The POC for Picture-in-Picture Calling UI pop-out.

There's a new calling view mode added - `DETACHED_WINDOW`, when enabled, calling ui will be gone from the webapp and new window with the calling ui will be displayed instead.

Currently, for PoC there's no support for:
- call participant list actions
- choosing a screen to share

as those need some more (product/design) investigation (where those items should be put in the PiP view).

For browsers that do not support opening an external window, the calling ui will be opened in the new browser tab instead.

The feature is hidden behind the debug util, to enable it, call `wire.app.debug.toggleDetachedCallFeature(true)` and then start/join the call (or just force the component to rerender).

Desktop wrapper:
Screenshot 2024-04-23 at 14 37 32

Chrome:
Screenshot 2024-04-23 at 14 36 17

Arc (not supporting window popups - it will open in a new tab):
Screenshot 2024-04-23 at 14 34 17

## Checklist


src/script/components/DetachedWindow/DetachedWindow.tsx
```diff
@@ -0,0 +1,101 @@
...
+      `,
+    );
+  }, [height, name, url, width]);
+
+  useEffect(() => {
+    if (!newWindow) {
+      return () => {};
+    }
+
+    setTimeout(() => copyStyles(window.document, newWindow.document), 0);
```
atomrc, line 65: isn't there a dom event that we could listen to something like dom ready or something along those lines? 
This `setTimeout` seems a little hacky

src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -149,6 +149,7 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
           <Video
             autoPlay
             playsInline
+            muted // This is needed to keep playing the video when detached to a new window
```
atomrc: could you link some external doc that reference this and the reason why?

PatrykBuniX: sure, will do

atomrc (approved):
A few things to consider, but looking good overall 👌

src/script/components/DetachedWindow/DetachedWindow.tsx
PatrykBuniX: I think the new window being opened on a different domain is an issue here, seems like events like this are not fired in this case.

atomrc: Makes sense. Let's add inline doc to this then :)

src/script/components/calling/GroupVideoGrid.tsx
```diff
@@ -164,6 +164,10 @@ const GroupVideoGrid: React.FunctionComponent<GroupVideoGripProps> = ({
             className="group-video__thumbnail-video"
             autoPlay
             playsInline
+            // This is needed to keep playing the video when detached to a new window,
+            // only muted video can be played automatically without user interacting with the window first,
+            // see https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide.
+            muted
```
atomrc: How about using multi-line comments for this `/* ... */`?

src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -149,6 +149,10 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
           <Video
             autoPlay
             playsInline
+            // This is needed to keep playing the video when detached to a new window,
+            // only muted video can be played automatically without user interacting with the window first,
+            // see https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide.
```
atomrc: How about using multi-line comments for this /* ... */?

src/script/components/calling/GroupVideoGrid.tsx
PatrykBuniX: Sure 👌

atomrc (approved):
LGTM 😎

PR #17305, opened 2024-04-22
runfix: Avoid race condition when team is deleted

atomrc:
## Description

When a team is deleted, we receive a `team.delete` event. This one is parsed and, currently, triggers an async, not awaited, process that will kill the database and the connection to the websocket. 
Because this process is not awaited (since it's a simple event fanned out through `amplify`), then other subsequent events can come in a be handled. 

This could cause a subsequent event to try to access the DB, while it's being deleted. 

The fix is to properly inject a callcack in the `TeamRepository` that will be called when the user is being deleted. 
We can then cleanly await this callback and not trigger any other event handling before this event is completely handled

## Checklist

### DIFF
--- a/src/script/team/TeamRepository.ts
+++ b/src/script/team/TeamRepository.ts
@@ -277,7 +277,7 @@ export class TeamRepository extends TypedEventEmitter<Events> {
         break;
       }
       case TEAM_EVENT.MEMBER_LEAVE: {
-        this.onMemberLeave(eventJson);
+        await this.onMemberLeave(eventJson);
         break;
       }
       case TEAM_EVENT.FEATURE_CONFIG_UPDATE: {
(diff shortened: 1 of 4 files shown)


src/script/team/TeamRepository.ts
```diff
@@ -351,13 +351,11 @@ export class TeamRepository extends TypedEventEmitter<Events> {
...
 
-  private onDelete(eventJson: TeamDeleteEvent | TeamMemberLeaveEvent): void {
+  private async onDelete(eventJson: TeamDeleteEvent | TeamMemberLeaveEvent) {
     const {team: teamId} = eventJson;
     if (this.teamState.isTeam() && this.teamState.team().id === teamId) {
       this.teamState.isTeamDeleted(true);
-      window.setTimeout(() => {
-        amplify.publish(WebAppEvents.LIFECYCLE.SIGN_OUT, SIGN_OUT_REASON.ACCOUNT_DELETED, true);
-      }, 50);
+      await this.onMemberDetete();
```
atomrc, line 358: We can now properly await for this operation to be successful before we process the next event

PatrykBuniX (approved):
Nice 😄

V-Gira (approved):

PR #17306, opened 2024-04-22
fix: prevent SSO login redirect for guest conversations (8827)

V-Gira:
#### Issue:
Users from a backend that provides default SSO code were not able to invite guest to conversations, as the guest login would be skipped by the SSO redirection

#### Solution:
- we prevent the redirection to the sso login if a conversation code is provided
- we add a SSO login button to the guest login flow

#### Before:
[Screencast from 2024-04-22 15-29-18.webm](https://github.com/wireapp/wire-webapp/assets/78490891/f57ee636-6839-4e8a-abeb-6ab8c29e2a88)

#### After:
Guest login page:
Screenshot from 2024-04-22 17-18-57

### DIFF
--- a/src/script/auth/page/Login.tsx
+++ b/src/script/auth/page/Login.tsx
@@ -123,6 +123,9 @@ const LoginComponent = ({
 
   const features = Config.getConfig().FEATURE;
 
+  const showBackButton =
+    !embedded && (features.ENABLE_DOMAIN_DISCOVERY || features.ENABLE_SSO || features.ENABLE_ACCOUNT_REGISTRATION);
+
   const [showEntropyForm, setShowEntropyForm] = useState(false);
   const isEntropyRequired = features.ENABLE_EXTRA_CLIENT_ENTROPY;
   const onEntropyGenerated = useRef<((entropy: Uint8Array) => void) | undefined>();
@@ -350,12 +353,11 @@ const LoginComponent = ({
 
   return (
     <Page>
-      {!embedded &&
-        (features.ENABLE_DOMAIN_DISCOVERY || features.ENABLE_SSO || features.ENABLE_ACCOUNT_REGISTRATION) && (
-          <IsMobile>
-            <div style={{margin: 16}}>{backArrow}</div>
-          </IsMobile>
-        )}
+      {showBackButton && (
+        <IsMobile>
+          <div style={{margin: 16}}>{backArrow}</div>
+        </IsMobile>
+      )}
       {isEntropyRequired && showEntropyForm ? (
         <EntropyContainer onSetEntropy={storeEntropy} />
       ) : (
@@ -378,9 +380,7 @@ const LoginComponent = ({
             {!embedded && (
               <IsMobile not>
                 <Column style={{display: 'flex'}}>
-                  {(features.ENABLE_DOMAIN_DISCOVERY ||
-                    features.ENABLE_SSO ||
-                    features.ENABLE_ACCOUNT_REGISTRATION) && <div style={{margin: 'auto'}}>{backArrow}</div>}
+                  {showBackButton && <div style={{margin: 'auto'}}>{backArrow}</div>}
                 </Column>
               </IsMobile>
             )}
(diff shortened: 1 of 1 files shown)


src/script/auth/page/Login.tsx
```diff
@@ -120,11 +121,18 @@ const LoginComponent = ({
...
   const [showEntropyForm, setShowEntropyForm] = useState(false);
-  const isEntropyRequired = Config.getConfig().FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY;
+  const isEntropyRequired = features.ENABLE_EXTRA_CLIENT_ENTROPY;
   const onEntropyGenerated = useRef<((entropy: Uint8Array) => void) | undefined>();
   const entropy = useRef<Uint8Array | undefined>();
 
+  const queryConversationCode = UrlUtil.getURLParameter(QUERY_KEY.CONVERSATION_CODE) || null;
+  const queryConversationKey = UrlUtil.getURLParameter(QUERY_KEY.CONVERSATION_KEY) || null;
+
+  const keyAndCodeExistent = queryConversationKey && queryConversationCode;
```
atomrc: rather that adding some implicit dependencies here, let's rather use the `embedded` flag set to the `Login` form to decide if we should do the redirect or not. This should be cleaner (and the ConversationJoin page does set this flag)

atomrc (changes requested):
Let's consider not reading unrelated url params in the login page, but rather consume some props given by the parent component (in our case `ConversationJoin` component that sets the `embedded` flag)

src/script/auth/page/Login.tsx
V-Gira: nice, sounds much better

src/script/auth/page/Login.tsx
```diff
@@ -120,8 +121,10 @@ const LoginComponent = ({
 
   const isOauth = UrlUtil.hasURLParameter(QUERY_KEY.SCOPE, window.location.hash);
 
+  const features = Config.getConfig().FEATURE;
+
   const [showEntropyForm, setShowEntropyForm] = useState(false);
-  const isEntropyRequired = Config.getConfig().FEATURE.ENABLE_EXTRA_CLIENT_ENTROPY;
+  const isEntropyRequired = features.ENABLE_EXTRA_CLIENT_ENTROPY;
```
przemvs: You can also do this thing:
```
const {
  ENABLE_EXTRA_CLIENT_ENTROPY: isEntrophyRequired,
  ENABLE_DOMAIN_DISCOVERY,
  ENABLE_SSO,
  ...
} = features;
```

I think we will code more clean :)

src/script/auth/page/Login.tsx
```diff
@@ -348,9 +351,7 @@ const LoginComponent = ({
   return (
     <Page>
       {!embedded &&
-        (Config.getConfig().FEATURE.ENABLE_DOMAIN_DISCOVERY ||
-          Config.getConfig().FEATURE.ENABLE_SSO ||
-          Config.getConfig().FEATURE.ENABLE_ACCOUNT_REGISTRATION) && (
+        (features.ENABLE_DOMAIN_DISCOVERY || features.ENABLE_SSO || features.ENABLE_ACCOUNT_REGISTRATION) && (
```
przemvs: This one You can move to variable. You're using it also in lines 381-383 :)

atomrc (approved):

src/script/auth/page/Login.tsx
V-Gira: addressed here  https://github.com/wireapp/wire-webapp/pull/17306/commits/fe6040206f296cec4b304568d8c7df1a162cf43b

src/script/auth/page/Login.tsx
V-Gira: aaand here: https://github.com/wireapp/wire-webapp/pull/17306/commits/fe6040206f296cec4b304568d8c7df1a162cf43b

przemvs (approved):
LGTM! 🚀

PR #17320, opened 2024-04-25
ci: fix helm release version bumps

smatting:
This PR should fix the following bug: tagged commits don't get a proper version bump for their helm chart


smatting:
we don't want to use TAG anymore, it's messy

PR #17323, opened 2024-04-25
runfix: do not allow a new client to snooze the enrollment [WPB-8878]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8878" title="WPB-8878" target="_blank">BugWPB-8878</a>  [Web] Possible to log in on C3 without enrolling for certificate
  </td></table>
  <br />
 

## Description

We should not allow a completely new MLS client to snooze the enrolment retry-process after it has failed for the first time.

If there's no user data after coming back from the redirect, we should try the enrollment flow from scratch.

## Checklist

### DIFF
--- a/src/script/E2EIdentity/E2EIdentityEnrollment.ts
+++ b/src/script/E2EIdentity/E2EIdentityEnrollment.ts
@@ -330,14 +330,11 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
     // Clear the e2e identity progress
     await this.coreE2EIService.clearAllProgress();
 
-    // A fresh MLS client should be forced to enrol.
-    const shouldHideSnooze = (await isFreshMLSSelfClient()) || !snoozable;
-
-    return new Promise<void>(async resolve => {
+    return new Promise<void>(resolve => {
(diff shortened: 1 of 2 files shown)


atomrc (approved):

src/script/E2EIdentity/E2EIdentityEnrollment.ts
```diff
@@ -143,14 +143,30 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
...
   }
 
+  public wasJustRedirected() {
+    const searchParams = new URLSearchParams(window.location.search);
+
+    const hasState = searchParams.has('state');
+    const hasSessionState = searchParams.has('session_state');
+    const hasCode = searchParams.has('code');
+
+    return hasState && hasSessionState && hasCode;
```
atomrc: Here is my take on this:
We should, as much as possible, rely on the library to read url params. so I'd prefer a solution where:

- we are a little less performant (fetching the entire user)
- but we don't have to hard code url params in our code

This way we let the library handle the params and in case in the future they evolve we won't have to change our own logic. 

Does that make sense?

PR #17325, opened 2024-04-25
fix: Add "Unsupported browser" page with no complicated JS WPB-8820

svitovyda:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8820" title="WPB-8820" target="_blank">BugWPB-8820</a>  [web] The unsupported browser page is broken
  </td></table>
  <br />
 

## Description

Move most of the check of supported browsers logic from the app (with all the React etc. scripts loaded and potentially failing the app due to the usage of unsupported features) into a separate small JS script and a static HTML page

- Instead of checking the browsers and their versions, we use an optimistic approach and detect the most important features;
- Except for mobile browsers: they are not supported at all, but the oAuth page is allowed in mobile browsers (see [WPB-1524](https://wearezeta.atlassian.net/browse/WPB-1524));
- The CSS is static and doesn't get affected by the theme (and looks slightly different - the font is a bit bigger than it should be);
- So far the page is only in English, its localisation is the subject of a separate Jira ticket;
- For now, we agreed to add the link to the Wire support page as hardcoded, maybe later its URL will be taken from the config;
- `UnsupportedBrowser` had to stay to check if the app was built with a configuration to use only the Elektron app;

Static `/unsupported/` page:
Screenshot 2024-04-26 at 19 10 35

Former `UnsupportedBrowser` (`HasToUseDesktop`):
Screenshot 2024-04-26 at 19 10 11

## Checklist

- make sure to tell your opinion about the commented code in `src/script/browser/CheckBrowser.ts`

[WPB-1524]: https://wearezeta.atlassian.net/browse/WPB-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/routes/RedirectRoutes.ts
```diff
@@ -43,10 +43,6 @@ export const RedirectRoutes = (config: ServerConfig, clientConfig: ClientConfig)
     }
     const userAgent = req.header('User-Agent');
     const parseResult = BrowserUtil.parseUserAgent(userAgent);
-    if (!parseResult) {
```
svitovyda, line 46: After the check in `checkBrowser.js` this can't happen

src/page/browser.ejs
```diff
@@ -0,0 +1,160 @@
...
+
+    <meta property="og:title" content="<%= OPEN_GRAPH_TITLE %>"/>
+    <meta property="og:description" name="description" content="<%= OPEN_GRAPH_DESCRIPTION %>"/>
+    <meta property="og:type" content="website"/>
+    <meta property="og:url" content="<%= APP_BASE %>"/>
+    <meta property="og:image" content="<%= OPEN_GRAPH_IMAGE_URL %>"/>
+
+    <title><%= BRAND_NAME %></title>
+
+    <style>
```
svitovyda: I don't know how to inject `less` into a simple static HTML and is it worse the effort

src/page/browser.ejs
```diff
@@ -0,0 +1,160 @@
...
+
+<body>
+<div style="box-sizing: border-box;">
+    <div style="display: flex; height: 100%; min-height: 100vh;">
+        <div style="display: flex; flex-direction: column; min-height: 100vh; width: 100%;">
+            <div class="content">
+                <header style="height: 40px; margin-left: 8px; margin-top: 20px;">
+                    <svg viewBox="0 0 78 25" width="71" height="23" aria-hidden="true" style="overflow: visible;">
+                        <g>
+                            <path xmlns="http://www.w3.org/2000/svg"
```
svitovyda: Same with SVG: I don't know if there is an easy way to inject here the icon from our assets instead of hardcode

src/script/browser/CheckBrowser.ts
```diff
@@ -0,0 +1,135 @@
...
+ */
+
+import {QUERY_KEY} from '../auth/route';
+
+const isMobile = (): boolean => 'ontouchstart' in window || !!navigator.maxTouchPoints;
+
+const isOauth = (): boolean => location?.hash?.includes(QUERY_KEY.SCOPE) ?? false;
+
+const supportsCookies = (): boolean => navigator.cookieEnabled;
+/* return new Promise((resolve, reject) => {
```
svitovyda: I'm not sure if we want to have here such an advanced check, with Promise - then we will have to postpone the other code on the page, show the loader etc. But this is the way our browser detection worked

src/script/browser/CheckBrowser.ts
```diff
@@ -0,0 +1,135 @@
...
+         if (Cookies.get(cookieName)) {
+           Cookies.remove(cookieName);
+           return resolve();
+         }
+         return reject(new Error());
+     }
+   });*/
+
+const supportsIndexDB = (): boolean => 'indexedDB' in window;
+/*{
```
svitovyda: Same. I'm not sure if we have to check the IndexDB for Firefox in this complex way

src/page/browser.ejs
aweiss-dev: We want to get rid of less anyways. So we should never add some new code to it, instead move less code into CSSinJS for the main application. 
For this page plain css is the best solution 👍

src/script/auth/component/HasToUseDesktop.tsx
```diff
@@ -0,0 +1,59 @@
...
+import {unsupportedJoinStrings, unsupportedStrings} from '../../strings';
+
+const hasToUseDesktopApplication = () =>
+  !Runtime.isDesktopApp() && Config.getConfig().FEATURE.ENABLE_ENFORCE_DESKTOP_APPLICATION_ONLY;
+
+interface UnsupportedBrowserProps extends React.HTMLProps<HTMLDivElement> {
+  isTemporaryGuest?: boolean;
+}
+
+export const HasToUseDesktop = ({children, isTemporaryGuest}: UnsupportedBrowserProps) => {
```
aweiss-dev: I am not too fond with the naming of this component. It should rather be something like _SupportedBrowserWrapper_ or something in this direction. 

has*, is* - naming scheme indicates that the variable is a boolean.

src/page/browser.ejs
atomrc: For this page, as @aweiss-dev says, inlining the css directly in the page is a great solution.

src/page/browser.ejs
atomrc: same than for css, all good here :)

src/script/browser/CheckBrowser.ts
atomrc: we do need such a check, I'm afraid. 

If you are afraid the async will not work, it's going to be okay since it's going to be safe guarded by the initial `isBrowserSupported` call. 

So you can go


```
if (isBrowserSupported()) {
  checkIndexedDBSupport().then((isSupported) => {if (!isSupported) { redirect() }});
}
```

svitovyda: done

src/page/auth.ejs
```diff
@@ -18,6 +18,9 @@
     <meta property="og:image" content="<%= OPEN_GRAPH_IMAGE_URL %>" />
 
     <title><%= BRAND_NAME %></title>
+
+    <script src="../browser.js?<%= VERSION %>"></script>
```
svitovyda: If the code is converted by `webpack`, due to `sourceMap` it adds to the compiled code some `webpack` script, that uses `globalThis`. So the decision is to have a very primitive script as plane JS, copied by `webpack`, and a more advanced TS browser checker, that can have things like config and constants from the codebase

src/page/browser.js
```diff
@@ -0,0 +1,49 @@
+function isMinimumSupported() {
```
atomrc: Let's add some comments here that explains why this file is `js` only and what it is for exactly

src/page/auth.ejs
atomrc: This would be a valuable comment to leave in the code itself ;)

src/page/browser.js
```diff
@@ -0,0 +1,49 @@
...
+    'endsWith' in String.prototype &&
+    'replaceAll' in String.prototype &&
+    'sticky' in RegExp.prototype &&
+    'toString' in RegExp &&
+    parseFloat('1.23') === 1.23 &&
+    (1.23456789).toFixed(2) === '1.23'
+  );
+}
+
+if (!isMinimumSupported()) {
```
atomrc: wasn't it possible to move the async checks here as well? (cookies and indexeddb)?

src/script/browser/CheckBrowser.ts
```diff
@@ -0,0 +1,102 @@
...
+          clearTimeout(connectionTimeout);
+          resolve(true);
+        };
+      }
+      resolve(true);
+    }
+  }).catch(() => false);
+
+const checkBrowser = (): void => {
+  if (isOauth()) {
```
svitovyda, line 92: we can also add here check if the browser is mobile, but I thought it should be enough to just check if this is the `oAuth` path and allow any browser. Initial feature check happened anyway in the static `browser.js`

src/script/browser/CheckBrowser.ts
```diff
@@ -17,119 +17,86 @@
...
+      default:
+        Cookies.set(cookieName, 'yes');
+        if (Cookies.get(cookieName)) {
+          Cookies.remove(cookieName);
+          resolve(true);
+        } else {
+          resolve(false);
+        }
     }
+  });
```
atomrc: why is this one async, I cannot see any async operation in there 🤔

src/script/browser/CheckBrowser.ts
```diff
@@ -0,0 +1,102 @@
...
+  }).catch(() => false);
+
+const checkBrowser = (): void => {
+  if (isOauth()) {
+    return;
+  }
+  if (!('RTCPeerConnection' in window)) {
+    location.href = '/unsupported/';
+  }
+  supportsIndexDB()
```
svitovyda, line 99: can be simplified with 
```
await const supportsDB = supportsIndexDB();
await const supportsC = supportsCookies();
...
```
to simplify the code readability

webpack.config.common.js
```diff
@@ -150,7 +150,8 @@ module.exports = {
           to: `${dist}/min/core-crypto.wasm`,
         },
         // copying all static resources (audio, images, fonts...)
-        {from: 'resource', to: dist},
+        {from: 'src/page/browser.js', to: dist},
+        {from: 'resource', to: path.resolve(dist, 'browser.js')},
```
atomrc: do we need those 2 lines, I feel only the first one is needed, isn't it? 

Let's move the `browser.js` file to `${dist}/min` so that all the js are there in the end

src/script/browser/CheckBrowser.ts
```diff
@@ -17,119 +17,86 @@
...
-  parseFloat('1.23') === 1.23 &&
-  (1.23456789).toFixed(2) === '1.23' &&
-  'RTCPeerConnection' in window;
+const checkBrowser = (): void => {
+  if (isOauth()) {
+    return;
+  }
+  if (!('RTCPeerConnection' in window)) {
+    location.href = '/unsupported/';
+  }
```
atomrc: why is this one here and not in `browser.js`?

src/page/browser.js
atomrc, line 1: let's maybe call this file `basicBrowserFeatureCheck.js`, so that it's a little more explicit what this file does, WDYT?

src/script/auth/component/HasToUseDesktop.tsx
svitovyda: In the end, the component is not needed at all, as the same check is already happening in `src/script/main/index.tsx`

src/page/auth.ejs
svitovyda: done

src/page/browser.js
svitovyda: done

src/page/browser.js
svitovyda: I would like to have more advanced code as a TS, then the types are clear, and I think some better syntax sugar (I'm not sure what pure JS offers, babel injects cool syntax features support on the fly). Here we just primitively check base functions availability and I like to keep it at a minimum. This will never be the code you'd like to debug. But if you see the value of having it here, I can try to move it.

src/script/browser/CheckBrowser.ts
svitovyda: I was just thinking of leaving in `browser.js` only the most primitive things. At first, I even thought to put there only `globalThis`, `Promise`, `Object`, `Array`. My logic was like... what if we identify similar issues with `RTC` support, as with `IndexDB`? We are forced to have 2 separate files anyway, so I was dividing the logic somehow intuitively - basic primitive things into the static JS files, some more advanced code - into the TS file. Let me know if you think it's better to move it there

atomrc: Ok, it makes sense I think :)

Maybe, since this one is a proper ts file, maybe we can 'import' it from the source files (rather than from html) WDYT?

src/page/browser.ejs
atomrc, line 1: This files should probably be named 'unsupported.ejs'. Right?

webpack.config.common.js
svitovyda: done

src/script/browser/CheckBrowser.ts
svitovyda: Not applicable anymore, as `supportsCookies()` is not async

src/page/browser.js
svitovyda: ah, I forgot: in the TS `CheckBrowser` I allow to run the browser for the oAuth case. I don't want to move oAuth to here, as it uses the constant `QUERY_KEY.SCOPE` - hardcoding it here would mean we must remember to fix it here too if some change happens in the main code. And I guess we don't need `IndexDB` or cookies for the oAuth  part, while things like `globalThis` and some other basics will be still required at least due to `sourceMap` injection by `webpack`

src/script/browser/CheckBrowser.ts
svitovyda: I'm not sure the browser won't get blocked by some unsupported code inside, for instance, `vendor.js`, if we import it in the main code. It might work, as the base features are already detected within `browser.js`, so the chances are already pretty low. But we wanted to completely move it out of the main codebase initially for this reason. Right now the check browser scripts run instantly in the `<head>` of the HTMLs. Also, they are very readable this way - you see them both imported one by one in the HTML. WDYT?

svitovyda: and regarding the `RTC` check - I forgot that the main reason is to not block the oAuth scenario. I would be happy if we would move the oAuth user interaction path from the main code too.

atomrc: yeah, it makes sense to do the rtc check later in the app. (and only for `app.ts`, not the auth page). 

But since those are application specific checks, they do belong in the app code I believe. It would make sense to import this file from `src/script/main/index.tsx` I believe

src/script/browser/CheckBrowser.ts
svitovyda: right! I just copied it from the app and didn't think to double-check!
done

src/script/browser/CheckBrowser.ts
svitovyda: I'm not sure about not adding it to the auth page: if a user logs in and then immediately gets the page "your browser is not supported" - then what is the point of allowing the user to log in at all?

atomrc (approved):

PR #17326, opened 2024-04-26
feat(debug): Allow blurring video background during calls [WPB-724]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-724" title="WPB-724" target="_blank">StoryWPB-724</a>  [Web] Blurred background on calls
  </td></table>
  <br />
 

## Description

This is a first MVP for the blurred background. 
It enabled blurring the self background using a debug util `wire.debug.enableCameraBlur()`. 

Next iteration:
- allow pre-loading the segmenter's models ahead of time so that blurring feels instant
- experiment a little more with webGL and see if we can optimize general blurring performances (see https://github.com/wireapp/wire-webapp/pull/17512/files)

## Performances

Currently the performances are not quite there yet. The blurring is quite costly. 
This is a typical example of the computation of a single frame of the video feed

image

We can see that the segmentation is taking very little time (some 2-3ms) but blurring takes 20ms+. Since we are supposed to have a budget of 16ms, this is not really our target performance. 
It's also quite intense on the main thread and leaves very little resources for the user. Those issues will need to be tackled.

### Using webWorker to compute the blur:

- 👍  leaves the main thread rather free
- 👎  frame rate drops and video gets quite laggy

### Using webgl to compute blurring on the GPU

- no performance gain have been found yet (https://github.com/wireapp/wire-webapp/pull/17512/files#diff-6ecbcd2983f79d5985879c4aa68724d2f00b4297aab3868a654217d6f3757768) Will need more experimenting


package.json
```diff
@@ -49,10 +50,12 @@
     "redux-logdown": "1.0.4",
     "redux-thunk": "3.1.0",
     "speakingurl": "14.0.1",
+    "stackblur-canvas": "^2.7.0",
```
aweiss-dev: this can go

assets/mediapipe-models/selfie_multiclass_256x256.tflite
aweiss-dev, line 1: this can go

.husky/post-checkout
aweiss-dev, line 1: no need for LFS without the large file

.husky/post-commit
aweiss-dev, line 1: no need for LFS without the large file

.husky/post-merge
aweiss-dev, line 1: no need for LFS without the large file

.husky/pre-push
aweiss-dev, line 1: no need for LFS without the large file

.gitattributes
```diff
@@ -13,3 +11,4 @@
 *.ttf binary
 *.woff binary
 *.woff2 binary
+*.tflite filter=lfs diff=lfs merge=lfs -text
```
aweiss-dev: no need for LFS without the large file

PatrykBuniX (approved):

PR #17344, opened 2024-04-30
runfix: Correctly check if user has mls device before performing mls operations

atomrc:
## Description

see https://github.com/wireapp/wire-web-packages/pull/6185/files

## Checklist

### DIFF
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -57,7 +57,7 @@ export class MLSConversationVerificationStateHandler {
     private readonly core: Core,
   ) {
     this.logger = getLogger('MLSConversationVerificationStateHandler');
-    // We need to check if the core service is available and if the e2eIdentity is available
+    // We need to check if the core account has a valid MLS device and that e2ei is enabled
     if (!this.core.hasMLSDevice || !this.core.service?.e2eIdentity) {
       return;
     }
--- a/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
+++ b/src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
@@ -58,12 +58,12 @@ export class MLSConversationVerificationStateHandler {
   ) {
     this.logger = getLogger('MLSConversationVerificationStateHandler');
     // We need to check if the core service is available and if the e2eIdentity is available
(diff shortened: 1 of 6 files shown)


src/script/main/app.ts
```diff
@@ -443,7 +443,7 @@ export class App {
       // We load all the users the self user is connected with
       await userRepository.loadUsers(selfUser, connections, conversations, teamMembers);
 
-      if (supportsMLS()) {
+      if (this.core.hasMLSDevice) {
```
atomrc, line 446: if we have a MLS device it means that:
- the env supportsMLS (previous check)
- the team has MLS enabled
- the user has a valid mls device

(it a more exhaustive check than the previous `supportsMLS` check)

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
```diff
@@ -58,12 +58,12 @@ export class MLSConversationVerificationStateHandler {
   ) {
     this.logger = getLogger('MLSConversationVerificationStateHandler');
     // We need to check if the core service is available and if the e2eIdentity is available
-    if (!this.core.service?.mls || !this.core.service?.e2eIdentity) {
+    if (!this.core.hasMLSDevice || !this.core.service?.e2eIdentity) {
```
V-Gira: is this correct? If so, it doesn't match the code comment anymore right?

src/script/main/app.ts
```diff
@@ -472,7 +472,7 @@ export class App {
 
       await conversationRepository.init1To1Conversations(connections, conversations);
 
-      if (supportsMLS()) {
+      if (this.core.hasMLSDevice) {
```
V-Gira, line 475: this one is the fix to automation correct?
Seems like the change in `MLSConversationVerificationStateHandler` isn't necessarry? (that's the one breaking unit tests)

src/script/conversation/ConversationVerificationStateHandler/MLS/MLSStateHandler.ts
atomrc: indeed the comment needs updating, good catch :) 
The fix is needed though (we should never check if mls is enabled by checking that the service is there, anymore)

V-Gira: gotcha, I guess we're just not spying on that `hasMLSDevice` in the unit test then

src/script/main/app.ts
atomrc: exactly, I'm working on the fix right now :)

V-Gira (approved):

PR #17350, opened 2024-05-03
runfix: check if IndexDB enabled in FireFox and extend the message in "Unsupported browser" page WPB-8820

svitovyda:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8820" title="WPB-8820" target="_blank">BugWPB-8820</a>  [web] The unsupported browser page is broken
  </td></table>
  <br />
 

## Description
- Fix check of IndexDB in FireFox private mode (was broken);
- Add console logs to identify which part of browser feature support triggered;
~- Message in "Unsupported browser" includes settings for cookies and IndexDB;~
- Messages in "Unsupported browser" page is up to date with our product/design/copy team and are final;

## Checklist


src/page/basicBrowserFeatureCheck.js
```diff
@@ -76,5 +76,7 @@ function isMinimumSupported() {
 }
 
 if (!isMinimumSupported()) {
+  // eslint-disable-next-line no-console
+  console.error('The minimum browser features are not supported to run the Wire app!');
```
V-Gira, line 80: I don't have a lot of insight on that unsupported browser subject.
But is this a log we intend to keep in production or is it just for testing purposes?

svitovyda: I would like it on production, not just for testing purposes. Imagine we get some bug about an unsupported browser, and if a user can provide us with this log we will be able to immediately know where to look for the issues. In any case, this log will be almost the only one in the console, as the app won't run, so it should not spam it

src/script/browser/CheckBrowser.ts
```diff
@@ -58,49 +58,60 @@ const supportsCookies = (): boolean => {
...
+    };
+
+    dbOpenRequest.onsuccess = _event => {
+      clearTimeout(connectionTimeout);
+      return resolve(true);
+    };
   });
 
 const checkBrowser = (): void => {
+  if (!supportsCookies()) {
```
svitovyda, line 94: Cookies are required for oAuth logic, but we cannot move it into the `basicBrowserFeatureCheck.js` as it has a dependency on `js-cookie`

src/script/browser/CheckBrowser.ts
```diff
@@ -58,49 +58,60 @@ const supportsCookies = (): boolean => {
 const supportsIndexDB = (): Promise<boolean> =>
   new Promise<boolean>((resolve, _reject) => {
     if (!('indexedDB' in window)) {
-      resolve(false);
-    } else {
-      // some versions of FF don't allow access to IndexDB in the private mode
-      if (navigator.userAgent.toLowerCase().indexOf('firefox') !== -1) {
-        let dbOpenRequest: IDBOpenDBRequest;
+      return resolve(false);
```
PatrykBuniX, line 61: **question**: Did we change any logic in `supportsIndexDB` or it's just the refactoring?

svitovyda: I did change the logic there, as it was returning `true` in one of the branches when it had to be `false`, and then I also refactored the code to be simpler

src/page/basicBrowserFeatureCheck.js
svitovyda: Also, since we removed anything technical from the text shown to the user, someone technical will still see that there is a problem with cookies or with IndexDB. There is a very low probability it is needed, but it is still nice to have

tlebon: id also lean towards another solution as console logs arent really something we tend to do in prod.

src/page/unsupported.ejs
```diff
@@ -130,24 +134,25 @@
                 </header>
                 <main style="flex: 1 1 0; padding-left: 8px; width: 100%;">
                     <div class="textContent">
-                        <h2 data-uie-name="element-unsupported-headline">This browser is not supported.</h2>
+                        <h2 data-uie-name="element-unsupported-headline">Wire doesn't support this browser.</h2>
```
tlebon: these should all be new translation strings

tlebon (changes requested):
add the translations please and lets see if we can come up with a better solution than console log

src/page/basicBrowserFeatureCheck.js
svitovyda: Any ideas how else we can implement this? Or you prefer to remove it completely? I'm against console logs for the application too, but this is even the code that is not a part of the main app codebase. When we are showing just one screen "Your browser is not supported" without anything technical about cookies or IndexDB (private mode), I thought it was ok to throw an error into the console.

src/page/unsupported.ejs
svitovyda: There is a separate ticket for localization of the strings in the "Unsupported browser" page, we agreed with Tom and Sascha to implement it hardcoded EN for now and think about configuring the link URL and translation later.
Since this one should be the simplest plain HTML that can be shown in any oldest browser possible and it has no scripts at all, it might be really difficult to implement the usage of `intl` here.

tlebon (approved):

src/page/unsupported.ejs
svitovyda: https://wearezeta.atlassian.net/browse/WPB-9005 - the translation ticket

src/script/browser/CheckBrowser.ts
```diff
@@ -58,49 +58,60 @@ const supportsCookies = (): boolean => {
...
 
 const checkBrowser = (): void => {
+  if (!supportsCookies()) {
+    location.href = '/unsupported/';
+    console.error("This browser doesn't support cookies to run the Wire app!");
+    return;
+  }
   if (isOauth()) {
+    // eslint-disable-next-line no-console
+    console.log('This browser is valid to handle oAuth scenario.');
```
V-Gira: Errors I can see why, not sure we need that log though

src/script/browser/CheckBrowser.ts
```diff
@@ -58,49 +58,60 @@ const supportsCookies = (): boolean => {
...
     return;
   }
-  if (!('RTCPeerConnection' in window) || !supportsCookies()) {
+  if (!('RTCPeerConnection' in window)) {
     location.href = '/unsupported/';
+    console.error("This browser doesn't support RTCto run the Wire app!");
```
V-Gira:
```suggestion
    console.error("This browser doesn't support RTC to run the Wire app!");
```

lil space missing

src/script/browser/CheckBrowser.ts
svitovyda: removed it

V-Gira (approved):

src/page/basicBrowserFeatureCheck.js
svitovyda: the decision is: remove the `console.log` but leave all the the `console.error`

PR #17369, opened 2024-05-07
refactor: readonly conversation indication

PatrykBuniX:
## Description

Applies changes already merged on dev branch - refactoring of readonly mls conversation indication + improved react localizer util. This is needed for handling new case of marking a conversation as readonly - migrating a conversation with a user that don't have any keys available.

## Checklist


src/i18n/en-US.json
```diff
@@ -1166,7 +1166,7 @@
   "ongoingGroupAudioCall": "Ongoing conference call with {{conversationName}}.",
   "ongoingGroupVideoCall": "Ongoing video conference call with {{conversationName}}, your camera is {{cameraStatus}}.",
   "ongoingVideoCall": "Ongoing video call with {{conversationName}}, your camera is {{cameraStatus}}.",
-  "otherUserNotSupportMLSMsg": "You can't communicate with [bold]{{participantName}}[/bold] anymore, as you two now use different protocols. When [bold]{{participantName}}[/bold] gets an update, you can call and send messages and files again.",
+  "otherUserNotSupportMLSMsg": "You can't communicate with {{participantName}} anymore, as you two now use different protocols. When {{participantName}} gets an update, you can call and send messages and files again.",
```
V-Gira, line 1169: Question: Is localisation in German done for those? Might be worth it to cherrypick the localisation commit too if we want to avoid the `[bold]` tag to appear as a string

PatrykBuniX: When merged to dev, these changes were made automatically for German

PatrykBuniX: So I guess it will happen automatically here as well

V-Gira: Let's see, it didn't work out that way on master last time :| 
Let's worry about it when we get there :crossed_fingers:

V-Gira (approved):

PR #17371, opened 2024-05-07
runfix: 1:1 conversation with user without key packages

PatrykBuniX:
## Description

Handles a 1:1 conversation with a user that don't have any key packages available to establish a MLS 1:1 conversation.

## Checklist


src/script/conversation/ConversationRepository.ts
```diff
@@ -1503,14 +1510,6 @@ export class ConversationRepository {
     }
   };
 
-  private readonly updateConversationReadOnlyState = async (
-    conversationEntity: Conversation,
-    conversationReadOnlyState: CONVERSATION_READONLY_STATE | null,
-  ) => {
-    conversationEntity.readOnlyState(conversationReadOnlyState);
-    await this.saveConversationStateInDb(conversationEntity);
-  };
```
PatrykBuniX, line 1512: We don't need to save the state manually, there's a subscription that updates the db whenever the value changes. See https://github.com/wireapp/wire-webapp/blob/runfix/1to1-migration-no-keys/src/script/entity/Conversation.ts#L582

przemvs (approved):

PR #17386, opened 2024-05-10
feat: Jump to last message WPB-6518

svitovyda:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6518" title="WPB-6518" target="_blank">StoryWPB-6518</a>  [Web] Jump to last message in a conversation
  </td></table>
  <br />
 

## Description

By the product/design team request, the user should jump to the very last message in the conversation, marking all the messages as read

https://github.com/wireapp/wire-webapp/assets/24770159/03c63fba-468f-4f39-ac81-0f0836f8c619

https://github.com/wireapp/wire-webapp/assets/24770159/38e094ca-8af9-4c10-a661-971d0bfd9ef5

https://github.com/wireapp/wire-webapp/assets/24770159/be9d6045-6103-4c6f-a3d5-7f2333121007

## Checklist


src/script/components/LastMessageTracker/LastMessageTracker.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import {Conversation as ConversationEntity} from '../../entity/Conversation';
+import {Message} from '../../entity/message/Message';
+
+export interface MessageVisibility {
+  message: Message;
+  isVisible: boolean;
+}
+
+export interface LastMessageTrackerProps extends HTMLProps<HTMLElement> {
```
svitovyda: This should be a separate component as we want to avoid the entire content of `Conversation` to re-render each time messages are scrolled and messages visibility changes

src/script/components/MessagesList/Message/index.tsx
```diff
@@ -88,6 +89,7 @@ export const Message: React.FC<MessageParams & {scrollTo?: ScrollToElement}> = p
     isHighlighted,
     hideHeader,
     onVisible,
+    onVisibilityChange,
```
svitovyda: I know this is weird, but as `isVisible` was implemented till now, it would be triggered only the first time (and if it's added as the item of the list one by one, and then disappear from the visible area - it will still be called). I don't mind merging them at least on the `Message` and the level of its usage, but I want to hear your opinions on how to do it the best

src/script/components/Conversation/Conversation.tsx
```diff
@@ -460,6 +464,19 @@ export const Conversation = ({
     [addReadReceiptToBatch, repositories.conversation, repositories.integration, updateConversationLastRead],
   );
 
+  const onGoToLastMessage = () => {
+    if (activeConversation) {
+      activeConversation.initialMessage(undefined);
```
svitovyda: clean up anything like search result

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -168,7 +166,7 @@ export const MessagesList: FC<MessagesListParams> = ({
     scrollHeight.current = newScrollHeight;
   }, [messagesContainer?.parentElement, loaded, filteredMessages, selfUser?.id]);
 
-  // Listen to resizes of the the content element (if it's resized it means something has changed in the message list, link a link preview was generated)
+  // Listen to resizes of the content element (if it's resized it means something has changed in the message list, link a link preview was generated)
```
svitovyda, line 171: just fix a typo

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -126,7 +124,7 @@ export const MessagesList: FC<MessagesListParams> = ({
 
   const groupedMessages = groupMessagesBySenderAndTime(filteredMessages, conversationLastReadTimestamp.current);
 
-  const [messagesContainer, setMessageContainer] = useState<HTMLDivElement | null>(null);
+  const [messagesContainer, setMessagesContainer] = useState<HTMLDivElement | null>(null);
```
svitovyda, line 129: just fix a typo

setupTests.js
```diff
@@ -68,5 +68,12 @@ window.z = {userPermission: {}};
 window.URL.createObjectURL = jest.fn();
 window.URL.revokeObjectURL = jest.fn();
 
+Object.defineProperty(document, 'elementFromPoint', {
```
svitovyda, line 71: I fixed it globally for tests like `MessageList`, where we don't want to test this functionality, but it gets called now. It is still possible to mock it like `document.elementFromPoint = () => element;`

src/script/components/Conversation/Conversation.tsx
```diff
@@ -460,6 +464,19 @@ export const Conversation = ({
...
+      if (!activeConversation.hasLastReceivedMessageLoaded()) {
```
svitovyda: there are two scenarios: all messages are loaded and we only need to scroll down - then we call the `rerenderMessageList()` and MessageList gets re-rendered (unread messages will be automatically marked as read). Or we have unloaded messages in the conversation and need to reload it entirely and mark all messages as read to enforce jumping to the very last message instead of the first unread one

src/script/view_model/ContentViewModel.ts
```diff
@@ -239,7 +239,7 @@ export class ContentViewModel {
       exposeMessage: exposeMessageEntity,
       openFirstSelfMention = false,
       openNotificationSettings = false,
-    } = options;
+    } = options || {};
```
aweiss-dev, line 242: Options are not supposed to be null | undefined, what is this change for?

src/script/components/Conversation/Conversation.tsx
```diff
@@ -118,8 +121,11 @@ export const Conversation = ({
 
   const [isMsgElementsFocusable, setMsgElementsFocusable] = useState(true);
 
+  const [messagesListRerenderKey, rerenderMessageList] = useComponentRerenderKey('messages-list');
```
atomrc: This seems like a non-trivial hook. Can you add comment to explain why this is needed and what's the purpose of it?

src/script/view_model/ContentViewModel.ts
svitovyda: Our TS config allowed me to use this function without providing this parameter and I got an exception. And actually, in many cases we provide just `{}`, so I'd rather make `options` optional and remove those `{}`, WDYT?

src/script/components/Conversation/Conversation.tsx
```diff
@@ -460,6 +464,19 @@ export const Conversation = ({
     [addReadReceiptToBatch, repositories.conversation, repositories.integration, updateConversationLastRead],
   );
 
+  const onGoToLastMessage = () => {
```
atomrc: This is a concrete implementation of a handler. The name should not start with `on` it should have a imperative naming (explaining what it does exactly). 

I'd suggest `loadLastReceivedMessages` or something along those lines

src/script/components/Conversation/Conversation.tsx
atomrc: This would be a valuable comment to add to the code directly

src/script/components/Conversation/Conversation.tsx
atomrc: This is also valuable code comment :)

src/script/components/Conversation/Conversation.tsx
```diff
@@ -566,3 +596,26 @@ export const Conversation = ({
...
+  if (isLastMessageVisible) {
+    return null;
+  }
+
+  return (
+    <IconButton data-uie-name="jump-to-last-message-button" onClick={onGoToLastMessage}>
+      <ChevronIcon css={{rotate: '90deg', height: 16, width: 16, path: {fill: '#0667C8'}}} />
+    </IconButton>
+  );
+};
```
atomrc: Let's extract that to its own file

src/script/components/Conversation/Conversation.tsx
svitovyda: This is the way to trigger the re-render of `MessageList`, by changing its `key` (KUDOs to @PatrykBuniX btw). Should I add a comment to the code?

src/script/view_model/ContentViewModel.ts
aweiss-dev: That would be the preferred way to act on this issue. If it can be undefined, it needs to be typed as such 👍

src/script/components/Conversation/Conversation.tsx
svitovyda: Initially I did so, but then I saw we have such small components in the same file, will do so

src/script/entity/Conversation.ts
```diff
@@ -163,6 +167,7 @@ export class Conversation {
   public readonly receiptMode: ko.Observable<RECEIPT_MODE>;
   public readonly removed_from_conversation: ko.PureComputed<boolean>;
   public readonly roles: ko.Observable<Record<string, string>>;
+  public readonly isLastMessageVisible: ko.Observable<boolean>;
```
atomrc, line 166: This should not belong in the low level entity. Should be a view-only thing. 
It should leave in `Conversation.tsx` only, I believe

src/script/components/Conversation/Conversation.tsx
atomrc: That's clever solution actually. I like it!!

But, because it's clever, it does call for a code-comment :)

src/script/components/Conversation/Conversation.tsx
```diff
@@ -460,6 +465,19 @@ export const Conversation = ({
...
 
+  const onGoToLastMessage = () => {
+    if (activeConversation) {
+      activeConversation.initialMessage(undefined);
+      if (!activeConversation.hasLastReceivedMessageLoaded()) {
+        updateConversationLastRead(activeConversation);
+        activeConversation.release();
+        amplify.publish(WebAppEvents.CONVERSATION.SHOW, activeConversation, {});
+      } else {
+        rerenderMessageList();
```
atomrc: I think in this case we should just run a scroll animation and not reload the entire conversation. This would feel more natural

src/script/components/Conversation/Conversation.tsx
svitovyda: done

src/script/entity/Conversation.ts
svitovyda: It was my idea too, but then we should pass it to the `MessageList` to implement its updating, or pull all the updating logic into the `Conversation`, chaining props of `onVisible` and `onVisibilityLost`. It was quite ugly implementation. I also thought that some of the props and methods of `ConversationEntity` might not be in the perfect place (like `initialMessage`), and maybe we need a `ConversationView` service, WDYT?

atomrc: I believe we do need a cleaner architecture for those values, yes. But maybe it's not the correct PR to tackle this. Probably a follow up PR would be nice.

src/script/view_model/ContentViewModel.ts
svitovyda: done

src/script/components/Conversation/Conversation.tsx
svitovyda: done, I called it `jumpToLastMessage`

src/script/components/Conversation/Conversation.tsx
svitovyda: done

src/script/components/Conversation/Conversation.tsx
svitovyda: done

src/script/components/Conversation/JumpToLastMessageButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {FC, HTMLProps} from 'react';
+
+// eslint-disable-next-line import/no-unresolved
```
svitovyda: without this type, there will be test failure due to:
```
   Error: src/script/components/Conversation/JumpToLastMessageButton.tsx(45,6): error TS2322: Type '{ children: Element; accept?: string; acceptCharset?: string; action?: string; allowFullScreen?: boolean; allowTransparency?: boolean; alt?: string; as?: string; async?: boolean; autoComplete?: string; ... 358 more ...; css: { ...; }; }' is not assignable to type 'IconButtonProps<HTMLButtonElement>'.
  Types of property 'ref' are incompatible.
    Type 'LegacyRef<HTMLElement>' is not assignable to type 'Ref<HTMLButtonElement>'.
      Type 'string' is not assignable to type 'Ref<HTMLButtonElement>'.
```

to fix this ESLint issue, StackOverflow recommends to add 
```
extends: [
    "plugin:import/errors",
    "plugin:import/warnings",
    "plugin:import/typescript",
  ],
```
but it doesn't help (and we already have some of them in the `@wireapp/eslint-config`

Without using `{...rest}` the CSS will not get applied on `IconButton`, and setting the position CSS inside of the `JumpToLastMessageButton` looks as a bad solution, it should be positioned where it is used

src/script/components/Conversation/Conversation.tsx
svitovyda: done

src/script/components/Conversation/JumpToLastMessageButton.tsx
tlebon: seems like a typing issue

tlebon: i would prefer the css inside of the component and just pass the variable in and decide the size based on var

svitovyda: yes, it is a type issue, and using `WithConditionalCSSProp` solves it. but this import causes the `ESLint` error

this is the CSS positioning we have to set:
```
  position: 'absolute',
  bottom: mdBreakpoint? '100px' : '56px',
  right: '10px',
  height: '40px',
```
Do you mean you'd prefer to have a separate custom `CSSObject` prop in `JumpToLastMessageButtonProps` instead of `css`  and apply it on `IconButton`?

src/script/components/Conversation/Conversation.tsx
```diff
@@ -526,11 +549,22 @@ export const Conversation = ({
...
+
+          <JumpToLastMessageButton
+            onGoToLastMessage={jumpToLastMessage}
+            conversation={activeConversation}
+            css={{
+              position: 'absolute',
+              bottom: mdBreakpoint ? '100px' : '56px',
+              right: '10px',
+              height: '40px',
+            }}
```
przemvs: Could You move this styles to conversation.style.ts file? :)

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +255,167 @@ export const MessagesList: FC<MessagesListParams> = ({
     syncScrollPosition();
   };
 
+  const jumpToLastMessage = () => {
+    if (conversation) {
+      // clean up anything like search result
+      setHighlightedMessage(undefined);
```
svitovyda, line 256: It feels that this cleanup should happen somewhere in the `useEffect`, when props and observables change, but I'm afraid to break something this way

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +255,167 @@ export const MessagesList: FC<MessagesListParams> = ({
...
+            });
+          })}
+        </div>
+      </FadingScrollbar>
+      <JumpToLastMessageButton onGoToLastMessage={jumpToLastMessage} conversation={conversation} />
+    </>
+  );
+};
+
+export interface JumpToLastMessageButtonProps extends HTMLProps<HTMLElement> {
```
svitovyda: we agreed to keep this component inside of `MessageList.tsx` as it contains positioning CSS and we didn't want to overcomplicate it by passing the custom CSS. It still should be a separate component with it's own state, as otherwise `isLastMessageVisible` will make the entire `MessageList` to re-render, unless we will implement memoisation of each element rendered.

svitovyda: `export` is for the test

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +255,167 @@ export const MessagesList: FC<MessagesListParams> = ({
...
+  onGoToLastMessage,
+  conversation,
+}: JumpToLastMessageButtonProps) => {
+  // To be changed when design chooses a breakpoint, the conditional can be integrated to the ui-kit directly
+  const mdBreakpoint = useMatchMedia('max-width: 768px');
+
+  const [isLastMessageVisible, setIsLastMessageVisible] = useState(conversation.isLastMessageVisible());
+
+  useEffect(() => {
+    const subscription = conversation.isLastMessageVisible.subscribe(
```
svitovyda: I can't use here `useKoSubscribableChildren` as I need to add debouncing to the change of the state.

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +255,167 @@ export const MessagesList: FC<MessagesListParams> = ({
...
+      debounce(value => {
```
svitovyda: debouncing is needed for messages, sent by `self`: at first the message from self gets mounted as not yet delivered, then it gets unmounted and mounted as received from the server. this makes the jump to the last message button blink. there is a complicated way to check if a message that is unmounted (its `onVisibilityLost` was called) has not yet been delivered, but we should somehow distinguish whether it was unmounted and not hidden by scrolling. I think debouncing is better, while it does add a bit of noticeable delay even with 200 ms

src/script/components/utils/InViewport.tsx
```diff
@@ -25,6 +25,7 @@ import {viewportObserver} from 'Util/DOM/viewportObserver';
 interface InViewportParams {
   onVisible: () => void;
```
svitovyda, line 26: in the current implementation, this callback will be called when most of the element is visible. and we need it this way to mark the message as read. if I reuse this callback, the button "jump to the last message" will be disappearing slightly different from how it behaves on our android app - not when a bit of the last message appears, but when the big chunk of it is visible. @atomrc suggested we leave it as it is. the other option would be to implement one callback `onPartialVisibilityChange(visible: boolean)` instead of `onVisibilityLost()` and call it with `true` differently than `onVisible` (as `onVisible` is used in many places, not only for messages)

src/script/components/Conversation/JumpToLastMessageButton.tsx
svitovyda: we decided to leave this component inside of the `MessageList` file, as it will only be used there, so that we can hardcode all the styles inside of it without any `...rest` or other props way logic

src/script/components/Conversation/Conversation.tsx
svitovyda: done

src/script/components/MessagesList/MessageList.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CSSObject} from '@emotion/react';
+
+export const jumpToLastMessageButtonStyles: (mdBreakpoint: boolean) => CSSObject = (mdBreakpoint: boolean) => ({
+  position: 'absolute',
+  bottom: mdBreakpoint ? '100px' : '56px',
```
przemvs: You can do smth like that:

```suggestion
export const jumpToLastMessageButtonStyles: (mdBreakpoint: boolean) => CSSObject = (mdBreakpoint: boolean) => ({
  position: 'absolute',
  bottom: '56px',
  '@media (max-width: @screen-md-min)': {
     bottom: '100px',
  }
```

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +252,158 @@ export const MessagesList: FC<MessagesListParams> = ({
...
+    </>
+  );
+};
+
+export interface JumpToLastMessageButtonProps extends HTMLProps<HTMLElement> {
+  onGoToLastMessage: () => void;
+  conversation: Conversation;
+}
+
+export const JumpToLastMessageButton: FC<JumpToLastMessageButtonProps> = ({
```
przemvs: You don't need to pass FC<..> here, You have already typed on line 379 :)

src/script/components/MessagesList/MessageList.tsx
przemvs: I think You can move it outside MessageList, and use - memo(Component) with validating props :)

src/script/components/MessagesList/MessageList.tsx
przemvs: You can do debounce with useCallback component, then You can use useKoSubscribableChildren :)

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -245,78 +252,158 @@ export const MessagesList: FC<MessagesListParams> = ({
...
+    return null;
+  }
+
+  return (
+    <IconButton
+      data-uie-name="jump-to-last-message-button"
+      onClick={onGoToLastMessage}
+      css={jumpToLastMessageButtonStyles(mdBreakpoint)}
+    >
+      <ChevronIcon css={{rotate: '90deg', height: 16, width: 16, path: {fill: '#0667C8'}}} />
```
przemvs: Please use variables for fill :)

src/script/entity/Conversation.ts
```diff
@@ -82,6 +82,10 @@ enum TIMESTAMP_TYPE {
   MUTED = 'mutedTimestamp',
 }
 
+export const isLastReceivedMessage = (messageEntity: Message, conversationEntity: Conversation): boolean => {
+  return messageEntity.timestamp() >= conversationEntity.last_event_timestamp();
+};
```
przemvs: Why not to use hasLastReceivedMessageLoaded, You have declared it in this Class ;)

svitovyda: because in this case, I need to identify if the given (rendered) message is the last message, to add to it `onVisible` and `onVisibilityLost`. but in many cases it can be replaced with `hasLastReceivedMessageLoaded` - I'm planning to do the refactoring of it later, within the bug that I discovered. here I just wanted to move it out of React component, as a first step.

przemvs: So please move this from Class :) We have some utils/helpers for cases like this one :)

src/script/components/MessagesList/MessageList.tsx
svitovyda: where should I add it? this is a completely new color AFAIK, we don't have it in our ui-kit

V-Gira: Let me check, a new color is always a red flag

V-Gira:
```suggestion
      <ChevronIcon css={{rotate: '90deg', height: 16, width: 16, path: {fill: 'var(--accent-color)'}}} />
```

src/script/components/MessagesList/MessageList.tsx
svitovyda: there were quite some issues with passing the `css`, we moved it back and forth a few times already. I'd leave it as is for now

src/script/components/MessagesList/MessageList.tsx
svitovyda: done

src/script/components/MessagesList/MessageList.tsx
svitovyda: done

src/script/components/MessagesList/MessageList.styles.ts
svitovyda: Looks like media queries don't accept less variables, so I just hardcoded 768px

src/script/components/MessagesList/MessageList.tsx
przemvs: You can do this smth like that:

```
  const {isLastMessageVisible} = useKoSubscribableChildren(conversation, ['isLastMessageVisible']);

  // eslint-disable-next-line react-hooks/exhaustive-deps
  const debounced = useCallback(
    debounce((value: boolean) => value, DEBOUNCE_TIME),
    [],
  );

  if (!debounced(isLastMessageVisible)) {
    return null;
  }

  return ...
```

src/script/components/Conversation/JumpToLastMessageButton.tsx
przemvs: I think we don't need to pass css from parent component, so I think better to move this component outside MessageList.tsx :)

svitovyda: done

src/script/components/MessagesList/MessageList.tsx
svitovyda: done

src/script/entity/Conversation.ts
svitovyda: done. but I think I will soon move it back as a method :)

src/script/components/MessagesList/MessageList.tsx
svitovyda: I tried it, but there is also one more thing - the initial value of `isLastMessageVisible` should be set by the last fired value. Just waiting for the new event and debouncing makes it behave really weird. I suggest to keep my existing implementation

src/script/components/MessagesList/MessageList.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+export const jumpToLastMessageButtonStyles: CSSObject = {
+  position: 'absolute',
+  right: '10px',
+  height: '40px',
+  borderRadius: '100%',
+  bottom: '56px',
+
+  '@media (max-width: 768px)': {
+    bottom: '100px',
+  },
```
atomrc, line 31: @V-Gira I see this value being copied a few times in the app. Not a topic for now, but we should have a single place where the different breakpoints are defined so we avoid hardcoding breakpoints like this (and when possible those breakpoints should not be used in favor of just relying on the natural DOM flow).



I did try to use the DOM flow to have the button placing naturally, but turns out it's not really possible to have a button that is fixed in a scrolling container sadly. 



Let's try to optimize this in a followup PR

atomrc (approved):

src/script/components/MessagesList/MessageList.tsx
przemvs: Checked, and for now we keep Your code ;)

src/script/components/MessagesList/MessageList.styles.ts
svitovyda: I also see that we have lots of hardcoded breakpoints in the code, and they are different. We should define key breakpoints, and then each component should use them and introduce new ones as `theme.breakpoints.tablet - COMPONENT_CUSTOM_WIDTH_CONSTANT` etc.

PR #17378, opened 2024-05-08
fix: open a 1:1 with a bot from conversation details [WPB-5887]

PatrykBuniX:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-5887" title="WPB-5887" target="_blank">BugWPB-5887</a>  Opening conversation with bots from member list doesn't work
  </td></table>
  <br />
 

## Description

It was not possible to open a 1:1 conversation with a bot (service) from the conversation panel. The reason was we were overwriting service's id with user's id (a user that was added to a conversation as a service). 

The fix is to use original service id and to keep the ability to remove service's user, just try to find a conversation member with the matching `serviceId` field.

Before
before

After
after

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

PR #17405, opened 2024-05-14
runfix: do not overwrite service id when opening the service details

PatrykBuniX:
## Description

After fixing a different bug (see https://github.com/wireapp/wire-webapp/pull/17378/files#diff-42332a315c252fd2b9da98fb360d34d231ad1fd02b62c115ee191dfd6dc86fa1R236), there was an oversight and I've forgot to update another place where are opening the right conversation panel with service entity. 

We should not overwrite the service entity id with user's id when opening a right sidebar panel. It needs a service id (the one set on backend) to work properly.

Also the isAddMode was moved to the right panel opening params. When it was kept inside the react state (and right panel was not closed before opening again with the avatar click), the add button was visible even though the service was already added.

## Checklist

### DIFF
--- a/src/script/components/Conversation/Conversation.tsx
+++ b/src/script/components/Conversation/Conversation.tsx
@@ -254,11 +254,7 @@ export const Conversation = ({
 
     const serviceEntity = userEntity.isService && (await repositories.integration.getServiceFromUser(userEntity));
 
-    if (serviceEntity) {
-      openRightSidebar(panelId, {entity: serviceEntity}, true);
-    } else {
-      openRightSidebar(panelId, {entity: userEntity}, true);
-    }
+    openRightSidebar(panelId, {entity: serviceEntity || userEntity}, true);
   };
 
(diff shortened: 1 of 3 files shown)


V-Gira (approved):

src/script/components/Conversation/Conversation.tsx
```diff
@@ -255,7 +255,7 @@ export const Conversation = ({
     const serviceEntity = userEntity.isService && (await repositories.integration.getServiceFromUser(userEntity));
 
     if (serviceEntity) {
-      openRightSidebar(panelId, {entity: {...serviceEntity, id: userEntity.id}}, true);
+      openRightSidebar(panelId, {entity: serviceEntity}, true);
```
przemvs: **suggestion:** maybe You can do smth like that:
```suggestion
      openRightSidebar(panelId, {entity: serviceEntity || userEntity }, true);
```

przemvs (approved):
Nice! 🚀

PR #17409, opened 2024-05-15
runfix: adress responsivness of the preferences sections (WPB-7214)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7214" title="WPB-7214" target="_blank">BugWPB-7214</a>  [Web] Implement responsiveness using container queries instead of viewport media queries
  </td></table>
  <br />
 

## Description

We can use [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_containment/Container_queries) to have the content of the preference pages be responsive with its container instead of the viewport.

Standard media queries are used as a fallback.

This allows the preferences page to be responvive regardless of the state of the new sidebar

#### Before: (layout breaks just before going into responsive design with the sidebar opened)
Kooha-2024-05-15-14-58-52

#### After:
Kooha-2024-05-15-15-11-30

### DIFF
--- a/src/script/page/MainContent/panels/preferences/components/PreferencesPage.styles.ts
+++ b/src/script/page/MainContent/panels/preferences/components/PreferencesPage.styles.ts
@@ -1,0 +1,44 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


src/style/content/preferences.less
```diff
@@ -37,26 +37,6 @@
...
-    padding: 40px;
-  }
-
-  @media (max-width: @screen-sm) {
-    padding: 20px;
-  }
-
-  input {
-    color: var(--background);
-  }
```
V-Gira, line 57: This appears to be a years old leftover that we can remove

src/script/page/MainContent/panels/preferences/components/PreferencesPage.styles.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const contentStyle: CSSObject = {
+  width: '100%',
+  height: '100%',
+  padding: '32px 32px 32px 72px',
+  overflowX: 'hidden',
+
+  // Rely on viewport media queries if container queries are not supported by the browser
+  '@media (max-width: 768px)': {
```
V-Gira, line 40: css variables are not supported in media queries and less variables don't work outside of less files unfortunately

PatrykBuniX (approved):

PR #17417, opened 2024-05-16
feat: view port observer on lost visibility

PatrykBuniX:
## Description

Adds an ability to subscribe to visibility loss after the changing the visibility from "visible", to "not visible at all" (not intersecting at all).

## Checklist

### DIFF
--- a/src/script/components/utils/InViewport.tsx
+++ b/src/script/components/utils/InViewport.tsx
@@ -88,6 +88,7 @@ const InViewport: React.FC<InViewportParams & React.HTMLProps<HTMLDivElement>> =
       });
     }
     return () => {
+      // If the element is unmounted, we can trigger the onVisibilityLost callback and release the trackers
       onVisibilityLost?.();
       releaseTrackers();
     };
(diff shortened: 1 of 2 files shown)


src/script/components/utils/InViewport.tsx
```diff
@@ -58,15 +60,23 @@ const InViewport: React.FC<InViewportParams & React.HTMLProps<HTMLDivElement>> =
     const triggerCallbackIfVisible = () => {
       if (inViewport && visible) {
         onVisible();
-        releaseTrackers();
+
+        if (!onVisibilityLost) {
+          releaseTrackers();
+        }
```
PatrykBuniX, line 66: We don't want to release the trackers after detecting element's visibility (we want to detect its visibility lost later).

src/script/util/DOM/viewportObserver.ts
```diff
@@ -17,12 +17,21 @@
...
  */
 
-const observedElements = new Map();
+const observedElements = new Map<
+  Element,
+  {
+    allowBiggerThanViewport?: boolean;
+    requireFullyInView?: boolean;
+    onVisible?: Function;
+    onVisibilityChange?: (isVisible: boolean, isPartiallyVisible: boolean) => void;
```
PatrykBuniX, line 26: Just added types here and renamed `onChange` to `onVisibilityChange` for more clarity.

przemvs (approved):

PR #17422, opened 2024-05-17
fix: show enrollment modal for existing clients after e2ei activation [WPB-9816]

aweiss-dev:
We did automatically "snooze" the certificate enrollment for existing clients, after e2ei feature activation.
This was not supposed to happen, instead we need to show the activation modal to the user, and he/she needs to take action to either enroll or snooze him-/herself

### DIFF
--- a/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.ts
+++ b/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.ts
@@ -91,6 +91,7 @@ export function getEnrollmentTimer(
   identity: WireIdentity | undefined,
   e2eiActivatedAt: number,
   teamGracePeriodDuration: number,
+  isFirstActivation: boolean = false,
 ) {
   if (identity?.status === MLSStatuses.EXPIRED) {
     return {isSnoozable: false, firingDate: Date.now()};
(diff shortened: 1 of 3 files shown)


tlebon (approved):
seems fine / please add.. something to the PR body.

src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.ts
```diff
@@ -99,6 +100,11 @@ export function getEnrollmentTimer(
   const deadline = getGracePeriod(identity, e2eiActivatedAt, teamGracePeriodDuration);
   const nextTick = getNextTick(deadline);
 
+  // For the first activation, we want to trigger the timer immediately
+  if (isFirstActivation) {
+    return {isSnoozable: nextTick > 0, firingDate: Date.now()};
```
PatrykBuniX, line 105: question: Will it be snoozable for the existing mls client (basic credential type) during the grace period? 🤔

aweiss-dev: Depends on the nextTick value I guess.

Nothing changed code related to this case. We just show the modal now with the option to snooze after first activation instead of snoozing automatically.

PatrykBuniX (approved):
Looks fine

PR #17432, opened 2024-05-20
refactor: store initial message in conversation entity

PatrykBuniX:
## Description

`initialMessage` is used to highlight the message when we use the conversation full search. Since it is tied to conversation, it should not live in a global state. This PR moves this value from ContentViewModel to the conversation entity.

## Checklist


przemvs (approved):
🚀

PR #17479, opened 2024-05-28
runfix: don't escape folder name char

PatrykBuniX:
## Description

Since the context menu option is rendered as a react element (and not the `dangerouslySetInnerHTML`) it's safe to skip the char escaping since it will be stringified in the end enyway. See https://github.com/wireapp/wire-webapp/blob/runfix/remove-from-char-escaped/src/script/ui/ContextMenu.tsx#L217. 

This fixes https://github.com/wireapp/wire-webapp/pull/17475 in a proper way (the prev solution won't work for at least german translation).

## Checklist

### DIFF
--- a/src/script/view_model/ListViewModel.ts
+++ b/src/script/view_model/ListViewModel.ts
@@ -374,7 +374,7 @@ export class ListViewModel {
       if (customLabel) {
         entries.push({
           click: () => conversationLabelRepository.removeConversationFromLabel(customLabel, conversationEntity),
-          label: t('conversationsPopoverRemoveFrom', customLabel.name),
+          label: t('conversationsPopoverRemoveFrom', customLabel.name, {}, true),
         });
       }
 


przemvs (approved):

src/script/view_model/ListViewModel.ts
```diff
@@ -374,7 +374,7 @@ export class ListViewModel {
       if (customLabel) {
         entries.push({
           click: () => conversationLabelRepository.removeConversationFromLabel(customLabel, conversationEntity),
-          label: t('conversationsPopoverRemoveFrom', customLabel.name),
+          label: t('conversationsPopoverRemoveFrom', customLabel.name, {}, true),
```
V-Gira, line 377: had to look that up :sweat_smile: 
For reference `{}` : passed to `dangerousSubstitutions` and `true`: passed to `skipEscape`

V-Gira (approved):

PR #17483, opened 2024-05-29
runfix: add breakpoint to bottom positioning

V-Gira:
We can adjusst the bottom positioning according to a breakpoint
Kooha-2024-05-29-11-20-50


src/script/components/Conversation/Conversation.tsx
```diff
@@ -559,7 +560,7 @@ export const Conversation = ({
             conversation={activeConversation}
             css={{
               position: 'absolute',
-              bottom: '56px',
+              bottom: mdBreakpoint ? '100px' : '56px',
```
V-Gira, line 563: replace `"100px"` by the correct px value

svitovyda (approved):
lgtm

PR #17449, opened 2024-05-23
fix: Incoming connection request is not highlighted

przemvs:
https://wearezeta.atlassian.net/browse/WPB-4715

## Description
While user send a connection request and user which got connection request click on pending request, then conversation-list-cell is not highlighted. Also Avatar now is center to the cell.

## Screenshots/Screencast (for UI changes)

Before:
image

After: 
image

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

PR #17503, opened 2024-06-03
fix: Incoming connection request is not highlighted (#17449)

V-Gira:
* fix: Incoming connection request is not highlighted

* return null instead of undefined

## Description

https://github.com/wireapp/wire-webapp/pull/17449 was reverted while merging dev to the new navigation as it was introducing too many conflicts.
This cherry-picks it back and solve the merge conflicts


przemvs (approved):
LGTM! :)

PR #17512, opened 2024-06-04
refactor: Use webgl to compute video blurring

atomrc:
## Description

This wil move the computation of blurred background from the CPU to GPU using webgl. 

This drastically improves the performances of the computation and reduce the load on the CPU. 
Computing a single frame goes from ~30-40ms to <1ms. 

## Technical Details

We have 2 steps to execute for each frame of the video:
- segmenting the video (differentiating  what is background from what is foreground)
- blurring the pixels that were detected as background

### Previous implementation 

The previous implementation did the following:

- doing the segmentation on the GPU
- bringing back the results from the GPU to the main thread (costly operation)
- doing the blurring on the main thread (very costly operation)

This would usually bring the CPU to its knees. 

### Full GPU solution

Now what this PR does is:

- computes the segmentation on the GPU (not changed, this was already the case)
- keep the result only inside the GPU (avoid bringing the mask back to Javascript)
- run the blurring directly on the GPU with the result from the segmentation already in the GPU memory

### Before

Video is laggy and CPU if over-burdened. 
Every frame takes up to 45ms to render (way above our 16ms budget)

https://github.com/wireapp/wire-webapp/assets/1090716/5e158fda-2ce8-4926-b9c0-84264cc5cacd

### After

Video is fluid and each frame takes less than 1ms to render. 
CPU is free to do other tasks

https://github.com/wireapp/wire-webapp/assets/1090716/ec83f970-24dd-4949-a01e-d553f7c0cec9

### Ressources 

Webgl documentation https://webglfundamentals.org/


src/script/media/Blurrer/Blurrer.ts
```diff
@@ -0,0 +1,175 @@
+/*
+ * Wire
+ * Copyright (C) 2020 Wire Swiss GmbH
```
thisisamir98:
```suggestion
 * Copyright (C) 2024 Wire Swiss GmbH
```

src/script/media/Blurrer/Blurrer.ts
```diff
@@ -0,0 +1,177 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {ImageSegmenterResult} from '@mediapipe/tasks-vision';
+// @ts-ignore
+import {createProgramFromSources} from 'webgl-utils.js';
```
atomrc: Sadly the webgl-utils do not have typescript types. 
Since we use a single simple function from it's probably no big deal.

src/script/media/Blurrer/Blurrer.ts
```diff
@@ -0,0 +1,177 @@
...
+
+// @ts-ignore
+import fragmentShader from './fragmentShader.glsl';
+// @ts-ignore
+import vertexShader from './vertexShader.glsl';
```
atomrc: We import those as `raw` no need for a typecheck there

src/script/media/BackgroundBlurrer/fragmentShader.glsl
```diff
@@ -0,0 +1,28 @@
...
+    // This creates a rather rough transition. We need to find a way to smooth the edges and avoid pixelated transitions.
+    if (texture2D(u_mask, v_texCoord).r == 1.0) {
+        gl_FragColor = texture2D(u_image, v_texCoord);
+    } else {
+        vec4 colorSum;
+        for (int i = -8; i < 8; i++) {
+            for (int j = -8; j < 8; j++) {
+                colorSum += texture2D(u_image, v_texCoord + onePixel * vec2(i, j));
+            }
+        }
```
atomrc: This is the blur algo. We run this part of the shader only if the segmentation mask value is != 1

webpack.config.common.js
```diff
@@ -81,6 +81,10 @@ module.exports = {
         },
         test: /\.svg$/,
       },
+      {
+        loader: 'raw-loader',
+        test: /\.glsl$/,
+      },
```
atomrc, line 87: We need the raw loader in order to load the glsl files as text into our js bundle

package.json
```diff
@@ -133,6 +134,7 @@
     "postcss-preset-env": "9.5.14",
     "postcss-scss": "4.0.9",
     "prettier": "3.3.0",
+    "raw-loader": "^4.0.2",
```
PatrykBuniX: `--exact`?

src/script/media/BackgroundBlurrer/BackgroundBlurrer.ts
```diff
@@ -0,0 +1,181 @@
...
+  // Bind it to ARRAY_BUFFER (think of it as ARRAY_BUFFER = positionBuffer)
+  gl.bindBuffer(gl.ARRAY_BUFFER, buffers.position);
+  // Set a rectangle the same size as the image.
+  setRectangle(gl, 0, 0, width, height);
+
+  // provide texture coordinates for the rectangle.
+  gl.bindBuffer(gl.ARRAY_BUFFER, buffers.textcoord);
+  gl.bufferData(
+    gl.ARRAY_BUFFER,
+    // This represents the 6 points of the 2 triangles reprensenting the restangle we want to paint
```
PatrykBuniX:
```suggestion
    // This represents the 6 points of the 2 triangles reprensenting the rectangle we want to paint
```

src/script/media/BackgroundBlurrer/BackgroundBlurrer.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  // start with the original image
+  // Create a texture and put the image in it.
+  const originalImageTexture = createAndSetupTexture(gl);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB, gl.RGB, gl.UNSIGNED_BYTE, videoElement);
+  gl.activeTexture(gl.TEXTURE0);
+  gl.bindTexture(gl.TEXTURE_2D, originalImageTexture);
+
+  // Assiging the segmentation mask to the mask uniform (so that it's accessible to the shader)
+  const segmentationMask = segmentationResults.confidenceMasks?.[0].getAsWebGLTexture() ?? null;
```
PatrykBuniX: How about an extra `?` if we get an empty list? 
```suggestion
  const segmentationMask = segmentationResults.confidenceMasks?.[0]?.getAsWebGLTexture() ?? null;
```

src/script/media/VideoBackgroundBlur.ts
```diff
@@ -62,111 +57,68 @@ function startBlurProcess(
...
+    videoEl.onplay = async () => {
+      const glContext = document.createElement('canvas');
+      glContext.height = videoDimensions.height;
+      glContext.width = videoDimensions.width;
+
+      const gl = initShaderProgram(glContext, videoDimensions);
+      const segmenter = await createSegmenter(glContext);
+
+      const stopBlurProcess = startBlurProcess(segmenter, gl, videoEl, videoDimensions);
+      const videoStream = glContext.captureStream(QualitySettings.framerate).getVideoTracks()[0];
```
PatrykBuniX, line 121: **suggestion**: How about an extra check if the videoStream is defined? Maybe we should add `--noUncheckedIndexedAccess` flag in our tsconfig in general (not in this PR)?

atomrc: humm that's good suggestion. We shall try to add it in a follow up for sure (it might need some code adjustment)

package.json
atomrc: ahah, good catch eagle eye 👀

PatrykBuniX (approved):
Looks solid, good job 🚀

PR #17553, opened 2024-06-10
feat: Improved delivered status indicator WPB-9170

svitovyda:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9170" title="WPB-9170" target="_blank">StoryWPB-9170</a>  Improve  hover states and paddings for delivered status
  </td></table>
  <br />
 

## Description

Now the conversation should have additional empty space at the right, and the last delivered message should contain a translated "Delivered" indicator, visible always (instead of on mouse over).
This feature design was made considering that most of the webapp/desktop users use it on res 800px+, design improvements for smaller screens will be defined later.
After this version is deployed, we might implement another design with checkboxes (under a separate ticket) that can be toggled using debug

_**Note 1:**_ _for the first implementation, where each message had an additional div to the right:_ I assumed that any message header would be short enough to not wrap it into the container that takes `100% - ${deliveredIndicatorWidth}px`. It might be the improvement for the next design iteration to address the small screen look

**_Note 2:_** _for the second implementation, where the entire `MessageList` has a margin:_ Now the `MessageList` has to add and immediately remove the "Delivered" indicator to identify its width. Another option could be to add a static margin (of ~100px) and listen to the message components when one of them will have to add this indicator, then adjust the margin width. Another option could be that each message component could have a margin, but I'm not sure what we will win here

Screenshot 2024-06-10 at 21 32 32

Screenshot 2024-06-10 at 21 32 59

Screenshot 2024-06-10 at 21 33 36

Screenshot 2024-06-10 at 21 34 06

Screenshot 2024-06-10 at 21 35 36

Screenshot 2024-06-12 at 11 50 17

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
@@ -23,6 +23,7 @@ import {QualifiedId} from '@wireapp/api-client/lib/user';
 import cx from 'classnames';
 import ko from 'knockout';
 
+import {DeliveredIndicator} from 'Components/MessagesList/Message/DeliveredIndicator/DeliveredIndicator';
 import {ReadIndicator} from 'Components/MessagesList/Message/ReadIndicator';
 import {Conversation} from 'src/script/entity/Conversation';
 import {CompositeMessage} from 'src/script/entity/message/CompositeMessage';
@@ -48,6 +49,7 @@ import {ContextMenuEntry} from '../../../../ui/ContextMenu';
 import {EphemeralTimer} from '../EphemeralTimer';
 import {MessageTime} from '../MessageTime';
 import {useMessageFocusedTabIndex} from '../util';
+
 export interface ContentMessageProps extends Omit<MessageActions, 'onClickResetSession'> {
   contextMenu: {entries: ko.Subscribable<ContextMenuEntry[]>};
   conversation: Conversation;
@@ -88,6 +90,7 @@ export const ContentMessageComponent = ({
   onClickDetails,
 }: ContentMessageProps) => {
   const messageRef = useRef<HTMLDivElement | null>(null);
+  const [deliveryIndicatorRef, setDeliveryIndicatorRef] = useState<HTMLDivElement | null>(null);
 
   // check if current message is focused and its elements focusable
   const msgFocusState = useMemo(() => isMsgElementsFocusable && isFocused, [isMsgElementsFocusable, isFocused]);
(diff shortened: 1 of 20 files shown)


src/script/components/MessagesList/Message/DeliveredIndicator/DeliveredIndicator.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+  padding: '0 24px 0 16px',
+  height: height ? `${height}px` : 'min-content',
+});
+
+export const DeliveredIndicatorStyles = (isLastDeliveredMessage: boolean): CSSObject => ({
+  color: 'var(--content-message-timestamp)',
+  fontSize: 'var(--font-size-small)',
+  fontWeight: 'var(--font-weight-regular)',
+  wordWrap: 'normal',
+  width: 'min-content',
```
svitovyda, line 36: One translation of "Delivered" is two words, here I make sure they will be wrapped

src/script/components/MessagesList/Message/DeliveredIndicator/DeliveredIndicator.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+import {CSSObject} from '@emotion/react';
+
+export const DeliveryIndicatorContainerStyles = (height?: number): CSSObject => ({
+  display: 'flex',
+  boxSizing: 'border-box',
+  justifyContent: 'center',
+  alignItems: 'center',
+  position: 'absolute',
```
svitovyda, line 24: It's another self-positioned component

src/script/components/MessagesList/Message/ReadReceiptStatus.tsx
```diff
@@ -29,17 +29,11 @@ import {formatTimeShort} from 'Util/TimeUtil';
...
   onClickDetails?: (message: Message) => void;
 }
 
-export const ReadReceiptStatus = ({
-  message,
-  is1to1Conversation,
-  isLastDeliveredMessage,
-  onClickDetails,
-}: ReadReceiptStatusProps) => {
+export const ReadReceiptStatus = ({message, is1to1Conversation, onClickDetails}: ReadReceiptStatusProps) => {
```
svitovyda, line 36: This one can be moved to the `ReadIndicator` folder, which will make it easier to not forget about one of them when the other should be changed

src/style/components/asset/file-asset.less
```diff
@@ -22,7 +22,7 @@
   width: 100%;
   flex-wrap: wrap;
   align-self: center;
-  padding: 12px;
+  padding: 12px 0 12px 12px;
```
svitovyda, line 25: Since the general designer's idea was to introduce additional whitespace at the right of the conversation, I guess we don't need the right padding of the message bodies

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -163,134 +166,138 @@ export const ContentMessageComponent = ({
...
-        className={cx('message-body', {
-          'message-asset': isAssetMessage,
-          'message-quoted': !!quote,
-          'ephemeral-asset-expired': isObfuscated && isAssetMessage,
-          'icon-file': isObfuscated && isFileMessage,
-          'icon-movie': isObfuscated && isVideoMessage,
-        })}
-        {...(ephemeralCaption && {title: ephemeralCaption})}
+        aria-label={messageAriaLabel}
+        className="content-message-wrapper"
```
svitovyda: adding here ```className={cx('content-message-wrapper', { 'width': `calc(100% - ${deliveryIndicatorRef?.offsetWidth || 0}px)` })}``` doesn't work

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -163,134 +166,138 @@ export const ContentMessageComponent = ({
...
+        css={{
+          width: `calc(100% - ${deliveryIndicatorRef?.offsetWidth || 0}px)`,
```
svitovyda: I tried playing with `display: flex` and `overflow` etc. and with some of the message content (images, `pre`) the message was stretching outside of the screen

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -163,134 +166,138 @@ export const ContentMessageComponent = ({
...
-          handleReactionClick={onClickReaction}
-          isMessageFocused={msgFocusState}
-          onTooltipReactionCountClick={() => onClickReactionDetails(message)}
-          onLastReactionKeyEvent={() => setActionMenuVisibility(false)}
-          isRemovedFromConversation={conversation.removed_from_conversation()}
-          users={conversation.allUserEntities()}
-        />
-      )}
-    </div>
+      <DeliveredIndicator
```
svitovyda: Another option was to add it on the level of the `MessageWrapper`, but I'm not sure it won't break again the stretching of the message for content like `pre`, and maybe we will need completely different behaviour for different kinds of messages. Also, it avoids additional `div` containers

src/script/components/MessagesList/Message/DeliveredIndicator/DeliveredIndicator.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+  height: height ? `${height}px` : 'min-content',
+});
+
+export const DeliveredIndicatorStyles = (isLastDeliveredMessage: boolean): CSSObject => ({
+  color: 'var(--content-message-timestamp)',
+  fontSize: 'var(--font-size-small)',
+  fontWeight: 'var(--font-weight-regular)',
+  wordWrap: 'normal',
+  width: 'min-content',
+  visibility: isLastDeliveredMessage ? 'unset' : 'hidden',
```
svitovyda: I don't know how else to add the empty space of the dynamic width (the width of the translated longest word of "Delivered" plus padding). I guess it is possible to set the width once for all somewhere globally, even set the global right margin for the messages container, and only add one `span` for the last delivered message.

src/style/content/conversation/message-list.less
```diff
@@ -274,7 +274,6 @@
 // MESSAGE - BODY
 .message-body {
   position: relative;
-  padding-right: 40px;
```
svitovyda, line 277: Same here - we already have one big whitespace at the right, we don't need this one anymore

tlebon:
338319899-724d7b8e-951b-4630-8cfe-b66947348f2a
regarding this image you shared -> i think the delivered status should be aligned with the time stamp at the top of longer mesages. (also timestamp should not wrap)

svitovyda:
> 338319899-724d7b8e-951b-4630-8cfe-b66947348f2a regarding this image you shared -> i think the delivered status should be aligned with the time stamp at the top of longer mesages. (also timestamp should not wrap)

I now aligned "Delivered" text to the top of the message, and also made the timestamp text `nowrap` (it was broken before my changes). And aligned the `MessageActions` to the most right - 16 px.

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageActions.styles.ts
```diff
@@ -25,20 +25,20 @@ export const messageActionsGroup: CSSObject = {
...
 
-export const messageBodyActions: CSSObject = {
+export const messageBodyActions = (rightMarginWidth: number): CSSObject => ({
   display: 'inline-flex',
   alignItems: 'center',
   minHeight: '32px',
   minWidth: '40px',
   position: 'absolute',
-  right: '16px',
+  right: `${16 - rightMarginWidth}px`,
```
svitovyda, line 34: I need to position reactions menu to look in the old way - 16px from the right of the screen, not of the message container. I can't use `position: fixed` as I need it only horizontally, and vertically it should still be relative to the parent, to be scrolled. I also tried to play with `left: (100vw - 100% - 16px)`, but `100%` is a message container, already resized to the right margin of the `MessageList` container with margin, so it gives the same result as `right: 16px`.

src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -57,14 +58,14 @@ export interface ContentMessageProps extends Omit<MessageActions, 'onClickResetS
...
   isFocused: boolean;
-  isLastDeliveredMessage: boolean;
   message: ContentMessage;
   onClickButton: (message: CompositeMessage, buttonId: string) => void;
   onRetry: (message: ContentMessage) => void;
   quotedMessage?: ContentMessage;
   selfId: QualifiedId;
   isMsgElementsFocusable: boolean;
   onClickReaction: (emoji: string) => void;
+  rightMarginWidth: number;
```
svitovyda, line 68: In this implementation, `rightMarginWidth` gets calculated in `MessageList` and must be used to correctly position `MessageActionsMenu` (in old way, aligned to the screen, ignoring the new margin). It can be made as a context to not do this long chain-propsing

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -268,15 +271,26 @@ export const MessagesList: FC<MessagesListParams> = ({
...
 
+  const setDeliveredIndicatorWidth = (element: HTMLDivElement) => {
+    if (element?.offsetWidth > 0) {
+      setRightMarginWidth(element?.offsetWidth);
+    }
+  };
+
   return (
     <>
+      {rightMarginWidth === 0 && <DeliveredIndicator ref={setDeliveredIndicatorWidth} isLastDeliveredMessage={false} />}
```
svitovyda, line 282: If we agree on this approach, it can be modified to 
```
if(rightMarginWidth === 0) return <DeliveredIndicator ref={setDeliveredIndicatorWidth} isLastDeliveredMessage={false} />

return ( ...
```
to avoid one extra render. Then the `MessageList.test` must be adjusted to ignore the first render

src/script/components/MessagesList/MessageList.tsx
```diff
@@ -121,6 +122,8 @@ export const MessagesList: FC<MessagesListParams> = ({
   const [highlightedMessage, setHighlightedMessage] = useState<string | undefined>(conversation.initialMessage()?.id);
   const conversationLastReadTimestamp = useRef(conversation.last_read_timestamp());
 
+  const [rightMarginWidth, setRightMarginWidth] = useState<number>(0);
```
svitovyda, line 125: This also can be ref, to avoid each time getting recalculated and re-rendered all the time, but should be recalculated on the language switch.

tlebon:
this was superceded by #17611 and #17606

PR #17562, opened 2024-06-12
fix: Display Media Device labels for Firefox 

EnricoSchw:
## Description
In Firefox, the labels of the Media devices can only be read under certain constrains. This means that the device names cannot be displayed in the select menu. This PR is a fix for it

Before this fix, the selection fields are displayed with blank labels or common names:

Bildschirmfoto 2024-06-12 um 14 06 46

Bildschirmfoto 2024-06-12 um 18 21 16

----

With this fix, the selection fields are displayed with device names:

Bildschirmfoto 2024-06-12 um 16 34 24

Bildschirmfoto 2024-06-12 um 18 21 58

----

# Notes
The fix has one disadvantage. I had to change the device list initialization process to avoid asking a user for audio access rights on first login starting page.

Bildschirmfoto 2024-06-24 um 15 06 23

In order to display the correct device labels, we need to make an access authorization query when creating the device list. But we should not ask for microphone and camera access when not needed. 

To avoid a permission query being displayed when starting the app, I have rewritten the `MediaDeviceHandler` and The device list is no longer created in the constructor. The method [`MediaDeviceHandler.initializeMediaDevices`](https://github.com/wireapp/wire-webapp/blob/58675eaa1ab85f2553ede4d49d875991b2627c94/src/script/media/MediaDevicesHandler.ts#L168-L176) is now integrated into the three points in our app where a call can be accepted or started. In addition, this method is also called in the device settings. So, a device list is only created before the first call or when the settings are called. 

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -882,7 +882,10 @@ export class CallingRepository {
       if (newState === VIDEO_STATE.STOPPED) {
         selfParticipant.releaseVideoStream(true);
       } else {
-        this.warmupMediaStreams(call, false, true);
+        this.mediaDevicesHandler
+          .initializeMediaDevices(true)
+          .then(() => this.warmupMediaStreams(call, false, true))
+          .catch(error => this.logger.warn('Failed to start video stream', error));
       }
     }
     this.wCall?.setVideoSendState(this.wUser, this.serializeQualifiedId(call.conversation.qualifiedId), newState);
--- a/src/script/media/MediaDevicesHandler.test.ts
+++ b/src/script/media/MediaDevicesHandler.test.ts
@@ -144,9 +144,14 @@ describe('MediaDevicesHandler', () => {
     ],
   };
 
+  let enumerateDevicesSpy: jasmine.Spy;
+
+  beforeEach(() => {
+    enumerateDevicesSpy = spyOn(window.navigator.mediaDevices, 'enumerateDevices');
+  });
   describe('refreshMediaDevices', () => {
-    it('filters duplicate microphones and keeps the ones marked with "communications"', () => {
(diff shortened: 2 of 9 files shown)


tlebon:
does this change anything for chromium browsers?

tlebon:
thanks for taking the initiative on this!

EnricoSchw:
@tlebon Thx for your fast response

Unfortunately it also has an impact on Chrome. The device lists are created during the initialization process of the app, so that the access popup also appears in Chrome. 

I see three solutions here:
1. A browser switch for Firefox (could be a temporary solution)
2. Only have a numbered device list for Firefox
3. Only initialize the device list when it is needed

What do you think?

More detail:
Some users will click away the pop-up and not allow access. When the mic is needed for a call, the pop-up will no longer appear and the mic will be blocked. We will receive many support requests because the mic is not working.
This is very bad UX and that's why I marked the PR as draft.

WolfiWire:
Do we have the possibility to create a custom dialog (that explains better why we are asking for these access rights), or are we bound to the browser default popup?

EnricoSchw:
@WolfiWire I would use this one.

Bildschirmfoto 2024-06-14 um 10 17 12

EnricoSchw:
@WolfiWire I have rewritten the device list initialization. This means that our app behaves the same for the user as before. Only internally the process is different.

EnricoSchw:
@tlebon Please, could someone take a look at this?

tlebon:
I'm still OoO (sick). Maybe @wireapp/web can take a look. Or else I will tomorrow.

EnricoSchw:
> I'm still OoO (sick). 

Get well soon!

> Maybe @wireapp/web can take a look. Or else I will tomorrow.

Thx. sometime this week would be enough

przemvs (approved):

PR #17567, opened 2024-06-13
runfix: Disable background blurring if webgl is not available

atomrc:
## Description

The blurred background feature needs a version of the wrapper that includes this fix https://github.com/wireapp/wire-desktop/pull/7940
In order to account for older wrapper version, we need to disable the feature in case the wrapper doesn't support hardware acceleration.

Screenshot 2024-06-13 at 10 11 03

### DIFF
--- a/server/Server.ts
+++ b/server/Server.ts
@@ -169,6 +169,7 @@ class Server {
     this.app.use('/proto', express.static(path.join(__dirname, 'static/proto')));
     this.app.use('/style', express.static(path.join(__dirname, 'static/style')));
     this.app.use('/worker', express.static(path.join(__dirname, 'static/worker')));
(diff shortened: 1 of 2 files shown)


svitovyda (approved):

PatrykBuniX (approved):

server/Server.ts
```diff
@@ -169,6 +169,7 @@ class Server {
     this.app.use('/proto', express.static(path.join(__dirname, 'static/proto')));
     this.app.use('/style', express.static(path.join(__dirname, 'static/style')));
     this.app.use('/worker', express.static(path.join(__dirname, 'static/worker')));
+    this.app.use('/assets', express.static(path.join(__dirname, 'static/assets')));
```
atomrc, line 172: This fixes accessing the model file that would not be available to the web server when bundled for production

PR #17569, opened 2024-06-13
runfix: Avoid loosing the blurred background state when call is minimized

atomrc:
## Description

Avoid duplication of state and use a single source of truth to know if the self user has blurred background enabled or not

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -70,6 +70,7 @@ export interface FullscreenVideoCallProps {
   conversation: Conversation;
   isChoosingScreen: boolean;
   isMuted: boolean;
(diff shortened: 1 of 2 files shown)


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -119,6 +119,9 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
     'sharesCamera',
   ]);
 
+  const {blurredVideoStream} = useKoSubscribableChildren(selfParticipant, ['blurredVideoStream']);
+  const hasBlurredBackground = !!blurredVideoStream;
```
atomrc, line 123: This avoids having duplicated states. The participant is now the single source of truth

PatrykBuniX (approved):

PR #17581, opened 2024-06-14
feat: Close "Create Group" modal on Esc key press on the first step WPB-9718

svitovyda:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9718" title="WPB-9718" target="_blank">TaskWPB-9718</a>  Close the "Create Group" modal on keypress "Escape"
  </td></table>
  <br />
 

## Description

If a user is on the first step of the "create new group" modal, the "Esc" key press should close it. On the next steps, it is not so straightforward what the "Esc" should do, as there is a search input and we might need to clear it on the "Esc" press, and there is already data that the user entered and will lose if the modal is closed. We can further discuss it with the design team.

This is part of the new navigation epic, but since it's an independent universal change, we decided to implement it on `dev`

## Checklist

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import React, {useContext, useEffect, useMemo, useState} from 'react';
+import React, {useCallback, useContext, useEffect, useMemo, useState} from 'react';
 
 import {RECEIPT_MODE} from '@wireapp/api-client/lib/conversation/data/ConversationReceiptModeUpdateData';
 import {ConversationProtocol} from '@wireapp/api-client/lib/conversation/NewConversation';
@@ -40,7 +40,7 @@ import {UserSearchableList} from 'Components/UserSearchableList';
 import {generateConversationUrl} from 'src/script/router/routeGenerator';
 import {createNavigate, createNavigateKeyboard} from 'src/script/router/routerBindings';
 import {useKoSubscribableChildren} from 'Util/ComponentUtil';
-import {handleEnterDown, isKeyboardEvent, offEscKey, onEscKey} from 'Util/KeyboardUtil';
+import {handleEnterDown, handleEscDown, isKeyboardEvent} from 'Util/KeyboardUtil';
 import {replaceLink, t} from 'Util/LocalizerUtil';
 import {sortUsersByPriority} from 'Util/StringUtil';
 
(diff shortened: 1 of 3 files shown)


src/script/util/KeyboardUtil.ts
```diff
@@ -83,40 +83,26 @@ export const isRemovalAction = (key: string): boolean => removalKeys.includes(ke
 
 export const isSpaceOrEnterKey = (key: string): boolean => key === KEY.SPACE || key === KEY.ENTER;
 
-type KeyboardHandler = (event: KeyboardEvent) => void;
-
-const escKeyHandlers: KeyboardHandler[] = [];
-
-document.addEventListener('keydown', event => {
```
svitovyda, line 90: 1. This doesn't even work
2. If that would work, we'd have this functionality unpredictably each time we import this file, makes no sense

src/script/util/KeyboardUtil.ts
```diff
@@ -83,40 +83,26 @@ export const isRemovalAction = (key: string): boolean => removalKeys.includes(ke
...
 
-export const handleEnterDown = (event: React.KeyboardEvent<HTMLElement> | KeyboardEvent, callback: () => void) => {
+export const handleEnterDown = (event: ReactKeyboardEvent<HTMLElement> | KeyboardEvent, callback: () => void): void => {
   if (event.key === KEY.ENTER) {
     callback();
   }
-  return true;
+};
+
+export const handleEscDown = (event: ReactKeyboardEvent<HTMLElement> | KeyboardEvent, callback: () => void): void => {
```
svitovyda, line 102: I'm not sure we need it here, but if we use `handleEnterDown` - it makes sense to add the same for `Esc`. I can create a follow-up refactoring MR to replace custom implementations with the usage of this one

src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -325,6 +326,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
       isShown={isShown}
       onClosed={onClose}
       data-uie-name="group-creation-label"
+      onKeyDown={stateIsPreferences ? handleEscape : undefined}
```
svitovyda, line 329: I need to double-check here if the `Esc` keydown should be handled (and allowed), as if not - it should be prevented. Otherwise, a tab switch will happen in the main bar in the background.

przemvs (approved):

PR #17602, opened 2024-06-18
feat: auto clear navigation search input on Escape press

thisisamir98:
## Description
This pull request introduces a feature that enhances the user experience by automatically clearing the navigation search input when the Escape key is pressed. This allows users to quickly reset their search input without manual deletion.

## Changes
- **Added Event Listener for Escape Key:**
  - An event listener is added to the navigation search input field to detect when the Escape key is pressed.
  - When the Escape key is pressed, the input field is cleared by setting its value to an empty string.

## Benefits
- Improves user experience by providing a quick and intuitive way to clear the search input field.
- Increases efficiency by reducing the need for manual deletion of search queries.

## Implementation Details
- The event listener is attached to the navigation search input field.
- The input field's value is set to an empty string (`""`) upon detecting the Escape key press.

### DIFF
--- a/src/script/page/LeftSidebar/panels/Conversations/ConversationHeader/ConversationHeader.tsx
+++ b/src/script/page/LeftSidebar/panels/Conversations/ConversationHeader/ConversationHeader.tsx
@@ -82,7 +82,21 @@ export const ConversationHeader = ({
     amplify.subscribe(WebAppEvents.SHORTCUT.SEARCH, () => {
       inputRef?.current?.focus();
     });
-  }, []);
+
+    const input = inputRef.current;
+
+    function onKeyDown(event: KeyboardEvent) {
+      if (event.key === 'Escape') {
+        setSearchValue('');
+      }
+    }
+
+    input?.addEventListener('keydown', onKeyDown);
+
+    return () => {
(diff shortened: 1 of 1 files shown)


src/script/page/LeftSidebar/panels/Conversations/ConversationHeader/ConversationHeader.tsx
```diff
@@ -82,7 +82,21 @@ export const ConversationHeader = ({
...
+        setSearchValue('');
+      }
+    }
+
+    input?.addEventListener('keydown', onKeyDown);
+
+    return () => {
+      input?.removeEventListener('keydown', onKeyDown);
+    };
+  }, [setSearchValue]);
```
przemvs: You don't need to addEventListener here. You can add on Input onKeyDown method and then do clear value on escape click :)

thisisamir98: @przemvs done in https://github.com/wireapp/wire-webapp/pull/17602/commits/8c1796ddc2eae8e3b76257dcc37e8ae8dae28bb0

przemvs (approved):

PR #17605, opened 2024-06-18
runfix: Allow blurred background to be enabled before the call starts

atomrc:
## Description

This PR will allow the video call background blur setting to be saved and reused across calls. 
We, for now, persist the user preference only for the current session. Later the setting will be saved to a persistent storage and allowed to be changed in the audio/video settings. 

This PR also increase the blurring effect.

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -130,7 +130,7 @@ export class CallingRepository {
   private readonly acceptVersionWarning: (conversationId: QualifiedId) => void;
   private readonly callLog: string[];
   private readonly logger: Logger;
-  private isBlurred = localStorage.getItem('isBlurred') === 'true';
+  private enableBackgroundBlur = false;
   private avsVersion: number = 0;
   private incomingCallCallback: (call: Call) => void;
   private isReady: boolean = false;
@@ -276,8 +276,7 @@ export class CallingRepository {
     if (!videoFeed) {
       return;
     }
-    this.isBlurred = enable;
-    localStorage.setItem('isBlurred', enable.toString());
+    this.enableBackgroundBlur = enable;
     const newVideoFeed = enable ? ((await selfParticipant.setBlurredBackground(true)) as MediaStream) : videoFeed;
     this.changeMediaSource(newVideoFeed, MediaType.VIDEO, false);
   }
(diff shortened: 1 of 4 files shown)


src/script/media/VideoBackgroundBlur.ts
```diff
@@ -32,7 +32,7 @@ enum FRAMERATE {
 
 const QualitySettings = {
   segmentationModel: SEGMENTATION_MODEL.PERFORMANCE,
-  framerate: FRAMERATE.HIGH,
+  framerate: FRAMERATE.LOW,
```
atomrc, line 35: This should improve performances a bit

src/script/media/BackgroundBlurrer/fragmentShader.glsl
```diff
@@ -1,4 +1,4 @@
-#define BLUR_QUALITY 8 //The higher the value, the more blur. Must be an even number.
+#define BLUR_QUALITY 14 //The higher the value, the more blur. Must be an even number.
```
atomrc, line 1: This will increase the blurring effect

src/script/calling/Participant.ts
```diff
@@ -18,68 +18,72 @@
...
+    this.hasPausedVideo = pureComputed(() => {
       return this.videoState() === VIDEO_STATE.PAUSED;
     });
-    this.videoStream = ko.observable();
-    this.audioStream = ko.observable();
-    this.isActivelySpeaking = ko.observable(false);
-    this.startedScreenSharingAt = ko.observable();
-    this.isMuted = ko.observable(false);
-    this.isSendingVideo = ko.pureComputed(() => {
+    this.isSendingVideo = pureComputed(() => {
```
atomrc, line 67: Those are only changes to make typescript and eslint happier. Nothing interesting to see there

src/script/calling/CallingRepository.ts
```diff
@@ -131,6 +130,7 @@ export class CallingRepository {
   private readonly acceptVersionWarning: (conversationId: QualifiedId) => void;
   private readonly callLog: string[];
   private readonly logger: Logger;
+  private enableBackgroundBlur = false;
```
atomrc, line 133: in a further feature, this will be moved to a proper setting in the `A/V` settings

PatrykBuniX (approved):

PR #17615, opened 2024-06-19
runfix: Improve background edge blurring

atomrc:
## Description

We were previously trying to manually do a smoothing of the edge detected by the segmenter. 
But turned out the result of the segmentation is already a float value that gives how much of the clear pixel should be used compared to the blurred pixel. 

So instead of manually computing a smooth value, we can just use the segmenter's raw value 

### Before
image

### After 
image


PatrykBuniX (approved):
Great finding!

PR #17617, opened 2024-06-19
chore: type safe icons

PatrykBuniX:
## Description

Adds type-safety to our svg icons utility.

I've defined a generate-icon-names script that will generate icon names and store them as an array of strings. The svg util then uses the array to create all the icon components. The same array of svg file names is used to generate types (string literals). The mentioned script is being run after the app config (including icon resources) is copied from the config repo (so in postinstall lifecycle script).

Wire 2024-06-19 at 15-21-09

Previously there was no typecheck at all and trying to render an icon that didn't exist could silently crash the app.

## Checklist


bin/generate_icon_names.ts
```diff
@@ -0,0 +1,58 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
```
atomrc: I'm not sure we need the header on this file since it's never going to be published on GH 🤔

PatrykBuniX: Oh right, my first idea was that it will be published, but I will remove it 👍

src/script/util/TypeUtil.ts
```diff
@@ -0,0 +1,24 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export type RemoveSuffix<S extends string, Suffix extends string> = S extends `${infer P}${Suffix}` ? P : S;
+
+export type PascalCase<S extends string> = S extends `${infer F}-${infer R}`
+  ? `${Capitalize<F>}${PascalCase<Capitalize<R>>}`
+  : Capitalize<S>;
```
atomrc, line 24: someone had some typescript fun :D

src/script/util/SVGProvider/SVGProvider.ts
atomrc, line 1: I'm actually thinking we could go one step further and generate the typescript icon at build time and avoid this runtime overhead. 

the `generate_icons.ts` could actually just copy/paste the raw input of the svg files into a single typescript file that `exports` every single icon. 

This way we would have a way, in dev, to easily see all the typescript icons. 

WDYT?

aweiss-dev (approved):

atomrc (approved):

PR #17618, opened 2024-06-19
perf: Improve caching of root js files

atomrc:
## Description

We currently only cache js files for 5min. 
But, we don't have a version query params that prevents the browser from using the cache in case a new version is detected. 
We can then cache the js files "forever" (max allowed caching value is 1 year) and let the `v` query param handle the cache invalidation.

### DIFF
--- a/server/Server.ts
+++ b/server/Server.ts
@@ -96,7 +96,8 @@ class Server {
       this.app.use((req, res, next) => {
         // If the user agent adds a v param, it means that its requesting a particular version of the file and that could be cached forever since the file will never change.
         const hasCacheVersionParam = req.query.v && typeof req.query.v === 'string';
-        const maxAge = hasCacheVersionParam ? 31536000 : this.config.CACHE_DURATION_SECONDS;
+        const oneYear = 31536000;
+        const maxAge = hasCacheVersionParam ? oneYear : this.config.CACHE_DURATION_SECONDS;
         const milliSeconds = 1000;
         res.header('Cache-Control', `public, max-age=${maxAge}`);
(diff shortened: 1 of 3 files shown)


src/page/auth.ejs
```diff
@@ -19,16 +19,16 @@
 
     <title><%= BRAND_NAME %></title>
 
-    <script src="../min/basicBrowserFeatureCheck.js?<%= VERSION %>"></script>
-    <script src="../min/checkBrowser.js?<%= VERSION %>"></script>
+    <script src="../min/basicBrowserFeatureCheck.js?v=<%= VERSION %>"></script>
+    <script src="../min/checkBrowser.js?v=<%= VERSION %>"></script>
```
atomrc, line 23: in the HTML files we manually add this `v` parameter that will allow the cache to be invalidated as soon as a new version is available.

server/Server.ts
```diff
@@ -94,9 +94,13 @@ class Server {
       this.app.use(nocache());
     } else {
       this.app.use((req, res, next) => {
+        // If the user agent adds a v param, it means that its requesting a particular version of the file and that could be cached forever since the file will never change.
+        const hasCacheVersionParam = req.query.v && typeof req.query.v === 'string';
+        const oneYear = 31536000;
+        const maxAge = hasCacheVersionParam ? oneYear : this.config.CACHE_DURATION_SECONDS;
```
atomrc, line 100: server side, when this `v` param is given we can assume the client handles the caching itself and add a max age of 1 year

PatrykBuniX (approved):

PR #17621, opened 2024-06-20
chore: generate react icons at build time

atomrc:
## Description

This will generate all the React icon components at build time and avoid the overhead of generating icons at runtime


przemvs (approved):

src/script/auth/page/Index.tsx
```diff
@@ -67,16 +66,13 @@ const IndexComponent = ({defaultSSOCode}: Props & ConnectedProps & DispatchProps
...
     <Page>
       <ContainerXS centerText verticalCenter style={{width: '380px'}}>
-        <SVGIcon
+        <LogoFullIcon
           aria-hidden="true"
-          scale={1.3}
-          realWidth={78}
-          realHeight={25}
+          width={102}
+          height={33}
```
atomrc, line 72: Here we apply the `scale` factor directly (that's why the height/width values are different)

PatrykBuniX (approved):
Nice one!

PR #17640, opened 2024-06-24
fix: Do not intercept right click on rendered message links [WPB-6546]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-6546" title="WPB-6546" target="_blank">BugWPB-6546</a>  Right click on email link opens standard email app instead of option to copy the link
  </td></table>
  <br />
 

## Description

A right click should not trigger the custom click handler and let the browser handle it with the default behavior. 

### Before 

https://github.com/wireapp/wire-webapp/assets/1090716/b0b6cf00-74c4-4f09-80a8-9be91f6110b3

### After 

https://github.com/wireapp/wire-webapp/assets/1090716/6a36fda0-78ed-4fbe-b484-f300c074031a

### DIFF
--- a/src/script/components/Conversation/Conversation.tsx
+++ b/src/script/components/Conversation/Conversation.tsx
@@ -279,10 +279,6 @@ export const Conversation = ({
   };
 
   const handleMarkdownLinkClick = (event: MouseEvent | KeyboardEvent, messageDetails: MessageDetails) => {
-    if (isAuxRightClickEvent(event)) {
-      // Default browser behavior on right click
-      return true;
-    }
     const href = messageDetails.href!;
     PrimaryModal.show(PrimaryModal.type.CONFIRM, {
       primaryAction: {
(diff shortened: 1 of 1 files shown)


src/script/components/Conversation/Conversation.tsx
```diff
@@ -279,10 +279,6 @@ export const Conversation = ({
   };
 
   const handleMarkdownLinkClick = (event: MouseEvent | KeyboardEvent, messageDetails: MessageDetails) => {
-    if (isAuxRightClickEvent(event)) {
-      // Default browser behavior on right click
-      return true;
-    }
```
atomrc, line 285: This check is now done in the parent function (see https://github.com/wireapp/wire-webapp/pull/17640/files#diff-3d29d09ab8471b4309568b6b6e1448ca6de1f9fafbb843c9430201c305bdfe92L328)

przemvs (approved):

V-Gira (approved):

PR #17644, opened 2024-06-25
fix: Save edited message state in editor draft [WPB-9846]

atomrc:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9846" title="WPB-9846" target="_blank">BugWPB-9846</a>  [web] original message of edit is lost when switching conversation
  </td></table>
  <br />
 

### Before 

https://github.com/wireapp/wire-webapp/assets/1090716/32393bf1-c607-445f-ba5a-f09af71184f7

### After

https://github.com/wireapp/wire-webapp/assets/1090716/18864f12-e925-466a-84cb-6150f25719ae

### DIFF
--- a/src/script/components/InputBar/util/DraftStateUtil.ts
+++ b/src/script/components/InputBar/util/DraftStateUtil.ts
@@ -25,6 +25,7 @@ import {StorageKey, StorageRepository} from '../../../storage';
 export interface DraftState {
   editorState: string | null;
   messageReply?: Promise<ContentMessage>;
+  editedMessage?: Promise<ContentMessage>;
 }
 
 const generateConversationInputStorageKey = (conversationEntity: Conversation): string =>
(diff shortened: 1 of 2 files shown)


src/script/components/InputBar/util/DraftStateUtil.ts
```diff
@@ -25,49 +25,62 @@ import {StorageKey, StorageRepository} from '../../../storage';
...
+  const editedMessageId = storageValue?.editedMessageId;
 
   let messageReply = null;
 
   if (replyMessageId) {
     const message =
-      (await messageRepository.getMessageInConversationById(conversationEntity, replyMessageId)) ||
-      (await messageRepository.getMessageInConversationByReplacementId(conversationEntity, replyMessageId));
+      (await messageRepository.getMessageInConversationById(conversation, replyMessageId)) ||
+      (await messageRepository.getMessageInConversationByReplacementId(conversation, replyMessageId));
```
przemvs: **_suggestion:_** Maybe we can make a function to get message, You using it in 2 places ;) 
Smth like this:

```
const getMessage = async (messageId) => {
      return (await messageRepository.getMessageInConversationById(conversation, messageId)) ||
      (await messageRepository.getMessageInConversationByReplacementId(conversation, messageId));
}

And in if's: 

const message = await getMessage(replyMessageId);

const message = await getMessage(editedMessageId);
```

atomrc: true that 👌  good call

przemvs (approved):
Looks solid! 🚀

PR #17654, opened 2024-06-26
chore: use node20 environment for staging

atomrc:
## Description

This will start using the `wire-webapp-staging` (running node20) as the deployment target for staging bumps.


tlebon:
this is superceded by #17598 .  i think that one slipped thru the cracks since i submitted it before vacation but didnt publicize it enough

PR #17656, opened 2024-06-26
chore: initialize datadog as soon as possible

atomrc:
## Description
 
Currently datadog does not receive events that are sent when the `core` is initialized. We are missing some critical data in there. 
This will make sure datadog is running as soon as we have loaded the self user

### DIFF
--- a/src/script/main/app.ts
+++ b/src/script/main/app.ts
@@ -372,6 +372,11 @@ export class App {
       onProgress(2.5);
       telemetry.timeStep(AppInitTimingsStep.RECEIVED_ACCESS_TOKEN);
 
+      const selfUser = await this.repository.user.getSelf([{position: 'App.initiateSelfUser', vendor: 'webapp'}]);
+
+      await initializeDataDog(this.config, selfUser.qualifiedId);
+      onProgress(5, t('initReceivedSelfUser', selfUser.name()));
+
       try {
         await this.core.init(clientType);
       } catch (error) {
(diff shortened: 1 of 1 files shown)


src/script/main/app.ts
```diff
@@ -384,7 +389,7 @@ export class App {
         userRepository.addClientToUser(userId, newClient, true);
       });
 
-      const selfUser = await this.initiateSelfUser();
+      await this.initiateSelfUser(selfUser);
```
atomrc, line 392: We can only initialize the self user here, since we need the DB to have been created by the core

src/script/main/app.ts
```diff
@@ -372,6 +372,11 @@ export class App {
       onProgress(2.5);
       telemetry.timeStep(AppInitTimingsStep.RECEIVED_ACCESS_TOKEN);
 
+      const selfUser = await this.repository.user.getSelf([{position: 'App.initiateSelfUser', vendor: 'webapp'}]);
+
+      await initializeDataDog(this.config, selfUser.qualifiedId);
```
atomrc, line 377: We first fetch (only fetching, not initializing at all) the self user, then we can use it to enable datadog

src/script/main/app.ts
```diff
@@ -372,6 +372,11 @@ export class App {
...
+      onProgress(5, t('initReceivedSelfUser', selfUser.name()));
```
atomrc, line 378: Nice little UX improvement, we show the user name in the loader earlier in the process :) 
No more empty loading bar for a while

V-Gira (approved):

src/script/main/app.ts
V-Gira: Was about to ask why this moved :D

PR #17724, opened 2024-07-08
fix: Hide delivery message for group conversation

przemvs:
## Description

Hide delivery message for group conversation. We want to display only for 1:1 conversation delivery state icon.

## Checklist

- use (x) data
- be sure to look at ...


src/script/components/MessagesList/Message/ContentMessage/ContentMessage.tsx
```diff
@@ -265,9 +267,11 @@ export const ContentMessageComponent = ({
...
         </div>
 
-        <div css={deliveredMessageIndicator}>
-          <DeliveredMessage isLastDeliveredMessage={isLastDeliveredMessage} />
-        </div>
+        {message.expectsReadConfirmation && (
+          <div css={deliveredMessageIndicator}>
+            <DeliveredMessage isLastDeliveredMessage={isLastDeliveredMessage} is1to1Conversation={is1to1} />
+          </div>
+        )}
```
PatrykBuniX: Wouldn't it be easier if we just didn't render the component at all if it's `!is1to1`? Passing a prop and rendering `null` doesn't really make sense to me 🤔  Same with `isLastDeliveredMessage`

przemvs: if we don't render this div, then we will have jumping content :(

przemvs: Also we using this component in 2 places, so better to pass inside component 2 props, and then component is reusable.

PatrykBuniX: But we can render the div anyway and just render the message conditionally

PatrykBuniX (approved):

PR #17757, opened 2024-07-15
fix:  enumerate media device label for firefox ii

EnricoSchw:
## Description
The labels of the media devices can only be read under certain constrains. This means that the device names cannot be displayed in the select menu. This PR is a fix for it.

I tried to not change the permission pop up flow, so that the user doesn't notice any difference. However, the device list is now only created before a call ore in the device setup.

**Note:**
I tried to implement the fix in a way that doesn't require additional security prompts. If the device list does not yet exist, it will be generated before each call. An active stream must exist to access the labels. It should be noted that we cannot read video labels if a call is answered without a video camera. We can only solve this special case by making a camera security prompt, which we tried to avoid. 

Therefore, this special case remains unresolved. But I have prepared the corresponding solution and left a TODO. Another option would be to change the GUI. We would not offer the user a camera selection as long as they have never turned on a camera.

Before this fix, the selection fields are displayed with blank labels or common names:

Bildschirmfoto 2024-06-12 um 14 06 46

Bildschirmfoto 2024-06-12 um 18 21 16

----

With this fix, the selection fields are displayed with device names:

Bildschirmfoto 2024-06-12 um 16 34 24

Bildschirmfoto 2024-06-12 um 18 21 58

----

## Checklist

- use (x) data
- be sure to look at ...


src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,35 @@ export class MediaStreamHandler {
...
+  requestMediaSreamAccess(video: boolean): Promise<MediaStream> {
+    const audio = true;
+    const screen = false;
+    return window.navigator.mediaDevices
+      .getUserMedia({audio, video})
+      .then((mediaStream: MediaStream) => {
+        return mediaStream;
+      })
+      .catch((error: Error) => {
+        const name = error.name as MEDIA_STREAM_ERROR;
```
przemvs: Please to try not to use as. We have some TypeGuards, You. can take a look there how it looks: src/script/util/TypePredicateUtil.ts

przemvs (approved):

PR #17766, opened 2024-07-17
feat: remove usage of user privacy field [WPB-10025]

aweiss-dev:
We need to remove the send anonymous usage data and crash reports checkbox.
We should use the telemetry_sharing attribute in the future for everything related to data sharing.


package.json
```diff
@@ -6,9 +6,10 @@
     "@lexical/history": "0.16.1",
     "@lexical/react": "0.16.1",
     "@mediapipe/tasks-vision": "0.10.14",
+    "@wireapp/api-client": "^27.1.0",
```
V-Gira: don't think api-client should show up here, isn't it handled by core?

aweiss-dev: That is true. Gonna remove it

src/script/page/MainContent/panels/preferences/accountPreferences/DataUsageSection.tsx
```diff
@@ -60,31 +59,21 @@ const DataUsageSection: React.FC<DataUsageSectionProps> = ({
...
-      setOptionPrivacy(settings.privacy.improve_wire);
       setOptionTelemetry(settings.privacy.telemetry_sharing);
     };
     amplify.subscribe(WebAppEvents.PROPERTIES.UPDATED, updateProperties);
     return () => amplify.unsubscribe(WebAppEvents.PROPERTIES.UPDATED, updateProperties);
   }, []);
 
+  const showCountly = !isTeam || (isTeam && isCountlyEnabled);
+
+  if (!showCountly && !isActivatedAccount) {
```
V-Gira, line 70: Looks like it should be an `OR` at first glance, is that not the case?

aweiss-dev: No. If both statements are falsy, the component should return null

If one of them is truthy, we want to render the component

przemvs (approved):

PR #17770, opened 2024-07-17
feat: Add feature flag for in call reactions (WPB-10240)

thisisamir98:
Only enable in call reactions if FEATURE_ENABLE_IN_CALL_REACTIONS is set to "true"


tlebon (approved):
youll add this to the config repo also?

thisisamir98:
> youll add this to the config repo also?

sure. here is the PR: https://github.com/wireapp/wire-web-config-default/pull/113

V-Gira (approved):

PR #17775, opened 2024-07-18
runfix: use removedFromConversation method to check if user is removed from conversation

przemvs:
## Description

Fix for check if user is removed from conversation. Now using proper method to checking if user is removed.

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/components/Conversation/Conversation.tsx
+++ b/src/script/components/Conversation/Conversation.tsx
@@ -538,7 +538,8 @@ export const Conversation = ({
           />
 
           {isConversationLoaded &&
-            (isReadOnlyConversation || isSelfUserRemoved ? (
+            !isSelfUserRemoved &&
+            (isReadOnlyConversation ? (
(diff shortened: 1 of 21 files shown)


src/script/components/Conversation/Conversation.tsx
```diff
@@ -540,7 +538,7 @@ export const Conversation = ({
           />
 
           {isConversationLoaded &&
-            (isReadOnlyConversation || (!isTemporaryGuest && !isActiveParticipant) ? (
+            (isReadOnlyConversation || isSelfUserRemoved ? (
```
PatrykBuniX: It would make more sense not to render neither inputbar nor readonly message (so `isConversationLoaded && !isSelfUserRemoved`), I don't think we have a readonly message for this case, do we?

tlebon (approved):
looks like a nice improvement!

PatrykBuniX (approved):

V-Gira (approved):

PR #17812, opened 2024-07-23
runfix: update the view on opening deep link from the connect tab (WPB-9507)

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9507" title="WPB-9507" target="_blank">BugWPB-9507</a>  [Web] Opening conversation via deeplink does not update the view
  </td></table>
  <br />
 

## Description

Opening a deep link from the connect tab should switch the sidebar to a view where the conversation is displayed, this switches to the Recent view

Since the app initialize opening the last conversation, we do not store the state of the sidebar when it's set to Connect (in addition to Preferences), since this state does not show the conversation list

### DIFF
--- a/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.ts
+++ b/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.ts
@@ -60,8 +60,9 @@ const useSidebarStore = create<SidebarStore>()(
       storage: createJSONStorage(() => localStorage),
       partialize: state => ({
         status: state.status,
-        currentTab:
-          state.currentTab === SidebarTabs.PREFERENCES || SidebarTabs.CONNECT ? SidebarTabs.RECENT : state.currentTab,
+        currentTab: [SidebarTabs.PREFERENCES, SidebarTabs.CONNECT].includes(state.currentTab)
+          ? SidebarTabs.RECENT
+          : state.currentTab,
       }),
     },
   ),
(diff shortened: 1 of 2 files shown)


src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.ts
```diff
@@ -60,7 +60,8 @@ const useSidebarStore = create<SidebarStore>()(
       storage: createJSONStorage(() => localStorage),
       partialize: state => ({
         status: state.status,
-        currentTab: state.currentTab === SidebarTabs.PREFERENCES ? SidebarTabs.RECENT : state.currentTab,
+        currentTab:
+          state.currentTab === SidebarTabs.PREFERENCES || SidebarTabs.CONNECT ? SidebarTabs.RECENT : state.currentTab,
```
PatrykBuniX: Shouldn't it be more like `[SidebarTabs.PREFERENCES, SidebarTabs.CONNECT].includes(state.currentTab)  ? SidebarTabs.RECENT : state.currentTab`

if state.currentTab is different than preferences, it will resolve to `false` and go to `OR` statement and then `SidebarTabs.CONNECT` itself is always truthy so it will always be `SidebarTabs.RECENT` 🤔

PatrykBuniX (approved):

thisisamir98 (approved):

PR #17819, opened 2024-07-25
chore: Use @roamhq/wrtc instead of @koush/wrtc

thisisamir98:
## Description

@koush/wrtc includes [ip](https://www.npmjs.com/package/ip) as a dependency, which is flagged as a high vulnerability, see https://github.com/wireapp/wire-webapp/security/dependabot/129

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@emotion/eslint-plugin": "11.11.0",
     "@faker-js/faker": "8.4.1",
     "@formatjs/cli": "6.2.12",
-    "@koush/wrtc": "0.5.3",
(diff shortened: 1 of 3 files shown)


V-Gira (approved):

thisisamir98:
Unfortunately removing this dependency did not fully fix this issue as we still 3 dependencies which have a dependency on `ip` package. 

- `jest-jasmine2` 
- `jest`
- `workbox-webpack-plugin`

For now we just have to wait until the maintainer of `ip` publishes a fix for https://github.com/github/advisory-database/pull/4619

PR #17733, opened 2024-07-10
feat: calling popout version 2

PatrykBuniX:
## Description

Adds a popout calling view in place of fullscreen view. 

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

PR #17822, opened 2024-07-26
feat: calling popout version 2 (#17733)

PatrykBuniX:
The original PR was mistakenly merged (and reverted). Original PR: https://github.com/wireapp/wire-webapp/pull/17733


src/script/calling/CallState.ts
```diff
@@ -84,8 +116,86 @@ export class CallState {
     });
     this.isSpeakersViewActive = ko.pureComputed(() => this.activeCallViewTab() === CallViewTab.SPEAKERS);
 
-    this.isChoosingScreen = ko.pureComputed(
+    this.hasAvailableScreensToShare = ko.pureComputed(
```
przemvs, line 116: We need 2 same variables for checking this? 🤔 You have similar on line 107

src/script/components/calling/ChooseScreen.tsx
```diff
@@ -17,23 +17,38 @@
...
 export interface ChooseScreenProps {
-  cancel: () => void;
   choose: (screenId: string) => void;
-  screens: Screen[];
-  windows: Screen[];
+  callState?: CallState;
 }
 
-const ChooseScreen: React.FC<ChooseScreenProps> = ({cancel, choose, screens = [], windows = []}: ChooseScreenProps) => {
+const ChooseScreen: React.FC<ChooseScreenProps> = ({
```
przemvs: You don't need there additional types
```suggestion
const ChooseScreen = ({
```

src/script/components/calling/ChooseScreen.tsx
```diff
@@ -56,18 +71,18 @@ const ChooseScreen: React.FC<ChooseScreenProps> = ({cancel, choose, screens = []
         data-uie-name={uieName}
         onClick={() => choose(id)}
       >
-        <img className="choose-screen-list-image" src={thumbnail.toDataURL()} role="presentation" alt="" />
+        <img className="choose-screen-list-image" src={thumbnail?.toDataURL()} role="presentation" alt="" />
```
przemvs: What we will display if there is no thumbnail? 🤔

src/script/components/calling/ChooseScreen.tsx
thisisamir98: because of lint error I applied the changes locally and pushed https://github.com/wireapp/wire-webapp/pull/17822/commits/bfbb157a5e1e729a6ac9f4c3f6c5b279fffbe3a3

src/script/calling/CallState.ts
thisisamir98: deleted in https://github.com/wireapp/wire-webapp/pull/17822/commits/d5172685e52df703b0b0132864f8631c31171dc4

src/script/components/calling/ChooseScreen.tsx
thisisamir98: we don't have a mock image to use when there are no thumbnails unfortunately

przemvs: Ask Wolfgang about this case, because it will be nice to show smth instead of empty thing

przemvs (approved):

src/script/components/calling/ChooseScreen.tsx
thisisamir98: According to electron docs thumbnail property is always there, https://github.com/wireapp/wire-webapp/pull/17822/commits/4e5fd435a6113941e9fb0968ecdff1c487c111fe

https://www.electronjs.org/docs/latest/api/structures/desktop-capturer-source

PR #17840, opened 2024-07-29
fix: Correctly match users with umlaut characters in new navigation search (WPB-10394)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10394" title="WPB-10394" target="_blank">BugWPB-10394</a>  [web] search in new-nav misses special characters
  </td></table>
  <br />
 

### BEFORE:
image

### AFTER
image

### DIFF
--- a/src/script/page/LeftSidebar/panels/Conversations/ConversationsList.tsx
+++ b/src/script/page/LeftSidebar/panels/Conversations/ConversationsList.tsx
@@ -137,15 +137,11 @@ export const ConversationsList = ({
   const isFolderView = currentTab === SidebarTabs.FOLDER;
 
   const getConversationView = () => {
-    const filterByName = (conversation: Conversation) =>
-      conversation.display_name().toLowerCase().includes(conversationsFilter.toLowerCase());
-
     if (isFolderView && currentFolder) {
       return (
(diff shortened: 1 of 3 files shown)


src/script/page/LeftSidebar/panels/Conversations/ConversationsList.tsx
```diff
@@ -155,7 +151,7 @@ export const ConversationsList = ({
 
     return (
       <>
-        {conversations.filter(filterByName).map((conversation, index) => (
```
thisisamir98, line 158: While investigating this bug I found out that we are actually filtering the conversations twice when a filter word is there and second filtering is unnecessarily as it's done already in src/script/page/LeftSidebar/panels/Conversations/helpers.tsx

przemvs (approved):

V-Gira (approved):

thisisamir98:
https://github.com/wireapp/wire-webapp/pull/17840/commits/06a29751da8f1f240b80c0a1bae9d1db35c1fa20

This test had to be removed as filtering conversations isn't done in `<ConversationsList />` anymore.

PR #17843, opened 2024-07-30
feat: change screen share constrains [WPB-9950]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9950" title="WPB-9950" target="_blank">StoryWPB-9950</a>  [AVS] Change Screen Share constrains for better Screen Share quality in 1:1 Call
  </td></table>
  <br />
 

## Description

With this pull request, we are increasing the resolution for screen sharing. To do this, we had to remove the bandwidth limitation in [wire-avs](https://github.com/wireapp/wire-avs/releases/tag/9.8.14) for One to one call. Because Chrome and Firefox can handle this automatically, so it is no necessary.

## Checklist


src/script/media/MediaConstraintsHandler.ts
```diff
@@ -57,28 +57,19 @@ export class MediaConstraintsHandler {
           DESKTOP_CAPTURER: {
             mandatory: {
               chromeMediaSource: 'desktop',
-              maxHeight: 1080,
-              minHeight: 1080,
             },
           },
           DISPLAY_MEDIA: {
             frameRate: 5,
```
tlebon, line 63: will this value be superceded in a 1:1 call?

src/script/media/MediaConstraintsHandler.ts
```diff
@@ -57,28 +57,19 @@ export class MediaConstraintsHandler {
           DESKTOP_CAPTURER: {
             mandatory: {
               chromeMediaSource: 'desktop',
-              maxHeight: 1080,
-              minHeight: 1080,
```
tlebon, line 61: can you also discuss removing the height values now?

src/script/media/MediaConstraintsHandler.ts
EnricoSchw: Yes, and in a conference call too.

By removing this value, the device determines the resolution. If you have a high resolution screen, this is it. WebRTC then reduces the resolution according to the bandwidth. This works well in 1:1 call. In Conference Call WebRTC can not reduces the resolution, currently.

src/script/media/MediaConstraintsHandler.ts
EnricoSchw: By removing this value, the device determines the resolution. If you have a high resolution screen, this is it. WebRTC then reduces the resolution according to the bandwidth. This works well in 1:1 call. In Conference Call WebRTC can not reduces the resolution, currently.

src/script/media/MediaConstraintsHandler.ts
EnricoSchw: Usually not
In our test, the value was exceeded briefly, but that is probably due to the sending buffer. But the value hardly increased noticeably.

tlebon (approved):

PR #17844, opened 2024-07-31
chore: Allow targetting locally running backend

akshaymankar:
## Description

`http://localhost:4570` serves the fake S3 service when running backend locally. So it needs to be part of `connect-src` in the CSP header.

The backend doesn't run with TLS for client API locally, so it needs to be connected to without TLS. If the webapp server runs on TLS, the websocket connection is automatically upgraded to TLS too, so the webapp server also needs to run without TLS.

Finally there were some CSP issues when running webapp server on runs on `http://local.zinfra.io:8081`, moving it to `http://localhost:8081` fixes it. I am not sure what this one is.

## Checklist

### DIFF
--- a/.env.localhost
+++ b/.env.localhost
@@ -5,7 +5,7 @@ APP_BASE="https://local.zinfra.io:8081"
 
 FORCED_CONFIG_URL="https://github.com/wireapp/wire-web-config-wire"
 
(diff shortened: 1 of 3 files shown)


.env.localhost
```diff
@@ -30,3 +30,7 @@ BACKEND_WS="wss://staging-nginz-ssl.zinfra.io"
 # BACKEND_REST="https://prod-nginz-https.wire.com"
 # BACKEND_WS="wss://prod-nginz-ssl.wire.com"
 
+# local
```
tlebon:
```suggestion
# Use these vars if you are running a local backend. (Rare)
```

tlebon (approved):

tlebon:
thanks for supporting the web team, we are hiring ;)

PR #17875, opened 2024-08-05
fix: clone audio track to avoid ending track

EnricoSchw:
## Description

When the device selection page is current track will be stopped.. Currently cloning saves us from ending the track in the call.

## Checklist

Cloning tracks is bad behavior because. This can cause side effects. The original track can still exist and record even though the conference has ended. Or the sent track cannot be muted.

- use (x) data
- be sure to look at ...


tlebon (approved):
are there other instances of cloning that were removed as well?

PR #17877, opened 2024-08-08
fix: Allow user to type while calling pop out window is open (WPB-10463)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10463" title="WPB-10463" target="_blank">BugWPB-10463</a>  [Web Edge] With popped out calling view, i cannot type into the message input bar
  </td></table>
  <br />
 

## Description
This PR removes the `useEffect` hook that was previously used to handle keyboard events in the video calling feature.

#### Context

Back when we did not have Picture-in-Picture (PiP) functionality, we added a `useEffect` to listen for `keydown` events and prevent typing outside of the video calling interface. This was intended to avoid accidental typing or interactions in the main application while the video calling view was active.

However, with the introduction of PiP, the video calling maximized view now operates as a separate window. As a result, the previous workaround of blocking all key interactions in the parent window is no longer necessary and actually interferes with normal usage.

#### Changes

- **Deleted the entire `useEffect` block** responsible for handling and preventing default actions on `keydown` events.

This change restores normal keyboard interaction in the parent window while the PiP window is open, ensuring a better user experience.

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import React, {useState, useEffect} from 'react';
+import React, {useState} from 'react';
 
 import {DefaultConversationRoleName} from '@wireapp/api-client/lib/conversation/';
 import {TabIndex} from '@wireapp/react-ui-kit/lib/types/enums';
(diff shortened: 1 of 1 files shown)


V-Gira (approved):

aweiss-dev (approved):

e-maad (approved):
LGTM 🚀

PR #17883, opened 2024-08-09
fix: handle leaving a 1:1 call over SFT [WPB-7151]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-7151" title="WPB-7151" target="_blank">StoryWPB-7151</a>  [Web] Add flag for customer specific 1:1 calls in federated environment
  </td></table>
  <br />
 

## Description
see https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1314750477/2024-07-29+1+1+calls+over+SFT#Do-not-leave-the-subconversation

There's currently 2 issues with having 1:1 calls over SFT:
- SFT calls are treated like group calls, so one user would stay in the call if the second participants hangs up (the call should terminate in that case if 1:1)
- Participants were atempting to leave the MLS subconversation, which may cause issue with users from different federated backends.

### Solution (Temporary)
- The remaining participant leaves the call if audio is no longer established from the other participant
- Both participant attempt to delete the subconversation when the call is terminated see https://github.com/wireapp/wire-web-packages/pull/6434

## Checklist

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -985,6 +985,14 @@ export class CallingRepository {
     return this.conversationState.findConversation(conversationId);
   }
 
+  private readonly leave1on1MLSConference = async (conversationId: QualifiedId) => {
+    await this.subconversationService.leave1on1ConferenceSubconversation(conversationId);
+
+    const conversationIdStr = this.serializeQualifiedId(conversationId);
+    this.wCall?.end(this.wUser, conversationIdStr);
+    callingSubscriptions.removeCall(conversationId);
+  };
+
   private readonly leaveMLSConference = async (conversationId: QualifiedId) => {
     await this.subconversationService.leaveConferenceSubconversation(conversationId);
     callingSubscriptions.removeCall(conversationId);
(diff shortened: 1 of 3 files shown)


src/script/calling/CallingRepository.ts
```diff
@@ -1666,13 +1679,43 @@ export class CallingRepository {
...
+
+    if (!currentOtherParticipant) {
+      return;
+    }
+
+    const isCurrentlyEstablished = currentOtherParticipant.isAudioEstablished();
+    const {aestab: newEstablishedStatus} = nextOtherMember;
+
+    if (isCurrentlyEstablished && newEstablishedStatus === AUDIO_STATE.CONNECTING) {
+      void this.leave1on1MLSConference(conversationId);
```
V-Gira, line 1723: `AUDIO_STATE.CONNECTING` after it was `AUDIO_STATE.ESTABLISHED` means we no longer receive audio from the other participant, which happens when they leave

src/script/calling/CallingRepository.ts
```diff
@@ -1666,13 +1679,43 @@ export class CallingRepository {
...
   };
 
+  private readonly handleOneToOneMlsCallParticipantLeave = (
+    conversationId: QualifiedId,
+    members: QualifiedWcallMember[],
+  ) => {
+    const conversation = this.getConversationById(conversationId);
+    const call = this.findCall(conversationId);
+
+    const [, nextOtherMember] = members;
```
e-maad: Are we sure that the next member will always be the other person who left the call?
I am not sure how members are placed in array, maybe the first one will be the one who started the call.
Is it not better if we find the other member using current userId, for example:
`const nextOtherMember = members.find(member => member.id !== -currentUserId-)`

aweiss-dev: I agree with Immad. Even if the order is guaranteed right now, there is no guarantee that someone might change some code and, as a side effect, breaks this part. 
Searching for the right value should be a better approach.

V-Gira: Yes, seems we overlooked this, we check that the id hasn't changed, but not that it's the expected one in the first place :sweat_smile: 
Let me take a look

e-maad (approved):
LGTM 🚀

PR #17905, opened 2024-08-16
fix: revert enumerate media device label for firefox iii [WPB-9688] 

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9688" title="WPB-9688" target="_blank">BugWPB-9688</a>  [web] In Firefox the device label are not read from getUserMedia. This means that the menus are empty.
  </td></table>
  <br />
 

This reverts commit 84b89bdabc9ea27f85178d8be131dcdb85b7b3c4.

## Description

This PR brings back the fix for Firefox labels. Please compare it with the previous one that was reverted:
https://github.com/wireapp/wire-webapp/pull/17757

Please check this ticket as well to understand the reasons for the need for this PR.:

<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10545" title="WPB-10545" target="_blank">BugWPB-10545</a>  [AVS-WEB] Improvements of the concept of Device List.
  </td></table>

## Checklist

Since the changes are relatively large, I'll leave comments and create it as a draft for now.

And this solution is not perfect because the new picture-in-picture mode hides the permission pop-up.

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -1197,12 +1197,8 @@ export class CallingRepository {
     return this.mediaStreamHandler.requestMediaStream(audio, camera, screen, isGroup).then(stream => {
       return this.mediaDevicesHandler
         .initializeMediaDevices(camera)
-        .then(() => {
-          return stream;
-        })
-        .catch(() => {
-          return stream;
-        });
+        .then(() => stream)
+        .catch(() => stream);
     });
   }
 
--- a/src/script/calling/CallingRepository.ts
+++ b/src/script/calling/CallingRepository.ts
@@ -1194,7 +1194,16 @@ export class CallingRepository {
   };
 
   private getMediaStream({audio = false, camera = false, screen = false}: MediaStreamQuery, isGroup: boolean) {
-    return this.mediaStreamHandler.requestMediaStream(audio, camera, screen, isGroup);
+    return this.mediaStreamHandler.requestMediaStream(audio, camera, screen, isGroup).then(stream => {
+      return this.mediaDevicesHandler
+        .initializeMediaDevices(camera)
+        .then(() => {
+          return stream;
+        })
+        .catch(() => {
+          return stream;
+        });
+    });
   }
 
   private handleMediaStreamError(call: Call, requestedStreams: MediaStreamQuery, error: Error | unknown): void {
--- a/src/script/media/MediaDevicesHandler.ts
+++ b/src/script/media/MediaDevicesHandler.ts
@@ -365,7 +365,7 @@ export class MediaDevicesHandler {
    * permissionm
    * @returns All enumerated MediaDeviceInfos
    */
-  private async enumerateMediaDevices(forceCamera: boolean = false): Promise<MediaDeviceInfo[]> {
+  private async enumerateMediaDevices(forceCamera = false): Promise<MediaDeviceInfo[]> {
     // The browser will set the device labels obtained from navigator.mediaDevices.enumerateDevices() to blank strings
     // if there is no longer an active MediaStream, even if the application was previously temporarily authorized to
     // access the devices by calling navigator.mediaDevices.getUserMedia().
--- a/src/script/media/MediaDevicesHandler.ts
+++ b/src/script/media/MediaDevicesHandler.ts
@@ -289,7 +289,7 @@ export class MediaDevicesHandler {
 
     try {
       this.removeAllDevices();
-      const mediaDevices = await this.enumerateMediaDevices(camera);
+      const mediaDevices = await window.navigator.mediaDevices.enumerateDevices();
 
       if (!mediaDevices) {
         throw new Error('No media devices found');
@@ -357,34 +357,6 @@ export class MediaDevicesHandler {
     this.logger.info(`Detected '${screenSources.length}' sources for screen sharing from Electron`);
     this.availableDevices.screeninput(screenSources);
     return screenSources;
-  }
-
-  /**
-   * Enumerates all known Media Devices Infos of a System.
-   * @param forceCamera If `forceCamera=true`, then an audio and video track is created to enforce the device access
-   * permissionm
-   * @returns All enumerated MediaDeviceInfos
-   */
-  private async enumerateMediaDevices(forceCamera = false): Promise<MediaDeviceInfo[]> {
-    // The browser will set the device labels obtained from navigator.mediaDevices.enumerateDevices() to blank strings
-    // if there is no longer an active MediaStream, even if the application was previously temporarily authorized to
-    // access the devices by calling navigator.mediaDevices.getUserMedia().
-
-    // @TODO: Fix Issue
-    // You can use this method to generate a camera stream even in the audio case. This way, the device labels can be displayed in any case.
-    // const constraints = (forceCamera)? {video: true} : {audio: true};
-    // const mediaPromise = navigator.mediaDevices.getUserMedia(constraints)
-    // return mediaPromise.then(stream => {
-    //   if (!stream) {
-    //     return [];
-    //   }
-    //   return window.navigator.mediaDevices.enumerateDevices().then( devices  => {
-    //     stream.getTracks().forEach((t) => t.stop());
-    //     return devices;
-    //   });
-    // });
-
-    return window.navigator.mediaDevices.enumerateDevices();
   }
 
   /**
--- a/src/script/media/MediaDevicesHandler.ts
+++ b/src/script/media/MediaDevicesHandler.ts
@@ -172,7 +172,7 @@ export class MediaDevicesHandler {
    * @camera: boolean, Only when the camera is queried can the entire device list be accessed.
    * @return Promise<void>
    */
-  public async initializeMediaDevices(camera: boolean = false): Promise<void> {
+  public async initializeMediaDevices(camera = false): Promise<void> {
     if (Runtime.isSupportingUserMedia() && !this.devicesAreInit) {
       return this.refreshMediaDevices(camera).then(() => {
         this.subscribeToObservables();
--- a/src/script/media/MediaDevicesHandler.ts
+++ b/src/script/media/MediaDevicesHandler.ts
@@ -267,7 +267,7 @@ export class MediaDevicesHandler {
    * This ensures that the video device labels can also be read. This is necessary for initializing the entire device list.
    * @returns Resolves with all MediaDevices when the list has been updated
    */
-  public async refreshMediaDevices(camera: boolean = false): Promise<MediaDeviceInfo[]> {
+  public async refreshMediaDevices(camera = false): Promise<MediaDeviceInfo[]> {
     const setDevices = (type: MediaDeviceType, devices: MediaDeviceInfo[]): void => {
       this.availableDevices[type](devices);
       const currentId = this.currentDeviceId[type];
--- a/src/script/media/MediaDevicesHandler.ts
+++ b/src/script/media/MediaDevicesHandler.ts
@@ -80,6 +80,7 @@ export class MediaDevicesHandler {
   public deviceSupport: DeviceSupport;
   private previousDeviceSupport: PreviousDeviceSupport;
   private onMediaDevicesRefresh?: () => void;
+  private devicesAreInit = false;
 
   static get CONFIG() {
     return {
@@ -154,7 +155,12 @@ export class MediaDevicesHandler {
       videoinput: this.availableDevices.videoinput().length,
     };
 
-    this.initializeMediaDevices();
+    // The device list must be queried once to obtain the device IDs. This way, the frontend knows whether cameras
+    // and microphones exist and can request them specifically during a call. Additionally, the app needs to know
+    // the device IDs; otherwise, we will receive a Media-Query-Constraint error when querying the devices, as we
+    // explicitly query by the device ID.
+    // The false parameter ensures that we only load the list temporarily.
+    this.initializeMediaDevices(false);
   }
 
   public setOnMediaDevicesRefreshHandler(handler: () => void): void {
@@ -163,14 +169,22 @@ export class MediaDevicesHandler {
 
   /**
    * Initialize the list of MediaDevices and subscriptions.
+   * @camera: boolean, Only when the camera is queried can the entire device list be accessed.
+   * @return Promise<void>
    */
-  private initializeMediaDevices(): void {
-    if (Runtime.isSupportingUserMedia()) {
-      this.refreshMediaDevices().then(() => {
+  public async initializeMediaDevices(camera: boolean = false): Promise<void> {
+    if (Runtime.isSupportingUserMedia() && !this.devicesAreInit) {
+      return this.refreshMediaDevices(camera).then(() => {
         this.subscribeToObservables();
         this.subscribeToDevices();
+        // The web browser cannot access the device labels without a camera stream.
+        // The device list is only complete when the camera was initialized.
+        if (camera) {
+          this.devicesAreInit = true;
+        }
       });
     }
+    return Promise.resolve();
   }
 
   /**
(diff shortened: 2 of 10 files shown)


src/script/calling/CallingRepository.ts
```diff
@@ -1194,7 +1194,16 @@ export class CallingRepository {
...
+      return this.mediaDevicesHandler
+        .initializeMediaDevices(camera)
+        .then(() => {
+          return stream;
+        })
+        .catch(() => {
+          return stream;
+        });
+    });
   }
```
EnricoSchw, line 1203: I've modified the device list so that it is no longer initialized when the app starts. However, this makes it necessary to create it either before a call or when opening the device selection. In these two situations, a media stream will be created.

src/script/media/MediaDevicesHandler.ts
```diff
@@ -154,7 +155,12 @@ export class MediaDevicesHandler {
       videoinput: this.availableDevices.videoinput().length,
     };
 
-    this.initializeMediaDevices();
+    // The device list must be queried once to obtain the device IDs. This way, the frontend knows whether cameras
+    // and microphones exist and can request them specifically during a call. Additionally, the app needs to know
+    // the device IDs; otherwise, we will receive a Media-Query-Constraint error when querying the devices, as we
+    // explicitly query by the device ID.
+    // The false parameter ensures that we only load the list temporarily.
+    this.initializeMediaDevices(false);
```
EnricoSchw, line 163: To maintain the old behavior and because we need the device IDs, I will temporarily initialize the device list at startup.

EnricoSchw: False' essentially means that no camera will be started. Without a video track, Firefox doesn't have permissions for the labels. Therefore, the device list will not be saved as final!

src/script/media/MediaDevicesHandler.ts
```diff
@@ -163,14 +169,22 @@ export class MediaDevicesHandler {
...
+    if (Runtime.isSupportingUserMedia() && !this.devicesAreInit) {
+      return this.refreshMediaDevices(camera).then(() => {
         this.subscribeToObservables();
         this.subscribeToDevices();
+        // The web browser cannot access the device labels without a camera stream.
+        // The device list is only complete when the camera was initialized.
+        if (camera) {
+          this.devicesAreInit = true;
+        }
       });
```
EnricoSchw, line 185: Only when the device list is initialized with a camera stream will it be saved as complete.

src/script/media/MediaDevicesHandler.ts
```diff
@@ -343,6 +359,34 @@ export class MediaDevicesHandler {
...
+    //     return [];
+    //   }
+    //   return window.navigator.mediaDevices.enumerateDevices().then( devices  => {
+    //     stream.getTracks().forEach((t) => t.stop());
+    //     return devices;
+    //   });
+    // });
+
+    return window.navigator.mediaDevices.enumerateDevices();
+  }
```
EnricoSchw: This whole method is how I believe it should be done. However, this would also start a camera for an audio call, or on startup the app. I’ve kept the code in place so we can find a better solution.

EnricoSchw: If a camera is started when launching the app or elsewhere, the user will be prompted for permissions. This can be inconvenient if the user does not intend to make a call.

EnricoSchw: To solve this the device list is now created only when a call is about to be made or when the device selection is opened.

EnricoSchw: To avoid any conflicts with the old behavior, I never start my own stream. I always use only the currently active streams.

tlebon:
Question: is this ready to be reviewed? @EnricoSchw

EnricoSchw:
> Question: is this ready to be reviewed? @EnricoSchw

Its ready 👍

src/script/media/MediaStreamErrorTypes.ts
```diff
@@ -29,3 +29,17 @@ export const MEDIA_STREAM_ERROR_TYPES = {
   ],
   PERMISSION: [MEDIA_STREAM_ERROR.NOT_ALLOWED_ERROR, MEDIA_STREAM_ERROR.SECURITY_ERROR],
 };
+
+export function isMediaStreamDeviceError(errorStr: string): boolean {
+  return MEDIA_STREAM_ERROR_TYPES.DEVICE.filter(errorType => errorType == errorStr).length > 0;
```
przemvs: Result will be the same with this fn from suggestion? If yes please use :)

```suggestion
  return MEDIA_STREAM_ERROR_TYPES.DEVICE.includes(errorStr);
```

src/script/media/MediaStreamErrorTypes.ts
```diff
@@ -29,3 +29,17 @@ export const MEDIA_STREAM_ERROR_TYPES = {
...
+}
+
+export function isMediaStreamReadDeviceError(errorStr: string): boolean {
+  return (
+    [
+      MEDIA_STREAM_ERROR.NOT_READABLE_ERROR,
+      MEDIA_STREAM_ERROR.NOT_ALLOWED_ERROR,
+      MEDIA_STREAM_ERROR.NOT_FOUND_ERROR,
+    ].filter(errorType => errorType == errorStr).length > 0
```
przemvs: I think this one will do the same result. So code will be more clean :)

```suggestion
    [
      MEDIA_STREAM_ERROR.NOT_READABLE_ERROR,
      MEDIA_STREAM_ERROR.NOT_ALLOWED_ERROR,
      MEDIA_STREAM_ERROR.NOT_FOUND_ERROR,
    ].includes(errorStr)
```

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
     }
   }
 
+  /**
+   * The method creates a media stream to enforce access rights to the camera and the microphone. If access is not possible, it starts a user pop-up
+   * @param video When `video=true` then the camera is also addressed. In many cases this is not necessary, because one
+   * track is enough to enforce the general permissions.
+   * @returns Promise with active MediaStream
+   */
+  requestMediaSreamAccess(video: boolean): Promise<MediaStream> {
```
przemvs:
```suggestion
  requestMediaStreamAccess(video: boolean): Promise<MediaStream> {
```

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
...
+    const audio = true;
+    const screen = false;
```
przemvs: You don't need additional variables for true & false method. Pass directly boolean value to fn :)

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
...
+    return window.navigator.mediaDevices
+      .getUserMedia({audio, video})
+      .then((mediaStream: MediaStream) => {
+        return mediaStream;
+      })
+      .catch((error: Error) => {
+        if (isMediaStreamReadDeviceError(error.name)) {
+          this.schedulePermissionHint(audio, video, screen);
+        }
+        throw error;
```
przemvs: Try to do early return instead of checking inside. :)

```suggestion
        if (!isMediaStreamReadDeviceError(error.name)) {
          throw error;
        }
        this.schedulePermissionHint(audio, video, screen);
```

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
...
+   * @returns Promise with active MediaStream
+   */
+  requestMediaSreamAccess(video: boolean): Promise<MediaStream> {
+    const audio = true;
+    const screen = false;
+    return window.navigator.mediaDevices
+      .getUserMedia({audio, video})
+      .then((mediaStream: MediaStream) => {
+        return mediaStream;
+      })
```
przemvs: You can do early return ;)

```suggestion
      .then((mediaStream: MediaStream) => mediaStream)
```

src/script/page/MainContent/panels/preferences/AVPreferences.tsx
```diff
@@ -43,22 +48,47 @@ const AVPreferences = ({
...
+    } catch (error) {
+      logger.warn(`Initialization of media devices failed: ${error.message}`, error);
+    } finally {
+      setCheckingPermissions(false);
+    }
+  };
+
+  useEffect(() => {
+    initializeMediaDevices();
+  }, []);
```
przemvs: I'll do hook for it ;) e.g

```
const { isMediaDevicesAreInitialized } = useInitializeMediaDevices(mediaRepository);

// useInitializeMediaDevices

export const useInitializeMediaDevices = (mediaRepository: MediaRepository) => {
  const [isCheckingPermissions, setCheckingPermissions] = useState(false);
  
  const initializeMediaDevices = async () => {
    setCheckingPermissions(true);
    try {
      await streamHandler.requestMediaSreamAccess(true).then(stream => {
        devicesHandler?.initializeMediaDevices().then(() => {
          stream?.getTracks().forEach(track => track.stop());
        });
      });
    } catch (error) {
      logger.warn(`Initialization of media devices failed: ${error.message}`, error);
    } finally {
      setCheckingPermissions(false);
    }
  };

  useEffect(() => {
    initializeMediaDevices();
  }, []);

  return { isMediaDevicesAreInitialized:  isCheckingPermissions }
}
```

And we don't need to have some function on unmount? :)

src/script/calling/CallingRepository.ts
```diff
@@ -1194,7 +1194,16 @@ export class CallingRepository {
...
-    return this.mediaStreamHandler.requestMediaStream(audio, camera, screen, isGroup);
+    return this.mediaStreamHandler.requestMediaStream(audio, camera, screen, isGroup).then(stream => {
+      return this.mediaDevicesHandler
+        .initializeMediaDevices(camera)
+        .then(() => {
+          return stream;
+        })
+        .catch(() => {
+          return stream;
+        });
```
przemvs:
```suggestion
        .then(() => stream)
        .catch(() => stream);
```

src/script/media/MediaDevicesHandler.ts
```diff
@@ -163,14 +169,22 @@ export class MediaDevicesHandler {
 
   /**
    * Initialize the list of MediaDevices and subscriptions.
+   * @camera: boolean, Only when the camera is queried can the entire device list be accessed.
+   * @return Promise<void>
    */
-  private initializeMediaDevices(): void {
-    if (Runtime.isSupportingUserMedia()) {
-      this.refreshMediaDevices().then(() => {
+  public async initializeMediaDevices(camera: boolean = false): Promise<void> {
```
przemvs: You don't have to assign type there, because default value assign type to arg ;)

```suggestion
  public async initializeMediaDevices(camera = false): Promise<void> {
```

src/script/media/MediaDevicesHandler.ts
```diff
@@ -249,9 +263,11 @@ export class MediaDevicesHandler {
 
   /**
    * Update list of available MediaDevices.
+   * @param [camera=false] If `camera=true`, a video track is also created when the device list is read.
+   * This ensures that the video device labels can also be read. This is necessary for initializing the entire device list.
    * @returns Resolves with all MediaDevices when the list has been updated
    */
-  public async refreshMediaDevices(): Promise<MediaDeviceInfo[]> {
+  public async refreshMediaDevices(camera: boolean = false): Promise<MediaDeviceInfo[]> {
```
przemvs: You don't have to assign type there, because default value assign type to arg ;)

```suggestion
  public async refreshMediaDevices(camera = false): Promise<MediaDeviceInfo[]> {
```

src/script/media/MediaDevicesHandler.ts
```diff
@@ -343,6 +359,34 @@ export class MediaDevicesHandler {
     return screenSources;
   }
 
+  /**
+   * Enumerates all known Media Devices Infos of a System.
+   * @param forceCamera If `forceCamera=true`, then an audio and video track is created to enforce the device access
+   * permissionm
+   * @returns All enumerated MediaDeviceInfos
+   */
+  private async enumerateMediaDevices(forceCamera: boolean = false): Promise<MediaDeviceInfo[]> {
```
przemvs: You don't have to assign type there, because default value assign type to arg ;)

```suggestion
  private async enumerateMediaDevices(forceCamera = false): Promise<MediaDeviceInfo[]> {
```

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
...
+      .catch((error: Error) => {
+        if (isMediaStreamReadDeviceError(error.name)) {
+          this.schedulePermissionHint(audio, video, screen);
```
przemvs: Maybe refactor this function: schedulePermissionHint to use object as props, and then You will pass smth like this:

```suggestion
          this.schedulePermissionHint({audio: true, video, screen: false});
```

src/script/media/MediaStreamHandler.ts
```diff
@@ -72,6 +72,28 @@ export class MediaStreamHandler {
...
+   * The method creates a media stream to enforce access rights to the camera and the microphone. If access is not possible, it starts a user pop-up
+   * @param video When `video=true` then the camera is also addressed. In many cases this is not necessary, because one
+   * track is enough to enforce the general permissions.
+   * @returns Promise with active MediaStream
+   */
+  requestMediaSreamAccess(video: boolean): Promise<MediaStream> {
+    const audio = true;
+    const screen = false;
+    return window.navigator.mediaDevices
+      .getUserMedia({audio, video})
```
przemvs:
```suggestion
      .getUserMedia({audio: true, video})
```

src/script/page/MainContent/panels/preferences/AVPreferences.tsx
EnricoSchw: I think I need some help here. I'm not an Reakt developer. 

I'm not really happy with the bahavior either.. I start a video stream here to enforce the permissions and block the UI for it

EnricoSchw: @przemvs I built a hook like you described. Could you please check if everything is OK.

No unmount  func because no state needs to be reset. Because ``devicesHandler: MediaDevicesHandler``  is a Singleton and uses the local storage to check its state.

And in the current implementation, the permissions are always checked when open the settings.. that's ok.

src/script/media/MediaDevicesHandler.ts
```diff
@@ -343,6 +359,34 @@ export class MediaDevicesHandler {
     return screenSources;
   }
 
+  /**
+   * Enumerates all known Media Devices Infos of a System.
+   * @param forceCamera If `forceCamera=true`, then an audio and video track is created to enforce the device access
+   * permissionm
+   * @returns All enumerated MediaDeviceInfos
+   */
+  private async enumerateMediaDevices(forceCamera = false): Promise<MediaDeviceInfo[]> {
```
przemvs: This param is unnecessary there ;)

przemvs (approved):

src/script/media/MediaDevicesHandler.ts
EnricoSchw: Actually the whole method is unnecessary. I can remove them if you want.

I would have liked to implement it like this.. I cannot use this method because otherwise we would create additional popups to ask user for permissions. So currently I check the permissions:

1. if a call is active I use the running stream or
2. I use the method .. to create a stream in the device selection, 

```typescript
streamHandler.requestMediaSreamAccess(true)
```

przemvs: If we don't have to use this whole method You can comment/remove whole fragment of code. And use method directly ;)

EnricoSchw: I'll remove it

PR #17933, opened 2024-08-23
fix(countly): collect data for non team users, start countly in offline mode

aweiss-dev:
## Description

To track apm data we need to collect them early on.

The solution is to start countly in offline-mode before our application script runs. 
This leads to data being collected locally (NOT BEING SENT TO THE SERVER). 
And as soon as the users consent is read, we switch the countly userId, turn off offline-mode and migrate the previously collected data.

Also removes the block for collecting data of non-team members.

### DIFF
--- a/server/config/server.config.ts
+++ b/server/config/server.config.ts
@@ -42,7 +42,7 @@ const defaultCSP = {
   imgSrc: ["'self'", 'blob:', 'data:', 'https://*.giphy.com'],
   manifestSrc: ["'self'"],
   mediaSrc: ["'self'", 'blob:', 'data:'],
-  scriptSrc: ["'self'", "'unsafe-eval'"],
+  scriptSrc: ["'self'", "'unsafe-eval', 'nonce-countly'"],
   styleSrc: ["'self'", "'unsafe-inline'"],
   workerSrc: ["'self'", 'blob:'],
 };
--- a/server/config/server.config.ts
+++ b/server/config/server.config.ts
@@ -94,7 +94,7 @@ function mergedCSP(
     manifestSrc: [...defaultCSP.manifestSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_MANIFEST_SRC)],
     mediaSrc: [...defaultCSP.mediaSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_MEDIA_SRC)],
     objectSrc: objectSrc.length > 0 ? objectSrc : ["'none'"],
-    scriptSrc: [...defaultCSP.scriptSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_SCRIPT_SRC), `${countlyCSPNonce}`],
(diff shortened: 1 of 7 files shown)


src/page/auth.ejs
```diff
@@ -25,6 +25,16 @@
     <script src='../libs/countly/boomerang.min.js?v=<%= VERSION %>' type='text/javascript'></script>
     <script src='../libs/countly/countly.min.js?v=<%= VERSION %>' type='text/javascript'></script>
     <script src='../libs/countly/countly_boomerang.js?v=<%= VERSION %>' type='text/javascript'></script>
+    <script type='text/javascript' nonce="countly">
+      Countly.init({
+        debug:false,
+        app_key:"<%= COUNTLY_API_KEY %>",
+        url: "https://countly.wire.com/",
```
przemvs, line 32: Can we store also url in envs ?

src/page/index.ejs
```diff
@@ -25,6 +25,16 @@
     <script src='./libs/countly/boomerang.min.js?v=<%= VERSION %>' type='text/javascript'></script>
     <script src='./libs/countly/countly.min.js?v=<%= VERSION %>' type='text/javascript'></script>
     <script src='./libs/countly/countly_boomerang.js?v=<%= VERSION %>' type='text/javascript'></script>
+    <script type='text/javascript' nonce="countly">
+      Countly.init({
+        debug:false,
+        app_key:"<%= COUNTLY_API_KEY %>",
+        url: "https://countly.wire.com/",
```
przemvs, line 32: Can we store also url in envs ?

przemvs (approved):
2 things, LGTM! :)

src/page/index.ejs
aweiss-dev: I dont think it is necessary. But we could

server/config/server.config.ts
```diff
@@ -79,7 +94,7 @@ function mergedCSP({urls}: ConfigGeneratorParams, env: Record<string, string>):
     manifestSrc: [...defaultCSP.manifestSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_MANIFEST_SRC)],
     mediaSrc: [...defaultCSP.mediaSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_MEDIA_SRC)],
     objectSrc: objectSrc.length > 0 ? objectSrc : ["'none'"],
-    scriptSrc: [...defaultCSP.scriptSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_SCRIPT_SRC)],
+    scriptSrc: [...defaultCSP.scriptSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_SCRIPT_SRC), `${countlyCSPNonce}`],
```
aweiss-dev:
```suggestion
    scriptSrc: [...defaultCSP.scriptSrc, ...parseCommaSeparatedList(env.CSP_EXTRA_SCRIPT_SRC), countlyCSPNonce],
```

PR #17952, opened 2024-08-26
fix: remove dynamically generated nonce

aweiss-dev:
## Description

Dynamically generating the nonce causes the whole application to crash since our index.html is statically generated during build time and the server generates a new nonce at server start time, which causes them to be out of sync. 

If we want to use dynamic nonces, the server should not use a static index.html and replace the html variables at the server start.


package.json
```diff
@@ -6,6 +6,7 @@
     "@lexical/history": "0.17.0",
     "@lexical/react": "0.17.0",
     "@mediapipe/tasks-vision": "0.10.14",
+    "@types/uuid": "10.0.0",
```
V-Gira: moved from dev deps to deps? :thinking:

V-Gira (approved):

PR #17956, opened 2024-08-27
fix: handled screen share on popup window [WPB-10672]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10672" title="WPB-10672" target="_blank">BugWPB-10672</a>  [Web] Pop out call window covers browser's screen selection screen
  </td></table>
  <br />
 

## Description
As a user, when we try to share the screen from popup window, the screen sharing options gets hidden behind the popup and we have to manually focus the main window and select options there.

Technical limitation:
It is impossible to focus back on main window from popup window.

Solution:
The best possible solution is to minimize the call view when the screen sharing initiated from popup window.
For better UX, it is better not to move focus back and forth between windows.

https://github.com/user-attachments/assets/efe57355-bc06-4fc1-9090-3983311d8df8

## Checklist

- use (x) data
- be sure to look at ...

### DIFF
--- a/src/script/components/calling/CallingOverlayContainer.tsx
+++ b/src/script/components/calling/CallingOverlayContainer.tsx
@@ -52,13 +52,16 @@ const CallingContainer: React.FC<CallingContainerProps> = ({
   toggleScreenshare,
 }) => {
   const {devicesHandler: mediaDevicesHandler} = mediaRepository;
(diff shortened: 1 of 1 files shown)


tlebon (approved):
looking good

PR #17959, opened 2024-08-28
fix(countly): remove user ids from URL traces [WPB-8918]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8918" title="WPB-8918" target="_blank">EpicWPB-8918</a>  [Web] Performance Analytics
  </td></table>
  <br />
 

## Description

Performance metrics sent customer private data in the URL to the countly servers before. Now the ids will be replaced.

## Checklist

### DIFF
--- a/src/page/countlyBoomerangCustom.js
+++ b/src/page/countlyBoomerangCustom.js
@@ -1,3 +1,27 @@
+/*
+ * Wire
+ * Copyright (C) 2024 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


src/page/countlyBoomerangCustom.js
```diff
@@ -0,0 +1,186 @@
+'use strict';
```
tlebon, line 25: do we need our copy info here?

aweiss-dev: I added it, just in case

thisisamir98 (approved):

PR #17969, opened 2024-08-29
feat: Call quality feedback

przemvs:
https://wearezeta.atlassian.net/browse/WPB-10379

## Description

We want to ask users about call Quality Feedback. So we displaying Quality Feedback Modal while call is ended every 3 days, until user set don't show Quality Feedback Modal checkbox to true.

After end call Quality Feedback Modal is displayed, user sending rate from 1-5 or can skip Quality Feedback.
Date about time to next Feedback is stored in Local Storage. So if user join with another device Quality Feedback Modal will be displayed again on end of call. We displaying this Modal while Countly is activated.

image

## Checklist

- use (x) data
- be sure to look at ...


src/script/calling/CallingRepository.ts
```diff
@@ -1170,6 +1172,24 @@ export class CallingRepository {
     const conversationIdStr = this.serializeQualifiedId(conversationId);
     delete this.poorCallQualityUsers[conversationIdStr];
     this.wCall?.end(this.wUser, conversationIdStr);
+
+    if (this.selfUser) {
```
thisisamir98: maybe use early return here 🤔

src/script/components/calling/useCallAlertState.ts
```diff
@@ -24,11 +24,19 @@ type CallAlertState = {
...
   clearShowAlert: () => void;
+  qualityFeedBackModalShown: boolean;
+  toggleQualityFeedbackModal: (isVisible: boolean) => void;
 };
 
 const useCallAlertState = create<CallAlertState>((set, get) => ({
   showAlert: false,
   isGroupCall: false,
+  qualityFeedBackModalShown: false,
+  toggleQualityFeedbackModal: isVisible =>
```
thisisamir98: Suggestion: what this method is doing is not really toggling but actually setting the value, I think `setQualityFeedbackModalShown` might be a better name.

src/script/calling/CallingRepository.ts
przemvs: Ahh.. I accidentally removed check for if countly is enabled. So I don't make a early return here ;)

src/script/calling/CallingRepository.ts
```diff
@@ -1165,11 +1168,37 @@ export class CallingRepository {
...
+
+    const {setQualityFeedbackModalShown} = useCallAlertState.getState();
+
+    try {
+      const qualityFeedbackStorage = localStorage.getItem(CALL_QUALITY_FEEDBACK_KEY);
+      const currentStorageData = qualityFeedbackStorage ? JSON.parse(qualityFeedbackStorage) : {};
+      const currentUserDate = currentStorageData?.[this.selfUser.id];
+      const currentDate = new Date().toISOString();
+
+      if (typeof currentUserDate === 'undefined' || (currentUserDate !== null && currentDate >= currentUserDate)) {
```
e-maad: Do we need typeof here?
```suggestion
      if (currentUserDate === undefined || (currentUserDate !== null && currentDate >= currentUserDate)) {
```

src/script/calling/CallingRepository.ts
```diff
@@ -1165,11 +1168,37 @@ export class CallingRepository {
     this.wCall?.requestVideoStreams(this.wUser, convId, VSTREAMS.LIST, JSON.stringify(payload));
   }
 
+  readonly showCallQualityFeedbackModal = (): void => {
```
e-maad:
```suggestion
  readonly showCallQualityFeedbackModal = () => {
```

src/script/calling/CallingRepository.ts
e-maad: `currentDate >= currentUserDate`
We are comparing two strings here, would be better if we compare date or unicode timestamps.

src/script/calling/CallingRepository.ts
```diff
@@ -1165,11 +1168,37 @@ export class CallingRepository {
...
+      const currentStorageData = qualityFeedbackStorage ? JSON.parse(qualityFeedbackStorage) : {};
+      const currentUserDate = currentStorageData?.[this.selfUser.id];
+      const currentDate = new Date().toISOString();
+
+      if (typeof currentUserDate === 'undefined' || (currentUserDate !== null && currentDate >= currentUserDate)) {
+        setQualityFeedbackModalShown(true);
+      }
+    } catch (error) {
+      this.logger.warn(`Storage data can't found: ${(error as Error).message}`);
+      setQualityFeedbackModalShown(true);
```
e-maad, line 1189: We are showing 'Quality Feedback Modal' for both cases. Is it okay?
If yes then it would be nicer if we remove it from `try` and` catch` block and move it to `finally` block (only once).

src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+  it('should not render if qualityFeedBackModalShown is false', () => {
+    const {setQualityFeedbackModalShown} = useCallAlertState.getState();
+    setQualityFeedbackModalShown(false);
+
+    const {container} = render(<QualityFeedbackModal />);
+    expect(container.firstChild).toBeNull();
+  });
+
+  it('should render correctly when qualityFeedBackModalShown is true', () => {
+    render(withTheme(withIntl(<QualityFeedbackModal />)));
```
e-maad: Minor suggestion, would be nice if we create custom render function, like this:
`const renderQualityFeedbackModal = () => render(withTheme(withIntl(<QualityFeedbackModal />)))`
and use it everywhere. Totally your call.

src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+import {Segmentation} from '../../../tracking/Segmentation';
+import {UserState} from '../../../user/UserState';
+
+const logger = getLogger('CallQualityFeedback');
+
+interface QualityFeedbackModalProps {
+  userState?: UserState;
+}
+
+export const QualityFeedbackModal = ({userState = container.resolve(UserState)}: QualityFeedbackModalProps) => {
```
e-maad: I think it is a bit misleading to accept an optional param which we are not passing anywhere.
```suggestion
export const QualityFeedbackModal = () => {
  const userState = container.resolve(UserState);
```

src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const [isChecked, setIsChecked] = useState(false);
+  const {setQualityFeedbackModalShown, qualityFeedBackModalShown} = useCallAlertState();
```
e-maad: Typo:
```suggestion
  const {setQualityFeedbackModalShown, qualityFeedbackModalShown} = useCallAlertState();
```
Maybe we have to update the callee too.

src/script/components/calling/useCallAlertState.ts
```diff
@@ -24,11 +24,19 @@ type CallAlertState = {
   isGroupCall: boolean;
   showStartedCallAlert: (isGroupCall: boolean, isVideoCall?: boolean) => void;
   clearShowAlert: () => void;
+  qualityFeedBackModalShown: boolean;
+  setQualityFeedbackModalShown: (isVisible: boolean) => void;
```
e-maad, line 28: I would suggest a better name if you like:
`qualityFeedbackModalVisibility` and `setQualityFeedbackModalVisibility`

e-maad (commented):
Amazing work 🚀 
Added few suggestions if you like to address.

src/script/calling/CallingRepository.ts
przemvs: We can't add it to finally, because if there is currentDate >= savedDate, we want to show it and if there will be no local storage ;)

src/script/components/calling/useCallAlertState.ts
przemvs: I'll go with qualityFeedbackModalShown - sounds like a boolean, Your proposal sounds like expects string for me :D

aweiss-dev: A Boolean for me should begin with isXxx, hasXxx or something like that.

thisisamir98 (approved):

src/script/calling/CallingRepository.ts
e-maad: You are right 👍

e-maad (approved):
Great job 🚀

PR #17970, opened 2024-08-29
feat: Use new telemetry key (WPB-9972)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-9972" title="WPB-9972" target="_blank">StoryWPB-9972</a>  [Web] data collection for free users 
  </td></table>
  <br />
 

## Description
This PR integrates the new telemetry key name implemented in https://github.com/wireapp/wire-web-packages/pull/6485

With this PR we also fix the issue that when user clicks on disagree on the data sharing consent modal and then goes to preferences, the corresponding checkbox for it is checked even though user has clicked on disagree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@mediapipe/tasks-vision": "0.10.14",
     "@wireapp/avs": "9.9.3",
     "@wireapp/commons": "5.2.10",
-    "@wireapp/core": "46.3.1",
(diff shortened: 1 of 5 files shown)


aweiss-dev (approved):

e-maad (approved):
Great 🚀

PR #17974, opened 2024-08-30
chore: update docker file for new yarn version

aweiss-dev:
The way we use yarn in the docker file was outdated (yarn v1 syntax, 5y old).

I updated the command to match the 2024 version of yarn.


tlebon (approved):
please add some details to the PR

PR #18003, opened 2024-09-04
fix: dont track apm data when user disabled tracking.

aweiss-dev:
## Description

We had multiple problems:

1. Countly didn't recognise the telemetry-sharing change anymore (didn't activate or deactivate after first init)
2. Countly sent APM reports even if the user disabled tracking

## Fixes

1. Added a listener to react to telemetry-sharing changes
2. Added the consent feature and removed consent for all Countly features as soon as someone disables tracking

## Additional Work
Changed to async Countly implementation since it is the recommended way by the API


src/page/auth.ejs
```diff
@@ -26,15 +26,22 @@
...
     <script type='text/javascript' nonce="<%= COUNTLY_NONCE %>">
-      Countly.init({
-        debug: false,
-        app_key:"<%= COUNTLY_API_KEY %>",
-        url: "https://countly.wire.com/",
-        offline_mode: true,
-        use_session_cookie: false,
-        storage: 'localstorage'
-      });
-      Countly.track_performance();
```
tlebon, line 37: can you explain why this change? just curious

src/script/tracking/countly-skd-web.d.ts
```diff
@@ -41,6 +41,8 @@ export interface CountlyEvent {
 }
 
 export interface Countly {
+  q: any[];
```
tlebon, line 53: we dont have any expectations about this array?

aweiss-dev: Countly has no types. We could implement all of them by ourselves, but this is a waste of time for me, since the library can change any time and our types will be outdated without our knowledge. 
Its better to open the countly documentation when working on it.

src/page/auth.ejs
aweiss-dev: While parsing through the documentation I saw that they recommend the asynchronous implementation with the q - array. The synchronous one is optional but inferior (resource blocking).
Since we can not import the SDK directly anymore, because of the performance plugin integration, our use case for a synchronous implementation is out of the window. 
Therefore it is a good call to change how we use countly to the recommended way.

src/script/tracking/countly-skd-web.d.ts
tlebon: maybe theres a link you could insert here then? to make it easier for the next person

tlebon: as a code comment.

aweiss-dev: Oh, for sure. Good call

tlebon (approved):
I'm happy.

PR #18036, opened 2024-09-11
feat: Add config toolbar

thisisamir98:
## Description

This pull request introduces a new configuration tool for internal use in our web application. The tool allows users to dynamically toggle and update configuration variables that come from environment variables. Key features include:

- A fixed-position panel that can be toggled on and off using the hotkey `Cmd + Shift + 2` (Mac) or `Ctrl + Shift + 2` (Windows/Linux).
- A form that displays all configuration variables, allowing users to update them directly.
- Improved sorting to prioritize the `FEATURE` key at the top of the list.

### Changes

1. **Added `ConfigTool` Component**:
   - A React component that renders a configuration panel.
   - Supports dynamic rendering of nested configuration objects.
   - Allows users to update boolean values, arrays, and string values.

2. **Sorting**:
   - The configuration list starts with the `FEATURE` key if it exists, followed by other configuration keys.

3. **Caution Message**:
   - Added a warning message at the top of the configuration panel to inform users about the potential impact of modifying settings.

https://github.com/user-attachments/assets/5bb8095d-6bd2-4821-8da6-e51bab62dab9

### DIFF
--- a/src/script/Config.ts
+++ b/src/script/Config.ts
@@ -33,7 +33,7 @@ export const ACCENT_ID = {
   TURQUOISE: 6,
 };
 
-const config = {
+let config = {
   ...env,
   APP_INSTANCE_ID: createUuid(),
   FEATURE: {
@@ -94,6 +94,9 @@ export type Configuration = typeof config;
 const Config = {
   getConfig: () => {
     return config;
+  },
+  _dangerouslySetConfigForDebug: (newConfig: Configuration) => {
+    config = newConfig;
   },
   getDesktopConfig: () => {
     if (!Runtime.isDesktopApp) {
--- a/src/script/Config.ts
+++ b/src/script/Config.ts
@@ -33,7 +33,7 @@ export const ACCENT_ID = {
   TURQUOISE: 6,
 };
 
-let config = {
+const config = {
   ...env,
   APP_INSTANCE_ID: createUuid(),
   FEATURE: {
@@ -95,8 +95,8 @@ const Config = {
   getConfig: () => {
     return config;
   },
-  _dangerouslySetConfigForDebug: (newConfig: Configuration) => {
-    config = newConfig;
+  _dangerouslySetConfigFeaturesForDebug: (newConfigFeatures: Configuration['FEATURE']) => {
+    (config.FEATURE as unknown) = newConfigFeatures;
   },
   getDesktopConfig: () => {
     if (!Runtime.isDesktopApp) {
--- a/src/script/components/ConfigToolbar/ConfigToolbar.styles.tsx
+++ b/src/script/components/ConfigToolbar/ConfigToolbar.styles.tsx
@@ -25,7 +25,7 @@ export const wrapperStyles: CSSObject = {
   right: 0,
   width: '400px',
   height: '100vh',
-  backgroundColor: '#f4f4f4',
+  backgroundColor: 'var(--sidebar-bg)',
   padding: '20px',
   overflow: 'auto',
   zIndex: 100000000,
--- a/src/script/components/ConfigToolbar/ConfigToolbar.styles.tsx
+++ b/src/script/components/ConfigToolbar/ConfigToolbar.styles.tsx
@@ -1,0 +1,33 @@
+/*
+ * Wire
+ * Copyright (C) 2022 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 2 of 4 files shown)


V-Gira (commented):
Not entirely sure from looking at the PR, would this only appear when `FEATURE_ENABLE_DEBUG` is set to `true`?

tlebon:
> Not entirely sure from looking at the PR, would this only appear when `FEATURE_ENABLE_DEBUG` is set to `true`?

same question, i think this is missing something to tie it to debug mode?

tlebon:
Otherwise, looks awesome! you whipped this up super fast

thisisamir98:
@tlebon @V-Gira tied to FEATURE_ENABLE_DEBUG in https://github.com/wireapp/wire-webapp/pull/18036/commits/21d25aa2705ce7d30122cb914f8dad1d4c213788

src/script/Config.ts
```diff
@@ -95,6 +95,9 @@ const Config = {
   getConfig: () => {
     return config;
   },
+  _dangerouslySetConfigFeaturesForDebug: (newConfigFeatures: Configuration['FEATURE']) => {
+    (config.FEATURE as unknown) = newConfigFeatures;
```
przemvs, line 99: Hmm... can't be as const? 🤔

src/script/components/ConfigToolbar/ConfigToolbar.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export function ConfigToolbar() {
+  const [showConfig, setShowConfig] = useState(false);
+  const [configFeaturesState, setConfigFeaturesState] = useState<Configuration['FEATURE']>(Config.getConfig().FEATURE);
+  const wrapperRef = useRef(null);
+
+  // Toggle config tool on 'cmd/ctrl + shift + 2'
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if ((event.metaKey || event.ctrlKey) && event.shiftKey && event.key === '2') {
```
przemvs: You can use isMetaKey for ctrl/meta key from KeyboardUtil.ts

```suggestion
      if (isMetaKey(event) && event.shiftKey && event.key === '2') {
```

V-Gira: > Nope, Meta key is the Windows/Super key and prone to issue (OS function would have priority)

Ah that's wrong sorry, didn't read the util. To be fair that's a pretty bad name for it :sweat_smile:

przemvs: image

przemvs: @V-Gira - as You can see this is same what Bardia do :D But has additional check keyboardEvent.ley :)

V-Gira: realized later you were not referring to the actual meta key, my bad https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey

src/script/Config.ts
thisisamir98: no because config variable is defined as const and normal config.FEATURE = something ends up with error:
`Cannot assign to 'FEATURE' because it is a read-only property.ts(2540)`

src/script/components/ConfigToolbar/ConfigToolbar.tsx
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/18036/commits/21e07ed284bff271e786eb71e6337f7220907ab6

src/script/components/ConfigToolbar/ConfigToolbar.tsx
```diff
@@ -0,0 +1,140 @@
...
+        top: 0,
+        right: 0,
+        width: '400px',
+        height: '100vh',
+        backgroundColor: '#f4f4f4',
+        padding: '20px',
+        overflow: 'auto',
+        zIndex: 100000000,
+        boxShadow: '3px 1px 6px 1px #000',
+      }}
```
przemvs: Maybe You can move styles to ConfigToolbar.styles.tsx ? 🤔

przemvs (approved):
LGTM!

src/script/components/ConfigToolbar/ConfigToolbar.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/18036/commits/6d6b2a827d50332289c8db8907ead78d19327af1

src/script/components/ConfigToolbar/ConfigToolbar.styles.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import {CSSObject} from '@emotion/react';
+
+export const wrapperStyles: CSSObject = {
+  position: 'fixed',
+  top: 0,
+  right: 0,
+  width: '400px',
+  height: '100vh',
+  backgroundColor: '#f4f4f4',
```
V-Gira: We need a CSS variable here, give me a minute, I'll find the correct one

V-Gira:
```suggestion
  backgroundColor: 'var(--sidebar-bg)',
```

PR #18081, opened 2024-09-23
chore: downgrade mediapipe/tasks-vision to 0.10.15

e-maad:
## Description
Downgrade [@mediapipe/tasks-vision](https://www.npmjs.com/package/@mediapipe/tasks-vision/v/0.10.16) to 0.10.15

The new release (0.10.16) does not contain wasm folder. Until we have new release we will prevent dependabot to update @mediapipe/tasks-vision.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,6 +35,8 @@ updates:
 
     ignore:
       - dependency-name: '@wireapp/avs'
+      # Wait until we have next release with wasm folder
+      - dependency-name: '@mediapipe/tasks-vision'
 
   # Server dependencies
   - package-ecosystem: npm
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,6 +37,8 @@ updates:
       - dependency-name: '@wireapp/avs'
       # Wait until we have next release with wasm folder
       - dependency-name: '@mediapipe/tasks-vision'
+        versions:
+          - '0.10.16'
 
   # Server dependencies
   - package-ecosystem: npm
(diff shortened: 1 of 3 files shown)


przemvs (approved):

.github/dependabot.yml
```diff
@@ -35,6 +35,8 @@ updates:
 
     ignore:
       - dependency-name: '@wireapp/avs'
+      # Wait until we have next release with wasm folder
+      - dependency-name: '@mediapipe/tasks-vision'
```
V-Gira, line 39: maybe we could specify only 0.10.16?
Let me look if it's doable

e-maad: Yes, it's doable. 
The plan is to remove it once we have new release.
As it is not a long term solution, I didn't mention version.

V-Gira:
```suggestion
      - dependency-name: '@mediapipe/tasks-vision'
        versions:
          - '0.10.16'
```

V-Gira: > Yes, it's doable. The plan is to remove it once we have new release. As it is not a long term solution, I didn't mention version.

Kinda my reasoning for specifying the version, it would bump to the next version and either break or not.
Relying on us manually checking a single npm package for update sounds very liable to be forgoten about :sweat_smile:

e-maad: I am not sure if there are multiple ways to do it. As per documentation:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore-updates-for-a-specific-version

V-Gira (approved):

PR #18093, opened 2024-09-25
chore: Replace rest common related less vars with native css vars and sort them

przemvs:
## Description

CSS Variables doesn't work with media queries. :(

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageActions.styles.ts
+++ b/src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageActions.styles.ts
@@ -34,7 +34,7 @@ export const messageBodyActions: CSSObject = {
   right: '-40px',
   top: '-34px',
   userSelect: 'none',
-  '@media (max-width: @screen-md-min)': {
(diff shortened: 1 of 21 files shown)


src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageActions.styles.ts
```diff
@@ -34,7 +34,7 @@ export const messageBodyActions: CSSObject = {
   right: '-40px',
   top: '-34px',
   userSelect: 'none',
-  '@media (max-width: @screen-md-min)': {
+  '@media (max-width: 768px)': {
```
aweiss-dev, line 37: not a fan of the fixed value here

aweiss-dev (approved):

src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageActions.styles.ts
przemvs: We have to find proper solution for media queries, because we can't pass css variables in media queries :(

PR #18097, opened 2024-09-26
fix: Weird transition in group info

przemvs:
https://wearezeta.atlassian.net/browse/WPB-11300

## Description

Fixed transitions after refactor less variables.

## Checklist

### DIFF
--- a/src/style/list/list.less
+++ b/src/style/list/list.less
@@ -250,7 +250,7 @@ body.theme-dark {
   &:hover {
     .right-column > .icon-more {
       opacity: 1;
-      transition-delay: calc(var(--animation-timing-faster) + 0.05);
+      transition-delay: calc(var(--animation-timing-faster) + 0.05s);
       transition-duration: var(--animation-timing-slow);
     }
 
--- a/src/style/panel/panel.less
+++ b/src/style/panel/panel.less
@@ -81,7 +81,7 @@
     &--move-in--right {
       & .panel__content,
       & .panel__footer {
-        transform: translateX(-var(--panel-width));
+        transform: translateX(calc(var(--panel-width) * -1));
       }
     }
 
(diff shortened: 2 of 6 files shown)


src/style/panel/panel.less
```diff
@@ -81,7 +81,7 @@
     &--move-in--right {
       & .panel__content,
       & .panel__footer {
-        transform: translateX(-var(--panel-width));
+        transform: translateX(calc(var(--panel-width) * -1));
```
aweiss-dev, line 84: * -1 is a hack? what does it do? :D

aweiss-dev (approved):

src/style/panel/panel.less
przemvs: `-var(--panel-width))` - this code doesn't work for negative values so we have to calc * -1.

e-maad: Why not create another variable like:
`--nagative-panel-width: -304px;`
Just a suggestion.

src/style/list/list.less
```diff
@@ -250,7 +250,7 @@ body.theme-dark {
   &:hover {
     .right-column > .icon-more {
       opacity: 1;
-      transition-delay: calc(var(--animation-timing-faster) + 0.05);
+      transition-delay: calc(var(--animation-timing-faster) + 0.05s);
```
e-maad, line 253: We have these variables for animation-timing:
```
  --animation-timing-faster: 0.15s;
  --animation-timing-fast: 0.35s;
```

Can we create a new one with `0.20s`?

e-maad (commented):
I would prefer to make new variables instead of using calc.

src/style/list/list.less
przemvs: Ahh... I already merged :( But this is a delay.

PR #18098, opened 2024-09-27
feat: filter remote domain users in conversation creation list [WPB-10792]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10792" title="WPB-10792" target="_blank">StoryWPB-10792</a>  [Web] Do not display remote domain users when Proteus is default protocol of the team
  </td></table>
  <br />
 

## Description

When the conversation type is not MLS, we should not show federated team users in the conversation creation list.
For Proteus and Mixed, both the Proteus protocol is used, therefore the filter applies to these conversation protocol types.

This PR also takes care of the ticket [WPB-10790](https://wearezeta.atlassian.net/browse/WPB-10790)

[WPB-10790]: https://wearezeta.atlassian.net/browse/WPB-10790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -440,6 +440,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
               conversationRepository={conversationRepository}
               noUnderline
               allowRemoteSearch
+              filterRemoteTeamUsers={true}
             />
           </FadingScrollbar>
         )}
--- a/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
+++ b/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
@@ -440,7 +440,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
               conversationRepository={conversationRepository}
               noUnderline
               allowRemoteSearch
-              filterRemoteTeamUsers={true}
+              filterRemoteTeamUsers
             />
           </FadingScrollbar>
         )}
--- a/src/script/components/UserSearchableList.tsx
+++ b/src/script/components/UserSearchableList.tsx
@@ -92,7 +92,8 @@ export const UserSearchableList: React.FC<UserListProps> = ({
 
       // We shouldn't show any members that have the 'external' role and are not already locally known.
       const nonExternalMembers = await teamRepository.filterExternals(uniqueMembers);
-      setRemoteTeamMembers(nonExternalMembers);
+      const nonRemoteDomainMembers = teamRepository.filterRemoteDomainUsers(nonExternalMembers);
+      setRemoteTeamMembers(nonRemoteDomainMembers);
     }, 300),
     [],
   );
@@ -116,13 +117,13 @@ export const UserSearchableList: React.FC<UserListProps> = ({
     }
 
     if (!selfFirst) {
-      setFilteredUsers(results);
+      setFilteredUsers(teamRepository.filterRemoteDomainUsers(results));
       return;
     }
 
     // make sure the self user is the first one in the list
     const [selfUser, otherUsers] = partition(results, user => user.isMe);
-    setFilteredUsers(selfUser.concat(otherUsers));
+    setFilteredUsers(teamRepository.filterRemoteDomainUsers(selfUser.concat(otherUsers)));
   }, [filter, users.length]);
 
   const foundUserEntities = () => {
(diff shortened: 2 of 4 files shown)


przemvs (approved):

V-Gira (commented):
Is this component (`UserSearchableList`) only used during the conversation creation flow?

src/script/team/TeamRepository.ts
```diff
@@ -236,6 +236,21 @@ export class TeamRepository extends TypedEventEmitter<Events> {
     await this.updateTeamMembersByIds(selfTeamId, newTeamMemberIds, true);
   };
 
+  filterRemoteDomainUsers(users: User[]): User[] {
+    const isMLS = this.getTeamSupportedProtocols().includes(ConversationProtocol.MLS);
```
V-Gira: still feels like we should find a way to get the default protocol instead of if MLS is supported.
The tickets read like we should filter fed users if Proteus is the default protocol even if MLS is supported

V-Gira: for reference, that's the check in the group creation modal
```typescript
  //if feature flag is set to false or mls is disabled for current team use proteus as default
  const defaultProtocol =
    isMLSFeatureEnabled && isMLSEnabledForTeam
      ? teamState.teamFeatures()?.mls?.config.defaultProtocol
      : ConversationProtocol.PROTEUS;

File: src/script/components/Modals/GroupCreation/GroupCreationModal.tsx [hunk 2]
```diff
@@ -440,6 +440,7 @@ const GroupCreationModal: React.FC<GroupCreationModalProps> = ({
               conversationRepository={conversationRepository}
               noUnderline
               allowRemoteSearch
+              filterRemoteTeamUsers={true}
```
przemvs:
```suggestion
              filterRemoteTeamUsers
```

src/script/components/UserSearchableList.tsx
```diff
@@ -92,7 +94,11 @@ export const UserSearchableList: React.FC<UserListProps> = ({
 
       // We shouldn't show any members that have the 'external' role and are not already locally known.
       const nonExternalMembers = await teamRepository.filterExternals(uniqueMembers);
-      setRemoteTeamMembers(nonExternalMembers);
+      if (!filterRemoteTeamUsers) {
+        setRemoteTeamMembers(nonExternalMembers);
+        return;
+      }
+      setRemoteTeamMembers(teamRepository.filterRemoteDomainUsers(nonExternalMembers));
```
przemvs: WDYT? :)

```suggestion
      const filteredRemoteTeamMembers = filterRemoteTeamUsers ? teamRepository.filterRemoteDomainUsers(nonExternalMembers) : nonExternalMembers;
      setRemoteTeamMembers(filteredRemoteTeamMembers);
```

src/script/page/RightSidebar/AddParticipants/AddParticipants.tsx
```diff
@@ -248,6 +248,7 @@ const AddParticipants: FC<AddParticipantsProps> = ({
               selfUser={selfUser}
               isSelectable
               allowRemoteSearch
+              filterRemoteTeamUsers={true}
```
przemvs:
```suggestion
              filterRemoteTeamUsers
```

src/script/components/UserSearchableList.tsx
przemvs: Or pass directly ternary operator into setRemoteTeamMembers :)

src/script/team/TeamRepository.ts
```diff
@@ -236,6 +236,22 @@ export class TeamRepository extends TypedEventEmitter<Events> {
     await this.updateTeamMembersByIds(selfTeamId, newTeamMemberIds, true);
   };
 
+  filterRemoteDomainUsers(users: User[]): User[] {
+    const isMLS = this.teamState.teamFeatures()?.mls?.config.defaultProtocol === ConversationProtocol.MLS;
```
przemvs, line 240: Maybe we can add guard for isMLS somewhere in helpers? Because I see we checking in a lot of places isMLS but passing conversation.

WDYT? :)
Smth like this:
`const isMLSProtocol = (protocol: string) => protocol === ConversationProtocol.MLS;`

And inside this fn:
`const isMLS = isMLSProtocol(this.teamState.teamFeatures()?.mls?.config.defaultProtocol);`

aweiss-dev: Why would that be helpful here? The value of the string is typed, it can be accessed directly :)

PR #18099, opened 2024-09-27
fix: Disallow typing special characters when in full screen calling (WPB-10888)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10888" title="WPB-10888" target="_blank">TaskWPB-10888</a>  [Web] Make sending messages possible while in pop out view, but not possible while in fullscreen calling view
  </td></table>
  <br />
 

## Description
- We prevent keyboard events while in full screen call ui, it was however still possible to type special event that require composed events.
- This fixes the issue by preventing propagation as well as prevent default 

## Checklist

### DIFF
--- a/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/src/script/components/calling/FullscreenVideoCall.tsx
@@ -187,6 +187,8 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
       }
 
       event.preventDefault();
+      event.stopPropagation();
+
       preventFocusOutside(event, 'video-calling');
     };
 


src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -187,6 +187,8 @@ const FullscreenVideoCall: React.FC<FullscreenVideoCallProps> = ({
       }
 
       event.preventDefault();
+      event.stopPropagation();
```
V-Gira, line 190: this prevents the `compositionstart` event from being fired.
( `compositionstart`  is an event that is fired when you press a key that waits for a second key press, like accents and diacritics in some layouts)

aweiss-dev (approved):

PR #18163, opened 2024-10-16
fix: add better logging for countly

aweiss-dev:
## Description

We have the Issue that Countly reporting works locally, but not on the internal environment anymore.
This PR enhances our logging ability by a lot, to be able to spot issues easier.


src/script/tracking/EventTrackingRepository.ts
```diff
@@ -226,24 +231,27 @@ export class EventTrackingRepository {
...
       return;
     }
 
     if (!isCountlyEnabledAtCurrentEnvironment() || this.isProductReportingActivated) {
+      this.countlyLogger.warn('Countly is not enabled at this environment');
       return;
     }
     this.isProductReportingActivated = true;
 
-    const {isProduction} = getWebEnvironment();
```
V-Gira, line 237: Maybe we keep this for peace of mind as a condition in addition to the new env var?

src/script/tracking/EventTrackingRepository.ts
```diff
@@ -226,24 +231,27 @@ export class EventTrackingRepository {
...
-
     // Add Parameters to previous Countly object
 
-    window.Countly.app_version = Config.getConfig().VERSION;
-    window.Countly.debug = !isProduction;
+    const {COUNTLY_ENABLE_LOGGING, VERSION} = Config.getConfig();
+
+    window.Countly.app_version = VERSION;
+    window.Countly.debug = COUNTLY_ENABLE_LOGGING;
```
V-Gira, line 249: I mean this as a guarantee countly logging doesn't happen on prod environments, even if a mistake happens with the env var

```suggestion
    window.Countly.debug = COUNTLY_ENABLE_LOGGING && !isProduction;
```

src/script/tracking/EventTrackingRepository.ts
aweiss-dev: I dont believe we need that. 

This was just added because we didnt had an env variable.

V-Gira: All good then

V-Gira (approved):

src/script/tracking/EventTrackingRepository.ts
aweiss-dev: Actually i believe this setting has no effect in any way.

Since countly doesnt react on this variable change dynamically and it is set the first time in the html files

PR #18173, opened 2024-10-18
feat(RightSidebar): Incorrect separator in conversation details

przemvs:
## Description

Changed separator in conversation details for read receipts and changed design for conversation details according to the design.

Before:
image

After:
image

## Checklist

### DIFF
--- a/src/script/components/panel/UserActions.tsx
+++ b/src/script/components/panel/UserActions.tsx
@@ -29,6 +29,7 @@ import {WebAppEvents} from '@wireapp/webapp-events';
 
 import * as Icon from 'Components/Icon';
 import {PrimaryModal} from 'Components/Modals/PrimaryModal';
+import {listCSS} from 'Components/panel/PanelActions/PanelActions.styles';
 import {ConversationState} from 'src/script/conversation/ConversationState';
 import {SidebarTabs, useSidebarStore} from 'src/script/page/LeftSidebar/panels/Conversations/useSidebarStore';
 import {TeamState} from 'src/script/team/TeamState';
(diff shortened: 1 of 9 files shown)


src/script/components/panel/UserActions.tsx
```diff
@@ -367,7 +368,9 @@ const UserActions: React.FC<UserActionsProps> = ({
       onCancel={onAction}
     />
   ) : (
-    <PanelActions items={items} />
+    <ul css={listCSS}>
+      <PanelActions items={items} />
+    </ul>
```
olafsulich, line 373: Wouldn't it be simpler to leave `<ul>` inside PanelActions? I found it a little bit tricky that the PanelActions component returns a list of `<li>`, and has to be wrapped with an additional `<ul>` when called (because `<li>` can't be used outside of `<ul>`). And also it uses styles from the `PanelActions.styles`. What's the difference there?

przemvs: Nah, because PanelActions is used in 2 places, which 2nd place has parent ul :(

olafsulich: By 2nd place you mean `ConversationDetails.tsx`? I believe the parent there is `FadingScrollbar` which has a `<div>` inside

przemvs: image

przemvs: It's here: src/script/page/RightSidebar/ConversationDetails/components/ConversationDetailsOptions/ConversationDetailsOptions.tsx

olafsulich: Ohh I see now :) Never mind then!

olafsulich (approved):
looks good ✅

PR #18174, opened 2024-10-18
fix: Do not use detached window for camera source

thisisamir98:
we should explicitly check that requested media stream is screen and then use detached window for it's source if we have a detached window, otherwise if we also request video from the detached window when user closes it the video ends up blank black.

### DIFF
--- a/src/script/media/MediaStreamHandler.ts
+++ b/src/script/media/MediaStreamHandler.ts
@@ -166,7 +166,8 @@ export class MediaStreamHandler {
 
     const detachedWindow = callState.detachedWindow();
     const isInDetachedMode = callState.viewMode() === CallingViewMode.DETACHED_WINDOW;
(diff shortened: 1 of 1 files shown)


src/script/media/MediaStreamHandler.ts
```diff
@@ -166,7 +166,8 @@ export class MediaStreamHandler {
 
     const detachedWindow = callState.detachedWindow();
     const isInDetachedMode = callState.viewMode() === CallingViewMode.DETACHED_WINDOW;
-    const useDetachedWindowForScreenSharingSource = isInDetachedMode && detachedWindow !== null;
+    const useDetachedWindowForScreenSharingSource =
+      screen === true && video === false && isInDetachedMode && detachedWindow !== null;
```
olafsulich:
```suggestion
    const useDetachedWindowForScreenSharingSource =
      screen && !video && isInDetachedMode && detachedWindow !== null;
```

olafsulich (approved):
✅

PR #18175, opened 2024-10-18
feat: Disabled enforce Countly by env variable

przemvs:
## Description

Added flag to disable enforcing Countly.

## Checklist


src/script/tracking/Countly.helpers.ts
```diff
@@ -35,10 +35,10 @@ export const getForcedErrorReportingStatus = (): boolean => forceActivateErrorRe
 
 // Init the forced activation of error reporting based on the environment
 export const initForcedErrorReporting = () => {
-  const {COUNTLY_ENABLE_LOGGING} = Config.getConfig();
+  const {COUNTLY_ENABLE_LOGGING, DISABLE_ENFORCE_COUNTLY} = Config.getConfig();
   const {isDev, isEdge, isInternal, isLocalhost, isStaging, name} = getWebEnvironment();
 
-  if (isDev || isEdge || isInternal || isLocalhost || isStaging) {
+  if (!DISABLE_ENFORCE_COUNTLY && (isDev || isEdge || isInternal || isLocalhost || isStaging)) {
```
V-Gira, line 41: I feel like this would achieve the same thing:
```suggestion
  if (isEdge || isInternal || isStaging) {
```

We only want to enforce it on internal envs, not really dev envs

aweiss-dev (changes requested):
This PR should be dropped. 

It used the "wrong" logic, to disable enforcement via a env variable. We already changed the enforcement to use a env variable instead of in-app logic

PR #18199, opened 2024-10-23
feat: Adjust in call reactions styles (WPB-10241)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10241" title="WPB-10241" target="_blank">TaskWPB-10241</a>  [Web] Adjust in-call reaction design for use within pop out calling view
  </td></table>
  <br />
 

Before:
image
image

After:
image
image

### DIFF
--- a/src/style/foundation/video-calling.less
+++ b/src/style/foundation/video-calling.less
@@ -199,6 +199,10 @@
     display: none;
   }
 
+  &__button_primary {
+    .button-icon-primary;
+  }
+
   &__button {
     .button-reset-default;
     position: relative;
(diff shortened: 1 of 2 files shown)


e-maad (approved):
LGTM 🚀

src/style/foundation/video-calling.less
```diff
@@ -272,16 +276,16 @@
     animation:
       flyUp 4s linear forwards,
       fadeIn 1s;
-    font-size: 2rem;
+    font-size: 1.75rem;
     gap: 0.1rem;
   }
 
   .emoji-text {
```
olafsulich, line 283: Correct me if I'm wrong, but as I'm looking [at design](https://www.figma.com/design/jnygr7PEOk1yHgjBF18udc/Desktop?node-id=3312-61738&node-type=frame&t=7NqH7vrGaNIYxOqC-0) I see the button should be blue, instead of green, as you've posted in PR's description. 
 
Screenshot 2024-10-23 at 13 19 32

thisisamir98: it should use the accent color of the account @olafsulich

olafsulich: got it!

PR #18200, opened 2024-10-24
feat(VideoAsset): handle unplayable videos [WPB-11667]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11667" title="WPB-11667" target="_blank">TaskWPB-11667</a>  [Web] Display unsupported video as file + countly event
  </td></table>
  <br />
 

## Description

Handles unsupported videos:
- if we can detect unsupported mime type => display it as a file

Screenshot 2024-10-24 at 08 12 19

- if we have to perform more advanced checks, e.g. to detect unsupported codecs => error message + download button (consulted with Design)

Screenshot 2024-10-24 at 08 05 29

I added logs to Countly, let's measure how big the problem is.

## Checklist

- [] If it is a core feature, unit tests have been added;


przemvs:
We want to display big placeholder for unsupported video? :)

olafsulich:
> We want to display big placeholder for unsupported video? :)

@przemvs It's the size of the video player, hmm I wouldn't shrink that down. Imagine you click the play button, and then, because of the error, the player disappears and some smaller error message shows up 😕 

https://github.com/user-attachments/assets/08b0e624-20a2-431c-9dbd-b2a930f94b86

przemvs (approved):

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -77,12 +82,39 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
...
       videoPreview?.dispose();
       videoSrc?.dispose();
     };
   }, []);
 
+  // Initial check if video is supported with `canPlayType` method, which checks for MIME type, e.g. 'video/mp4' or 'video/mov'.
+  // It's not 100% reliable (e.g. doesn't check codecs), but it's synchorous, which is helpful for initial rendering.
+  const isVideoMimeTypeSupported = (mimeType: string): boolean => {
+    const video = document.createElement('video');
+    const canPlay = video.canPlayType(mimeType) !== '';
```
e-maad, line 96: Just a suggestion, if we create an enum for the return types it would be more readable:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType#return_value

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
```diff
@@ -153,6 +205,7 @@ const VideoAsset: React.FC<VideoAssetProps> = ({
               poster={videoPreview?.url}
               onError={event => {
                 setVideoPlaybackError(true);
+                amplify.publish(WebAppEvents.ANALYTICS.EVENT, EventName.MESSAGES.VIDEO.PLAY_FAILED);
```
e-maad, line 208: There are 4 place where we logged `PLAY_FAILED` event.
Just curious if it will only log once per error.

e-maad (approved):
LGTM 🚀 
Added few comments, please have a look.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
olafsulich: I don't think that's necessary because we don't use it directly - we just check that it's not an empty string.

src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset.tsx
olafsulich: I've checked and it shouldn't fire a few at once. 

1. fires when mime type is not supported (the simple check)
2. fires when a video is not playable (the advanced check)
3. fires when something fails in the `try{}` block inside `onPlayButtonClicked`  (e.g. `getAssetUrl` function)
4. fires on the video error callback - I guess now it's not gonna be a common case (because we handled unplayable videos before that), but in case something fails, then we can log it 

I wanted to add logs in many places, so that we can later analyse how many videos were played successfully, how many ended with "unsupported mime type", and how many failed due to unsupported codecs.

e-maad: Thanks for explanation 🚀

PR #17307, opened 2024-04-22
fix: User availability state

przemvs:
## Description

In this PR updating users availability state was fixed. When user set availability state and reload app, availability state disappears. Currently added also **localUsers** to update all users availability state, not only incompleteUsers and nonQualified users.

## Screenshots/Screencast (for UI changes)

Before:
https://github.com/wireapp/wire-webapp/assets/13432884/50e38806-1b32-41a9-ac4a-882ec0f0a166

After:
https://github.com/wireapp/wire-webapp/assets/13432884/dd8eeb71-0c28-4f81-8e05-0e7929cdf48d

## Checklist

- [x] PR has been self reviewed by the author;
- [x] Hard-to-understand areas of the code have been commented;
- [x] If it is a core feature, unit tests have been added;

### Important details for the reviewers

(Delete this section if unnecessary)

- use (x) data
- can be reviewed commit-by-commit
- be sure to look at ...

PR #18304, opened 2024-11-13
feat(UserRepository): retain username after deleting account [WPB-11166]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11166" title="WPB-11166" target="_blank">TaskWPB-11166</a>  [web] Maintain user name when they are removed from the backend.
  </td></table>
  <br />
 

## Description

After a user deletes their account, we're displaying the "default" label instead of their name. 

Now, we would like to retain the original name in the local database and display it as shown before deletion.

To do this, I had to remove the logic for handling orphan users. Because of this, we were removing the deleted users from the database. It takes users who are participants in a conversation and decides to remove them from the database based on that.

I've checked some related PRs from the past but didn't find any reason (besides deleted users) for keeping the "orphanUsers" logic. Related PRs:

- https://github.com/wireapp/wire-webapp/pull/15076
- https://github.com/wireapp/wire-webapp/pull/15002
- https://github.com/wireapp/wire-webapp/pull/16405
- https://github.com/wireapp/wire-webapp/pull/17307
- https://github.com/wireapp/wire-webapp/pull/15005

## Checklist

### DIFF
--- a/src/script/user/UserRepository.ts
+++ b/src/script/user/UserRepository.ts
@@ -217,18 +217,6 @@ export class UserRepository extends TypedEventEmitter<Events> {
     const nonQualifiedUsers = await this.userService.clearNonQualifiedUsers();
 
     const dbUsers = await this.userService.loadUserFromDb();
-    /* prior to April 2023, we were only storing the availability in the DB, we need to refetch those users */
-    const [localUsers] = partition(dbUsers, user => !!user.qualified_id);
-
-    // We can remove users that are not linked to any "known users" from the local database.
-    // Known users are the users that are part of the conversations or the connections (or some extra users)
-    const orphanUsers = localUsers.filter(
-      localUser => !users.find(user => matchQualifiedIds(user, localUser.qualified_id)),
-    );
-
-    for (const orphanUser of orphanUsers) {
-      await this.userService.removeUserFromDb(orphanUser.qualified_id);
-    }
 
     // The self user doesn't need to be re-fetched
     const usersToFetch = users.filter(user => !matchQualifiedIds(selfUser.qualifiedId, user));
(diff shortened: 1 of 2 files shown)


src/script/user/UserRepository.ts
```diff
@@ -586,7 +574,27 @@ export class UserRepository extends TypedEventEmitter<Events> {
     );
   }
 
-  private mapUserResponse(found: APIClientUser[], failed: QualifiedId[], dbUsers?: UserRecord[]): User[] {
+  // Replaces a deleted user name ("default") with the name from the local database.
+  private replaceDefaultUserName(apiUsers: APIClientUser[], dbUsers: UserRecord[]): UserRecord[] {
+    return apiUsers.map(user => {
+      if (user.name !== 'default') {
```
e-maad: Just curious, did we decide to use `deleted` prop instead of name?

aweiss-dev: If we talk about deleted users then we should rely on the is isDeleted attribute of the User Entity instead of the name.

olafsulich: You're right, good catch. I've just updated the check (and the function name)

src/script/user/UserRepository.ts
```diff
@@ -586,7 +574,27 @@ export class UserRepository extends TypedEventEmitter<Events> {
     );
   }
 
-  private mapUserResponse(found: APIClientUser[], failed: QualifiedId[], dbUsers?: UserRecord[]): User[] {
+  // Replaces a deleted user name ("default") with the name from the local database.
+  private replaceDeletedUserName(apiUsers: APIClientUser[], dbUsers: UserRecord[]): UserRecord[] {
```
e-maad: Just a minor suggestion, completely your call.
```suggestion
  private replaceDeletedUserNames(apiUsers: APIClientUser[], dbUsers: UserRecord[]): UserRecord[] {
```

e-maad (approved):
LGTM 🚀 
One minor suggestion for function name.

src/script/user/UserRepository.ts
olafsulich: Yeah, it's gonna be better, thanks :) 

Also, I've changed `replace` to `restore` - it should be more self-descriptive now

przemvs (approved):
LGTM!

PR #18305, opened 2024-11-13
runfix: block mobile browsers

aweiss-dev:
## Description

The PR from a while ago which reworked our way of detecting browsers [#17325](https://github.com/wireapp/wire-webapp/pull/17325) had the side effect that our mobile detection has been removed as well.

This PR adds the mobile detection back in and blocks mobile devices.

### DIFF
--- a/src/script/browser/CheckBrowser.ts
+++ b/src/script/browser/CheckBrowser.ts
@@ -96,6 +96,7 @@ const checkBrowser = (): void => {
     console.error("This browser doesn't support cookies to run the Wire app!");
     return;
   }
+  // Skip the mobile browser check for OAuth
   if (isOauth()) {
     return;
   }
--- a/src/script/browser/CheckBrowser.ts
+++ b/src/script/browser/CheckBrowser.ts
@@ -99,12 +99,13 @@ const checkBrowser = (): void => {
   if (isOauth()) {
     return;
   }
-  if (!('RTCPeerConnection' in window)) {
+  const isMobile = window.matchMedia('(any-pointer:coarse)').matches;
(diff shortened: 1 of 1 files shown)


src/script/browser/CheckBrowser.ts
```diff
@@ -96,15 +96,17 @@ const checkBrowser = (): void => {
     console.error("This browser doesn't support cookies to run the Wire app!");
     return;
   }
+  // Skip the mobile browser check for OAuth
   if (isOauth()) {
     return;
   }
-  if (!('RTCPeerConnection' in window)) {
+  const isMobile = window.matchMedia('(any-pointer:coarse)').matches;
```
olafsulich: I suggest we enhance the mobile detection logic for better reliability. The current `window.matchMedia('(any-pointer:coarse)')` check has some limitations.

It detects touch input, which could give false positives on touch-enabled laptops and desktops (such as Surface devices, touch-screen Windows laptops, etc.). Some desktop users might incorrectly see the "unsupported device" message.

It might be a good idea to look at other checks, like checking the user agent:

```ts
const isMobileDevice = () => {
    const isTouch = ('ontouchstart' in window) || (navigator.maxTouchPoints > 0);
    const isSmallScreen = window.innerWidth <= 768;
    const isMobileUserAgent = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
        navigator.userAgent
    );
    
    return isTouch && (isSmallScreen || isMobileUserAgent);
}
```

olafsulich (approved):
lgtm ✅ 

it should be working great, but i would give it to qa to test on some different mobile device user agents, including the most common androids and iphones

PR #18312, opened 2024-11-15
fix: guest link password copy [WPB-11041]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11041" title="WPB-11041" target="_blank">BugWPB-11041</a>  [web] Guest link password copy is broken.
  </td></table>
  <br />
 

## Description

**Problem:** 

Our translation files have a different syntax for passing dynamic properties inside a text string. One uses single brackets `{text}`, other uses double `{{text}}`.

`strings.ts`:
```ts
export const joinGuestLinkPasswordModalStrings = defineMessages({
  headline: {
    defaultMessage: '{conversationName} \n Enter password',
    id: 'guestLinkPasswordModal.headline',
  },
}
```

vs

`en-US.json`:
```
"guestLinkPasswordModal.headline": "{{conversationName}} Enter password",
```

It caused a problem with the guest link password copy. These two ways weren't compatible which caused the `conversationName` not to be rendered properly. 

image

**Solution:**

Unify these approaches, use single brackets for both cases - in the translation files, and in the `string.ts` the `defineMessages` function.

Why not the other way around? (double curly brackets everywhere)

We don't control the `defineMessages` function, it comes from `react-intl`. However we do control the `t` function, it's written by us. 

I've tried to "override" the `defineMessages` with our own implementation (I prefer the double ones :)), but got some errors. If it's just a matter of preference, let's stick with the single brackets.

## Checklist


e-maad (commented):
The changes look fine but I have a question:
Can we solve this issue by using our custom function for translations `t` from `LocalizerUtil.ts`? 
I think for new features we are consuming it and we don't have to provide default message in that case.

olafsulich:
> The changes look fine but I have a question: Can we solve this issue by using our custom function for translations `t` from `LocalizerUtil.ts`? I think for new features we are consuming it and we don't have to provide default message in that case.

@e-maad do you mean, just don't use `strings.ts` in that case, and use the `t` function?

e-maad:
> > The changes look fine but I have a question: Can we solve this issue by using our custom function for translations `t` from `LocalizerUtil.ts`? I think for new features we are consuming it and we don't have to provide default message in that case.
> 
> @e-maad do you mean, just don't use `strings.ts` in that case, and use the `t` function?

Yes, using `t` function instead of `formatMessage` from `useIntl()`

PR #18347, opened 2024-11-20
feat(i18n): enhance types

olafsulich:
## Description

Automatically generates types for internalization file (`en-US.json`). The script runs on the server start and on a file change (webpack / hot reload). Every change to the `en-US.json` will cause an update of `i18n.d.ts`.

The `t` function now checks for dynamic value types. There's also a unified syntax for passing variables, we eliminate multiple ways (with an object and without) of passing dynamic values. Works with single and multiple values.

**What it means?**

- the only allowed variables would be those presented in the translation text
- ts will let us know when we forgot to pass a variable / pass a wrong one / pass a variable for a text that doesn't have specified one

**Before:**

```ts
// en-US.json
{
 hi: 'hi',
 hello: 'hello {text}'
}

---

t('hi') // ✅ <-- correct

t('hi', 'world') // ✅ <-- incorrect, doesn't throw an error
t('hi', {name: 'world'}) // ✅ <-- incorrect, doesn't throw an error

---

t('hello', 'world') // ✅ <-- correct
t('hello', {text: 'world'}) // ✅ <-- correct

t('hello', {name: 'world'}) // ✅ <-- incorrect, doesn't throw an error
t('hello') // ✅ <-- incorrect, doesn't throw an error

``` 

**After:**

```ts
// en-US.json
{
 hi: 'hi',
 hello: 'hello {text}'
}

---

t('hi') // ✅ <-- correct

t('hi', 'world') // ❌ <-- incorrect, throws an error (redundant argument, wrong syntax)
t('hi', {name: 'world'}) // ❌ <-- incorrect, throws an error (redundant argument)

---

t('hello', {text: 'world'}) // ✅ <-- correct

t('hello', 'world') // ❌ <-- incorrect, throws an error (wrong syntax)
t('hello', {name: 'world'}) // ❌ <-- incorrect, throws an error (wrong value "name")
t('hello') // ❌ <-- incorrect, throws an error (missing value)

```  

I've fixed many places to use the new way. I've left some places with `@ts-expect-error` - I don't want to break anything / it should be done in other PRs.

## Checklist


bin/transalations_generate_types.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs');
```
olafsulich, line 1: an important file

src/script/components/MessagesList/Message/FailedToAddUsersMessage.tsx
```diff
@@ -108,25 +108,58 @@ const MessageDetails = ({failure, isMessageFocused, allUsers}: MessageDetailsPro
     </>
   );
 
+  const getText = () => {
```
olafsulich: a lot of changes there, but it had to be done, because of the type, logic remains the same as it was

src/script/util/LocalizerUtil/LocalizerUtil.ts
```diff
@@ -19,11 +19,28 @@
```
olafsulich, line 19: an important file

src/script/util/LocalizerUtil/LocalizerUtil.ts
```diff
@@ -135,14 +162,14 @@ export const setStrings = (newStrings: typeof strings): void => {
...
-  dangerousSubstitutions?: Record<string, string>,
-  skipEscape: boolean = false,
-): string {
-  return LocalizerUtil.translate(identifier, substitutions, dangerousSubstitutions, skipEscape);
-}
+export const t = <Id extends StringIdentifer>(
+  identifier: Id,
+  ...args: ExtractSubstitutionKeys<TranslationStrings[Id]> extends never
+    ? [substitutions?: undefined, dangerousSubstitutions?: Record<string, string>, skipEscape?: boolean]
+    : [substitutions: Substitutions<Id>, dangerousSubstitutions?: Record<string, string>, skipEscape?: boolean]
```
olafsulich, line 169: unfortunately, the `args` are mandatory here (instead of typing each parameter separately), because we have to mark `substitutions` either **optional** (the `?` sign) with undefined type, or **required** with the `Substitutions<Id>` type

src/types/i18n.d.ts
```diff
@@ -0,0 +1,1654 @@
+declare module 'I18n/en-US.json' {
```
olafsulich, line 4: an important file

olafsulich: i've thought about creating a normal `.ts` file, but IMO `.d.ts` is fine in this case

the normal `.ts` file would require additional import + type assertion the json

webpack.config.dev.js
```diff
@@ -25,6 +25,35 @@ const commonConfig = require('./webpack.config.common');
 
 const srcScript = 'src/script/';
 
+const updateTranslationTypesPlugin = {
```
olafsulich, line 30: an important file

przemvs (approved):
LGTM!

PR #18349, opened 2024-11-20
feat: Uploading assets

przemvs:
## Description

Changed method of uploading assets to avoid not sent files. If user uploading image, and want to close app action will be got native alert about uploading assets.

Before:

https://github.com/user-attachments/assets/6b974994-1f7e-471e-8385-718448c25342

After:

https://github.com/user-attachments/assets/276f973f-7fb6-4ef0-9a4e-30455f958363

https://github.com/user-attachments/assets/c3c5a1ad-6530-420a-b20c-eb42998618a0

## Checklist


src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import {AssetRepository} from '../../../../../assets/AssetRepository';
+
+interface Props {
+  assetRepository: AssetRepository;
+  message: GenericMessage;
+}
+
+export const UploadAssetItem = ({assetRepository, message}: Props) => {
+  const [uploadProgress, setUploadProgress] = useState<number>(0);
```
olafsulich: minor, the `number` here seems unnecessary, TypeScript is going to infer the type: 
```suggestion
  const [uploadProgress, setUploadProgress] = useState(0);
```

src/script/components/MessagesList/UploadAssets/UploadAssets.tsx
```diff
@@ -0,0 +1,63 @@
...
+      {processQueue.map((processingMessage, key) => {
+        const processingAsset = uploadProgressQueue.find(
+          message => message.messageId === processingMessage.message.messageId,
+        );
+
+        if (!processingAsset) {
+          return null;
+        }
+
+        return <UploadAssetItem assetRepository={assetRepository} message={processingMessage.message} key={key} />;
```
thisisamir98: maybe let's use key={processingMessage.message.messageId}

for the key, map indexes can lead to suboptimal React reconciliation

src/script/conversation/MessageRepository.ts
```diff
@@ -602,7 +613,47 @@ export class MessageRepository {
...
+      meta.audio = metadata as AudioMetaData;
+    } else if (isVideo(file)) {
+      meta.video = metadata as VideoMetaData;
+    } else if (allowImageDetection && isImage(file)) {
+      meta.image = metadata as ImageMetaData;
+    }
+
+    const message = MessageBuilder.buildFileMetaDataMessage({metaData: meta as FileMetaDataContent}, originalId);
+    this.assetRepository.addToProcessQueue(message, conversation.id);
+    return {message, metaData: meta as FileMetaDataContent};
```
olafsulich: (minor) We could refactor it a little:

- we could wrap the whole code inside the try/catch block, avoiding reassigning the metadata variable
- is the `Partical` required here? I think we could omit it and just use the`FileMetaDataContent`, because we would have all required variables (based on `packages/core/src/conversation/content/FileContent.ts`)
- by that, we could also create the meta object without dynamically adding properties to it, perform a condition, and if it's true, assign the `metadata`, otherwise leave it `undefined`

```suggestion
    try {
    const metadata = await buildMetadata(file);
    
    const meta = {
      length: file.size, 
      name: (file as File).name, 
      type: file.type,
      audio: isAudio(file) && metadata,
      video: isVideo(file) && metadata,
      image allowImageDetection && isImage(file) && metadata
     } as FileMetaDataContent;

    const message = MessageBuilder.buildFileMetaDataMessage({metaData: meta}, originalId);
    
    this.assetRepository.addToProcessQueue(message, conversation.id);
    
    return {message, metaData: meta};
    } catch (error) {
      const logMessage = `Couldn't render asset preview from metadata. Asset might be corrupt: ${
        (error as Error).message
      }`;
      this.logger.warn(logMessage, error);
    }
```

src/script/components/MessagesList/UploadAssets/UploadAssets.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const currentConversationProcessQueue = processQueue.filter(item => item.conversationId === conversationId);
+
+  if (!currentConversationProcessQueue.length) {
+    return null;
+  }
+
+  return (
+    <div css={uploadAssetsContainer} data-uie-name="upload-assets">
+      {processQueue.map((processingMessage, key) => {
```
thisisamir98: isn't it better to use currentConversationProcessQueue here?

```suggestion
      {currentConversationProcessQueue.map((processingMessage, key) => {
```

src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  useEffect(() => {
+    const progressSubscribable = assetRepository.getUploadProgress(message.messageId);
```
thisisamir98, line 39: can't we do something like 

const progressSubscribable = assetRepository.getUploadProgress(message.messageId);

progress = useKoSubscribableChildren(progressSubscribable...)

instead of a useEffect here?

src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
thisisamir98: I actually like the explicit approach, this way later on if you try to do setUploadProgress("somestring") typescript will have an error that this value must be a number

src/script/components/MessagesList/UploadAssets/UploadAssets.tsx
```diff
@@ -0,0 +1,63 @@
...
+        const processingAsset = uploadProgressQueue.find(
```
thisisamir98: this find inside the map sounds inefficient for large queues, maybe we could build a map of uploadProgressQueue before rendering to reduce redundant lookups:

```typescript
const uploadProgressMap = new Map(
  uploadProgressQueue.map(item => [item.messageId, item])
);

return (
  <div css={uploadAssetsContainer} data-uie-name="upload-assets">
    {currentConversationProcessQueue.map(processingMessage => {
      const processingAsset = uploadProgressMap.get(processingMessage.message.messageId);

      if (!processingAsset) {
        return null;
      }

      return (
        <UploadAssetItem
          assetRepository={assetRepository}
          message={processingMessage.message}
          key={processingMessage.message.messageId}
        />
      );
    })}
  </div>
);
```

src/script/components/MessagesList/UploadAssets/UploadAssets.tsx
przemvs: You're right.. should be this one :)

src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
olafsulich: If the initial value is `0`, then TS won't let you pass anything else than a number:

Screenshot 2024-11-26 at 10 08 42

src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
przemvs: It won't work :( We have to subscribe event ;(

src/script/components/MessagesList/UploadAssets/components/UploadAssetItem/UploadAssetItem.tsx
przemvs: Yee, if You add default value it's automatically getting type ;)

olafsulich (approved):

thisisamir98 (approved):

PR #18359, opened 2024-11-22
chore: Serialize object arguments in desktop runtime

thisisamir98:
#### **Overview**  
This PR enhances the logging functionality for the desktop runtime by serializing object arguments into JSON format. This prevents object arguments from being logged as `[object Object]`, improving the readability and utility of log messages.  

#### **Changes Introduced**  
- Updated the `getLogger` function to:
  - Inspect arguments passed to logger methods.
  - Serialize objects to JSON using `JSON.stringify(arg, null, 2)` for better formatting.  
- Applied the new behavior specifically for the desktop runtime (`Runtime.isDesktopApp()`).
- Ensured non-object arguments are logged as-is to maintain the expected behavior.  

#### **Why is This Change Needed?**  
Currently, when logging object arguments, they are output as `[object Object]`, which is unhelpful for debugging and diagnostics. This is particularly problematic for:
- Configuration objects.
- Error payloads.
- Other structured data critical for understanding runtime behavior.

By serializing objects to JSON:
- Logs are more informative and easier to read.
- Developers can quickly identify issues without needing additional context or manual object inspection.
- It aligns with best practices for structured logging.

#### **Testing**  
- Verified that log messages containing objects are now serialized as JSON in the desktop runtime.
- Confirmed non-object arguments (e.g., strings, numbers) remain unaffected.
- Ensured logging behavior is unchanged in non-desktop runtimes.


aweiss-dev (approved):

src/script/util/Logger.ts
```diff
@@ -17,15 +17,42 @@
...
+        logger.log(...serializeArgs(args));
+      },
+      warn: (...args: any[]): void => {
+        logger.warn(...serializeArgs(args));
+      },
+    };
+  }
+
+  return logger;
 }
```
olafsulich, line 56: Not a big thing, but `unknown` wouldn't be better for this case?

thisisamir98: I've followed the signature of the original logdown library:

image

olafsulich: sure, it's not a big deal either way 👌

PR #18445, opened 2024-12-06
feat: Screen Share improvement for Group Calls [WPB-10134]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10134" title="WPB-10134" target="_blank">StoryWPB-10134</a>  [AVS] Change Screen Share constrains for better Screen Share quality in Group Call
  </td></table>
  <br />
 

###  This is an improvement to the readability of screen sharing:

- I have removed the downscaling on the sender side.
- I mark screen share videos as "All Content Detailed."
- If someone receives a 4K screen share on their 1080p monitor, it will naturally be downscaled on the receiver's side. For this case, I have added a zoom feature.
- We now send screen shares at 5 fps by default.

Bildschirmfoto 2024-12-06 um 17 48 37

@WolfiWire 

I think you'll have an opinion on this, but I'd like to adjust the zoom feature in a fallow up PR. I hope this is ok 
(These PRs always take so long :) )

## Checklist

- [x ] mentions the JIRA issue in the PR name (Ex. [WPB-XXXX])


src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -105,8 +108,29 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
...
   };
 
+  const handleZoomClick = () => {
+    if (isZoomedIn) {
+      setIsZoomedIn(false);
+    } else {
+      setIsZoomedIn(true);
+    }
+  };
+
```
thisisamir98:
```suggestion
  const handleZoomClick = () => {
    setIsZoomedIn(prev => !prev); 
```

src/script/components/calling/GroupVideoGridTile.tsx
```diff
@@ -105,8 +108,29 @@ const GroupVideoGridTile: React.FC<GroupVideoGridTileProps> = ({
...
   const participantNameColor = getParticipantNameColor({isActivelySpeaking, isAudioEstablished});
 
+  const actionItem = !minimized && sharesScreen && (
+    <button
+      data-uie-name="item-grid"
+      data-user-id={participant?.user.id}
+      className="group-video-grid__element__action_icon"
+      onClick={handleZoomClick}
+      onKeyDown={handleZoomClick}
```
thisisamir98: two are unnecessarily, only onClick should be enough

src/script/team/TeamState.ts
```diff
@@ -117,9 +117,7 @@ export class TeamState {
       () => this.teamFeatures()?.mls?.config.protocolToggleUsers.includes(this.userState.self().id) ?? false,
     );
 
-    this.isConferenceCallingEnabled = ko.pureComputed(
-      () => this.teamFeatures()?.conferenceCalling?.status === FeatureStatus.ENABLED,
-    );
+    this.isConferenceCallingEnabled = ko.pureComputed(() => true);
```
thisisamir98, line 120: if it's gonna be always true I suggest a bigger refactor to just get rid of this property every where, but for now we could at least set it to an always true observable maybe

```suggestion
    this.isConferenceCallingEnabled = ko.observable(true);
```

EnricoSchw: O good finding.. ❤️  Thx. this is a hack to activate group call.. This should never go to staging. I will remove this!

src/script/components/calling/GroupVideoGridTile.tsx
EnricoSchw: You use both events in many places... I thought it was a workaround.  I will remove it

thisisamir98 (approved):

PR #18470, opened 2024-12-10
feat: press space to unmute [WPB-8580]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8580" title="WPB-8580" target="_blank">StoryWPB-8580</a>  [Web] Press space to unmute (MVP)
  </td></table>
  <br />
 

## Description

**How it works?**

- users can enable/disable the option "Press and hold to unmute" in audio/video settings (disabled by default) 
- they can unmute themselves by pressing and holding the space bar, with a delay of 200ms (tested, worked best for me)
- automatically mutes the user upon releasing the space bar
- works only when the user is already muted
- works only in the fullscreen and detached window view
- a notification is shown when holding a space
- it's behind a feature flag

**What's changed?**

- we no longer have an option to specify a key for unmuting the mic, we had it in the code, but it wasn't a product/UI feature - as long as we don't have such a need, we don't have to specify it

---

**Enabled:**

https://github.com/user-attachments/assets/8b3fd712-50f5-4f43-9e1b-5a4183140955

---

**Disabled:**

https://github.com/user-attachments/assets/2cc17ac6-31a4-4808-96be-dcaaa5653997

---

**Dynamic:**

plus a demo of it doesn't work outside fullscreen

https://github.com/user-attachments/assets/882f20e1-474c-4736-9126-826e96feba08

## Checklist


src/script/components/calling/CallingCell/usePressSpaceToUnmute/usePressSpaceToUnmute.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const HOLD_DELAY_MS = 200;
+
+export const usePressSpaceToUnmute = ({callState, toggleMute, isMuted, enabled}: UsePressSpaceToUnmuteParams) => {
+  const isInCallAndViewMode = checkUserInCallAndViewMode(callState);
+
+  const {detachedWindow, viewMode} = callState;
+
+  const activeWindow = viewMode() === CallingViewMode.DETACHED_WINDOW ? detachedWindow()! : window;
```
przemvs, line 47: I think for this one we can make helper to check if is window or detachedWindow :) Some hook maybe ?

przemvs (approved):

src/script/components/calling/CallingCell/usePressSpaceToUnmute/usePressSpaceToUnmute.ts
olafsulich: @przemvs, how would you do that? Either way, we have to get `viewMode()` out of the `callState`. So the hooks/function would only be "responsible" for this part `viewMode() === CallingViewMode.DETACHED_WINDOW ? detachedWindow()! : window;`. Is it that much? I don't think so 🤷‍♂️ 

If we could like not pass the `callState` every time, then sure, it'd be a great addition 👌

src/script/page/MainContent/panels/preferences/avPreferences/CallOptions.tsx
```diff
@@ -111,6 +117,26 @@ const CallOptions: React.FC<CallOptionsProps> = ({constraintsHandler, properties
...
         </p>
       </div>
+      {isPressSpaceToUnmuteFlagEnabled && (
+        <div className="checkbox-margin">
+          <Checkbox
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              const isChecked = event.target.checked;
+              propertiesRepository.savePreference(PROPERTIES_TYPE.CALL.ENABLE_PRESS_SPACE_TO_UNMUTE, isChecked);
+              setPressSpaceToUnmuteEnabled(isChecked);
+            }}
```
thisisamir98: too much logic, maybe extract to a separate function

thisisamir98 (approved):

PR #18476, opened 2024-12-11
fix: app notification

olafsulich:
## Description

**Limitations with the current notifications system:**

The press space to unmute feature is coming. Along with the hand-raised one, it will be another notification displayed in calls.

With the current approach, it would be problematic:

1. A user named "Jacob" clicked the "raise hand" button
2. I see a notification "Jacob raised his hand"
3. I'm muted, but I want to unmute myself with the "press space to unmute" feature
4. I press space, and the notification about Jacob's raised hand disappears, and in its place, the new one shows up ("Mic temporarily on")

Support for multiple notifications is needed, I'm pretty sure that we'll have more of these cases sometime in the future.

---

**Notifications in the detached window:**

The detached window is the biggest problem, though. Here's the demo (current implementation) of it being broken, on the example of hand-raised notification:

https://github.com/user-attachments/assets/da9a587e-d89d-481f-a064-de7c9fcc735d

If I'm in the fullscreen view (the default window), I see the hand-raised notifications ✅ 

But if I change the call to be in the detached view, then I don't see them ❌ They're displayed in the "old" (the default window) view.

The above situation is kinda problematic, especially for the call notifications, and when someone uses the detached window. Every time someone receives/sends a notification in the detached view, it'll transmit to the default window.

--- 

**Solution:**

Use an external library for notifications production. We don't need to reinvent the wheel, and handling all of those cases is pretty problematic. At first, I tried to expand our implementation to support multiple notifications, but I gave up - the implementation started to grow quickly. 

That's why, I picked the [Sonner Library](https://github.com/emilkowalski/sonner) instead, it's easy to use, and it has some great defaults.

Using the Sonner, I build a custom wrapper on top of it. It contains a system for handling windows. As a customer of the hook, you're able to pick in which window you want your notifications to appear, it can be dynamic:

```ts
const {show} = useAppNotification({
    activeWindow: viewMode === CallingViewMode.DETACHED_WINDOW ? detachedWindow! : window,
  });
  
  // ...
  
  show();
```

**Tradeoffs:**

- We no longer can't use notifications in non-react components (such as repositories), we can argue if it's a good or bad thing, but it's a limitation somehow. If we want to display a notification after some action inside e.g. a repository, we have to publish a new event, call the hook inside appropriate component, subscribe to the event, and then show a notification

--- 

**Multiple notifications:**

https://github.com/user-attachments/assets/30be540e-13ac-44ba-b9c6-88eb7a4ee7f6

---

**Transmit notifications to another window:**

https://github.com/user-attachments/assets/dfb4cbea-801d-46f7-abc3-5b3c68d440cb

## Checklist


package.json
```diff
@@ -13,7 +13,7 @@
     "@wireapp/react-ui-kit": "9.28.0",
     "@wireapp/store-engine-dexie": "2.1.15",
     "@wireapp/telemetry": "0.1.4",
-    "@wireapp/webapp-events": "0.24.3",
+    "@wireapp/webapp-events": "file:.yalc/@wireapp/webapp-events",
```
olafsulich: To be replaced after https://github.com/wireapp/wire-web-packages/pull/6750 is merged

src/style/components/app-notification.less
```diff
@@ -0,0 +1,40 @@
+.app-notification {
```
olafsulich, line 1: Pretty much the same (added little design touches) styles as before, see the deleted file `src/script/components/AppNotification/AppNotification.styles.ts`

src/script/calling/CallingRepository.ts
```diff
@@ -866,7 +865,7 @@ export class CallingRepository {
           ? t('videoCallParticipantRaisedSelfHandUp')
           : t('videoCallParticipantRaisedTheirHandUp', {name});
 
-        showAppNotification(handUpMessage);
+        amplify.publish(WebAppEvents.CALL.HAND_RAISED, {notificationMessage: handUpMessage});
```
przemvs: I think we should avoid to use amplify.publish. Maybe we could use this example:

`window.dispatchEvent(new CustomEvent(WebAppEvents.CALL.HAND_RAISED, {notificationMessage: handUpMessage}))`

WDYT? :)

src/script/components/calling/CallingCell/CallingCell.tsx
```diff
@@ -159,6 +162,21 @@ export const CallingCell = ({
...
+  });
+
+  useEffect(() => {
+    amplify.subscribe(WebAppEvents.CALL.SCREEN_SHARING_ENDED, () => {
+      screenSharingEndedNotification.show();
+    });
+
+    return () => {
+      amplify.unsubscribeAll(WebAppEvents.CALL.SCREEN_SHARING_ENDED);
+    };
```
przemvs: And there:

```suggestion
    window.addEventListener(WebAppEvents.CALL.SCREEN_SHARING_ENDED, () => {
      screenSharingEndedNotification.show();
    });

    return () => {
      window.removeEventListener(WebAppEvents.CALL.SCREEN_SHARING_ENDED);
    };
```

src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -237,6 +240,20 @@ const FullscreenVideoCall = ({
...
+
+  useEffect(() => {
+    amplify.subscribe(WebAppEvents.CALL.HAND_RAISED, (event: {notificationMessage: string}) => {
+      handRaisedNotification.show({message: event.notificationMessage});
+    });
+
+    return () => {
+      amplify.unsubscribeAll(WebAppEvents.CALL.HAND_RAISED);
+    };
+  }, [handRaisedNotification]);
```
przemvs: And there also You can use this method ;)

przemvs (approved):

PR #18520, opened 2024-12-17
fix: rich text editor [WPB-12089]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12089" title="WPB-12089" target="_blank">StoryWPB-12089</a>  [Web] Text formatting via UI in text input field
  </td></table>
  <br />
 

## Description

**Features:**

- nested lists support

**Fixes:**

- mentions while editing a message (works with multiple mentions, a mention with spaces, etc.)
- inserting heading (now works like every other button, earlier works only when the text was selected
- remove link and blockquote markdown auto-changing (from markdown string to rich text)
- shred a few pixels shift when clicking the "show formatting" button aka "Aa" button
- increase line height for lists
- change the order of lists (first ordered, then ordered), like on iOS
- adjust the emoji picker position (display it above the button, not on the side)

**Other**
- lexical v0.21.0

**What's missing:**
- fixing list styles after pasting (waiting for https://github.com/facebook/lexical/issues/6949)
- a setting for the disabling auto transformation of markdown to rich text in preview, to be done in a separate PR
- the issue with adding a new paragraph while doing a soft break (Shift + Enter)
  - currently, those two produce the same output, because we're adding a new paragraph every time someone presses Shift + Tab (it's because we can't use only Enter because it sends a message)
     
     ```
     first line 
     second line
     
     fourth line
     ```

     and
     
      ```
     first line
 
     second line
     
     fourth line
     ```

## Checklist


thisisamir98 (approved):

przemvs (approved):
LGTM!

PR #18540, opened 2024-12-19
fix: Do not show notification when call feedback is skipped

thisisamir98:


src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.tsx
```diff
@@ -66,7 +66,7 @@ export const QualityFeedbackModal = () => {
     return null;
   }
 
-  const handleCloseModal = () => {
+  const handleCloseModal = (skipNotification = false) => {
```
olafsulich: **suggestion:** imo, it would be more clear to have an object, because now, by looking at the function call, I can't tell what this `true` is for

```suggestion
  const handleCloseModal = ({skipNotification}: {skipNotification: boolean}) => {
```

PR #18566, opened 2025-01-08
fix: mic space click [WPB-8580]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-8580" title="WPB-8580" target="_blank">StoryWPB-8580</a>  [Web] Press space to unmute (MVP)
  </td></table>
  <br />
 

## Description

**Problem:**

1. Have enabled the "press space to unmute" setting
2. Click with mouse/touchpad on the mute/unmute button
3. Button became focused
4. It's toggling when hitting space

We want to avoid this. When someone presses the space, their gonna see the mic state toggle very quickly. Other apps with the "press space to unmute" system have this implemented this way.

Why leave the "Enter" instead of completely disabling it?

The "Enter" natively (by default in the browser) is also a valid option

We should provide some option, without it someone who uses a keyboard can't change their mind about turning the mic on - "Now I want to speak for a longer period, and I'm gonna turn my mic on, instead of holding the space", seems like a common case

If we have the option to click with the  mouse/touchpad, there should be an option to use the keyboard only

**What's changed?**

1. Refactored the handleKeyDown to take an object as an arg, and add the ability to pass keys, instead of heaving hardcoded values
2. Enable only Enter for mic mute/unmute when the press space to unmute option is enabled

## Checklist


src/script/util/KeyboardUtil.ts
```diff
@@ -83,11 +83,16 @@ export const isRemovalAction = (key: string): boolean => removalKeys.includes(ke
...
 
-export const handleKeyDown = (
-  event: ReactKeyboardEvent<Element> | KeyboardEvent,
-  callback: (event?: ReactKeyboardEvent<Element> | KeyboardEvent) => void,
-) => {
-  if (event.key === KEY.ENTER || event.key === KEY.SPACE) {
+export const handleKeyDown = ({
+  event,
+  callback,
+  keys,
```
przemvs, line 89: Why not to add default keys here? 🤔  Then You don't have to add in every function this keys :)

olafsulich: I was thinking about this, but I have decided not to. Currently, I preserved the Space and the Enter keys for every function, because that what was the previous implementation - but do we event need both of them in any case, I don't know.

For the future usage, I rather have it required. If we mark the keys as optional with some default, imagine looking at this function, you wouldn't know what are the defaults, you'd have to go into the implementation and check those defaults - the function name doesn't communicate "hey, it's only for the Space and the Enter".

przemvs (approved):

PR #18574, opened 2025-01-09
feat: Show drafted message [WPB-10991]

przemvs:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-10991" title="WPB-10991" target="_blank">TaskWPB-10991</a>  [Web] Show "drafted message" in second line in conversation list
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-10991

## Description

Showing drafted message while user change tab to inform about there is drafted message.
All another messages/alerts will shown before drafted message.

image

## Checklist


src/script/components/InputBar/InputBar.tsx
```diff
@@ -599,6 +599,7 @@ export const InputBar = ({
               </div>
               {!isSelfUserRemoved && !pastedFile && (
                 <RichTextEditor
+                  conversation={conversation}
```
olafsulich: As I see we only use the conversation's id there, let's not pass the whole object, but just the id

```suggestion
                  conversationId={conversation.id}
```

src/script/components/ConversationListCell/components/CellDescription/CellDescription.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const CellDescription = ({conversation, mutedState, isActive, isRequest, unreadState}: Props) => {
+  const storageService = container.resolve(StorageService);
+  const {draftMessage} = useMessageDraftState();
+  const currentConversationDraftMessage = isActive ? '' : draftMessage[conversation.id];
+
+  const cellState = useMemo(() => generateCellState(conversation), [unreadState, mutedState, isRequest]);
+
+  const getDraftMessageContent = useCallback(async () => {
+    const storageKey = generateConversationInputStorageKey(conversation);
+    const storageValue = await storageService.loadFromSimpleStorage<any>(storageKey);
```
olafsulich: I'd add a type here

src/script/hooks/useMessageDraftState.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import {create} from 'zustand';
+
+type MessageDraftState = {
+  draftMessage: Record<string, string>;
+  setDraftMessage: (conversationId: string, message: string) => void;
+};
+
+export const useMessageDraftState = create<MessageDraftState>((set, get) => ({
+  draftMessage: {},
```
olafsulich: I think it should be `draftMessages`, because you store multiple messages there
```suggestion
  draftMessages: {},
```

src/script/components/ConversationListCell/components/CellDescription/CellDescription.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+interface Props {
+  conversation: Conversation;
+  mutedState: number;
+  isActive: boolean;
+  isRequest: boolean;
+  unreadState: UnreadState;
+}
+
+export const CellDescription = ({conversation, mutedState, isActive, isRequest, unreadState}: Props) => {
```
olafsulich, line 41: It's not handling formatting correctly, instead it's giving the raw markdown. The question is do we want it to be like that? I'd discuss this, IMO we shouldn't leave it like that

image

src/script/components/ConversationListCell/components/CellDescription/CellDescription.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const storageService = container.resolve(StorageService);
+  const {draftMessage} = useMessageDraftState();
+  const currentConversationDraftMessage = isActive ? '' : draftMessage[conversation.id];
+
+  const cellState = useMemo(() => generateCellState(conversation), [unreadState, mutedState, isRequest]);
+
+  const getDraftMessageContent = useCallback(async () => {
```
olafsulich: I'd move the logic for getting the draft out of the component, e.g. look how it's handled in the saveDraftState fn

src/script/components/InputBar/util/DraftStateUtil.ts
```diff
@@ -83,3 +87,19 @@ export const loadDraftState = async (
...
+  };
+
+  const editor = createEditor(config);
+  const editorState = editor.parseEditorState(storageValue);
+  const textContent = editorState.read(() => $getRoot().getTextContent());
+
+  const {setDraftMessage} = useMessageDraftState.getState();
+
+  setDraftMessage(conversation.id, textContent);
+};
```
olafsulich: We already have saving/loading draft logic. By adding another layer (state with messages), we're kinda duplicating the logic, and we add more complexity. Can we reuse what we already have? Example of saving the draft:

```tsx
// RichTextEditor.tsx

saveDraftState(JSON.stringify(editorState.toJSON()));
```

I know it's JSON, not text, but maybe we can modify it somehow and reuse it?

Because without that, look how my info we store:

1. Draft state in local storage (json)
2. Draft state in local storage (text)
3. Draft messages external store (text, zustand)
4. Editor state (lexical)
5. ...

olafsulich: The way this function is named gives me a sense that I'll receive the draft message content from it, but it's not working like that. 

```ts
const messageContent = getDraftMessageContent(...);
```

Instead, it puts the message in a store. It increases [cognitive load](https://github.com/zakirullin/cognitive-load) by not giving me the sense that I know what's going to happen next. 

---

Let's break through the whole process:
1. The `handleChange` fires, it sets editor json state to local storage and register message to zustand store (setDraftMessage).
2. In the `CellDescription` we're getting the `draftMessage` from `useMessageDraftState`. Then we pick a draft for the message we have (based on the conversation's id)
3. In the `CellDescription`, in the `useEffect`, we call `getDraftMessageContent`. It gets the editor json state from the localstorage an pass it to the `getDraftTextMessageContent` function
4. In this function we create new editor and call setDraftMessage again.

From my perspective, it's a lot. The state of the draft editor is set in many different places. We're switching between zustand/localstorage, between editor state and messages state. The whole process happens in many places, I have spent some time to figure out what's happening here.

src/script/components/ConversationListCell/components/CellDescription/CellDescription.tsx
przemvs: Let me ask Olga ;)

src/script/components/InputBar/InputBar.tsx
```diff
@@ -459,8 +459,15 @@ export const InputBar = ({
...
-    await saveDraftState(storageRepository, conversation, editorState, replyMessageEntity?.id, editedMessage?.id);
+  const saveDraft = async (editorState: string, plainMessage: string) => {
+    await saveDraftState(
+      storageRepository,
+      conversation,
+      editorState,
+      plainMessage,
+      replyMessageEntity?.id,
+      editedMessage?.id,
+    );
```
olafsulich: Could you refactor `saveDraftState` to take an object instead of those multiple params? It's error prone, hard to maintain

```suggestion
    await saveDraftState({
      storageRepository,
      conversation,
      editorState,
      plainMessage,
      replyMessageEntityId: replyMessageEntity?.id,
      editedMessageId: editedMessage?.id
      },
    );
```

src/script/components/InputBar/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -101,23 +101,18 @@ export const RichTextEditor = ({
...
-
+  const handleChange = (editorState: EditorState, editor: LexicalEditor) => {
     editorState.read(() => {
-      if (!editorRef.current) {
-        return;
-      }
-
       const markdown = $convertToMarkdownString(markdownTransformers);
 
+      saveDraftState(JSON.stringify(editorState.toJSON()), replaceEmojis ? findAndTransformEmoji(markdown) : markdown);
```
olafsulich: Let's add debounce for it

src/script/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const subscribe = (listener: () => void) => {
+    window.addEventListener('storage', listener);
+    return () => window.removeEventListener('storage', listener);
+  };
+
+  const store = useSyncExternalStore(subscribe, getSnapshot);
+
+  return [store, setLocalStorage] as const;
+};
```
olafsulich: Let's enhance types a little:

```suggestion
import { useSyncExternalStore } from 'react';

const parseJSON = <Value>(value: string | null): Value | null => {
    try {
      return value === null ? null : JSON.parse(value);
    } catch {
      console.error(`Error parsing JSON for key "${key}"`);
      return null;
    }
  };


export const useLocalStorage = <Value>(key: string, initialValue: Value) => {
  const setLocalStorage = (newValue: Value): void => {
    const serializedValue = JSON.stringify(newValue);
    window.localStorage.setItem(key, serializedValue);
    window.dispatchEvent(
      new StorageEvent('storage', { key, newValue: serializedValue })
    );
  };

  const getSnapshot = (): T | null => {
    const storedValue = localStorage.getItem(key);
    if (storedValue === null && initialValue !== undefined) {
      setLocalStorage(initialValue);
      return initialValue;
    }
    return parseJSON<T>(storedValue);
  };

  const subscribe = (listener: () => void): (() => void) => {
    window.addEventListener('storage', listener);
    return () => window.removeEventListener('storage', listener);
  };

  const store = useSyncExternalStore(subscribe, getSnapshot);

  return [store, setLocalStorage] as const;
};

// Usage examples:
type User = {
  name: string;
  age: number;
};

// With primitive types
const [count, setCount] = useLocalStorage<number>('count', 0);

// With complex types
const [user, setUser] = useLocalStorage<User>('user', { name: 'John', age: 30 });
```

src/script/components/InputBar/util/DraftStateUtil.ts
```diff
@@ -43,6 +45,7 @@ export const saveDraftState = async (
 
   await storageRepository.storageService.saveToSimpleStorage<any>(storageKey, {
```
olafsulich: Can we add a type here?

src/script/components/InputBar/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -71,7 +71,7 @@ interface RichTextEditorProps {
   showFormatToolbar: boolean;
   showMarkdownPreview: boolean;
   getMentionCandidates: (search?: string | null) => User[];
-  saveDraftState: (editor: string) => void;
+  saveDraftState: (editor: string, plainMessage: string) => void;
```
olafsulich, line 76: What about passing editorState as is, without stringifing it? The `saveToSimpleStorage` should stringify it automatically + can we make it an object?

```suggestion
  saveDraft: ({editorState, plainMessage}: {editorState: EditorState, plainMessage: string}) => void;
```

src/script/hooks/useLocalStorage.ts
przemvs: getSnapshot should be static one

src/script/components/InputBar/components/RichTextEditor/RichTextEditor.tsx
przemvs: I'll keep it like we have currently implemented for now :)

olafsulich (approved):

PR #18579, opened 2025-01-13
fix: Clean up connections when user joins team (WPB-15263)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15263" title="WPB-15263" target="_blank">BugWPB-15263</a>  [Web] Cannot create 1:1 conversation with personal user that migrated to my team after cancelling connection request
  </td></table>
  <br />
 

## Description

After the implementation of converting personal users to team users it turned out we need to clear any previous pending connection user A could possibly have with user B before user B joined user A's team. 

in this PR we use `user.update` event to find out when a user has joined our team then:

If the user is our self user -> we delete all pending connections old previous - connections that are not accepted - with the new team members we find.

If the user is someone else -> we try to find if we have a pending old connection with them and delete it. 

notes:
- A pending connection is a connection that's not accepted. for already accepted connections we will use the existing connection to interact with the newly joined to our team user.

### DIFF
--- a/src/script/connection/ConnectionRepository.test.ts
+++ b/src/script/connection/ConnectionRepository.test.ts
@@ -25,6 +25,8 @@ import {StatusCodes} from 'http-status-codes';
 
 import {WebAppEvents} from '@wireapp/webapp-events';
 
+import {SelfService} from 'src/script/self/SelfService';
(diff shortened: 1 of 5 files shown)


przemvs (approved):
lgtm

PR #18592, opened 2025-01-16
fix(BlockquotePlugin): block sending a message [WPB-12089]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-12089" title="WPB-12089" target="_blank">StoryWPB-12089</a>  [Web] Text formatting via UI in text input field
  </td></table>
  <br />
 

## Description

Fix BlockquotePlugin to not block sending a message, when user edits the blockquote

## Checklist

### DIFF
--- a/src/script/components/InputBar/components/RichTextEditor/plugins/BlockquotePlugin/BlockquotePlugin.tsx
+++ b/src/script/components/InputBar/components/RichTextEditor/plugins/BlockquotePlugin/BlockquotePlugin.tsx
@@ -73,13 +73,15 @@ const registerBlockquoteEnterCommand = (editor: LexicalEditor) => {
         return false;
       }
 
+      if (!event.shiftKey) {
+        return false;
+      }
+
       event.preventDefault();
 
-      if (event.shiftKey) {
-        editor.update(() => {
-          editor.dispatchCommand(INSERT_LINE_BREAK_COMMAND, false);
-        });
-      }
+      editor.update(() => {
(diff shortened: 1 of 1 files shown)


przemvs (approved):

V-Gira (approved):

src/script/components/InputBar/components/RichTextEditor/plugins/BlockquotePlugin/BlockquotePlugin.tsx
```diff
@@ -73,13 +73,15 @@ const registerBlockquoteEnterCommand = (editor: LexicalEditor) => {
         return false;
       }
 
+      if (!event.shiftKey) {
+        return false;
```
e-maad, line 77: There are multiple conditions above returning `false`, would be nice if we can use the same block

e-maad (approved):
LGTM 🚀 

Added one suggestion.

src/script/components/InputBar/components/RichTextEditor/plugins/BlockquotePlugin/BlockquotePlugin.tsx
olafsulich: I don't think it would improve readbility in this case. Above this condition you have that one:

```
if(!$isQuoteNode(quoteBlock) && !$isQuoteNode(anchorNode)) {...}
```

So adding another thing actually could cause more of [cognitive load](https://github.com/zakirullin/cognitive-load) than something good.

PR #18649, opened 2025-01-23
fix: Filter dead connection conversations (WPB-15377)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15377" title="WPB-15377" target="_blank">BugWPB-15377</a>  [Web] Name not available conversations appear for "dead" contact requests on fresh client
  </td></table>
  <br />
 

## Description

When a connection request is sent from user A to user B or vice verse a conversation is created for this connection.
If the connection request is in the not accepted state and user B joins user A's team we have to ignore this conversation.
 
Backend persists that conversation which was initially created when user B wasn't part of user A's team and they had a pending connection request.

This PR aims to filter and block that dead conversation.

### DIFF
--- a/src/script/conversation/ConversationRepository.ts
+++ b/src/script/conversation/ConversationRepository.ts
@@ -615,12 +615,15 @@ export class ConversationRepository {
    * Will load all the conversations in memory
    * @returns all the conversations from backend merged with the locally stored conversations and loaded into memory
    */
-  public async loadConversations(connections: ConnectionEntity[]): Promise<Conversation[]> {
+  public async loadConversations(
+    connections: ConnectionEntity[],
+    deadConnections: ConnectionEntity[],
+  ): Promise<Conversation[]> {
     const remoteConversations = await this.conversationService.getAllConversations().catch(error => {
       this.logger.error(`Failed to get all conversations from backend: ${error.message}`);
       return {found: []} as RemoteConversations;
     });
-    return this.loadRemoteConversations(remoteConversations, connections);
+    return this.loadRemoteConversations(remoteConversations, connections, deadConnections);
   }
 
   /**
@@ -722,6 +725,7 @@ export class ConversationRepository {
     localConversations: ConversationDatabaseData[],
     remoteConversations: RemoteConversations,
     connections: ConnectionEntity[],
+    deadConnections: ConnectionEntity[],
   ): Promise<{localConversations: ConversationDatabaseData[]; remoteConversations: RemoteConversations}> {
     //If there's any local conversation of type 3 (CONNECT), but the connection doesn't exist anymore (user was deleted),
     // we delete the conversation and blacklist it so it's never refetched from the backend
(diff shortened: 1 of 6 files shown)


src/script/conversation/ConversationRepository.ts
```diff
@@ -743,38 +747,60 @@ export class ConversationRepository {
...
-      deletedConnectionRequests.map(async ({id, domain, qualified_id}) => {
-        const qualifiedId = qualified_id || {id, domain};
-        await this.conversationService.deleteConversationFromDb(qualifiedId.id);
-        await this.blacklistConversation(qualifiedId);
-      }),
+    filteredRemoteConversations = filteredRemoteConversations?.filter(backendConversation => {
+      return !deadConnections.some(deadConnection =>
+        matchQualifiedIds(backendConversation.qualified_id, deadConnection.conversationId),
+      );
+    });
```
olafsulich: **suggestion:** I'm not a big fan of reassigning variables when it's not necessary. IMO, the more "pure" code is, the more predictable it becomes.

Here we don't reassigning the `filteredRemoteConversations` in other places, it's used only here. So, we don't even need to split it into two separate blocks, here's everything in the one `filter()`:

```ts
const filteredRemoteConversations = remoteConversations.found?.filter(remoteConversation => {
      const isNotDeletedConnection = !deletedConnectionRequests.find(({qualified_id, id, domain}) =>
        matchQualifiedIds(qualified_id || {id, domain}, remoteConversation.qualified_id),
      );

      const isNotDeadConnection = !deadConnections.some(deadConnection =>
        matchQualifiedIds(remoteConversation.qualified_id, deadConnection.conversationId),
      );

      return isNotDeletedConnection && isNotDeadConnection;
    });
```

src/script/conversation/ConversationRepository.ts
```diff
@@ -743,38 +747,60 @@ export class ConversationRepository {
...
+
+    await this.deleteAndBlockConversations(
+      deletedConnectionRequests.map(
+        deletedConnection =>
+          deletedConnection.qualified_id || {id: deletedConnection.id, domain: deletedConnection.domain ?? ''},
+      ),
     );
 
+    await this.deleteAndBlockConversations(deadConnections.map(deadConnection => deadConnection.conversationId));
```
olafsulich: **suggestion:** More like a personal prefence, but I think it's more clear to move mapping out of the function call, like here:

```suggestion
const deletedConnectionIds = deletedConnectionRequests.map(
      deletedConnection =>
        deletedConnection.qualified_id || {id: deletedConnection.id, domain: deletedConnection.domain ?? ''},
    );

    const deadConnectionIds = deadConnections.map(deadConnection => deadConnection.conversationId);

    await this.deleteAndBlockConversations(deletedConnectionIds);
    await this.deleteAndBlockConversations(deadConnectionIds);
```

src/script/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/18649/commits/bccda996dac5ffe36b1d746a8d815a0911e55f06

olafsulich (approved):

PR #18684, opened 2025-01-28
feat: update call_review_feedback countly event [WPB-14981]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-14981" title="WPB-14981" target="_blank">TaskWPB-14981</a>  [Web] Modify Segmentations for calling.call_quality_review
  </td></table>
  <br />
 

## Description
Updated call_review_feedback countly event and added a condition to ignore feedback if the call is too short (< 60 secs).

## Checklist

### DIFF
--- a/src/script/calling/Call.ts
+++ b/src/script/calling/Call.ts
@@ -46,7 +46,7 @@ interface ActiveSpeaker {
 export class Call {
   public readonly reason: ko.Observable<number | undefined> = ko.observable();
   public readonly startedAt: ko.Observable<number | undefined> = ko.observable();
-  public readonly endedAt: ko.Observable<number | undefined> = ko.observable();
+  public readonly endedAt: ko.Observable<number> = ko.observable(0);
   public readonly state: ko.Observable<CALL_STATE> = ko.observable(CALL_STATE.UNKNOWN);
   public readonly muteState: ko.Observable<MuteState> = ko.observable(MuteState.NOT_MUTED);
   public readonly participants: ko.ObservableArray<Participant>;
--- a/src/script/calling/Call.ts
+++ b/src/script/calling/Call.ts
@@ -46,6 +46,7 @@ interface ActiveSpeaker {
 export class Call {
   public readonly reason: ko.Observable<number | undefined> = ko.observable();
   public readonly startedAt: ko.Observable<number | undefined> = ko.observable();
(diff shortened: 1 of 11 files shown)


src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.tsx
```diff
@@ -45,17 +45,16 @@ import {
...
 import {UserState} from '../../../user/UserState';
 
 const logger = getLogger('CallQualityFeedback');
 
-export const QualityFeedbackModal = () => {
+export const QualityFeedbackModal = ({callingRepository}: {callingRepository: CallingRepository}) => {
   const userState = container.resolve(UserState);
-
+  const {conversationId} = useCallAlertState();
+  const call = conversationId && callingRepository.findCall(conversationId);
```
przemvs, line 59: Maybe You can check if there is no conversationId do not render component? 🤔

src/script/calling/Call.ts
```diff
@@ -46,6 +46,7 @@ interface ActiveSpeaker {
 export class Call {
   public readonly reason: ko.Observable<number | undefined> = ko.observable();
   public readonly startedAt: ko.Observable<number | undefined> = ko.observable();
+  public readonly endedAt: ko.Observable<number | undefined> = ko.observable();
```
przemvs: I think You can add default value in `ko.observable(0)` and then You don't have to check if this is a undefined and remove check `endedAt() || 0` ? WDYT?

src/script/components/Modals/QualityFeedbackModal/QualityFeedbackModal.tsx
```diff
@@ -45,17 +45,16 @@ import {
...
 } from './QualityFeedbackModal.styles';
 
-import {EventName} from '../../../tracking/EventName';
-import {Segmentation} from '../../../tracking/Segmentation';
 import {UserState} from '../../../user/UserState';
 
 const logger = getLogger('CallQualityFeedback');
 
-export const QualityFeedbackModal = () => {
+export const QualityFeedbackModal = ({callingRepository}: {callingRepository: CallingRepository}) => {
```
przemvs: And maybe create interface Props and pass there instead doing like that :) Will be more readable :)

przemvs (approved):
LGTM! :)

PR #18685, opened 2025-01-28
feat: Save draft message reply

przemvs:
## Description

While user click on reply message (without writing text) and change tab, reply draft message is not properly saved.
Also while user want to remove reply message box, and change tab then reply draft message is present after change conversation.

## Checklist

### DIFF
--- a/src/script/components/InputBar/InputBar.tsx
+++ b/src/script/components/InputBar/InputBar.tsx
@@ -19,9 +19,10 @@
 
 import {useCallback, useEffect, useRef, useState} from 'react';
 
+import {$convertToMarkdownString} from '@lexical/markdown';
 import {amplify} from 'amplify';
 import cx from 'classnames';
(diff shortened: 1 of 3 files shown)


src/script/components/InputBar/InputBar.tsx
```diff
@@ -214,6 +217,23 @@ export const InputBar = ({
     ),
   });
 
+  const handleSaveEditorDraft = (replyId = '') => {
```
olafsulich, line 220: Shouldn't it be like that? (looking at other usage of `replyId`):

```suggestion
  const handleSaveEditorDraft = (replyId?: string) => {
```

przemvs: We have to pass empty string for proper update of localStorage :( So I do it like that

olafsulich (approved):

PR #18688, opened 2025-01-28
fix: Resolve virtual rendering focus issues (WPB-15154)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15154" title="WPB-15154" target="_blank">TaskWPB-15154</a>  [Web] Improve loading of conversation list by only loading what is visible in viewport
  </td></table>
  <br />
 

## Description

- changed align behavior of scrollToIndex of rowVirtualizer to auto to avoid always bringing the conversation in center even in cases they are already visible
- removed useEffect for focusing as it was causing unexpected focus jumping behavior and it had no use case when opening a folder

### DIFF
--- a/src/script/components/ConversationListCell/ConversationListCell.tsx
+++ b/src/script/components/ConversationListCell/ConversationListCell.tsx
@@ -17,13 +17,7 @@
  *
  */
 
-import React, {
-  useEffect,
-  useRef,
-  useState,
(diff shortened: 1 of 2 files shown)


przemvs (approved):
LGTM!

olafsulich (approved):

PR #18693, opened 2025-01-30
fix: margin in message details

aweiss-dev:
## Description

Change 
Screenshot 2025-01-30 at 14 31 13

to 
Screenshot 2025-01-30 at 14 53 09

### DIFF
--- a/src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
+++ b/src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
@@ -151,7 +151,7 @@ const MessageDetails: FC<MessageDetailsProps> = ({
   const onParticipantClick = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
   return (
-    <div id="message-details" className="panel__page message-details">
+    <div id="message-details" className="panel__page panel__message-details">
       <PanelHeader
         onClose={onClose}
         title={panelTitle}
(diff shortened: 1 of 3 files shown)


przemvs (approved):

V-Gira (approved):

src/script/page/RightSidebar/MessageDetails/MessageDetails.tsx
```diff
@@ -151,7 +151,7 @@ const MessageDetails: FC<MessageDetailsProps> = ({
   const onParticipantClick = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
   return (
-    <div id="message-details" className="panel__page message-details">
+    <div id="message-details" className="panel__page panel__message-details">
```
dariaoldenburg, line 154: I think I'd prefer to stick with the message-details class here, because it's the name of the component, and we don't use it outside of that component (except for my mistake). From my perspective, it would be better to change the name of that class inside FailedToAddUsersMessage

dariaoldenburg (changes requested):
I think I'd prefer to stick with the message-details class here, because it's the name of the component, and we don't use it outside of that component (except for my mistake). From my perspective, it would be better to change the name of that class inside FailedToAddUsersMessage

dariaoldenburg (commented):
:)

e-maad (approved):

thisisamir98 (approved):

PR #18698, opened 2025-01-31
fix(applock): prevent unintentional logout when applock is active [15545]

V-Gira:
## Description

The app would log out on app initialization for team users if app lock is active, without giving them the chance to disable app lock

There is an observer that makes sure the app or the modal is not tempered with to prevent people to simply removing the app lock from the DOM

It appears the observer is too aggressive, and would logout a user automatically

## Checklist

### DIFF
--- a/src/script/page/AppLock/AppLock.tsx
+++ b/src/script/page/AppLock/AppLock.tsx
@@ -84,6 +84,8 @@ const AppLock: React.FC<AppLockProps> = ({
     'isAppLockEnforced',
   ]);
 
+  // We log the user out if there is a style change on the app element
+  // i.e. if there is an attempt to remove the blur effect
   const {current: appObserver} = useRef(
     new MutationObserver(mutationRecords => {
       const [{attributeName}] = mutationRecords;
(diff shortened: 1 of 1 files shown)


src/script/page/AppLock/AppLock.tsx
```diff
@@ -156,11 +159,16 @@ const AppLock: React.FC<AppLockProps> = ({
     app?.style.setProperty('pointer-events', isVisible ? 'none' : 'auto', 'important');
 
     if (isVisible) {
-      modalObserver.observe(document.querySelector('#wire-main'), {
-        childList: true,
-        subtree: true,
```
V-Gira, line 161: I was unable to pinpoint the exact cause, but this observer was checking for both `childList` and `subtree` for a removal attempt of the modal.
It doesn't seem to be necessary to include the subtree with our current code structure, as the app correctly logs out when removing the app lock modal from the DOM

aweiss-dev (approved):

olafsulich (approved):

PR #18769, opened 2025-02-17
feat: bump to AVS 10.0.7 [WPB-16177]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16177" title="WPB-16177" target="_blank">StoryWPB-16177</a>  |Web] Bump AVS to 10.0.7
  </td></table>
  <br />
 

## Description

In this PR, we are upgrading the AVS version to 10.0.7. Version 10.0.6 does not contain any features and was skipped for technical reasons.

The change is that AVS now includes a browser switch.

Firefox will always send the high resolution first, while Chrome will always send the lower resolution first. Additionally, Firefox will only send the high resolution during screen sharing. The reason for this is that Firefox has only properly supported simulcast for screen sharing since version 135 (as of January 2025). However, since it is not yet perfect, we are temporarily disabling it.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "@lexical/rich-text": "0.24.0",
     "@mediapipe/tasks-vision": "0.10.21",
     "@tanstack/react-virtual": "^3.13.0",
-    "@wireapp/avs": "10.0.5",
+    "@wireapp/avs": "10.0.7",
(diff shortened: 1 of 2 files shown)


EnricoSchw:
I close this PR because we revert simulcast from dev and working in a feature branch.

PR #18774, opened 2025-02-20
feat(InputBar): files dropzone draft [WPB-16106]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16106" title="WPB-16106" target="_blank">TaskWPB-16106</a>  [Web] Draft message with multiple files - UI
  </td></table>
  <br />
 

## Description

**About the change:**

- it's a first iteration, only from the UI side (no logic)
- a user would be able to upload multiple files into the input
- it uses a new library, [react-dropzone](https://react-dropzone.js.org/), which is a nice addition and allows me to easily upload multiple files & display their previews
- it's not working with the "files" icon button yet, but I'll in the future

**Animation library:**

- introducing a new library for animation handling [AutoAnimate](https://auto-animate.formkit.com/), it's trivial to use
- with the animations, the deleting feature is looking way nicer
- we can use the library to animate other parts of the app, like the conversation list, so after archiving a conversation, the effect will be smoother
- didn't want to introduce a bigger library like Framer Motion, we don't need it at the moment, so AutoAnimate sounds like a fair alternative

## Demo

**Uploading and deleting multiple files:**

https://github.com/user-attachments/assets/db5b8cbf-f8b0-43c6-8497-6818e1926ffa

**[Error] Too many files:**

https://github.com/user-attachments/assets/43f8292f-7e14-4321-86d4-70a8b48536d7

**[Error] Images too large:**

https://github.com/user-attachments/assets/0c10e4ca-e4a1-4225-8f9c-458488a955b0

**[Error] Files and/or images too large:**

https://github.com/user-attachments/assets/470dd83f-0269-400f-bc22-a193e02b3ebc

## Checklist


src/i18n/en-US.json
```diff
@@ -489,7 +489,19 @@
...
+  "conversationFileVideoPreviewLabel": "Video file preview for: {src}",
   "conversationFileAssetRestricted": "Receiving files is prohibited",
+  "conversationFileUploadFailedHeading": "File upload failed",
+  "conversationFileUploadFailedMessage": "Error occurred in the upload. Please try again.",
+  "conversationFileUploadFailedTooManyFilesHeading": "Too many files",
+  "conversationFileUploadFailedTooManyFilesMessage": "Please select a maximum of {maxFiles} files.",
+  "conversationFileUploadFailedTooLargeFilesHeading": "Files too large",
+  "conversationFileUploadFailedTooLargeFilesMessage": "Please select files smaller than {maxSize}MB.",
+  "conversationFileUploadFailedTooLargeImagesMessage": "Please select images smaller than {maxSize}MB.",
+  "conversationFileUploadFailedTooLargeFilesAndImagesMessage": "Please select files smaller than {maxImageSize}MB and images smaller than {maxFileSize}MB.",
```
przemvs, line 503: Maybe we can also provide unit ? MB/GB/ etc? WDYT ?

src/script/components/Conversation/useFiles/useFiles.ts
```diff
@@ -0,0 +1,43 @@
...
+  files: FileWithPreview[];
+  addFiles: (files: FileWithPreview[]) => void;
+  deleteFile: (fileId: string) => void;
+}
+
+export const useFileUploadState = create<FileUploadState>(set => ({
+  files: [],
+  addFiles: newFiles =>
+    set(state => ({
+      files: [...newFiles, ...state.files],
```
przemvs, line 37: Shouldn't be ...state.files first? Because it will override newFiles if will be the same :)

src/script/components/InputBar/FilePreviews/VideoPreviewCard/VideoPreviewCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import {FilePreviewDeleteButton} from '../common/FilePreviewDeleteButton/FilePreviewDeleteButton';
+import {FilePreviewPlayButton} from '../common/FilePreviewPlayButton/FilePreviewPlayButton';
+interface VideoPreviewCardProps {
+  src: string;
+  onDelete: () => void;
+}
+
+export const VideoPreviewCard = ({src, onDelete}: VideoPreviewCardProps) => {
+  const videoRef = useRef<HTMLVideoElement>(null);
```
przemvs: Hmm.. You need this videoRef there ? :) It's not used anywhere. Or this is a trick? :D

olafsulich: Oh right, it shouldn't be there!

src/script/components/Conversation/useFiles/useFiles.ts
olafsulich: The case was to have new uploaded files first, on top of the list - then we can see there's a new file

src/i18n/en-US.json
olafsulich: Currently we don't allow big files, like in GB (max is 25/50 MB, depends if user is in a team), so I would leave it like that 🤷‍♂️ If we would have a need, then sure

przemvs (approved):

PR #18797, opened 2025-02-25
feat(TeamRepository): Prevent removal of MLS from supported protocols if it was previously supported

przemvs:
https://wearezeta.atlassian.net/browse/WPB-16255

## Description
Prevent removal of MLS from supported protocols if it was previously supported

## Checklist


src/script/team/TeamRepository.ts
```diff
@@ -134,6 +142,17 @@ export class TeamRepository extends TypedEventEmitter<Events> {
 
     this.teamState.teamFeatures(newFeatureList);
 
+    if (newFeatureList[FEATURE_KEY.MLS]?.config?.supportedProtocols?.includes(ConversationProtocol.MLS)) {
+      this.updatePersistedSupportedProtocols();
```
mathias-niboulies, line 146: Not sure I'm reading it well, but this should be based on the client capability to support MLS, not the configuration of the team

typfel:
I think there's a confusion between the **team**SupportedProtocols (feature config) and the **user**SupportedProtocols (self fuser). Here you seem to prevent removing MLS from the `teamSupportedProtocols` but we should actually prevent removing MLS from the `userSupportedProtocols` which are updated by calling `PUT /self/supported-protocols`.

thisisamir98:
@typfel addressed your concern in here: https://github.com/wireapp/wire-webapp/pull/18797/commits/c1c6030ba5cff2749011a4e71407c1516d0380b6

typfel (approved):

PR #18805, opened 2025-02-27
fix: clicking on pagination dots to switch pages [WPB-16183]

waseemansar:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16183" title="WPB-16183" target="_blank">BugWPB-16183</a>  [Web] Popped out calling view when resized small, clicking on bubble to change page should not do anything
  </td></table>
  <br />
 

## Description
Fixes an issue where clicking on pagination dot buttons had no effect, preventing users from switching pages.

### Before
https://github.com/user-attachments/assets/c44c1336-4e80-4c08-bfd5-8086d9afe2a7

### After
https://github.com/user-attachments/assets/74b72ecb-4a71-45f5-9f43-8e70fc26fd33

## Checklist

### DIFF
--- a/src/script/components/calling/Pagination/PaginationArrow.tsx
+++ b/src/script/components/calling/Pagination/PaginationArrow.tsx
@@ -20,6 +20,7 @@
 import {ChevronIcon, IconButton, IconButtonVariant} from '@wireapp/react-ui-kit';
 
 import {handleKeyDown, KEY} from 'Util/KeyboardUtil';
+import {t} from 'Util/LocalizerUtil';
 
 import {chevronLeftStyles, chevronRightStyles, iconButtonStyles} from './Pagination.styles';
 
@@ -30,7 +31,9 @@ interface PaginationArrowProps {
   'data-uie-name': string;
 }
 
-export function PaginationArrow({onClick, disabled, direction, 'data-uie-name': uieName}: PaginationArrowProps) {
+export const PaginationArrow = ({onClick, disabled, direction, 'data-uie-name': uieName}: PaginationArrowProps) => {
+  const ariaLabel = direction === 'left' ? t('paginationLeftArrowAriaLabel') : t('paginationRightArrowAriaLabel');
+
   return (
     <IconButton
       variant={IconButtonVariant.SECONDARY}
@@ -46,8 +49,9 @@ export function PaginationArrow({onClick, disabled, direction, 'data-uie-name': 
       disabled={disabled}
       data-uie-name={uieName}
       type="button"
+      aria-label={ariaLabel}
     >
-      <ChevronIcon css={direction === 'left' ? chevronLeftStyles : chevronRightStyles} />
+      <ChevronIcon css={direction === 'left' ? chevronLeftStyles : chevronRightStyles} aria-hidden="true" />
     </IconButton>
   );
-}
+};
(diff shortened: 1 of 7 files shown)


src/script/components/calling/Pagination/PaginationDot.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface PaginationDotProps {
+  page: number;
+  isCurrentPage: boolean;
+  isSmaller: boolean;
+  onClick: (page: number) => void;
+}
+
+export function PaginationDot({page, isCurrentPage, isSmaller, onClick}: PaginationDotProps) {
+  return (
+    <div
```
olafsulich: It should be button not div, imo

olafsulich: Missing some "accessible label" like "aria-label" - without it, users who use screen readers wouldn't be able to know what they're clicking - I think it wasn't there before, but would be awesome if we add it if we change something here

src/script/components/calling/Pagination/PaginationArrow.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface PaginationArrowProps {
+  onClick: () => void;
+  disabled?: boolean;
+  direction: 'left' | 'right';
+  'data-uie-name': string;
+}
+
+export function PaginationArrow({onClick, disabled, direction, 'data-uie-name': uieName}: PaginationArrowProps) {
+  return (
+    <IconButton
```
olafsulich, line 38: Missing some "accessible label" like "aria-label" - I think it wasn't there before, but would be awesome if we add it if we change something here

src/script/components/calling/Pagination/PaginationArrow.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {chevronLeftStyles, chevronRightStyles, iconButtonStyles} from './Pagination.styles';
+
+interface PaginationArrowProps {
+  onClick: () => void;
+  disabled?: boolean;
+  direction: 'left' | 'right';
+  'data-uie-name': string;
+}
+
+export function PaginationArrow({onClick, disabled, direction, 'data-uie-name': uieName}: PaginationArrowProps) {
```
olafsulich: Not sure if it's a strict rule (probably not), but I've seen that for new components we use arrow functions instead

src/script/components/calling/Pagination/PaginationDot.tsx
```diff
@@ -0,0 +1,54 @@
...
+import {dotStyles, dotWrapperStyles} from './Pagination.styles';
+
+interface PaginationDotProps {
+  page: number;
+  isCurrentPage: boolean;
+  isSmaller: boolean;
+  onClick: (page: number) => void;
+}
+
+export function PaginationDot({page, isCurrentPage, isSmaller, onClick}: PaginationDotProps) {
```
olafsulich: same here, normal fn => arrow fn

src/script/components/calling/Pagination/PaginationDot.tsx
waseemansar: changed to button

waseemansar: added "aria-label"

olafsulich (approved):

PR #18829, opened 2025-03-05
feat: update group avatars, select from 15 options [WPB-15794]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15794" title="WPB-15794" target="_blank">StoryWPB-15794</a>  [Web] Adjust group avatars (not channels) to be in line with channel mocks
  </td></table>
  <br />
 

## Description

Updated group avatars, select from 15 options of colors set.

### Conversation List
Screenshot 2025-03-12 at 10 20 07

### Calling header
Screenshot 2025-03-12 at 10 23 21

### Connection Requests
Screenshot 2025-03-12 at 10 24 05

## Checklist


przemvs (approved):

src/script/util/avatarUtil.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ *
+ * @param id - unique id
+ * @param theme - default or dark
+ * @returns - array of colors for group avatar
+ */
+export function getGroupAvatarColors(
+  id = '',
+  theme: typeof PropertiesRepository.prototype.properties.settings.interface.theme,
```
V-Gira: we have the accent-colors available as css variable that changes depending on theme, for example `var(--amber-500` is #a25915 in light mode and #ffd426 in dark mode.

Seems like we could use those instead of relying on the property repository to get access to the theme, WDYT?

src/script/components/Avatar/GroupAvatar.tsx
```diff
@@ -52,22 +51,27 @@ export const GroupAvatar: React.FC<GroupAvatarProps> = ({users, className}) => {
...
-              color: user.accent_color(),
-              flex: '0 0 auto',
-              fontSize: 9,
-              fontWeight: 900,
-            }}
-          >
-            {Array.from(user.initials())[0]}
-          </div>
-        ))}
+        <svg
```
olafsulich, line 54: Looks like the border color is different on designs:

Screenshot 2025-03-12 at 09 44 16

olafsulich: plus border radius should be 8px, it looks smaller on the screenshot

olafsulich (approved):

V-Gira (approved):

PR #18900, opened 2025-03-24
fix: allow personal users to select mentions and replies notification setting [WPB-16650]

waseemansar:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16650" title="WPB-16650" target="_blank">BugWPB-16650</a>  [Web] cannot select "Mentions and replies" as notification setting for group as personal user
  </td></table>
  <br />
 

## Description
Fixes the issue where personal users cannot select "Mentions and Replies" notification setting in group conversations. When attempting to select this option, it defaults to "Nothing". This was previously a deliberate restriction in the codebase that limited this feature to team users only.

### Before

https://github.com/user-attachments/assets/ac323c18-8d30-4493-967d-fe8b00683104

### After

https://github.com/user-attachments/assets/b904082d-8a9a-41c9-aed5-28aa34f54db1

## Checklist

### DIFF
--- a/src/script/entity/Conversation.ts
+++ b/src/script/entity/Conversation.ts
@@ -315,22 +315,11 @@ export class Conversation {
       }
       const mutedState = this.mutedState();
 
-      const knownNotificationStates = Object.values(NOTIFICATION_STATE);
-      if (knownNotificationStates.includes(mutedState)) {
-        const isStateMentionsAndReplies = mutedState === NOTIFICATION_STATE.MENTIONS_AND_REPLIES;
-        const isInvalidState = isStateMentionsAndReplies && !this.selfUser()?.teamId;
-
-        return isInvalidState ? NOTIFICATION_STATE.NOTHING : mutedState;
-      }
-
       if (typeof mutedState === 'boolean') {
-        const migratedMutedState = !!this.selfUser()?.teamId
-          ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES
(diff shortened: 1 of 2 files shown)


src/script/entity/Conversation.ts
```diff
@@ -315,22 +315,11 @@ export class Conversation {
...
       if (typeof mutedState === 'boolean') {
-        const migratedMutedState = !!this.selfUser()?.teamId
-          ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES
-          : NOTIFICATION_STATE.NOTHING;
-        return this.mutedState() ? migratedMutedState : NOTIFICATION_STATE.EVERYTHING;
+        return mutedState ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES : NOTIFICATION_STATE.EVERYTHING;
       }
 
-      return NOTIFICATION_STATE.EVERYTHING;
+      return mutedState;
```
V-Gira, line 322: surprisingly confusing to review (monday morning?)
So we remove all the logic tied to the muted state (blocking personal users to select "mentions and replies), we only keep one bit, presumably backward compability with an older backend API where mutedState is a boolean instead of an enum, correct?

waseemansar: @V-Gira yes correct only keeping backward compatibility layer where `mutedState` is boolean.

V-Gira (approved):

PR #18910, opened 2025-03-26
feat(cells): increase table files limit [WPB-15679]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-15679" title="WPB-15679" target="_blank">TaskWPB-15679</a>  [Web] Conversation “files” tab view
  </td></table>
  <br />
 

## Description

Increase table files limit, implements https://github.com/wireapp/wire-web-packages/pull/6990 changes

## Checklist


src/script/cells/CellsRepository.ts
```diff
@@ -80,8 +80,8 @@ export class CellsRepository {
     return this.apiClient.api.cells.deleteFile({uuid});
   }
 
-  async getAllFiles({path}: {path: string}) {
-    return this.apiClient.api.cells.getAllFiles({path: path || this.basePath});
+  async getAllFiles({path, limit}: {path: string; limit?: number}) {
```
przemvs: Maybe You can create variable in CellRepository with MAX_FILES_LIMIT instead of doing in hooks :)

olafsulich: hmm, i think I'd rather do it in the place where I use it, in the future we'll add pagination and other stuff, so I will have to be configurable either way

przemvs: Sure, but default limit is 100. So better to pass there default limit, and then if You'll add pagination You'll extend functionality :)

olafsulich: The default is actually 10 (in the api-client package), this 100 will change, it's only for demo purposes

przemvs (approved):

PR #18942, opened 2025-04-01
feat: implement screen sharing with camera overlay POC [WPB-16186]

waseemansar:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16186" title="WPB-16186" target="_blank">SpikeWPB-16186</a>  [Web] Prototype simultaneous screenshare and camera [timebox: 3 days]
  </td></table>
  <br />
 

## Description

This PR implements a prototype/ POC for screen sharing with camera overlay functionality, including Picture-in-Picture (PiP) support.

- Screen sharing with camera overlay using Document PiP API.
- Feature flag to enable/disable screen sharing with video.
- Position tracking for PiP window movement.

https://github.com/user-attachments/assets/d4504e8c-0635-4b84-8b06-5c625eee6923

## Checklist


src/script/calling/CallingRepository.ts
```diff
@@ -159,6 +160,8 @@ export class CallingRepository {
   private wUser: number = 0;
   private nextMuteState: MuteState = MuteState.SELF_MUTED;
   private isConferenceCallingSupported = false;
+  private readonly canvasMixer: CanvasMediaStreamMixer;
```
EnricoSchw: This is a theoretical note: Maybe the canvas should be assigned to the Call object. If you assign the canvas to the Repo, there can only be one per app. However, there can be multiple Calls happening simultaneously. I’m not sure if it’s possible to have two different screen shares in two concurrent Calls. If it is possible, there is a risk that the streams might get mixed up.

src/script/calling/CallingRepository.ts
```diff
@@ -1075,25 +1091,61 @@ export class CallingRepository {
...
+
+    if (selfParticipant.sharesScreen()) {
+      this.stopScreenShare(selfParticipant, conversation, call);
+      return;
+    }
+
+    try {
+      // Get both screen and camera streams
+      const screenStream = await this.getMediaStream({audio: true, screen: true}, call.isGroupOrConference);
+      const cameraStream = await this.getMediaStream({camera: true}, call.isGroupOrConference);
```
EnricoSchw: Theoretically, the current camera stream should be closed… so restarting the camera shouldn’t be a problem, right?

src/script/calling/CallingRepository.ts
```diff
@@ -1075,25 +1091,61 @@ export class CallingRepository {
...
+
+      // Initialize mixer and get mixed stream
+      const mixedStream = await this.canvasMixer.startMixing(screenStream, cameraStream);
+
+      // If the screen share is stopped by the os system or the browser, an "ended" event is triggered. We listen for
+      // this event to clean up the screen share state in this case.
+      screenStream.getVideoTracks()[0].onended = () => {
```
EnricoSchw: In that case, you also need to close the camera video track. With  _stopScreenShare_  you only finish the mixed video stream.

src/script/calling/CallingRepository.ts
```diff
@@ -1465,7 +1517,14 @@ export class CallingRepository {
 
   readonly leaveCall = (conversationId: QualifiedId, reason: LEAVE_CALL_REASON): void => {
     const call = this.findCall(conversationId);
-    call?.endedAt(Date.now());
+    if (call) {
+      call.endedAt(Date.now());
+      // Stop screen sharing if active
+      if (call.getSelfParticipant().sharesScreen()) {
+        this.stopScreenShare(call.getSelfParticipant(), call.conversation, call);
```
EnricoSchw, line 1564: Be careful. In the mixed case, you have three running video tracks plus one audio track. Make sure all of them are properly closed and handled.

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,317 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+// Canvas configuration
```
EnricoSchw, line 20: This you can not know, but at AVS, we put a lot of effort into transmitting screen shares in the best possible quality. Here You’re drawing a canvas with fixed dimensions. That means if someone captures a screen with 4000 x 2000 dimensions, you’re scaling it down. This results in a loss of resolution.

I know that on the receiver side, the screen share is also scaled down to fit the dimensions (one of the reasons why we don’t have 100% best screen share quality). Nevertheless, we should avoid downscaling on the sender side.

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,317 @@
...
+const CANVAS_WIDTH = 1920;
+const CANVAS_HEIGHT = 1080;
+const FRAME_RATE = 30;
```
EnricoSchw, line 24: Keep in mind that we capture screen shares at only 5 FPS… but with this approach, you’re rendering as many frames as needed.

The video in the top corner will be very choppy… but 5 FPS is the trade-off that allows us to transmit Ultra HD screen shares. (On the other end, not much "Ultra" will actually arrive—but without this trade-off, it would be even worse.)

src/script/calling/CallingRepository.ts
EnricoSchw: Okay, I see that you're ending the tracks in the canvas.

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,317 @@
...
+      this.animationFrame = null;
+    }
+
+    if (window.documentPictureInPicture?.window) {
+      window.documentPictureInPicture.window.close();
+    }
+
+    [this.screenVideo, this.cameraVideo].forEach(video => {
+      if (video) {
+        const tracks = video.srcObject as MediaStream;
```
EnricoSchw, line 297: Theoretically that's fine, because the tracks are singletons... but it's cleaner to keep the track reference and end it and then remove the HTML element. but ok.

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  async startMixing(screenShare: MediaStream, camera: MediaStream): Promise<MediaStream> {
+    try {
+      this.screenVideo = this.createVideoElement(screenShare, {
+        muted: true,
+        autoplay: true,
+      });
+
+      // Handle screen share track ending
+      screenShare.getVideoTracks().forEach(track => {
```
EnricoSchw: Ok, you now treat the event twice, we should only treat it once.

EnricoSchw (commented):
Your code is very clean and easy to read, reviewing it is fun. :) I hope there aren’t too many comments from me.

src/script/calling/CallingRepository.ts
waseemansar: Thank you for raising this important architectural point. You're absolutely right - having the `canvasMixer` as a repository-level property could potentially cause issues with concurrent calls.

Currently, the `canvasMixer` is initialized in the `CallingRepository` constructor:
```typescript
private readonly canvasMixer: CanvasMediaStreamMixer;

constructor(...) {
  this.canvasMixer = new CanvasMediaStreamMixer();
}
```

This means there's only one mixer instance for all calls in the application, which could indeed lead to issues if multiple calls are sharing screens simultaneously.

We should move the `canvasMixer` to be a property of the `Call` class instead. This would:
1. Ensure each call has its own dedicated mixer
2. Prevent any potential stream mixing between different calls
3. Better encapsulate the screen sharing functionality within the call context
4. Make the code more maintainable and safer for concurrent calls

src/script/media/CanvasMediaStreamMixer.ts
waseemansar: agreed. will avoid downscaling.

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,316 @@
...
+            this.smallOffsetX += (this.targetOffsetX - this.smallOffsetX) * POSITION_SMOOTHING;
+            this.smallOffsetY += (this.targetOffsetY - this.smallOffsetY) * POSITION_SMOOTHING;
+          }
+
+          // Draw camera overlay
+          if (this.cameraVideo.readyState >= this.cameraVideo.HAVE_CURRENT_DATA) {
+            const out_h = this.canvas.height / CAMERA_OVERLAY_SCALE;
+            const out_w = (this.cameraVideo.videoWidth / this.cameraVideo.videoHeight) * out_h;
+
+            const x = this.isPipActive
```
przemvs: Could we name it offsetX ? Will be more readable than x :)

src/script/media/CanvasMediaStreamMixer.ts
```diff
@@ -0,0 +1,316 @@
...
+              ? Math.round(this.canvas.width - out_w + this.smallOffsetX)
+              : Math.round(this.canvas.width - out_w - CAMERA_OVERLAY_PADDING);
+
+            const y = this.isPipActive ? Math.round(this.smallOffsetY) : CAMERA_OVERLAY_PADDING;
```
przemvs: Could we name it offsetY ? Will be more readable than y :)

przemvs (approved):
LGTM! :) 2 comments only ;)

PR #18944, opened 2025-04-02
feat: simulcast

EnricoSchw:
## Description

This pull request brings Adaptive Video Streaming back into the main development branch. After being temporarily removed in favor of MLS, it's now time to reintroduce it.

Adaptive Video Streaming improves the adaptation to network throughput conditions and enhances our screen share quality in group chats.

This allows us to remove the distinction between 1:1 calls and group calls for screen sharing and define them uniformly. See 'src/script/media/MediaConstraintsHandler.ts'.

### Bandwidth Usage

**Tile view**
In tile mode, the received video data is reduced to a quarter.

**Fullscreen**
When switching to full screen mode, only the full screen video data is received. In the past, video streams that were not being displayed were still received in the background.
**Note:** The new scaling feature, which adjusts the number of videos based on screen size, may not yet be included in the optimization.

**Sending**
The required upload bandwidth increases by approximately a factor of 1.3. This is necessary because two resolutions are now being transmitted.

**Adaptivity**
_sending client_
If the sending client has a poor connection, the transmission of the high resolution is automatically disabled.
If the internet connection is extremely poor, the low resolution stream will also be disabled.
However, in this case, the conference would no longer be realistically usable.

_receiving client_
If the internet connection is very poor, the client automatically switches to the low resolution video stream.
In the current version, the client cannot recover on its own when the connection improves.

### Video Resolution
**Video View:**
In the tile view, streams are now displayed with a maximum resolution of 320x180 pixels.
In full screen mode, videos are displayed at a maximum resolution of 1280x720 pixels.

**Screen Shares:**
The resolution of screen shares is based on the root stream.
In tile view, screen shares are shown at a quarter of their original resolution.
In full screen mode, screen shares are displayed in their original resolution.
**Note**: The web client still scales down screen shares using CSS. For example, 4000-pixel screen shares lose resolution due to this downscaling.

**Optimisations**
Videos are now specifically optimized for motion, resulting in less pixelation.
Screen shares are specifically optimized for resolution.

**Note**: The new screen share prototype (video within screen share) is not optimized in this way.

### Changes in Video View Behavior

When the app GUI switches between video pages or full screen mode, the last frames of the previous video were displayed for 1 second.
However, since Simulcast now causes many more transitions and this became quite noticeable, we have added a black overlay view during the switch.

### Backward compatibility

The AVS API itself is not backward compatible. However, we have ensured that clients using AVS 9 are compatible with clients using AVS 10.
We have achieved the same level of compatibility between SFT 4 and SFT 5.

**Feature Flag**
Adaptive Video Streaming (Simulcast) is enabled or disabled via the SFT.
Only SFT 5 activates Simulcast for an AVS 10 client.

## Checklist


src/script/calling/CallingRepository.ts
```diff
@@ -256,14 +258,47 @@ export class CallingRepository {
...
       if (!call) {
         return;
       }
       const isSpeakersViewActive = this.callState.isSpeakersViewActive();
       if (isSpeakersViewActive) {
-        this.requestVideoStreams(call.conversation.qualifiedId, call.activeSpeakers());
+        const videoQuality = call.activeSpeakers().length > 2 ? RESOLUTION.LOW : RESOLUTION.HIGH;
+
+        const speakes = call.activeSpeakers();
+        speakes.forEach(p => {
```
przemvs: Could you give different var name for p ? :) Will be more readable ;) Try to give min 3 chars to variable :)

src/script/media/MediaConstraintsHandler.ts
```diff
@@ -153,14 +130,12 @@ export class MediaConstraintsHandler {
     };
   }
 
-  getScreenStreamConstraints(method: ScreensharingMethods, isGroup: boolean): MediaStreamConstraints | undefined {
+  getScreenStreamConstraints(method: ScreensharingMethods): MediaStreamConstraints | undefined {
     switch (method) {
       case ScreensharingMethods.DESKTOP_CAPTURER:
-        this.logger.info(`Enabling screen sharing from desktopCapturer (with fULL resolution: ${isGroup})`);
+        this.logger.info(`Enabling screen sharing from desktopCapturer (with fULL resolution)`);
```
przemvs:
```suggestion
        this.logger.info(`Enabling screen sharing from desktopCapturer (with FULL resolution)`);
```

src/script/media/MediaConstraintsHandler.ts
```diff
@@ -172,22 +147,18 @@ export class MediaConstraintsHandler {
 
         return streamConstraints;
       case ScreensharingMethods.DISPLAY_MEDIA:
-        this.logger.info(`Enabling screen sharing from getDisplayMedia (with fULL resolution: ${isGroup})`);
+        this.logger.info(`Enabling screen sharing from getDisplayMedia (with fULL resolution)`);
```
przemvs:
```suggestion
        this.logger.info(`Enabling screen sharing from getDisplayMedia (with FULL resolution)`);
```

src/script/media/MediaConstraintsHandler.ts
```diff
@@ -172,22 +147,18 @@ export class MediaConstraintsHandler {
...
-          : MediaConstraintsHandler.CONFIG.CONSTRAINTS.SCREEN.DISPLAY_MEDIA_FULL;
+        const display = MediaConstraintsHandler.CONFIG.CONSTRAINTS.SCREEN.DISPLAY_MEDIA;
 
         return {
           audio: false,
           video: display,
         };
       case ScreensharingMethods.USER_MEDIA:
-        this.logger.info(`Enabling screen sharing from getUserMedia (with fULL resolution: ${isGroup})`);
+        this.logger.info(`Enabling screen sharing from getUserMedia (with fULL resolution)`);
```
przemvs:
```suggestion
        this.logger.info(`Enabling screen sharing from getUserMedia (with FULL resolution)`);
```

przemvs (approved):

src/script/calling/CallingRepository.ts
thisisamir98: don't merge this as it'll break your code, just suggesting `speaker` instead of `p`
```suggestion
        speakes.forEach(speaker => {
```

src/script/calling/CallingRepository.ts
```diff
@@ -2435,6 +2500,13 @@ export class CallingRepository {
     PrimaryModal.show(PrimaryModal.type.ACKNOWLEDGE, modalOptions);
   }
 
+  private isSelfUser(participant: Participant): boolean {
+    if (this.selfUser == null || this.selfClientId == null) {
+      return false;
+    }
+    return participant.doesMatchIds(this.selfUser.qualifiedId, this.selfClientId);
+  }
```
aweiss-dev, line 2633: we already have the "matchQualifiedIds" helper function that is used for selfUser checks, we could use this here as well just to be aligned

src/script/calling/CallingRepository.ts
```diff
@@ -1406,24 +1445,47 @@ export class CallingRepository {
...
+  /**
+   * This method monitors every change in the call and is therefore the main method for handling video requests.
+   * These changes include mute/unmute, screen sharing, or camera switching, joining or leaving of participants, or...
+   * @param call
+   * @param newPage
+   */
   changeCallPage(call: Call, newPage: number): void {
     call.currentPage(newPage);
-    if (!this.callState.isSpeakersViewActive()) {
+    if (!this.callState.isSpeakersViewActive() && !this.callState.isMaximisedViewActive()) {
```
thisisamir98, line 1538: good optimization here 👍

src/script/calling/CallingRepository.ts
V-Gira: little typo here
```suggestion
        const speakers = call.activeSpeakers();
        speakers.forEach(p => {
```

src/style/components/group-video-grid.less
```diff
@@ -163,9 +163,31 @@
         object-fit: cover;
       }
 
+      &__action_icon {
```
aweiss-dev: We actually would like to get rid of less, maybe we really should stopping adding any new less code to finally force us to refactor :/

aweiss-dev (approved):

src/script/calling/Participant.ts
```diff
@@ -39,6 +40,7 @@ export class Participant {
   public readonly hasPausedVideo: ko.PureComputed<boolean>;
   public readonly sharesScreen: ko.PureComputed<boolean>;
   public readonly sharesCamera: ko.PureComputed<boolean>;
+  public readonly isSwitchingVideoResolution = observable(false);
```
thisisamir98, line 43: maybe instead of a public property this variable could be private and we could add a method like: toggleIsSwitchingVideoResolution which would set it to true and set it to false after a second, this way you could avoid writing the setTimeouts twice in here

https://github.com/wireapp/wire-webapp/pull/18944/files#diff-815c4a1834d94a723e8cd6eb18d5608e92dd62d3fb18809251711b9239baef96R275-R279

and here:

https://github.com/wireapp/wire-webapp/pull/18944/files#diff-815c4a1834d94a723e8cd6eb18d5608e92dd62d3fb18809251711b9239baef96R293-R298

thisisamir98 (commented):
@EnricoSchw can you explain the changes in this file in the PR description please and maybe add some comments as well? 

src/script/media/MediaConstraintsHandler.ts

src/script/calling/CallingRepository.ts
EnricoSchw: I did it in this case because "doesMatchIds" uses "matchQualifiedIds" under the hood plus additionally, it also compares the client ID. This is necessary in this case because to identify the specific user and not just another client of the same user.

src/script/calling/Participant.ts
EnricoSchw: A separate method certainly makes sense. It also makes what's happening clearer.

I will not make **isSwitchingVideoResolution** private. The property is used in the view object. All such kind of properties are **public readonly** in your application.

EnricoSchw:
> @EnricoSchw can you explain the changes in this file in the PR description please and maybe add some comments as well?
> 
> src/script/media/MediaConstraintsHandler.ts

I've updated the PR description. I'm not sure what exactly I should document in the code... I'm removing the resolution constraints and setting a fixed frame rate. The config object is self-explanatory.

```js
{
  SCREEN: {
    DESKTOP_CAPTURER: {
      mandatory: {
        chromeMediaSource: 'desktop',
        maxFrameRate: 5,
      },
    },
    DISPLAY_MEDIA: {
      frameRate: 5,
    },
    USER_MEDIA: {
      frameRate: 5,
      mediaSource: 'screen',
    },
  },
}
```

src/style/components/group-video-grid.less
EnricoSchw: I don't think we need the code, so let's get rid of it.

thisisamir98 (approved):

PR #18977, opened 2025-04-07
fix: Resolve timestamp issues of migrated mls conversation (WPB-16715)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16715" title="WPB-16715" target="_blank">BugWPB-16715</a>  [Web] When a 1:1 conversation migrates to MLS it ends up on top of the conversation list
  </td></table>
  <br />
 

## Description

when we fetch this new MLS conversation, we get last_event_time equal to 1970-01-01T00:00:00.000Z as there haven't been any events for this particular new conversation yet.

since there are no previous proteus conversation there is no last_event_time we can copy from so we set it to Date.now() to avoid having a conversation with last_event_time equal to 1970-01-01T00:00:00.000Z this is important because we use last_event_time to sort our list of conversations.


src/script/conversation/EventBuilder.ts
```diff
@@ -305,7 +305,7 @@ export const EventBuilder = {
   build1to1MigratedToMLS(conversationEntity: Conversation): OneToOneMigratedToMlsEvent {
     return {
       ...buildQualifiedId(conversationEntity),
-      time: new Date(conversationEntity.getNextTimestamp()).toISOString(),
+      time: new Date().toISOString(),
```
V-Gira, line 308: leftover from testing?
I guess we're not planning to change this, this is probably what fails the tests

src/script/conversation/ConversationRepository.ts
```diff
@@ -1654,6 +1654,20 @@ export class ConversationRepository {
...
+        we get last_event_time equal to 1970-01-01T00:00:00.000Z as there
+        haven't been any events for this particular new conversation yet
+        since there are no previous proteus conversation there is no last_event_time
+        we can copy from so we set it to Date.now() to avoid
+        having a conversation with last_event_time equal to 1970-01-01T00:00:00.000Z
+        this is important because we use last_event_time to sort our list of conversations
+      */
+      if (mlsConversation.last_event_timestamp() < 10) {
+        mlsConversation.last_event_timestamp(Date.now());
+      }
```
V-Gira: doesn't this guarantees that a conversation with no event will end up at the top of the list?
Do we know why a conv would appear on top if it's last timestamp is set to unix time zero?

src/script/conversation/EventBuilder.ts
thisisamir98: this is part of the fix, not left over.

src/script/conversation/ConversationRepository.ts
thisisamir98: we sort conversations based on last timestamp, so the older the timestamp the lower it goes. this is for the case you don't have a previous proteus conversation to copy properties from and there is no event in the conversation. if we stick to setting system message time to new Date().toISOString(), we don't need this.

only visible issue with these timestamps is this:

image

src/script/conversation/EventBuilder.ts
V-Gira: do you think we can keep the old logic or does it cause issues?
It seems it's intended to prevent new timestamps older than the last one which the backend wouldn't like
```typescript
  getNextTimestamp(): number {
    const sentMessages = this.messages().filter(message => message?.status() !== StatusType.SENDING);
    if (sentMessages.length === 0) {
      return this.getLastKnownTimestamp() + 1;
    }
    const lastMessageTimestamp = sentMessages[sentMessages.length - 1].timestamp();
    // The next timestamp can never be before the last known timestamp, so we need to take the max between the last message and the last known server timestamp
    return Math.max(lastMessageTimestamp, this.getLastKnownTimestamp()) + 1;
  }
  ```
  
  maybe something like this would work?
  ```typescript
      time:
        conversationEntity.getNextTimestamp() > 10
          ? new Date(conversationEntity.getNextTimestamp()).toISOString()
          : new Date().toISOString(),

aweiss-dev (approved):

PR #18989, opened 2025-04-09
fix: default notification state for newly created conversations [WPB-16859]

waseemansar:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16859" title="WPB-16859" target="_blank">BugWPB-16859</a>  [Web] Newly created group has muted notifications by default
  </td></table>
  <br />
 

## Description
Fixes issue where newly created group and 1:1 conversations have muted notifications by default.

### Private Users

https://github.com/user-attachments/assets/0c4f49c9-7c34-45c1-9669-51a8ef70251c

### Team Users

https://github.com/user-attachments/assets/961fbdf8-3962-477d-90de-39bab9094b37

## Checklist

- [x] PR has been self reviewed by the author
- [ ] Hard-to-understand areas of the code have been commented
- [x] If it is a core feature, unit tests have been added

### DIFF
--- a/src/script/entity/Conversation.test.ts
+++ b/src/script/entity/Conversation.test.ts
@@ -956,14 +956,6 @@ describe('Conversation', () => {
       conversationEntity.mutedState(NOTIFICATION_STATES.EVERYTHING);
       expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.EVERYTHING);
 
-      //@ts-expect-error
-      conversationEntity.mutedState(true);
-      expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.MENTIONS_AND_REPLIES);
-
-      //@ts-expect-error
-      conversationEntity.mutedState(false);
-      expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.EVERYTHING);
-
       conversationEntity.mutedState(NOTIFICATION_STATES.NOTHING);
       expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.NOTHING);
 
--- a/src/script/entity/Conversation.ts
+++ b/src/script/entity/Conversation.ts
@@ -332,13 +332,7 @@ export class Conversation {
       if (!this.selfUser()) {
         return NOTIFICATION_STATE.NOTHING;
       }
-      const mutedState = this.mutedState();
-
-      if (typeof mutedState === 'boolean') {
-        return mutedState ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES : NOTIFICATION_STATE.EVERYTHING;
-      }
-
-      return mutedState;
+      return this.mutedState();
     });
 
     this.is_archived = this.archivedState;
--- a/src/script/conversation/ConversationMapper.ts
+++ b/src/script/conversation/ConversationMapper.ts
@@ -37,6 +37,7 @@ import {LegalHoldStatus} from '@wireapp/protocol-messaging';
 import {ACCESS_STATE} from './AccessState';
 import {ConversationStatus} from './ConversationStatus';
 import {ConversationVerificationState} from './ConversationVerificationState';
+import {NOTIFICATION_STATE} from './NotificationSetting';
 
 import {Conversation} from '../entity/Conversation';
 import {BaseError, BASE_ERROR_TYPE} from '../error/BaseError';
@@ -282,6 +283,10 @@ export class ConversationMapper {
     if (!conversationEntity.last_event_timestamp() && initialTimestamp) {
       conversationEntity.last_event_timestamp(initialTimestamp);
       conversationEntity.last_server_timestamp(initialTimestamp);
+    }
+
+    if (conversationEntity.mutedState() === null) {
+      conversationEntity.mutedState(NOTIFICATION_STATE.EVERYTHING);
     }
 
     // Active participants from database or backend payload


src/script/entity/Conversation.ts
```diff
@@ -334,11 +334,16 @@ export class Conversation {
...
+        return mutedState;
+      }
+
       if (typeof mutedState === 'boolean') {
-        return mutedState ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES : NOTIFICATION_STATE.EVERYTHING;
+        return mutedState ? NOTIFICATION_STATE.NOTHING : NOTIFICATION_STATE.EVERYTHING;
       }
 
-      return mutedState;
+      return NOTIFICATION_STATE.EVERYTHING;
```
e-maad: Minor suggestion, will make the code more readable:
```
if (typeof mutedState === true) {
        return NOTIFICATION_STATE.NOTHING;
      }

      return NOTIFICATION_STATE.EVERYTHING;
```

przemvs: I will don't doing this. Because typeof mutedState will boolean or falsy value. If want to do this way, will be better to check like this: typeof mutedState === 'boolean' && mutedState { nothing }

src/script/entity/Conversation.test.ts
```diff
@@ -948,29 +948,58 @@ describe('Conversation', () => {
...
 
       expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.NOTHING);
-
       conversationEntity.selfUser(selfUserEntity);
-      conversationEntity.mutedState(NOTIFICATION_STATES.EVERYTHING);
+      conversationEntity.mutedState(undefined);
+
       expect(conversationEntity.notificationState()).toBe(NOTIFICATION_STATES.EVERYTHING);
+      //@ts-expect-error
+      conversationEntity.mutedState('true');
```
przemvs: Hmm... There is smth wrong, we can't disable ts. Types are wrong in mutedState. It should have number instead of boolean.

src/script/entity/Conversation.ts
```diff
@@ -334,11 +334,16 @@ export class Conversation {
       }
       const mutedState = this.mutedState();
 
+      const knownNotificationStates = Object.values(NOTIFICATION_STATE);
+      if (knownNotificationStates.includes(mutedState)) {
+        return mutedState;
+      }
```
przemvs: I think this is not necessary. There is some problems with a mapping new Conversation

src/script/entity/Conversation.ts
```diff
@@ -334,11 +334,16 @@ export class Conversation {
...
+
       if (typeof mutedState === 'boolean') {
-        return mutedState ? NOTIFICATION_STATE.MENTIONS_AND_REPLIES : NOTIFICATION_STATE.EVERYTHING;
+        return mutedState ? NOTIFICATION_STATE.NOTHING : NOTIFICATION_STATE.EVERYTHING;
```
przemvs: What I see mutedState should be a number all the time. So there is smth wrong with this logic.

przemvs (changes requested):
There is smth wrong with current code. We passing idk, boolean, undefined, number and someone else into mutedState. We can't do this like that.

src/script/entity/Conversation.ts
waseemansar: Let me explain why we have this boolean check in the code:
### Backward Compatibility:

- The mutedState property was originally designed to work with boolean values (true/false) for simple mute/unmute functionality
- Later, it was enhanced to support more granular notification states (NOTHING, EVERYTHING, MENTIONS_AND_REPLIES) using numbers
- The boolean check is there to maintain compatibility with older code that might still be setting boolean values

src/script/entity/Conversation.ts
waseemansar: Let me explain why the `knownNotificationStates` check is actually important:

- The mutedState observable is defined as `ko.Observable<number>`, but it could potentially receive any number value
- The `knownNotificationStates` check ensures that we only return values that are actually defined in our `NOTIFICATION_STATE` enum
- This prevents invalid or unexpected notification states from being used
- Regarding the comment about problems with mapping new conversations - this check actually helps prevent such problems so When a new conversation is created or mapped, we want to ensure it only uses valid notification states

thisisamir98: this is also a question to me

> The knownNotificationStates check ensures that we only return values that are actually defined in our NOTIFICATION_STATE enum

why is that? can't we avoid it?

src/script/entity/Conversation.ts
thisisamir98: @przemvs 
muted state can be a boolean for some reason it sounds like:

https://github.com/wireapp/wire-webapp/blob/dev/src/script/entity/Conversation.ts#L337-L339

this change is from 5 years ago.

maybe we could find why is that and possibly change it

przemvs: Agree.. If we have task for notification, will be nice to find why we getting a boolean.

przemvs: But what I found in the code, we setting true/false only in testing files :D

przemvs: What I'm thinking is from first app notifications while user can have only enabled/disabled notifications,

waseemansar: > What I'm thinking is from first app notifications while user can have only enabled/disabled notifications,

yes i think that was the case

przemvs (approved):
LGTM! :)

PR #18996, opened 2025-04-10
fix: react on a detected client mismatch event [WPB-16934]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16934" title="WPB-16934" target="_blank">BugWPB-16934</a>  [Web] Wiping crypto identity without changing client ID
  </td></table>
  <br />
 

## Description

There is an issue where our existing RegisteredClient and the CryptoClient no longer have matching signatures. This makes the client unusable for MLS purposes.

This PR add the event handling and displays a logout modal for the user / disables any interaction with the app.


src/script/main/app.ts
```diff
@@ -430,9 +432,14 @@ export class App {
       }
       this.core.on(CoreEvents.NEW_SESSION, ({userId, clientId}) => {
         const newClient = {class: ClientClassification.UNKNOWN, id: clientId};
-        userRepository.addClientToUser(userId, newClient, true);
+        void userRepository.addClientToUser(userId, newClient, true);
```
e-maad, line 435: Don't we care if the promise didn't resolve?

src/script/main/app.ts
```diff
@@ -940,4 +939,28 @@ export class App {
 
     PrimaryModal.show(modalType, modalOptions);
   };
+
+  // Todo: Move this to a separate hook or service
+  private showForceLogoutModal = async (reason: SIGN_OUT_REASON) => {
```
e-maad: Do we have to make it async?

src/script/main/app.ts
aweiss-dev: i just added void for the compiler, we never waited

src/script/main/app.ts
aweiss-dev: no, actually not

e-maad (approved):
LGTM 🚀

PR #19012, opened 2025-04-14
feat: Single button for starting a call (WPB-16259)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16259" title="WPB-16259" target="_blank">TaskWPB-16259</a>  [Web] Single button for starting a 1:1 or conference call
  </td></table>
  <br />
 

## Description

- Only starting a call as audio.
- Removed redundant old code for starting a call as video
- Updated react-ui-kit version to adopt the new icon
- Updated broken tests

image
image


src/script/calling/CallingRepository.ts
```diff
@@ -946,7 +946,8 @@ export class CallingRepository {
     return CONV_TYPE.ONEONONE;
   }
 
-  async startCall(conversation: Conversation, callType: CALL_TYPE): Promise<void | Call> {
+  async startCall(conversation: Conversation): Promise<void | Call> {
+    const callType = CALL_TYPE.NORMAL;
```
przemvs: Use directly CALL_TYPE.NORMAL instead of creating new variable :)

przemvs (approved):

src/script/calling/CallingRepository.ts
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/19012/commits/38b306c88c237b6bbdf1681547afc9b9237241b6

PR #19027, opened 2025-04-16
feat: Redirect to unsupported page when desktop is enforced (WPB-17261)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17261" title="WPB-17261" target="_blank">BugWPB-17261</a>  [Web] Only desktop supported flag does not redirect user to the correct page when using browser
  </td></table>
  <br />


V-Gira (approved):
this should target `release/q2-2025` instead of dev

PR #19039, opened 2025-04-23
feat(cells): tables columns fixed width

olafsulich:
## Description

Set table columns to have a fixed width, instead of adjusting based on the data. It allows us to have, e.g., the "Shared" column be as small as it should be but grow others, like the "Name" column.

**Before:**
Screenshot 2025-04-23 at 11 22 41

**After:**
Screenshot 2025-04-23 at 11 22 52

## Checklist


src/script/components/CellsGlobalView/CellsTable/CellsTableSharedColumn/CellsTableSharedColumn.styles.ts
```diff
@@ -24,16 +24,16 @@ import {CSS_VISUALLY_HIDDEN} from 'Util/CSSMixin';
...
     alignItems: 'center',
-    paddingLeft: '10px',
+    justifyContent: 'center',
   },
 };
 
 export const wrapperStylesHidden: CSSObject = {
   ...wrapperStyles,
-  '@media (min-width: 900px)': {
+  '@media (min-width: 1000px)': {
```
aweiss-dev: This media query should be a reusable variable, seeing the same change over and over again is not feat :(

e-maad: I would also suggest to use predefined screen size breakpoints (if you can):
```
[media.desktopMin]: {
 // css
}
```

src/script/components/Conversation/ConversationCells/CellsTable/CellsTable.tsx
```diff
@@ -183,6 +178,9 @@ export const CellsTable = ({
                     key={cell.id}
                     css={cell.column.id === 'id' ? tableActionsCellStyles : tableCellStyles}
                     data-cell={cell.column.id === 'id' ? undefined : cell.column.columnDef.header}
+                    style={{
+                      width: cell.column.id == 'name' ? undefined : cell.column.getSize(),
+                    }}
```
e-maad, line 183: Minor suggestion, if you want to do it:
```suggestion
               {...(cell.column.id == 'name' && {
                style: {
                   width: cell.column.getSize(),
                },
              })}
```

olafsulich: I think it complicates the whole thing

aweiss-dev: Not a fan of this approach as well. It's less readable, and it also prints "false" for the falsy case

aweiss-dev (approved):

e-maad (approved):

PR #19041, opened 2025-04-23
fix: Avoid conversation moving to top after migrating to MLS (WPB-16715)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16715" title="WPB-16715" target="_blank">BugWPB-16715</a>  [Web] When a 1:1 conversation migrates to MLS it ends up on top of the conversation list
  </td></table>
  <br />
 

Root Cause:
The issue occurred because a system message was injected too early, which triggered another flow that unexpectedly updated the last_event_timestamp of the conversation. This led to inconsistencies in the conversation state.

Additionally, the saveConversation logic had a flaw:
It skipped saving a conversation into the state if it was already present, assuming it didn't need to be updated. However, this prevented backend-fetched conversations from being reliably synced into the state.

Fix:

Delayed the system message injection to avoid prematurely triggering dependent flows.

Replaced the conditional saving logic in saveConversation with an upsert operation to ensure that any fetched conversation—new or existing—is always saved into the state correctly.


olafsulich (approved):

e-maad (approved):
LGTM 🚀

PR #19051, opened 2025-04-28
feat(cells): folders display & navigation [WPB-17384]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17384" title="WPB-17384" target="_blank">TaskWPB-17384</a>  [Web] Folders display and navigation
  </td></table>
  <br />
 

## Description

Introducing folders for the conversation cells. Folders are available via URL, users can navigate to them by clicking on the folder name, picking an option from the menu, or using the breadcrumbs menu.

Folders can be nested, each folder represents a new segment in the URL.

Plans for the near future:
- navigating to a specific folder from the global cells view
- creating nested folders
- moving files between folders
- deleting a folder (API is missing)

## Demo

https://github.com/user-attachments/assets/306e1868-e584-430d-9368-a68bf43862be

## Checklist


src/script/router/Router.ts
```diff
@@ -75,17 +79,16 @@ const parseRoute = () => {
...
-  setHistoryParam(path, stateObj);
+export const navigate = (path: string) => {
+  setHistoryParam(path);
   parseRoute();
 };
 
-export const setHistoryParam = (path: string, stateObj: {} = window.history.state) => {
-  window.history.replaceState(stateObj, '', `#${path}`);
+export const setHistoryParam = (path: string) => {
+  window.location.hash = path;
```
olafsulich, line 93: Replace to retain history and enable browser navigation

src/script/components/Conversation/ConversationTabs/ConversationTabs.tsx
```diff
@@ -17,61 +17,88 @@
...
+      onIndexChange(0);
+    }
+  }, [onIndexChange]);
+
+  useEffect(() => {
+    handleHashChange();
+    window.addEventListener('hashchange', handleHashChange);
+    return () => window.removeEventListener('hashchange', handleHashChange);
+  }, [handleHashChange]);
+
```
thisisamir98, line 80: could you explain why we need to listen to hashchange event here?

olafsulich: We want to react to what's in the URL, and based on that activate the right tab. Consider these scenarios:

1. "/#/conversation/:conversationId/files" -> mark "files" tab as active
2. "/#/conversation/:conversationId/" -> mark "messages" tab as active

Users can enter the app with having the "files/" in the URL

olafsulich: This is the first scenario (entering the app), but we need mark the right tab also when a user clicks in the app itself. For example: I'm in the cells search view and I click on a folder, the app redirects me to ".../files/" route, the URL is changing, and the tabs have to change as well

przemvs (approved):

PR #19059, opened 2025-04-30
feat(router): replace switch-path library with path-to-regexp [WPB-17384]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17384" title="WPB-17384" target="_blank">TaskWPB-17384</a>  [Web] Folders display and navigation
  </td></table>
  <br />
 

## Description

Replacing the `switch-path` library with `path-to-regexp`.

**Reasons:**
- `switch-path` is outdated, and the last update was in 2015
- the library doesn't allow the dynamic URL segments, which will be needed for the future Cells folders changes (e.g. `conversation/:conversationId/files/segment1/segment2`)
- `path-to-regexp` is pretty popular, actively maintained, and simple
- the new library looks like an easy replacement - I've tested all available routes, and everything seems to be working as before

**Drawbacks:**
- `path-to-regexp` doesn't support optional segments like `switch-path`, that's why we have to duplicate the route instead and have both conversation with domain and without - not a big deal

**Alternatives:**
- we could think about other alternatives for routing (like react-router, which we already have for auth), but it'd require a complete rewrite, which I wanted to avoid

## Checklist


src/script/page/AppMain.tsx
```diff
@@ -182,10 +182,22 @@ export const AppMain: FC<AppMainProps> = ({
       window.history.replaceState(historyState, '', window.location.hash);
     }
 
+    const showConversationMessages = (conversationId: string, domain = apiContext.domain ?? '') => {
+      void mainView.content.showConversation({id: conversationId, domain});
+    };
+
+    const showConversationFiles = (conversationId: string, domain = apiContext.domain ?? '', filePath = '') => {
+      void mainView.content.showConversation({id: conversationId, domain}, {filePath: `files/${filePath}`});
+    };
```
przemvs, line 191: Can't You prepare 1 function for showingConversation? You can pass filePath as optional arg.

przemvs (approved):
LGTM, 1 thing

thisisamir98 (approved):

src/script/page/AppMain.tsx
olafsulich: Hmm, I don't think so. I can have ".../files/" and that's it, or I can have ".../files/something/another/...". That said, for the base "files" route (without dynamic segments), I have to include only "files/".

PR #19061, opened 2025-04-30
test(Playwright): Set up Playwright [WPB-17365]

iskvortsov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17365" title="WPB-17365" target="_blank">EpicWPB-17365</a>  Set up automation for critical flow with Playwright
  </td></table>
  <br />
 

## Description

Refer to the epic: https://wearezeta.atlassian.net/browse/WPB-17365

## Checklist


package.json
```diff
@@ -79,6 +79,7 @@
     "@emotion/eslint-plugin": "11.11.0",
     "@faker-js/faker": "9.6.0",
     "@formatjs/cli": "6.6.1",
+    "@playwright/test": "^1.52.0",
```
V-Gira: let's pin that bad boy
```suggestion
    "@playwright/test": "1.52.0",
```

playwright.config.ts
```diff
@@ -0,0 +1,91 @@
...
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
```
iskvortsov: Not sure which browsers we need, so I've commented our all except for chromium

test/e2e_tests/backend/backend.ts
```diff
@@ -0,0 +1,138 @@
...
+const BASE_URL = getCredentials(onePasswordItemName, 'backendUrl');
+
+export async function createPersonalUserViaBackdoor(user: ClientUser) {
+  const axiosInstance = axios.create({
+    baseURL: BASE_URL,
+    withCredentials: true,
+    headers: {
+      'Content-Type': 'application/json',
+    },
+  });
```
e-maad: Can we move out this initialization?
I would prefer to initialize `axiosInstance` once for each test session.

test/e2e_tests/backend/backend.ts
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  const deleteResponse = await axiosInstance.request({
+    url: '/self',
+    method: 'DELETE',
+    headers: {
+      Authorization: `Bearer ${user.token}`,
+    },
+    data: deletePayload,
+  });
```
e-maad: Can we create a module that expose api helper functions (get, put .. etc).
Then we don't have to pass the token every time. We can also maintain the user data using singleton class.

test/e2e_tests/utils/credentialsReader.ts
```diff
@@ -0,0 +1,54 @@
...
+    // if environment variable is not set search in 1password
+    return readFrom1Password(itemName, fieldName);
+  }
+
+  return environmentVariable;
+}
+
+function readFrom1Password(itemName: string, fieldName: string): string {
+  const homeDir = os.homedir();
+  const command = 'op';
```
nkaramarko: The op CLI is trusted, but to make it robust, we should check the path with 'which op' and then hardcode that path in command instead of 'op'

command = 'path_from_which_op'

test/e2e_tests/backend/backend.ts
```diff
@@ -0,0 +1,138 @@
...
+ *
+ */
+
+import axios from 'axios';
+
+import {ClientUser} from './clientUser';
+
+import {getCredentials} from '../utils/credentialsReader';
+
+const onePasswordItemName = 'BackendConnection staging-with-webapp-master';
```
nkaramarko: Looks like onePasswordItemName is gonna be used a lot in various tests?
It would be great to extract it as reusable const then

nkaramarko (changes requested):

test/e2e_tests/backend/clientUser.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  get token(): string | null {
+    return this.accessToken;
+  }
+
+  set token(value: string) {
+    this.accessToken = value;
+  }
+}
```
e-maad: I would prefer to have object value factories (TS friendly) for mock data:
```
interface User {
  email: string;
  password: string;
  firstName: string;
  lastName: string;
  username: string;
  token: string;
}

export const getUser = (user: Partial<User> = {}) => ({
  ...user,
  email: user?.email ?? faker.internet.email({lastName: user?.lastName, provider: 'wire.engineering'}).toLowerCase(),
  password: user?.password ?? faker.internet.password({length: 8, pattern: /[A-Za-z\d!@#$]/}),
  firstName: user?.firstName ?? faker.person.firstName(),
  lastName: user?.lastName ?? faker.person.lastName(),
  username: user?.username ?? `${user?.lastName}${faker.string.alpha({length: 5, casing: 'lower'})}`.toLowerCase(),
  token: user?.token ?? null,
});

const testUser1 = getUser();
const testUser2 = getUser({email: 'test@example.com'});
```
This will be more readable.

e-maad: https://medium.com/@skovy/object-factories-for-testing-in-typescript-501c8f42768e

test/e2e_tests/pages/login.page.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    this.backButton = page.locator('[data-uie-name="go-index"]');
+    this.signInButton = page.locator('[data-uie-name="do-sign-in"]');
+    this.loginForm = page.locator('[data-uie-name="login"]');
+    this.emailInput = this.loginForm.locator('[data-uie-name="enter-email"]');
+    this.passwordInput = this.loginForm.locator('[data-uie-name="enter-password"]');
+    this.loginErrorText = page.locator('[data-uie-name="error-message"]');
+  }
+
+  async isEmailFieldVisible(): Promise<boolean> {
```
e-maad: No return type needed.
```suggestion
  async isEmailFieldVisible() {
```

test/e2e_tests/utils/credentialsReader.ts
```diff
@@ -0,0 +1,54 @@
...
+  const args = ['item', 'get', '--vault', 'Test Automation', itemName, '--fields', fieldName, '--reveal'];
+  const options = {
+    cwd: homeDir,
+    // eslint-disable-next-line no-undef
+    encoding: 'utf8' as BufferEncoding,
+  };
```
e-maad, line 41: I am curious about how it will work in CI/CD env.
Should we add any fallback strategy? Or any env file from where we could read the credentials.

e-maad (commented):
Overall looks good, I really like the usage of page objects and data factories 🚀 
Added few concerns.

test/e2e_tests/utils/credentialsReader.ts
```diff
@@ -0,0 +1,54 @@
...
+import * as os from 'os';
+
+export function getCredentials(itemName: string, fieldName: string = 'password'): string {
+  const environmentVariable = process.env[itemName];
+  if (!environmentVariable) {
+    // if environment variable is not set search in 1password
+    return readFrom1Password(itemName, fieldName);
+  }
+
+  return environmentVariable;
```
e-maad:
```suggestion
  return environmentVariable || readFrom1Password(itemName, fieldName);
```

e-maad: What if we don't have credentials on any of the places?

playwright.config.ts
e-maad: We can decide what browsers we need, maybe we can pick those we support.
Do you think we should configure it using env var?

araneforseti: If you can configure it to use env var, I'd try for that; then you'd be able to set up tests using different browsers with tagging the ones which cover items important to try in other browsers

test/e2e_tests/README.md
```diff
@@ -0,0 +1,19 @@
+## Please refer to [Playwright doc](https://playwright.dev/docs/intro) for detailed overview of the framework, troubleshooting, and best practices.
+
+# Tests
+
+E2E tests (login.spec.ts and credentialsReader.spec.ts) can be found inside [test folder](/test/e2e_tests/). The folder contains [page objects](/test/e2e_tests/pages), [backend classes](/test/e2e_tests/backend), and [credentialsReader.ts](/test//e2e_tests/utils/credentialsReader.ts) for access 1Password credentials.
```
araneforseti: Small item: Given the e2e tests are going to grow beyond those two, should their names be removed from this item?

test/e2e_tests/backend/backend.ts
araneforseti: Question: Does this method mean it'll prompt for 1Password interaction multiple times in a test run?

test/e2e_tests/backend/backend.ts
```diff
@@ -0,0 +1,138 @@
...
+  // Find the zuid cookie string in setCookieHeader
+  if (!setCookieHeader) {
+    throw Error('Cookies were not provided in register response');
+  }
+  const zuidCookie = setCookieHeader.find(cookieStr => cookieStr.startsWith('zuid='));
+  if (!zuidCookie) {
+    throw new Error('zuid cookie not found in register response');
+  }
+
+  // 2. Get activation code
```
araneforseti: Getting activation/verification code is something which is gonna be needed separately once we get to automating 2FA scenarios again. While not necessary for this PR, consider extracting this part

test/e2e_tests/backend/clientUser.ts
araneforseti: +1. Builder or Factories for data like this tend to be a good idea long term.

test/e2e_tests/credentialsReader.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {test, expect} from '@playwright/test';
+
+import {getCredentials} from './utils/credentialsReader';
+
+// This test is only needed to test the integration with 1Password
+// It can be removed later
```
araneforseti: Is this test still necessary? Looks like the rest is working

test/e2e_tests/pages/login.page.ts
araneforseti: > No return type needed.

@e-maad but it does return a boolean and this fits with the page object model design. Why wouldn't a return type be needed?

test/e2e_tests/login.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+import {LoginPage} from './pages/login.page';
+import {WelcomePage} from './pages/welcome.page';
+import {getCredentials} from './utils/credentialsReader';
+
+const onePasswordItemName = 'BackendConnection staging-with-webapp-master';
+const webAppPath = getCredentials(onePasswordItemName, 'webappUrl');
+const createdUsers: ClientUser[] = [];
+
+test('Verify sign in error appearance in case of wrong credentials', {tag: ['@TC-3465', '@smoke']}, async ({page}) => {
+  const incorrectEmail = 'blablabla@gmail.com';
```
araneforseti:
```suggestion
  const incorrectEmail = 'blablabla@wire.engineering';
```

Nitpick: Instead of gmail can we use wire.engineering?

test/e2e_tests/README.md
iskvortsov: Sure, I've just tried to explain what was done in this PR. I'll rewrite it better! Thx for the input

test/e2e_tests/pages/login.page.ts
e-maad: You don't need to explicitly define return type, in TS it should be auto handled based on the returned value.

test/e2e_tests/backend/clientUser.ts
iskvortsov: Thanks, initially I have created a constructor for different cases, but I'm not sure it will be used. So I've decided not to overcomplicate things in the beginning but probably it makes sense to have it from the get go.

test/e2e_tests/backend/backend.ts
iskvortsov: Do you mean initialize BE client with `axiosInstance` per each test method? Or did I misunderstand your implications?

test/e2e_tests/backend/backend.ts
iskvortsov: Maybe I can read it one time per test suite run and store it in some data class. Would that be ok?

test/e2e_tests/backend/backend.ts
iskvortsov: Could you elaborate of maintaining the user data using a singleton? I thought of just using ClientUser class instances with this user data. Could be useful if we have more than one user in test case.

test/e2e_tests/utils/credentialsReader.ts
iskvortsov: Understood. I'll throw an error for this case.

test/e2e_tests/backend/backend.ts
e-maad: For each call you are initializing `axiosInstance`, you don't need a new instance for each call. You can initialize only once and then you could use it everywhere.

test/e2e_tests/backend/backend.ts
araneforseti: Yeah and there's a few options for how to go about that, for example in Android and iOS we are creating an env file using 1Password's template capabilities

test/e2e_tests/utils/credentialsReader.ts
iskvortsov: So if credentials are not found in env -> we look in 1Password. If credentials can't be found there we throw: 

``` 
if (result.error) {
    throw new Error(`Do you have 1Password CLI installed? Starting 1Password CLI failed: ${result.error.message}`);
  }

  if (result.status === 1) {
    throw new Error(`1Password found none or multiple items for id '${itemName}':\n${result.stderr}`);
  }
```

test/e2e_tests/utils/credentialsReader.ts
iskvortsov: I followed slightly different approach this time and it separates the steps in two. First you need to run op client to populate the .env file and then the tests are running taking the variables from that file.

araneforseti (approved):
lgtm - the new changes look very nice!

test/e2e_tests/backend/backend.ts
```diff
@@ -0,0 +1,140 @@
...
+      {handle: username},
+      {
+        headers: {
+          Authorization: `Bearer ${token}`,
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+  }
+}
```
e-maad, line 140: Great work 👏 
For the future use I would be prefer to not put all the api calls in the single class.
For example:

```
export class UserRepository extends Backend {
 ...
}

export class AuthRepository extends Backend {
 ...
}
```

test/e2e_tests/backend/user.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const getUser = (user: Partial<User> = {}): User => {
+  const firstName = user.firstName ?? faker.person.firstName();
+  const lastName = user.lastName ?? faker.person.lastName();
+
+  return {
+    ...user,
+    email: user.email ?? faker.internet.email({lastName, provider: 'wire.engineering'}).toLowerCase(),
+    password: user.password ?? faker.internet.password({length: 8, pattern: /[A-Za-z\d!@#$]/}),
+    firstName,
```
e-maad: No need for a variable.
```suggestion
    firstName: user.firstName ?? faker.person.firstName(),
```

e-maad (approved):
LGTM 🚀

nkaramarko (approved):

test/e2e_tests/backend/backend.ts
iskvortsov: Yep, that's the plan! Thank you :)

PR #19066, opened 2025-05-07
fix: Do not mark conversation as read with self user system messages (WPB-17209)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17209" title="WPB-17209" target="_blank">BugWPB-17209</a>  [Web] - After migrating the team to MLS number of unread msgs from 1-1 conversations disappears
  </td></table>
  <br />
 

After migrating the team to MLS number of unread msgs from 1-1 conversations disappears because of the inserted system message


e-maad (approved):
LGTM

V-Gira (approved):

PR #19079, opened 2025-05-12
fix: Correct copy on login flow (WPB-17270)

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17270" title="WPB-17270" target="_blank">TaskWPB-17270</a>  DNS for Bund ID staging WebApp under zinfra
  </td></table>
  <br />
 

small copy fix


olafsulich (approved):

src/i18n/en-US.json
```diff
@@ -1029,7 +1029,7 @@
   "login.passwordPlaceholder": "Password",
   "login.publicComputer": "This is a public computer",
   "login.ssoLogin": "Company log in",
-  "login.subhead": "Enter your email address or username.",
+  "login.subhead": "Enter your password to log in",
```
przemvs: I think this is wrong... Because You're change also for default login page.

thisisamir98: alright then we need a separate string

przemvs (approved):

V-Gira (approved):

PR #19133, opened 2025-06-04
fix(cells): multipart message name [WPB-17930]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17930" title="WPB-17930" target="_blank">BugWPB-17930</a>  [Web] Sanitize cells file title from Android
  </td></table>
  <br />
 

## Description

Messages' file names from Android are sent with their full path (folders). Because of this, we need to trim the "path" from the name and display only the file name, without the extension.

## Checklist

### DIFF
--- a/src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/MultipartAssets.tsx
+++ b/src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/MultipartAssets.tsx
@@ -84,7 +84,7 @@ const MultipartAsset = ({
   senderName,
   timestamp,
 }: MultipartAssetProps) => {
-  const name = trimFileExtension(initialName!);
+  const name = getName(initialName!);
   const extension = getFileExtension(initialName!);
   const size = formatBytes(Number(initialSize));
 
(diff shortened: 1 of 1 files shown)


src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/MultipartAssets.tsx
```diff
@@ -84,7 +84,7 @@ const MultipartAsset = ({
   senderName,
   timestamp,
 }: MultipartAssetProps) => {
-  const name = trimFileExtension(initialName!);
+  const name = getName(initialName!);
```
thisisamir98, line 87: how about we make getName accept nullable parameters and remove this ! from here?

thisisamir98 (approved):

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/MultipartAssets.tsx
olafsulich: I already use `!` in other places; I'd like to stick with this convention. It's shouldn't be optional, we'll have to change it in the protocol

PR #19143, opened 2025-06-11
feat: Implement async notifications [WPB-17183]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17183" title="WPB-17183" target="_blank">TaskWPB-17183</a>  [Web] Async Notification implementation
  </td></table>
  <br />
 

see: https://github.com/wireapp/wire-web-packages/pull/7066


src/script/Config.ts
```diff
@@ -84,7 +84,7 @@ const config = {
   ALLOWED_IMAGE_TYPES: ['image/bmp', 'image/gif', 'image/jpeg', 'image/jpg', 'image/png', 'image/webp'],
 
   /** Which min and max version of the backend api do we support */
-  SUPPORTED_API_RANGE: [1, env.ENABLE_DEV_BACKEND_API ? Infinity : 8],
+  SUPPORTED_API_RANGE: [8, env.ENABLE_DEV_BACKEND_API ? Infinity : 8],
```
aweiss-dev, line 87: Can we use a shared min variable instead of a magic number? We have the hardcoded 8 now in several different places

src/script/event/EventRepository.ts
```diff
@@ -201,7 +201,7 @@ export class EventRepository {
       // We make sure there is only be a single active connection to the WebSocket.
       this.disconnectWebSocket?.();
       return new Promise<void>(async resolve => {
-        this.disconnectWebSocket = account.listen({
+        this.disconnectWebSocket = await account.listen({
```
e-maad, line 204: Do we have to use `await` here? I think account.listen is not a promise.

thisisamir98: it will be after my new changes in https://github.com/wireapp/wire-web-packages/pull/7066

thisisamir98: https://github.com/wireapp/wire-web-packages/pull/7066/files#diff-0a7e6d4a8ee3b5e3153dbc2336fecd1741ea0f39e7226d6b7a5d63be31b2747fR572

aweiss-dev (approved):

V-Gira (approved):

e-maad (approved):

PR #19145, opened 2025-06-12
feat(cells): presigned urls expiration handling [WPB-17809]

olafsulich:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17809" title="WPB-17809" target="_blank">TaskWPB-17809</a>  [Web] Rename cell node modal
  </td></table>
  <br />
 

## Description

**Problem:**

Every node has a url to its destination, which expires after a certain amount of time. When the url expires, we need to refresh the nodes to get the new urls. If we don't refresh the nodes, the user will see a broken image.

---

**Solution:**

Refresh lists after the url expiration passes with a new hook `useDatePassed` - by that, we get a fresh url every time, even when the app is stale for a long period.

## Checklist


e-maad (approved):
LGTM 👍

PR #19207, opened 2025-07-04
fix(externalUrls): avoid appending locale to urls in custom envs [WB18474]

V-Gira:
## Description

We have logic to append the locale directly to the website url for German users
This is only intended to work with `https://wire.com` and can create issues with custom deployment where locales are handled differently or not at all.

This PR includes a check to only append the website url with locale for the production website. 

## Checklist

### DIFF
--- a/src/script/externalRoute.ts
+++ b/src/script/externalRoute.ts
@@ -22,6 +22,8 @@ import {Config} from './Config';
 
 const URL = Config.getConfig().URL;
 
+const isProductionWebsite = URL.WEBSITE_BASE && URL.WEBSITE_BASE === 'https://wire.com';
+
 const getTeamSettingsUrl = (path: string = '', utmSource?: string): string | undefined => {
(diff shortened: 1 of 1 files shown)


src/script/externalRoute.ts
```diff
@@ -72,6 +74,9 @@ export const addLocaleToUrl = (url?: string): string | undefined => {
   if (!url) {
     return undefined;
   }
+  if (!isProductionWebsite) {
+    return url;
+  }
   const language = currentLanguage().slice(0, 2);
   const websiteLanguage = language == 'de' ? language : 'en';
   return url.replace(Config.getConfig().URL.WEBSITE_BASE, `${Config.getConfig().URL.WEBSITE_BASE}/${websiteLanguage}`);
```
V-Gira, line 82: Without the added check, this would return `mycustomwebsiteurl.com/en` or  `mycustomwebsiteurl.com/de` regardless of the way the custom url handle locales

thisisamir98 (approved):

PR #19242, opened 2025-07-11
fix: update critical flow tests file structure [WPB-18713]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18713" title="WPB-18713" target="_blank">TaskWPB-18713</a>  [Web] Implement "Group Management" test (TC-8636)
  </td></table>
  <br />
 

## Description

Updated critical flow tests file structure, move each test into it's own file.

## Checklist

### DIFF
--- a/test/e2e_tests/specs/criticalFlow/util.ts
+++ b/test/e2e_tests/specs/criticalFlow/util.ts
@@ -1,0 +1,52 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {ApiManagerE2E} from 'test/e2e_tests/backend/apiManager.e2e';
+
+import {User} from '../../data/user';
+
+const createdUsers: User[] = [];
+const createdTeams: Map<User, string> = new Map();
+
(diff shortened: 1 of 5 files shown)


thisisamir98 (approved):

test/e2e_tests/specs/criticalFlow/util.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const addCreatedUser = (user: User) => {
+  createdUsers.push(user);
+};
+
+export const addCreatedTeam = (user: User, teamId: string) => {
+  createdTeams.set(user, teamId);
+};
+
+export const tearDown = async (api: ApiManagerE2E) => {
```
iskvortsov: I think it would be great to move util.ts somewhere else, as it will most likely be used by other tests as well

iskvortsov (changes requested):
Let's move util.ts somewhere else. It's to handy to be used with only criticalFlow tests. And I would probably rename it to teardownUtils.ts or something.

V-Gira (approved):

test/e2e_tests/specs/criticalFlow/util.ts
e-maad: Would you like to suggest any place?

aweiss-dev: we can also have an index.ts that holds the variables etc for this folder, and if we need to reuse variables for different folders then we move them one layer higher

e-maad:
> Let's move util.ts somewhere else. It's to handy to be used with only criticalFlow tests. And I would probably rename it to teardownUtils.ts or something.

This file contains the data used by the tests during run, we should get a generic name.

test/e2e_tests/specs/criticalFlow/util.ts
e-maad: Then we have to decouple the constants from tearDown function.

iskvortsov: I would move it to `test/e2e_test/utils` maybe? Or into the root folder of specs `test/e2e_test/specs`

e-maad: > we can also have an index.ts that holds the variables etc for this folder, and if we need to reuse variables for different folders then we move them one layer higher

We can later decide it, for now lets move the file to better place.

iskvortsov (approved):
Thanks for the move! LGTM

PR #19253, opened 2025-07-15
chore(e2e): rewrite PageManager

aweiss-dev:
## Description

- This PR organises the folder structure for pages and rewrites the PageManager used for playwright with ease of usability and good organisation in mind. 
- This PR adds assets that can be used in tests
- This PR already adds a new test file, but this test will be written in a different PR

### Refactoring of `PageManager`
* Change the  `PageManager` class to centralise the management of web app and team management pages and modals, with caching for reusable components (`test/e2e_tests/pageManager/index.ts`).


test/e2e_tests/pageManager/index.ts
```diff
@@ -0,0 +1,152 @@
+/*
```
aweiss-dev, line 1: relevant file

test/e2e_tests/specs/CriticalFlow/messagesIn1On1-TC-8750.spec.ts
```diff
@@ -0,0 +1,121 @@
+/*
```
aweiss-dev, line 1: this can be ignored. i just didnt want to delete / exclude it

test/e2e_tests/test.fixtures.ts
```diff
@@ -20,20 +20,20 @@
...
 
 import {ApiManagerE2E} from './backend/apiManager.e2e';
-import {PageManager} from './pages/pageManager';
+import {PageManager} from './pageManager';
 
 // Define custom test type with axios fixture
 type Fixtures = {
   api: ApiManagerE2E;
-  pages: PageManager;
+  pm: PageManager;
```
aweiss-dev: we didnt pass pages. we passed the PM

test/e2e_tests/utils/tearDownUtil.ts
```diff
@@ -54,7 +54,7 @@ export const removeCreatedTeam = async (api: ApiManagerE2E, user: User) => {
 
 export const removeCreatedUser = async (api: ApiManagerE2E, user: User) => {
   if (!user.id) {
-    throw new Error('User must have an ID to remove from createdUsers');
+    return; // No user to remove if user ID is not set
```
aweiss-dev: error made tests fail

test/e2e_tests/pageManager/index.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  openMainPage = () => {
+    return this.page.goto(webAppPath, {waitUntil: 'networkidle'});
+  };
+
+  openTeamManagementPage = () => {
+    return this.page.goto(teamManagementPath, {waitUntil: 'networkidle'});
+  };
+
+  refreshPage = (options: {waitUntil?: 'load' | 'domcontentloaded' | 'networkidle'} = {waitUntil: 'networkidle'}) => {
```
aweiss-dev, line 88: We need to use arrow functions to keep "this bindings" while destructuring

With common functions we would have following result:

const { refreshPage } = pageManager;
await refreshPage(); //  ❌ this.page is undefined

test/e2e_tests/test.fixtures.ts
e-maad: I would suggest using pageManager instead of pm.

test/e2e_tests/utils/tearDownUtil.ts
iskvortsov: I guess in the case of deleting team first, right? This issue shouldn't happen if the user was created beforehand and is not part of any team.

iskvortsov: I would avoid deleting the error throwing tbh.

test/e2e_tests/test.fixtures.ts
iskvortsov: Agree with Immad. I avoided using `pm` for the same reason.

test/e2e_tests/utils/tearDownUtil.ts
aweiss-dev: We can log the error, but should a test fail because a user.id wasn't found?

test/e2e_tests/specs/CriticalFlow/accountManagement-TC-8639.spec.ts
```diff
@@ -30,9 +30,12 @@ const teamName = 'Critical';
 const conversationName = 'Tracking';
 const appLockPassphrase = generateSecurePassword();
 
-test('Account Management', {tag: ['@TC-8639', '@crit-flow-web']}, async ({pages, api}) => {
+test('Account Management', {tag: ['@TC-8639', '@crit-flow-web']}, async ({pm, api}) => {
   test.slow(); // Increasing test timeout to 90 seconds to accommodate the full flow
 
+  const {refreshPage} = pm;
+  const {pages, modals, components} = pm.webapp;
```
aweiss-dev: How comes? 

Destructuring is imho an elegant and readable way to access object entities. The change I will incorporate is to name pm => packageManager

What we pass into the tests is not really "pages", it's the page manager. And since we split the utilities more into different aspects (pages, modals) and also separate webapp and tm, the page names would be duplicated

pages.webapp.pages.PAGE for example

test/e2e_tests/utils/tearDownUtil.ts
iskvortsov: But if the user.id is not found it means that teardown either 1) used in appropriately (like when deleting team members or deleting user who was not created in the first place)  2) user.id got lost in the process because of some bug

iskvortsov (approved):
Thank you for applying all the changes. LGTM!

e-maad (approved):
LGTM

PR #19258, opened 2025-07-16
fix(files): check for sharing permission on pasting files only [WPB-18723]

V-Gira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18723" title="WPB-18723" target="_blank">BugWPB-18723</a>  [Web] When Team does not allow file transfer, pasting text from clip board triggers mutliple warnings
  </td></table>
  <br />
 

## Description

Every paste event in the input bar currently performs a check to whether or not file sharing is permitted for the team.

This triggers the check (and prevent pasting for files if the setting is enabled) only when the clipboard contains a file

## Checklist

### DIFF
--- a/src/script/components/InputBar/useFileHandling/useFilePaste/useFilePaste.test.ts
+++ b/src/script/components/InputBar/useFileHandling/useFilePaste/useFilePaste.test.ts
@@ -67,6 +67,7 @@ describe('useFilePaste', () => {
     const calledWithFile = mockOnFilePasted.mock.calls[0][0];
     expect(calledWithFile instanceof File).toBe(true);
     expect(calledWithFile.name).toBe(`Pasted file from ${mockFormattedDate}.txt`);
+    expect(checkFileSharingPermissionModule.checkFileSharingPermission).toHaveBeenCalled();
   });
 
   it('ignores paste events with text/plain content', () => {
@@ -79,6 +80,7 @@ describe('useFilePaste', () => {
     });
 
     expect(mockOnFilePasted).not.toHaveBeenCalled();
+    expect(checkFileSharingPermissionModule.checkFileSharingPermission).not.toHaveBeenCalled();
   });
 
   it('does nothing when no files are pasted', () => {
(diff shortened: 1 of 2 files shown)


src/script/components/InputBar/useFileHandling/useFilePaste/useFilePaste.ts
```diff
@@ -60,14 +60,15 @@ export const useFilePaste = ({onFilePasted}: UseFilePasteParams) => {
...
       }
     },
     [processClipboardFiles],
   );
 
   useEffect(() => {
-    document.addEventListener('paste', checkFileSharingPermission(handlePasteEvent));
-    return () => document.removeEventListener('paste', checkFileSharingPermission(handlePasteEvent));
+    document.addEventListener('paste', handlePasteEvent);
+    return () => document.removeEventListener('paste', handlePasteEvent);
```
V-Gira, line 72: context:
```ts
export const checkFileSharingPermission = <T extends (...args: any[]) => void>(callback: T): T | (() => void) => {
  const teamState = container.resolve(TeamState);

  if (teamState.isFileSharingSendingEnabled()) {
    return callback;
  }
  return () => {
    showWarningModal(
      t('conversationModalRestrictedFileSharingHeadline'),
      t('conversationModalRestrictedFileSharingDescription'),
    );
  };
};
```
The modal was being trigerred every time anything was pasted if file sharing was disabled

src/script/components/InputBar/useFileHandling/useFilePaste/useFilePaste.test.ts
```diff
@@ -92,12 +94,6 @@ describe('useFilePaste', () => {
 
     expect(mockOnFilePasted).not.toHaveBeenCalled();
   });
-
-  it('uses checkFileSharingPermission wrapper', () => {
-    renderHook(() => useFilePaste({onFilePasted: mockOnFilePasted}));
-
-    expect(checkFileSharingPermissionModule.checkFileSharingPermission).toHaveBeenCalled();
-  });
```
V-Gira, line 100: using `checkFileSharingPermission` as a wrapper for every paste event is what causes the issue, we need to check permissions only when pasting files

przemvs (approved):

PR #19304, opened 2025-07-30
chore: Do not use api version 10

thisisamir98:
api version 10 has breaking changes so as long as they are not adopted we shouldn't be using it.


e-maad (approved):
LGTM 👍

PR #19311, opened 2025-08-01
fix: individual file preview on multiple file upload [WPB-19231]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19231" title="WPB-19231" target="_blank">BugWPB-19231</a>  [Web] Click on file not working
  </td></table>
  <br />
 

## Description
When multiple assets uploaded at the same time, the click to preview file does not work.

**Issue:**

https://github.com/user-attachments/assets/0062fe39-8050-4a4a-b3ba-2d17ee684df4

**Fix:**

https://github.com/user-attachments/assets/a3f587e0-9a02-4b17-93ea-49e8d459dd1e

## Checklist


src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
```diff
@@ -53,27 +55,29 @@ export const FileAssetSmall = ({
...
-        fileExtension={extension}
-        senderName={senderName}
-        timestamp={timestamp}
-        isOpen={isOpen}
-        onClose={() => setIsOpen(false)}
-        isError={isError}
-        isLoading={isLoading}
-      />
-    </FileCard.Root>
+    <button onClick={() => setIsOpen(true)} css={contentWrapperStyles}>
```
arjita-mitra: <button> should have an accessible name, especially since it doesn't contain direct text, something like aria-label={`Open preview for ${name}`}

arjita-mitra: Suggestion: We could use useCallback to avoid function creation on every render

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
```diff
@@ -53,27 +55,29 @@ export const FileAssetSmall = ({
...
+      <FileCard.Root extension={extension} name={name} size={size}>
+        <FileCard.Header>
```
arjita-mitra, line 68: Suggestion: To improve readability, we could extract FileCard.Header, FileCard.Name, and the modal into separate render functions or components.

```
function renderFileHeader() {
  return (
    <FileCard.Header>
      <FileCard.Icon type={isError ? 'unavailable' : 'file'} />
      {!isError && <FileCard.Type />}
      <FileAssetOptions ... />
    </FileCard.Header>
  );
}

```

arjita-mitra (changes requested):

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
przemvs: It's not needed ;) It's doesn't contain huge actions inside and also using a lot of useCallbacks will break our performance :)

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
przemvs: This also not needed :) This is why we have component composition there :)

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
przemvs: You can extract () => setIsOpen(true) to separate function, You calling for this 2 times ;)

przemvs (approved):
LGTM, 1 thing :)

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.styles.tsx
```diff
@@ -0,0 +1,28 @@
+/*
```
zskhan: There is some inconsistency with styles file name.
Some of them has `.tsx` extension and some of them has `.ts`

Also any specific reason to duplicate the styles?
Can we extract out them in some common place?

e-maad: It should be .ts always.

src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/FileAssetCard/FileAssetSmall/FileAssetSmall.tsx
e-maad: This will reduce the readability, if we create a render function inside a component. Also the component is not big enough to break into smaller ones.

PR #19329, opened 2025-08-06
feat: Implement breaking changes of api V10 & V11 [WPB-19039]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19039" title="WPB-19039" target="_blank">TaskWPB-19039</a>  [Web] Upgrade to API v10
  </td></table>
  <br />
 

SEE https://github.com/wireapp/wire-web-packages/pull/7195


src/script/auth/hooks/useEnterpriseLoginV2.ts
```diff
@@ -73,7 +73,7 @@ export const useEnterpriseLoginV2 = ({
         }
 
         case DomainRedirect.BACKEND: {
-          const url = await getRedirectURL(response.backend_url);
+          const url = await getRedirectURL(response.backend.webapp_url);
```
e-maad: @thisisamir98 you need to update the types in web-packages repo.
Also, now we don't need to call `getRedirectURL` as we are getting it from BE directly.

thisisamir98: https://github.com/wireapp/wire-webapp/pull/19329/commits/c9ad9b372c13d0ea40131362d1aae00ac8c27cc1

PR #19434, opened 2025-08-21
fix(media-stream): prevent camera light from staying on after call settings change [WPB-11786]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-11786" title="WPB-11786" target="_blank">BugWPB-11786</a>  [Web] camera light stayed on after a 1:1 call
  </td></table>
  <br />
 

Prevent getting multiple instances of media devices on AV Preferences screen

## Description
When a user visits Audio / Video Settings view, due to re rendering, it acquires multiple media streams.
Which are not destroyed when the component is completely removed from the DOM and re-rendered.

Re-rendering is happening because of `useKoSubscribableChildren`.
I moved all instances of `useKoSubscribableChildren` to the parent component(MainContent.tsx) and pass the required value to the child components (AVPreferences.tsx, CameraPreferences.tsx) and memoized them.
In this way now both of these components are rendered only once and hence only 1 media stream of Microphone and Camera is in use and properly destroyed on component destroy.

## Checklist


src/script/page/MainContent/MainContent.tsx
```diff
@@ -187,6 +205,10 @@ const MainContent: FC<MainContentProps> = ({
                   callingRepository={repositories.calling}
                   mediaRepository={repositories.media}
                   propertiesRepository={repositories.properties}
+                  isMediaDevicesAreInitialized={isMediaDevicesAreInitialized}
```
aweiss-dev: areMediaDevicesInitialised is already a boolean, no need for 'is'

aweiss-dev (approved):

src/script/page/MainContent/panels/preferences/avPreferences/CameraPreferences.tsx
```diff
@@ -40,29 +39,25 @@ interface CameraPreferencesProps {
   hasActiveCameraStream: boolean;
   refreshStream: () => Promise<MediaStream | void>;
   streamHandler: MediaStreamHandler;
+  availableDevices: (MediaDeviceInfo | ElectronDesktopCapturerSource)[];
+  currentDeviceId: string;
 }
 
-const CameraPreferences: React.FC<CameraPreferencesProps> = ({
+const CameraPreferencesComponent: React.FC<CameraPreferencesProps> = ({
```
przemvs: While You updating component, please remove FC props :) And pass them next to props.

src/script/page/MainContent/MainContent.tsx
```diff
@@ -105,9 +107,25 @@ const MainContent: FC<MainContentProps> = ({
...
+  );
+  const {[MediaDeviceType.VIDEO_INPUT]: currentDeviceId} = useKoSubscribableChildren(devicesHandler?.currentDeviceId, [
+    MediaDeviceType.VIDEO_INPUT,
+  ]);
+  const deviceSupport = useKoSubscribableChildren(devicesHandler?.deviceSupport, [
+    MediaDeviceType.AUDIO_INPUT,
+    MediaDeviceType.AUDIO_OUTPUT,
+    MediaDeviceType.VIDEO_INPUT,
+  ]);
+  const {isMediaDevicesAreInitialized} = useInitializeMediaDevices(devicesHandler, mediaRepo.streamHandler);
```
przemvs: Please be careful with this, it can throw some of errors :)

zskhan: Can you please guide me what should I care for?

src/script/page/MainContent/panels/preferences/avPreferences/CameraPreferences.tsx
```diff
@@ -80,16 +75,18 @@ const CameraPreferences: React.FC<CameraPreferencesProps> = ({
         setStream(stream);
       }
     } catch (error) {
-      logger.warn(`Requesting MediaStream for type "${MediaType.VIDEO}" failed: ${error.message}`, error);
+      if (error instanceof Error) {
```
e-maad, line 78: Do we need this check?

EnricoSchw:
@zskhan I see what you're trying to do.

I think, because you've moved access to the device handler to a higher component, the app will now aske for video and audio permissions immediately when the the user logging in.

However, we only want to ask for permissions when we do the first call. React is not my strongest point, do you see a way around that?

EnricoSchw:
@zskhan Please not mere!

EnricoSchw (changes requested):
Let us discuss first my question before you merge, please.

zskhan:
> @zskhan I see what you're trying to do.
> 
> I think, because you've moved access to the device handler to a higher component, the app will now aske for video and audio permissions immediately when the the user logging in.
> 
> However, we only want to ask for permissions when we do the first call. React is not my strongest point, do you see a way around that?

@EnricoSchw As we discussed this thing yesterday, I fixed the issue.
Whenever you have time please take a look.
I verified the ongoing call scenario also.

EnricoSchw:
> @EnricoSchw As we discussed this thing yesterday, I fixed the issue.
> Whenever you have time please take a look.
> I verified the ongoing call scenario also.

@zskhan Good work it works 👍

EnricoSchw (approved):

src/script/page/MainContent/panels/preferences/avPreferences/CameraPreferences.tsx
zskhan: Not needed but my IDE was giving error so I added this check as it was also there in other places.

PR #19442, opened 2025-08-25
feat: fix flickering messages

przemvs:
## Checklist


src/script/components/MessagesList/VirtualizedMessagesList/useScrollMessages.ts
```diff
@@ -32,6 +32,8 @@ interface Props {
   setAlreadyScrolledToLastMessage: (scrolled: boolean) => void;
 }
 
+let scrollTimeout: number | null = null;
```
e-maad, line 35: I am thinking about the case when we have more than one list at the same time.
There may be conflict between keeping this value correct, what do you think ?

przemvs: It's not possible to have more than one list :)

PR #19452, opened 2025-08-29
fix: focus password input when email is pre-filled and locked (WPB-17650)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17650" title="WPB-17650" target="_blank">TaskWPB-17650</a>  [Web] On second screen of the new login flow, the cursor focus should be on the Password Entry field
  </td></table>
  <br />
 

## Description
When the login flow advances to the second screen (email disabled), the
password field is now programmatically focused. This allows users to
start typing their password immediately after pressing Enter on the
email step, without needing an extra click.

- Added useEffect to focus password input only when defaultEmail is set
- Avoided using HTML autofocus prop (accessibility best practice)
- Preserved caret at end if password manager pre-filled a value
- fixed no-floating-promises eslint issue inside handleSubmit

Screenshot 2025-08-29 at 10 45 55

## Checklist


src/script/auth/component/LoginForm.tsx
```diff
@@ -95,9 +95,46 @@ const LoginForm = ({isFetching, onSubmit}: LoginFormProps) => {
...
+    attemptPasswordFieldFocus();
+
+    // Strategy 2: Post-layout attempt (handles browser autofill timing)
+    const animationFrameId = requestAnimationFrame(attemptPasswordFieldFocus);
+
+    return () => {
+      cancelAnimationFrame(animationFrameId);
+    };
+  }, [defaultEmail]);
+
```
e-maad, line 106: I think it is a bit over-engineered solution, a simple approach would be:
```
useEffect(() => {
  if (!defaultEmail || !passwordInput.current) return;

  requestAnimationFrame(() => {
    passwordInput.current?.focus({ preventScroll: true });
    const value = passwordInput.current?.value ?? "";
    passwordInput.current?.setSelectionRange(value.length, value.length);
  });
}, [defaultEmail]);
```

I also think this hook should be moved outside and we can use it for other components, for example:
```
useAutoFocus(any-field);
```

src/script/auth/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,60 @@
...
+export const useAutoFocus = ({
+  elementRef,
+  preventScroll = true,
+  cursorAtEnd = true,
+  shouldFocus = true,
+}: UseAutoFocusProps) => {
+  useEffect(() => {
+    // Early return if should not focus
+    if (!shouldFocus) {
+      return;
```
przemvs: You have to return function with undefined value :)


```suggestion
      return () => undefined;
```

src/script/auth/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      // Check if element exists at animation frame time
+      if (!element) {
+        return;
+      }
+
+      element.focus({preventScroll});
+
+      if (cursorAtEnd && 'setSelectionRange' in element) {
+        const inputElement = element as HTMLInputElement | HTMLTextAreaElement;
```
przemvs, line 52: Please add typeguard in guards/HTMLElement.ts for checkign if it's a input or textarea :)

src/script/auth/hooks/useAutoFocus.ts
```diff
@@ -0,0 +1,60 @@
...
+      const element = elementRef.current;
+
+      // Check if element exists at animation frame time
+      if (!element) {
+        return;
+      }
+
+      element.focus({preventScroll});
+
+      if (cursorAtEnd && 'setSelectionRange' in element) {
```
przemvs, line 51: If You check if is input or textarea element then You don't have to check if setSelectionRange exist in element :)

e-maad (approved):

PR #19482, opened 2025-09-03
fix: More detailed monitoring of the epoch handover to AVS [WPB-19989]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19989" title="WPB-19989" target="_blank">BugWPB-19989</a>  [Calling] Epoch Update not passed on AVS
  </td></table>
  <br />
 

## Description

When passing the epoch information to AVS, we notice significant things.

- Epochs are much larger than the join/leave events would expect.
- Epoch updates are not always passed to AVS.
- Different clients send with different epochs.

To narrow down the cause, we are monitoring the epoch generation and forwarding to the AVS SDK.

## Checklist


src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1386,7 +1389,8 @@ export class CallingRepository {
     if (isTelemetryEnabledAtCurrentEnvironment()) {
       this.showCallQualityFeedbackModal(conversationId);
     }
-
+    const serializeSelfUser = this.selfUser ? this.serializeQualifiedId(this.selfUser) : 'unknown';
+    this.logger.info(`Call Epoch Info: _leave, user: ${serializeSelfUser}, conversation: ${conversationId}`);
```
zskhan: Do we need something to identify that it is `MLS 1:1` or `MLS Conference`

zskhan (approved):

src/script/repositories/calling/CallingRepository.ts
EnricoSchw: I also thought about it, but the information is already somewhere in the logs, and actually we are only interested in the group calls.. and the logs themselves will only be temporary (I think). If you insist, I'll change it

thisisamir98 (approved):

PR #19521, opened 2025-09-08
feat(media-repository): Replace KO with zustand in media handler and remove MediaRepository [WPB-19835]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19835" title="WPB-19835" target="_blank">SpikeWPB-19835</a>  [Web] Remove Knockout - extension, integration, media repository
  </td></table>
  <br />
 

## Checklist


src/script/components/calling/VideoControls/VideoControls.tsx
```diff
@@ -472,7 +465,7 @@ export const VideoControls: React.FC<VideoControlsProps> = ({
         <li className="video-controls__item">
           <button
             className="video-controls__button"
-            data-uie-value={!isMuted ? 'inactive' : 'active'}
+            data-uie-value={isMuted ? 'inactive' : 'active'}
```
zskhan, line 475: CSS we are providing below is 
`css={!isMuted ? videoControlActiveStyles : videoControlInActiveStyles}`
so the `data-uie-value` should also have same condition.

src/script/repositories/calling/Call.ts
```diff
@@ -88,6 +88,7 @@ export class Call {
     public readonly conversationType: CONV_TYPE,
     private readonly selfParticipant: Participant,
     callType: CALL_TYPE,
+    // @ts-ignore
     private readonly mediaDevicesHandler: MediaDevicesHandler,
```
zskhan, line 92: It is not used anymore but there are quite some places where it is passed from. In the next PR I will remove it.

src/script/components/calling/VideoControls/VideoControls.tsx
```diff
@@ -297,7 +290,7 @@ export const VideoControls: React.FC<VideoControlsProps> = ({
     .concat(hasBlurredBackground ? blurredBackgroundOptions.options[0] : blurredBackgroundOptions.options[1]);
 
   const updateVideoOptions = (selectedOption: string | BlurredBackgroundStatus) => {
-    const camera = videoOptions[0].options.find(item => item.value === selectedOption) ?? selectedVideoOptions[0];
+    const camera = videoOptions[0].options.find(({value}) => value === selectedOption) ?? selectedVideoOptions[0];
```
aweiss-dev, line 300: nitpick, but how about we use array.at(0) in this case (and similar cases)

src/script/repositories/media/useMediaDevicesStore.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export const defaultAudioInputId = 'default';
+export const defaultAudioOutputId = 'default';
+export const defaultVideoInputId = 'default';
+export const defaultScreenInputId = 'screen';
+
+export type MediaDevicesState = {
+  // device lists
+  audioInputDevices: MediaDeviceInfo[];
+  audioOutputDevices: MediaDeviceInfo[];
```
aweiss-dev: does this flat structure make sense?

We could also group them up:

audio: {
  inputDevices: 
  selectedDeviceId:
  isSupported:
}
...

zskhan: I would like to avoid grouping because updating a nested object replaces that object's reference and any subscriber selecting the whole audio object re-renders even if only one property is changed.

But open for suggestion.

src/script/components/calling/VideoControls/VideoControls.tsx
zskhan: `Array.prototype.at() is typed as T | undefined`
We can use `.at()` but need some changes for handling undefineds.

e-maad (approved):
LGTM 🚀

PR #19606, opened 2025-09-29
fix: 1:1 conversations gives false unread after CPB import(WPB-20062)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20062" title="WPB-20062" target="_blank">BugWPB-20062</a>  [All clients] 1:1 conversations should ignore MLS migration message when restoring backup
  </td></table>
  <br />
 

## Description
mapConversationRecord currently returns null if the conversation name is falsy. Since 1:1 conversations are exported with an empty string as their name, they get filtered out during CPB import. This prevents their
last_event_timestamp from being restored, causing them to appear unread
after backup restore.

Fix:
Ensure that empty-string names for 1:1 conversations are preserved instead of filtered out.

## Checklist


V-Gira (approved):
LGTM :rocket: 

I tested locally after review, no regression :+1:

PR #19642, opened 2025-10-08
chore(release): use git branch --format for clean contains check

thisisamir98:
## Description

it wrongly rejected valid commits because git branch --contains outputs a “*” on the current branch; now we use the formatted branch list (git branch --contains <sha> --format="%(refname:short)") so the containment check is clean and tagging no longer fails on commits that are actually on the branch.


e-maad (approved):

zskhan:
Then we can close this [PR](https://github.com/wireapp/wire-webapp/pull/19561/files) in favor of yours.

zskhan (approved):

PR #19679, opened 2025-10-16
fix: Recover from having a local unestablished MLS group [WPB-21122]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21122" title="WPB-21122" target="_blank">BugWPB-21122</a>  [Web] Recover from having a local unestablished MLS group
  </td></table>
  <br />
 

### Problem scenario
If a client fails to establish a MLS conversation it might be left with a local MLS group in core crypto. Since this MLS group was never established it will be stuck epoch 0.

Since conversationExists()will return true clients will consider this conversation to be established even though it’s not. 

### Recovery 
When trying to send a message in a conversation with a local MLS group stuck at epoch 0, delete the local MLS group and re-join the group / establish the group.

### DIFF
--- a/src/script/repositories/conversation/ConversationRepository.ts
+++ b/src/script/repositories/conversation/ConversationRepository.ts
@@ -2198,7 +2198,7 @@ export class ConversationRepository {
           this.logger.error('Epoch is 0, but retry is false, not retrying again', {conversationId, groupId, epoch});
           return;
         }
-        return this.recoverFromLocalUnestblishedMLSConversations({conversationId, groupId, epoch, core});
+        return this.recoverFromLocalUnestablishedMLSConversations({conversationId, groupId, epoch, core});
       }
       return;
     }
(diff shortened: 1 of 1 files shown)


src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2229,6 +2217,91 @@ export class ConversationRepository {
...
+      conversationId,
+    );
+  };
+
+  /**
+   * Recovers from local unestablished MLS conversations by refetching metadata and re-establishing the conversation.
+   * This is typically needed when the local epoch is 0 but the epoch on backend is greater than 0
+   * indicating that the conversation has not been properly established.
+   */
+  private recoverFromLocalUnestblishedMLSConversations = async ({
```
typfel:
```suggestion
  private recoverFromLocalUnestablishedMLSConversations = async ({
```

e-maad (commented):
LGTM 👍

e-maad (approved):
LGTM 👍

PR #19714, opened 2025-10-28
fix: accessibility improvements login(https://wearezeta.atlassian.net/browse/WPB-20819)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20819" title="WPB-20819" target="_blank">EpicWPB-20819</a>  [Web] Accessibility Improvements: Login
  </td></table>
  <br />
 

## Description
 make password toggle button accessible and localised [(Ticket)](https://wearezeta.atlassian.net/browse/WPB-21228)
 Login-new view not announced [(Ticket)](https://wearezeta.atlassian.net/browse/WPB-21230)
make back button accessible
code input new label and each digit accessible name

## Checklist


src/script/auth/component/RouteA11y.tsx
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {useRouteA11y} from '../hooks/useRouteA11y';
+
+export const RouteA11y: React.FC = () => {
```
aweiss-dev: React.FC type is unneccesary, especially since React 18. It should be removed

src/script/auth/hooks/useRouteA11y.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import {useEffect} from 'react';
+
+import {useLocation} from 'react-router-dom';
+
+export function useRouteA11y(screenKey?: string) {
+  const location = useLocation();
+
+  useEffect(() => {
+    const focusTarget =
```
aweiss-dev: const focusTarget: HTMLElement = 

Declaring the type from the start is a better approach than casting the type later on.

src/script/auth/page/Root.tsx
```diff
@@ -146,6 +154,7 @@ const RootComponent: FC<RootProps & ConnectedProps & DispatchProps> = ({
           </ContainerXS>
         ) : (
           <Router>
+            <RouteA11y />
```
aweiss-dev, line 159: Instead of writing a route for this that does nothing but include a hook, how about loading the hook directly in Root.tsx?
Seems like the RouteAlly component is not necessary

src/script/auth/page/Login.tsx
```diff
@@ -129,6 +132,11 @@ const LoginComponent = ({
 
   const isOauth = UrlUtil.hasURLParameter(QUERY_KEY.SCOPE, window.location.hash);
 
+  const [screen, setScreen] = useState<'login' | 'verify'>('login');
```
aweiss-dev: maybe "routeAction" would be a clearer naming. Screen is usually used literally for the end device screen, like desktop / mobile etc

src/script/auth/page/Login.tsx
```diff
@@ -464,11 +494,17 @@ const LoginComponent = ({
...
+                          role="heading"
+                          aria-level={1}
+                          data-page-title
+                          tabIndex={-1}
+                        >
                           {t('index.welcome', {brandName: Config.getConfig().BACKEND_NAME})}
                         </div>
                       ) : (
-                        <Heading level={embedded ? '2' : '1'} center>
+                        <Heading level={embedded ? '2' : '1'} center data-page-title tabIndex={-1}>
```
aweiss-dev, line 507: The TabIndex -1 is usually an antipattern and needs to be avoided. 
Could you please explain briefly why it is needed here?

src/script/auth/page/Root.tsx
```diff
@@ -73,6 +74,13 @@ const RootComponent: FC<RootProps & ConnectedProps & DispatchProps> = ({
   isFetchingSSOSettings,
   doGetSSOSettings,
 }) => {
+  useEffect(() => {
+    const style = document.createElement('style');
+    style.textContent = `.sr-only-focus:focus, .sr-only-focus:focus-visible { outline: none !important; }`;
```
aweiss-dev, line 81: Is it possible to write a CSS class and apply it instead of manipulating the style directly?

src/script/auth/component/RouteA11y.tsx
arjita-mitra: added it to fix the 'any' return type error, now i added null return type

src/script/auth/hooks/useRouteA11y.ts
arjita-mitra: focusTarget can be null | HTMLElement so updating it to const focusTarget: HTMLElement | null =

src/script/auth/page/Login.tsx
arjita-mitra: tabIndex={-1} makes the heading programmatically focusable so useRouteA11y() can shift focus to it after navigation and trigger the screen reader announcement. With a plain heading there’s nothing focusable to target, so the hook couldn’t announce the new page without temporarily opting for focusability

src/script/auth/page/Root.tsx
arjita-mitra: as discussed added inline comment

src/script/auth/page/Root.tsx
arjita-mitra: useRouteA11y() calls useLocation(), and useLocation() only works when the component is already inside a router context. In Root component the <Router> is rendered later in the tree so when I call useRouteA11y() near the top of the component there’s no router provider above it yet and useLocation() throws.

aweiss-dev (dismissed):

aweiss-dev (approved):

PR #19749, opened 2025-11-07
test: remove util `isMessageVisible()` in favor of web first assertions

markbrockhoff:
## Description

This is a performance and stability optimization for our tests. I noticed that sending a message takes a lot of time. Turns out the util always waits 5s after sending the message to ensure it was sent.
Since this util was creating a manual assertion I removed it in favor of a new `getMessage()` util which integrates nicely with playwrights web first assertions aligning with best practices: https://playwright.dev/docs/best-practices#dont-use-manual-assertions
Using web first assertions not only makes the tests more stable but also speeds them up a lot since the test no longer waits a fixed time after sending the message but can continue as soon as the assertion is fulfilled.

> **Note:** I manually compared the results of the last nightly CI run with a local test run to ensure only tests which also failed in CI fail locally to avoid breaking any of the already working tests with this change.

## Checklist

### DIFF
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -202,17 +202,26 @@ export class ConversationPage {
   }
 
   /**
-   * Util to get a message in the conversation sent by a given user
-   * @param user Optional parameter to specify user who sent the message. If undefined the last message matching the content will be returned.
+   * Util to get a message in the conversation
+   * @param options.content Only match messages containing this text
+   * @param options.sender Only match messages send by this user
+   * @returns a Locator to the matching message(s)
    */
-  getMessage(messageContent: string, user?: User) {
-    const message = this.messageItems.filter({hasText: messageContent});
+  getMessage(options?: {content?: string | RegExp; sender?: User}): Locator {
+    let message = this.messageItems;
 
-    if (user !== undefined) {
-      return message.filter({has: this.page.getByTestId('sender-name').getByText(user.fullName)});
+    if (options?.content) {
+      message = message.filter({hasText: options.content});
     }
 
-    return message.last();
+    if (options?.sender?.fullName) {
+      message = message.filter({
+        // Using getByLabel doesn't work here as the aria label is just placed on a div with no input inside which could be located
+        has: this.page.locator(`.content-message-wrapper[aria-label*="${options.sender.fullName}"]`),
+      });
+    }
+
+    return message;
   }
 
   /**
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -152,9 +152,7 @@ export class ConversationPage {
   async sendMessage(message: string) {
     await this.messageInput.fill(message);
     await this.sendMessageButton.click();
-    // Wait for the specific message to appear in the conversation
-    const messageLocator = this.messages.filter({hasText: message}).last();
-    await messageLocator.waitFor({state: 'visible', timeout: 20_000});
+    await this.isMessageVisible(message); // isMessageVisible not only checks for visibility but ensures the optimistic update was persisted
   }
 
   async typeMessage(message: string) {
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -152,7 +152,6 @@ export class ConversationPage {
   async sendMessage(message: string) {
     await this.messageInput.fill(message);
     await this.sendMessageButton.click();
-    await this.isMessageVisible(message); // isMessageVisible not only checks for visibility but ensures the optimistic update was persisted
   }
 
   async typeMessage(message: string) {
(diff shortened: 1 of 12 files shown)


test/e2e_tests/specs/CriticalFlow/conversationManagement-TC-8636.spec.ts
```diff
@@ -73,21 +73,24 @@ test('Conversation Management', {tag: ['@TC-8636', '@crit-flow-web']}, async ({p
 
   await test.step('Team owner signed in to the application and verify messages', async () => {
     await pages.conversationList().openConversation(conversationName);
-    for (const member of members) {
-      expect(await pages.conversation().isMessageVisible(`Hello team! ${member.firstName} here.`)).toBeTruthy();
-    }
+    await Promise.all(
+      members.map(async member => {
```
markbrockhoff, line 76: I noticed this assertion across 4 browsers can run concurrently maybe speeding things up a bit :)

test/e2e_tests/utils/userActions.ts
```diff
@@ -77,13 +77,6 @@ export const sendTextMessageToConversation = async (
   const {pages} = pageManager.webapp;
   await pages.conversationList().openConversation(conversation);
   await pages.conversation().sendMessage(message);
-
-  // TODO: Bug [WPB-18226] Message is not visible in the conversation after sending it
-  await pages.conversation().page.waitForTimeout(500); // Wait for the message to be sent
```
markbrockhoff, line 82: While the bug is probably still valid waiting 500ms after sending the message doesn't fix it and in my opinion the assertion shouldn't be part of the util.

test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -224,6 +199,20 @@ export class ConversationPage {
     return await locator.screenshot();
   }
 
+  /**
+   * Util to get a message in the conversation sent by a given user
+   * @param user Optional parameter to specify user who sent the message. If undefined the last message matching the content will be returned.
+   */
+  getMessage(messageContent: string, user?: User) {
```
aweiss-dev: similar to my comment in the other PR, i mersonally am a big fan of speaking variable and function names, as well as pure, one task only, functions.

I would recommend to have a getLastMessage() and getLastMessageFromUser(user) function

test/e2e_tests/specs/CriticalFlow/addMembersToChat-TC-8631.spec.ts
```diff
@@ -124,80 +124,80 @@ test(
...
-      expect(await pages.conversation().isMessageVisible(`Hello from ${member2.firstName}!`)).toBeTruthy();
+      await expect(pages.conversation().getMessage(`Hello from ${member1.firstName}!`)).toBeVisible();
+      await expect(pages.conversation().getMessage(`Hello from ${member2.firstName}!`)).toBeVisible();
     });
 
     await test.step('Team owner and group members react on received messages with reactions', async () => {
       // Owner reacts to member1's message with +1 (thumbs up)
       await pages.conversationList().openConversation(conversationName);
-      const member1MessageForOwner = pages.conversation().getMessageByText(`Hello from ${member1.firstName}!`).first();
+      const member1MessageForOwner = pages.conversation().getMessage(`Hello from ${member1.firstName}!`).first();
```
aweiss-dev: the .first() is not needed right, since the last message is already returned

(in case the function would be called getLastMessage() you might have noticed already just by naming that .first() is not needed

test/e2e_tests/specs/CriticalFlow/addMembersToChat-TC-8631.spec.ts
```diff
@@ -124,80 +124,80 @@ test(
...
       await member1MessageForOwner.waitFor({state: 'visible'}); // Wait for message to be ready
       await pages.conversation().reactOnMessage(member1MessageForOwner, 'plus-one');
 
       // Owner reacts to member2's message with +1 (thumbs up)
-      const member2MessageForOwner = pages.conversation().getMessageByText(`Hello from ${member2.firstName}!`).first();
+      const member2MessageForOwner = pages.conversation().getMessage(`Hello from ${member2.firstName}!`).first();
```
aweiss-dev: same here

test/e2e_tests/specs/CriticalFlow/addMembersToChat-TC-8631.spec.ts
```diff
@@ -124,80 +124,80 @@ test(
...
       await pages.conversation().reactOnMessage(member2MessageForOwner, 'plus-one');
 
       // Member1 reacts to owner's message with heart (❤️)
       await member1PageManager.webapp.pages.conversationList().openConversation(conversationName);
       const ownerMessageForMember1 = member1PageManager.webapp.pages
         .conversation()
-        .getMessageByText(`Hello from ${owner.firstName}!`)
+        .getMessage(`Hello from ${owner.firstName}!`)
```
aweiss-dev: and here (and so on)

aweiss-dev (changes requested):
we need to remove the .first() calls

Besides that, renaming / refactoring would be welcome :)

test/e2e_tests/specs/CriticalFlow/addMembersToChat-TC-8631.spec.ts
markbrockhoff: Even without the `.last()` within the implementation it would not have been needed ^^'
But I removed them anyhow ;)

test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
markbrockhoff: Yeah, as already answered in the other comment (in the end we're talking about the same commit cherry-picked between two branches): I prefer to leverage playwrights strict mode in this case and keep the util generic by removing the `.last()` in the implementation. If there would be multiple messages with the same content and possibly even from the same user it's up to the tester to figure out how to differentiate them. (E.g. by changing the test data)

markbrockhoff: I updated the implementation, let me know what you think :)

aweiss-dev (approved):

PR #19760, opened 2025-11-11
test(WPB-19966): write "Reply" regression tests

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19966" title="WPB-19966" target="_blank">TaskWPB-19966</a>  [Web/QA] Write the Reply regression tests in Playwright
  </td></table>
  <br />
 

## Description

This PR adds the regression tests for replying to different kinds of  messages in different scenarios

## Checklist

### DIFF
--- a/test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
+++ b/test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
@@ -68,58 +68,41 @@ export class ConversationRepositoryE2E extends BackendClientE2E {
     }
   }
 
-  async getMLSConversationWithUser(token: string, conversationPartnerId: string, timeout = 10000) {
-    const retryDelay = 1000;
-    const maxRetries = timeout / retryDelay;
-    let mlsConversationId = null;
-
-    for (let attempt = 0; attempt < maxRetries && !mlsConversationId; attempt++) {
-      const listIdsResponse = await this.axiosInstance.post(
-        'conversations/list-ids',
-        {},
-        {
-          headers: {
-            Authorization: `Bearer ${token}`,
-          },
+  async getConversationWithUser(token: string, conversationPartnerId: string) {
+    const listIdsResponse = await this.axiosInstance.post(
+      'conversations/list-ids',
+      {},
+      {
+        headers: {
+          Authorization: `Bearer ${token}`,
         },
-      );
+      },
+    );
 
-      const qualifiedIds = listIdsResponse.data.qualified_conversations;
+    const qualifiedIds = listIdsResponse.data.qualified_conversations;
 
-      if (!qualifiedIds) {
-        throw new Error('No qualified conversations found');
-      }
-
-      const response = await this.axiosInstance.post(
-        'conversations/list',
-        {
-          qualified_ids: qualifiedIds,
-        },
-        {
-          headers: {
-            Authorization: `Bearer ${token}`,
-            'Content-Type': 'application/json',
-          },
-        },
-      );
-
-      // Filtering out conversations that are not MLS;
-      const responseData = response.data;
-      mlsConversationId =
-        responseData.found.find(
-          (conversation: {protocol: string; members: {others: {conversation_role: string; id: string}[]}}) =>
-            conversation.protocol === 'mls' &&
-            conversation.members.others.some(
-              (member: {conversation_role: string; id: string}) =>
-                member.conversation_role === 'wire_member' && member.id === conversationPartnerId,
-            ),
-        )?.qualified_id?.id ?? null;
-
-      await new Promise(resolve => setTimeout(resolve, retryDelay));
+    if (!qualifiedIds) {
+      throw new Error('No qualified conversations found');
     }
-    if (!mlsConversationId) {
-      throw new Error(`No MLS conversation found with user ID: ${conversationPartnerId}`);
-    }
-    return mlsConversationId;
+
+    const response = await this.axiosInstance.post(
+      'conversations/list/v2',
+      {
+        qualified_ids: qualifiedIds,
+      },
+      {
+        headers: {
+          Authorization: `Bearer ${token}`,
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+
+    // Find a conversation with the given user
+    const conversation = response.data.found.find(
+      (conversation: {members: {others: {conversation_role: string; id: string}[]}}) =>
+        conversation.members.others.some((member: {id: string}) => member.id === conversationPartnerId),
+    );
+    return conversation?.qualified_id?.id;
   }
 }
--- a/test/e2e_tests/backend/testServiceClient.e2e.ts
+++ b/test/e2e_tests/backend/testServiceClient.e2e.ts
@@ -51,4 +51,15 @@ export class TestServiceClientE2E {
       text,
     });
   }
+
+  async sendLocation(
+    instanceId: string,
+    conversationId: string,
+    message: {locationName: string; latitude: number; longitude: number; zoom: number},
+  ) {
+    return await this.axiosInstance.post(`/api/v1/instance/${instanceId}/sendLocation`, {
+      conversationId,
+      ...message,
+    });
+  }
 }
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -162,6 +162,12 @@ export class ConversationPage {
     await this.createGroupSubmitButton.click();
   }
 
+  async sendTimedMessage(message: string) {
+    await this.timerMessageButton.click();
+    await this.timerTenSecondsButton.click();
+    await this.sendMessage(message);
+  }
+
   async sendMessageWithUserMention(userFullName: string, messageText?: string) {
     await this.messageInput.fill(`@`);
     await this.page
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -35,6 +35,7 @@ export class ConversationPage {
   readonly createGroupSubmitButton: Locator;
   readonly messageInput: Locator;
   readonly sendMessageButton: Locator;
+  readonly searchButton: Locator;
   readonly conversationTitle: Locator;
   readonly watermark: Locator;
   readonly timerMessageButton: Locator;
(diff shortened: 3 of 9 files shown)


test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -224,6 +236,37 @@ export class ConversationPage {
     return await locator.screenshot();
   }
 
+  /**
+   * Util to get a message in the conversation sent by a given user
+   * @param user Optional parameter to specify user who sent the message. If undefined the last message matching the content will be returned.
```
aweiss-dev: This function exhibits unexpected behaviour for me.

Since it returns the last message by default, the naming convention getMostRecentMessage() would be helpful. 

I personally am a big fan of pure and dedicated functions, as well as descriptive function and variable names.

getLastMessage() => Message
getLastMessageByUser(user) => Message

test/e2e_tests/specs/Reply/reply.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+  },
+  userB: async ({api}, use) => {
+    const userB = getUser();
+    await api.createPersonalUser(userB);
+    await use(userB);
+    await removeCreatedUser(api, userB);
+  },
+});
+
+const createPagesForUser = async (browser: Browser, user: User, options?: {openConversationWith?: User}) => {
```
aweiss-dev: This approach can be shared in a common refactoring. We create pages for multiple users in a lot of test cases :)

test/e2e_tests/specs/Reply/reply.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+test.describe('Reply', () => {
+  test.beforeEach(async ({api, userA, userB}) => {
+    await api.connectUsers(userA, userB);
+  });
+
+  test(
+    'I should not be able to reply to a ping',
+    {tag: ['@TC-8038', '@regression']},
+    async ({browser, userA, userB}) => {
+      const pages = await createPagesForUser(browser, userA, {openConversationWith: userB});
```
aweiss-dev: So, as far as I understand, you always create new pages. Dont these pages have to be cleaned up?

test/e2e_tests/specs/Reply/reply.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+
+      await expect(message.getByTestId('do-reply-message')).not.toBeAttached();
+    },
+  );
+
+  test(
+    'I want to see a placeholder text as quote when original message is not available anymore',
+    {tag: ['@TC-2994', '@regression']},
+    async ({browser, userA, userB}) => {
+      const [userAPages, userBPages] = await Promise.all([
```
aweiss-dev, line 63: This could also be abstracted away in a helper function, which takes an array of users, to be able to be easily shared with other functions.

test/e2e_tests/specs/Reply/reply.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+  });
+
+  test('I want to reply to a file', {tag: ['@TC-3006', '@regression']}, async ({browser, userA, userB}) => {
+    const pages = await createPagesForUser(browser, userA, {openConversationWith: userB});
+    const {page} = pages.conversation();
+    await shareAssetHelper(getTextFilePath(), page, page.getByRole('button', {name: 'Add file'}));
+
+    const messageWithFile = pages.conversation().getMessage('', userA);
+    await pages.conversation().replyToMessage(messageWithFile);
+    await pages.conversation().sendMessage('Reply');
```
aweiss-dev, line 182: Since you always use the same string for each test, it would make sense to abstract it away into a shared const

test/e2e_tests/utils/userActions.ts
```diff
@@ -86,20 +86,21 @@ export const sendTextMessageToConversation = async (
...
+  await pages.groupCreation().setGroupName(conversationName);
+  await pages.startUI().selectUsers(user.map(user => user.username));
+  await pages.groupCreation().clickCreateGroupButton();
 };
 
-export const createChannel = async (pageManager: PageManager, conversationName: string, user: User[]) => {
-  await pageManager.webapp.pages.conversationList().clickCreateGroup();
-  await pageManager.webapp.pages.groupCreation().setGroupName(conversationName);
-  await pageManager.webapp.pages.groupCreation().clickNextButton();
+export const createChannel = async (pages: UserPages, conversationName: string, user: User[]) => {
```
aweiss-dev, line 90: this is propably the change which breaks the other tests

test/e2e_tests/specs/Reply/reply.spec.ts
markbrockhoff: Not necessarily. The pages are bound to the browser context which by itself is bound to the browser fixture. We run all tests in fully parallel mode, so after each test the contexts will be closed automatically closing the pages associated with them beforehand.
But I do agree that this is a bit more intricate knowledge of playwrights inner workings and manually closing at least the manually opened context probably won't hurt :)

test/e2e_tests/specs/Reply/reply.spec.ts
markbrockhoff: I disagree on this point. If the string had any actual meaning shared across the tests I'd agree, however it is just a dummy value required in two places of a test.
(What I thought of doing was to maybe turn some of the tests into data driven ones within a for loop. However, since they are all described as individual test cases and differ in setup and assertion I decided to leave them split.

test/e2e_tests/utils/userActions.ts
markbrockhoff: You think so? But I made sure to update all usages of it and ran tsc to verify types still match. I guess this would have failed if I forgot to update an occurrence, right?

test/e2e_tests/specs/Reply/reply.spec.ts
markbrockhoff: Yes, that's also my plan for the long run. However, I didn't want to abstract prematurely. Otherwise we'll just end up with more redundant utils and abstractions one being more tailored to one use case than the other. That's why I tend to keep something like this local at first and apply the ["Rule of three"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), refactoring once the same logic was required a third time. This gives me the needed insight at which point to abstract and not be to generic nor to specific :)

test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
markbrockhoff: I get your point but would propose to keep the name and instead change the implementation.
You're right returning the last message by default can be confusing as it's not reflected by the name. So I'd just return the locator here even though it might resolve to multiple messages. In most cases the locator will be unique so no extra code will be needed. But if it would resolve to multiple playwrights strict mode would kick in and require this one test to be more specific.

test/e2e_tests/specs/Reply/reply.spec.ts
markbrockhoff: Yes, but this would assume I have the new util depend on the `createPagesForUser()` one which, as stated in an other comment, is I would ideally like to be generic across all tests first. Without the coupling the abstraction would only abstract `Promise.all()` which doesn't need abstraction in my opinion :D
Don't worry, I get the idea of reusing things, but abstracting to early in my past experience typically only made things worse.

test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
```diff
@@ -68,58 +68,41 @@ export class ConversationRepositoryE2E extends BackendClientE2E {
...
-        {
-          qualified_ids: qualifiedIds,
-        },
-        {
-          headers: {
-            Authorization: `Bearer ${token}`,
-            'Content-Type': 'application/json',
-          },
+    const response = await this.axiosInstance.post(
+      'conversations/list/v2',
```
e-maad, line 89: If we are introducing versioning here, maybe it's the right time to introduce it globally for all endpoints.

test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
```diff
@@ -68,58 +68,41 @@ export class ConversationRepositoryE2E extends BackendClientE2E {
...
 
-      await new Promise(resolve => setTimeout(resolve, retryDelay));
-    }
-    if (!mlsConversationId) {
-      throw new Error(`No MLS conversation found with user ID: ${conversationPartnerId}`);
-    }
-    return mlsConversationId;
+    // Find a conversation with the given user
+    const conversation = response.data.found.find(
+      (conversation: {members: {others: {conversation_role: string; id: string}[]}}) =>
```
e-maad, line 103: Do we need this type `{members: {others: {conversation_role: string; id: string}[]}`, can we use existing types, like: `Conversation`

test/e2e_tests/backend/testServiceClient.e2e.ts
```diff
@@ -51,4 +51,15 @@ export class TestServiceClientE2E {
       text,
     });
   }
+
+  async sendLocation(
+    instanceId: string,
+    conversationId: string,
+    message: {locationName: string; latitude: number; longitude: number; zoom: number},
```
e-maad, line 58: Maybe we can check if we already have this type.

e-maad (dismissed):
LGTM 👍 
Added few suggestions

test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
markbrockhoff: That's actually not introducing versioning. Please take a look at the OpenAPI spec for api version 1 (which we're currently using in the tests due to it being the minimum API version in the wire client: https://staging-nginz-https.zinfra.io/v1/api/swagger-ui/#/default/post_conversations_list_v2
The endpoint literally has that path ^^'

test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
markbrockhoff: Please note that this util was already there, I just made it work. Generally speaking though, I'm a fan of only specifying the interface you actually need. However, in this case I'd be a even bigger fan of typing the whole api request and response so TS can use inference instead of us having to provide a type at all :D
So Tldr: We could use an already existing type but can't be sure it really matches so I'd suggest to have no manual types here at all in the near future :)

test/e2e_tests/backend/testServiceClient.e2e.ts
markbrockhoff: We do have a `Location` class, but that's a class extending `Asset` which would be very inconvenient to provide to a util like this and resembles the wrong interface for this use case.

thisisamir98 (approved):

PR #19765, opened 2025-11-12
fix: Do not fetch backend to update MLS conversation properties upon reset event arrival [WPB-21618]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21618" title="WPB-21618" target="_blank">TaskWPB-21618</a>  [Web] Recover from MlsErrorWrongEpoch when decrypting messages
  </td></table>
  <br />
 

## Description

When a reset event arrives from the backend there is a possibility that the new welcome for the conversation that was reset has already arrived before and with that conversation epoch is now 1 not 0.

previously when a reset event arrived we fetched the backend to update the epoch which introduced a race condition. with this PR first we check if the new group_id exists on core crypto (meaning welcome has already arrived) and if so we skip updating the conversation epoch otherwise we set it to 0.

### DIFF
--- a/src/script/repositories/conversation/ConversationRepository.ts
+++ b/src/script/repositories/conversation/ConversationRepository.ts
@@ -4369,8 +4369,14 @@ export class ConversationRepository {
 
       const {new_group_id: newGroupId, group_id: oldGroupId} = eventJson.data;
 
-      await this.core.service?.conversation.wipeMLSConversation(oldGroupId);
-      const existingConversation = this.core.service?.conversation.mlsGroupExistsLocally(newGroupId);
+      const conversationService = this.core.service?.conversation;
+
+      if (!conversationService) {
+        throw new Error('Conversation service is not available!');
+      }
+
+      await conversationService.wipeMLSConversation(oldGroupId);
+      const existingConversation = await conversationService.mlsGroupExistsLocally(newGroupId);
 
       if (existingConversation) {
         this.logger.info(
--- a/src/script/repositories/conversation/ConversationRepository.ts
+++ b/src/script/repositories/conversation/ConversationRepository.ts
@@ -4370,9 +4370,9 @@ export class ConversationRepository {
       const {new_group_id: newGroupId, group_id: oldGroupId} = eventJson.data;
 
       await this.core.service?.conversation.wipeMLSConversation(oldGroupId);
-      const exisitingConversation = this.core.service?.conversation.mlsGroupExistsLocally(newGroupId);
+      const existingConversation = this.core.service?.conversation.mlsGroupExistsLocally(newGroupId);
 
-      if (exisitingConversation) {
+      if (existingConversation) {
         this.logger.info(
           'An MLS conversation with the new group ID already exists on core crypto. no need to update epoch',
         );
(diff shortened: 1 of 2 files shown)


e-maad (approved):
LGTM 👍

aweiss-dev (approved):

src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4359,13 +4359,36 @@ export class ConversationRepository {
...
+          `Received MLS reset event for a conversation that is not MLS capable: ${conversationEntity.id}`,
+        );
         return;
       }
 
-      await this.core.service?.conversation.wipeMLSConversation(eventJson.data.group_id);
+      const {new_group_id: newGroupId, group_id: oldGroupId} = eventJson.data;
+
+      await this.core.service?.conversation.wipeMLSConversation(oldGroupId);
+      const exisitingConversation = this.core.service?.conversation.mlsGroupExistsLocally(newGroupId);
```
arjita-mitra: exisitingConversation is misspelled. Rename to existingConversation

src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4359,13 +4359,36 @@ export class ConversationRepository {
...
+        this.logger.warn(
+          `Received MLS reset event for a conversation that is not MLS capable: ${conversationEntity.id}`,
+        );
         return;
       }
 
-      await this.core.service?.conversation.wipeMLSConversation(eventJson.data.group_id);
+      const {new_group_id: newGroupId, group_id: oldGroupId} = eventJson.data;
+
+      await this.core.service?.conversation.wipeMLSConversation(oldGroupId);
```
arjita-mitra: IMO its better to handle this check inside if block and add a logger.warn  when CC conversation service unavailable.

src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4359,13 +4359,36 @@ export class ConversationRepository {
...
+      const updatedConversation = ConversationMapper.updateProperties(conversationEntity, {
+        epoch: 0,
+        groupId: newGroupId,
+      });
+
+      await this.saveConversationStateInDb(updatedConversation);
+
+      this.logger.info(
+        `Updated conversation group ID from ${oldGroupId} to ${newGroupId} for conversation ${conversationEntity.id} and set epoch to 0`,
+      );
```
arjita-mitra, line 4400: if we don't have any unit tests covering this use case its good to add one.

arjita-mitra (changes requested):

src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/19765/commits/fd56e8941b37e28bfa765d31bbdfab7c7524bb4c

src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/19765/commits/c73fb5470db9b6f70293bb0bd55ef35a3b056758

src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/19765/commits/625f28f37a50a5819c4eb538d10447a3972a35e5

arjita-mitra (approved):

PR #19767, opened 2025-11-13
test: add new fixtures to create users and pages

markbrockhoff:
## Description

In a lot of cases we need to create multiple pages to simulate multiple devices / users within one test. The two new added fixtures `createPage()` and `createUser()` allow us to do that in a save way, ensuring tests are isolated and created contexts / users are correctly cleaned up. (Even if the test fails in the middle)
The fixtures also allow us to generalize some more advanced use cases through the API. For example preventing the Telemetry modal from appearing after every login, reducing the time a test takes by ~5s per login. The design of the fixtures is extensible and more options may be added in the future if they really provide benefit to a lot of tests.
I started by refactoring the login tests, using them as a POC. The ultimate goal would be to refactor all tests creating a uniform codebase but I'd like to avoid doing so within one PR, so please see this as the first step of an incremental refactoring.

> **Note:** Since the Telemetry modal is now skipped by default I added a new test to ensure it still works. Please let me know if you think it should be moved to a different category or expanded somehow.

## Checklist

## Notes for reviewers
- I also refactored some of the POM/COMs I encountered while refactoring the login tests. I removed the "getter functions" in favor of using the locators in assertions directly and removed some unused or only used once functions.

### DIFF
--- a/test/e2e_tests/specs/Edit/edit.spec.ts
+++ b/test/e2e_tests/specs/Edit/edit.spec.ts
@@ -195,11 +195,14 @@ test.describe('Edit', () => {
     {tag: ['@TC-692', '@regression']},
     async ({createPage, userA, userB}) => {
       const [userAPage, userBPage] = await Promise.all([
-        createPage(withLogin(userA), withConversation(userB)),
-        createPage(withLogin(userB), withConversation(userA)),
+        PageManager.from(createPage(withLogin(userA), withConversation(userB))),
+        PageManager.from(createPage(withLogin(userB), withConversation(userA))),
       ]);
-      const userAPages = PageManager.from(userAPage).webapp.pages;
-      const userBPages = PageManager.from(userBPage).webapp.pages;
+
+      // ToDo: Can this boiler plate be abstracted? E.g. overload Pagemanager.from to accept a promise?
+      // Maybe it would be best to split the PageManager a bit, since the .webapp.pages often prevents optimization?
+      const userAPages = userAPage.webapp.pages;
+      const userBPages = userBPage.webapp.pages;
 
       await userAPages.conversation().sendMessage('Test');
       const sentMessage = userAPages.conversation().getMessage({sender: userA});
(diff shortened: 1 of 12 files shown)


test/e2e_tests/specs/LoginSpecs/login.spec.ts
```diff
@@ -17,49 +17,45 @@
...
-    const pages = pageManager.webapp.pages;
-
-    await pageManager.openMainPage();
-    await pages.singleSignOn().enterEmailOnSSOPage(incorrectEmail);
-    await pages.login().inputPassword(incorrectPassword);
-    await pages.login().clickSignInButton();
+    const {pages} = pageManager.webapp;
 
-    const errorMessage = await pages.login().getErrorMessage();
+    await pageManager.openLoginPage();
```
markbrockhoff, line 30: Going to the login page directly instead of opening the main page and being redirected to the sso page and navigating to it from there is much faster, less code and still realistic since it is it's own URL.
I intentionally kept the flow in the test below though to not accidentally remove a test case.

test/e2e_tests/specs/Edit/edit.spec.ts
```diff
@@ -115,39 +77,36 @@ test.describe('Edit', () => {
...
-      await expect(messageOnDevice1).toContainText('Message from device 1');
-      await expect(messageOnDevice2).toContainText('Message from device 1');
+      await deviceA.conversation().sendMessage('Message from device 1');
 
-      await device1.conversation().editMessage(messageOnDevice1);
-      await expect(device1.conversation().messageInput).toContainText('Message from device 1');
-      await device1.conversation().sendMessage('Updated message from device 1');
+      const messageOnDeviceA = deviceA.conversation().getMessage({sender: userA});
+      const messageOnDeviceB = deviceB.conversation().getMessage({sender: userA});
+      await expect(messageOnDeviceA).toContainText('Message from device 1');
```
e-maad, line 92: Can we move these strings to constants ?

e-maad (approved):
LGTM 👍 
Added minor suggestion.

test/e2e_tests/specs/Edit/edit.spec.ts
markbrockhoff: We could, why though? The message is only used two times within the same test, if it was used by multiple tests or the strings were far apart I'd agree, but for such a simple case I find actual values often easier to read and understand than some abstract variable name :)

thisisamir98 (approved):

PR #19768, opened 2025-11-13
fix(conversation): add retry attempts for conversationNotFound Error [WPB-18227]

aweiss-dev:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18227" title="WPB-18227" target="_blank">BugWPB-18227</a>  [Web] Can't open this conversation - modal is shown after login
  </td></table>
  <br />
 

## Description

We show this error when an conversation we try to load from backend is not available anymore (Backend returns a not found error).
This could be triggered in rare cases for fresh users due to race conditions. 

Wire 2025-11-13 at 4_47 PM

## Fix

Added a retry mechanism for fetching the conversation, before accepting the error and continuing to handle the error state. 

## Result

- The retry mechanism handles edge case
- Every other case that causes the backend to deliver us the "not found" - error is handled as before

### DIFF
--- a/src/script/repositories/conversation/ConversationRepository.ts
+++ b/src/script/repositories/conversation/ConversationRepository.ts
@@ -611,7 +611,9 @@ export class ConversationRepository {
       await this.updateParticipatingUserEntities(conversationEntity);
       await this.saveConversation(conversationEntity);
 
-      fetching_conversations[conversationId].forEach(({resolveFn}) => resolveFn(conversationEntity));
+      for (const {resolveFn} of fetching_conversations[conversationId]) {
+        resolveFn(conversationEntity);
+      }
       delete fetching_conversations[conversationId];
 
       return conversationEntity;
@@ -628,11 +630,15 @@ export class ConversationRepository {
           ConversationError.MESSAGE.CONVERSATION_NOT_FOUND,
           originalError,
         );
-        fetching_conversations[conversationId].forEach(({rejectFn}) => rejectFn(error));
-        delete fetching_conversations[conversationId];
 
+        for (const {rejectFn} of fetching_conversations[conversationId]) {
+          rejectFn(error);
+        }
+
+        delete fetching_conversations[conversationId];
         throw error;
       }
+      throw new Error('unkown error encountered', {cause: originalError});
     }
   }
 
--- a/src/script/repositories/conversation/ConversationRepository.ts
+++ b/src/script/repositories/conversation/ConversationRepository.ts
@@ -57,7 +57,7 @@ import {ClientMLSError, ClientMLSErrorLabel} from '@wireapp/core/lib/messagingPr
 import {amplify} from 'amplify';
 import {StatusCodes as HTTP_STATUS} from 'http-status-codes';
 import {container} from 'tsyringe';
-import {flatten} from 'underscore';
+import {flatten, isError} from 'underscore';
 
 import {Account} from '@wireapp/core';
 import {Asset as ProtobufAsset, Confirmation, LegalHoldStatus} from '@wireapp/protocol-messaging';
@@ -615,19 +615,24 @@ export class ConversationRepository {
       delete fetching_conversations[conversationId];
 
       return conversationEntity;
-    } catch (originalError) {
-      if (originalError.code === HTTP_STATUS.NOT_FOUND) {
-        this.deleteConversationLocally(qualifiedId, false);
-      }
-      const error = new ConversationError(
-        ConversationError.TYPE.CONVERSATION_NOT_FOUND,
-        ConversationError.MESSAGE.CONVERSATION_NOT_FOUND,
-        originalError,
-      );
-      fetching_conversations[conversationId].forEach(({rejectFn}) => rejectFn(error));
-      delete fetching_conversations[conversationId];
+    } catch (originalError: unknown) {
+      if (isError(originalError)) {
+        const code =
+          originalError && typeof originalError === 'object' && 'code' in originalError ? originalError.code : null;
+        this.logger.error(originalError.message);
+        if (code === HTTP_STATUS.NOT_FOUND) {
+          await this.deleteConversationLocally(qualifiedId, false);
+        }
+        const error = new ConversationError(
+          ConversationError.TYPE.CONVERSATION_NOT_FOUND,
+          ConversationError.MESSAGE.CONVERSATION_NOT_FOUND,
+          originalError,
+        );
+        fetching_conversations[conversationId].forEach(({rejectFn}) => rejectFn(error));
+        delete fetching_conversations[conversationId];
 
-      throw error;
+        throw error;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/script/view_model/ContentViewModel.ts
```diff
@@ -218,8 +219,25 @@ export class ContentViewModel {
...
 
-  private isConversationNotFoundError(error: any): boolean {
-    return error.type === ConversationError.TYPE.CONVERSATION_NOT_FOUND;
+  private isConversationNotFoundError(error: unknown): boolean {
+    return isError(error) && 'type' in error && error.type === ConversationError.TYPE.CONVERSATION_NOT_FOUND;
+  }
+
+  private async retryFetchConversationWithBackoff(
+    conversationId: QualifiedId,
+    maxRetries: number = 3,
```
arjita-mitra, line 228: is it a standardised reattempt number in our app?

src/script/view_model/ContentViewModel.ts
```diff
@@ -274,8 +291,17 @@ export class ContentViewModel {
       const messageEntity = openFirstSelfMention ? conversationEntity.getFirstUnreadSelfMention() : exposeMessageEntity;
       this.changeConversation(conversationEntity, messageEntity);
       this.showAndNavigate(conversationEntity, openNotificationSettings, filePath);
-    } catch (error: any) {
+    } catch (error: unknown) {
       if (this.isConversationNotFoundError(error)) {
+        // Retry fetching the conversation to handle race conditions
+        const fetchSucceeded = await this.retryFetchConversationWithBackoff(conversation, 5);
```
arjita-mitra: how are we deciding on retry attempts?

src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -615,19 +615,24 @@ export class ConversationRepository {
...
-      }
-      const error = new ConversationError(
-        ConversationError.TYPE.CONVERSATION_NOT_FOUND,
-        ConversationError.MESSAGE.CONVERSATION_NOT_FOUND,
-        originalError,
-      );
-      fetching_conversations[conversationId].forEach(({rejectFn}) => rejectFn(error));
-      delete fetching_conversations[conversationId];
+    } catch (originalError: unknown) {
+      if (isError(originalError)) {
```
arjita-mitra, line 621: Read a bit about isError looks like it expects an error instance, is it possible API can throw plain objects then this block won't run right?

src/script/view_model/ContentViewModel.ts
```diff
@@ -218,8 +219,25 @@ export class ContentViewModel {
     );
   }
 
-  private isConversationNotFoundError(error: any): boolean {
-    return error.type === ConversationError.TYPE.CONVERSATION_NOT_FOUND;
+  private isConversationNotFoundError(error: unknown): boolean {
+    return isError(error) && 'type' in error && error.type === ConversationError.TYPE.CONVERSATION_NOT_FOUND;
```
arjita-mitra, line 223: same question, we are assuming error is an error instance, will it be always true?

src/script/view_model/ContentViewModel.ts
```diff
@@ -218,8 +219,25 @@ export class ContentViewModel {
...
+  }
+
+  private async retryFetchConversationWithBackoff(
+    conversationId: QualifiedId,
+    maxRetries: number = 3,
+    initialDelayMs: number = 100,
+  ): Promise<boolean> {
+    for (let attempt = 0; attempt < maxRetries; attempt++) {
+      try {
+        await new Promise(resolve => setTimeout(resolve, initialDelayMs * (attempt + 1)));
```
arjita-mitra, line 233: I didn't know about backoff retry policy so read a bit. From my high level reading i got the idea its better to use exponential backoff is better than linear this gives the system more time to recover without overwhelming a failing system. Please explain your reasoning to go with linear.

src/script/view_model/ContentViewModel.ts
arjita-mitra, line 228: it is better to avoid a default retry attempt it has following disadvantages-
1. caller has to look up to know real delay
2. if caller rely on them and in future default value changes then the behaviour will change and it will be hard to spot. passing both maxRetries, initialDelayMs explicitely is better here.

src/script/repositories/conversation/ConversationRepository.ts
aweiss-dev: It checks if the object inherits from the Error class.

Since we are in a catch phrase, this must be the case (you only reach a catch by throwing), and in a catch, we separate one specific Error or every other case.

src/script/view_model/ContentViewModel.ts
aweiss-dev: In this function? No, it depends on what you give it. But this typeguard has one job only, check if whatever you give it is a ConversationNotFoundError.

src/script/view_model/ContentViewModel.ts
aweiss-dev: We dont have any standard defined yet

aweiss-dev: I disagree. Defining standard parameters in a function has advantages over not doing it. Especially when it comes to refactorings or changing the standard later.

The caller shouldn't care about what the function does unless there is a special case where the parameters need to be adjusted.
If there is no standardisation, then every caller will have different, arbitrary numbers. It's better to find a standard that works and use that by default.

src/script/view_model/ContentViewModel.ts
aweiss-dev: I decided against an exponential backoff because in this use case, the server is not a concern. The main difference is that if too many clients adopt a linear approach when an endpoint goes down, it can overwhelm the server. 
The tradeoff is that the client needs to wait exponentially longer for an answer.

In this case, where the race condition is solved below 500ms, an exponential strategy is an overcomplication of the codebase.
We want to continue with the code execution as fast as possible, which the linear variant satisfies best.

src/script/view_model/ContentViewModel.ts
aweiss-dev: Thats a good call, I wanted to use standard and missed the removal

src/script/repositories/conversation/ConversationRepository.ts
aweiss-dev: Implemented a fallback to satisfy typescript

src/script/view_model/ContentViewModel.ts
arjita-mitra: i don't totally agree with you. caller can still call with arbitrary numbers and override default. Atleast you removed the caller passing the retry number and using the default so now its better

src/script/view_model/ContentViewModel.ts
arjita-mitra: my question was is it possible a non error instance arrives here then isError(error) will fail and rest of the condition won't be checked right?

aweiss-dev: This is a typeguard. It is a pure function that has one job only and doesnt care about anything else. 

How the case is handled when it returns false, it doesn't know and it is the job of the caller to implement. 

In this case, we only want to show a modal if the error type matches; otherwise, we ignore the error. My refactor just fixed the typeguard implementation, not the logic around it

thisisamir98 (approved):

PR #19769, opened 2025-11-14
feat(Wpb-18861): Selectable tabs, conversation filters

emil-wire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18861" title="WPB-18861" target="_blank">TaskWPB-18861</a>  [Web] add filter for unread messages, mentions and replies
  </td></table>
  <br />
 

## user selectable conversation filters + tabs

Bildschirmfoto 2025-11-14 um 01 25 03
Bildschirmfoto 2025-11-14 um 01 25 10
Bildschirmfoto 2025-11-14 um 01 25 34

This PR is a rewrite of a previous PR that didn't work out so well. Using the feature flag FEATURE_ENABLE_ADVANCED_FILTERS you can turn this on and test it. 

## Checklist


emil-wire:
Bildschirmfoto 2025-11-14 um 13 23 11
Design update

src/script/page/LeftSidebar/panels/Conversations/ConversationTabs/ConversationTabs.tsx
```diff
@@ -78,43 +87,91 @@ interface ConversationTabsProps {
...
-    !conversation.is_archived() && conversation.hasUnread();
+    conversationFilters.notArchived(conversation) && conversationFilters.hasUnread(conversation);
 
   const isTeamCreationEnabled =
     Config.getConfig().FEATURE.ENABLE_TEAM_CREATION &&
     core.backendFeatures.version >= Config.getConfig().MIN_TEAM_CREATION_SUPPORTED_API_VERSION;
 
-  const channelConversationsLength = channelConversations.filter(filterUnreadAndArchivedConversations).length;
-  const groupConversationsLength = groupConversations.filter(filterUnreadAndArchivedConversations).length;
+  const channelConversationsLength = useMemo(
```
nkaramarko: useMemo only helps when the calculation is expensive enough to worry about and its dependencies do not change on every render.
If those arrays are new on every render, then useMemo recomputes every time anyway. In that case, useMemo just adds React bookkeeping overhead with zero gain.

Even if the arrays are referentially stable, these are still cheap O(n) filters which is rarely huge on the UI level. So from a performance POV, most of these are premature micro-optimizations.

I would remove these, but maybe I'm missing some context - others can provide their opinion as well

src/script/page/LeftSidebar/panels/Conversations/ConversationTabs/ConversationTabs.tsx
```diff
@@ -131,6 +188,41 @@ export const ConversationTabs = ({
       Icon: <StarIcon />,
       unreadConversations: totalUnreadFavoriteConversations,
     },
+    {
+      type: SidebarTabs.UNREAD,
+      title: t('conversationLabelUnread'),
+      dataUieName: 'go-unread-view',
+      Icon: <Icon.MarkAsUnreadIcon />,
```
nkaramarko, line 160: @emil-wire MarkAsUnreadIcon is missing on Icon component and locally this breaks the app - popup shows and the only thing to do is reload, which again shows same error. Please fix

nkaramarko (changes requested):

src/script/page/LeftSidebar/panels/Conversations/utils/draftUtils.ts
```diff
@@ -51,8 +51,16 @@ export const conversationHasDraft = (conversation: Conversation): boolean => {
...
+
+    if (!draft) {
+      return false;
+    }
+
+    // Check if plainMessage has actual content (not just whitespace)
+    const plainMessage = draft.plainMessage || '';
+    const hasTextContent = plainMessage.trim().length > 0;
+
+    return hasTextContent;
```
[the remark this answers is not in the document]
emil-wire: this was done on purpose because of UX

src/script/repositories/calling/enum/LeaveCallReason.ts
```diff
@@ -25,7 +25,7 @@ export enum LEAVE_CALL_REASON {
   MANUAL_LEAVE_TO_JOIN_ANOTHER_CALL = 'manual_leave_to_join_another_call',
   MANUAL_LEAVE_BY_UI_CLICK = 'manual_leave_by_ui_click',
   USER_MANUALY_LEFT_CONVERSATION = 'user_manualy_left_conversation',
-  USER_IS_REMOVED_BY_AN_ADMIN_OR_LEFT_ON_ANOTHER_CLIENT = 'user_is_removed_by_an_admin_or_left_on_another_client',
+  USER_IS_REMOVED_FROM_CONVERSATION = 'user_is_removed_from_conversation',
```
EnricoSchw, line 28: I find the old name more descriptive. What has changed? Are there more reasons than these two why someone might have to leave the group?

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1674,6 +1680,11 @@ export class CallingRepository {
       if (call.getSelfParticipant().sharesScreen()) {
         this.stopScreenShare(call.getSelfParticipant(), call.conversation, call);
       }
+
+      // If the user is not part of the conversation, the call must be removed from the state
+      if (reason === LEAVE_CALL_REASON.USER_IS_REMOVED_FROM_CONVERSATION) {
```
EnricoSchw, line 1685: After the call is terminated, it is ended via AVS (see the last line of code). AVS then triggers an `onClose` event. The call state is then cleaned up there in this event handler. Why is it cleaned up this particular case here now? And what makes it so special?

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1692,7 +1703,7 @@ export class CallingRepository {
       call.muteState(shouldMute ? MuteState.SELF_MUTED : MuteState.NOT_MUTED);
       return;
     }
-    if (call.hasWorkingAudioInput === false && call.muteState() !== MuteState.NOT_MUTED) {
+    if (!shouldMute && call.hasWorkingAudioInput === false && call.muteState() !== MuteState.NOT_MUTED) {
```
EnricoSchw, line 1706: Here is the Client now in lead. But the mute state should be managed in AVS, not in the client. If isMute is faulty, for whatever reason, AVS can't correct it because there's no access to the audio in some cases.

The client should always be controlled by AVS.

please compare with:
https://github.com/wireapp/wire-webapp/blob/0f1c57e15cafdb9309c4b3546dd08400a939d74d/src/script/repositories/calling/CallingRepository.ts#L445C3-L445C36

the call mute state is set by AVS via update call back

EnricoSchw (dismissed):

src/script/repositories/calling/enum/LeaveCallReason.ts
EnricoSchw: @zskhan I think this commit is from you

src/script/repositories/calling/CallingRepository.ts
EnricoSchw: @zskhan I think this commit is from you

src/script/repositories/calling/CallingRepository.ts
EnricoSchw: @zskhan I think this commit is from you

emil-wire:
New attempt because of monorepo: https://github.com/wireapp/wire-webapp/pull/20142

PR #19782, opened 2025-11-18
feat: meeting list item states [WPB-20284]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20284" title="WPB-20284" target="_blank">TaskWPB-20284</a>  [Web] Meeting list items for all states 
  </td></table>
  <br />
 

## Description
Meeting list items for all states 
Buttons to start or schedule a meeting - WPB-20285
Meeting options menu - WPB-20286

https://github.com/user-attachments/assets/34076acc-e676-454d-a122-77ca480ae7b3

## Checklist


src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -29,7 +30,9 @@ import {EmptyMeetingList} from 'Components/Meeting/EmptyMeetingList/EmptyMeeting
 import {meetingListContainerStyles, showAllButtonStyles} from 'Components/Meeting/MeetingList/MeetingList.styles';
 import {MeetingListItemGroup} from 'Components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItemGroup';
 import {MeetingTab, MeetingTabs} from 'Components/Meeting/MeetingList/MeetingTabs/MeetingTabs';
-import {getTodayTomorrowLabels, groupByStartHour} from 'Components/Meeting/utils/MeetingDatesHandler';
+import {TodayAndOngoingSection} from 'Components/Meeting/MeetingList/TodayAndOngoingSection/TodayAndOngoingSection';
+import {MEETINGS_PAST, MEETINGS_TODAY, MEETINGS_TOMORROW} from 'Components/Meeting/mocks/MeetingMocks';
+import {getTodayTomorrowLabels, groupByStartHour} from 'Components/Meeting/utils/MeetingDatesUtil';
 import {t} from 'Util/LocalizerUtil';
 
 export interface Meeting {
```
aweiss-dev: Maybe MeetingEntry or MeetingEntity?

I think Meeting is a bit too generic for the interface

src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -38,83 +41,37 @@ export interface Meeting {
   schedule: string;
   conversation_id: string;
   title: string;
+  // Ask iOS and Android about how to identify this status
+  attending?: boolean;
```
aweiss-dev, line 45: Usually a Boolean is not sufficient,

Yes, Maybe, No

src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -126,10 +83,14 @@ export const MeetingList = () => {
   }
 
   if (isNextTab) {
```
aweiss-dev: Instead of next I would rather use the word upcoming. Next is often used to describe an action, but we are looking for upcoming meetings

src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -126,10 +83,14 @@ export const MeetingList = () => {
...
-    // Next tab
     content = hasMeetingsToday ? (
```
aweiss-dev: Better use early return with regular if instead of the shorthand. It makes the code more readable

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingListItem.styles.ts
```diff
@@ -20,6 +20,7 @@
 import {CSSObject} from '@emotion/react';
 
 export const itemStyles: CSSObject = {
+  fontSize: '14px',
```
aweiss-dev: Since our app has accessible font sizes that can be adjusted, we need to work with em or rem instead of px

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingListItem.tsx
```diff
@@ -27,31 +30,73 @@ import {
...
+  const meetingStatus = useMemo(
+    () => getMeetingStatusAt(nowMs, start_date, end_date, attending),
+    [nowMs, start_date, end_date, attending],
+  );
+
+  const isOngoing = meetingStatus === MeetingStatuses.ON_GOING || meetingStatus === MeetingStatuses.PARTICIPATING;
 
   return (
-    <div css={itemStyles}>
+    <div css={[itemStyles, isOngoing && onGoingMeetingStyles]}>
```
aweiss-dev: This will print false into the css attribute for falsely cases

Emotion might strip them, we should double check, but keep that in mind when you use that syntax in class names or other parts of the app. 

It's not advices to use

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingListItem.tsx
```diff
@@ -61,8 +106,12 @@ export const MeetingListItem = ({title, start_date, end_date, schedule}: Meeting
...
       </div>
       <div css={rightStyles}>
+        <MeetingStatus start_date={start_date} end_date={end_date} attending={attending} />
         <MeetingAction />
       </div>
     </div>
   );
 };
+
+export const MeetingListItem = memo(MeetingListItemComponent);
```
aweiss-dev, line 118: Is there a reason why you decided to memo the full component here?

In react 19 it is not advices to memo components unless absolutely necessary, since the compiler takes care of it.

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.styles.ts
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CSSObject} from '@emotion/react/dist/emotion-react.cjs';
+
+export const participatingStatusStyles = {
+  color: 'var(--accent-color)',
+  fontSize: '14px',
```
aweiss-dev: No pixel sizes please

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.styles.ts
```diff
@@ -0,0 +1,59 @@
...
+export const joinButtonContainerStyles = {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+};
+
+export const joinButtonStyles = {
+  height: '32px',
+  borderRadius: '8px',
+  fontSize: '14px',
```
aweiss-dev: ⬆️

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+const MeetingStatusComponent = ({start_date, end_date, attending}: MeetingStatusProps) => {
+  const nowMs = useContext(ClockContext);
+
+  const meetingStatus = useMemo(
+    () => getMeetingStatusAt(nowMs, start_date, end_date, attending),
+    [nowMs, start_date, end_date, attending],
+  );
+
+  if (meetingStatus === MeetingStatuses.PARTICIPATING) {
```
aweiss-dev: I would propably use a switch case here, since we already have three entries

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -0,0 +1,82 @@
...
+      const seconds = getCountdownSeconds(nowMs, start_date);
+      return formatSeconds(seconds);
+    }, [nowMs, start_date]);
+    return <div css={startingSoonStatusStyles}>{t('meetings.meetingStatus.startingIn', {countdown})}</div>;
+  }
+
+  return null;
+};
+
+export const MeetingStatus = memo(MeetingStatusComponent);
```
aweiss-dev, line 79: Same question about the memo, if there is no good reason for it let's not add it

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItemGroup.tsx
```diff
@@ -40,7 +42,7 @@ export enum MeetingGroupBy {
   HOUR = 'hour',
 }
 
-export const MeetingListItemGroup = ({
+const MeetingListItemGroupComponent = ({
   header,
   groupedMeetings,
   view = MeetingTabsTitle.NEXT,
```
aweiss-dev: Upcoming

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItemGroup.tsx
```diff
@@ -95,3 +97,6 @@ export const MeetingListItemGroup = ({
     </section>
   );
 };
+
+export const MeetingListItemGroup = memo(MeetingListItemGroupComponent);
```
aweiss-dev, line 101: Same question about the memo

src/script/components/Meeting/MeetingList/TodayAndOngoingSection/TodayAndOngoingSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+          view={MeetingTabsTitle.PAST}
+          groupedMeetings={{0: onGoingMeetings}}
+        />
+      )}
+      <MeetingListItemGroup header={headerForToday} groupedMeetings={groupedMeetingsToday} />
+    </>
+  );
+};
+
+export const TodayAndOngoingSection = memo(TodayAndOngoingSectionComponent);
```
aweiss-dev, line 68: Same question about the memo

I get the feeling you just added them "in case". Let's not do that and let react decide

src/script/components/Meeting/MeetingList/TodayAndOngoingSection/TodayAndOngoingSection.tsx
```diff
@@ -0,0 +1,60 @@
...
+  // Exclude ongoing items from today's grouped list
+  const todayNotOngoing = useMemo(
+    () => meetingsToday.filter(meeting => !ongoingIds.has(meeting.conversation_id)),
+    [meetingsToday, ongoingIds],
+  );
+  const groupedMeetingsToday = useMemo(() => groupByStartHour(todayNotOngoing), [todayNotOngoing]);
+
+  return (
+    <>
+      {onGoingMeetings.length > 0 && (
```
aweiss-dev: Maybe an early return? It's better readable than chaining

src/script/components/Meeting/utils/MeetingStatusUtil.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Meeting} from 'Components/Meeting/MeetingList/MeetingList';
+import {TIME_IN_MILLIS} from 'Util/TimeUtil';
+
+export enum MeetingStatuses {
```
aweiss-dev: The enum should be written uppercase MEETING_STATUS, the entries can be written camelcase.

You mixed it up 😄

src/script/components/Meeting/utils/MeetingStatusUtil.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  if (nowMs >= startMs) {
+    return attending ? MeetingStatuses.PARTICIPATING : MeetingStatuses.ON_GOING;
+  }
+
+  if (startMs <= nowMs + STARTING_SOON_THRESHOLD_MS) {
+    return MeetingStatuses.STARTING_SOON;
+  }
+
+  return MeetingStatuses.UPCOMING;
```
aweiss-dev: Here you used upcoming. Let's use that everywhere and replace next

src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -38,83 +41,37 @@ export interface Meeting {
   schedule: string;
   conversation_id: string;
   title: string;
+  // Ask iOS and Android about how to identify this status
+  attending?: boolean;
 }
 
 export enum MeetingTabsTitle {
   NEXT = 'next',
```
aweiss-dev: Upcoming?

src/script/components/Meeting/MeetingList/MeetingList.tsx
zskhan: Momentarily now sure as there is no backend.
I will take care of it once API is ready.
Meanwhile I will discuss with other clients team how they are handling this.

src/script/components/Meeting/MeetingList/MeetingList.tsx
zskhan: `Next` is actually the tab name in the UI thats why I started with this variable name.

src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingListItem.tsx
zskhan: The reason is on every second I have to refresh the data.
And also currently we are using React 18 so it is really important to avoid heavy re-renders by using memo.

aweiss-dev (changes requested):
As discussed in the 1:1 call, I would like to see a different solution for the part of the code that is required to rerender every second just to display a timer and / or the list. 

This needs to be implemented via "cron-jobs", either directly in the backend (best solution since they can notify all clients, not just web) or via our lowPrecisionTimeScheduler functionality in the webapp. A cron-job like implementation in our application already exists and can be utilised. 
You can also reuse these events to send browser notifications or display information to the user in general troughout our application.

For reference: https://github.com/wireapp/wire-webapp/pull/16935/files

PR #19784, opened 2025-11-20
chore: blocking user spec [19940]

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19940" title="WPB-19940" target="_blank">TaskWPB-19940</a>  [Web/QA] Write the block regression tests in Playwright
  </td></table>
  <br />
 

## Description
Add regression tests for blocking users in eight cases

## Checklist

- [x] mentions the JIRA issue in the PR name ([WPB-19940]

[WPB-19940]: https://wearezeta.atlassian.net/browse/WPB-19940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/e2e_tests/specs/Folders/Folders.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+  },
+});
+
+export {expect} from '@playwright/test';
+
+// Generating test users
+let userA: User;
+let userB: User;
+
+test.describe('Folders', () => {
```
markbrockhoff: Should the folder tests really be part of this PR? Its name suggests something else, maybe they should be split into their own PR? (I'm skipping the review of this file for now as I feel like it should be part of a different PR)

test/e2e_tests/pageManager/webapp/modals/conversationNotConnected.modal.ts
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Page} from '@playwright/test';
+
+import {BaseModal} from './base.modal';
+
+export class ConversationNotConnectedModal extends BaseModal {
```
markbrockhoff: This POM is identical to the already existing `AcknowledgeModal` (see: `test/e2e_tests/pageManager/webapp/modals/acknowledge.modal.ts`). Since I can't seem to find a dedicated ConversationNotConnected Modal component this POM would match we should not create a duplicate abstraction for it :)

test/e2e_tests/pageManager/webapp/modals/createNewFolder.modal.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import {BaseModal} from './base.modal';
+
+export class CreateNewFolderModal extends BaseModal {
+  readonly cancelButton: Locator;
+  readonly createButton: Locator;
+  readonly folderNameInput: Locator;
+
+  constructor(page: Page) {
+    super(page, 'modal-template-option');
```
markbrockhoff, line 29: Are you sure the modal shown here is of type "modal-template-option"? Looking into the function where it's opened it rather looks like a modal of type "input" to me. (Source: src/script/repositories/conversation/ConversationLabelRepository.ts - addConversationToNewLabel)

To explain:
Almost all modals used throughout the application are just sub-versions of the so called `PrimaryModal`. This modal is opened in different variations like `OPTION` which will show the cancel and do action buttons.
Ideally we don't want the POMs (Page Object Models) like this one to macht the real pages / components in the application. In the application there's no "CreateNewFolderModal" but there are multiple occurrences of an option modal just with different titles. So instead of creating a new modal POM for every opened modal, let's try to stick to the kinds of modals which can exist.
You can see all types of modals by looking at the enum `PrimaryModalType` in PrimaryModalTypes.ts

test/e2e_tests/pageManager/webapp/modals/unblockWarning.modal.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.page = page;
+
+    this.modal = page.locator("[data-uie-name='primary-modals-container'][aria-label*='Unblock?']");
+    this.modalTitle = this.modal.locator("[data-uie-name='status-modal-title']");
+    this.modalText = this.modal.locator("[data-uie-name='status-modal-text']");
+    this.unblockButton = this.modal.locator("[data-uie-name='do-action']");
+    this.cancelButton = this.modal.locator("[data-uie-name='do-secondary']");
+  }
+
+  async isModalPresent() {
```
markbrockhoff: Please avoid such "is*" utils on POMs. They will require you to write the expect for it like `expect(await is*()).toBe(true);` which is a bad practice in Playwright. Instead we want to use native assertions like `await expect(something).toBeVisible()`.
To archive that you can just use the locator already available in here instead of creating a function wrapping it, e.g.:
`await expect(modals.unblockWarningModal.modal).toBeVisible()`

test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
```diff
@@ -50,6 +59,15 @@ export class ConversationListPage {
     this.searchConversationsInput = page.locator(selectByDataAttribute('search-conversations'));
     this.archiveConversationMenuButton = page.locator(selectById('btn-archive'));
     this.unarchiveConversationMenuButton = page.locator(selectById('btn-unarchive'));
+    this.moveConversationToFolderButton = page.locator(selectById('btn-move-to'));
+    this.createNewFolderButton = page.locator(selectById('btn-create-new-folder'));
+    this.customFolderButton = page.locator(selectByDataAttribute('conversation-label-context-menu'));
```
markbrockhoff: Are there no better selectors for the buttons than their id? Usually depending on the id of an element is not a really good practice within tests

test/e2e_tests/specs/Block/block2.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  await userAComponents.conversationSidebar().clickPreferencesButton();
+  await userBComponents.conversationSidebar().clickPreferencesButton();
+
+  await expect(userAPages.account().teamName).toHaveText(teamName);
+  await expect(userBPages.account().teamName).toHaveText(teamName);
+});
+
+test.afterAll(async ({api}) => {
+  await tearDownAll(api);
+});
```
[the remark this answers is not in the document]
markbrockhoff: One of the rare occasions a copilot comment is actually right xD
We only added block2.spec.ts while trying out things. Please remove it from the PR, if anything in here is still relevant it should be moved over to block.spec.ts :)

test/e2e_tests/utils/userActions.ts
```diff
@@ -112,3 +113,95 @@ export const handleAppLockState = async (pageManager: PageManager, appLockPassCo
     }
   }
 };
+
+/**
+ * Logs in a user and handles initial setup
+ */
+export async function loginAndSetup(user: User, pageManager: PageManager) {
```
markbrockhoff: Please prefer the fixtures `createUser()` and `createPage()` with the plugins `withLogin` instead of creating a new util. You can find examples on how to use them in `edit.spec.ts`

test/e2e_tests/utils/userActions.ts
```diff
@@ -112,3 +113,95 @@ export const handleAppLockState = async (pageManager: PageManager, appLockPassCo
...
+) {
+  const {modals: userAModals, components: userAComponents, pages: userAPages} = userAPageManager.webapp;
+  const {pages: userBPages} = userBPageManager.webapp;
+
+  await userAComponents.conversationSidebar().clickConnectButton();
+  await userAPages.startUI().searchInput.fill(userB.username);
+  await userAPages.startUI().selectUser(userB.username);
+  await userAModals.userProfile().clickConnectButton();
+
+  expect(await userAPages.conversationList().isConversationItemVisible(userB.fullName)).toBeTruthy();
```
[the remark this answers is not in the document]
markbrockhoff: Copilot is right about the await but that's just because you tried to re-use the `isConversationItemVisible` function which is one of the "is*" functions we'd like to slowly get rid of.
Generally speaking I'd advise against using an `expect` within a util. The util should only perform the stated action, waiting before continuing is not part of it's responsibility. Either the test should do the expect after triggering the action, or the action performed after it should be stable enough to only continue once ready.

test/e2e_tests/utils/userActions.ts
```diff
@@ -112,3 +113,95 @@ export const handleAppLockState = async (pageManager: PageManager, appLockPassCo
...
+  const {pages, modals} = pageManager.webapp;
+  const {handleUnableToOpenModal = false} = options;
+
+  await pages.conversationList().openConversation(userToBlock.fullName);
+  await pages.conversationList().clickConversationOptions(userToBlock.fullName);
+  await pages.conversationList().clickBlockConversation();
+  await modals.blockWarning().clickBlock();
+
+  // Optional handling for modals that appear after blocking
+  if (handleUnableToOpenModal) {
```
markbrockhoff: This sound's more like the workaround for a bug than a feature to me xD
Is there really a reasonable situation in which a modal should pop up telling you you can't open a conversation which you're already on and just blocked?

test/e2e_tests/utils/userActions.ts
```diff
@@ -112,3 +113,95 @@ export const handleAppLockState = async (pageManager: PageManager, appLockPassCo
...
+    }
+  }
+}
+
+/**
+ * Blocks a user from their profile view (from a 1:1 conversation)
+ * @param pageManager PageManager of the blocking user
+ * @param userToBlock User object of the user to be blocked
+ */
+export async function blockUserFromProfileView(pageManager: PageManager, userToBlock: User) {
```
markbrockhoff: Are you sure it's reasonable to already abstract this function even though it is only used once? Not sure but it doesn't sound to me like it would be re-used a lot, so maybe it would be better to just keep it as a step in the test for now. What do you think?

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        await userBPageManager.openMainPage();
+
+        await test.step('Preconditions: Signing in User A and User B', async () => {
+          await Promise.all([loginAndSetup(userA, userAPageManager), loginAndSetup(userB, userBPageManager)]);
+        });
+      },
+    );
+
+    test.afterEach(async ({api}) => {
+      await tearDownAll(api);
```
markbrockhoff: FYI, once #19804 is merged, there will be a fixture available for setting up teams ;)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        // Step 1: User A opens conversation with User B
+        await userAPages.conversationList().openConversation(userB.fullName);
+        // Step 2: User A opens the options menu for user B
+        await userAPages.conversationList().clickConversationOptions(userB.fullName);
+        // Step 3: User A opens modal and clicks 'Block' button
+        await userAPages.conversationList().clickBlockConversation();
+        // Step 4: User A clicks 'Cancel' button
+        await userAModals.blockWarning().clickCancel();
+        // Step 5: Conversation is still present, and User A can open it
+        await userAPages.conversationList().openConversation(userB.fullName);
```
markbrockhoff, line 104: This test doesn't have any assertion, you just rely on the call to openConversation to not fail but this is not a valid way of testing the behavior. For example: If you change line 104 to call `clickBlock()` instead of `clickCancel()` the test still passes.

Tip: Always try to make your test fail once before making it pass ;)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        await userAPages.conversationDetails().clickBlockConversationButton();
+        // Step 4: User A clicks 'Confirm' button
+        await userAModals.blockWarning().clickBlock();
+        // Step 5: User A cannot send message to User B
+        await userAPages.conversationList().openConversation(userB.fullName);
+        await expect(userAPages.conversation().messageInput).not.toBeVisible();
+        // Step 6: User B cannot send message to User A
+        await userBPages.conversationList().openConversation(userA.fullName);
+        await userBPages.conversation().sendMessage('Message after block');
+        expect(await userAPages.conversation().messageCount()).toBe(0);
```
markbrockhoff: Hint: A more reliable way of checking the number of messages would be:
```suggestion
        await expect(userAPages.conversation().messages).toHaveCount(0);
```

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        await test.step('Preconditions: Users A and B are in a group', async () => {
+          await userAComponents.conversationSidebar().isPageLoaded();
+          await userAPages.conversationList().clickCreateGroup();
+          await userAPages.groupCreation().setGroupName(conversationName);
+          await userAPages.startUI().selectUsers([userB.username]);
+          await userAPages.groupCreation().clickCreateGroupButton();
+          await userBPages.conversationList().openConversation(conversationName);
+        });
+
+        // Step 1: User B sends message to group chat with User A
```
markbrockhoff: [Nitpick] Maybe it's sufficient to only have a comment or a step with a description and not both ;)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+
+      test(
+        'Verify you can block a user you sent a connection request from conversation list 0',
+        {tag: ['@TC-144', '@regression']},
+        async ({pageManager: userAPageManager, userBPageManager}) => {
+          const {pages: userBPages} = userBPageManager.webapp;
+
+          // Step 1: User A sends connection request to User B
+          await test.step('Preconditions: User A connects with User B', async () => {
+            await connectUsersManually(userA, userB, userAPageManager, userBPageManager, false);
```
markbrockhoff: Are you sure this util is the right one to use here? To me the test title reads like the user should be blocked before the connection request is accepted, otherwise there wouldn't be a difference to previous tests or would it?

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+          });
+
+          // Step 2: User A blocks User B from conversation list
+          await blockUserFromConversationList(userAPageManager, userB);
+          // Step 3: User B does not see Connection Request from User A
+          expect(userBPages.conversationList()).not.toContain(userA.fullName);
```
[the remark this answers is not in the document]
markbrockhoff: Copilot is right, just wanted to point out the same thing. toContain will check if the given substring is part of the string within `expect`. But conversationList returns a class instance so this will always be true. E.g. you can comment out line 255 and not block the user and the test would still pass.

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        },
+      );
+    });
+
+    test(
+      'Verify you can unblock someone from conversation list options',
+      {tag: ['@TC-148', '@regression']},
+      async ({pageManager: userAPageManager, userBPageManager}) => {
+        test.slow();
```
markbrockhoff: I don't think is is necessary now that we increased the default test timeout to 90s. On my maschine the test takes 29s so no need to increase the timeout to 270s ;)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+          await userAPages.conversation().sendMessage('Message after unblocking');
+          await userBPages.conversationList().openConversation(userA.fullName);
+          expect(await userBPages.conversation().messageCount()).toBe(1);
+        });
+
+        await test.step('User A receives message sent by User B', async () => {
+          await userBPages.conversationList().openConversation(userA.fullName);
+          await userBPages.conversation().sendMessage('Message after being unblocked');
+          await userAPages.conversationList().openConversation(userB.fullName);
+          expect(await userBPages.conversation().messageCount()).toBe(2);
```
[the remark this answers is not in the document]
markbrockhoff: Copilot is right

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+    );
+  });
+
+  test.describe('Block: User A and User B are in the same team', () => {
+    test.beforeEach('Team Test', async ({api, pageManager, browser}) => {
+      userA = getUser();
+      userB = getUser();
+      const teamName = 'Team Name';
+      const context = await browser.newContext();
+      const page = await context.newPage();
```
markbrockhoff: Is it really necessary to open a new page just to log in userB and then do nothing with it? I was able to remove the whole context without the test failing.

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+      async ({pageManager: userAPageManager}) => {
+        const {pages: userAPages} = userAPageManager.webapp;
+
+        // Step 1: User A opens conversation with User B
+        await userAPages.conversationList().openConversation(userB.fullName);
+
+        // Step 2: User A opens the options menu for user B
+        await userAPages.conversationList().clickConversationOptions(userB.fullName);
+
+        await expect(userAPages.conversationList().blockConversationMenuButton).not.toBeAttached();
```
markbrockhoff: That's a nice assertion 💪🏻

markbrockhoff (changes requested):

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -0,0 +1,344 @@
...
+        });
+      },
+    );
+
+    test.afterEach(async ({api}) => {
+      await tearDownAll(api);
+    });
+
+    test(
+      'I want to cancel blocking a 1on1 conversation from conversation list 0',
```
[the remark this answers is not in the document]
JacquelineLehner: fixed typo

test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
```diff
@@ -123,4 +135,15 @@ export class ConversationListPage {
     await this.searchConversationsInput.fill(conversationName);
     await this.openConversation(conversationName);
   }
+
+  async getUserAvatarWrapper(user: User): Promise<Locator> {
+    return this.page
```
markbrockhoff: Question: Could you maybe reuse the `getConversationLocator()` function in here to get the conversation itself and then only locate the avatar within it?

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+        await userAPages.conversation().clickConversationInfoButton();
+        // Step 3: User A clicks "Block conversation" button
+        await userAPages.conversationDetails().clickBlockConversationButton();
+        // Step 4: User A clicks 'Confirm' button
+        await userAModals.blockWarning().clickBlock();
+        // Step 5: User A cannot send message to User B
+        await userAPages.conversationList().openConversation(userB.fullName);
+        await expect(userAPages.conversation().messageInput).not.toBeVisible();
+        // Step 6: User B cannot send message to User A
+        await userBPages.conversationList().openConversation(userA.fullName);
```
markbrockhoff, line 134: Checking for the sent message to not be received isn't really reliable, since it creates a race condition where the test could pass simply because the message way not yet received.
My general question to you would be why you even send a message here. I appreciate the effort but checking that the message from a blocked user is not received is not part of the test's scope. My suggestion would by to ensure the user is blocked like in the test above :)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+        const blockedIcon = avatarWrapperUserB.locator('[data-uie-value="blocked"]');
+        await expect(blockedIcon).toBeVisible();
+      },
+    );
+
+    test(
+      'Verify you still receive messages from blocked person in a group chat 0',
+      {tag: ['@TC-141', '@regression']},
+      async ({createPage}) => {
+        const userAContext = await PageManager.from(createPage(withLogin(userA), withConnectionRequest(userB)));
```
markbrockhoff: Why is this constant named `userAContext`? It's not a browser context but a PageManager instance

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+        const conversationName = 'Group Conversation';
+
+        // Preconditions: User B accepts the connection request
+        const userBContext = await PageManager.from(createPage(withLogin(userB)));
+        const userBPages = userBContext.webapp.pages;
+        await userBPages.connectRequest().clickConnectButton();
+
+        // Step 1: User B blocks User A
+        await test.step('User B blocks User A', async () => {
+          await userAPages.conversationList().openConversation(userC.fullName);
```
markbrockhoff, line 245: Could you please add a comment explaining why userA opens a conversation with userC here? I know that it's to avoid the can't open conversation modal since we looked at that together, but others don't :)

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+          await blockUserFromConversationList(userBContext, userA);
+        });
+
+        // Step 2: User A wants to add B to a group chat after being blocked by User B
+        await test.step('Users A tries to add B to a group', async () => {
+          await userAComponents.conversationSidebar().isPageLoaded();
```
markbrockhoff: I'm pretty sure the page is loaded at this point of the test or is there a valid reason to manually check for it here?

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+          await userAPages.conversationList().clickCreateGroup();
+          await userAPages.groupCreation().setGroupName(conversationName);
+          await userAPages.startUI().selectUsers([userB.username]);
+          await userAPages.groupCreation().clickCreateGroupButton();
+
+          // Step 3: Modal 'modalConversationNotConnectedMessageOne' is visible
```
markbrockhoff: [Nitpick] I think this comment is rather misleading due to its "Step 3:" prefix within Step 2

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+      async ({createPage}) => {
+        const userAContext = await PageManager.from(createPage(withLogin(userA), withConnectionRequest(userB)));
+
+        // Preconditions: User B does not accept connection request
+        const userBPages = (await PageManager.from(createPage(withLogin(userB)))).webapp.pages;
+
+        // Step 2: User A blocks User B from conversation list
+        await blockUserFromConversationList(userAContext, userB);
+        // Step 3: User B does not see Connection Request from User A
+        await expect(userBPages.conversationList().getConversationLocator(userA.fullName)).not.toBeVisible();
```
markbrockhoff: This test passes even if the action above is commented out. Please make sure the test can fail, e.g. by making sure the conversation is visible for userB before A blocks him.

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -104,222 +103,241 @@ test.describe('Block', () => {
...
+
+        await test.step('User A unblocks User B from Search List', async () => {
+          await userAComponents.conversationSidebar().clickConnectButton();
+          await userAPages.startUI().searchInput.fill(userB.username);
+          await userAPages.startUI().selectUser(userB.username);
+          await userAModals.userProfile().unblockButton.click();
+          await userAModals.unblockWarningModal().unblockButton.click();
+        });
+
+        await test.step('User B receives message sent by User A', async () => {
```
markbrockhoff, line 310: Is it really part of the tests scope to exchange messages between the two users? I'd argue that if A is able to send a message to B we can be sure B isn't blocked. Making sure a blocked user can't send a message is the responsibility of a different test

test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -17,83 +17,82 @@
...
+/* ==================================== */
+
+test.describe('User Blocking', () => {
+  test.describe('Block: User A and User B are NOT in the same team', () => {
+    let userA: User;
+    let userB: User;
+    let userC: User;
+
+    test.beforeEach(async ({createTeam}) => {
+      const teamA = await createTeam('Team A', {withMembers: 1});
```
markbrockhoff: Since this runs before each test it would be great to parallelize the team creation:
```suggestion
      const [teamA, teamB] = await Promise.all([createTeam('Team A', {withMembers: 1}), createTeam('Team B')]);
```

test/e2e_tests/pageManager/webapp/modals/unblockWarning.modal.ts
```diff
@@ -0,0 +1,48 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Locator, Page} from '@playwright/test';
+
+export class UnblockWarningModal {
```
markbrockhoff: There is no component matching this POM, it's just a wrapper around the "modal-template-confirm" modal.
Something like this would allow us to re-use the POM for more than 2 cases ;)

```ts
export class ConfirmModal extends BaseModal {
  constructor(page: Page) {
    super(page, 'modal-template-confirm');
  }
}
```

This already includes all utils to execute the action or click cancel etc, since they are part of the BaseModal

test/e2e_tests/pageManager/webapp/pages/account.page.ts
```diff
@@ -43,6 +43,7 @@ export class AccountPage {
   readonly resetPasswordButton: Locator;
   readonly receiveNewsletterCheckbox: Locator;
   readonly typingIndicator: Locator;
+  readonly teamName: Locator;
```
markbrockhoff: This variable is only used within this file and never used anywhere else -> [YAGNI](https://de.wikipedia.org/wiki/YAGNI)

test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
```diff
@@ -33,6 +35,10 @@ export class ConversationListPage {
   readonly searchConversationsInput: Locator;
   readonly archiveConversationMenuButton: Locator;
   readonly unarchiveConversationMenuButton: Locator;
+  readonly statusLabel: Locator;
```
markbrockhoff: `statusLabel` and `profilePicture` are initialized but never used -> [YAGNI](https://de.wikipedia.org/wiki/YAGNI)

markbrockhoff (changes requested):

test/e2e_tests/specs/Block/block.spec.ts
JacquelineLehner: The test specifications in Testiny require User A and B sending messages to each other

test/e2e_tests/specs/Block/block.spec.ts
JacquelineLehner: Assertion is changed now. The messages between A and B are required from the Testiny Specifications.

markbrockhoff (dismissed):

e-maad (dismissed):
LGTM 👍

e-maad (approved):

markbrockhoff (approved):

PR #19788, opened 2025-11-20
chore: update glob dependency version [WPB-29104]

nkaramarko:
Fix dependency review failure by forcing glob to a patched, non-vulnerable version.

Details
rimraf@6.1.0 brings in glob@11.0.3, which is affected by GitHub advisory GHSA-5j98-mcp5-4vw2.
This PR adds a Yarn resolutions entry to force glob@11.1.0 (patched version) across the dependency tree.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -36,6 +36,9 @@
     "rimraf": "6.1.0",
     "typescript": "5.6.3"
   },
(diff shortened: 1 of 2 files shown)


e-maad (approved):
LGTM 🚀

PR #19790, opened 2025-11-21
fix: remove call from the call state when a user is left the group [WPB-21813]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21813" title="WPB-21813" target="_blank">BugWPB-21813</a>  [Web] Issues when leaving a group during an ongoing call
  </td></table>
  <br />
 

# Pull Request

## Summary
Call should be removed from the call state once the user left the group with ongoing call.

- What did I change and why?
-- I updated the method of leave call so that the user will not see CallingCell in the conversation list after leaving the group.
- Risks and how to roll out / roll back (e.g. feature flags): No risk in rolling this out.

---

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/src/script/repositories/calling/CallingRepository.ts
+++ b/src/script/repositories/calling/CallingRepository.ts
@@ -35,16 +35,16 @@ import 'webrtc-adapter';
 
 import {
   AUDIO_STATE,
-  ENV as AVS_ENV,
-  STATE as CALL_STATE,
   CALL_TYPE,
   CONV_TYPE,
+  ENV as AVS_ENV,
   ERROR,
   getAvsInstance,
   LOG_LEVEL,
   QUALITY,
   REASON,
   RESOLUTION,
+  STATE as CALL_STATE,
   VIDEO_STATE,
   VSTREAMS,
   Wcall,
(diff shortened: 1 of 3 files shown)


thisisamir98 (approved):

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -35,16 +35,16 @@ import 'webrtc-adapter';
...
   CALL_TYPE,
   CONV_TYPE,
+  ENV as AVS_ENV,
   ERROR,
   getAvsInstance,
   LOG_LEVEL,
   QUALITY,
   REASON,
   RESOLUTION,
+  STATE as CALL_STATE,
```
aweiss-dev, line 47: Could we export the call state as call state immediately? We want clearly named exports in the code

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1674,6 +1680,11 @@ export class CallingRepository {
       if (call.getSelfParticipant().sharesScreen()) {
         this.stopScreenShare(call.getSelfParticipant(), call.conversation, call);
       }
+
+      // If the user is not part of the conversation, the call must be removed from the state
+      if (reason === LEAVE_CALL_REASON.USER_IS_REMOVED_FROM_CONVERSATION) {
```
aweiss-dev, line 1685: This enum would be ripe for a refactor. 

Upper Case just for the enum, camelcase of the enum attributes

aweiss-dev (approved):

src/script/repositories/calling/CallingRepository.ts
zskhan: From the AVS library?

src/script/repositories/calling/CallingRepository.ts
zskhan: Yes I will raise a subsequent PR to refactor it.

PR #19805, opened 2025-11-25
test(WPB-19968): add tests for self deleting messages

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19968" title="WPB-19968" target="_blank">TaskWPB-19968</a>  [Web/QA] Write the Self Deleting Messages regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- Add regression tests for self deleting messages

---

---

## Notes for reviewers

- I noticed that the timed message button is not really disabled but only ignores the click event, so I fixed the accessibility there
- For now I kept all test cases separate like they are defined in Testiny, however we might want to merge some of them since they are very similar

### DIFF
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -53,7 +53,8 @@ export class ConversationPage {
   readonly callButton: Locator;
   readonly conversationInfoButton: Locator;
   readonly pingButton: Locator;
-  readonly messages: Locator; // Only contains message items which have been sent successfully
+  /** Messages in conversation, only contains message items which have been sent successfully */
+  readonly messages: Locator;
   readonly messageDetails: Locator;
   readonly messageItems: Locator;
   readonly filesTab: Locator;
--- a/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -53,7 +53,7 @@ export class ConversationPage {
   readonly callButton: Locator;
   readonly conversationInfoButton: Locator;
   readonly pingButton: Locator;
-  readonly messages: Locator;
+  readonly messages: Locator; // Only contains message items which have been sent successfully
   readonly messageDetails: Locator;
   readonly messageItems: Locator;
   readonly filesTab: Locator;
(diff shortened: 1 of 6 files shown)


test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -52,7 +53,7 @@ export class ConversationPage {
   readonly callButton: Locator;
   readonly conversationInfoButton: Locator;
   readonly pingButton: Locator;
-  readonly messages: Locator;
+  readonly messages: Locator; // Only contains message items which have been sent successfully
```
thisisamir98: might be useful to use a jsdoc style comment instead


```suggestion
  /**
    * Only contains message items which have been sent successfully
  */
  readonly messages: Locator; // Only contains message items which have been sent successfully
```

test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
```diff
@@ -102,6 +104,15 @@ export class ConversationDetailsPage {
     await this.archiveButton.click();
   }
 
+  /** Opens the self deleting messages panel, selects the given value and closes it again */
+  async setSelfDeletingMessages(value: 'Off' | '10 seconds') {
+    await this.selfDeletingMessageButton.click();
+    const selfDeletingMessagesPanel = this.page.locator('#timed-messages');
+    // I tried to use accessible locators but the radio options are currently not accessible
```
thisisamir98:
```suggestion
    // Radio options are currently not accessible to be able to use accessible locators
```

test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
markbrockhoff: Good point, then it will be shown when using the property somewhere 👍

thisisamir98 (approved):

test/e2e_tests/test.fixtures.ts
```diff
@@ -67,12 +68,25 @@ export const test = baseTest.extend<Fixtures>({
   createPage: async ({browser}, use) => {
     const contexts: BrowserContext[] = [];
 
-    await use(async (...setup) => {
-      const context = await browser.newContext();
-      const page = await context.newPage();
-      contexts.push(context);
+    // Due to the function overload firstParam can be either a PagePlugin or a BrowserContext to reuse
+    await use(async (firstParam, ...plugins) => {
```
e-maad, line 72: Can't we break it down into two different functions to improve readability? Like:
`createPageFromContext` and `createPageFromPlugin`

e-maad (approved):
LGTM 👍 
Added one suggestion if you want to address.

test/e2e_tests/test.fixtures.ts
markbrockhoff: I thought about this as well, but then we'd end up with two fixtures for the same purpose. That's why I decided to rather go with an overload, also considering this use case is pretty limited and only relevant a few times where the context needs to be reused.

PR #19807, opened 2025-11-25
ci(WPB-21987): shard playwright tests across multiple runners

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21987" title="WPB-21987" target="_blank">TaskWPB-21987</a>  [WEB/QA] POC - Use sharding to run e2e tests in parallel in CI
  </td></table>
  <br />
 

# Pull Request

## Summary

- Build and run Testservice as part of CI
- Shard playwright tests across multiple runners
- Include regression tests when running tests
- Fix script reporting test results as github comment to work with nested test suites

---

---

## Notes for reviewers

- I updated the node setup within the pipeline to respect the `.nvmrc` so we'll always use the same node version in CI as we do locally


.github/workflows/precommit-crit-flows.yml
```diff
@@ -124,25 +124,75 @@ jobs:
...
-    runs-on: [self-hosted, Linux, X64, office]
-    needs: [deploy_to_aws]
+    if: ${{ !cancelled() && github.repository == 'wireapp/wire-webapp' && github.actor != 'dependabot[bot]' && github.actor != 'dependabot' }}
+    runs-on: ubuntu-latest
+    needs: [deploy_to_aws, build_testservice]
     timeout-minutes: 35
+    strategy:
+      fail-fast: false
+      matrix:
+        shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```
arjita-mitra: Do we always need a fixed matrix? that won't balance the runners for small PR and might overload when there's a reduced capacity. Could we make it dynamic based on number of tests and average duration?

.github/workflows/precommit-crit-flows.yml
```diff
@@ -156,12 +206,49 @@ jobs:
           echo "Using precommit URL: https://wire-webapp-precommit.zinfra.io/"
           curl -s -o /dev/null -w "HTTP %{http_code}\n" https://wire-webapp-precommit.zinfra.io/
 
+      - name: Start Testservice in background
+        run: java -jar testservice/build/libs/testservice-*-all.jar server testservice/config.yml &
```
arjita-mitra: I think its recommended to run a healthcheck loop to wait until it responds before running tests and also need to ensure to kill the background process in the end.

.github/workflows/precommit-crit-flows.yml
markbrockhoff: First of all the number of tests we're executing doesn't depend on how big or small the PR is, it's constant based of the number of tests with the given tags in the repo.
Second, while making the matrix dynamic is possible it would add a lot of extra  scripting and complexity for very little benefit. Simply going with a reasonable number of agents depending on the overall number of tests seems like a way simpler approach to me. Also please consider this is the outcome of a POC, optimizing for something like this would be way out of scope right now.

.github/workflows/precommit-crit-flows.yml
markbrockhoff: About running a loop to wait for the background task to be ready, yes I thought about doing so as well. But I checked the start up time of the testservice which is only ~600ms. Since there's nothing else running on the Agent at this point we can also be very sure it will be relatively constant. So as long as the service starts up this quickly I don't think there's a need to complicate things.

Regarding killing the background process in the end, yeah I can see that point. Usually the agent stopping should clean it up automatically but I agree that it's probably better to clean up the resource explicitly :)

.github/workflows/precommit-crit-flows.yml
```diff
@@ -124,25 +124,75 @@ jobs:
...
+          ref: main
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Gradle Cache
+        uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # SHA of tag v5.0.0
```
markbrockhoff, line 143: Fyi: the only reason I pinned the actions version to a specific commit is because sonar was worried of a supply chain attack. Not sure what's better, having to manually update every time to get e.g. security patches or the risk of malicious code being merged into the action unnoticed.

.github/workflows/precommit-crit-flows.yml
```diff
@@ -156,12 +206,57 @@ jobs:
...
+          echo TESTSERVICE_PID=$! >> "$GITHUB_OUTPUT"
+
       - name: Run critical flow tests
         env:
           # TODO: remove hardcoded precommit env in the future when ephemeral PR envs will exist
           # Overrides URL from .env file
           WEBAPP_URL: https://wire-webapp-precommit.zinfra.io/
-        run: yarn e2e-test --grep "@crit-flow-web"
+          TEST_SERVICE_URL: http://localhost:8080 # Run tests against locally running test service
+        run: yarn e2e-test --shard=${{ matrix.shard }}/${{ strategy.job-total }} --grep "@crit-flow-web" --grep "@regression"
```
nkaramarko: @markbrockhoff this will run intersection of @crit-flow-web and @regression - meaning it will not run all @regression tests and all @crit-flow-web tests? I think grep flags are AND operation, not OR

.github/workflows/precommit-crit-flows.yml
nkaramarko: @markbrockhoff lets stick with manual pinning for now. We could update dependabot workflow to scan all workflow files in .github/workflows/*.yml and make PRs when there is newer versions, so we dont rely on manual updates in the future

.github/workflows/precommit-crit-flows.yml
markbrockhoff: Good catch, I might have been a bit to quick with the grep filtering.
My understanding was that multiple grep filters behave like a OR, but in reality they just overwrite each other. Meaning the critical flow tests were not executed at all ^^'
For an actual OR expression on tags a regex needs to be utilized, so `--grep "@crit-flow-web|@regression"`. I updated the command

.github/workflows/precommit-crit-flows.yml
markbrockhoff: Considering the most recent supply chain attack in the npm ecosystem strict pinning might also not be too bad of an idea xD

test/e2e_tests/scripts/create-playwright-report-summary.js
```diff
@@ -29,58 +29,66 @@ const ansiRegex = new RegExp(`${ESC}\\[[0-?]*[ -/]*[@-~]`, 'g');
...
+      const lastResult = test.results[test.results.length - 1];
+
+      if (lastResult.status !== 'passed' && lastResult.status !== 'skipped') {
+        // Show only the last (final) failure
+        let failureInfo = `<details> \n <summary> ❌ ${title} </summary><br> \n\n  Location: **${specLocation}**\n  Duration: **${lastResult.duration}ms**\n`;
+
+        if (lastResult.errors?.length) {
+          failureInfo += `\n**Errors:**\n`;
+          lastResult.errors.forEach(e => {
+            failureInfo += `\n\`\`\`\n${stripAnsi(e.message)}\n\`\`\``;
```
nkaramarko: @markbrockhoff you forgot to safeguard this stripAnsi function:
stripAnsi(e?.message || '')

markbrockhoff: Not really, I only changed it's indentation ^^'
Small tip in case you don't know already: In Github you can enable a setting to hide whitespace changes:
image

This makes reviewing indentation changes like this a lot easier ;)

markbrockhoff: But I updated the passed parameter anyhow ;)

nkaramarko (dismissed):

nkaramarko (approved):

PR #19826, opened 2025-12-01
test(WPB-19939): add authentication regression tests

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19939" title="WPB-19939" target="_blank">TaskWPB-19939</a>  [Web/QA] Write the authentication regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- Adds the regression tests for authentication

---

---

## Notes for reviewers

- The test case for only allowing TLS v1.3 and higher was skipped for now as currently only v1.2 seems to be enforced. A bug story will be linked or the test updated once it's clarified which minimum version to test for.

### DIFF
--- a/test/e2e_tests/pageManager/webapp/pages/login.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/login.page.ts
@@ -28,6 +28,7 @@ export class LoginPage {
   readonly emailInput: Locator;
   readonly passwordInput: Locator;
   readonly loginErrorText: Locator;
+  readonly publicComputerCheckbox: Locator;
 
   constructor(page: Page) {
     this.page = page;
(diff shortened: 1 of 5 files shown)


test/e2e_tests/pageManager/webapp/pages/login.page.ts
```diff
@@ -36,6 +37,7 @@ export class LoginPage {
     this.emailInput = page.locator('[data-uie-name="enter-email"]');
     this.passwordInput = page.locator('[data-uie-name="enter-password"]');
     this.loginErrorText = page.locator('[data-uie-name="error-message"]');
+    this.publicComputerCheckbox = page.getByText('This is a public computer');
```
e-maad, line 40: I would suggest to use a better locator. Maybe by role or input type.

e-maad (approved):
LGTM 👍 
Added a suggestion

test/e2e_tests/pageManager/webapp/pages/login.page.ts
markbrockhoff: @e-maad I wanted to do so as well, ideally locating the checkbox by label. However turns out the checkboxes are absolutely not accessible ^^'
That's also the reason I have to use `.click()` on the locator instead of the native `.check()` which I'd prefer. Maybe at some point it would be possible to make the underlying component library accessible.

PR #19833, opened 2025-12-03
test(WPB-22129): wait for login to finish before continuing

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22129" title="WPB-22129" target="_blank">TaskWPB-22129</a>  [Web] Investigate and improve login e2e flow (its currently flaky)
  </td></table>
  <br />
 

# Pull Request

## Summary

Stabilize test to avoid flake due to login taking so long. Normally this kind of waiting for a util to finished would be an antipattern. However considering the login sometimes taking a very long time and the fact that we can't easily improve it to be 3 times as fast this is a valid intermediate solution to avoid distractions due to flake. Ideally at some point the login would be fast enough this statement can be removed.

---

---

## Notes for reviewers

- Since the assertion is done right after the login was triggered its timeout needs to be extended to account for the login taking up to 30s

### DIFF
--- a/test/e2e_tests/pageManager/webapp/pages/login.page.ts
+++ b/test/e2e_tests/pageManager/webapp/pages/login.page.ts
@@ -21,6 +21,8 @@ import type {Page, Locator} from '@playwright/test';
 
 import type {User} from 'test/e2e_tests/data/user';
 
+import {webAppPath} from '../..';
+
 export class LoginPage {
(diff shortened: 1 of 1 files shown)


aweiss-dev (approved):

e-maad (approved):
LGTM 👍

PR #19837, opened 2025-12-03
fix: add guard on join/call buttons [WPB-21715].

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21715" title="WPB-21715" target="_blank">BugWPB-21715</a>  Tapping call join button repeatedly leads to incorrect state
  </td></table>
  <br />
 

## Summary
Add guard on join/call buttons on `CallingCell.tsx`, `TitleBar.tsx` and `ConversationList.tsx`.
Clicking on join/call button multiple times will not try to connect call multiple times.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ Prerequisites:
 ## 1. Fetching dependencies and configurations
 
 1. Run `yarn`
-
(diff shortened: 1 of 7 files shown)


e-maad (approved):
LGTM 👍

PR #19871, opened 2025-12-09
fix(WPB-19951): Disable primary action of input modal if empty

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19951" title="WPB-19951" target="_blank">TaskWPB-19951</a>  [Web/QA] Write the Folders regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

This is to improve the accessability of the application, instead of just doing nothing upon clicking the button.

---

---

## Notes for reviewers

-

### DIFF
--- a/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
+++ b/src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
@@ -143,6 +143,9 @@ export const PrimaryModalComponent: FC = () => {
     if (isConfirm) {
       return false;
     }
+    if (isInput) {
+      return !inputActionEnabled;
+    }
     return !inputActionEnabled && !actionEnabled;
   };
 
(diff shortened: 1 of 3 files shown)


src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -143,6 +143,9 @@ export const PrimaryModalComponent: FC = () => {
     if (isConfirm) {
       return false;
     }
+    if (isInput) {
+      return !inputActionEnabled;
```
arjita-mitra, line 147: As discussed its best to also add aria-describeby to help sr user understand why its disabled or aria-live to announce when the button is enabled again. It doesn't have to be part of this PR so approving.

arjita-mitra (approved):

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
markbrockhoff: Thank you, IMHO the solution here should be to mark the input above as required solving a11y and general validation considerations. Optionally providing a custom error message to the input element by calling `setCustomValidity()` on it. But all of that raises further issues like the submit button not being part of the form and so on... That's why as explained I just wanted to do a simple improvement for now unblocking testing. :)

PR #19874, opened 2025-12-09
feat: add file versioning for editable files [WPB-21659]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21659" title="WPB-21659" target="_blank">TaskWPB-21659</a>  [Web] Implement versioning
  </td></table>
  <br />
 

# Pull Request

## Summary
Added file versioning for editable files.

https://github.com/user-attachments/assets/ec985b03-d279-4d3e-9683-b67d390c19b7

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


src/script/components/Modals/FileHistoryModal/hooks/useFileVersions.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+    setIsLoading(true);
+    try {
+      await cellsRepository.promoteNodeDraft({
+        uuid: nodeUuid,
+        versionId: toBeRestoredVersionId,
+      });
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : 'Failed to restore file version';
+      setError(errorMessage);
```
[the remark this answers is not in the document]
e-maad: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/wireapp/wire-webapp/pull/19874#discussion_r2604752949)

src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
```diff
@@ -0,0 +1,226 @@
...
+export const versionButtonCss: CSSObject = {
+  visibility: 'hidden',
+  height: '32px',
+  minHeight: 'auto',
+  minWidth: 'auto',
+  padding: '0px 16px',
+  alignSelf: 'center',
+  marginBottom: '0px',
+  borderRadius: '12px',
+};
```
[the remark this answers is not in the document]
e-maad: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/wireapp/wire-webapp/pull/19874#r2604752964)

src/script/components/Modals/FileHistoryModal/FileVersionItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+          <span css={versionOwnerSpanCss}>{version.ownerName}</span>
+          {version.size}
+        </p>
+      </div>
+      <div css={versionActionsWrapperCss}>
+        <Button
+          variant={ButtonVariant.SECONDARY}
+          css={versionButtonCss}
+          onClick={() => void onDownload(version.downloadUrl)}
+          aria-label={t('cells.versionHistory.downloadAriaLabel', {time: version.time})}
```
arjita-mitra, line 69: Because you set aria-label:

The accessible name of the button is whatever your translation returns
(e.g 'Download version from 10:30').

The visible text “Download” is ignored for the accessible name.

The text in the <p> (time, ownerName, size) is not tied to the button at all.

So a screen reader user focusing the button hears something like:

“Download version from 10:30, button”

**A simple, solid pattern in your setup is:**

Keep the button’s label short (“Download” / “Restore”).

Attach the file description via aria-describedby pointing to the existing text.

`<p css={versionMetaTextCss} id={descriptionId}>`

```
<Button
          variant={ButtonVariant.SECONDARY}
          css={versionButtonCss}
          onClick={() => void onDownload(version.downloadUrl)}
          aria-label={t('cells.versionHistory.download')}          // short action name
          aria-describedby={descriptionId}                         // + file description
        >
```

Now when a screen reader focuses

1. It gets name: “Download” / “Restore”.
3. Plus description: “{ownerName} {size}” and if you also include time in that <p> it will also get time.

src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
```diff
@@ -88,11 +93,13 @@ export const FileFullscreenModal = ({
         onEditModeChange={setIsEditableState}
         isEditable={isEditable}
         id={id}
+        onFileContentRefresh={refreshModalContent}
       />
       {isEditableState && isEditable ? (
-        <FileEditor id={id} />
+        <FileEditor key={refreshKey} id={id} />
       ) : (
         <ModalContent
```
arjita-mitra, line 101: Make sure modal has focus trap to prevent users from tabbing to or interacting with elements on the page behind it. I saw TM app uses `import FocusTrap from 'focus-trap-react';` you could make a use of it.

src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
```diff
@@ -0,0 +1,226 @@
...
+
+// Version item styles
+export const fileVersionItemWrapperCss: CSSObject = {
+  display: 'flex',
+  gap: '12px',
+  padding: '8px 16px',
+  borderRadius: '8px',
+  ':hover': {
+    backgroundColor: 'var(--gray-20)',
```
arjita-mitra, line 125: Add `&:focus-within` along with hover

Then when the user presses Tab, the moment the keyboard focus enters the row:

- The row becomes :focus-within
- Actions appear
- User can tab into the buttons normally

**How it works**

- The button is still focusable even when it’s visually hidden.
- When it receives focus via Tab, the row becomes :focus-within.
- Your :focus-within CSS then shows the actions, so they appear exactly when focus reaches them.

zskhan (approved):

zskhan (approved):

PR #19881, opened 2025-12-11
feat: debounce microphone and camera permission to avoid multiple track initialization [WPB-16648]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-16648" title="WPB-16648" target="_blank">BugWPB-16648</a>  [Web] Microphone selection when in the call
  </td></table>
  <br />
 

## Summary
On device change, there should be only one media track call initilization.
Also During the call if an AV device is added/removed then media devices should be refreshed.

- What did I change and why?

---

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


src/script/repositories/calling/CallingRepository.ts
```diff
@@ -347,7 +350,7 @@ export class CallingRepository {
     return {wCall: this.wCall, wUser: this.wUser};
   }
 
-  private onMediaDevicesRefresh = () => {
+  private onMediaDevicesRefresh = debounce(() => {
```
EnricoSchw: I think, we shouldn't do that... `navigator.mediaDevices.ondevicechange` --> `onMediaDevicesRefresh` is sometimes called very quickly in succession. With debounce, we lose events.

src/script/page/MainContent/panels/preferences/avPreferences/MicrophonePreferences.tsx
```diff
@@ -68,10 +72,12 @@ const MicrophonePreferences = ({streamHandler, refreshStream, hasActiveCall}: Mi
     }
   };
 
+  // Debounce to handle rapid device changes (removeAllDevices + enumerateDevices)
+  const debouncedRequestStream = useDebouncedCallback(requestStream, DEBOUNCE_TIMEOUT);
```
EnricoSchw, line 76: This is a good Idea, I think!

src/script/hooks/useInitializeMediaDevices.ts
```diff
@@ -30,11 +33,18 @@ export const useInitializeMediaDevices = (devicesHandler: MediaDevicesHandler, s
 
   const initializeMediaDevices = useCallback(async () => {
     try {
-      const stream = await streamHandler.requestMediaStreamAccess(true);
-      await devicesHandler?.initializeMediaDevices();
-      if (stream) {
-        stream.getTracks().forEach(track => track.stop());
+      const permissionStatus = await queryBrowserPermission(PermissionType.MICROPHONE);
```
EnricoSchw: There's one important rule:

Only once we've generated a stream do we know we can use a device.

Your test will not work reliably, because permissions only indicate that a device was allowed at some point , not that the currently available device still exists or is functional. So we need to activate a device at least once to be absolutely sure.

src/script/hooks/useInitializeMediaDevices.ts
```diff
@@ -31,10 +31,11 @@ export const useInitializeMediaDevices = (devicesHandler: MediaDevicesHandler, s
   const initializeMediaDevices = useCallback(async () => {
     try {
       const stream = await streamHandler.requestMediaStreamAccess(true);
-      await devicesHandler?.initializeMediaDevices();
+
```
EnricoSchw, line 34: 👍

EnricoSchw (approved):
LGTM

PR #19955, opened 2026-01-02
feat: enhance accessibility in modal components on call view [WPB-21193]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21193" title="WPB-21193" target="_blank">TaskWPB-21193</a>  [Web] Calling - Users have to navigate backwards and in some cases no event is triggered
  </td></table>
  <br />
 

# Pull Request

## Summary
Enhance accessibility in modal components on call view.

- What did I change and why?

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -207,21 +209,38 @@ export const PrimaryModalComponent: FC = () => {
...
+      if (closeButtonRef.current) {
+        closeButtonRef.current.focus();
+      }
+    }, 0);
+
+    return () => clearTimeout(timeoutId);
+  }, [isModalVisible]);
+
+  useEffect(() => {
+    const onKeyDown = (event: KeyboardEvent) => {
```
arjita-mitra: he keydown effect currently always adds the listener and only checks isModalVisible inside the handler. Would it be possible to attach the listener only when the modal is visible and include all external references in the dependency array [isModalVisible, onKeyDown] . This avoids unnecessary global listeners.

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -207,21 +209,38 @@ export const PrimaryModalComponent: FC = () => {
...
 
+  // Auto-focus close button when modal opens
   useEffect(() => {
-    const onKeyPress = (event: KeyboardEvent) => {
+    if (!isModalVisible) {
+      return undefined;
+    }
+
+    // Use setTimeout to ensure the modal is fully rendered before focusing
+    const timeoutId = setTimeout(() => {
```
arjita-mitra, line 226: Using setTimeout works but useLayoutEffect or requestAnimationFrame is more deterministic and avoids a paint flicker.

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
```diff
@@ -207,21 +209,38 @@ export const PrimaryModalComponent: FC = () => {
...
+      if (closeButtonRef.current) {
+        closeButtonRef.current.focus();
```
arjita-mitra: Focusing the close button every time might not be ideal. 

Better approach-
If primaryBtnFirst is true, focus the primary action button.
Otherwise focus the close button if visible, or the first focusable input. This gives a better keyboard experience.

src/script/components/Modals/PrimaryModal/PrimaryModalShell/PrimaryModalShell.tsx
```diff
@@ -43,14 +46,48 @@ export const PrimaryModalShell = ({
...
 }: PrimaryModalShellProps) => {
   const modalsRef = useRef<HTMLDivElement | null>(null);
 
+  // Make detached window background inert when modal is shown
   useEffect(() => {
+    if (!container) {
+      return undefined;
+    }
+
+    const detachedWindowRoot = (container as HTMLElement).querySelector?.('#detached-window') as HTMLElement;
```
arjita-mitra: I would avoid force casting if possible. Instead can we do this?

const detachedWindowRoot = (container as Element | DocumentFragment)?.querySelector('#detached-window') as HTMLElement | null; if (!detachedWindowRoot) return;

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -2752,10 +2774,42 @@ export class CallingRepository {
...
     };
     PrimaryModal.show(PrimaryModal.type.ACKNOWLEDGE, modalOptions);
   }
 
+  /**
+   * Helper method to get modal container and restore focus callback
+   * @returns Object containing container, targetDocument, and a focus restoration callback
+   */
+  private getModalContainerAndRestoreFocusCallback(): {
+    container?: HTMLElement;
```
arjita-mitra: can we make the stateless part of this function reusable util thats not calling specific as we want this behaviour across our app?

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -2752,10 +2774,42 @@ export class CallingRepository {
...
+      container: isDetachedWindow && detachedWindow ? detachedWindow.document.body : undefined,
+      targetDocument,
+      restoreFocusCallback: (additionalCallback?: () => void) => () => {
+        // Execute additional callback if provided
+        if (additionalCallback) {
+          additionalCallback();
+        }
+        // Restore focus to the button that triggered the modal
+        if (previouslyFocusedElement && typeof previouslyFocusedElement.focus === 'function') {
+          previouslyFocusedElement.focus();
```
arjita-mitra: If the previouslyFocusedElement was removed or window closed calling focus() can throw. Adding a try/catch here to ensure closing a modal never throws and breaks caller code.

src/script/repositories/calling/CallingRepository.ts
```diff
@@ -2752,10 +2774,42 @@ export class CallingRepository {
...
+   */
+  private getModalContainerAndRestoreFocusCallback(): {
+    container?: HTMLElement;
+    targetDocument: Document;
+    restoreFocusCallback: (additionalCallback?: () => void) => () => void;
+  } {
+    const detachedWindow = this.callState.detachedWindow();
+    const isDetachedWindow = this.callState.viewMode() === CallingViewMode.DETACHED_WINDOW;
+    const targetDocument = isDetachedWindow && detachedWindow ? detachedWindow.document : document;
+    const previouslyFocusedElement = targetDocument.activeElement as HTMLElement;
```
arjita-mitra: You capture previouslyFocusedElement at the time getModalContainerAndRestoreFocusCallback is called. If you call this helper long before the modal is shown, the activeElement might change. Please add a comment documenting capture activeElement as late as possible (call this helper right before PrimaryModal.show)

arjita-mitra (changes requested):

src/script/components/Modals/PrimaryModal/PrimaryModal.tsx
zskhan: I tried with useLayoutEffect or requestAnimationFrame but they are not working in the desired way.

arjita-mitra (approved):

PR #19971, opened 2026-01-05
fix: Use production url instead of environment variable to enable asset logging [WPB-22566]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22566" title="WPB-22566" target="_blank">BugWPB-22566</a>  [Web] Schwarz - Asset Audit logs not enabled in production 
  </td></table>
  <br />
 

# Pull Request

## Summary

- What did I change and why?
We should not rely on the webapp build type to control asset logging. Since we don’t want asset logging enabled on the wire.com deployment, I used isProductionWebsite from externalRoute.ts to ensure asset logging is disabled on cloud deployments, without depending on the build type anymore.
 
---

- [X] **External inputs are validated & sanitized** on client and/or server where applicable.
- [X] **API responses are validated**; unexpected shapes are handled safely (fallbacks or errors).
- [X] **No unsafe HTML is rendered**; if unavoidable, sanitization is applied **and** documented where it happens.
- [X] **Injection risks (XSS/SQL/command) are prevented** via safe APIs and/or escaping.

## Accessibility (required)

- [X] I have read and this PR **upholds** our [Accessibility Best Practices](https://github.com/wireapp/wire-webapp/blob/dev/docs/accessibility-practices.md).

- [X] I have read and this PR **upholds** our [Coding Standards](https://github.com/wireapp/wire-webapp/blob/dev/docs/coding-standards.md) and [Tech Radar Choices](https://github.com/wireapp/wire-webapp/blob/dev/docs/tech-radar.md).


aweiss-dev (approved):

src/script/externalRoute.ts
```diff
@@ -24,6 +24,15 @@ const URL = Config.getConfig().URL;
 
 const isProductionWebsite = URL.WEBSITE_BASE && URL.WEBSITE_BASE === 'https://wire.com';
 
+/**
+ * Check if audit logging is allowed for the current backend.
+ * Audit logging must be disabled for the production backend.
+ */
+export const isAuditLogAllowedForBackend = (): boolean => {
+  const {BACKEND_REST} = Config.getConfig();
+  return BACKEND_REST !== 'https://prod-nginz-https.wire.com';
```
zskhan: This URL can be in environment variable.

PR #20045, opened 2026-01-09
test(WPB-19967): Add regression tests for user search functionality

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19967" title="WPB-19967" target="_blank">TaskWPB-19967</a>  [Web/QA] Write the Search regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

Add search regression tests.

---

## Accessibility (required)

---

## Notes for reviewers
The test case TC-1658 regarding search by username with '@' symbol
(mentions context) is currently skipped due to bug [WPB-222687].

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/startUI.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/startUI.page.ts
@@ -66,4 +66,8 @@ export class StartUIPage {
     }
     throw new Error(`User ${username} not found in search results`);
   }
+
+  getSearchResultForSpecificUser(username: String) {
+    return this.page.getByRole('button', {name: `Open profile of ${username}`});
+  }
 }
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/startUI.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/startUI.page.ts
@@ -67,7 +67,7 @@ export class StartUIPage {
     throw new Error(`User ${username} not found in search results`);
   }
 
-  getSearchResultForSpecificUser(username: String) {
+  getSearchResultForSpecificUser(username: string) {
     return this.page.getByRole('button', {name: `Open profile of ${username}`});
   }
 }
--- a/apps/webapp/test/e2e_tests/specs/Search/search.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Search/search.spec.ts
@@ -1,0 +1,93 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


apps/webapp/test/e2e_tests/specs/Search/search.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    {tag: ['@TC-1658', '@regression']},
+    async ({createUser, createPage}) => {
+      const [userA, userB] = await Promise.all([createUser(), createUser()]);
+      const userAPageManager = await PageManager.from(createPage(withLogin(userA)));
+      const userAPages = userAPageManager.webapp.pages;
+
+      const conversationName = 'Group conversation';
+      await createGroup(userAPages, conversationName, [userB]);
+      await userAPages.conversationList().openConversation(conversationName);
+      await userAPages.conversation().sendMessage('@' + userB.username + 'Group message with mention of User B');
```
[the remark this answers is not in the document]
aweiss-dev: you can use ``@${userB.name} Group message with mention of User B`` instead

apps/webapp/test/e2e_tests/pageManager/webapp/pages/startUI.page.ts
```diff
@@ -66,4 +66,8 @@ export class StartUIPage {
     }
     throw new Error(`User ${username} not found in search results`);
   }
+
+  getSearchResultForSpecificUser(username: string) {
+    return this.page.getByRole('button', {name: `Open profile of ${username}`});
```
markbrockhoff, line 71: I didn't try it, but couldn't you use the existing `searchResults` locator instead of looking for a button on the whole page? Feels a bit vague ^^'

If that works you could also use `searchResults` in the assertion directly and wouldn't need the util at all :)

markbrockhoff: Example:

```ts
await expect(userAPages.startUI().searchResults).toContainText(userB.username);
```

instead of: 
```ts
const searchResult = userAPages.startUI().getSearchResultForSpecificUser(userB.fullName);
await expect(searchResult).toBeVisible();
await expect(searchResult).toContainText(userB.fullName);
```

markbrockhoff: It might also be worth it to refactor the `clickUserFromSearchResults` function a bit as it manually does retries etc. which seems overly complicated to me.

JacquelineLehner: I was using the `searchResults` in the assertion at first, but it didn't find User B at all in there, although the user could see it. Therefore, I opted for the solution with the button :)

And I agree, the `clickUserFromSearchResults` function seems too complicated.

markbrockhoff: Hmm, the current locator points only to the usernames of the search results, e.g. `@username`. That's a bit unfortunate. Using the button is already much better but is as pointed out pretty vague as it is based off the whole page.
Let's see if this improves things: #20057 ;)

> Whoever merges last has to rebase :P

aweiss-dev (approved):

PR #20140, opened 2026-01-19
test(WPB-20613): fix critical flow test for 1 on 1 calls TC-8754

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20613" title="WPB-20613" target="_blank">TaskWPB-20613</a>  [Web] Find a solution of failing critical tests due to timeout.
  </td></table>
  <br />
 

# Pull Request

## Summary

- Refactor mocks for audio and video devices -> The fake devices now work and don't just show up in the dropdown
- Fix test for 1 on 1 calls by fixing the video mock and improving the locators for the device dropdowns
- Mute audio of the test browser so e.g. ringtones within a test don't make you think you're being called...

---

## Accessibility (required)

---

## Notes for reviewers

- I noticed that the `Select` component used for the audio / video in- and outputs doesn't reflect the disabled state correctly. This is heavily related to the input tag within it not really being linked to the actual value and options. I had to use a few workarounds but this should be improved at the source.

### DIFF
--- a/apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
+++ b/apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
@@ -26,8 +26,11 @@
  * No matter which of the devices is selected the default input is returned. (The default input is mocked via launchArgs in the playwright config)
  */
 export const mockAudioAndVideoDevices = () => {
+  // If media devices aren't defined on the current device there's nothing to mock
+  if (!navigator.mediaDevices) return;
+
   // Keep a copy of the original function so it can be used within its own stub
-  const originalGetUserMedia = navigator.mediaDevices?.getUserMedia.bind(navigator.mediaDevices);
+  const originalGetUserMedia = navigator.mediaDevices.getUserMedia.bind(navigator.mediaDevices);
 
   // Mock the devices the browser sees, 3 for each device type
   navigator.mediaDevices.enumerateDevices = async () => [
--- a/apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
+++ b/apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
@@ -19,116 +19,93 @@
 
 import {BrowserContext} from '@playwright/test';
 
-// Define a type for our fake devices
-interface FakeDevice {
-  deviceId: string;
-  label: string;
-  kind: MediaDeviceKind;
-}
-
-const fakeVideoDevices: FakeDevice[] = [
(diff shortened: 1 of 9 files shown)


apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
```diff
@@ -17,118 +17,54 @@
...
+      toJSON: () => `{ "deviceId": "video-camera-${i + 1}" }`,
+    })),
+
+    // Create 3 fake audio inputs
+    ...Array.from<unknown, MediaDeviceInfo>({length: 3}, (_, i) => ({
+      deviceId: `audio-input-${i + 1}`,
+      kind: 'audioinput',
+      label: `Fake Audio Input ${i + 1}`,
+      groupId: `audio-input-group-${i + 1}`,
+      toJSON: () => `{ "deviceId": "audio-input-${i + 1}" }`,
```
arjita-mitra: why are you returning a JSON string not an object?

apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
```diff
@@ -17,118 +17,54 @@
...
+ * Init script to mock the available audio and video devices the browser sees.
+ *
+ * Usage: `context.addInitScript(mockAudioAndVideoDevices);`
+ *
+ * This will add 3 devices for audio in- and output as well as 3 cameras.
+ * No matter which of the devices is selected the default input is returned. (The default input is mocked via launchArgs in the playwright config)
+ */
+export const mockAudioAndVideoDevices = () => {
+  // Keep a copy of the original function so it can be used within its own stub
+  const originalGetUserMedia = navigator.mediaDevices?.getUserMedia.bind(navigator.mediaDevices);
```
arjita-mitra: If navigator.mediaDevices or getUserMedia is undefined, originalGetUserMedia will be undefined and later calls will throw an error, its good to handle that and throw an error with a clear message.

apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
markbrockhoff: Whops, valid point (fixed)

apps/webapp/test/e2e_tests/utils/mockVideoDevice.util.ts
markbrockhoff: TBH in reality that can't happen since we're using a browser controlled by us we can be sure it will always be defined. But I get the point so I'll add a small guard to ensure we don't try to mock something which doesn't exist.

arjita-mitra (approved):

PR #20146, opened 2026-01-20
feat: self-deleting messages always off for Cells [WPB-20952]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20952" title="WPB-20952" target="_blank">TaskWPB-20952</a>  [Web] New conversation system messages with Cells on
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-20952

Display a banner in group conversations when self-deleting messages are disabled. The banner appears on conversation creation with a timer icon and localized text.

- Add isSelfDeletingMessagesOff prop to MemberMessage component
- Subscribe to hasGlobalMessageTimer in MessageWrapper
- Show banner when Cells is enabled (self-deleting always off for Cells)
- Add consistent spacing between consecutive info banners
- Add test coverage for banner visibility states

---
Screenshot 2026-01-20 at 13 11 30

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/script/components/MessagesList/Message/MessageWrapper.test.tsx
```diff
@@ -0,0 +1,186 @@
+/*
```
cristianoliveira, line 1: Went ahead and added tests for MessageWrapper test so codecov is happy

apps/webapp/src/script/components/MessagesList/Message/MemberMessage.tsx
```diff
@@ -169,6 +171,17 @@ export const MemberMessage = ({
         </div>
       )}
 
+      {isGroupCreation && isSelfDeletingMessagesOff && (
```
arjita-mitra, line 182: Accessibility improvements needed-

- Since the icon is purely visual, use aria-hidden="true" on the icon wrapper.
- This is a status message, give the container role="status" or aria-live="polite" so assistive tech announces the change when it appears.
- Ensure the header has a semantic structure: consider using a heading element (h2/h3) if it’s a section label, or keep it as a plain div but avoid “header” semantics if it’s not actually a heading(Preferred can skip if too much changes)
- ellipsis truncates content, add title or aria-label with the full text.

Example snippet

```
<div
          className="message-header"
          data-uie-name="label-self-deleting-messages-off"
          role="status"
          aria-live="polite"
        >
          <div className="message-header-icon message-header-icon--svg text-foreground" aria-hidden="true">
            <Icon.TimerIcon />
          </div>
          <p className="message-header-label">
            <span className="ellipsis" title={timedMessagesDisabledLabel}>
              {timedMessagesDisabledLabel}
            </span>
          </p>
        </div>
```

arjita-mitra: apply the same a11y pattern to the other message-header blocks for consistency.

arjita-mitra (changes requested):

apps/webapp/src/script/components/MessagesList/Message/MemberMessage.tsx
cristianoliveira: Done! ✅

arjita-mitra (approved):

PR #20150, opened 2026-01-21
test(WPB-19948): implement delete regression tests

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19948" title="WPB-19948" target="_blank">TaskWPB-19948</a>  [Web/QA] Write the Delete regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- Implemented multiple testcases for the delete messages functionality

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
@@ -17,10 +17,12 @@
  *
  */
 
-import {Locator, Page} from 'playwright/test';
+import {Locator} from 'playwright/test';
 import {User} from 'test/e2e_tests/data/user';
 import {PageManager} from 'test/e2e_tests/pageManager';
 import {test, expect, withConnectedUser, withLogin} from 'test/e2e_tests/test.fixtures';
+import {getAudioFilePath, getTextFilePath, getVideoFilePath, shareAssetHelper} from 'test/e2e_tests/utils/asset.util';
+import {getImageFilePath} from 'test/e2e_tests/utils/sendImage.util';
 import {createGroup} from 'test/e2e_tests/utils/userActions';
 
 test.describe('Delete', () => {
@@ -151,19 +153,20 @@ test.describe('Delete', () => {
     'Message gets deleted even when I was offline on time of deletion',
     {tag: ['@TC-573', '@regression']},
     async ({context, createPage}) => {
-      let userBPage!: Page;
-      let [userAPages, userBPages] = await Promise.all([
-        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
-        PageManager.from(createPage(context, withLogin(userB), withConnectedUser(userA))).then(pm => {userBPage = pm.page; return pm.webapp.pages}),
-      ]);
+      const userAPages = (await PageManager.from(createPage(withLogin(userA)))).webapp.pages;
+      let userBPage = await createPage(context, withLogin(userB));
+      let userBPages = PageManager.from(userBPage).webapp.pages;
 
-      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      await withConnectedUser(userA)(userBPage);
+
+      await userAPages.conversationList().openConversation(userB.fullName);
       await userAPages.conversation().sendMessage('Test Message');
 
       const messageA = userAPages.conversation().getMessage({sender: userA});
       await expect(messageA).toContainText('Test Message');
 
-      const messageB = userBPages.conversation().getMessage({sender: userA});
+      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      let messageB = userBPages.conversation().getMessage({sender: userA});
       await expect(messageB).toContainText('Test Message');
 
       await userBPage.close();
@@ -171,6 +174,8 @@ test.describe('Delete', () => {
 
       userBPages = (await PageManager.from(createPage(context, withLogin(userB)))).webapp.pages;
       await userBPages.conversationList().openConversation(userA.fullName);
+
+      messageB = userBPages.conversation().getMessage({sender: userA});
 
       await expect(messageA).not.toBeAttached();
       await expect(messageB).not.toBeAttached();
@@ -189,15 +194,19 @@ test.describe('Delete', () => {
       await createGroup(userAPages, 'Test Group', [userB]);
       await userBPages.conversationList().openConversation('Test Group');
 
-      await userAPages.conversation();
+      await userAPages.conversation().sendMessage('Test Message');
 
       const messageA = userAPages.conversation().getMessage({sender: userA});
       const messageB = userBPages.conversation().getMessage({sender: userA});
+      await expect(messageA).toContainText('Test Message');
       await expect(messageB).toContainText('Test Message');
 
       await userAPages.conversation().deleteMessage(messageA, 'Everyone');
 
-      const text = await messageB.textContent();
+      // Deletion of a message converts it to a system message with a trash can symbol
+      await expect(userBPages.conversation().systemMessages.getByTitle('Deleted')).toBeAttached();
+
+      await expect(messageB).not.toBeAttached();
       await expect(messageB).not.toBeAttached();
     },
   );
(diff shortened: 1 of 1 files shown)


apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+    await test.step('Create test group', async () => {
+      await createGroup(userAPages, 'Test Group', [userB]);
+      await userBPages.conversationList().openConversation('Test Group');
+    });
+
+    let messageA: Locator;
+    let messageB: Locator;
+
+    await test.step('Send and delete message from user A to B', async () => {
```
markbrockhoff: I think the "and delete" part of the test step could be removed ;)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+      await test.step('Unarchive conversation', async () => {
+        await components.conversationSidebar().clickArchive();
+        await pages.conversationList().clickConversationOptions(userB.fullName);
+        await pages.conversationList().unarchiveConversation();
+        await components.conversationSidebar().clickAllConversationsButton();
+      });
+
+      await test.step('Verify message remains deleted', async () => {
+        await expect(message).not.toBeAttached();
```
markbrockhoff, line 149: Just to be extra save I'd suggest to call `openConversation()` before making that assertion. Just to ensure the conversation is open before looking for the message not to exist :)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+
+  test(
+    'Message gets deleted even when I was offline on time of deletion',
+    {tag: ['@TC-573', '@regression']},
+    async ({context, createPage}) => {
+      const userAPages = (await PageManager.from(createPage(withLogin(userA)))).webapp.pages;
+      let userBPage = await createPage(context, withLogin(userB));
+      let userBPages = PageManager.from(userBPage).webapp.pages;
+
+      await withConnectedUser(userA)(userBPage);
```
markbrockhoff: Tip: You could speed this up by creating both pages in parallel. E.g.:
```suggestion
      const [userAPage, userBPage] = await Promise.all([
        createPage(withLogin(userA), withConnectedUser(userB)),
        createPage(context, withLogin(userB), withConnectedUser(userA)),
      ]);

      const userAPages = PageManager.from(userAPage).webapp.pages;
      let userBPages = PageManager.from(userBPage).webapp.pages;
```

> It might be necessary to specify the protocol when calling `openConversation()` though to ensure we don't run into flake :)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+  test(
+    'I see "Deleted" status message if other user deletes their message "For Everyone" (1:1)',
+    {tag: ['@TC-576', '@regression']},
+    async ({createPage}) => {
+      const [userAPages, userBPages] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      await createGroup(userAPages, 'Test Group', [userB]);
```
markbrockhoff: I think the name of the test implies that it should be executed for a one on one conversation instead of a group ^^'

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+  test('Delete "For Everyone" works for images', {tag: ['@TC-580', '@regression']}, async ({createPage}) => {
+    const [userAPages, userBPages] = await Promise.all([
+      PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+      PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+    ]);
+
+    await userBPages.conversationList().openConversation(userA.fullName);
+
+    const {page} = userAPages.conversation();
+    await shareAssetHelper(getImageFilePath(), page, page.getByRole('button', {name: 'Add picture'}));
```
markbrockhoff: You should probably think about turning this and the very similar tests after it into a parameterized one. Essentially only the action to send the initial message is different. Here's a link to the docs on how to do that within playwright: https://playwright.dev/docs/test-parameterize ;)

> Hint: Please make sure the tag with the test id stays unique and every test has one

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+      await test.step('Send message from user A to B', async () => {
+        await userAPages.conversationList().openConversation(userB.fullName);
+        await userAPages.conversation().sendMessage('Test Message');
+      });
+
+      await test.step('Check user B has a unread conversation with A containing the sent message', async () => {
+        const conversation = userBPages.conversationList().getConversationLocator(userA.fullName);
+        await expect(conversation.getByTestId('status-unread')).toBeVisible();
+
+        await userBPages.conversationList().openConversation(userA.fullName);
```
markbrockhoff: I think you don't need to open the conversation before the message is deleted. The test case is supposed to ensure the unread dot is removed if the unread message was deleted. This is also written in the description of the test case: "If someone sends a message before i read it, and deletes it before i read it - i should not see the Unread count on Conversation in Conversation list"
(You probably didn't see it since you didn't have access to testiny initially :))

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -251,75 +251,89 @@ test.describe('Delete', () => {
...
-    await expect(messageB).not.toBeAttached();
-    await expect(messageB).not.toBeAttached();
-  });
+  type SendActionParams = {
+    pageA: Page;
+    pageB: Page;
+    api: ApiManagerE2E;
+  }
+
+  const testCases = [
```
markbrockhoff: 💡Tip (optional): You could add the the type for the test cases here instead of repeating the definition in each test case:
```ts
const testCases: {
  name: string,
  tc: string,
  sendAction: (params: SendActionParams) => Promise<void>,
}[] = [];
```

Or even inline the `SendActionsParams` type since it would then only be used once. :)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -251,75 +251,89 @@ test.describe('Delete', () => {
...
+    },
+    {
+      name: 'location sharing',
+      tc: '@TC-582',
+      sendAction: async ({pageB, api}: SendActionParams) => {
+        const userBPages = PageManager.from(pageB).webapp.pages;
+
+        await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+
+        await test.step('Prerequisite: Send location via TestService', async () => {
```
markbrockhoff: 🎨 Guess it's more of a stylistic choice, but maybe it would be more uniform to wrap the call to `sendAction()` in the test in a step instead of having a step within one of the actions? Optional nitpic though :)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
```diff
@@ -251,75 +251,89 @@ test.describe('Delete', () => {
...
+      sendAction: async ({pageA}: SendActionParams) => {
+        const userAPages = PageManager.from(pageA).webapp.pages;
+        await userAPages.conversation().sendMessage('https://www.lidl.de/');
+      },
+    },
+    {
+      name: 'location sharing',
+      tc: '@TC-582',
+      sendAction: async ({pageB, api}: SendActionParams) => {
+        const userBPages = PageManager.from(pageB).webapp.pages;
```
markbrockhoff: If you'd use pageA to send the location like in the other cases you wouldn't even need to pass pageP :)

apps/webapp/test/e2e_tests/specs/Delete/delete.spec.ts
panuschek: You're right. I copied the code from the standalone testcase and forgot to remove the test step in the parameter.

markbrockhoff (approved):

markbrockhoff (approved):

aweiss-dev (approved):

PR #20232, opened 2026-01-26
test(WPB-19964): add regression tests for message reactions

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19964" title="WPB-19964" target="_blank">TaskWPB-19964</a>  [Web/QA] Write the Reactions regression tests in Playwright
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19964
# Pull Request

## Summary
Add regression tests for reactions on messages and assets

---

## Accessibility (required)

## Note for reviewers
Code duplication cannot be reduced any further because it only refers to signing in users.

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -89,7 +89,7 @@ export class ConversationPage {
     this.removeUserButton = page.locator(selectByDataAttribute('do-remove-item-text'));
     this.addMemberButton = page.locator(selectByDataAttribute('go-add-people'));
     this.systemMessages = page.locator(
-      `${selectByDataAttribute('item-message')}${selectByClass('system-message')}${selectByDataAttribute('-1', 'send-status')}`,
+      `${selectByDataAttribute('item-message')}${selectByClass('system-message')}:not(${selectByDataAttribute('1', 'send-status')})`,
     );
     this.callButton = page.locator(selectByDataAttribute('do-call'));
     this.conversationInfoButton = page.locator(selectByDataAttribute('do-open-info'));
@@ -99,7 +99,7 @@ export class ConversationPage {
      * Status type -1 ensures that system messages do NOT count as sent messages
      */
     this.messages = page.locator(
-      `${selectByDataAttribute('item-message')}:not(${selectByDataAttribute('1', 'send-status')}):not(${selectByDataAttribute('-1', 'send-status')})`,
+      `${selectByDataAttribute('item-message')}:not(${selectByDataAttribute('1', 'send-status')}):not(${selectByDataAttribute('-1', 'send-status')}):not(${selectByClass('system-message')})`,
     );
     this.messageDetails = page.locator('#message-details');
     this.filesTab = page.locator('#conversation-tab-files');
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -60,6 +60,7 @@ export class ConversationPage {
   readonly itemPendingRequest: Locator;
   readonly ignoreButton: Locator;
   readonly cancelRequest: Locator;
+  readonly reactionWithHeartEmoji: Locator;
 
   readonly getImageAltText = (user: User) => `Image from ${user.fullName}`;
 
@@ -107,6 +108,7 @@ export class ConversationPage {
     this.itemPendingRequest = page.locator(selectByDataAttribute('item-pending-requests'));
     this.ignoreButton = page.getByTestId('do-ignore');
     this.cancelRequest = page.getByTestId('do-cancel-request');
+    this.reactionWithHeartEmoji = page.getByRole('button', {name: /react with heart emoji/i});
   }
 
   protected getImageLocator(user: User): Locator {
@@ -264,12 +266,10 @@ export class ConversationPage {
 
   async reactOnMessage(message: Locator, emojiType: EmojiReaction) {
     await message.hover();
-    const reactionButton = message.getByRole('group').getByRole('button').first();
-    await reactionButton.click();
 
     switch (emojiType) {
       case 'plus-one':
-        // The first quick reaction button is +1 (thumbs up), so we just clicked it
+        await message.getByRole('group').getByRole('button').first().click();
         break;
       case 'heart':
         await this.page.getByTestId('reactwith-love-message').click();
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -88,7 +88,9 @@ export class ConversationPage {
     this.makeAdminToggle = page.locator(selectByDataAttribute('do-allow-admin'));
     this.removeUserButton = page.locator(selectByDataAttribute('do-remove-item-text'));
     this.addMemberButton = page.locator(selectByDataAttribute('go-add-people'));
-    this.systemMessages = page.locator(`${selectByDataAttribute('item-message')}${selectByClass('system-message')}`);
+    this.systemMessages = page.locator(
+      `${selectByDataAttribute('item-message')}${selectByClass('system-message')}${selectByDataAttribute('-1', 'send-status')}`,
+    );
     this.callButton = page.locator(selectByDataAttribute('do-call'));
     this.conversationInfoButton = page.locator(selectByDataAttribute('do-open-info'));
     this.pingButton = page.locator(selectByDataAttribute('do-ping'));
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
@@ -60,7 +60,6 @@ export class ConversationPage {
   readonly itemPendingRequest: Locator;
   readonly ignoreButton: Locator;
   readonly cancelRequest: Locator;
-  readonly reactionWithHeartEmoji: Locator;
 
   readonly getImageAltText = (user: User) => `Image from ${user.fullName}`;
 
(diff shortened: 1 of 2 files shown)


apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {getAudioFilePath, getTextFilePath, getVideoFilePath, shareAssetHelper} from 'test/e2e_tests/utils/asset.util';
+import {getImageFilePath} from 'test/e2e_tests/utils/sendImage.util';
+import {selectByDataAttribute} from '../../utils/selector.util';
+
+type PM = Awaited<ReturnType<typeof PageManager.from>>;
+type Pages = PM['webapp']['pages'];
+
+type LikeCase = {
+  tc: '@TC-1527' | '@TC-1528' | '@TC-1529' | '@TC-1530' | '@TC-1532' | '@TC-1536';
```
markbrockhoff: Nice that you use the features of TS but using a string literal union just for the test ids seems a bit overkill ^^'
Maybe a plain string would do as well, so you don't need to define each id twice.

> Or you could just not define the type at all, `likeCases` will infer it automatically from what you assign to it ;)

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+import {User} from 'test/e2e_tests/data/user';
+import {PageManager} from 'test/e2e_tests/pageManager';
+import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {getAudioFilePath, getTextFilePath, getVideoFilePath, shareAssetHelper} from 'test/e2e_tests/utils/asset.util';
+import {getImageFilePath} from 'test/e2e_tests/utils/sendImage.util';
+import {selectByDataAttribute} from '../../utils/selector.util';
+
+type PM = Awaited<ReturnType<typeof PageManager.from>>;
+type Pages = PM['webapp']['pages'];
```
markbrockhoff: Let's destructure those two lines a bit:
- You take the type returned by `PageManager.from` which is a static constructor so it will return a PageManager or a promise of a PageManager
- You need to wrap it with `Awaited` since it might be a promise
- You then get the type of the nested `pages` prop

So the first two steps get the type of PageManager from the PageManager type ^^'
E.g. why not just use it directly?


```suggestion
type Pages = PageManager['webapp']['pages'];
```

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+    await expect(userAPages.conversation().reactionWithHeartEmoji).toBeVisible();
+  });
+
+  test('Verify you cannot like a system message', {tag: ['@TC-1535', '@regression']}, async ({createPage}) => {
+    const userAPages = (await PageManager.from(createPage(withLogin(userA), withConnectedUser(userB)))).webapp.pages;
+
+    const systemMessage = userAPages
+      .conversation()
+      .page.locator(`${selectByDataAttribute('item-message')}${selectByDataAttribute('-1', 'send-status')}`)
+      .last();
```
markbrockhoff: Couldn't you use the existing locator `userAPages.conversation().systemMessages`? Or maybe adapt it a bit if necessary?

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -60,6 +60,7 @@ export class ConversationPage {
   readonly itemPendingRequest: Locator;
   readonly ignoreButton: Locator;
   readonly cancelRequest: Locator;
+  readonly reactionWithHeartEmoji: Locator;
```
markbrockhoff: IMHO it's too specific to have a locator on the whole POM just for the heart emoji reaction.
Couldn't you use the existing function `getReactionOnMessage()`?

Also since the reaction is part of the message you could do `messageLocator.getByRole('button', { name: 'react with heart emoji' })` instead of searching the whole page for it.

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+      const message = userBPages.conversation().getMessage({sender: userB});
+      await userBPages.conversation().reactOnMessage(message, 'heart');
+
+      const reaction = userAPages.conversation().reactionWithHeartEmoji;
+      await reaction.hover();
+
+      const tooltip = userAPages
+        .conversation()
+        .page.locator('[data-testid="tooltip-content"] div')
```
markbrockhoff: Ouch, that's not a pretty locator, but I see why you had to do it like this as the tooltips implementation also isn't pretty at all. I created https://github.com/wireapp/wire-web-packages/pull/7843 to help with that ;)

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+
+      await userAPages.conversation().reactOnMessage(messageInUserA, 'heart');
+
+      const messageInUserB = userBPages.conversation().getMessage({sender: userB});
+      await userBPages.conversation().reactOnMessage(messageInUserB, 'heart');
+
+      const reactionPill = userAPages.conversation().reactionWithHeartEmoji;
+
+      await expect(reactionPill).toBeVisible();
+      await expect(reactionPill).toHaveAttribute('aria-label', '2 reactions, react with heart emoji');
```
markbrockhoff: Let's try to not assert based on attributes if possible, the label attribute set here isn't displayed to the user so using it makes the assertion less realistic. We're only interested in the number of reactions, so it's sufficient to do `.toHaveText('2')`

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      await userBPages.conversation().sendMessage('Message for emoji picker test');
+      const message = userAPages.conversation().getMessage({sender: userB});
+
+      await message.hover();
+
+      const thumbsUpButton = message.getByRole('button', {name: 'React with thumbs up'});
```
markbrockhoff: I don't think the intention of the test is to simply react to a message and remove the reaction. Next to the default emojis 👍🏻 & ❤️ there's the option to pick from all available emojis. From my understanding this is what the test is supposed to do ^^'

> Hint: You can use the existing util `reactOnMessage()` which can do that e.g. by using the "joy" emoji from the emojis pannel.

markbrockhoff (approved):

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+      const message = userBPages.conversation().getMessage({sender: userB});
+      await userBPages.conversation().reactOnMessage(message, 'heart');
+
+      const reaction = userAPages.conversation().getReactionOnMessage(messageUserB, 'heart');
+      await reaction.hover();
+
+      const tooltip = userAPages
+        .conversation()
+        .page.getByRole('tooltip')
+        .filter({hasText: `${userA.fullName} reacted with`});
```
arjita-mitra, line 198: filters tooltips using the full English text ${userA.fullName} reacted with which can be localized or have different punctuation/format, this makes the assertion brittle.

Can we assert presence of the username inside the tooltip, not an exact English phrase.

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    const reaction = userAPages.conversation().getReactionOnMessage(messageUserB, 'heart');
+    await expect(reaction).toBeVisible();
+
+    await userBPages.conversation().editMessage(userBPages.conversation().getMessage({sender: userB}));
+    await userBPages.conversation().sendMessage('Edited Message');
+    const editedMessage = userAPages.conversation().getMessage({content: 'Edited Message'});
+    await expect(editedMessage).toBeVisible();
+
+    await expect(reaction).not.toBeVisible();
```
arjita-mitra, line 172: Storing an element handle for the reaction before the message is edited and later checking that same handle is not visible can be causing stale element problems. The DOM containing that element may be replaced during edit.

// Re-query the reaction on the edited message to avoid stale element handle
await expect(userAPages.conversation().getReactionOnMessage(editedMessage, 'heart')).not.toBeVisible();

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -88,9 +88,7 @@ export class ConversationPage {
     this.makeAdminToggle = page.locator(selectByDataAttribute('do-allow-admin'));
     this.removeUserButton = page.locator(selectByDataAttribute('do-remove-item-text'));
     this.addMemberButton = page.locator(selectByDataAttribute('go-add-people'));
-    this.systemMessages = page.locator(
-      `${selectByDataAttribute('item-message')}${selectByClass('system-message')} ${selectByClass('message-header')}`,
-    );
+    this.systemMessages = page.locator(`${selectByDataAttribute('item-message')}${selectByClass('system-message')}`);
```
arjita-mitra: why are you removing ${selectByClass('message-header')}? Now you’re targeting the whole system message container (item-message.system-message) any reason to do that?

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
markbrockhoff: This is not an element handle, it's a playwright locator. Locators in playwright are re-querried every time an assertion or action is triggered on them. (See: https://playwright.dev/docs/locators#locating-elements)

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
markbrockhoff: IMHO it's totally fine to use english text within locators, it's done all the time. We're not going to re-test the whole application for each supported language, so english is the default language set for all tests. (Except e.g. a test which might test switching the locale but that's an edge case)

JacquelineLehner: I agree with Mark. What do you think, Arjita?

apps/webapp/test/e2e_tests/specs/Reactions/reactions.spec.ts
JacquelineLehner: I agree

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
JacquelineLehner: To align this locator with the other message-locators. I added now the selector via the attribute send status to make it clearer.

arjita-mitra (approved):

PR #20240, opened 2026-01-27
test(WPB-22420): fix TC-60 for SCIM integration

markbrockhoff:
<table>
<td>
  </td></table>
  <br />
 

# Pull Request

## Summary

- Fixes the test for a user managed by SCIM by cleaning up the device added during test after each run

---

## Accessibility (required)

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/AccountSettingsSpecs/accountSettings.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/AccountSettingsSpecs/accountSettings.spec.ts
@@ -84,42 +84,55 @@ test.describe('account settings', () => {
     },
   );
 
-  test(
-    'I should not be able to change email of user managed by SCIM',
-    {tag: ['@TC-60', '@regression']},
-    async ({context, createPage}) => {
-      const ssoUser = getUser({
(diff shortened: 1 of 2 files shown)


apps/webapp/test/e2e_tests/specs/AccountSettingsSpecs/accountSettings.spec.ts
```diff
@@ -84,42 +84,55 @@ test.describe('account settings', () => {
...
+    test.beforeEach(async ({context, createPage}) => {
+      pageManager = PageManager.from(await createPage(context));
+    });
+
+    test.afterEach(async () => {
+      const {pages, modals} = pageManager.webapp;
+
+      // Log out the user to ensure the previous login won't affect future runs
+      // If this isn't done the test will start to fail due to too many clients being registered for this one user
+      await pages.sidebar().clickPreferencesButton();
```
markbrockhoff, line 104: We don't need to worry about the case where the test failed because the login wasn't successful. If the login fails there's nothing to clean up, if something after the login failed or the test succeeded this will clean up the device as intended.

arjita-mitra (approved):

PR #20247, opened 2026-01-28
test(WPB-19963): Add e2e regression tests for sending a ping

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19963" title="WPB-19963" target="_blank">TaskWPB-19963</a>  [Web/QA] Write the Ping regression tests in Playwright
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-19963

# Pull Request

## Summary

Add regression tests for sending ping
---

## Accessibility (required)

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
@@ -1,0 +1,80 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {User} from 'test/e2e_tests/data/user';
+import {PageManager} from 'test/e2e_tests/pageManager';
+import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {createGroup} from '../../utils/userActions';
+
+test.describe('Ping', () => {
+  let userA: User;
+  let userB: User;
+
+  test.beforeEach(async ({createTeam}) => {
+    const team = await createTeam('Test Team', {withMembers: 1});
+    userA = team.owner;
+    userB = team.members[0];
+  });
+
+  test('Verify I can receive ping in 1:1', {tag: ['@TC-1490', '@regression']}, async ({createPage}) => {
+    const [userAPages, userBPages] = await Promise.all([
+      PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+      PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+    ]);
+
+    await userBPages.conversationList().openConversation(userA.fullName);
+    await userBPages.conversation().sendPing();
+
+    await userAPages.conversationList().openConversation(userB.fullName);
+    await expect(userAPages.conversation().getPing()).toBeVisible();
+  });
+
+  test('Verify I can receive ping several times in a row', {tag: ['@TC-1491', '@regression']}, async ({createPage}) => {
+    const [userAPages, userBPages] = await Promise.all([
+      PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+      PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+    ]);
+
+    await userBPages.conversationList().openConversation(userA.fullName);
+    await userBPages.conversation().sendPing();
+    await userBPages.conversation().sendPing();
+    await userBPages.conversation().sendPing();
+    await userBPages.conversation().sendPing();
+    await userBPages.conversation().sendPing();
+
+    await userAPages.conversationList().openConversation(userB.fullName);
+    await expect(userAPages.conversation().getPing()).toHaveCount(5);
+  });
+
+  test('Verify I can receive ping in group conversation', {tag: ['@TC-1492', '@regression']}, async ({createPage}) => {
+    const [userAPages, userBPages] = await Promise.all([
+      PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+      PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+    ]);
+
+    const conversationName = 'Test Group';
+    await createGroup(userAPages, conversationName, [userB]);
+
+    await userBPages.conversationList().openConversation(conversationName);
+    await userBPages.conversation().sendPing();
+
+    await userAPages.conversationList().openConversation(conversationName);
+    await expect(userAPages.conversation().getPing()).toBeVisible();
+  });
+});
--- a/apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
@@ -19,11 +19,11 @@
 
 import {User} from 'test/e2e_tests/data/user';
 import {PageManager} from 'test/e2e_tests/pageManager';
-import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {test, expect, withLogin, withConnectedUser, Team} from 'test/e2e_tests/test.fixtures';
 import {createGroup} from '../../utils/userActions';
 
 test.describe('Ping', () => {
-  let team: any;
+  let team: Team;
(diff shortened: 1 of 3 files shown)


apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {createGroup} from '../../utils/userActions';
+
+test.describe('Ping', () => {
+  let userA: User;
+  let userB: User;
+  let userC: User;
+  let userD: User;
+  let userE: User;
+  let userF: User;
```
markbrockhoff: The users C,D,E and F only get used in one of the tests, please only create them for this one test instead of every test.

> The team has a `addMember()` function to add a new member to it e.g.: `team.addMember(await createUser())`

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+  const pingScenarios = [
+    {
+      id: '@TC-1490',
+      title: 'Verify I can receive ping in 1:1',
+      isGroup: false,
+      pingCount: 1,
+    },
+    {
+      id: '@TC-1491',
+      title: 'Verify I can receive ping several times in a row',
```
markbrockhoff: IMHO this one doesn't really fit into the parameterized test. Maybe it's just me but if there are too many parameters which are not all permutations of the same test I think readability suffers.
Don't force yourself to merge tests just because they e.g. have a similar setup and sonar complains about test coverage, if two tests are testing different things, they should be separate tests.

E.g. in this case I'd have this parameterized test only change whether a the ping happens in a 1on1 or Group conversation and have TC-1491 be its own test.

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    test(scenario.title, {tag: [scenario.id, '@regression']}, async ({createPage}) => {
+      const [userAPages, userBPages] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+      ]);
+
+      let conversationNameForB = userA.fullName;
+      let conversationNameForA = userB.fullName;
+
+      if (scenario.isGroup) {
```
markbrockhoff: It's a good practice to not have stand alone if statements inside a test, meaning there should always also be a else block. E.g. in this case you could inline the part where the conversation is opened for both users and either open the group conversation or the 1on1s:

```ts
if (scenario.isGroup) {
        await createGroup(userAPages, 'Test Group', [userB]);
        await userAPages.conversationList().openConversation('Test Group');
        await userBPages.conversationList().openConversation('Test Group');
      } else {
        await userAPages.conversationList().openConversation(userB.fullName);
        await userBPages.conversationList().openConversation(userA.fullName);
      }
```

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    'Verify I see a warning when I ping in a big group',
+    {tag: ['@TC-1496', '@regression']},
+    async ({createPage}) => {
+      const userAPage = await createPage(
+        withLogin(userA),
+        withConnectedUser(userB),
+        withConnectedUser(userC),
+        withConnectedUser(userD),
+        withConnectedUser(userE),
+        withConnectedUser(userF),
```
markbrockhoff: I don't think you need to connect userA to all the other users. They are all part of the same team so they will automatically show up when creating the group. This will speed up the test quiet a bit ;)

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+      );
+      const userAPages = PageManager.from(userAPage).webapp.pages;
+      const userAModals = PageManager.from(userAPage).webapp.modals;
```
markbrockhoff: It seems a bit excessive to create two PageManager instances for the same page. How about just using one:

```suggestion
      const {pages: userAPages, modals: userAModals} = PageManager.from(userAPage).webapp;
```

> You could just keep the default names for pages and models instead of renaming them since there are no other users involved in the test

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      const conversationName = 'Test Group';
+      await createGroup(userAPages, conversationName, [userB, userC, userD, userE, userF]);
+
+      await userAPages.conversationList().openConversation(conversationName);
+      await userAPages.conversation().sendPing();
+
+      await expect(userAModals.confirm().modal).toBeVisible();
```
markbrockhoff, line 104: Maybe it would be a good idea to not just assert that some confirm modal is visible but to make sure e.g. it's title contains some text about pinging a big group. Just to make sure it's the right modal ;)

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+        const conversationName = 'Test Group';
+        await createGroup(userAPages, conversationName, [userB]);
+        conversationNameForB = conversationName;
+        conversationNameForA = conversationName;
+      }
+
+      await userBPages.conversationList().openConversation(conversationNameForB);
+
+      for (let i = 0; i < scenario.pingCount; i++) {
+        await userBPages.conversation().sendPing();
```
markbrockhoff: You're just spamming the ping button here, but the test case also mentions to check for the ping button to be disabled between pings:

Image

> You can try to use the `.toBeDisabled()` assertion of playwright on the ping button (hoping it is accessible enough for that)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    async ({createPage, createUser}) => {
+      const usersForBigGroup = [];
+      for (let i = 0; i<4; i++) {
+        const newMember = await createUser();
+        await team.addMember(newMember);
+        usersForBigGroup.push(newMember);
+      }
+
+      const userAPage = await createPage(withLogin(userA), withConnectedUser(userB));
+      const {pages: userAPages, modals: userAModals} = PageManager.from(userAPage).webapp;
```
arjita-mitra, line 96: PageManager.from() is async, so the destructure runs on a Promise instead of the resolved object. shouldn't we use await here?

JacquelineLehner: Since `userAPage` is not a promise, `PageManager.from` returns no promise, therefore no await is needed.

arjita-mitra (approved):

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  test('Verify I can receive ping several times in a row', {tag: ['@TC-1491', '@regression']}, async ({createPage}) => {
+    const [userAPages, userBPages] = await Promise.all([
+      PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+      PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+    ]);
+
+    await userBPages.conversationList().openConversation(userA.fullName);
+    await userBPages.conversation().sendPing();
+    await expect(userBPages.conversation().pingButton).toBeDisabled();
```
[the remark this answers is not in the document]
markbrockhoff: Playwright will wait for the button to be enabled before trying to click it by default, no need for redundant manual checks

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import {test, expect, withLogin, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+import {createGroup} from '../../utils/userActions';
+
+test.describe('Ping', () => {
+  let team: any;
+  let userA: User;
+  let userB: User;
+
+  test.beforeEach(async ({createTeam}) => {
+    team = await createTeam('Test Team', {withMembers: 1});
```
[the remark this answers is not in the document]
markbrockhoff: I 100% agree with copilot on this one. Please don't use the any type and since team is only needed in the beforeEach, keep it local ;)

apps/webapp/test/e2e_tests/specs/Ping/ping.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  test(
+    'Verify I see a warning when I ping in a big group',
+    {tag: ['@TC-1496', '@regression']},
+    async ({createPage, createUser}) => {
+      const usersForBigGroup = [];
+      for (let i = 0; i<4; i++) {
+        const newMember = await createUser();
+        await team.addMember(newMember);
+        usersForBigGroup.push(newMember);
+      }
```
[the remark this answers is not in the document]
markbrockhoff: I agree with Copilot on this one, why did you disagree?

markbrockhoff (changes requested):

markbrockhoff (approved):

PR #20310, opened 2026-02-04
fix: persist audio elements in DOM and forward mic change events [WPB-23195]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23195" title="WPB-23195" target="_blank">BugWPB-23195</a>  [desktop] Not being able to change audio device during call
  </td></table>
  <br />
 

# Pull Request

## Summary

Audio output device switching was unreliable doing a call. The browser would drop the
audio graph because audio elements were detached from the DOM, causing
`setSinkId()` to be ignored. Additionally, microphone change events were not
propagated consistently, leading to unnecessary audio reinitialization.

### Solution

- Keep all audio elements attached to the DOM for their entire lifecycle of a call to
  ensure a stable audio graph and reliable speaker binding.
- Forward microphone change events explicitly to ensure audio input updates
  without resetting audio output state.

---

## Accessibility (required)

---

## Notes for reviewers

This aligns with browser audio graph requirements where audio elements must be
DOM-bound to reliably control output devices.


apps/webapp/src/script/repositories/calling/Call.ts
```diff
@@ -162,6 +162,7 @@ export class Call {
       audioElement.srcObject = audio.stream;
       audioElement.play();
       audio.audioElement = audioElement;
+      document.body.appendChild(audioElement);
```
zskhan: It is good idea to append it to the DOM.
I have a suggestion that instead of appending in the `document.body` we can have an element like we have `audio-elements` for ringtone to keep the DOM clean

EnricoSchw: It must append it to the DOM. If not the browse ignore them in some cases. And then speaker input not changed. I saw this existing 'audio-elements'.  I wouldn't attach them to this element. These audio elements are static. The audio elements in the call are dynamic. It must be possible to simply discard all audio elements of a call. However, the static ones shouldn't be accidentally removed as well.

I would create a new one. But this is React. Let me try.

zskhan (approved):

PR #20362, opened 2026-02-09
chore: Publish libraries on merge workflow [WPB-23275]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23275" title="WPB-23275" target="_blank">TaskWPB-23275</a>  [Web] Fix publish workflow
  </td></table>
  <br />
 

# Pull Request

## Why `preserveLocalDependencyProtocols: false`

`@wireapp/core` depends on `@wireapp/api-client` using the `workspace:^` protocol. Since Nx v21, this protocol is preserved by default during `nx release version`. At publish time, `nx release publish` expects the package manager to dynamically replace `workspace:^` with a real semver in the published tarball — but only pnpm and bun support this. Yarn does not, causing publish to fail:

> _Cannot publish package "@wireapp/core" because it contains a local dependency protocol in its "dependencies", and your package manager is yarn._

Setting `preserveLocalDependencyProtocols: false` tells `nx release version` to resolve `workspace:^` → the actual version (e.g., `27.95.5`) in the source `package.json` during the versioning step, so by the time `nx release publish` runs, the dependency is already a valid semver. This aligns with our fixed version convention and has no impact on local development — Yarn workspaces still symlinks the local package as long as the version matches.

---

- [X] **External inputs are validated & sanitized** on client and/or server where applicable.
- [X] **API responses are validated**; unexpected shapes are handled safely (fallbacks or errors).
- [X] **No unsafe HTML is rendered**; if unavoidable, sanitization is applied **and** documented where it happens.
- [X] **Injection risks (XSS/SQL/command) are prevented** via safe APIs and/or escaping.

## Accessibility (required)

- [X] I have read and this PR **upholds** our [Accessibility Best Practices](https://github.com/wireapp/wire-webapp/blob/dev/docs/accessibility-practices.md).

- [X] I have read and this PR **upholds** our [Coding Standards](https://github.com/wireapp/wire-webapp/blob/dev/docs/coding-standards.md) and [Tech Radar Choices](https://github.com/wireapp/wire-webapp/blob/dev/docs/tech-radar.md).

### DIFF
--- a/.github/workflows/publish-libraries-on-merge.yml
+++ b/.github/workflows/publish-libraries-on-merge.yml
@@ -11,7 +11,7 @@ permissions:
   id-token: write
 
 concurrency:
-  group: publish-libraries-${{ github.event.pull_request.base.ref }}
+  group: publish-libraries
   cancel-in-progress: false
 
 jobs:
--- a/.github/workflows/publish-libraries-on-merge.yml
+++ b/.github/workflows/publish-libraries-on-merge.yml
@@ -16,7 +16,6 @@ concurrency:
 
 jobs:
   publish:
-    name: Publish Libraries to NPM
     if: |
       github.event.pull_request.merged == true &&
       contains(github.event.pull_request.labels.*.name, 'publish-to-npm')
--- a/.github/workflows/publish-libraries-on-merge.yml
+++ b/.github/workflows/publish-libraries-on-merge.yml
@@ -19,7 +19,7 @@ jobs:
     name: Publish Libraries to NPM
     if: |
       github.event.pull_request.merged == true &&
-      startsWith(github.event.pull_request.head.ref, 'chore/library-release-')
+      contains(github.event.pull_request.labels.*.name, 'publish-to-npm')
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/publish-libraries-on-merge.yml
+++ b/.github/workflows/publish-libraries-on-merge.yml
@@ -26,7 +26,9 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
-          ref: dev
+          # Pin to the exact merge commit to avoid publishing code from a newer dev head
+          # if other PRs merge between trigger and job start
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
 
       - name: Setup Node.js
         uses: actions/setup-node@v6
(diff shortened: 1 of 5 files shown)


V-Gira (approved):

V-Gira (approved):

V-Gira (approved):

libraries/api-client/src/APIClient.ts
```diff
@@ -405,7 +405,7 @@ export class APIClient extends EventEmitter {
       const self = await this.api.self.getSelf();
       selfDomain = self.qualified_id?.domain;
     } catch (error) {
-      this.logger.warn('Could not get self user:', (error as BackendError).message);
```
screendriver, line 408: This change is not related to the semantic meaning of this pull request and should be moved into its own pull request.

.github/workflows/publish-libraries-on-merge.yml
```diff
@@ -0,0 +1,57 @@
+name: Publish Libraries on Release PR Merge
```
screendriver:
```suggestion
name: Publish packages to npm
```

Describes the intention more clearly 🙂 I'm also thinking about if it would be better to give the workflow (the YAML file itself). What do you think @thisisamir98 @V-Gira ?

.github/workflows/publish-libraries-on-merge.yml
```diff
@@ -0,0 +1,57 @@
...
+  contents: write
+  id-token: write
+
+concurrency:
+  group: publish-libraries
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: Publish Libraries to NPM
```
screendriver:
```suggestion

```

I think the workflow name (see comment above) already describes the intention pretty good (so it's duplicated at the moment)

libraries/api-client/src/APIClient.ts
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/20362/commits/2b9f5d32f906ac989971d0c91524c4694941bab5

.github/workflows/publish-libraries-on-merge.yml
thisisamir98: done in https://github.com/wireapp/wire-webapp/pull/20362/commits/86044ce9f6f2c3fffc31e49a6150818bf7c06384

screendriver (approved):

PR #20429, opened 2026-02-17
test: add participant profile regression tests [WPB-19962]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19962" title="WPB-19962" target="_blank">TaskWPB-19962</a>  [Web/QA] Write the Participant Profile regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

New Tests: Added E2E regression suite for Participant Profiles.
Fixture Update: Enhanced addMember in the createTeam fixture to support role specification.
Locator Logic: Refactored getLocatorByUser to select users regardless of their role (Admin vs. Member) or list placement.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
@@ -23,9 +23,8 @@ export class ParticipantDetails {
   readonly page: Page;
 
   readonly userPicture: Locator;
-  readonly adminStatus: Locator;
-  readonly guestStatus: Locator;
-  readonly externalStatus: Locator;
+  readonly userName: Locator;
+  readonly userStatus: Locator;
   readonly createGroup: Locator;
   readonly block: Locator;
   readonly closeButton: Locator;
@@ -36,12 +35,11 @@ export class ParticipantDetails {
   constructor(page: Page) {
     this.page = page;
     this.userPicture = this.page.getByTestId('status-profile-picture');
-    this.adminStatus = this.page
+    this.userName = this.page.locator('.panel-participant').getByTestId('status-name');
+    this.userStatus = this.page
       .locator('#group-participant-user')
-      .getByTestId('status-admin')
-      .filter({hasText: 'Group Admin'});
-    this.guestStatus = this.page.locator('#group-participant-user').getByTestId('status-guest');
-    this.externalStatus = this.page.locator('#group-participant-user').getByTestId('status-external');
+      .getByTestId(/^status-(external|guest|admin)$/)
+      .and(this.page.locator('.panel-participant__label'));
     this.createGroup = this.page.locator('#conversation-details').getByRole('button', {name: 'Create group'});
     this.block = this.page.getByRole('button', {name: 'Block'});
     this.closeButton = this.page.getByRole('button', {name: 'Close conversation info'});
@@ -56,10 +54,6 @@ export class ParticipantDetails {
 
   getUserEmailLocator(email: string) {
     return this.page.getByTestId('item-enriched-value').and(this.page.locator(`[data-uie-value="${email}"]`));
-  }
-
-  getUserNameLocator(userName: string) {
-    return this.page.getByTitle(userName).and(this.page.getByTestId('status-username'));
   }
 
   async closeParticipantDetails() {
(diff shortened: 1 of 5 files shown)


apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
```diff
@@ -22,15 +22,51 @@ import {Locator, Page} from '@playwright/test';
...
+    this.unblockButton = this.page.getByRole('button', {name: 'Unblock'});
+    this.removeFromGroup = this.page.getByRole('button', {name: 'Remove from group'});
   }
 
   async blockUser() {
-    this.block.click();
+    await this.block.click();
+  }
+
+  getUserEmailLocator(email: string) {
```
markbrockhoff, line 55: I noticed that this locator is always negated e.g. `.not.toBeVisible()`. Are you sure it works as expected? Since locating something which doesn't exist easily leads to false positives.

apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
```diff
@@ -22,15 +22,51 @@ import {Locator, Page} from '@playwright/test';
...
+    return this.page.getByTestId('item-enriched-value').and(this.page.locator(`[data-uie-value="${email}"]`));
+  }
+
+  getUserNameLocator(userName: string) {
+    return this.page.getByTitle(userName).and(this.page.getByTestId('status-username'));
```
markbrockhoff: TBH I'm not so stoked about having to pass the user name to the locator for the user name ^^'
You could instead just look for the testId "status-username" only within the participant details, this would also resolve the need for using a function. (Sadly I also couldn't find a better way to locate this text in Wire than using a testid :/)

E.g.
```ts
const panel = page.getByRole('complementary').and(page.getByRole('heading', {name: 'Conversation info'}));
const userName = panel.getByTestId('status-name');
```

apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
```diff
@@ -22,15 +22,51 @@ import {Locator, Page} from '@playwright/test';
...
     this.page = page;
-
-    this.block = this.page.getByTestId('do-block-item-text');
+    this.userPicture = this.page.getByTestId('status-profile-picture');
+    this.adminStatus = this.page
+      .locator('#group-participant-user')
+      .getByTestId('status-admin')
+      .filter({hasText: 'Group Admin'});
+    this.guestStatus = this.page.locator('#group-participant-user').getByTestId('status-guest');
+    this.externalStatus = this.page.locator('#group-participant-user').getByTestId('status-external');
```
markbrockhoff: Do you think it would be possible to merge the locators for the user status into one? E.g. maybe using a regex like this:

```ts
const userStatus = this.page
      .locator('#group-participant-user')
      .getByTestId(/^status-admin|status-external|status-guest$/);
```

Having multiple locators for effectively the same element might not be as maintainable. Instead of filtering we could then do assertions like `await expect(userStatus).toContainText('Group Admin')`

apps/webapp/test/e2e_tests/specs/ParticipantProfile/participantProfile.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+      await expect(userBPages.participantDetails().getUserNameLocator(userA.username)).toBeVisible;
+    },
+  );
+
+  test(
+    'I want to see the external icon on the profile of an external participant',
+    {tag: ['@TC-1484', '@regression']},
+    async ({createPage, createUser}) => {
+      const externalUser = await createUser();
+      team.addMember(externalUser, {role: Role.EXTERNAL});
```
markbrockhoff:
```suggestion
      await team.addMember(externalUser, {role: Role.EXTERNAL});
```

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
zhannach: I added a check for this selector in TC-1477

apps/webapp/test/e2e_tests/pageManager/webapp/pages/participantDetails.page.ts
zhannach: Thank you for the tip, unfortunately the group conversation panel doesn't have this heading. I rewrote the function, but I'm still not happy with the selector

markbrockhoff: Of course only the one on one panel has a heading... 😤
Well, was worth a try

markbrockhoff (approved):

zskhan (approved):

PR #20439, opened 2026-02-18
feat(api-client): Update WebSocket connection logic to fallback from /websocket to /await on 404 [WPB-23436]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23436" title="WPB-23436" target="_blank">TaskWPB-23436</a>  [Web] Update WebSocket connection logic to fallback from /websocket to /await on 404
  </td></table>
  <br />
 

# Pull Request

## Summary

- What did I change and why?

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/libraries/api-client/src/tcp/WebSocketClient.ts
+++ b/libraries/api-client/src/tcp/WebSocketClient.ts
@@ -66,6 +66,7 @@ export class WebSocketClient extends EventEmitter {
   private bufferedMessages: string[];
   private abortHandler?: AbortController;
   private versionPrefix = '';
+  private cachedSocketAddress?: 'websocket' | 'await';
 
   public static readonly TOPIC = TOPIC;
 
@@ -268,7 +269,10 @@ export class WebSocketClient extends EventEmitter {
 
     const queryString = queryParams.toString();
 
-    const socketAddressToUse = await this.findSocketAddressToUse(this.baseUrl, queryString);
+    if (!this.cachedSocketAddress) {
+      this.cachedSocketAddress = await this.findSocketAddressToUse(this.baseUrl, queryString);
+    }
+    const socketAddressToUse = this.cachedSocketAddress;
 
     const websocketAddress = this.useLegacySocket
       ? `${this.baseUrl}/${socketAddressToUse}?${queryString}`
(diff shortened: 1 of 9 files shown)


libraries/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -283,15 +269,51 @@ export class WebSocketClient extends EventEmitter {
 
     const queryString = queryParams.toString();
 
+    if (!this.cachedSocketAddress) {
+      this.cachedSocketAddress = await this.findSocketAddressToUse(this.baseUrl, queryString);
+    }
+    const socketAddressToUse = this.cachedSocketAddress;
```
screendriver: This variable is just an alias to another one. Do we really need that?

libraries/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -283,15 +269,51 @@ export class WebSocketClient extends EventEmitter {
 
     const queryString = queryParams.toString();
 
+    if (!this.cachedSocketAddress) {
```
screendriver:
```suggestion
    if (this.cachedSocketAddress !== undefined) {
```

I know this is spread across the whole code base **but** it would be great to make the intention more clear and to prevent something that is called __future drift__:

`if (!this.cachedSocketAddress)` could mean (but you don't know it):

- if `cachedSocketAddress` changes in the future to be string instead of a string literal and uses an empty string this statement will be truthy
- if it will become `null` then `null` will be also treated as missing

libraries/api-client/src/tcp/WebSocketClient.ts
```diff
@@ -283,15 +269,51 @@ export class WebSocketClient extends EventEmitter {
...
+    // We try to connect to the websocket endpoint first, if it fails we fallback to the await endpoint
+    return new Promise<'websocket' | 'await'>(resolve => {
+      const testSocket = new WebSocket(websocketUrl);
+
+      // In case the connection hangs, we fallback to the await endpoint after 5 seconds
+      const timeoutId = setTimeout(() => {
+        if (testSocket.readyState !== WebSocket.OPEN) {
+          handleFailure();
+        }
+      }, 5000);
```
screendriver: Instead of an hard timer wouldn't it be better to listen for error cases? So for example

```ts
testSocket.addEventListener('error', handleFailure);
```

If the WebSocket connection could really "hang", wouldn't it be easier to introduce [timing out promises](https://github.com/sindresorhus/p-timeout) so that you would not have to take care about cleaning up things afterwards manually. It also provides [fallbacks](https://github.com/sindresorhus/p-timeout#fallback) (which would be exactly what I want to solve here: use `/websocket` and fallback to `/await`)

screendriver (approved):

PR #20492, opened 2026-02-23
test: implement regression tests for in conversation search [WPB-19955]

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19955" title="WPB-19955" target="_blank">TaskWPB-19955</a>  [Web/QA] Write the In Conversation Search regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- add regression tests for in conversation search

---

## Accessibility (required)

---

## Notes for reviewers
Code duplication cannot be reduced any further since it's log in of the users.

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
@@ -23,10 +23,12 @@ import {Locator, Page} from '@playwright/test';
 export class CollectionPage {
   readonly page: Page;
   readonly component: Locator;
+  readonly searchResults: Locator;
 
   constructor(page: Page) {
     this.page = page;
     this.component = page.locator('#collection');
+    this.searchResults = this.component.getByTestId('full-search-item');
   }
 
   get searchInput() {
@@ -41,27 +43,14 @@ export class CollectionPage {
     await this.searchInput.fill(search);
   }
 
-  get imagesSection() {
-    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Images'})});
+  getSection(type: 'Images' | 'Audio' | 'Files') {
+    const section = this.component.locator('section').filter({has: this.page.locator('header', {hasText: type})});
+    return Object.assign(section, {
+      showAllButton: section.getByRole('button', {name: 'Show all'}),
+    });
   }
 
-  get audioSection() {
-    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Audio'})});
-  }
-
-  get filesSection() {
-    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Files'})});
-  }
-
-  get overviewImagesButton() {
-    return this.imagesSection.getByRole('button', {name: /Show all/i});
-  }
-
-  get overviewFilesButton() {
-    return this.filesSection.getByRole('button', {name: /Show all/i});
-  }
-
-  get fullSearchBar() {
+  get searchBar() {
     return this.component.getByRole('textbox', {name: 'Search text messages'});
   }
 
@@ -69,21 +58,4 @@ export class CollectionPage {
     return this.component.getByText('No results.', {exact: true});
   }
 
-  /**
-   * Returns a locator for the highlighted (marked) text within the search results.
-   * @param searchTerm The text you expect to be highlighted.
-   */
-  getMarkedSearchResult(searchTerm: string): Locator {
-    return this.component.locator('mark').filter({hasText: searchTerm});
-  }
-
-  /** Locates the text content of every search result item */
-  get resultTexts() {
-    return this.component.locator('[data-uie-name="full-search-item-text"]');
-  }
-
-  /** Gets all result texts as an array of strings */
-  async getAllResultTexts(): Promise<string[]> {
-    return await this.resultTexts.allInnerTexts();
-  }
 }
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
@@ -21,21 +21,69 @@ import {Locator, Page} from '@playwright/test';
 
 /** POM for the "collection". This page is accessible by searching for a message within a conversation. */
 export class CollectionPage {
-  readonly page: Locator;
+  readonly page: Page;
+  readonly component: Locator;
 
   constructor(page: Page) {
-    this.page = page.locator('#collection');
+    this.page = page;
+    this.component = page.locator('#collection');
   }
 
   get searchInput() {
-    return this.page.getByRole('textbox', {name: 'Search text messages'});
+    return this.component.getByRole('textbox', {name: 'Search text messages'});
   }
 
   get searchItems() {
-    return this.page.getByTestId('full-search-item');
+    return this.component.getByTestId('full-search-item');
   }
 
   async searchForMessages(search: string) {
     await this.searchInput.fill(search);
+  }
+
+  get imagesSection() {
+    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Images'})});
+  }
+
+  get audioSection() {
+    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Audio'})});
+  }
+
+  get filesSection() {
+    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Files'})});
+  }
+
+  get overviewImagesButton() {
+    return this.imagesSection.getByRole('button', {name: /Show all/i});
+  }
+
+  get overviewFilesButton() {
+    return this.filesSection.getByRole('button', {name: /Show all/i});
+  }
+
+  get fullSearchBar() {
+    return this.component.getByRole('textbox', {name: 'Search text messages'});
+  }
+
+  get noResultsMessage() {
+    return this.component.getByText('No results.', {exact: true});
+  }
+
+  /**
+   * Returns a locator for the highlighted (marked) text within the search results.
+   * @param searchTerm The text you expect to be highlighted.
+   */
+  getMarkedSearchResult(searchTerm: string): Locator {
+    return this.component.locator('mark').filter({hasText: searchTerm});
+  }
+
+  /** Locates the text content of every search result item */
+  get resultTexts() {
+    return this.component.locator('[data-uie-name="full-search-item-text"]');
+  }
+
+  /** Gets all result texts as an array of strings */
+  async getAllResultTexts(): Promise<string[]> {
+    return await this.resultTexts.allInnerTexts();
   }
 }
--- a/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
@@ -183,7 +183,8 @@ test.describe('In Conversation Search', () => {
     },
   );
 
-  test('Verify I can search my own message', {tag: ['@TC-385', '@regression']}, async ({createPage}) => {
+  // TODO: [WPB-23814] - search behavior is broken
+  test.skip('Verify I can search my own message', {tag: ['@TC-385', '@regression']}, async ({createPage}) => {
     const [userAPages, userBPages] = await Promise.all([
       PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
       PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
@@ -199,11 +200,11 @@ test.describe('In Conversation Search', () => {
     const collection = userAPages.collection();
     await collection.searchBar.fill('User A');
 
-    // TODO: uncomment the next two lines when search behavior is fixed
+    // TODO: uncomment the next two lines when bug [WPB-23814] is fixed
     // await expect(collection.searchResults).toHaveCount(1);
     // await expect(collection.searchResults).toContainText('User A Message');
 
-    // TODO: remove this line when search behavior is fixed
+    // TODO: remove this line when bug [WPB-23814] is fixed
     await expect(collection.searchResults.last()).toContainText('User A Message');
   });
 
@@ -252,7 +253,6 @@ test.describe('In Conversation Search', () => {
 
     await expect(collection.searchResults).toHaveCount(1);
     await expect(collection.searchResults).toContainText('https://www.kaufland.de');
-
   });
 
   test('Verify I can not search for a deleted message', {tag: ['@TC-392', '@regression']}, async ({createPage}) => {
@@ -355,8 +355,7 @@ test.describe('In Conversation Search', () => {
       await userBPages.conversation().sendMessage('Papaya');
 
       await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
-      await expect(userAPages.conversation().getMessage({content: 'Papaya'})).toBeVisible()
-      ;
+      await expect(userAPages.conversation().getMessage({content: 'Papaya'})).toBeVisible();
       await userAPages.conversation().sendMessage(`Message from User A: 1`);
       await userAPages.conversation().sendMessage('Empty\n'.repeat(50));
       await userAPages.conversation().sendMessage(`Message from User A: 2`);
@@ -372,10 +371,9 @@ test.describe('In Conversation Search', () => {
       await expect(collection.searchResults).toHaveCount(1);
       await expect(collection.searchResults).toContainText('Papaya');
 
-      const markedSearchResult = userAPages.collection().component.locator('mark').filter({hasText: 'Papaya'});
-      await markedSearchResult.click();
+      await userAPages.collection().searchResults.click();
       const messageFromUserB = userAPages.conversation().getMessage({sender: userB});
-      await expect(messageFromUserB).toBeVisible();
+      await expect(messageFromUserB).toBeInViewport();
     },
   );
 });
--- a/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
@@ -24,7 +24,7 @@ import {getAudioFilePath, getTextFilePath, shareAssetHelper} from 'test/e2e_test
 import {getImageFilePath} from 'test/e2e_tests/utils/sendImage.util';
 import {createGroup} from '../../utils/userActions';
 
-test.describe('Reactions', () => {
+test.describe('In Conversation Search', () => {
   let userA: User;
   let userB: User;
 
--- a/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
@@ -29,9 +29,8 @@ test.describe('In Conversation Search', () => {
   let userB: User;
 
   test.beforeEach(async ({createTeam}) => {
-    const team = await createTeam('Test Team', {withMembers: 1});
+    const team = await createTeam('Test Team', {users: [userB]});
     userA = team.owner;
-    userB = team.members[0];
   });
 
   // TODO: links are not shown to the collection - this part of the test is blocked by [WPB-22484]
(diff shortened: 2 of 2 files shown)


apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+  });
+
+  // TODO: links are not shown to the collection - this part of the test is blocked by [WPB-22484]
+  test(
+    'Verify main overview shows media from all categories',
+    {tag: ['@TC-352', '@regression']},
+    async ({createPage}) => {
+      const [userAPages, userBPages] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
```
markbrockhoff: Please no longer use `withConnectedUser` on both pages anymore. I posted a message about this in our dev channel last week, please take a look there regarding the "why".

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+      ]);
+
+      // Preconditions: User B sends media from all categories (images, links, audio and files)
+      await userBPages.conversationList().openConversation(userA.fullName);
```
markbrockhoff: Please specify the protocol for 1on1 conversations to avoid flake:

```suggestion
      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls');
```

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      let {page} = userBPages.conversation();
+      await shareAssetHelper(getImageFilePath(), page, page.getByRole('button', {name: 'Add picture'}));
+
+      await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+      ({page} = userAPages.conversation());
```
markbrockhoff: Are you really sure you want to re-assign the page variable user by userB to now be of userA? That sound's like a recipe for potential disaster when trying to debug this later on...

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+      await shareAssetHelper(getImageFilePath(), page, page.getByRole('button', {name: 'Add picture'}));
+
+      await userAPages.conversation().searchButton.click();
+      await userAPages.collection().page.getByRole('presentation').click();
+      await expect(userAPages.collection().page.getByRole('presentation')).toBeVisible();
```
markbrockhoff: This locator looks for an item with the role "presentation" on the whole page, this will fail due to strict mode as there are multiple such items on the page. Please limit the area you're searching and don't use the whole page. E.g.: `userAPages.collection().imagesSection.getByRole('presentation').first()`

You're also just clicking the image and checking that it's visible. I don't think this check is sufficient to verify the image was opened in the full screen detail view.

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+      const [userAPages, userBPages] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      let {page} = userBPages.conversation();
+      await shareAssetHelper(getImageFilePath(), page, page.getByRole('button', {name: 'Add picture'}));
+
+      const messageWithImage = userAPages.conversation().getMessage({sender: userB});
```
markbrockhoff, line 175: Please ensure userA has the conversation with B opened before trying to get the message -> Flake prevention

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+    await userBPages.conversation().sendMessage('User B message');
+
+    await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+    await userAPages.conversation().sendMessage('User A Message');
+
+    await userAPages.conversation().searchButton.click();
+    await userAPages.collection().fullSearchBar.fill('User A');
+    await expect(userAPages.collection().getMarkedSearchResult('User A')).toBeVisible();
```
markbrockhoff: IMHO this util might be a bit overkill, I don't think we'll need it anywhere else than this suite of tests. Maybe it's sufficient to just assert the marked text (there should only be one match anyway):

`await expect(userAPages.collection().component.locator('mark')).toHaveText('User A');`

I think there might be some error with the implementation though, since the marked text doesn't really match the search term:
Image

Maybe this is something to report ;)

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+    await userBPages.conversation().sendMessage('User B message');
+
+    await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+    await userAPages.conversation().sendMessage('https://www.kaufland.de');
+
+    await userAPages.conversation().searchButton.click();
+    await userAPages.collection().fullSearchBar.fill('kaufland');
+    await expect(userAPages.collection().getMarkedSearchResult('kaufland').first()).toBeVisible();
```
markbrockhoff: Maybe it would be better to not look for the "mark" tag on the page but instead check the number of results shown to be 1 and for that result to contain the link. Then you also wouldn't need to use .first().

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+      await userAPages.conversation().searchButton.click();
+      const collection = userAPages.collection();
+      await collection.fullSearchBar.fill('Message');
+
+      await expect(collection.resultTexts).toHaveCount(2);
+
+      const results = await collection.getAllResultTexts();
+      expect(results[0]).toContain('Newer Message from User A');
+      expect(results[1]).toContain('Older Message from User B');
```
markbrockhoff: Adding a util to just call ".allInnerTexts()" and only using it once might be ab bit early to abstract.

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      // User B sends the first (older) message
+      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      await userBPages.conversation().sendMessage('Older Message from User B');
+
+      //  User A sends the second (newer) message
+      await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
```
markbrockhoff, line 294: It might be good to wait for A to receive the message from B before sending the second one. Just to rule out any potential for a race condition where due to network or whatever user A would send his message before receiving the one from user B.

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+        PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(pm => pm.webapp.pages),
+      ]);
+
+      await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      await userBPages.conversation().sendMessage('Papaya');
+
+      await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+      for (let i = 0; i < 20; i++) {
+        await userAPages.conversation().sendMessage(`Message from User A: ${i}`);
+      }
```
markbrockhoff: Tipp: Instead of sending multiple messages you could also just send one very long message containing line breaks ;)


```suggestion
      await pages.conversation().sendMessage('Empty\n'.repeat(20));
```

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+      await expect(userAPages.conversation().getMessage({content: 'Message from User A: 19'})).toBeVisible();
```
markbrockhoff: Instead of asserting the last message is visible I'd suggest to rather check the first message is not visible as that's the state you're trying to reach here.

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
```diff
@@ -21,21 +21,69 @@ import {Locator, Page} from '@playwright/test';
...
+    return this.component.locator('section').filter({has: this.page.locator('header', {hasText: 'Files'})});
+  }
+
+  get overviewImagesButton() {
+    return this.imagesSection.getByRole('button', {name: /Show all/i});
+  }
+
+  get overviewFilesButton() {
+    return this.filesSection.getByRole('button', {name: /Show all/i});
+  }
```
markbrockhoff: What happens if there's a new type of section, e.g. the one for the links? -> You'd have to add two new functions, one to get the section and one for the show all button. This is just code duplication and not very scaleable.

Here's a trick how you can combine all of this into one small package, by leveraging [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign):

```suggestion
  getSection(type: 'Images' | 'Audio' | 'Files') {
    const section = this.component.locator('section').filter({has: this.page.locator('header', {hasText: type})});
    return Object.assign(section, {
      showAllButton: section.getByRole('button', {name: 'Show all'}),
    });
  }
```

Instead of returning only the locator for the wanted section, we add a new property `showAllButton` to it and then return it. This way the returned locator can still be used like any other e.g. inside an expect or to chain off other locators from it. But we can also access the showAllButton using `getSection('Images').showAllButton` on it. A neat trick for when you want to add util to something returned by an other util but it's not worth creating a whole new POM for it ;)

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
```diff
@@ -21,21 +21,69 @@ import {Locator, Page} from '@playwright/test';
...
+
+  get fullSearchBar() {
```
markbrockhoff: ❓ Why "fullSearchBar" and not just "searchBar"?

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
```diff
@@ -21,21 +21,69 @@ import {Locator, Page} from '@playwright/test';
...
+
+  get noResultsMessage() {
+    return this.component.getByText('No results.', {exact: true});
+  }
+
+  /**
+   * Returns a locator for the highlighted (marked) text within the search results.
+   * @param searchTerm The text you expect to be highlighted.
+   */
+  getMarkedSearchResult(searchTerm: string): Locator {
```
markbrockhoff: I think it makes a util useless if you need to pass in what you're searching for. Maybe a locator for all searchResults in general could be more useful. Then we could assert on the number of results, or filter them or find the marked texts inside. :)

> This would also replace the `resultTexts()` util below as we could just assert on the text in the results in general instead of  having a util for it.

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: removed it

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: good point, added the protocol 👍

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: I absolutely agree, don't know why I did that initially- changed it

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: good point, changed it

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: changed

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: assertion is now in the test suite file, the search behavior is reported

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: agree, changed it

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: check is inserted

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: way more efficient now, thanks :D

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: changed it

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
JacquelineLehner: that's way more readable and changed 👍

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
JacquelineLehner: changed

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
JacquelineLehner: renamed it :D

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: changed it

apps/webapp/test/e2e_tests/pageManager/webapp/pages/collection.page.ts
JacquelineLehner: renamed it

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -193,8 +196,15 @@ test.describe('In Conversation Search', () => {
     await userAPages.conversation().sendMessage('User A Message');
 
     await userAPages.conversation().searchButton.click();
-    await userAPages.collection().fullSearchBar.fill('User A');
-    await expect(userAPages.collection().getMarkedSearchResult('User A')).toBeVisible();
+    const collection = userAPages.collection();
+    await collection.searchBar.fill('User A');
+
+    // TODO: uncomment the next two lines when search behavior is fixed
```
markbrockhoff: Could you please add the ticket number for this issue to the comment? This way it's easier to uncomment as part of the issue being resolved

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+
+      const collection = userAPages.collection();
+      await collection.searchBar.fill('Papaya');
+      await expect(collection.searchResults).toHaveCount(1);
+      await expect(collection.searchResults).toContainText('Papaya');
+
+      const markedSearchResult = userAPages.collection().component.locator('mark').filter({hasText: 'Papaya'});
+      await markedSearchResult.click();
+      const messageFromUserB = userAPages.conversation().getMessage({sender: userB});
+      await expect(messageFromUserB).toBeVisible();
```
markbrockhoff: This will always be true, for testing scrolling you need to use `.toBeInViewport()` like you did above ;)

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+      await userAPages.conversation().searchButton.click();
+      await userAPages.collection().searchBar.fill('Papaya');
+
+      const collection = userAPages.collection();
+      await collection.searchBar.fill('Papaya');
+      await expect(collection.searchResults).toHaveCount(1);
+      await expect(collection.searchResults).toContainText('Papaya');
+
+      const markedSearchResult = userAPages.collection().component.locator('mark').filter({hasText: 'Papaya'});
+      await markedSearchResult.click();
```
markbrockhoff: Do you really need to check the marked text for this test? If it's only about scrolling to the message when clicked couldn't you just click the searchResult? E.g.:

```suggestion
      await userAPages.collection().searchResults.click();
```

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: No, that would be too easy 😂 I changed it :)

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: true- changed it ✅

apps/webapp/test/e2e_tests/specs/InConversationSearch/inConversationSearch.spec.ts
JacquelineLehner: added ticket number

markbrockhoff (approved):

arjita-mitra (approved):

PR #20498, opened 2026-02-24
test: add mention regression tests [WPB-19960]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19960" title="WPB-19960" target="_blank">TaskWPB-19960</a>  [Web/QA] Write the Mention regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- Implement regression tests for mentioning other users in messages
- Fix aria label of decline button within call cell
- Group conversation list item locators e.g. for unread or mention status
- Fix a11y within UserList to correctly label the group admin and member lists
- Create setup for adding guest users to a group conversation

---

## Accessibility (required)

---

## Notes for reviewers

- This PR is on the bigger side, but I made sure all commits inside it are atomic. So instead of looking at everything at once, you may step through the individual commits one after an other during review to reduce the cognitive overhead ;)

### DIFF
--- a/apps/webapp/playwright.config.ts
+++ b/apps/webapp/playwright.config.ts
@@ -64,7 +64,7 @@ module.exports = defineConfig({
       name: 'Chromium',
       use: {
         ...devices['Desktop Chrome'],
-        permissions: ['notifications'],
+        permissions: ['notifications', 'clipboard-read', 'clipboard-write'],
         launchOptions: {
           args: [
             '--use-fake-device-for-media-stream', // Provide fake devices for audio & video device input
--- a/apps/webapp/src/script/components/calling/CallingCell/CallingControls/CallingControls.tsx
+++ b/apps/webapp/src/script/components/calling/CallingCell/CallingControls/CallingControls.tsx
@@ -39,7 +39,6 @@ interface CallingControlsProps {
   answerCall: () => void;
   call: Call;
   callActions: CallActions;
-  call1To1StartedAlert: string;
   isDetachedWindow: boolean;
   isFullUi?: boolean;
   isMuted?: boolean;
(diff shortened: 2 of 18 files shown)


apps/webapp/src/script/components/calling/CallingCell/CallingControls/CallingControls.tsx
```diff
@@ -176,8 +176,8 @@ export const CallingControls = ({
               className="call-ui__button call-ui__button--red call-ui__button--large"
               onClick={() => (isIncoming ? callActions.reject(call) : callActions.leave(call))}
               onBlur={() => clearShowAlert()}
-              title={!isGroup && showAlert ? call1To1StartedAlert : t('videoCallOverlayHangUp')}
```
markbrockhoff, line 179: The `call1To1StartedAlert` message is already placed on the calling cell itself. It should not be used as title / label of a button since it just says e.g. "You are calling {conversationName}." which adds no accessible meaning to a button and makes it even less accessible. So I removed it in favor for the descriptive label describing the button to hang up which is what it's doing.

apps/webapp/playwright.config.ts
```diff
@@ -64,7 +64,7 @@ module.exports = defineConfig({
       name: 'Chromium',
       use: {
         ...devices['Desktop Chrome'],
-        permissions: ['notifications'],
+        permissions: ['notifications', 'clipboard-read', 'clipboard-write'],
```
markbrockhoff, line 67: Turns out it's not possible to create a guest link if the browser doesn't have clipboard permission...
After setting a password a modal is forced on you to copy the password. If the user doesn't copy it or the browser can't write to the clipboard it will just fail and do nothing.

arjita-mitra (dismissed):

arjita-mitra (approved):

e-maad (approved):

PR #20507, opened 2026-02-25
test: write the Status regression e2e tests[WPB-19971]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19971" title="WPB-19971" target="_blank">TaskWPB-19971</a>  [Web/QA] Write the Status regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary
1. Add end-to-end (e2e) status regression tests.
2. Update getConversationWithUser with a conversationName filter to support group conversation lookups.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/test/e2e_tests/pageManager/webapp/modals/statusChange.modal.ts
```diff
@@ -17,18 +17,12 @@
...
-export function runClientVersionCheck(options: RunClientVersionCheckOptions): void {
-  const {ky} = options;
-
-  const requestClientVersionCheck = async () => {
-    await ky.get('/client-version-check');
-  };
+import {BaseModal} from './base.modal';
 
-  void requestClientVersionCheck();
+export class StatusChangeModal extends BaseModal {
```
markbrockhoff: This modal is identical to the `OptionModal` (option.modal.ts)
There are only a limited number of modal types within wire, so unless a modal is very specialized through child components there's no need to create a custom modal type for it. Please just use the OptionModal in this case instead of creating a new one :)

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
```diff
@@ -111,6 +111,11 @@ export class ConversationDetailsPage {
     await userLocator.click();
   }
 
+  async statusParticipant(fullName: string) {
```
markbrockhoff: 🎨 

```suggestion
  async getParticipantStatus(fullName: string) {
```

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+      await updateUserStatus(userBPageManager.webapp, userB.fullName, UserStatus.Away);
+
+      const {getNotifications: getUserBNotifications} = await interceptNotifications(pages.conversation().page);
+
+      for (const conversation of ['1on1', 'group']) {
+        await test.step(`User A opens the ${conversation} conversation`, async () => {
+          if (conversation === '1on1') {
+            // User B sends a message to User A
+            await prepareConversationForReply(userBPageManager.webapp, userA.fullName, conversation);
+            conversationId = await api.conversation.getConversationWithUser(userA.token, userB.id!, {
```
markbrockhoff: Did I overlook something or is this variable assigned within the test but never used? Maybe it shouldn't be global but local to the test to avoid something like this?

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+        for (const systemTestCase of systemTestCases) {
+          await test.step(`Action: ${systemTestCase.name}`, async () => {
+            await systemTestCase.sendAction({pageA: userAPageManager});
+            await expect.poll(() => getUserBNotifications()).toHaveLength(0);
+          });
+        }
+      });
+
+      await test.step(`User B should receive mentions, replies and calls notifications in 1to1`, async () => {
+        const specialTasteCases = commonTestCases.filter(
```
markbrockhoff: Oh, yes those test cases look very "tasty" 😋


```suggestion
        const specialTestCases = commonTestCases.filter(
```

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
markbrockhoff: Oh, I see they're implicitly used within the commonTestCases...
Please just pass the conversationId to the `sendAction()` function of them then instead of having it as implicit global dependency

apps/webapp/test/e2e_tests/backend/ConversationRepository/conversationRepository.e2e.ts
```diff
@@ -71,7 +71,7 @@ export class ConversationRepositoryE2E extends BackendClientE2E {
   async getConversationWithUser(
     token: string,
     conversationPartnerId: string,
-    options?: {protocol?: 'proteus' | 'mls'},
+    options?: {protocol?: 'proteus' | 'mls'; conversationName?: string},
```
markbrockhoff: Hmm, I don't think mixing group and 1on1 conversations within this util is a good idea. Ideally I'd prefer to just get the conversationId by the name of the conversation e.g. "Max Mustermann" or "Group Conversation" but since the name property you added only seems to be defined for group conversations I think it's probably best to split the functions.
E.g. leave this function as is for only finding 1on1 conversations as the name implies and create a second util e.g. `getGroupConversation(token: string, conversationName: string)`. That should also simplify the logic for filtering a lot.

But what do you think? :)

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+          await test.step(`Action: ${systemTestCase.name}`, async () => {
+            await systemTestCase.sendAction({pageA: userAPageManager});
+            await expect.poll(() => getUserBNotifications()).toHaveLength(0);
+          });
+        }
+      });
+
+      await test.step(`User B should not receive notification from call in 1to1`, async () => {
+        await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+        await userAPages.conversation().startCall();
```
markbrockhoff: Could you please make sure userB received the call before asserting he didn't get the notification? E.g. by checking the call cell is visible beforehand. Otherwise this probably won't ever fail as the check for length to be 0 is done right after the call button was clicked. The network would take longer for the call to arrive than the negative assertion.

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+        });
+
+        await test.step(`User B should not receive any notifications for certain messages (text, file, audio, ping, images, link preview)`, async () => {
+          for (const testCase of commonTestCases) {
+            if (testCase.name === 'Mention' || testCase.name === 'Reply') {
+              continue; // Skip mention and reply test cases as they are handled separately
+            }
+            await test.step(`Action: ${testCase.name} message in ${conversation}`, async () => {
+              await testCase.sendAction({pageA: userAPages.conversation().page, api});
+              await expect.poll(() => getUserBNotifications()).toHaveLength(0);
```
markbrockhoff: Just a suggestion to make this test a bit more reliable: Since you're only interested in the fact that no notifications were received you could just send all messages after each other and do one general assertion at the end. Before doing this assertion you could verify userB received the messages e.g. using the conversationList: `await expect(userBPages.conversationList().getConversationLocator(userA.fullName)).toContainText('1 message');`

Since it's many messages the timing should be ok but this way we could be sure the user received the messages but didn't get a system notification

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+
+      await test.step('Prerequisite: User B sets status to Busy and mutes group conversation ', async () => {
+        await updateUserStatus(userBPageManager.webapp, userB.fullName, UserStatus.Busy);
+
+        await pages.conversationList().openConversation(groupName);
+        await pages.conversation().clickConversationInfoButton();
+        await pages.conversationDetails().setNotifications('Mentions and replies');
+      });
+
+      const {getNotifications: getUserBNotifications} = await interceptNotifications(pages.conversation().page);
```
markbrockhoff: 🎨 Please don't access the page object via the POM. If you need access to the page itself don't wrap it with the PageManager immediately after creating it.

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+        });
+
+        await prepareConversationForReply(userBPageManager.webapp, userA.fullName, '1on1');
+
+        await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+
+        for (const testCase of specialTasteCases) {
+          await test.step(`Action: ${testCase.name} message`, async () => {
+            await testCase.sendAction({pageA: userAPages.conversation().page, api});
+            await expect.poll(() => getUserBNotifications()).toHaveLength(1);
```
markbrockhoff: Are you sure this should only assert for 1 notification? shouldn't there be 3 after the test cases ran?

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+
+      await test.step(`User B should receive system notifications only for group creation and renaming`, async () => {
+        await userAPages.conversationList().openConversation(groupName);
+        for (const systemTestCase of systemTestCases) {
+          await test.step(`Action: ${systemTestCase.name}`, async () => {
+            await systemTestCase.sendAction({pageA: userAPageManager});
+          });
+        }
+
+        await expect.poll(() => getUserBNotifications()).toHaveLength(notificationCount + 2);
```
markbrockhoff: The "+ 2" here is a bit to random for my taste, could you at least use systemTestCases.length? Or send the messages all together and just check for a static number

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+      });
+    },
+  );
+
+  test('I want to set my status on profile page', {tag: ['@TC-1766', '@regression']}, async ({createPage}) => {
+    const userAPageManager = await PageManager.from(createPage(withLogin(userA)));
+    const userBPages = await PageManager.from(createPage(withLogin(userB), withConnectedUser(userA))).then(
+      pm => pm.webapp.pages,
+    );
+    const {pages, components, modals} = userAPageManager.webapp;
```
markbrockhoff:
```suggestion
    const [userAPage, userBPage] = await Promise.all([
      createPage(withLogin(userA)),
      createPage(withLogin(userB), withConnectedUser(userA)),
    ]);

    const {pages: userAPages, components, modals} = PageManager.from(userAPage).webapp;
    const {pages: userBPages} = PageManager.from(userBPage).webapp;
```

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
markbrockhoff: 🎨 Tipp: You could just remove this line and make the function sync. That would remove some redundant awaits in your tests ;)

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+}
+
+async function updateUserStatus(pageManager: PageManager['webapp'], userName: string, status: UserStatus) {
+  const {components, modals} = pageManager;
+  await components.conversationSidebar().openStatusMenu(userName);
+  await components.conversationSidebar().setStatus(status);
+  await modals.statusChangeModal().clickAction();
+}
+
+async function prepareConversationForReply(
```
markbrockhoff: TBH I feel like this function might be abstracted a bit too early. Or it would need to abstract more than just the "preparation". But maybe that's just my feeling, what do you think?

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+  let groupName: string;
+  let conversationId: string | undefined;
+
+  test.beforeEach(async ({createTeam, createUser}) => {
+    userB = await createUser();
+    userC = await createUser();
+
+    team = await createTeam('Test Team', {users: [userB, userC]});
+    userA = team.owner;
+    groupName = 'Test group';
```
markbrockhoff: Why not just use `const groupName = 'Test group'` when declaring the variable?

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,540 @@
...
+        const message = userAPages.conversation().getMessage({sender: userA});
+        await expect(message).toBeVisible();
+      },
+    },
+    {
+      name: 'Image',
+      sendAction: async ({pageA}) => {
+        const userAPages = PageManager.from(pageA).webapp.pages;
+        await shareAssetHelper(getImageFilePath(), pageA, pageA.getByRole('button', {name: 'Add picture'}));
+        expect(await userAPages.conversation().isImageFromUserVisible(userA)).toBeTruthy();
```
markbrockhoff: Please try to no longer depend on the `is*` functions which don't return a locator for assertions. In general I think you don't need to even assert the message was sent here, since it's not relevant for the tests.

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -224,79 +215,135 @@ test.describe('Status', () => {
...
         await expect.poll(() => getUserBNotifications()).toHaveLength(0);
       });
-    },
-  );
+
+      // Handle the "Special" Notifications (Calls/Mentions/Replies)
+      if (config.expectSpecialNotifications) {
+        await test.step('User B should receive mentions, replies, and calls (Busy status)', async () => {
+          const specialTestCases = [
+            ...commonTestCases.filter(tc => tc.name === 'Mention' || tc.name === 'Reply'),
```
markbrockhoff, line 302: When working with strings like this it's easy to mess up due to typos or if something is renamed later on. To prevent this you can initialize `commonTestCases` using a combination of const and satisfies, e.g.:

```ts
const commonTestCases = [/* Test cases */] as const satisfies {
    name: string;
    sendAction: (params: {pageA: Page; api: ApiManagerE2E; conversation?: string}) => Promise<void>;
  }[];
```

This will make properties like `name` here type safe

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -224,79 +215,135 @@ test.describe('Status', () => {
...
+                const userAPages = PageManager.from(pageA).webapp.pages;
+                await userAPages.conversation().startCall();
+              },
+            },
+          ];
+
+          await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+          for (const testCase of specialTestCases) {
+            await test.step(`Action: ${testCase.name} message`, async () => {
+              await testCase.sendAction({pageA: userAPages.conversation().page, api});
```
markbrockhoff:
```suggestion
              await testCase.sendAction({pageA: userAPage, api});
```

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -311,176 +358,97 @@ test.describe('Status', () => {
...
+
+      for (const {type, target, options} of scenarios) {
+        await test.step(`User A opens the ${type} conversation`, async () => {
+          await userAPages.conversationList().openConversation(target, options);
         });
 
         await test.step(`User B should receive all conversation notifications`, async () => {
           for (const testCase of commonTestCases) {
             await test.step(`Action: ${testCase.name}`, async () => {
               await testCase.sendAction({pageA: userAPages.conversation().page, api});
```
markbrockhoff:
```suggestion
              await testCase.sendAction({pageA: userAPage, api});
```

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+        }
+
+        await expect.poll(() => getUserBNotifications()).toHaveLength(notificationCount);
+      });
+    },
+  );
+
+  test('I want to set my status on profile page', {tag: ['@TC-1766', '@regression']}, async ({createPage}) => {
+    const [userAPageManager, userBPageManager] = await Promise.all([
+      PageManager.from(createPage(withLogin(userB))),
```
markbrockhoff:
```suggestion
      PageManager.from(createPage(withLogin(userA))),
```

apps/webapp/test/e2e_tests/specs/Status/status.spec.ts
```diff
@@ -0,0 +1,508 @@
...
+    name: string;
+    sendAction: (params: {pageA: Page; api: ApiManagerE2E; conversation?: string}) => Promise<void>;
+  }[] = [
+    {
+      name: 'Text',
+      sendAction: async ({pageA}) => {
+        const userAPages = PageManager.from(pageA).webapp.pages;
+        await userAPages.conversation().sendMessage('Test message');
+        const message = userAPages.conversation().getMessage({sender: userA});
+        await expect(message).toBeVisible();
```
markbrockhoff: I just tried removing all the assertions inside the sendAction functions in here as well as for the systemMessages and at least locally it ran just fine for me. Are you sure the issue you mentioned is related to this or could it maybe be the version of the testservice you're using locally?

Image

markbrockhoff (changes requested):

markbrockhoff (approved):

zskhan (approved):

PR #20508, opened 2026-02-25
fix(webapp): Implement recovery mechanism for epoch number mismatch between CC and local state [WPB-23452]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23452" title="WPB-23452" target="_blank">TaskWPB-23452</a>  [Web] Implement recovery mechanism for epoch number mismatch between CC and local web state
  </td></table>
  <br />
 

## Problem:

if we read the epoch number from core crypto in case welcome is processed sooner than reset conversation will exists on core crypto but epoch number will be 0, on this scenario user will have a local unestablished MLS conversation

## Solution:

In order to fix this we should not rely on the epoch number that we have on our local state and always read the epoch number from core crypto.

### DIFF
--- a/apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
+++ b/apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
@@ -2210,39 +2210,52 @@ export class ConversationRepository {
   public ensureConversationExists = async ({
     conversationId,
     groupId,
-    epoch,
     core = this.core,
     retry = true,
   }: {
     conversationId: QualifiedId;
     groupId: string;
-    epoch: number;
     core?: Account;
     retry?: boolean;
   }): Promise<void> => {
-    this.logger.info('Ensuring conversation exists', {conversationId, groupId, epoch});
+    const coreCryptoEpochNumber = await this.core.service?.mls?.getEpoch(groupId);
+
+    this.logger.info('Ensuring conversation exists', {conversationId, groupId, epoch: coreCryptoEpochNumber});
     if (await this.conversationService.mlsGroupExistsLocally(groupId)) {
-      this.logger.info('Conversation already exists locally', {conversationId, groupId, epoch});
-      if (epoch === 0) {
+      this.logger.info('Conversation already exists locally', {conversationId, groupId, epoch: coreCryptoEpochNumber});
+      if (coreCryptoEpochNumber === 0) {
         if (!retry) {
-          this.logger.error('Epoch is 0, but retry is false, not retrying again', {conversationId, groupId, epoch});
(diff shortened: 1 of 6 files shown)


screendriver (commented):
Shouldn't some tests fail now?

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2210,39 +2210,52 @@ export class ConversationRepository {
...
+
+    this.logger.info('Ensuring conversation exists', {conversationId, groupId, epoch: coreCryptoEpochNumber});
     if (await this.conversationService.mlsGroupExistsLocally(groupId)) {
-      this.logger.info('Conversation already exists locally', {conversationId, groupId, epoch});
-      if (epoch === 0) {
+      this.logger.info('Conversation already exists locally', {conversationId, groupId, epoch: coreCryptoEpochNumber});
+      if (coreCryptoEpochNumber === 0) {
         if (!retry) {
-          this.logger.error('Epoch is 0, but retry is false, not retrying again', {conversationId, groupId, epoch});
+          this.logger.error('Epoch is 0, but retry is false, not retrying again', {
```
screendriver, line 2228: This is a lie: `retry` could also be `undefined` and not just `false`

thisisamir98: it has a default value: retry = true,

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2210,39 +2210,52 @@ export class ConversationRepository {
...
+      });
+      await this.establishMlsGroupConversation({conversationId, groupId, epoch: coreCryptoEpochNumber, core});
       return;
     }
 
     // join by external commit
-    this.logger.info('Joining conversation by external commit', {conversationId, epoch});
-    if (epoch && epoch > 0) {
+    this.logger.info('Joining conversation by external commit', {conversationId, epoch: coreCryptoEpochNumber});
+    if (coreCryptoEpochNumber && coreCryptoEpochNumber > 0) {
```
screendriver:
```suggestion
    if (coreCryptoEpochNumber !== undefined && coreCryptoEpochNumber > 0) {
```

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
screendriver: You are absolutely right 👍

typfel (approved):

screendriver (approved):

screendriver (approved):

PR #20519, opened 2026-02-27
test: re-enable crit flow group call [WPB-23747]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23747" title="WPB-23747" target="_blank">TaskWPB-23747</a>  [Web/QA] Fix and re-enable TC-8632 Group call
  </td></table>
  <br />
 

# Pull Request

## Summary

This PR re-enables the previously skipped critical flow test for group calling. It was skipped due to a limitation with enterprise features not being unlockable which was resolved. The PR also:
- Makes the emails of generated users follow their first and last name instead of just last name (for easier debugging)
- Refactors the critical flow test to follow best practices
- Removes a lot of no longer needed utils

---

## Accessibility (required)

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/CriticalFlow/groupCalls-TC-8632.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/CriticalFlow/groupCalls-TC-8632.spec.ts
@@ -25,88 +25,73 @@ import {PageManager} from 'test/e2e_tests/pageManager';
 import {test, expect, withLogin} from '../../test.fixtures';
 import {createGroup} from 'test/e2e_tests/utils/userActions';
 
+let owner: User;
+let member: User;
+const conversationName = 'Calling';
+
+test.beforeEach(async ({api, createUser, createTeam}) => {
+  member = await createUser();
+  const team = await createTeam('Calling', {users: [member]});
+  owner = team.owner;
+
+  // The team will be reset right after initialization, so we need to wait a short time for it to finish before changing feature configs since they would otherwise be overwritten
+  await new Promise(resolve => setTimeout(resolve, 5000));
+  await api.enableConferenceCallingFeature(owner.teamId!);
+  await api.waitForFeatureToBeEnabled(FEATURE_KEY.CONFERENCE_CALLING, owner.teamId!, owner.token);
+});
+
 test(
   'Planning group call with sending various messages during call',
   {tag: ['@TC-8632', '@crit-flow-web']},
-  async ({createUser, createTeam, createPage, api}) => {
-    test.setTimeout(150_000);
-
-    let owner: User;
-    let member: User;
-    let ownerPageManager: PageManager;
-    let memberPageManager: PageManager;
-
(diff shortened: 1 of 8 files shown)


apps/webapp/test/e2e_tests/specs/HistoryBackup/historyBackup.spec.ts
```diff
@@ -209,7 +209,9 @@ test.describe('History Backup', () => {
 
         // User A sees system message that User B had renamed the conversation
         await userAPages.conversationList().openConversation(renamedConversationName);
-        const renamedSystemMessage = userAPages.conversation().systemMessages.filter({hasText: `${userB.fullName} renamed the conversation`});
+        const renamedSystemMessage = userAPages
+          .conversation()
+          .systemMessages.filter({hasText: `${userB.fullName} renamed the conversation`});
         await expect(renamedSystemMessage).toContainText(`${userB.fullName} renamed the conversation`);
```
zhannach: I see it was there before, but it seems a bit redundant to filter on specific text and check for ContainText. Maybe we can use the last message here userAPages.conversation().systemMessages.last()

apps/webapp/test/e2e_tests/specs/CriticalFlow/groupCalls-TC-8632.spec.ts
```diff
@@ -25,99 +25,85 @@ import {PageManager} from 'test/e2e_tests/pageManager';
...
+const conversationName = 'Calling';
+
+test.beforeEach(async ({api, createUser, createTeam}) => {
+  member = await createUser();
+  const team = await createTeam('Calling', {users: [member]});
+  owner = team.owner;
+
+  // The team will be reset right after initialization, so we need to wait a short time for it to finish
+  // before changing feature configs since they would otherwise be overwritten (See WPB-23698)
+  await new Promise(resolve => setTimeout(resolve, 5000));
```
thisisamir98: what if we abstract this to createTeam and add a waitForReset parameter?

apps/webapp/test/e2e_tests/specs/HistoryBackup/historyBackup.spec.ts
markbrockhoff: Agreed, but I'd still prefer a content based locator over a positional one. I replaced the redundant assertion to only check the system message containing this text is visible.

apps/webapp/test/e2e_tests/specs/CriticalFlow/groupCalls-TC-8632.spec.ts
markbrockhoff: Yeah you're probably right, we will need this in more places so it makes sense to include it in the fixture. Although I disagree with the proposed parameter since I don't think this should be configurable. Once the linked story is resolved we should have a deterministic way for checking when the team is ready instead of just waiting "some time".

thisisamir98 (approved):

PR #20521, opened 2026-02-27
refactor: establish MLS groups with one single transaction [WPB-23744]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23744" title="WPB-23744" target="_blank">TaskWPB-23744</a>  [Web] Utilize Core Crypto Transactions API for Full MLS Establish/Join Flow
  </td></table>
  <br />
 

## Problem statement

We identified race conditions during MLS establish/reset flows.
If multiple clients try to establish or reset the same group concurrently, only one “wins” the race. Others may end up in a broken epoch 0 state.

Currently, the flow is split across multiple transactions:

1. Register conversation on Core Crypto
2. Download key packages
3. Establish MLS group

This allows partial state to persist if a failure or race occurs.

## Solution

Refactor the Web MLS establish flow to run fully inside a single Core Crypto transaction.

- Ensure atomic establish/join
- Automatically rollback if:
  - Client loses the race
  - A Welcome arrives mid-establish
  - Any step fails
- Prevent broken or partial MLS state


libraries/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -609,41 +634,43 @@ export class MLSService extends TypedEventEmitter<Events> {
...
-        }
-        return user;
-      }),
-    );
+    return this.coreCryptoClient.transaction(async transactionContext => {
+      await this.registerEmptyConversation(groupId, transactionContext, options?.parentGroupId);
+
+      const creator = options?.creator;
+
+      const {keyPackages, failures: keysClaimingFailures} = await this.getKeyPackagesPayload(
```
screendriver: Side effects in transactions. Is this valid? Just imagine `await` takes longer to respond for any reason. Then the whole transaction would be blocked, right?

libraries/core/src/messagingProtocols/mls/MLSService/MLSService.ts
```diff
@@ -328,16 +329,36 @@ export class MLSService extends TypedEventEmitter<Events> {
...
     const groupIdBytes = Decoder.fromBase64(groupId).asBytes;
 
     if (keyPackages.length < 1) {
       throw new Error('Empty list of keys provided to addUsersToExistingConversation');
     }
 
-    const crlNewDistributionPoints = await this.coreCryptoClient.transaction(cx =>
-      cx.addClientsToConversation(new ConversationId(groupIdBytes), keyPackages),
-    );
+    let crlNewDistributionPoints: NewCrlDistributionPoints;
```
screendriver, line 343: What is `crl`?

thisisamir98: image
@screendriver it's `certificate revocation list`, coming from core crypto apis

screendriver (dismissed):

screendriver (approved):

PR #20565, opened 2026-03-03
fix(core): Do not reset & re-establish conversations if they do not belong to self backend [WPB-22511]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22511" title="WPB-22511" target="_blank">BugWPB-22511</a>  [Web] Triggering MLS resets for a federated conversation will establish a broken group
  </td></table>
  <br />
 

## Summary

User should be able to reset & re-establish on if domain of the conversation is the same of their own user domain

### DIFF
--- a/libraries/core/src/conversation/ConversationService/ConversationService.ts
+++ b/libraries/core/src/conversation/ConversationService/ConversationService.ts
@@ -394,6 +394,12 @@ export class ConversationService extends TypedEventEmitter<Events> {
     selfClientId: string;
     conversationQualifiedId: QualifiedId;
   }): Promise<BaseCreateConversationResponse> {
+    if (selfUserId.domain !== conversationQualifiedId.domain) {
+      const errorMessage = `Self user domain (${selfUserId.domain}) does not match conversation domain (${conversationQualifiedId.domain}), cannot establish MLS group conversation`;
+      this.logger.error(errorMessage, {conversationQualifiedId, selfUserId});
+      throw new Error(errorMessage);
+    }
+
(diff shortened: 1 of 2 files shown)


libraries/core/src/conversation/ConversationService/ConversationService.ts
```diff
@@ -394,6 +394,12 @@ export class ConversationService extends TypedEventEmitter<Events> {
     selfClientId: string;
     conversationQualifiedId: QualifiedId;
   }): Promise<BaseCreateConversationResponse> {
+    if (selfUserId.domain !== conversationQualifiedId.domain) {
```
arjita-mitra: I see the same core validation logic is being used 3 places in this file, although the error message format slightly varies.

I think its worth making it a private method with the following signature

```
private validateDomainMatch(
  selfUserDomain: string,
  conversationDomain: string,
  operationContext: string,
  logContext?: Record<string, unknown>
)
```

then it could be called

```
this.validateDomainMatch(
  selfUserId.domain,
  conversationQualifiedId.domain,
  'cannot establish MLS group conversation',
  {conversationQualifiedId, selfUserId}
);
```

arjita-mitra: WDYT?

thisisamir98: @arjita-mitra thanks, see https://github.com/wireapp/wire-webapp/pull/20565/commits/ce61b988914a226a55d7db7cd3fab8e74f0e189f

arjita-mitra (approved):

PR #20567, opened 2026-03-04
fix: Conversation images previews do not seem to use actual server-provided previews, but original image(WPB-22331)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22331" title="WPB-22331" target="_blank">BugWPB-22331</a>  [Web] Conversation images previews do not seem to use actual server-provided previews, but original image
  </td></table>
  <br />
 

# Pull Request

## Summary

- Small image cards (multiple images in one message): continue using src={filePreviewUrl} correct for grid display, no change.

- Large image cards (single image in a message): now use the original fileUrl, falling back to filePreviewUrl only if unavailable.

- Fullscreen viewer: now uses the original fileUrl instead of the thumbnail.

HEIC handling

HEIC files uploaded from iOS cannot be decoded natively by Chrome, Firefox, or Edge. For these files, isPreviewableImage({extension}) returns false, so both the large card and fullscreen fall back to filePreviewUrl the server-generated JPEG conversion which is safe across all browsers. The original HEIC file is still accessible via the download button in the fullscreen header.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
```diff
@@ -148,7 +149,9 @@ const ModalContent = ({
   }
 
   if (type === 'image') {
-    return <ImageFileView src={filePreviewUrl} senderName={senderName} timestamp={timestamp} />;
+    const isBrowserPreviewable = isPreviewableImage({extension: fileExtension});
+    const imageSrc = isBrowserPreviewable ? fileUrl || filePreviewUrl : filePreviewUrl;
```
screendriver:
```suggestion
    const imageSrc = isBrowserPreviewable ? fileUrl ?? filePreviewUrl : filePreviewUrl;
```

But would you mind extracting that logic into a meaningful and self describing function to make the intention clear what we try to solve here? For example `resolvePreviewImageSource()` or `getBestPreviewSource()`

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/ImageAssetCard/ImageAssetLarge/ImageAssetLarge.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {render} from '@testing-library/react';
+
+import {ImageAssetLarge} from './ImageAssetLarge';
+
+jest.mock('Components/FileFullscreenModal/FileFullscreenModal', () => ({
```
screendriver: Isn't it possible to **not** use `jest.mock()` at all? Module mocking is always the wrong approach.

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
arjita-mitra: Sure actually both of your comment has the same root fix. The extracted utility eliminates the need for the mock because all behaviour now lives in a pure, directly testable function.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/MultipartAssets/ImageAssetCard/ImageAssetLarge/ImageAssetLarge.test.tsx
arjita-mitra: its no longer needed

apps/webapp/src/script/util/ImageUtil.ts
```diff
@@ -179,3 +179,11 @@ export const isPreviewableImage = ({
...
   return !!normalizedExtension && PREVIEWABLE_IMAGE_EXTENSIONS.has(normalizedExtension);
 };
+
+export const getBestPreviewSource = (
+  extension: string,
+  fileUrl?: string,
+  filePreviewUrl?: string,
+): string | undefined => {
+  return isPreviewableImage({extension}) ? (fileUrl ?? filePreviewUrl) : filePreviewUrl;
+};
```
screendriver:
```suggestion
type GetBestPreviewSourceOptions = {
  readonly fileExtension: string;
  readonly fileUrl: Maybe<string>;
  readonly filePreviewUrl: Maybe<string>;
};

export function getBestPreviewSource(options: GetBestPreviewSourceOptions): Maybe<string> {
  const {fileExtension, fileUrl, filePreviewUrl} = options;

  if (!isPreviewableImage({extension: fileExtension})) {
    return filePreviewUrl;
  }

  return fileUrl.or(filePreviewUrl);
}
```

I recommend making fileUrl and filePreviewUrl explicit Maybe<string> values in getBestPreviewSource’s options object. This removes nullable/undefined handling from the function body, keeps the API fully functional, and makes absence/presence semantics explicit at the call site. It also lets us compose with or directly and avoid ternary-based fallback logic.

screendriver (approved):

PR #20578, opened 2026-03-05
feat: add startup URL feature flag override and expose typed lookup in root context [WPB-23826]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23826" title="WPB-23826" target="_blank">TaskWPB-23826</a>  [Web] Optimize websocket implementation
  </td></table>
  <br />
 

# Pull Request

## Summary

We need a safe way to toggle specific webapp behavior at startup via URL for controlled rollout/verification scenarios, without broadening runtime flag surface or introducing re-evaluation during app usage.

- Added a dedicated startup feature-flag slice for URL parsing.
- Added support for enabled-features query parsing at startup only.
- Limited URL-overridable flags to one explicit whitelist entry:
  - reliable-websocket-connection
- Exposed a minimal, typed feature-flag lookup via root context:
  - isFeatureFlagEnabled(...)
- Updated relevant tests to cover parsing behavior and context contract.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/featureFlags/startupFeatureFlags.ts
+++ b/apps/webapp/src/script/featureFlags/startupFeatureFlags.ts
@@ -52,10 +52,8 @@ function toEnabledFeatureNameList(serializedFeatureNames: string): readonly Star
 function readEnabledFeatureNameListFromLocationSearch(locationSearch: string): readonly StartupFeatureFlagName[] {
   const queryParameters = new URLSearchParams(Maybe.of(locationSearch).unwrapOr(''));
   const enabledFeaturesParameterValue = queryParameters.get(startupFeatureFlagQueryParameterName);
-  const enabledFeatureNameListOrNull =
-    enabledFeaturesParameterValue === null ? null : toEnabledFeatureNameList(enabledFeaturesParameterValue);
 
-  return Maybe.of(enabledFeatureNameListOrNull).unwrapOr([]);
+  return Maybe.of(enabledFeaturesParameterValue).map(toEnabledFeatureNameList).unwrapOr([]);
 }
 
 export function createStartupFeatureFlagMapFromLocationSearch(locationSearch: string): StartupFeatureFlagMap {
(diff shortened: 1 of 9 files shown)


apps/webapp/src/script/featureFlags/startupFeatureFlags.ts
```diff
@@ -0,0 +1,83 @@
...
+function toEnabledFeatureNameList(serializedFeatureNames: string): readonly StartupFeatureFlagName[] {
+  return serializedFeatureNames
+    .split(',')
+    .map(trimFeatureFlagName)
+    .filter(isNonEmptyFeatureFlagName)
+    .filter(isAllowedStartupFeatureFlagName);
+}
+
+function readEnabledFeatureNameListFromLocationSearch(locationSearch: string): readonly StartupFeatureFlagName[] {
+  const queryParameters = new URLSearchParams(Maybe.of(locationSearch).unwrapOr(''));
```
thisisamir98, line 53: new URLSearchParams

is this supported in node.js? shouldn't it be a dependency?

screendriver: Available since [Node.js v6.13.0](https://nodejs.org/api/url.html#class-urlsearchparams) 🙂

apps/webapp/src/script/featureFlags/startupFeatureFlags.ts
```diff
@@ -0,0 +1,83 @@
...
+  const enabledFeaturesParameterValue = queryParameters.get(startupFeatureFlagQueryParameterName);
+  const enabledFeatureNameListOrNull =
+    enabledFeaturesParameterValue === null ? null : toEnabledFeatureNameList(enabledFeaturesParameterValue);
```
thisisamir98: usage of null smells a bit?

screendriver: Good point. Was also unsure because it is 100% internal to the scope of the module. But I changed it in 293df43cc52905e90b6369fbf1f6f0aa66f76568 🙂

thisisamir98 (approved):

PR #20589, opened 2026-03-06
feat(webapp): add cleanup stack utility as explicit resource management alternative [WPB-23826]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23826" title="WPB-23826" target="_blank">EpicWPB-23826</a>  [Web] Optimize websocket implementation
  </td></table>
  <br />
 

# Pull Request

## Summary

We need deterministic teardown that is simple, explicit, and not tied to language-level explicit resource management features. Instead of using [explicit resource management](https://github.com/tc39/proposal-explicit-resource-management), which does unfortunately [not exist in Safari](https://caniuse.com/wf-explicit-resource-management) we will have our own explicit resource management alternative.

We will use this in the upcoming WebSocket connection changes.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/cleanupStack/cleanupStack.test.ts
+++ b/apps/webapp/src/script/cleanupStack/cleanupStack.test.ts
@@ -1,0 +1,81 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


screendriver:
Not needed because we only support the newest Chrome versions which already support [disposable stack](https://caniuse.com/mdn-javascript_builtins_disposablestack_dispose)

PR #20620, opened 2026-03-09
feat: add an independent foundation for a future reliable WebSocket connection [WPB-23826]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23826" title="WPB-23826" target="_blank">EpicWPB-23826</a>  [Web] Optimize websocket implementation
  </td></table>
  <br />
 

# Pull Request

## Summary

We want a new WebSocket connection path that can be developed, tested, and observed independently from the existing application flow. The intent of this pull request is to create a safe, isolated starting point for a future reliable WebSocket connection without affecting the current production path.

This change introduces a separate, feature-flagged WebSocket connection foundation with clear ownership at the application boundary. This pull request does not replace the existing production WebSocket/event flow. It does not integrate the new connection path with the legacy event repository or Knockout-based implementation.

It also makes the connection state visible for manual verification when the feature is enabled. The new path is only enabled when the application is started with:

```?enabled-features=reliable-websocket-connection```

Unfortunately the previous one https://github.com/wireapp/wire-webapp/pull/20618 was closed accidentally. 

Screen Recording 2026-03-09 at 15 44 01

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### CODE REVIEW

File: apps/webapp/src/script/webSocketConnection/createManagedWebSocketConnection.ts [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+}
+
+export function createManagedWebSocketConnection(
+  dependencies: CreateManagedWebSocketConnectionDependencies,
+): ManagedWebSocketConnection {
+  const {createWebSocketConnection} = dependencies;
+  const webSocketConnectionStateMachineActor = createActor(createWebSocketConnectionStateMachine());
+  const connectionStateListenerSet = new Set<WebSocketConnectionStateListener>();
+  const managedWebSocketConnectionCleanupStack = createCleanupStack();
+  let activeManagedWebSocketConnection: ActiveManagedWebSocketConnection | undefined;
```
thisisamir98: why not use a maybe here?

apps/webapp/src/script/webSocketConnection/createManagedWebSocketConnection.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  webSocketConnectionStateMachineActor.start();
+
+  const actorSubscription = webSocketConnectionStateMachineActor.subscribe(snapshot => {
+    const currentConnectionState = snapshot.value as WebSocketConnectionStateMachineState;
```
thisisamir98: isn't there anyway to avoid the type casting?

screendriver: You were looking on an old commit. It was already changed in 9822b20770a681756e019aac07094fd185c08350

apps/webapp/src/script/webSocketConnection/createManagedWebSocketConnection.ts
screendriver: For that old version: yes, `Maybe` would have been the better model. that imperative `activeManagedWebSocketConnection` variable was removed when lifecycle ownership moved into the state machine in commit 67c05776d5a0173109c90a9dfb37c9ef58924d48. So you are looking at an old commit.

apps/webapp/src/script/webSocketConnection/managedWebSocketConnectionStateMachine.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export type ManagedWebSocketConnectionEventType = 'open' | 'close' | 'error';
+
+export type ManagedWebSocketConnectionTransport = {
+  readonly readyState: number;
+  readonly send: (message: string) => void;
+  readonly close: () => void;
+  readonly addEventListener: (type: ManagedWebSocketConnectionEventType, listener: (event: Event) => void) => void;
+  readonly removeEventListener: (type: ManagedWebSocketConnectionEventType, listener: (event: Event) => void) => void;
+};
```
thisisamir98, line 43: isn't this kind of same as https://github.com/wireapp/wire-webapp/pull/20620/changes/4350ac83f957481c6fa6d59671ddc40527909582#diff-8898c54f0df6a5647e06e1d37203b254476715f20d722c7803dfa89f9ec2fb49R34-R40

?

apps/webapp/src/script/main/index.tsx
```diff
@@ -70,15 +76,46 @@ document.addEventListener('DOMContentLoaded', async () => {
...
+        isConnectivityAvailable() {
+          return navigator.onLine;
+        },
+        subscribeToConnectivityStatusChanges(
+          subscribeToConnectivityStatusChangesArguments: SubscribeToConnectivityStatusChangesArguments,
+        ) {
+          const {onConnectivityBecameAvailable, onConnectivityBecameUnavailable} =
+            subscribeToConnectivityStatusChangesArguments;
+
+          window.addEventListener('online', onConnectivityBecameAvailable);
```
thisisamir98: can we Prefer `globalThis` over `window`?

screendriver: Good point 👍 Introduced by accident. Fixed in cf788451323409768d20a7a46d919c1f08564360

apps/webapp/src/script/webSocketConnection/managedWebSocketConnectionStateMachine.ts
screendriver: That change was intentional to keep DOM/browser types out of the core.

apps/webapp/src/script/webSocketConnection/managedWebSocketConnectionStateMachine.ts
```diff
@@ -0,0 +1,375 @@
...
+              [managedWebSocketConnectionStateMachineEventType.connect]: {
+                actions: managedWebSocketConnectionStateMachineSetup.assign({
+                  connectionUrl: (
+                    managedWebSocketConnectionStateMachineAssignArguments: ManagedWebSocketConnectionStateMachineAssignArguments,
+                  ) => {
+                    const {event} = managedWebSocketConnectionStateMachineAssignArguments;
+
+                    if (event.type !== managedWebSocketConnectionStateMachineEventType.connect) {
+                      throw new Error('Managed WebSocket connection state machine expected a connect event');
+                    }
```
e-maad, line 257: Do we even need it?
Are we not already in managedWebSocketConnectionStateMachineEventType.connect state ?

screendriver: We will iterate on the whole thing **a lot** as discussed internally 🙂

e-maad (approved):

screendriver:
We will split it into multiple, more focused, pull requests.

PR #20621, opened 2026-03-09
fix: collabora copy paste not working(WPB-23544)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23544" title="WPB-23544" target="_blank">BugWPB-23544</a>  [Web] Not possible to copy/paste in Collabora
  </td></table>
  <br />
 

# Pull Request

## Summary

Copy and paste did not work inside the Collabora editor when opened via Wire web or desktop.

- **Copy** did not work at all
- **Paste** only worked after multiple attempts
- Collabora showed a modal indicating it had limited clipboard access

The root cause:
1. The Collabora iframe was missing the `allow="clipboard-read; clipboard-write"` attribute, so the browser never granted clipboard access to the iframe
2. Permissions-Policy was missing too

## Solution

Added `allow="clipboard-read; clipboard-write"` to the Collabora iframe in `FileEditor.tsx`, but only after the editor URL passes `validateCollaboraUrl`. The attribute is never rendered for an invalid URL.

### URL validation (`validateCollaboraUrl.ts`)

New pure function `validateCollaboraUrl(url: Maybe<string>, trustedOrigin: string)` returns `Result<string, CollaboraUrlError>` with three typed failure reasons:

 `empty` | No editor URL in the API response |
  `insecure` | URL is not HTTPS — clipboard grant would go to an insecure origin |
  `untrusted` | URL origin does not match `CELLS_PYDIO_URL` — rejects unexpected HTTPS domains

### `Permissions-Policy` header — scoped to Collabora origin (`Server.ts`)

Permissions-Policy with an explicit allowlist scoped to `CELLS_PYDIO_URL`:

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/FileFullscreenModal/FileEditor/validateCollaboraUrl.ts
+++ b/apps/webapp/src/script/components/FileFullscreenModal/FileEditor/validateCollaboraUrl.ts
@@ -30,7 +30,7 @@ const HTTPS_URL_PATTERN = /^https:\/\/\S+$/;
  * Validates a Collabora editor URL is safe to embed with clipboard permissions.
  */
 export const validateCollaboraUrl = (url: Maybe<string>, trustedOrigin: string): Result<string, CollaboraUrlError> => {
-  if (!url.isJust) {
+  if (!url.isJust || url.value === '') {
     return Result.err({reason: 'empty'});
   }
 
--- a/apps/webapp/src/script/components/FileFullscreenModal/FileEditor/validateCollaboraUrl.ts
+++ b/apps/webapp/src/script/components/FileFullscreenModal/FileEditor/validateCollaboraUrl.ts
@@ -1,0 +1,46 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Maybe, Result} from 'true-myth';
+
+export type CollaboraUrlError =
+  | {readonly reason: 'empty'}
+  | {readonly reason: 'insecure'; readonly url: string}
+  | {readonly reason: 'untrusted'; readonly url: string; readonly trustedOrigin: string};
+
+const HTTPS_URL_PATTERN = /^https:\/\/\S+$/;
+
+/**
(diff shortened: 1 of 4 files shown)


thisisamir98 (dismissed):

zskhan (dismissed):

apps/webapp/src/script/components/FileFullscreenModal/FileEditor/validateCollaboraUrl.ts
```diff
@@ -0,0 +1,46 @@
...
+  | {readonly reason: 'empty'}
+  | {readonly reason: 'insecure'; readonly url: string}
+  | {readonly reason: 'untrusted'; readonly url: string; readonly trustedOrigin: string};
+
+const HTTPS_URL_PATTERN = /^https:\/\/\S+$/;
+
+/**
+ * Validates a Collabora editor URL is safe to embed with clipboard permissions.
+ */
+export const validateCollaboraUrl = (url: Maybe<string>, trustedOrigin: string): Result<string, CollaboraUrlError> => {
```
screendriver, line 48: The current regex is weaker than the subsequent `new URL(...)` parse. A malformed string can pass the regex and then throw (which is what we try to avoid when using a `Result`). Also there is a test missing for that.

```ts
function parseUrl(url: string): Maybe<URL> {
  try {
    return Maybe.just(new URL(url));
  } catch {
    return Maybe.nothing();
  }
}

function validateSecureUrl(url: string): Result<URL, CollaboraUrlError> {
  const parsedUrl = parseUrl(url);

  if (!parsedUrl.isJust || parsedUrl.value.protocol !== 'https:') {
    return Result.err({reason: 'insecure', url});
  }

  return Result.ok(parsedUrl.value);
}

/**
 * Validates a Collabora editor URL is safe to embed with clipboard permissions.
 */
export function validateCollaboraUrl(url: Maybe<string>, trustedOrigin: string): Result<string, CollaboraUrlError> {
  if (!url.isJust || url.value === '') {
    return Result.err({reason: 'empty'});
  }

  const collaboraUrlValidation = validateSecureUrl(url.value);

  if (result.isErr(collaboraUrlValidation)) {
    return Result.err(collaboraUrlValidation.error);
  }

  const trustedOriginUrl = parseUrl(trustedOrigin);

  if (
    trustedOriginUrl.isJust &&
    trustedOriginUrl.value.protocol === 'https:' &&
    collaboraUrlValidation.value.origin !== trustedOriginUrl.value.origin
  ) {
    return Result.err({reason: 'untrusted', url: url.value, trustedOrigin});
  }

  return Result.ok(url.value);
}
```

And a test could look like

```ts
it('returns insecure error for malformed https url', () => {
  const url = 'https://[cells.example.com/editor';
  const result = validateCollaboraUrl(Maybe.of(url), TRUSTED_ORIGIN);

  expect(unwrapErr(result)).toEqual({reason: 'insecure', url});
});
```

arjita-mitra: ok, understood your changes. 

- Replaced regex only validation with non throwing URL parsing
- Added validateSecureUrl helper to enforce https: after parsing
- Added missing regression test for malformed https:// URL that previously could pass regex and then throw

screendriver (approved):

PR #20640, opened 2026-03-11
test(WPB-19958): add regression tests localization

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19958" title="WPB-19958" target="_blank">TaskWPB-19958</a>  [Web/QA] Write the Localization regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

- Add regression test for localization for mostly german browser
- Add 'locale' option for creating a user to switch the language

---

## Accessibility (required)

---


apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+      await expect(messagePlaceholder).toHaveText('Eine Nachricht schreiben');
+
+      await components.conversationSidebar().allConverationsButton.click();
+      await pages.conversationList().clickConversationOptions(userB.fullName);
+
+      const menuList = pages.conversationList().page.getByTestId('conversation-list-options-menu');
+
+      // check for
+      //notifications button
+      await expect(menuList.first()).toHaveText('Benachrichtigungen');
```
JacquelineLehner: These locators are like this on purpose, otherwise they wouldn't be accessible in german.

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
```diff
@@ -0,0 +1,37 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Locator, Page} from '@playwright/test';
+
+export class LandingwelcomePage {
```
e-maad:
```suggestion
export class LandingWelcomePage {
```

Also the filename.

apps/webapp/test/e2e_tests/pageManager/index.ts
```diff
@@ -185,6 +194,7 @@ export class PageManager {
         this.getOrCreate('webapp.pages.verification', () => new EmailVerificationPage(this.page)),
       setUsername: () => this.getOrCreate('webapp.pages.setUsername', () => new SetUsernamePage(this.page)),
       conversationJoin: () => this.getOrCreate('webapp.pages.conversationJoin', () => ConversationJoinPage(this.page)),
+      landingwelcome: () => this.getOrCreate('webapp.pages.landingwelcome', () => new LandingwelcomePage(this.page)),
```
zhannach: Let's stick to a consistent naming convention:
`landingWelcome: () => this.getOrCreate('webapp.pages.landingwelcome', () => new LandingWelcomePage(this.page))`

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  readonly header: Locator;
+  readonly emailInput: Locator;
+  readonly signInButton: Locator;
+
+
+  constructor(page: Page) {
+    this.page = page;
+    this.header = page.locator('#sso-login-heading-text');
+    this.emailInput = page.locator('input[name="sso-code-email"]');
```
zhannach: Can you use getByRole here?
 e.g. page.getByRole('textbox', { name: 'sso-code-email' })

apps/webapp/test/e2e_tests/backend/authRepository.e2e.ts
```diff
@@ -30,6 +30,7 @@ export class AuthRepositoryE2E extends BackendClientE2E {
       password: user.password,
       name: `${user.firstName} ${user.lastName}`,
       email: user.email,
+      locale: 'de-DE',
```
markbrockhoff: Could you please make this dynamic instead of hardcoding the locale for newly registered users? I'm afraid it could break some of other tests since the default language for most of them is english ^^')

apps/webapp/test/e2e_tests/backend/inbucketClient.e2e.ts
```diff
@@ -116,8 +116,8 @@ export class InbucketClientE2E {
     return matchingUrl;
   }
 
-  private async getLatestEmail(email: string) {
-    return await this.axiosInstance.get<{subject: string; body: {text: string}}>(`/api/v1/mailbox/${email}/latest`, {
```
markbrockhoff, line 120: Why did you remove the type for the response? (Was this maybe a mistake while rebasing?)

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    const enPage = await createPage(enContext);
+    const enPageManager = PageManager.from(enPage);
+    const {pages: enPages} = enPageManager.webapp;
+
+    await enPageManager.openWelcomePage();
+
+    await expect(enPages.landingwelcome().header).toHaveText('Welcome to Wire!');
+    await expect(enPages.landingwelcome().emailInput).toHaveAttribute('placeholder', 'Email or SSO code');
+    await expect(enPages.landingwelcome().emailInput).toHaveAttribute('placeholder', 'Email or SSO code');
```
zhannach: It looks like lines 60 and 61 are identical. We should probably remove the duplicate.

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
markbrockhoff: This POM looks awfully similar to the `singleSignOn.page.ts` are you sure it should be a standalone page? If not we should instead add the missing locators to the existing POM.

apps/webapp/test/e2e_tests/pageManager/index.ts
```diff
@@ -109,6 +110,14 @@ export class PageManager {
     await this.page.goto(`${webAppPath}auth/#/login`);
   };
 
+  openRegistrationPage = async () => {
+    await this.page.goto(`${webAppPath}auth/#/createaccount`);
+  };
+
+  openWelcomePage = async () => {
```
markbrockhoff: TBH I don't think this should be called "WelcomePage", the old tests called it like that but I feel like this will just cause confusion to us devs since we usually refer to it as the SSO page.

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
e-maad: I can see that you are using hardcoded strings for assertion. Is it possible to use de-DE.json?

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    const user = getUser();
+
+    await pages.registration().nameInput.fill(user.fullName);
+    await pages.registration().emailInput.fill(user.email);
+    await pages.registration().passwordInput.fill(user.password);
+    await pages.registration().confirmPasswordInput.fill(user.password);
+
+    // Click the top-left corner of the label to toggle the checkbox and avoid clicking the embedded "Terms of Service" link.
+    await pages.registration().termsLabel.click({position: {x: 5, y: 5}});
```
e-maad: Can't we locate checkbox and trigger click?

e-maad (commented):
Added some suggestions.

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    await expect(enPages.landingwelcome().header).toHaveText('Welcome to Wire!');
+    await expect(enPages.landingwelcome().emailInput).toHaveAttribute('placeholder', 'Email or SSO code');
+    await expect(enPages.landingwelcome().emailInput).toHaveAttribute('placeholder', 'Email or SSO code');
+    await expect(enPages.landingwelcome().signInButton).toHaveText('Log in');
+  });
+
+  test('Verify support pages are opened in language de', {tag: ['@TC-3456', '@regression']}, async ({createPage}) => {
+    const mainPage = await createPage(withLogin(userA));
+
+    const {components: components} = (await PageManager.from(mainPage)).webapp;
```
markbrockhoff:
```suggestion
    const {components} = PageManager.from(mainPage).webapp;
```

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    const [newPage] = await Promise.all([
+      mainPage.waitForEvent('popup'),
+      components.conversationSidebar().supportButton.click(),
+    ]);
+
+    await newPage.waitForLoadState();
```
markbrockhoff: Why is this needed? Shouldn't the new page be opened with the url directly?
Also: the regex for the opened URL looks a bit strange, what is the "/hc/" part in there? Wouldn't it be sufficient to check for "/de/" to be in the URL e.g. `await expect(newPage).toHaveURL(/\/de/);`?

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    await expect(pages.login().header).toHaveText('Willkommen bei Wire!');
+    await expect(pages.login().emailInput).toHaveAttribute('placeholder', 'E-Mail-Adresse oder Benutzername');
+    await expect(pages.login().signInButton).toHaveText('Anmelden');
+  });
+
+  test(
+    'Verify conversation view and list has German-localized strings',
+    {tag: ['@TC-1275', '@regression']},
+    async ({createPage}) => {
+      const {pages, modals, components} = (await PageManager.from(createPage(withLogin(userA)))).webapp;
```
markbrockhoff: 🎨 Hint: You can make the sonar false positive go away and get rid of those extra brackets by putting the `await` inside the call to `PageManager.from()` e.g.:

```suggestion
      const {pages, modals, components} = PageManager.from(await createPage(withLogin(userA))).webapp;
```

`PageManager.from()` is overloaded to accept either a `Page` or `Promise<Page>` so you can pass either, but if you pass a Page / fulfilled page promise directly the function itself no longer needs to be awaited ;)

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      // Connect users manually due to the current locators don't work in german browser
+      await components.conversationSidebar().connectButton.click();
+      await pages.startUI().component.getByPlaceholder('Personen suchen').fill(userB.fullName);
+      await pages.startUI().component.getByRole('button', {name: userB.fullName}).click();
+      await modals.userProfile().startConversationButton.click();
+
+      await pages.conversationList().openConversation(userB.fullName);
+      const messagePlaceholder = pages.conversation().page.locator('[data-uie-name="input-placeholder"]');
```
markbrockhoff: Please don't access the page through the POM, if you need it use the page you already have in the test since it's returned by `createPage()`

markbrockhoff: (Same applies to all other places within this file where you did this ;))

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+      //notifications button
+      await expect(menuList.first()).toHaveText('Benachrichtigungen');
+      //favorite button
+      await expect(menuList.nth(1)).toHaveText('Zu Favoriten hinzufügen');
+      //move to button
+      await expect(menuList.nth(2)).toHaveText('Verschieben nach');
+      //archive button
+      await expect(menuList.nth(3)).toHaveText('Archivieren');
+      //clear content button
+      await expect(menuList.nth(4)).toHaveText('Unterhaltungsverlauf löschen');
```
markbrockhoff: 🔧 This seems a bit strict to me, what if the order of the menu items changes at some point? The goal of this test is not to verify their order, yet it would break if it changed. You can make this assertion independent of the order by using an [Asymmetric Matcher](https://jestjs.io/docs/expect#asymmetric-matchers) (Don't worry about the link pointing to jest while we're using playwright, they are based on the same `expect` library under the hood)

E.g.:
```suggestion
      await expect
        .poll(async () => await menuList.allInnerTexts())
        .toEqual(
          expect.arrayContaining([
            'Benachrichtigungen', //notifications button
            'Zu Favoriten hinzufügen', //favorite button
            'Verschieben nach', //move to button
            'Archivieren', //archive button
            'Unterhaltungsverlauf löschen', //clear content button
          ]),
        );
```

The `.poll()` is needed to make this retryable and avoid flake ;)

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    await pageManager.openRegistrationPage();
+    const {pages} = pageManager.webapp;
+
+    const user = getUser();
+
+    await pages.registration().nameInput.fill(user.fullName);
+    await pages.registration().emailInput.fill(user.email);
+    await pages.registration().passwordInput.fill(user.password);
+    await pages.registration().confirmPasswordInput.fill(user.password);
```
markbrockhoff: 🔧  There's an existing util for filling this form, feel free to use it ;)

```suggestion
    await pages.registration().fillInUserInfo(user);
```

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    // Click the top-left corner of the label to toggle the checkbox and avoid clicking the embedded "Terms of Service" link.
+    await pages.registration().termsLabel.click({position: {x: 5, y: 5}});
+    await pages.registration().submitButton.click();
+
+    await expect
+      .poll(
+        async () => {
+          const email = await api.inbucket.getLatestEmail(user.email);
+          return email.data?.body?.text as string;
```
markbrockhoff: The `as string` shouldn't be necessary here, please see my comment in `inbucketClient.e2e.ts` ;)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
markbrockhoff: It won't work for the emails but could make maintenance a bit easier for the ui translations 👍 

FYI @JacquelineLehner you can import the translations from the `de-DE.json` file like so:
```ts
import {conversationDetailsActionNotifications} from 'I18n/de-DE.json';
```

apps/webapp/test/e2e_tests/pageManager/index.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/pageManager/index.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
JacquelineLehner: two are better than one I guess xD
removed 👍

apps/webapp/test/e2e_tests/pageManager/webapp/pages/landingwelcome.page.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/backend/inbucketClient.e2e.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: the problem is that `checkbox.click()` doesn't work somehow, since Playwright couldn't find the locator. I tried now `termsCheckbox.dispatchEvent('click')` which seems to do the trick.

apps/webapp/test/e2e_tests/backend/authRepository.e2e.ts
JacquelineLehner: changed it

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
```diff
@@ -127,11 +128,11 @@ test.describe('Localization', () => {
         .poll(async () => await menuList.allInnerTexts())
         .toEqual(
           expect.arrayContaining([
-            'Benachrichtigungen', //notifications button
-            'Zu Favoriten hinzufügen', //favorite button
-            'Verschieben nach', //move to button
-            'Archivieren', //archive button
-            'Unterhaltungsverlauf löschen', //clear content button
+            deTranslations['conversationDetailsActionNotifications'], // notifications button
```
markbrockhoff: Nice, but could you please try to match the translation you use for the assertion not just by the actual string which is translated but also by the semantics of the translation key? E.g. here you used a key prefixed with "conversationDetailsAction" which means it's probably used within the conversation details and not the conversationPopover. E.g. `conversationsPopoverNotificationSettings` is probably the semantically more correct key. Reason is that if the translation key for only one of the locations is changed your test would break, which is what we're trying to avoid by importing the translations ;)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/pageManager/webapp/modals/confirmLogout.modal.ts
```diff
@@ -22,18 +22,16 @@ import {Page, Locator} from '@playwright/test';
...
 
   constructor(page: Page) {
     super(page);
 
     this.modalCheckbox = this.modal.getByTestId('modal-option-checkbox');
-    this.deleteDeviceCheckbox = this.modal.getByText('Delete all your personal information');
   }
 
   async toggleModalCheck() {
-    await this.deleteDeviceCheckbox.click();
```
markbrockhoff, line 36: Hint: This change is causing tests depending on the `logoutUser()` function to fail since it's now trying to click the checkbox instead of its label. One solution would be to bring back the second locator but rename it so it's clear that its referring to the label of the modalCheckbox. An other would be to remove the one usage `modalCheckbox` inside `logoutUser()` (it's redundant anyway) and always use the locator for the label in here :)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/pageManager/webapp/modals/confirmLogout.modal.ts
JacquelineLehner: ✅

e-maad (dismissed):

apps/webapp/test/e2e_tests/specs/Localization/localization.spec.ts
JacquelineLehner: ✅

markbrockhoff (approved):

zskhan (approved):

PR #20647, opened 2026-03-11
feat: inject incremental retry backoff through HTTP client construction [WPB-23381]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23381" title="WPB-23381" target="_blank">TaskWPB-23381</a>  [Web] Make sure that the web app applies incremental backoff on 420/429
  </td></table>
  <br />
 

# Pull Request

## Summary

Enable incremental HTTP retry backoff through explicit client construction so the rollout remains controlled by the incremental-http-retry-backoff startup feature flag without widening the shared config contract. This keeps the new transport behavior opt-in, preserves the existing retry path when the flag is disabled, and avoids changing the Core lifecycle inside React rendering.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/service/APIClientSingleton.ts
+++ b/apps/webapp/src/script/service/APIClientSingleton.ts
@@ -22,26 +22,39 @@ import {singleton} from 'tsyringe';
 import {APIClient as APIClientUnconfigured} from '@wireapp/api-client';
 
 import {Config} from '../Config';
+import {StartupFeatureToggleName} from '../featureToggles/startupFeatureToggles';
 
 const wireClientHeaderName = 'Wire-Client';
 const wireClientVersionHeaderName = 'Wire-Client-Version';
 const wireClientIdentifier = 'Web';
+const incrementalHttpRetryBackoffFeatureToggleName = 'incremental-http-retry-backoff';
+
+type APIClientSingletonConfiguration = {
+  readonly isFeatureToggleEnabled?: (featureToggleName: StartupFeatureToggleName) => boolean;
+};
 
 @singleton()
 export class APIClient extends APIClientUnconfigured {
-  constructor() {
+  constructor(apiClientSingletonConfiguration: APIClientSingletonConfiguration = {}) {
     const webAppConfiguration = Config.getConfig();
+    const {isFeatureToggleEnabled} = apiClientSingletonConfiguration;
 
-    super({
-      headers: {
-        [wireClientHeaderName]: wireClientIdentifier,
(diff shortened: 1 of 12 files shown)


libraries/api-client/src/http/HttpClient.ts
```diff
@@ -66,6 +83,10 @@ export class HttpClient extends EventEmitter {
   private connectionState: ConnectionState;
   private readonly requestQueue: PriorityQueue;
   private readonly backOffQueue: PriorityQueue;
+  private readonly incrementalRetryBackoffPolicy = createIncrementalRetryBackoffPolicy();
+  private readonly incrementalRetryBackoffRunner;
+  private incrementalRetryBackoffState: IncrementalRetryBackoffState;
+  private shouldUseIncrementalRetryBackoff = false;
```
thisisamir98, line 87: Member 'shouldUseIncrementalRetryBackoff' is never reassigned; mark it as `readonly`.

apps/webapp/src/script/service/APIClientSingleton.ts
```diff
@@ -22,26 +22,39 @@ import {singleton} from 'tsyringe';
 import {APIClient as APIClientUnconfigured} from '@wireapp/api-client';
 
 import {Config} from '../Config';
+import {StartupFeatureToggleName} from '../featureToggles/startupFeatureToggles';
 
 const wireClientHeaderName = 'Wire-Client';
 const wireClientVersionHeaderName = 'Wire-Client-Version';
 const wireClientIdentifier = 'Web';
+const incrementalHttpRetryBackoffFeatureToggleName = 'incremental-http-retry-backoff';
```
thisisamir98: would it be a good idea to have a dedicated file for all the feature toggle constants?

libraries/api-client/src/http/HttpClient.ts
screendriver: [It is](https://github.com/wireapp/wire-webapp/pull/20647/changes#diff-ad2c61641cac7f7e831dfc34f0fec9284f82275f9660b1a835327aeed9f5bb34R131)

thisisamir98 (dismissed):

apps/webapp/src/script/service/APIClientSingleton.ts
screendriver: Yes. Good point 👍 Changed in d7fbb4a9575482700e0ce2bd54221b348738ffe5

thisisamir98 (approved):

PR #20678, opened 2026-03-13
test: write the Sending Assets regression tests [WPB-19969]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19969" title="WPB-19969" target="_blank">TaskWPB-19969</a>  [Web/QA] Write the Sending Assets regression tests in Playwright
  </td></table>
  <br />
 

# Pull Request

## Summary

Add the Sending Assets regression e2e tests.
---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
@@ -78,48 +78,48 @@ test.describe('Sending Assets', () => {
   });
 
   test('Verify you can delete an audio message', {tag: ['@TC-111', '@regression']}, async ({createPage}) => {
-    const [userBPage] = await Promise.all([
-      createPage(withLogin(userB), withConnectedUser(userA)),
-      createPage(withLogin(userA)),
+    const [userAPage] = await Promise.all([
+      createPage(withLogin(userA), withConnectedUser(userB)),
+      createPage(withLogin(userB)),
     ]);
 
-    const {pages} = PageManager.from(userBPage).webapp;
+    const {pages} = PageManager.from(userAPage).webapp;
 
-    await pages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
-    await shareAssetHelper(getAudioFilePath(), userBPage, userBPage.getByRole('button', {name: 'Add file'}));
+    await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+    await shareAssetHelper(getAudioFilePath(), userAPage, userAPage.getByRole('button', {name: 'Add file'}));
 
-    const message = pages.conversation().getMessage({sender: userB});
+    const message = pages.conversation().getMessage({sender: userA});
     await expect(message).toBeVisible();
     await pages.conversation().deleteMessage(message, 'Everyone');
     await expect(message).not.toBeVisible();
   });
 
   test('I want to drag & drop a file into a conversation', {tag: ['@TC-497', '@regression']}, async ({createPage}) => {
-    const [userBPage] = await Promise.all([
-      createPage(withLogin(userB), withConnectedUser(userA)),
-      createPage(withLogin(userA)),
+    const [userAPage] = await Promise.all([
+      createPage(withLogin(userA), withConnectedUser(userB)),
+      createPage(withLogin(userB)),
     ]);
 
-    const {pages, modals} = PageManager.from(userBPage).webapp;
+    const {pages, modals} = PageManager.from(userAPage).webapp;
 
     const buffer = await readLocalFile(getTextFilePath());
     const targetSelector = '#conversation-input-bar';
 
     await test.step('Go to any 1:1 conversation', async () => {
-      await pages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+      await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
     });
 
     await test.step('User B can drag & drop a file into a conversation', async () => {
-      await dragAndDropFiles(userBPage, targetSelector, {
+      await dragAndDropFiles(userAPage, targetSelector, {
         buffer,
         fileName: TextFileName,
       });
-      await expect(pages.conversation().getMessage({sender: userB})).toBeVisible();
-      await expect(pages.conversation().getMessage({sender: userB})).toHaveCount(1);
+      await expect(pages.conversation().getMessage({sender: userA})).toBeVisible();
+      await expect(pages.conversation().getMessage({sender: userA})).toHaveCount(1);
     });
 
     await test.step('Verify User B sees an error message when he drops 11 files', async () => {
-      await dragAndDropFiles(userBPage, targetSelector, {
+      await dragAndDropFiles(userAPage, targetSelector, {
         buffer,
         fileName: TextFileName,
         count: 11,
@@ -127,7 +127,7 @@ test.describe('Sending Assets', () => {
 
       await expect(modals.acknowledge().modalTitle).toContainText(`Too many files at once`);
       await modals.acknowledge().clickAction();
-      await expect(pages.conversation().getMessage({sender: userB})).toHaveCount(1);
+      await expect(pages.conversation().getMessage({sender: userA})).toHaveCount(1);
     });
   });
 
@@ -135,12 +135,12 @@ test.describe('Sending Assets', () => {
     'I want to copy & paste an image into a conversation',
     {tag: ['@TC-498', '@regression']},
     async ({createPage}) => {
-      const [userBPage] = await Promise.all([
-        createPage(withLogin(userB), withConnectedUser(userA)),
-        createPage(withLogin(userA)),
+      const [userAPage] = await Promise.all([
+        createPage(withLogin(userA), withConnectedUser(userB)),
+        createPage(withLogin(userB)),
       ]);
 
-      const {pages} = PageManager.from(userBPage).webapp;
+      const {pages} = PageManager.from(userAPage).webapp;
 
       const buffer = await readLocalFile(getImageFilePath());
       const targetSelector = '#conversation-input-bar';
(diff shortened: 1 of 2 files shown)


apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+        dt.items.add(file);
+      }
+      return dt;
+    },
+    {buffer: Array.from(buffer), fileName, count},
+  );
+
+  await page.dispatchEvent(selector, 'dragenter', {dataTransfer});
+  await page.dispatchEvent(selector, 'dragover', {dataTransfer});
+  await page.dispatchEvent(selector, 'drop', {dataTransfer});
```
markbrockhoff: Do you really need all of the 3 events? Isn't the only relevant one the "drop" event?

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+  test('Verify you can delete an audio message', {tag: ['@TC-111', '@regression']}, async ({createPage}) => {
+    const [userAPage] = await Promise.all([
+      createPage(withLogin(userA), withConnectedUser(userB)),
+      createPage(withLogin(userB)),
+    ]);
+
+    const {pages} = PageManager.from(userAPage).webapp;
+
+    await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+    await shareAssetHelper(getAudioFilePath(), userAPage, userAPage.getByRole('button', {name: 'Add file'}));
```
markbrockhoff: If you don't use userB there's not really a reason to create a page and log him in

```suggestion
    const userAPage = await createPage(withLogin(userA), withConnectedUser(userB));
    const {pages} = PageManager.from(userAPage).webapp;

    await pages.conversationList().openConversation(userB.fullName);
    await shareAssetHelper(getAudioFilePath(), userAPage, userAPage.getByRole('button', {name: 'Add file'}));
```

markbrockhoff: (Same for all the other tests where only userAPage is used)

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  test('I want to drag & drop a file into a conversation', {tag: ['@TC-497', '@regression']}, async ({createPage}) => {
+    const [userAPage] = await Promise.all([
+      createPage(withLogin(userA), withConnectedUser(userB)),
+      createPage(withLogin(userB)),
+    ]);
+
+    const {pages, modals} = PageManager.from(userAPage).webapp;
+
+    const buffer = await readLocalFile(getTextFilePath());
```
markbrockhoff: Didn't know the repo contained such a "readLocalFile()" util, but IMHO it could just be replaced by the native "readFile" function from "node:fs/promises" e.g.:

```suggestion
    const buffer = await fs.readFile(getTextFilePath());
```

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+      await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+    });
+
+    await test.step('User B can drag & drop a file into a conversation', async () => {
+      await dragAndDropFiles(userAPage, targetSelector, {
+        buffer,
+        fileName: TextFileName,
+      });
+      await expect(pages.conversation().getMessage({sender: userA})).toBeVisible();
+      await expect(pages.conversation().getMessage({sender: userA})).toHaveCount(1);
```
markbrockhoff: `toBeVisible()` and `toHaveCount(1)` almost check the same thing (except visible also checks for the element to be visible while count only checks attached elements)
It seems kind of redundant in this test though to have both.

```suggestion
      await expect(pages.conversation().getMessage({sender: userA})).toHaveCount(1);
```

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+      };
+
+      const pastedFileControls = userAPage.getByTestId('pasted-file-controls');
+      await test.step('Go to any 1:1 conversation', async () => {
+        await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+      });
+
+      await test.step('User A copy paste image into the conversation', async () => {
+        await pages.conversation().messageInput.click();
+        await pasteImage(userAPage, targetSelector, buffer, ImageQRCodeFileName);
```
markbrockhoff: 🔧 You can call `evaluateHandle()` not just on the page but also on any Locator directly. This way you don't need to "re-locate" the element to dispatch the event on within the handle and can continue to use the existing locators ;)

```suggestion
      const buffer = await fs.readFile(getImageFilePath());

      const pasteImage = async (locator: Locator, buffer: Buffer, fileName: string): Promise<void> => {
        await locator.evaluateHandle(
          (target, {bufferArray, name}) => {
            const file = new File([Uint8Array.from(bufferArray)], name, {type: 'image/png'});

            const dataTransfer = new DataTransfer();
            dataTransfer.items.add(file);

            const pasteEvent = new ClipboardEvent('paste', {
              clipboardData: dataTransfer,
              bubbles: true,
              cancelable: true,
            });

            target.dispatchEvent(pasteEvent);
          },
          {bufferArray: Array.from(buffer), name: fileName},
        );
      };

      const pastedFileControls = userAPage.getByTestId('pasted-file-controls');
      await test.step('Go to any 1:1 conversation', async () => {
        await pages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
      });

      await test.step('User A copy paste image into the conversation', async () => {
        await pages.conversation().messageInput.click();
        await pasteImage(pages.conversation().messageInput, buffer, ImageQRCodeFileName);
```

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    const message = userAPages.conversation().getMessage({sender: userA});
+    await expect(message).toBeVisible();
+
+    const messageOptions = await userAPages.conversation().openMessageOptions(message);
+    await messageOptions.getByRole('button', {name: 'Copy'}).click();
+    const isMac = process.platform === 'darwin';
+
+    await userAPages.conversation().messageInput.press(isMac ? 'Meta+V' : 'Control+V');
+    await expect(userAPages.conversation().messageInput).toContainText('Message to copy');
```
markbrockhoff: If possible I think we should try to avoid using platform specific conditions within tests. It's sufficient for this test to verify the text was stored in the users clipboard, e.g.:

```suggestion
    await expect
      .poll(async () => await userAPage.evaluate(() => navigator.clipboard.readText()))
      .toBe('Message to copy');
```

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+
+        await test.step('User A re-downloads file from the conversation', async () => {
+          const downloadedPath = await pages.conversation().downloadFile();
+          const {size: actualSize} = await fs.stat(downloadedPath);
+          const {ext: downloadedExtension, base: downloadedName} = path.parse(downloadedPath);
+
+          expect(downloadedName).toBe(fileBase);
+          expect(downloadedExtension).toBe(extension);
+          expect(actualSize).toBe(expectedSize);
+        });
```
markbrockhoff, line 227: Please make sure to also clean up the previously downloaded file afterwards. Otherwise the test could e.g. pass because the file was already successfully down loaded in a previous run. E.g.:

```suggestion
        await test.step('User A re-downloads file from the conversation', async () => {
          const downloadedPath = await pages.conversation().downloadFile();
          try {
            const {size: actualSize} = await fs.stat(downloadedPath);
            const {ext: downloadedExtension, base: downloadedName} = path.parse(downloadedPath);

            expect(downloadedName).toBe(fileBase);
            expect(downloadedExtension).toBe(extension);
            expect(actualSize).toBe(expectedSize);
          } finally {
            await fs.rm(downloadedPath);
          }
        });
```

markbrockhoff: It might also be a good idea to use a unique output location for the downloaded file. If this parameterized test would run in parallel it would share the same file possibly causing a race condition. For cases like this playwright provides a unique [outputDir](https://playwright.dev/docs/api/class-testinfo#test-info-output-dir) for each test. You can access it on the testInfo which is passed as second parameter to the test function e.g. `test("Name", async ({createPage}, testInfo) => {});`

markbrockhoff (dismissed):

apps/webapp/test/e2e_tests/specs/SendingAssets/sendingAssets.spec.ts
zhannach: yes, I tried. In this case, do we create a conversation with the 'proteus' protocol? I tried filtering by 'mls' but it was unsuccessful.

zskhan (approved):
LGTM. What about the failing critical flow tests?

zhannach:
> LGTM. What about the failing critical flow tests?

These failures match the current nightly run results, so they aren't related to this PR. I’ll investigate the root cause separately.

e-maad (approved):

PR #20680, opened 2026-03-13
fix(webapp): bump lexical and preserve raw markdown in preview-off mode [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Upgrade Lexical packages to 0.41.0 and adjust input-bar serialization so markdown preview-off mode keeps the raw text the user typed.

When preview is enabled, the editor still uses Lexical markdown serialization. When preview is disabled, the editor now uses plain text content for message updates and draft persistence instead of exporting through Lexical markdown.

This fixes the regression where markdown markers like `**` and `_` were escaped before send, causing messages such as `**bold** _italic_` to be sent as literal text instead of formatted content.

Also adds a regression test covering both serialization modes.


apps/webapp/src/script/components/InputBar/InputBarEditor/RichTextEditor/utils/serializeMessage.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('uses Lexical markdown serialization when preview is enabled', () => {
+    const editor = createEditorWithText('**bold** _italic_');
+
+    const message = editor.getEditorState().read(() => serializeMessage(true));
+
+    expect(message).toBe('\\*\\*bold\\*\\* \\_italic\\_');
+  });
+});
```
screendriver: End of line missing

thisisamir98: fixed in https://github.com/wireapp/wire-webapp/pull/20680/commits/24180e7bf5da2f40d84d242925c0829194020a6f

screendriver (approved):

PR #20681, opened 2026-03-13
feat: interrupt active incremental HTTP backoff waits on reset [WPB-23381]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23381" title="WPB-23381" target="_blank">TaskWPB-23381</a>  [Web] Make sure that the web app applies incremental backoff on 420/429
  </td></table>
  <br />
 

# Pull Request

## Summary

Make incremental HTTP backoff resets wake requests that are already waiting so recovery signals such as foreground and connectivity improvements trigger an immediate retry instead of only affecting future delay state. This tightens the transport behavior to match the intended recovery semantics while preserving explicit request-abort handling and the existing flagged rollout boundary.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/libraries/api-client/src/http/HttpClient.ts
+++ b/libraries/api-client/src/http/HttpClient.ts
@@ -83,6 +83,8 @@ export class HttpClient extends EventEmitter {
   private readonly backOffQueue: PriorityQueue;
   private readonly incrementalRetryBackoffPolicy = createIncrementalRetryBackoffPolicy();
   private readonly incrementalRetryBackoffRunner;
+  private incrementalRetryBackoffResetAbortController = new AbortController();
+  private incrementalRetryBackoffResetCount = 0;
   private incrementalRetryBackoffState: IncrementalRetryBackoffState;
   private shouldUseIncrementalRetryBackoff = false;
   public static readonly TOPIC = TOPIC;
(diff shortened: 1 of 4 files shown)


libraries/api-client/src/http/IncrementalRetryBackoffRunner.ts
```diff
@@ -72,10 +80,23 @@ export function createIncrementalRetryBackoffRunner(
...
+          } catch (error: unknown) {
+            const hasRetryBackoffBeenReset = retryBackoffResetCount !== getRetryBackoffResetCount();
+            const wasRequestAborted = isRequestAborted();
+
+            if (hasRetryBackoffBeenReset && !wasRequestAborted) {
+              return runRequestAttemptWithRetry();
+            }
+
+            throw error;
+          }
```
thisisamir98, line 99: is the try catch throw intentional here? is it gonna be changed in future iteration?

libraries/api-client/src/http/HttpClient.ts
```diff
@@ -183,7 +185,20 @@ export class HttpClient extends EventEmitter {
   }
 
   public resetRetryBackoff(): void {
+    this.incrementalRetryBackoffResetCount += 1;
     this.incrementalRetryBackoffState = this.incrementalRetryBackoffPolicy.createInitialIncrementalRetryBackoffState();
+    this.incrementalRetryBackoffResetAbortController.abort();
+    this.incrementalRetryBackoffResetAbortController = new AbortController();
+  }
+
+  private getIncrementalRetryBackoffAbortSignal(abortController?: AbortController): Maybe<AbortSignal> {
```
thisisamir98, line 194: why not turn `this.incrementalRetryBackoffResetAbortController` to a maybe?

screendriver: I wouldn't make it a `Maybe` because it's not optional state. It always exists: we create it eagerly, abort it on reset, then immediately replace it with a new `AbortController`.

libraries/api-client/src/http/IncrementalRetryBackoffRunner.ts
screendriver: This is intentional. This code already lives in a Promise/async error contract. HttpClient, axios, and the tests all already model failures as rejected promises. Introducing `Result` or `Task` will blow up everything even more as it is already here.

thisisamir98 (approved):

PR #20727, opened 2026-03-17
feat: leave subConversation on error during MLS conference join [WPB-23723]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23723" title="WPB-23723" target="_blank">BugWPB-23723</a>  [Web Calling] Remove automatic rejection of all incoming calls on error MLS join in Web App
  </td></table>
  <br />
 

## Summary

- What did I change and why?
Make independent catch block for JoinMLSConversation and answerCall.
So when there is any error on JoinMLSConversation then it will not reject the calls.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/repositories/calling/CallingRepository.ts
+++ b/apps/webapp/src/script/repositories/calling/CallingRepository.ts
@@ -1556,7 +1556,7 @@ export class CallingRepository {
       if (error) {
         this.logger.error('Failed answering call', error);
       }
-      this.leaveCall(conversation.qualifiedId, LEAVE_CALL_REASON.CONVERSATION_DEGRADED);
+      this.leaveCall(conversation.qualifiedId, LEAVE_CALL_REASON.CALL_SETUP_ERROR);
       call.reason(REASON.ERROR);
       if (!!conversation && this.isMLSConference(conversation)) {
         await this.leaveMLSConferenceBecauseError(conversation);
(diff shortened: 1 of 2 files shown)


e-maad (dismissed):

screendriver (dismissed):

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1461,9 +1467,6 @@ export class CallingRepository {
         this.logger.error('Failed answering call', error);
       }
       this.rejectCall(conversation.qualifiedId);
```
EnricoSchw: Capturing the exception is fine, I think so. The real problem, however, is the rejection.

Rejecting means we refuse the call. This, in turn, means we also refuse the call on all other devices of this user. This means that if an error occurs on the web, the call cannot be accepted on any other device from this user.

EnricoSchw (changes requested):

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1460,7 +1460,8 @@ export class CallingRepository {
       if (error) {
         this.logger.error('Failed answering call', error);
       }
-      this.rejectCall(conversation.qualifiedId);
+      this.leaveCall(conversation.qualifiedId, LEAVE_CALL_REASON.CONVERSATION_DEGRADED);
```
zskhan: @EnricoSchw which reason we can reuse here?

e-maad (dismissed):

apps/webapp/src/script/repositories/calling/CallingRepository.ts
EnricoSchw: @zskhan If you don't want to define a new error, then:

MEDIA_STREAM_ERROR = 'media_stream_error',

or better a new error like: CALL_SETUP_ERRORR

EnricoSchw (dismissed):
LGTM

arjita-mitra (approved):

PR #20737, opened 2026-03-18
fix(webapp): avoid window-dependent imports in backup zip worker [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

The backup zip worker imported `toError` from `TypePredicateUtil`, which pulled in app-level dependencies including `@wireapp/api-client`, `@wireapp/commons`, and `logdown`. That dependency chain assumes a browser `window`, causing the worker to fail at runtime with `ReferenceError: window is not defined`.

Replace the shared utility import with a small worker-local error message helper so the worker bundle stays environment-safe and isolated.

### DIFF
--- a/apps/webapp/src/script/repositories/backup/zipWorker.ts
+++ b/apps/webapp/src/script/repositories/backup/zipWorker.ts
@@ -22,11 +22,25 @@ import sodium, {ready} from 'libsodium-wrappers-sumo';
 
 import {ImportError} from './Error';
 
-import {toError} from '../../util/TypePredicateUtil';
-
 type Payload =
   | {type: 'zip'; files: Record<string, ArrayBuffer | string>; encrytionKey?: Uint8Array}
   | {type: 'unzip'; bytes: ArrayBuffer; encrytionKey?: Uint8Array; headerLength?: number};
+
+function getErrorMessage(error: unknown): string {
+  if (error instanceof Error) {
+    return error.message;
+  }
+
+  if (typeof error === 'string') {
+    return error;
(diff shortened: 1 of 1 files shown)


apps/webapp/src/script/repositories/backup/zipWorker.ts
```diff
@@ -22,12 +22,26 @@ import sodium, {ready} from 'libsodium-wrappers-sumo';
 
 import {ImportError} from './Error';
 
-import {toError} from '../../util/TypePredicateUtil';
-
 type Payload =
   | {type: 'zip'; files: Record<string, ArrayBuffer | string>; encrytionKey?: Uint8Array}
   | {type: 'unzip'; bytes: ArrayBuffer; encrytionKey?: Uint8Array; headerLength?: number};
 
+function getErrorMessage(error: unknown): string {
```
screendriver, line 29: Instead of introducing a local `getErrorMessage()` we should change `toError()` to:

```ts
export function toError(errorCandidate: unknown): Error {
  if (is.error(errorCandidate)) {
    return errorCandidate;
  }

  if (isObject(errorCandidate)) {
    if ('message' in errorCandidate && typeof errorCandidate.message === 'string') {
      return new Error(errorCandidate.message, {cause: errorCandidate});
    }

    return new Error('Unknown error', {cause: errorCandidate});
  }

  if (typeof errorCandidate === 'string') {
    return new Error(errorCandidate);
  }

  return new Error('Unknown error', {cause: errorCandidate});
}

```

PR #20743, opened 2026-03-18
refactor(webapp): inject app lock crypto dependency [WPB-24141]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24141" title="WPB-24141" target="_blank">BugWPB-24141</a>  [Web] It's not possible to set an applock, error is shown in console.
  </td></table>
  <br />
 

Refactor AppLockRepository to receive its password-hashing dependency through constructor injection instead of reading libsodium directly at module scope.

This keeps production behavior unchanged through a default libsodium-backed adapter, while making tests simpler and more reliable. App-lock tests now pass plain fake crypto implementations rather than using jest.mock for libsodium.

Also preserve the runtime guard and upstream issue reference for the crypto_pwhash_str type mismatch, where the installed package returns a string despite the published TypeScript definitions declaring Uint8Array.

See: https://github.com/jedisct1/libsodium.js/issues/364


apps/webapp/src/script/repositories/user/AppLockRepository.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import {createUuid} from 'Util/uuid';
+
+import {AppLockCrypto, AppLockRepository} from './AppLockRepository';
+import {AppLockState} from './AppLockState';
+import {UserState} from './UserState';
+
+const mockCryptoPwhashStr = jest.fn();
+const mockCryptoPwhashStrVerify = jest.fn();
+
+const createAppLockRepository = () => {
```
screendriver:
```suggestion
const createAppLockRepository = (): ??? => {
```

apps/webapp/src/script/repositories/user/AppLockRepository.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  const appLockState = new AppLockState();
+  const appLockCrypto: AppLockCrypto = {
+    cryptoPwhashMemLimitInteractive: 1,
+    cryptoPwhashOpsLimitInteractive: 1,
+    ready: Promise.resolve(),
+    cryptoPwhashStr: (code: string, opsLimit: number, memLimit: number) => mockCryptoPwhashStr(code, opsLimit, memLimit),
+    cryptoPwhashStrVerify: (hashedCode: string, code: string) => mockCryptoPwhashStrVerify(hashedCode, code),
+  };
+
+  jest.spyOn(userState, 'self').mockImplementation(ko.observable({id: createUuid()}));
```
screendriver: Just out of curiosity: is it possible to **not** use `jest.spyOn()` at all here? Or do we need to do it because how everything else around is structured?

apps/webapp/src/script/repositories/user/AppLockRepository.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  return new AppLockRepository(userState, appLockState, appLockCrypto);
+};
+
+describe('AppLockRepository', () => {
+  beforeEach(() => {
+    mockCryptoPwhashStr.mockReset();
+    mockCryptoPwhashStrVerify.mockReset();
+    window.localStorage.clear();
```
screendriver:
```suggestion
    globalThis.localStorage.clear();
```

e-maad (approved):
LGTM 👍

apps/webapp/src/script/repositories/user/AppLockRepository.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/20743/commits/ea38bf220d3075b7791a1bfc1153a22e2c4e5ace

apps/webapp/src/script/repositories/user/AppLockRepository.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/20743/commits/83815c17f9714d54cbddd36dbe90f061ba47182b

screendriver (approved):

PR #20842, opened 2026-03-24
feat(WPB-23950): handle account deep links in-app

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23950" title="WPB-23950" target="_blank">TaskWPB-23950</a>  [Web] Improve Deep Link Handling to Avoid Forced Logout
  </td></table>
  <br />
 

## Summary

- add in-app handling for supported Wire account deep links clicked in conversation messages
- makes profile and conversation-join links reuse the current authenticated session instead of opening in a new browser tab

---

## Accessibility (required)

---

### DIFF
--- a/apps/webapp/src/script/components/Conversation/Conversation.tsx
+++ b/apps/webapp/src/script/components/Conversation/Conversation.tsx
@@ -299,6 +299,23 @@ export const Conversation = ({
     return false;
   };
 
+  const openUserProfile = async (id: string, domain?: string) => {
+    try {
+      const userEntity = await repositories.user.getUserById({
+        id,
+        domain: domain ?? '',
+      });
+
+      showUserModal(userEntity);
+    } catch (error: unknown) {
+      if (error instanceof UserError && error.type === UserError.TYPE.USER_NOT_FOUND) {
+        messageListLogger.warn('Could not resolve user profile deep link', {id, domain});
+        return;
+      }
+      throw error;
+    }
+  };
+
   const handleMarkdownLinkClick = (event: MouseEvent | KeyboardEvent, messageDetails: MessageDetails) => {
     const href = messageDetails.href!;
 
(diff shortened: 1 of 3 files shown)


zhannach (dismissed):

apps/webapp/src/script/components/Conversation/Conversation.tsx
```diff
@@ -298,6 +301,44 @@ export const Conversation = ({
 
   const handleMarkdownLinkClick = (event: MouseEvent | KeyboardEvent, messageDetails: MessageDetails) => {
     const href = messageDetails.href!;
+
+    const parsed = parseAccountDeepLink(href, CONFIG.URL.ACCOUNT_BASE);
+
+    if (parsed?.type === 'user-profile') {
+      event.preventDefault();
+
+      void (async () => {
```
zskhan: I believe you added IIFE to use the async method?
As it is a fire and forgot thingy then then just take it out in another method and we can get rid of IIFE.

apps/webapp/src/script/components/Conversation/Conversation.tsx
```diff
@@ -298,6 +301,44 @@ export const Conversation = ({
...
+        try {
+          const userEntity = await repositories.user.getUserById({
+            id: parsed.id,
+            domain: parsed.domain ?? '',
+          });
+
+          showUserModal(userEntity);
+        } catch (error: unknown) {
+          if (error instanceof UserError && error.type !== UserError.TYPE.USER_NOT_FOUND) {
```
zskhan: if the error is `UserError.TYPE.USER_NOT_FOUND`, it will be silently swallowed.

apps/webapp/src/script/components/Conversation/Conversation.tsx
JacquelineLehner: got rid of it 👍

apps/webapp/src/script/components/Conversation/Conversation.tsx
JacquelineLehner: true, thanks for pointing it out. Changed it 👍

zskhan (approved):

PR #20846, opened 2026-03-25
fix: handle e2ei certificate update [WPB-24220]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24220" title="WPB-24220" target="_blank">BugWPB-24220</a>  [Web] Certificate not updated when using update function with valid certificate
  </td></table>
  <br />
 

## Summary

Fix the handling of e2ei certificate update.

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/E2EIdentity/E2EIdentityEnrollment.ts
+++ b/apps/webapp/src/script/E2EIdentity/E2EIdentityEnrollment.ts
@@ -27,7 +27,6 @@ import {TypedEventEmitter} from '@wireapp/commons';
 import {WebAppEvents} from '@wireapp/webapp-events';
 
 import {PrimaryModal, removeCurrentModal} from 'Components/Modals/PrimaryModal';
-import {ConversationState} from 'Repositories/conversation/ConversationState';
 import {UserState} from 'Repositories/user/UserState';
 import {Core} from 'src/script/service/CoreSingleton';
 import {getLogger} from 'Util/Logger';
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/E2EIdentity/E2EIdentityEnrollment.ts
```diff
@@ -313,12 +312,13 @@ export class E2EIHandler extends TypedEventEmitter<Events> {
...
-        getAllConversations: () => {
-          const conversationState = container.resolve(ConversationState);
-          const conversations = conversationState.conversations().map(conversation => ({
-            group_id: conversation.groupId ?? '',
-          }));
-          return Promise.resolve(conversations);
+        getAllConversations: async () => {
+          const conversations = await this.core.service.conversation.getConversations();
+          return conversations.found
+            .filter(conversation => !!conversation.group_id)
```
screendriver:
```suggestion
            .filter(conversation => {
              return is.nonEmptyString(conversation.group_id);
            })
```

arjita-mitra: I think its important to increase test coverage for this use case. We already have E2EIdentityEnrollment.test.ts could you add there?

@e-maad

arjita-mitra (changes requested):

apps/webapp/src/script/E2EIdentity/E2EIdentityEnrollment.ts
e-maad: Basically, not much changed on the coverage side. We just consumed conversations from BE instead of local state which will always be empty during app startup

zskhan (approved):
LGTM 🚀

arjita-mitra (approved):

PR #20866, opened 2026-03-26
chore: optimize critical flow tests runner [WPB-17563]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-17563" title="WPB-17563" target="_blank">TaskWPB-17563</a>  [Web] - Add critical flow tests to precommit pipeline
  </td></table>
  <br />
 

## Summary

Optimize critical flow tests runner.

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -54,10 +54,6 @@ jobs:
     name: Prepare Dependencies (One Time Setup)
     runs-on: ubuntu-latest
     
-    # Use Playwright's official container - browsers already installed!
-    container:
-      image: mcr.microsoft.com/playwright:v1.49.1-noble
-    
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -50,9 +50,65 @@ on:
...
+        run: |
+          echo "=== Checking node_modules ==="
+          ls -la node_modules/ | head -5
+          echo "=== Checking Playwright browsers ==="
+          ls -la ~/.cache/ms-playwright/ | head -5
+          echo "=== Checking 1Password CLI ==="
+          ls -la $HOME/.local/bin/op
+          echo "=== All paths verified ==="
+
+      - name: Cache dependencies for test shards
```
screendriver, line 89: I'm not convinced the dependency caching is a good tradeoff here. The 64 Playwright shards do make repeated setup expensive, so I agree with the goal of avoiding 64 separate dependency installs. My concern is with the mechanism, not the optimization target.

Because the cache key includes `github.run_id`, this is not really a reusable cache across workflow runs. We save a large dependency tree once, restore it many times within the same run, and then delete it again. That makes `actions/cache` behave more like an intra-run artifact transport than a true cache.

That adds complexity and fragility around `node_modules`, Playwright binaries, and the 1Password CLI. The follow-up fixes below in this branch makes that brittleness already showing up.

So I'm not arguing against reducing repeated setup for 64 shards. I'm arguing that this implementation may be the wrong tradeoff.

e-maad: I am closing it then.

PR #20893, opened 2026-03-30
fix: ui overflow issues in Create Group modal [WPB-22420]

theham3d:
## Summary
Hi team 👋

I recently came across an open position at Wire, and before applying, I wanted to get a better feel for the product and the codebase. I usually like to explore the app a bit and check out the website, so I did the same here.

While playing around with the app, I noticed two small UI issues in Create Group modal. The first one is in the header, it slightly overflows the modal body and causes a small visual glitch in the corner. The second one happens when we're showing list of users: the modal becomes scrollable, but the current styling creates some inconsistencies, with overlapping overflow areas and slightly awkward scrolling.

**What I changed:**

- Removed the `wrapperCSS` property from the modal component (which sets overflow: unset)
- Adjusted the modal body styling to improve scrolling behavior when the user list grows

From what I could tell and checked the history, `wrapperCSS` was originally introduced for `FadingScrollbar`, but in this case it doesn’t seem to add value, and removing it didn’t introduce any issues.

**Result:**

- More consistent and predictable scrolling behavior
- Cleaner UI in the modal when the list grows
- Fixed header overflow glitch

I’ve attached before/after screenshots for reference in below.

---

## Accessibility (required)

---

## Screenshots or demo (if the user interface changed)

**Before:**

scroll-issue
dark-o
light-o

**After:**

fix

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/style/modal/group-creation.less
```diff
@@ -64,4 +69,4 @@
     overflow-x: hidden;
     overflow-y: auto;
   }
-}
+}
```
thisisamir98: @theham3d could you fix the EOL issue here please?

theham3d: @thisisamir98 Sure

apps/webapp/src/style/modal/group-creation.less
```diff
@@ -43,6 +44,10 @@
       line-height: var(--line-height-lg);
     }
 
+    .modal__body {
```
arjita-mitra, line 47: removing wrapperCSS={{overflow: 'unset', overflowY: 'unset'}} is necessary because this was overriding overflow: hidden of the wrapper.

adding modal_body{ overflow: hidden} could be dropped because scrolling is handled inside .group-creation &__list which already has

```
height: 374px;
overflow-y: auto;
```

could you explain why we need `.modal__body {overflow-y: hidden;} `too?

theham3d: Thanks @arjita-mitra for the review 👍 
Yeah, you’re right. From what I can see in the history, this was mainly introduced because of the `FadingScrollbar` component. It looks like that changed over time, and when you fixed the issue, this override was left behind.

As for your question, the main reason is that you’re already handling the height limit and overflow in `.group-creation__list`, like you mentioned. With the current setup, the active style on `.modal__body` ends up causing a double scroll. In reality, there’s only one scrollable area (the list), but the UI shows two scrollbars, which is just a visual glitch (I've atached a screenshot above). This override fixes that issue.

arjita-mitra: I couldn't reproduce the double scrollbar though. Please check the pipeline issues.

zskhan: Hey @theham3d 
I am also unable to reproduce the double scrollbar.
Can you provide some steps, browser details, resolution?
The other part of this PR is good to go which is artifacts at the top corners of modal.

theham3d: @arjita-mitra @zskhan Thanks for the review.
I’ve recorded a video on the live production app to show how this can be reproduced. Since I didn’t have enough contacts to make the list scrollable, I duplicated them manually to demonstrate it.

Just a note about the video, the file was over 10MB, so I uploaded it to another service. But will expire by the end of today due to platform limits.

If you want to reproduce it yourself, you can try it with a scrollable contact list in the group creation flow, you’ll see the double scroll issue. If not, you can also add contacts manually like I did.

After applying the fix, the double scroll should be gone. I’m using Chrome version 146.0.7680.165.

Here is the video link: https://vimeo.com/reviews/bd7d8d76-26ac-4063-a6e7-35baa9eb8eaf/videos/1181857734

theham3d: For the failed pipeline, aside from the Jira ticket number check, I think the other issue might be because the PR is coming from a fork. It probably needs someone with elevated permissions to run it.

theham3d: _Quick update as of April 13:_ I just checked the link again. Based on the platform message I thought the preview would expire by the end of Friday, but it’s still working on my side in incognito tab.

Let me know if you have any trouble viewing the video.

zskhan: Hello.
I saw the video and now I am able to reproduce it.
About the PR CI issues I will also take care of it.
Thanks.

theham3d: @zskhan Thanks 💯

zskhan (approved):

PR #20898, opened 2026-03-30
chore(Playwright): Optimize the size of playwright reports [WPB-24280]

iskvortsov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24280" title="WPB-24280" target="_blank">TaskWPB-24280</a>  [Web][QA] Optimize the size of Playwright reports
  </td></table>
  <br />
 

## Summary

https://wearezeta.atlassian.net/browse/WPB-24280

- Added removing of blob reports after they are merged.
- Added removing trace .zip archives for every failure
- Modified trace settings to exclude snapshots from it.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs: Removing snapshots in traces
- Follow-ups (linked issues):

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -26,6 +26,10 @@ on:
       testinyRunName:
         type: string
         description: Define Testiny Run name to upload results to
+      keep_traces:
+        description: 'Attach traces to HTML report'
+        type: boolean
+        default: false
   # Options for calling the workflow within an other workflow
   workflow_call:
     inputs:
@@ -38,6 +42,10 @@ on:
       grep:
         type: string
         description: Define which tests to run, the given value will be passed to the `--grep` flag of playwright
+      keep_traces:
+        type: boolean
+        description: 'Keep Playwright traces'
+        default: false
     secrets:
       OP_SERVICE_ACCOUNT_TOKEN:
         required: true
@@ -141,6 +149,17 @@ jobs:
           else
             echo "Trace data directory not found"
           fi  
+
+      - name: Delete trace artifacts
+        if: ${{ !inputs.keep_traces }}
+        working-directory: apps/webapp
+        run: |
+          if [ -d "playwright-report/html/data" ]; then
+            DELETED_COUNT=$(find playwright-report/html/data -name "*.zip" -type f -delete -print)
+            echo "Deleted $DELETED_COUNT trace archive files"
+          else
+            echo "Trace data directory not found, skipping trace cleanup"
+          fi
 
       - name: Delete blob report artifacts
         if: always()
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -136,6 +136,12 @@ jobs:
           yarn playwright merge-reports --config ./playwright.config.ts playwright-report/blob-reports
           rm -rf playwright-report/blob-reports
 
+      - name: Delete blob report artifacts
+        if: always()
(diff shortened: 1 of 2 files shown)


apps/webapp/playwright.config.ts
```diff
@@ -17,11 +17,11 @@
  *
  */
 
-import {defineConfig, devices} from '@playwright/test';
-import {config} from 'dotenv';
-import {resolve} from 'node:path';
+import { defineConfig, devices } from '@playwright/test';
```
zskhan: Please remove these extra whitespaces.

apps/webapp/playwright.config.ts
```diff
@@ -43,14 +43,19 @@ module.exports = defineConfig({
...
+    ['json', { outputFile: 'playwright-report/report.json' }],
     ['line'],
   ],
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'on-first-retry',
+    trace: {
+      mode: 'on-first-retry',
+      snapshots: false,
```
zskhan, line 55: `snapshots: false` saves space, but it also removes traces. That makes retry traces meaningfully less useful for debugging flaky UI / API failures. Was that trade-off intentional?

iskvortsov: Yep, are we using snapshots in traces? It's basically taking a screenshot every few milliseconds and it can take a lot of space. We will however have an error message and screenshot of the failure.

apps/webapp/playwright.config.ts
iskvortsov: Oh my bad, I used auto formatting in VS code. Let me revert that.

iskvortsov:
Check this resulting playwright report here and let me know if you would find it useful: 

https://github.com/wireapp/wire-webapp/actions/runs/23745976831

.github/workflows/e2e-tests.yml
```diff
@@ -129,12 +129,19 @@ jobs:
...
 
-      # Merge reports and delete blob reports, only keep resources to include within the artifact
+      # Merge reports and delete blob reports and zip archives for traces, only keep resources to include within the artifact
       - name: Merge playwright reports
         working-directory: apps/webapp
         run: |
           yarn playwright merge-reports --config ./playwright.config.ts playwright-report/blob-reports
-          rm -rf playwright-report/blob-reports/report*
+          rm -rf playwright-report/blob-reports
+          rm -f playwright-report/html/data/*.zip
```
arjita-mitra: rm -f playwright-report/html/data/*.zip

1. There's some risk here. This assumes playwright-report/html/data/ structure exists exactly as is which is not full-proof since Playwright could reorganize where traces are stored.

2. rm -f uses the -f flag, meaning it won't error if the path doesn't exist or if no .zip files are found

More robust solution would be to use error handling here. Something like this, please verify the exact command.

```
 if [ -d "playwright-report/html/data" ]; then
 DELETED_COUNT=$(find playwright-report/html/data -name "*.zip" -type f -delete -print)
      echo "Cleaned up $DELETED_COUNT trace archive files"
    else
      echo "Trace data directory not found"
```

iskvortsov: I can add this. Thanks

iskvortsov:
I've implemented a checkbox to keep Playwright traces in case we need it.

thisisamir98 (approved):

arjita-mitra (approved):

PR #20921, opened 2026-04-02
fix: Documents in Recycle Bin should not be editable(WPB-23547)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23547" title="WPB-23547" target="_blank">BugWPB-23547</a>  [Web] Documents in Recycle Bin should not be editable
  </td></table>
  <br />
 

## Summary

Documents that have been moved to the Recycle Bin remain editable when on web. Files in the Recycle Bin should be in a read-only state and not allow further modifications.

**Steps to Reproduce**
Go to a Wire Drive enabled conversation

Upload an editable document (e.g., .docx)

Move the document to the Recycle Bin

Open the document from the Recycle Bin

Attempt to edit the document

**Expected Result**
Documents in the Recycle Bin should:

Be read-only / Be on preview-only

No modifications should be possible

Editing capabilities (typing, formatting, saving changes) should be disabled

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
+++ b/apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
@@ -103,7 +103,6 @@ describe('FileFullscreenModal - File Version Restore', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
-    window.location.hash = '#/conversation/test/files/';
   });
 
   describe('Modal Rendering', () => {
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
```diff
@@ -103,6 +103,7 @@ describe('FileFullscreenModal - File Version Restore', () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
+    window.location.hash = '#/conversation/test/files/';
```
screendriver: `window.location.hash = '#/conversation/test/files/';` introduces a global side effect in test setup. There is no real `window` in plain Node.js, so this test depends on a browser-like test environment being present, and it mutates shared global state without restoring it. If this setup is required, the previous location state should be captured and reset in `afterEach`; otherwise the test can leak state into later tests.

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
```diff
@@ -65,7 +66,8 @@ export const FileFullscreenModal = ({
   badges,
   isEditMode,
 }: FileFullscreenModalProps) => {
-  const [isEditableState, setIsEditableState] = useState(isEditMode);
+  const notInRecycleBin = !isInRecycleBin();
+  const [isEditableState, setIsEditableState] = useState(!!isEditMode && notInRecycleBin);
```
screendriver:
```suggestion
  const [isEditableState, setIsEditableState] = useState(isEditMode && notInRecycleBin);
```

Why double negotiation of a boolean? `!!true` becomes still `true`

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
```diff
@@ -79,8 +81,8 @@ export const FileFullscreenModal = ({
   };
 
   useEffect(() => {
-    setIsEditableState(!!isEditMode);
-  }, [isEditMode]);
+    setIsEditableState(!!isEditMode && notInRecycleBin);
```
screendriver:
```suggestion
    setIsEditableState(isEditMode && notInRecycleBin);
```

The same as in https://github.com/wireapp/wire-webapp/pull/20921/changes#r3049895742

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
arjita-mitra: Removed the use of window.location.hash. Component now accept a isInRecycleBin prop (dependency injection) then tests pass isInRecycleBin={() => true} or () => false directly no global state.

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.tsx
arjita-mitra: isEditMode is typed as isEditMode?: boolean so it's boolean | undefined. I have used explicit default(false) in destructuring then just useState(isEditMode && notInRecycleBin) is enough.

screendriver: Then it would be even better to write `isEditMode === true` 🙂

screendriver (approved):

PR #20923, opened 2026-04-02
fix: call e2eRotate on existing conversations [WPB-24528]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24528" title="WPB-24528" target="_blank">BugWPB-24528</a>  [Web] User not able to send messages after e2ei was enbaled while user is logged in
  </td></table>
  <br />
 

## Summary

When we enable e2eI for existing MLS client, all the conversations must me called with `e2eiRotate`.

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/libraries/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
+++ b/libraries/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
@@ -276,8 +276,19 @@ export class E2EIServiceInternal {
       const newCrlDistributionPoints = await cx.saveX509Credential(identity, certificate);
       for (const conversation of conversations) {
         if (Boolean(conversation.group_id?.length)) {
-          const idAsBytes = Decoder.fromBase64(conversation.group_id).asBytes;
-          await cx.e2eiRotate(new ConversationId(idAsBytes));
+          try {
+            const idAsBytes = Decoder.fromBase64(conversation.group_id).asBytes;
+            const conversationId = new ConversationId(idAsBytes);
+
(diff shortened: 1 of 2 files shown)


libraries/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
```diff
@@ -276,8 +276,19 @@ export class E2EIServiceInternal {
...
         if (Boolean(conversation.group_id?.length)) {
-          const idAsBytes = Decoder.fromBase64(conversation.group_id).asBytes;
-          await cx.e2eiRotate(new ConversationId(idAsBytes));
+          try {
+            const idAsBytes = Decoder.fromBase64(conversation.group_id).asBytes;
+            const conversationId = new ConversationId(idAsBytes);
+
+            // Check if conversation exists before rotating
+            const conversationExists = await cx.conversationExists(conversationId);
+            if (conversationExists) {
```
zskhan, line 285: Can we add a log for else condition?
Otherwise it will be silently swallowed

zskhan (approved):

libraries/core/src/messagingProtocols/mls/E2EIdentityService/E2EIServiceInternal.ts
e-maad: We must only rotate conversation which exits in CC. So, we swallow else cases without any issue.

PR #20937, opened 2026-04-07
ci: parallelize precommit flows with slot-locked EB concurrency [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

The precommit critical-flow pipeline was bottlenecked by a single shared Elastic Beanstalk environment and a single global concurrency lane. That meant PRs queued behind each other even when they were unrelated, and a new push to one PR could not efficiently replace its own stale run. Separately, the previous runner migration left a risk of long-running or stuck jobs, so explicit job timeouts were needed as a safety net.

This change moves the workflow to a fixed pool of pre-created EB environments and updates the concurrency model so we can run multiple PRs in parallel while preventing overlapping deploy-and-test execution on the same target environment.

Environment mapping:
- Dependabot-only PRs use the dedicated main precommit environment: https://wire-webapp-precommit.zinfra.io/
- Regular PRs are distributed across the fixed slot pool:
  - https://wire-webapp-precommit-1.zinfra.io/
  - https://wire-webapp-precommit-2.zinfra.io/
  - https://wire-webapp-precommit-3.zinfra.io/

What changed:
- Replaced the old global workflow concurrency group with a per-PR concurrency key: `precommit-pr-${{ github.event.pull_request.number || github.run_id }}` with `cancel-in-progress: true`. This means a new push only cancels stale runs for the same PR, not runs from other PRs.
- Added environment selection logic in the build job and exposed it as `env_name`.
- Routed regular PRs across three pre-created environments: `wire-webapp-precommit-1`, `wire-webapp-precommit-2`, `wire-webapp-precommit-3`.
- Reserved `wire-webapp-precommit` for Dependabot PRs only, so automated dependency updates stay isolated from the main slot pool.
- Added a new reusable workflow, `precommit-slot-run.yml`, which wraps the deploy and e2e path under a slot-scoped workflow-level concurrency lock: `precommit-slot-${{ inputs.env_name }}`. This ensures the lock is held for the whole deploy-plus-test critical section instead of only for the deploy job.
- Replaced the old standalone `deploy_to_aws` plus direct `e2e_tests` flow in `precommit-crit-flows.yml` with a single `deploy_and_run_e2e` reusable workflow call.
- Updated the report jobs to consume artifact outputs from the reusable workflow instead of the old direct e2e job.
- Added missing `timeout-minutes` values to the jobs that previously had no explicit limit:
  - `check_edits_to_e2e_tests`: 5
  - `run_e2e_tests`: 5
  - `e2e-report`: 10
- Pinned the previously unpinned checkout action in `check_edits_to_e2e_tests` to the same commit SHA already used elsewhere in the workflow for supply chain consistency.

New concurrency model:
- Layer 1: PR-level cancellation in `precommit-crit-flows.yml` prevents stale runs for the same PR from piling up.
- Layer 2: slot-level locking in `precommit-slot-run.yml` serializes access to each EB environment across the full deploy and e2e lifecycle.
- Result:
  - different PRs can run in parallel if they map to different slots
  - the same PR supersedes its own older runs
  - Dependabot PRs are isolated onto https://wire-webapp-precommit.zinfra.io/
  - no second PR can redeploy over an environment while another PR is still executing e2e against that same slot

This gives us controlled parallelism, preserves isolation for Dependabot, adds timeout protection for stuck jobs, and removes the previous deploy-versus-test overlap on shared precommit environments.


.github/workflows/precommit-slot-run.yml
```diff
@@ -0,0 +1,99 @@
...
+          wait-for-deployment: true
+          wait-for-environment-recovery: true
+          use-existing-application-version-if-available: true
+
+      - name: Set precommit URL
+        id: fetch_url
+        env:
+          ENV_NAME: ${{ inputs.env_name }}
+        run: |
+          echo "url=https://$ENV_NAME.zinfra.io/" >> "$GITHUB_OUTPUT"
```
arjita-mitra: .zinfra.io/ suffix is hardcoded. you could do one of the following

Adding a configurable domain suffix as a workflow input
Fetching the actual CNAME from AWS (as the old code did)

.github/workflows/precommit-crit-flows.yml
```diff
@@ -53,6 +54,19 @@ jobs:
       - run: yarn nx run webapp:configure
       - run: yarn nx run server:package
 
+      - name: Compute target EB environment
+        id: slot
+        run: |
+          if [[ "${{ github.actor }}" == "dependabot[bot]" ]]; then
+            # Dependabot gets the original dedicated environment
+            echo "env_name=wire-webapp-precommit" >> "$GITHUB_OUTPUT"
+          else
```
arjita-mitra, line 63: Could you add telemetry or logging to verify:

That different PRs actually run in parallel on different slots
That slot locks are preventing concurrent deployments as intended

.github/workflows/precommit-crit-flows.yml
```diff
@@ -53,6 +54,19 @@ jobs:
...
+            # Spread other PRs across 3 pre-created EB environments; merge queue uses run_id
+            ID="${{ github.event.pull_request.number || github.run_id }}"
+            SLOT=$(( (ID % 3) + 1 ))
+            echo "env_name=wire-webapp-precommit-$SLOT" >> "$GITHUB_OUTPUT"
+          fi
+
```
arjita-mitra, line 69: Also could you add some monitoring to ensure even distribution of slot assignment.

AI output so please verify

```
- name: Log slot assignment
  run: |
    mkdir -p monitoring
    echo "pr_number=${{ github.event.pull_request.number || github.run_id }}" >> monitoring/slot-assignment.txt
    echo "slot=${{ steps.slot.outputs.env_name }}" >> monitoring/slot-assignment.txt
    echo "timestamp=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> monitoring/slot-assignment.txt

- name: Upload slot metrics
  uses: actions/upload-artifact@v4
  with:
    name: slot-assignment-${{ github.run_id }}
    path: monitoring/slot-assignment.txt
    retention-days: 30
```

zskhan (dismissed):

.github/workflows/precommit-slot-run.yml
thisisamir98: I think this might be extra unnecessary work as we are using zinfra.io since many many years.

.github/workflows/precommit-crit-flows.yml
thisisamir98: https://github.com/wireapp/wire-webapp/pull/20937/commits/fa9b3231c5ed5ca308ba5c3811e67fa1261f23da

.github/workflows/precommit-crit-flows.yml
thisisamir98: https://github.com/wireapp/wire-webapp/pull/20937/commits/fa9b3231c5ed5ca308ba5c3811e67fa1261f23da

.github/workflows/precommit-slot-run.yml
arjita-mitra: it was there before unless this command won't work anymore don't you think its better to fetch it?

```
environment-names wire-webapp-precommit \
            --query "Environments[0].CNAME" --output text)
```

arjita-mitra (dismissed):

zskhan (dismissed):

zskhan (approved):

PR #20501, opened 2026-02-24
feat: toggle apps in conversation creation [WPB-23669]

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23669" title="WPB-23669" target="_blank">StoryWPB-23669</a>  [Web][Integrations] Toggle app slider at conversation creation
  </td></table>
  <br />
 

# Pull Request

## Summary

- Toggles the availability of apps when creating channels or groups based on the default protocol and feature flags

---

## Security Checklist (required)

- [ X ] **External inputs are validated & sanitized** on client and/or server where applicable.
- [ X ] **API responses are validated**; unexpected shapes are handled safely (fallbacks or errors).
- [ X ] **No unsafe HTML is rendered**; if unavoidable, sanitization is applied **and** documented where it happens.
- [ X ] **Injection risks (XSS/SQL/command) are prevented** via safe APIs and/or escaping.

## Accessibility (required)

- [ X ] I have read and this PR **upholds** our [Accessibility Best Practices](https://github.com/wireapp/wire-webapp/blob/dev/docs/accessibility-practices.md).

## Standards Acknowledgement (required)

- [ X ] I have read and this PR **upholds** our [Coding Standards](https://github.com/wireapp/wire-webapp/blob/dev/docs/coding-standards.md) and [Tech Radar Choices](https://github.com/wireapp/wire-webapp/blob/dev/docs/tech-radar.md).

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):
- Linked PRs (e.g. web-packages):

PR #20941, opened 2026-04-08
feat: toggle apps in conversation creation [WPB-23669]

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23669" title="WPB-23669" target="_blank">StoryWPB-23669</a>  [Web][Integrations] Toggle app slider at conversation creation
  </td></table>
  <br />
 

## Summary

- Toggle apps slider and apps disabled banner based on protocol and feature flags

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/script/components/Modals/CreateConversation/CreateConversationSteps/Preference.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+            allowedCipherSuites: [1],
+            protocolToggleUsers: ['protocolToggleUsers']
+          },
+          status: FEATURE_STATUS.ENABLED
+        },
+      }),
+    };
+
+    container.clearInstances();
+    container.registerInstance(TeamState, mockTeamState as any);
```
markbrockhoff:
```suggestion
    container.registerInstance(TeamState, mockTeamState);
```

apps/webapp/src/script/components/Modals/CreateConversation/CreateConversationSteps/Preference.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+            defaultCipherSuite: 1,
+            allowedCipherSuites: [1],
+            protocolToggleUsers: ['protocolToggleUsers']
+          },
+          status: FEATURE_STATUS.ENABLED
+        },
+      }),
+    };
+
+    container.clearInstances();
```
markbrockhoff: 🔧 Maybe put the clearing into a beforeEach hook. Just to make sure we don't run into issues with test isolation e.g. if someone adds an other test to the describe block.

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+              allowedCipherSuites: [1],
+              protocolToggleUsers: ['protocolToggleUsers']
+            },
+            status: FEATURE_STATUS.ENABLED
+          },
+        }),
+      };
+
+      container.clearInstances();
+      container.registerInstance(UserState, mockUserState as any);
```
markbrockhoff:
```suggestion
      container.registerInstance(UserState, mockUserState);
```

panuschek: I removed the entire line as it was unnessecary

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+              search: {} as SearchRepository,
+              team: {} as TeamRepository,
+            }
+          }
+        }
+      };
+
+      // Act
+      const {getByTestId} = mountComponent(
+        <RootContext.Provider value={mockRootContext as any}>
```
markbrockhoff: When doing a type assertion e.g. `as Something` at least keep it as narrow as possible. Any disables all type checking, meanwhile asserting the needed type keeps at least a bit of typescripts ability to keep us from shooting our own foot :D

E.g.:

```suggestion
      const mockRootContext = {
        mainViewModel: {
          content: {
            repositories: {
              conversation: {},
              search: {},
              team: {},
            }
          }
        }
      } as RootContextValue;

      // Act
      const {getByTestId} = mountComponent(
        <RootContext.Provider value={mockRootContext}>
```

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+      const mockUser = new User('user-id', 'test-domain.wire.com');
+
+      const mockUserState: UserState = {
+        self: ko.observable(mockUser),
+        users: ko.observableArray<User>([]),
+        connectedUsers: ko.pureComputed(() => [] as User[]),
+        connectRequests: ko.pureComputed(() => [] as User[]),
+      };
+
+      const mockTeamState: Partial<TeamState> = {
```
markbrockhoff: FYI: I noticed that there's a type for a "deep partial", it's just called `RecursivePartial` in this project. You could use it to type mocks so you no longer have to pass properties which are irrelevant for your test.


```suggestion
      const mockTeamState: TypeUtil.RecursivePartial<TeamState> = {
```

apps/webapp/src/script/components/toggle/InfoToggle.tsx
```diff
@@ -64,19 +62,20 @@ const InfoToggle = ({
...
             onClick={() => setIsChecked(!isChecked)}
-            data-uie-name={dataUieNameLabelText}
             data-uie-value={isChecked ? 'checked' : 'unchecked'}
           >
             <span className="button-label__switch" />
             <span className="visually-hidden">{name}</span>
           </button>
         </div>
       </div>
+      {label}
```
markbrockhoff: Why is there a prop "name" which is rendered as label but also a label which is rendered at the end of it as plain text? Could this be a duplicate prop or be better named maybe?

apps/webapp/src/script/util/featureUtil.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    {
+      featureOptions: {
+        protocol: CONVERSATION_PROTOCOL.MLS,
+        isAppsEnabled: false,
+        hasWhitelistedServices: true,
+      },
+      expected: false
+    },
+  ])('apps feature is $expected when { protocol: $protocol, isAppsEnabled: $isAppsEnabled, hasWhitelistedServices: $hasWhitelistedServices }',
```
markbrockhoff, line 51: Are you sure this kind of interpolation within the test name works? E.g. $protocol is a nested property of "featureOptions", I don't think it will be interpolated like this. Maybe you could just not group the options within a "featureOptions" property, e.g.:

```ts
type AppsFeatureDataSet = AppsFeatureOptions & { expected: boolean };
```

apps/webapp/src/script/components/toggle/InfoToggle.tsx
panuschek: The label prop is used to append the "Apps aren't available" banner to the component.
That seems to be the only way to do it since the border around the element is drawn from within the component itself.
image

markbrockhoff: But could you then please name the prop differently? Maybe "children" or something since IMHO "label" implies the wrong thing

markbrockhoff (dismissed):

apps/webapp/src/script/components/Modals/CreateConversation/CreateConversationSteps/Preference.tsx
```diff
@@ -70,17 +83,17 @@ export const Preference = () => {
...
-        />
-      )}
+      <InfoToggle
+        className="modal-style"
+        dataUieName="info-toggle-services"
+        info={t('servicesRoomToggleInfoExtended')}
+        setIsChecked={setIsServicesEnabled}
+        isDisabled={!isAppsFeatureAvailable}
+        name={t('servicesOptionsTitle')}
+        isChecked={isServicesEnabled && isAppsFeatureAvailable}
```
zskhan, line 93: we can skip `isAppsFeatureAvailable` here as you are already checking that above in `isDisabled`

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -79,11 +81,13 @@ const GroupCreationModal = ({
     isMLSEnabled: isMLSEnabledForTeam,
     isProtocolToggleEnabledForUser,
     isCellsEnabled: isCellsEnabledForTeam,
+    isAppsEnabled: isAppsEnabledForTeam,
```
zskhan, line 84: hasWhitelistedServices: hasWhitelistedServicesForTeam,

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -79,11 +81,13 @@ const GroupCreationModal = ({
...
   } = useKoSubscribableChildren(teamState, [
     'isTeam',
     'isMLSEnabled',
     'isProtocolToggleEnabledForUser',
     'isCellsEnabled',
+    'isAppsEnabled',
```
zskhan, line 91: 'hasWhitelistedServices',

apps/webapp/src/script/components/Modals/GroupCreation/GroupCreationModal.tsx
```diff
@@ -154,7 +158,16 @@ const GroupCreationModal = ({
...
   const isGuestRoom = accessState === ACCESS_STATE.TEAM.GUEST_ROOM;
   const isGuestEnabled = isGuestRoom || isGuestAndServicesRoom;
-  const isServicesEnabled = isServicesRoom || isGuestAndServicesRoom;
+
+  const isAppsFeatureAvailable =
+    isTeam &&
+    checkAppsFeatureAvailability({
+      protocol: selectedProtocol.value,
+      isAppsEnabled: isAppsEnabledForTeam,
+      hasWhitelistedServices: teamState?.hasWhitelistedServices(),
```
zskhan: 'hasWhitelistedServices
```suggestion
      hasWhitelistedServices,
```

apps/webapp/src/script/components/toggle/InfoToggle.tsx
```diff
@@ -17,7 +17,7 @@
  *
  */
 
-import {useId} from 'react';
+import React, {useId} from 'react';
```
zskhan:
```suggestion
import {ReactNode, useId} from 'react';
```

apps/webapp/src/script/components/toggle/InfoToggle.tsx
```diff
@@ -29,6 +29,7 @@ interface InfoToggleProps {
   name: string;
   className?: string;
   setIsChecked: (checked: boolean) => void;
+  label?: React.ReactNode;
```
zskhan:
```suggestion
  label?: ReactNode;
```

apps/webapp/src/script/page/RightSidebar/guestServicesOptions/components/servicesOptions/servicesOptions.tsx
```diff
@@ -22,42 +22,60 @@ import {FC} from 'react';
...
+        <BaseToggle
+          isChecked={isServicesEnabled}
+          setIsChecked={toggleServicesAccessState}
+          isDisabled={isToggleDisabled}
+          infoText={t('servicesRoomToggleInfo')}
+          toggleName={t('servicesOptionsTitle')}
+          toggleId="services"
+        />
+      ) : (
+        <span>You team doesn&#39;t use apps yet.</span>
```
zskhan: This text should come from translations

zskhan:
Will [this PR](https://github.com/wireapp/wire-webapp/pull/20501) go away?

panuschek:
> Will [this PR](https://github.com/wireapp/wire-webapp/pull/20501) go away?

I closed it. 
I realized that the ticket I was working on was way too big for a single PR so I decided to split it into multiple which this is the first of.

apps/webapp/src/script/page/RightSidebar/guestServicesOptions/components/servicesOptions/servicesOptions.tsx
panuschek: This has been removed

apps/webapp/src/script/components/Modals/CreateConversation/CreateConversationSteps/Preference.tsx
panuschek: I don't see how this is redundant. "isServicesEnabled" is true by default but should be false if the app feature flag is false.
Removing the additional check would result in the checkbox being disabled but set to 'On'.

zskhan (approved):

PR #20953, opened 2026-04-09
fix: keep the version details always in a single line and introduce a hiding behind a button behaviour in case details are very long(WPB-24288)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24288" title="WPB-24288" target="_blank">BugWPB-24288</a>  [Web][versions] Reduce name and size to single line of text
  </td></table>
  <br />
 

## Summary

**Description**

Currently the version details (username and size) often appear in two lines (in order to make space for the buttons in hover state. 

**Acceptance criteria**

Instead, we want to keep the details always in a single line and introduce a hiding behind a button behavior in case details are very long. (two versions of design, depending on effort)

---

## Accessibility (required)

---

## Screenshots or demo (if the user interface changed)
Screenshot 2026-04-09 at 14 46 36

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
+++ b/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
@@ -123,8 +123,8 @@ export const fileVersionItemWrapperCss: CSSObject = {
   borderRadius: '8px',
   ':hover': {
     backgroundColor: 'var(--gray-20)',
-    button: {
-      visibility: 'visible',
+    '& [data-version-actions="true"]': {
+      display: 'flex',
     },
   },
 };
@@ -151,6 +151,7 @@ export const versionDotOldCss: CSSObject = {
 
 export const versionInfoContainerCss: CSSObject = {
   flex: 1,
+  minWidth: 0,
 };
 
 export const versionTimeTextCss: CSSObject = {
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
```diff
@@ -123,8 +123,8 @@ export const fileVersionItemWrapperCss: CSSObject = {
   borderRadius: '8px',
   ':hover': {
     backgroundColor: 'var(--gray-20)',
-    button: {
-      visibility: 'visible',
+    '& [data-version-actions="true"]': {
```
zskhan, line 126: Showing the button on hover is not good for touch devices and accessibility.
What is the plan to make it accessible?

zskhan (approved):

apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
arjita-mitra: @zskhan the entire section is not accessible as of now as they are non interactive elements. I agree buttons should be keyboard accessible as well. Since its not part of this ticket i will talk to my scope PM and address this in coming sprints.

PR #20954, opened 2026-04-09
test(WPB-24426): rewrite Wire Drive tests

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24426" title="WPB-24426" target="_blank">TaskWPB-24426</a>  [Web][QA] Rewrite Wire Drive tests to use the same test fixtures as Critical Flow tests to reduce flakiness
  </td></table>
  <br />
 

## Summary

- rewrite Wire Drive tests to use newer fail-proofed fixtures
- Test cases: TC-8786, TC-8787, TC-8788, TC-8785

---

## Accessibility (required)

---


apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -27,7 +27,7 @@ import {ConfirmModal} from '../modals/confirm.modal';
 type EmojiReaction = 'plus-one' | 'heart' | 'joy';
 
 export class ConversationPage {
-  private readonly page: Page;
+  protected readonly page: Page;
```
markbrockhoff: Without the CellsConversationPage extending this class the change from private to protected should no longer be necessary :)

apps/webapp/test/e2e_tests/utils/userActions.ts
```diff
@@ -64,8 +64,13 @@ export const sendTextMessageToConversation = async (
 };
 
 type UserPages = PageManager['webapp']['pages'];
-export const createGroup = async (pages: UserPages, conversationName: string, user: User[]) => {
+export const createGroup = async (pages: UserPages, conversationName: string, user: User[], cells?: boolean) => {
```
markbrockhoff: 🔧 This API isn't very expandable, what if we also need to enable e.g. read receipts when creating a group? We'd have to call a function like this `createGroup(pages, "Test Group", [userB], undefined, true);`.

Adding more parameters to a function becomes ugly quickly, instead try to group such optional settings as an object. This way the position of the parameter no longer matters and adding more is easily possible in the future, plus it's easier to read since `{cells: true}` has more semantic meaning than just `true`. E.g.:

```suggestion
export const createGroup = async (pages: UserPages, conversationName: string, user: User[], features?: {cells?: boolean}) => {
```

> For the `createGroup` function it could even be reasonable to also move the name and users into the object, but that would be too much change for this PR, so adding a new "features" object is enough for now ;)

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
```diff
@@ -491,4 +491,22 @@ export class ConversationPage {
...
+
+  public getVideoInMultipartMessageLocator(user: User): Locator {
+    return this.page.getByLabel(new RegExp(`^Image from ${user.fullName}`)).locator('video');
+  }
+
+  public getCellsImageLocator(user: User): Locator {
+    return this.page
+      .getByTestId('item-message')
+      .getByRole('button', {name: new RegExp(`^Image from ${user.fullName}`)});
+  }
```
markbrockhoff: Two of these helper functions are only used once and `getCellsImageLocator` is only used twice. Does it really make sense to have them as utils on the conversation POM? I could potentially see locators for images and videos as part of the response of "getMessage" but here I don't feel like there's much value in the abstraction, especially since they are very specific to the suite of drive tests.

Maybe it would also make sense to align the folder structure of the drive tests with the other test suites. E.g. move the tests all into one file called `drive.spec.ts`. This would also reduce code duplication slightly since e.g. the beforeEach could be reused. (If you merge them, please make sure to do it in one atomic commit so the PR stays easy to review)

markbrockhoff (changes requested):
So much deleted code, what a lovely refactoring 💪🏻

apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
```diff
@@ -40,66 +35,38 @@ const editedMessageText = 'Here is an image for you, friend';
 
 const imageFilePath = getImageFilePath();
 
+test.beforeEach(async ({createTeam, createUser}) => {
```
zhannach: Since this file contains only one test case, should we move the setup logic directly into the test block?

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
JacquelineLehner: changed it 👍

apps/webapp/test/e2e_tests/utils/userActions.ts
JacquelineLehner: Good point! Changed it 👍

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversation.page.ts
JacquelineLehner: refactored the files into one and bundled the helper function in the single file ✅

apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
JacquelineLehner: I bundled the separate files into one 👍

markbrockhoff (dismissed):

apps/webapp/test/e2e_tests/specs/Drive/drive.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    const team = await createTeam(teamName, {users: [userB], features: {cells: true}});
+    userA = team.owner;
+  });
+
+  test(
+    'Uploading an file in a group conversation',
+    {tag: ['@crit-flow-cells', '@regression', '@TC-8785']},
+    async ({createPage}, testInfo) => {
+      const [userAPageManager, userBPageManager] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))),
```
thisisamir98, line 71: Image

@JacquelineLehner could you check if this comment is relevant? looks like PageManager.from is not returning a promise?

JacquelineLehner: I checked it, and it's a known Sonar false positive. @markbrockhoff already raised this some time ago if I'm not mistaken.

`PageManager.from(...)` is overloaded: when it receives a `Promise<Page>`, it returns a `Promise<PageManager>`; when it receives a `Page`, it returns a `PageManager`. In this case we pass `createPage(...)`, so at runtime it does return a promise.

Sonar seems to lose that distinction because of the overloads and flags `Promise.all(...)` as if it received non-promises. So this is a static-analysis issue rather than an actual async bug.

zhannach (dismissed):

zskhan (dismissed):

zskhan (approved):

PR #20957, opened 2026-04-09
test: repair / improve wire drive tests after removal of testId [WPB-22420]

markbrockhoff:
<table>
<td>
  </td></table>
  <br />
 

## Summary

This is just an intermediate PR to resolve the currently failing drive tests with little changes. The full refactoring of those tests will be done in #20954

- Update locator for toggle to enable shared drive during group creation. The previously used testId has been removed
- Remove conditional logic from tests depending on which stage they run against
- Remove flaky non retryable util

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
@@ -67,9 +67,7 @@ test(
       const loginOwner = async () => {
         await userAPageManager.openMainPage();
         await loginUser(userA, userAPageManager);
-        if (process.env.ENV_NAME === 'staging') {
-          await userAModals.dataShareConsent().clickDecline();
-        }
+        await userAModals.dataShareConsent().clickDecline();
         await userAPages.conversationList().clickCreateGroup();
         // Files should be disabled by default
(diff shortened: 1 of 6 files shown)


apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
```diff
@@ -67,9 +67,7 @@ test(
       const loginOwner = async () => {
         await userAPageManager.openMainPage();
         await loginUser(userA, userAPageManager);
-        if (process.env.ENV_NAME === 'staging') {
-          await userAModals.dataShareConsent().clickDecline();
```
arjita-mitra, line 71: removing the env check because the feature is now available everywhere right?

arjita-mitra (approved):

apps/webapp/test/e2e_tests/specs/Drive/editMultipartMessage-TC-8786.spec.ts
markbrockhoff: Yes, at least we know from all other tests that this modal is showing up after login on all envs so this condition doesn't make sense to me (and caused the test to fail on dev & pre-commit)

PR #20973, opened 2026-04-13
feat: auto downgrade should stop to low instead of bypass because of privacy reasons [WPB-24687]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24687" title="WPB-24687" target="_blank">TaskWPB-24687</a>  [Web] Background effects should not be bypass due to performance
  </td></table>
  <br />
 

## Summary

- What did I change and why?
Auto downgrade should stop to low instead of bypass because of privacy reasons

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/quality/qualityController.test.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/quality/qualityController.test.ts
@@ -24,7 +24,8 @@ const MODE_BLUR = 'blur' as const;
 const MODE_VIRTUAL = 'virtual' as const;
 
 const TARGET_FPS = 30;
-const BUDGET_MS = 1000 / TARGET_FPS;
+const PERFORMANCE_THRESHOLD = 1000;
+const BUDGET_MS = PERFORMANCE_THRESHOLD / TARGET_FPS;
 const DOWNGRADE_THRESHOLD_MS = BUDGET_MS * DEFAULT_TUNING.downgradeThresholdRatio;
 const HYSTERESIS_FRAMES = DEFAULT_TUNING.hysteresisFrames;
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/repositories/media/backgroundEffects/quality/qualityController.test.ts
```diff
@@ -72,6 +72,12 @@ describe('QualityController', () => {
     }
 
     expect(params.tier).toBe('low');
+
+    guard = 0;
+    while (guard++ < 1000) {
```
arjita-mitra: lets make 1000 as const and reuse everywhere?

apps/webapp/src/script/repositories/media/backgroundEffects/quality/qualityController.ts
```diff
@@ -440,15 +440,12 @@ export class QualityController {
       // CPU-bound: normal step down to B
       return dominant === 'gpu' ? 'low' : 'medium';
     }
-    if (this.tier === 'medium') {
+    if (this.tier === 'medium' || this.tier === 'low') {
       return 'low';
```
arjita-mitra: i think we should use the typed variables here. you already have backgroundEffectsWorkerType we could define a const like and then use thoe values that will ensure single source of truth and type safety. Wdyt?

```
export const QUALITY_TIERS = {
  SUPERHIGH: 'superhigh' as const,
  HIGH: 'high' as const,
  MEDIUM: 'medium' as const,
  LOW: 'low' as const,
  BYPASS: 'bypass' as const,
  // Privacy boundary constant
  MIN_ALLOWED_TIER: 'low' as const,
} as const;
```

arjita-mitra (approved):

PR #20989, opened 2026-04-14
feat: added UserType and app fields to User entity (WPB-20361)

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20361" title="WPB-20361" target="_blank">TaskWPB-20361</a>  [Web] [Integration] Client should understand the User entity Type field
  </td></table>
  <br />
 

## Summary

- In preparation for implementing the search for apps, the user entity needs to know about the new UserType inidcating whether a UserEntity is a regular user or an app

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/repositories/user/UserMapper.ts
+++ b/apps/webapp/src/script/repositories/user/UserMapper.ts
@@ -17,6 +17,8 @@
  *
  */
 
+import {UserType} from '@wireapp/api-client/lib/user';
+
 import {MappedAsset, mapProfileAssets, updateUserEntityAssets} from 'Repositories/assets/assetMapper';
 import {User} from 'Repositories/entity/User';
 import {UserRecord} from 'Repositories/storage';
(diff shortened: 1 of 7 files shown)


apps/webapp/src/script/repositories/user/UserMapper.ts
```diff
@@ -181,6 +183,15 @@ export class UserMapper {
       userEntity.isDeleted = true;
     }
 
+    if (type) {
```
zskhan: In User.ts type is not optional so it should have a type in fallback
Maybe instead of 
```
if (type) {
    userEntity.type = type;
}
```
This 
`userEntity.type = type ?? UserType.REGULAR;`

apps/webapp/src/script/repositories/user/UserMapper.ts
```diff
@@ -181,6 +183,15 @@ export class UserMapper {
...
+      userEntity.type = type;
+    }
+
+    if (app) {
```
zskhan, line 190: When app is not present it should also clear the stale value from `description` and `category`


```suggestion
    if (app) {
      userEntity.description = app.description;
      userEntity.category = app.category;
    } else {
      userEntity.description = undefined;
      userEntity.category = undefined;
    }
```

zskhan (approved):

PR #20990, opened 2026-04-14
fix: Broken dashed line and dots behavior in the version history modal(WPB-24287)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24287" title="WPB-24287" target="_blank">BugWPB-24287</a>  [Web][versions] Broken dashed line and dots behavior in the version history modal
  </td></table>
  <br />
 

### Expected state:
1. if only one version is display per date, don’t display the dashed line
2. the dashed line should be displayed between the two middles of the dots
3. correct the shape of the dots
4. Fix dark mode issues on hover of file versions and close button

### Before:
The alignment and shape of dots and dashed line is off 
image-20260324-132229

Screenshot 2026-04-14 at 15 42 05

### After:
Screenshot 2026-04-14 at 15 40 15

Screenshot 2026-04-14 at 15 40 23

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryContent.tsx
+++ b/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryContent.tsx
@@ -47,6 +47,7 @@ export const FileHistoryContent = ({
                   key={version.versionId}
                   version={version}
                   isCurrentVersion={versionIndex === 0 && groupIndex === 0}
+                  showTimelineConnector={versionIndex < fileVersions[date].length - 1}
                   onDownload={handleDownload}
                   onRestore={handleRestore}
                 />
--- a/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
+++ b/apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryModal.styles.ts
@@ -65,19 +65,28 @@ export const fileHeaderFileInfoCss: CSSObject = {
   display: 'flex',
   alignItems: 'center',
   gap: '6px',
-  color: 'var(--gray-70)',
+  color: 'var(--base-secondary-text)',
   height: '21px',
 };
 
 export const fileHistoryCloseButtonCss: CSSObject = {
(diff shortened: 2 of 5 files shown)


apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryContent.tsx
```diff
@@ -47,6 +47,7 @@ export const FileHistoryContent = ({
                   key={version.versionId}
                   version={version}
                   isCurrentVersion={versionIndex === 0 && groupIndex === 0}
+                  showTimelineConnector={versionIndex < fileVersions[date].length - 1}
```
screendriver, line 50: What if `fileVersions[date]` returns `undefined` because the key does not exist?

screendriver (changes requested):

apps/webapp/src/script/components/Modals/FileHistoryModal/FileHistoryContent.tsx
arjita-mitra: This code path is executed under -  `{Object.keys(fileVersions).map((date, groupIndex) =>` and Object.keys(fileVersions) only returns keys that are actually present on the fileVersions object therefore fileVersions[date] shouldn't be undefined. In can only be undefined if the data gets malformed during runtime. 

If you want to harden it further i can add

```
const versions = fileVersions[date] ?? [];
then showTimelineConnector={versionIndex < versions.length - 1}
```

screendriver: Yes, I know. I just had in my mind "what could go wrong" when we extract the mapping function into its own module/function... but when I look at the outer scope I can see that we are already using that and even have `fileVersions[date]` access twice now... (which could be extracted)

So we can leave it as it is right now I guess.

screendriver (approved):

PR #20992, opened 2026-04-14
test: write the group conversation tests [WPB-19952]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19952" title="WPB-19952" target="_blank">TaskWPB-19952</a>  [Web/QA] Write the Group Conversation tests in Playwright
  </td></table>
  <br />
 

## Summary

This PR introduces new regression tests for Group Conversations.
---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
@@ -31,6 +31,7 @@ export class GroupCreationPage {
 
   readonly searchPeopleInput: Locator;
   readonly searchPeopleResults: Locator;
+  readonly errorGroupName: Locator;
 
   constructor(page: Page) {
     this.groupCreationModal = page.locator('#group-creation-modal');
@@ -44,6 +45,7 @@ export class GroupCreationPage {
     this.searchPeopleInput = page.getByRole('dialog').getByLabel('Search by name');
     this.searchPeopleList = page.getByRole('dialog').getByRole('list');
     this.searchPeopleResults = this.searchPeopleList.getByRole('listitem');
+    this.errorGroupName = page.getByTestId('error-group-name');
   }
 
   async setGroupName(name: string) {
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
@@ -31,7 +31,7 @@ export class GroupCreationPage {
 
   readonly searchPeopleInput: Locator;
   readonly searchPeopleResults: Locator;
-  readonly selectedPeopleList: Locator;
+  readonly selectedMembers: Locator;
   readonly toggleSelectedListButton: Locator;
   readonly errorGroupName: Locator;
 
@@ -47,7 +47,7 @@ export class GroupCreationPage {
     this.searchPeopleInput = page.getByRole('dialog').getByLabel('Search by name');
     this.searchPeopleList = page.getByRole('dialog').getByRole('list');
     this.searchPeopleResults = this.searchPeopleList.getByRole('listitem');
-    this.selectedPeopleList = page.getByTestId('selected-search-list');
+    this.selectedMembers = page.getByTestId('selected-search-list').getByRole('listitem');
     this.toggleSelectedListButton = page.getByTestId('do-toggle-selected-search-list');
     this.errorGroupName = page.getByTestId('error-group-name');
   }
@@ -91,6 +91,6 @@ export class GroupCreationPage {
 
   async deselectGroupMember(username: string) {
     await this.groupCreationModal.getByTestId('do-toggle-selected-search-list').click();
-    await this.selectedPeopleList.filter({hasText: username}).click();
+    await this.selectedMembers.filter({hasText: username}).click();
   }
 }
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
@@ -31,6 +31,8 @@ export class GroupCreationPage {
 
   readonly searchPeopleInput: Locator;
   readonly searchPeopleResults: Locator;
+  readonly selectedPeopleList: Locator;
+  readonly toggleSelectedListButton: Locator;
   readonly errorGroupName: Locator;
 
   constructor(page: Page) {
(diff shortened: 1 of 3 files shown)


apps/webapp/test/e2e_tests/pageManager/webapp/pages/groupCreation.page.ts
```diff
@@ -82,4 +88,9 @@ export class GroupCreationPage {
       await this.searchPeopleResults.filter({hasText: username}).click();
     }
   }
+
+  async deselectGroupMember(username: string) {
+    await this.groupCreationModal.getByTestId('do-toggle-selected-search-list').click();
+    await this.selectedPeopleList.filter({hasText: username}).click();
```
markbrockhoff: Does the locator "selectedPeopleList" really contain multiple matches? Since it targets a test-id it looks like it would only result in one match which is the list itself rather than individual items inside it. Are you sure there shouldn't be one more locator before filtering it? (Similar to how it is for the searchPeopleList and searchPeopleResults)
This might only work by accident since it clicks the whole list which currently only contains one user.

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Page} from 'playwright/test';
```
markbrockhoff: Looking at all these warnings you might want to clean up the imports a bit ^^'

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+  );
+
+  test('I want to delete a group as the Group Creator', {tag: ['@TC-1089', '@regression']}, async ({createPage}) => {
+    const [userAPage, userBPage] = await Promise.all([createPage(withLogin(userA)), createPage(withLogin(userB))]);
+    const {pages: userAPages, modals: userAModals} = PageManager.from(userAPage).webapp;
+    const userBPages = PageManager.from(userBPage).webapp.pages;
+
+    await createGroup(userAPages, groupName, [userB, userC]);
+    const adminGroupLocator = userAPages.conversationList().list.filter({hasText: groupName});
+    const memberGroupLocator = userBPages.conversationList().list.filter({hasText: groupName});
```
markbrockhoff: You're filtering a locator which will only match one element. This only works because within your test the list only contains one item, instead you can just use the existing `getConversationLocator()` util:

```suggestion
    const adminGroupLocator = userAPages.conversationList().getConversationLocator(groupName);
    const memberGroupLocator = userBPages.conversationList().getConversationLocator(groupName);
```

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    await expect(adminGroupLocator).toBeVisible();
+
+    // User can delete group as the group creator
+    await userAPages.conversationDetails().deleteGroupButton.click();
+    await userAModals.confirm().actionButton.click();
+
+    // Verify the group no longer exists for any user
+    await adminGroupLocator.waitFor({state: 'detached'});
+    await expect(adminGroupLocator).toBeHidden();
+    await expect(memberGroupLocator).toBeHidden();
```
markbrockhoff: Why the extra waitFor expression? This does the same as the assertion beneath it. Only difference would be the given default timeout for it, but if that's the reason for adding the extra waitFor please just increase the timeout for the assertion instead:

```suggestion
    await expect(adminGroupLocator).toBeHidden({timeout: 20_000});
    await expect(memberGroupLocator).toBeHidden({timeout: 20_000});
```

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    {tag: ['@TC-1093', '@regression']},
+    async ({createPage}) => {
+      const [userAPage, userBPage] = await Promise.all([
+        createPage(withLogin(userA), withConnectedUser(userB)),
+        createPage(withLogin(userB)),
+      ]);
+
+      const {pages: userAPages, modals: userAModals} = PageManager.from(userAPage).webapp;
+
+      await createGroup(userAPages, groupName, [userB]);
```
markbrockhoff, line 139: 🎨 Maybe it could be a good idea to first assert userB sees the created group before deleting it again. I'm not 100% sure all events are passed to the user in order so this way we could be sure we can't end up with some mismatched state

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
```diff
@@ -0,0 +1,333 @@
...
+    for (const group of groups) {
+      const {pages, modals} = PageManager.from(group.owner).webapp;
+
+      await pages.conversationList().openConversation(group.name);
+      await pages.conversation().clickConversationInfoButton();
+      await pages.conversationDetails().deleteGroupButton.click();
+      await modals.confirm().actionButton.click();
+      await pages.conversationList().list.filter({hasText: group.name}).waitFor({state: 'detached'});
+
+      await expect(userCPages.conversationList().list.filter({hasText: group.name})).not.toBeVisible();
```
markbrockhoff: Same as above, adding waitFor is just a less clear way of increasing the assertions timeout. Also please don't filter on locators which are not expected to match multiple elements.

```suggestion

      await expect(userCPages.conversationList().getConversationLocator(group.name)).not.toBeAttached({
        timeout: 20_000,
      });
```

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/GroupConversation/groupConversation.spec.ts
zhannach: Okay, I wasn't sure about the specific timeout value - is there just a 'safe' number or how did you come up to 20_000  timeout?

markbrockhoff: In the end the timeout should be the limit of what we think is the maximum acceptable time for a user to wait for something. E.g. for the login we defined this to be at max 40s. For all assertions this is 10s by default, e.g. if a user sent a message and the receiver asserts it to be visible it should appear within 10s otherwise the test is considered to have failed.

The main reason I suggested 20s in this case was because 20s is the default timeout for actions e.g. the `waitFor()` you used previously. This way the effective timeout stayed the same ;)
But I also think it's a reasonable value for the group to be removed for the other user, sure faster is better but I don't think a user should wait for more than max 30s for an update like this.

markbrockhoff (approved):

zskhan (approved):

PR #20994, opened 2026-04-14
test: avoid flake by scoping conversationList locators by protocol [WPB-24429]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24429" title="WPB-24429" target="_blank">TaskWPB-24429</a>  [Web][QA] Investigate E2E tests that are flaky because of possible race condition.
  </td></table>
  <br />
 

## Summary

While investigating flaky tests I noticed that one cause of flake is the usage of all the unscoped utils for the conversation list. While the `openConversattion()` function allows passing a protocol to ensure the right conversation will be picked independent of the background migration to MLS this isn't the case for the others.
So I took it as the opportunity to clean out the POM and have utils only relevant for individual conversations be scoped to them.

---

## Accessibility (required)

---

## Notes for reviewers

- This PR may look big and scary but it's atomic, when stepping through the commits you can review only one change at a time only touching a few files (or review the whole thing at once, whatever floats your boat :))

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
@@ -20,6 +20,8 @@
 import {Locator, Page} from '@playwright/test';
 import {GuestOptionsPage} from './guestOptions.page';
 
+type NotificationLevel = 'Everything' | 'Mentions and replies' | 'Nothing';
+
 export class ConversationDetailsPage {
   private readonly page: Page;
 
@@ -219,16 +221,16 @@ export class ConversationDetailsPage {
     await this.clearConversationContentButton.click();
   }
 
-  async setNotifications(value: Parameters<typeof this.selectNotificationsLevel>[0]) {
+  async setNotifications(level: NotificationLevel) {
     await this.notificationsButton.click();
-    await this.selectNotificationsLevel(value);
+    await this.selectNotificationsLevel(level);
 
     // Close the settings by clicking "Go back" button.
     await this.page.getByRole('button', {name: 'Go back'}).click();
   }
 
-  async selectNotificationsLevel(value: 'Everything' | 'Mentions and replies' | 'Nothing') {
-    await this.page.getByRole('radiogroup').getByText(value).click();
+  async selectNotificationsLevel(level: NotificationLevel) {
+    await this.page.getByRole('radiogroup').getByText(level).click();
   }
 
   async changeConversationName(newConversationName: string) {
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
@@ -219,12 +219,16 @@ export class ConversationDetailsPage {
     await this.clearConversationContentButton.click();
   }
 
-  async setNotifications(value: 'Everything' | 'Mentions and replies' | 'Nothing') {
+  async setNotifications(value: Parameters<typeof this.selectNotificationsLevel>[0]) {
     await this.notificationsButton.click();
-    await this.page.getByRole('radiogroup').getByText(value).click();
+    await this.selectNotificationsLevel(value);
 
     // Close the settings by clicking "Go back" button.
     await this.page.getByRole('button', {name: 'Go back'}).click();
+  }
+
+  async selectNotificationsLevel(value: 'Everything' | 'Mentions and replies' | 'Nothing') {
+    await this.page.getByRole('radiogroup').getByText(value).click();
   }
 
   async changeConversationName(newConversationName: string) {
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
@@ -28,7 +28,6 @@ export class ConversationListPage {
   readonly createGroupButton: Locator;
   readonly pendingConnectionRequest: Locator;
   readonly searchConversationsInput: Locator;
-  readonly moveToMenu: Locator;
   readonly conversationListHeaderTitle: Locator;
   readonly joinCallButton: Locator;
   readonly clearContentButton: Locator;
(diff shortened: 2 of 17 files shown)


zhannach (dismissed):

zskhan (dismissed):

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
```diff
@@ -219,14 +219,18 @@ export class ConversationDetailsPage {
     await this.clearConversationContentButton.click();
   }
 
-  async setNotifications(value: 'Everything' | 'Mentions and replies' | 'Nothing') {
+  async setNotifications(value: Parameters<typeof this.selectNotificationsLevel>[0]) {
```
e-maad: I would suggest not to extract parameter type due to reasons like readability and dependency.

Cleaner approach:
```
type NotificationsLevel = 'Everything' | 'Mentions and replies' | 'Nothing';

async selectNotificationsLevel(value: NotificationsLevel) {
    await this.page.getByRole('radiogroup').getByText(value).click();
  }

 async setNotifications(value: NotificationsLevel) {
...
 }
```

e-maad (changes requested):

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
markbrockhoff: Yes valid point. Thing is: this is a class, so your suggestion only works partially since it's not possible to declare the type in the same place it is used like this. I'd have to declare it outside of the class. Do you still think it's worth the tradeoff like this?

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
```diff
@@ -118,47 +63,39 @@ export class ConversationListPage {
     }
 
     return Object.assign(conversation, {
```
e-maad, line 71: Do we really need to use `Object.assign`?

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
```diff
@@ -118,47 +63,39 @@ export class ConversationListPage {
...
+   * @example
+   * const contextMenu = await pages.conversationList().openContextMenu(conversationLocator);
+   * await contextMenu.archiveButton.click();
+   */
+  async openContextMenu(conversation: Locator) {
+    await conversation.getByRole('button', {name: 'Open conversation options'}).click();
+
+    const contextMenu = this.page.getByRole('menu');
+
+    return Object.assign(contextMenu, {
```
e-maad, line 96: Same as https://github.com/wireapp/wire-webapp/pull/20994/changes#r3098978402

apps/webapp/test/e2e_tests/specs/Archive/archive.spec.ts
```diff
@@ -69,8 +69,11 @@ test.describe('Archive', () => {
       await memberBPages.conversationList().openConversation(memberA.fullName, {protocol: 'mls'});
 
       await test.step('MemberA archives conversation with memberB', async () => {
-        await memberAPages.conversationList().clickConversationOptions(memberB.fullName);
-        await memberAPages.conversationList().archiveConversation();
+        const contextMenu = await memberAPages
+          .conversationList()
+          .getConversationLocator(memberB.fullName, {protocol: 'mls'})
```
e-maad, line 74: Do we have a constant of `mls`?

apps/webapp/test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -94,9 +94,12 @@ test.describe('User Blocking', () => {
         // Step 1: User A opens conversation with User B
         await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
         // Step 2: User A opens the options menu for user B
-        await userAPages.conversationList().clickConversationOptions(userB.fullName);
+        const contextMenu = await userAPages
+          .conversationList()
+          .getConversationLocator(userB.fullName, {protocol: 'mls'})
```
e-maad, line 99: Same as https://github.com/wireapp/wire-webapp/pull/20994/changes#r3099014803

apps/webapp/test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -135,12 +138,15 @@ test.describe('User Blocking', () => {
         const message = userAPages.conversation().getMessage({sender: userB});
         await expect(message).not.toBeVisible();
         // Step 7: 'Blocked' chip is visible next to the name of User B in the conversation list
-        const statusTextElement = userAPages.conversationList().blockedChip;
+        const statusTextElement = userAPages
+          .conversationList()
+          .getConversationLocator(userB.fullName, {protocol: 'mls'}).blockedIndicator;
```
e-maad, line 143: Same as https://github.com/wireapp/wire-webapp/pull/20994/changes#r3099014803

apps/webapp/test/e2e_tests/specs/Block/block.spec.ts
```diff
@@ -204,8 +210,11 @@ test.describe('User Blocking', () => {
         await blockUserFromProfileView(userAPageManagerInstance);
         await expect(userAPages.conversation().messageInput).toBeHidden();
         // Step 3: User A unblocks User B from Conversation Details Options
-        await userAPages.conversationList().clickConversationOptions(userB.fullName);
-        await userAPages.conversationList().clickUnblockConversation();
+        const contextMenu = await userAPages
+          .conversationList()
+          .getConversationLocator(userB.fullName, {protocol: 'mls'})
```
e-maad, line 215: Same as https://github.com/wireapp/wire-webapp/pull/20994/changes#r3099014803

e-maad (changes requested):

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
e-maad: I would suggest, it will be more readable even outside the class.

apps/webapp/test/e2e_tests/specs/Archive/archive.spec.ts
markbrockhoff: IMHO it's pointless to create constants and import them all over the place for something which can be expressed via a simple string. E.g. I'd prefer `"mls"` over `PROTOCOL.MLS`.
Please keep in mind that `protocol` is typed as string literal union so only `undefined | "mls" | "proteus"` are valid values for it.

apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts
markbrockhoff: I'd say so or do you know a "better" (quotes because "better" is always subjective) way to extend an existing locator with new properties referencing the original locator while keeping everything fully type-safe without having to manually write types?

e-maad (approved):

PR #21000, opened 2026-04-15
chore: add eventListenerService [WPB-23186]

e-maad:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Add eventListenerService for purify functions and pass window.eventListener as a dependency.

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/eventListener/deterministicEventListenerService.ts
+++ b/apps/webapp/src/script/eventListener/deterministicEventListenerService.ts
@@ -1,0 +1,119 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/eventListener/deterministicEventListenerService.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      for (const entry of entries) {
+        if (entry.once) {
+          const current = listenerMap.get(type);
+
+          if (!is.undefined(current)) {
+            const index = current.indexOf(entry);
+
+            if (index !== -1) {
+              current.splice(index, 1);
```
zskhan, line 106: Mutating the same array inside a loop will skip the some array elements.
[eventOnce, event1, event2]
if you remove eventOnce after the first iteration then event1 will be skipped in the second iteration as it will be on index 0 and on index 1 it will be event2.
So we can store the events to be removed on another array and delete it after the loop.

screendriver:
I'm not convinced this module is a useful abstraction in its current form. To me this reads more like a generic event-emitter-style wrapper than a clean architecture boundary. `eventListenerService` is essentially a thin wrapper around `global` event APIs and `deterministicEventListenerService` adds another generic listener layer on top, but without making the domain model clearer.

I also find the name `deterministic` misleading here. In this context, determinism usually refers to controlling sources of runtime variability such as time, network responses, randomness, retries, or scheduling. This implementation does not seem to address that kind of nondeterminism; it mainly provides synchronous in-memory listener dispatch.

I'm also concerned about the typing. Accepting arbitrary string event names and generic `Event` values means the compiler cannot help us validate that subscriptions and dispatched payloads actually match.

My preference would be to keep global/browser event registration at the application entry point and pass concrete function callbacks inward into the application layer. That keeps the browser-specific wiring at the edge, makes dependencies explicit and gives us better type safety than introducing a generic event-listener abstraction.

e-maad:
Closing this PR as we have other prioritized tasks.

PR #21005, opened 2026-04-15
feat: add state-machine for appLockflow [WPB-23186]

e-maad:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Added state machine for appLock flow
https://stately.ai/registry/editor/5b37cbe9-127b-4b50-b3bb-3ffec900e2b6?mode=Simulate&machineId=9d4336ce-95a7-47a3-94a5-2ee7288d31a3
Screenshot 2026-04-15 at 21 04 39

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
+++ b/apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
@@ -1,0 +1,275 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {assign, createMachine, fromCallback} from 'xstate';
+
+import {WallClock} from '../../clock/wallClock';
+
+const DEFAULT_INACTIVITY_TIMEOUT_MS = 60_000; // 1 minute
+
+/**
+ * DOM events that count as user activity and reset the inactivity debounce timer.
+ */
+const ACTIVITY_EVENTS = ['mousemove', 'keydown', 'mousedown', 'touchstart', 'wheel', 'scroll'] as const;
+
+export const appLockEventType = Object.freeze({
+  /** User enables "Lock with passcode" in preferences. */
+  enable: 'ENABLE',
+  /** User disables "Lock with passcode" in preferences. */
+  disable: 'DISABLE',
+  /** Team policy begins enforcing app lock. */
+  lockEnforced: 'LOCK_ENFORCED',
+  /** User successfully set or changed their passcode. */
+  passcodeConfirmed: 'PASSCODE_CONFIRMED',
+  /** Inactivity debounce timer fired — no user activity for inactivityTimeoutMs. */
+  inactivityTimeout: 'INACTIVITY_TIMEOUT',
+  /** User entered the correct passcode on the lock screen. */
+  unlockSuccess: 'UNLOCK_SUCCESS',
+  /** User clicked "Forgot passcode" on the lock screen. */
+  forgotPasscode: 'FORGOT_PASSCODE',
+  /** User navigates back to locked screen from forgotPasscode. */
+  backToLocked: 'BACK_TO_LOCKED',
+  /** User wants to change their passcode from preferences. */
+  changePasscode: 'CHANGE_PASSCODE',
+  /** User chooses "Logout + delete all data" from forgotPasscode screen. */
+  logoutWithWipe: 'LOGOUT_WITH_WIPE',
+  /** User chooses "Logout without deleting data" from forgotPasscode screen. */
+  logoutWithoutWipe: 'LOGOUT_WITHOUT_WIPE',
+  /** User confirms the data wipe from wipeConfirm screen. */
+  confirmWipe: 'CONFIRM_WIPE',
+  /** User cancels the data wipe from wipeConfirm screen. */
+  cancelWipe: 'CANCEL_WIPE',
+} as const);
+
+export type AppLockMachineEvent = {
+  [K in keyof typeof appLockEventType]: {type: (typeof appLockEventType)[K]};
+}[keyof typeof appLockEventType];
+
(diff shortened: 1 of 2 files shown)


screendriver (changes requested):

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,275 @@
...
+  });
+
+  scheduleTimeout();
+
+  return () => {
+    if (debounceTimeoutId !== null) {
+      wallClock.clearTimeout(debounceTimeoutId);
+    }
+    ACTIVITY_EVENTS.forEach(event => {
+      globalThis.removeEventListener(event, scheduleTimeout);
```
screendriver: Same as in https://github.com/wireapp/wire-webapp/pull/21005#discussion_r3091521958

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,275 @@
...
+  if (!input.isEnabled && !input.isEnforced) {
+    return appLockStateName.unprotected;
+  }
+  if (input.hasPasscode) {
+    return appLockStateName.unlocked;
+  }
+  return appLockStateName.setup;
+}
+
+export function createAppLockStateMachine(input: AppLockMachineInput) {
```
screendriver, line 181: I think this should come with focused unit tests before merge. We are introducing stateful behavior with transitions, guards and timeout handling and I do not see corresponding test coverage in this branch. I would be much more comfortable merging this once the machine behavior is tested directly.

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,275 @@
...
+};
+
+type InactivityActorInput = {
+  readonly wallClock: WallClock;
+  readonly timeoutMs: number;
+};
+
+type InactivityActorEvent = {type: typeof appLockEventType.inactivityTimeout};
+
+const inactivityCallbackActor = fromCallback<InactivityActorEvent, InactivityActorInput>(({sendBack, input}) => {
```
screendriver: This actor currently mixes state-machine behavior with runtime wiring. I think the separation would be cleaner if the machine consumed explicit activity events and the browser event subscription lived outside the machine behind an injected boundary.

screendriver:
I think this pull request would be easier to review and safer to roll back if the CI / dependency-management changes were separated from the product behavior changes. Right now the branch combines app-lock work with workflow and automation updates, which makes review scope larger than necessary.

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,275 @@
...
+    if (debounceTimeoutId !== null) {
+      wallClock.clearTimeout(debounceTimeoutId);
+    }
+    debounceTimeoutId = wallClock.setTimeout(() => {
+      sendBack({type: appLockEventType.inactivityTimeout});
+    }, timeoutMs);
+  };
+
+  ACTIVITY_EVENTS.forEach(event => {
+    globalThis.addEventListener(event, scheduleTimeout, {passive: true});
```
screendriver: I would prefer not to access `globalThis` directly inside the state machine. This makes the runtime boundary implicit and couples the machine to browser APIs. Could we inject an activity source or a small runtime adapter instead? That would keep the machine more isolated and make deterministic unit testing much easier.

screendriver (changes requested):

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,275 @@
...
+
+type InactivityActorEvent = {type: typeof appLockEventType.inactivityTimeout};
+
+const inactivityCallbackActor = fromCallback<InactivityActorEvent, InactivityActorInput>(({sendBack, input}) => {
+  const {wallClock, timeoutMs} = input;
+  let debounceTimeoutId: ReturnType<typeof wallClock.setTimeout> | null = null;
+
+  const scheduleTimeout = () => {
+    if (debounceTimeoutId !== null) {
+      wallClock.clearTimeout(debounceTimeoutId);
```
screendriver: Small naming/behavior point: this does not seem to implement a real debounce of the activity events themselves. It resets an inactivity timeout on every event which is slightly different. Could we either rename this to reflect that more precisely or move to an explicit debounced activity source if we want true event-rate limiting?

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
e-maad: I intentionally kept the PR short to be able to merge quickly.

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
e-maad: There is another PR to add deterministic event listener service: https://github.com/wireapp/wire-webapp/pull/21000

e-maad: It Will be there in next PR.

e-maad:
@screendriver, waiting for your review.

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,293 @@
...
+        },
+      },
+
+      [appLockStateName.setup]: {
+        on: {
+          [appLockEventType.passcodeConfirmed]: {
+            target: appLockStateName.unlocked,
+            actions: assign({isChangingPasscode: false}),
+          },
+          [appLockEventType.disable]: {
```
screendriver, line 218: This changes the current product behavior for optional app-lock setup.

`DISABLE` in `setup` is currently guarded by `!context.isEnforced && context.isChangingPasscode`, which means first-time setup cannot be cancelled after the user enables app lock. In the existing user interface that cancel path is allowed when app lock is optional and not yet activated.

apps/webapp/src/script/page/AppLock/appLockStateMachine.ts
```diff
@@ -0,0 +1,293 @@
...
+          },
+          [appLockEventType.changePasscode]: {
+            target: appLockStateName.setup,
+            actions: assign({isChangingPasscode: true}),
+          },
+          [appLockEventType.disable]: {
+            target: appLockStateName.unprotected,
+            guard: ({context}) => !context.isEnforced,
+          },
+          [appLockEventType.lockEnforced]: {
```
screendriver, line 250: `LOCK_ENFORCED` only updates context in `unprotected` and `unlocked`. If enforcement changes while the machine is in `setup`, `locked`, `forgotPasscode` or `wipeConfirm`, `context.isEnforced` stays stale and later guard decisions can be wrong.

screendriver (changes requested):

e-maad:
Closing this PR as we have other prioritized tasks.

PR #21020, opened 2026-04-16
test: reduce flake by limiting number of parallel tests per agent to 1 [WPB-22420]

markbrockhoff:
<table>
<td>
  </td></table>
  <br />
 

## Summary

This PR is an attempt to reduce flake by setting the number of parallel tests per worker back to 1. It was increased to 3 some time ago but since we're recently encountering a lot of flake it might still be better to rather execute less tests in parallel but have them pass then having to retry a lot.

---

## Accessibility (required)

---

## Notes for reviewers

-

### DIFF
--- a/apps/webapp/playwright.config.ts
+++ b/apps/webapp/playwright.config.ts
@@ -23,8 +23,9 @@ import {resolve} from 'node:path';
 
 config({path: resolve(__dirname, './test/e2e_tests/.env'), quiet: true});
 
+// The number of parallel workers needs to be limited since otherwise the number of parallel requests could run into rate limiting by the backend
+const numberOfParallelWorkersOnCI = 1;
 const numberOfRetriesOnCI = 2;
-const numberOfParallelWorkersOnCI = 3;
 
 /**
  * See https://playwright.dev/docs/test-configuration.


apps/webapp/src/script/auth/page/SingleSignOnForm.tsx
```diff
@@ -343,7 +343,7 @@ const SingleSignOnFormComponent = ({
         <Button
           block
           type="submit"
-          disabled={!codeOrMail}
+          disabled={isFetching || !codeOrMail}
```
markbrockhoff: While investigating I noticed that sometimes the login button on the SSO page was enabled but clicking it would simply do nothing. I traced the cause of this down to the early return within `handleSubmit` if some auth request is still ongoing in the background. So I updated this condition to reflect reality and disable the button while clicking it would be pointless.

arjita-mitra: unrelated change to this PR right? Although small still we should only aim for changes aligning with the main purpose of the PR unless i am missing something.

arjita-mitra (dismissed):

apps/webapp/src/script/auth/page/SingleSignOnForm.tsx
markbrockhoff: Ok, then it'll be two 1 line PRs: #21044

arjita-mitra (approved):

PR #21025, opened 2026-04-16
feat(appLock): add logout flow to app lock modal [WPB-23186]

e-maad:
<table>
<td>
  </td></table>
  <br />
 

## Summary
The PR simplifies the App Lock feature by replacing the old “wipe data” flow with a clearer logout option (with optional data clearing), improves the “forgot passcode” experience, updates the UI with better components and styles, cleans up related code, and refreshes tests and wording for better usability and maintainability.

Reference.: https://github.com/wireapp/wire-webapp/pull/20391

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/script/page/AppLock/AppLock.tsx
```diff
@@ -73,11 +72,10 @@ const AppLock = ({
   appLockRepository = container.resolve(AppLockRepository),
 }: AppLockProps) => {
   const [state, setState] = useState<APPLOCK_STATE>(APPLOCK_STATE.NONE);
```
arjita-mitra: please rename this to setAppLockState setState isn't explicit enough

apps/webapp/src/script/page/AppLock/AppLock.tsx
```diff
@@ -264,20 +253,24 @@ const AppLock = ({
...
-      case APPLOCK_STATE.WIPE_PASSWORD:
-        return t('modalAppLockWipePasswordTitle', {brandName: Config.getConfig().BRAND_NAME});
       default:
         return '';
     }
   };
 
+  const ErrorMessage = () => (
+    <p className="modal__input__error" data-uie-name="label-applock-unlock-error">
+      {unlockError}
```
arjita-mitra: this comment makes sense?

apps/webapp/src/script/page/AppLock/AppLock.tsx
```diff
@@ -306,22 +299,16 @@ const AppLock = ({
...
-              className="modal__text modal__label"
-              data-uie-name="label-applock-unlock-text"
-              htmlFor="input-applock-set-a"
-            >
-              {t('modalAppLockPasscode')}
-            </label>
-
             {/* eslint jsx-a11y/no-autofocus : "off" */}
-            <input
+            <Input
```
arjita-mitra, line 297: Could you check TextInput component in webapp? it has the latest error icon, red border of label and field and a clear button to clear the input. that might be a better fit here?

apps/webapp/src/script/page/AppLock/AppLock.tsx
```diff
@@ -447,147 +434,98 @@ const AppLock = ({
...
-              className="modal__input"
               type="password"
+              placeholder={t('modalAppLockInputPlaceholder')}
               id={Math.random().toString()}
               name="password"
               onKeyDown={clearUnlockError}
               data-uie-name="input-applock-unlock"
-              autoComplete="new-password"
+              autoComplete="current-password"
+              error={ErrorMessage()}
```
arjita-mitra: Could you add aria-invalid={Boolean(unlockError)} to set/reset  error and aria-describedby with the id of the <p> element to describe the error if not already there as part of Input component?

apps/webapp/src/script/page/AppLock/AppLock.tsx
```diff
@@ -241,12 +226,16 @@ const AppLock = ({
...
 
-  const clearWipeError = () => setWipeError('');
   const clearUnlockError = () => setUnlockError('');
   const onGoBack = () => setState(APPLOCK_STATE.LOCKED);
   const onClickForgot = () => setState(APPLOCK_STATE.FORGOT);
-  const onClickWipe = () => setState(APPLOCK_STATE.WIPE_CONFIRM);
-  const onClickWipeConfirm = () => setState(APPLOCK_STATE.WIPE_PASSWORD);
+  const onClickLogout = async () => {
+    if (isTemporaryClient) {
+      await clientRepository.logoutClient();
```
arjita-mitra, line 234: no error handling? taken care of inside logout function?

e-maad: The error handled in the called function. Even if we can a general exception at this point of code we don't have any other way but swallowing it.

apps/webapp/src/script/page/AppLock/AppLock.tsx
e-maad: The usage of the component is very limited and the functionality is not enough (for example it does not support className attr) what we need for AppLock.
I will keep `<Input` for now.

apps/webapp/src/script/page/AppLock/AppLock.tsx
e-maad: Which comment?

arjita-mitra: the sonar cloud comment 

Move this component definition out of the parent component and pass data as props.
Check warning: Move this component definition out of the parent component and pass data as props.

See more on https://sonarcloud.io/project/issues?id=wireapp_wire-webapp&issues=AZ2YZ_IErAxzBqI9Lo0P&open=AZ2YZ_IErAxzBqI9Lo0P&pullRequest=21025

apps/webapp/src/script/page/AppLock/AppLock.tsx
arjita-mitra: will the error message be announced for the user? is it handled inside Input component? i see you have removed the <p> block so where are we showing the error and whether its announced for the sr-user?

e-maad: Yes, basically the error container is always there but when something is wrong, we have a string that we populate to show the error:   `const [unlockError, setUnlockError] = useState('');`

e-maad: Image

arjita-mitra (approved):

PR #21032, opened 2026-04-17
test(WPB-19947): Add e2e regression tests for deep links

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19947" title="WPB-19947" target="_blank">TaskWPB-19947</a>  [Web/QA] Write the Deep Links regression tests in Playwright
  </td></table>
  <br />
 

## Summary

- add regression tests for deep links

---

## Accessibility (required)

---

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
@@ -24,12 +24,39 @@ export class UserProfileModal {
   readonly connectButton: Locator;
   readonly startConversationButton: Locator;
   readonly unblockButton: Locator;
+  readonly blockButton: Locator;
+  readonly openConversationButton: Locator;
+  readonly guestChip: Locator;
+  readonly connectWarning: Locator;
+  readonly participantFullname: Locator;
+  readonly participantUsername: Locator;
+  readonly userEmailLabel: Locator;
+  readonly userEmailEntry: Locator;
+  readonly domainLabel: Locator;
+  readonly cancelButton: Locator;
+  readonly modalCloseButton: Locator;
 
   constructor(page: Page) {
     this.modal = page.getByTestId('modal-user-profile');
     this.connectButton = page.getByTestId('modal-user-profile').getByTestId('do-send-request');
     this.startConversationButton = page.getByTestId('modal-user-profile').getByTestId('start-conversation');
     this.unblockButton = page.getByTestId('modal-user-profile').getByTestId('do-unblock');
+    this.blockButton = page.getByRole('button', {name: 'Block...'});
+    this.openConversationButton = this.modal.getByRole('button', {name: 'Open conversation'}).getByTestId('go-conversation');
+    this.guestChip = page.getByTestId('status-guest');
+    this.connectWarning = page.getByText(/Get certainty about .*’s identity before connecting/);
+    this.participantFullname = this.modal.getByTestId('status-label');
+    this.participantUsername = this.modal.getByTestId('status-username');
+
+    const emailContainer = page.locator('.enriched-fields__entry').filter({
+      has: page.locator('[data-uie-name="item-enriched-key"]', {hasText: 'Email'}),
+    });
+    this.userEmailLabel = emailContainer.getByRole('paragraph').getByText('Email');
+    this.userEmailEntry = emailContainer.getByTestId('item-enriched-value');
+
+    this.domainLabel = page.getByRole('paragraph').getByText('Domain');
+    this.cancelButton = this.modal.getByRole('button', {name: 'Cancel', exact: true});
+    this.modalCloseButton = this.modal.getByTestId('do-close');
   }
 
   async isVisible() {
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
@@ -41,7 +41,7 @@ export class UserProfileModal {
     this.connectButton = page.getByTestId('modal-user-profile').getByTestId('do-send-request');
     this.startConversationButton = page.getByTestId('modal-user-profile').getByTestId('start-conversation');
     this.unblockButton = page.getByTestId('modal-user-profile').getByTestId('do-unblock');
-    this.blockButton = page.getByRole('button', {name: 'Block...'});
+    this.blockButton = page.getByRole('button', {name: 'Block…'}).getByTestId('do-block');
     this.openConversationButton = this.modal.getByRole('button', {name: 'Open conversation'}).getByTestId('go-conversation');
     this.guestChip = page.getByTestId('status-guest');
     this.connectWarning = page.getByText(/Get certainty about .*’s identity before connecting/);
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
@@ -54,7 +54,10 @@ export class UserProfileModal {
     this.userEmailLabel = emailContainer.getByRole('paragraph').getByText('Email');
     this.userEmailEntry = emailContainer.getByTestId('item-enriched-value');
 
-    this.domainLabel = page.getByRole('paragraph').getByText('Domain');
+    const domainContainer = page.locator('.enriched-fields__entry').filter({
+      has: page.locator('[data-uie-name="item-enriched-key"]', {hasText: 'Domain'}),
+    });
+    this.domainLabel = domainContainer.getByRole('paragraph').getByText('Domain');
     this.cancelButton = this.modal.getByRole('button', {name: 'Cancel', exact: true});
     this.modalCloseButton = this.modal.getByTestId('do-close');
   }
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
@@ -41,10 +41,9 @@ export class UserProfileModal {
     this.connectButton = page.getByTestId('modal-user-profile').getByTestId('do-send-request');
     this.startConversationButton = page.getByTestId('modal-user-profile').getByTestId('start-conversation');
     this.unblockButton = page.getByTestId('modal-user-profile').getByTestId('do-unblock');
-    this.blockButton = page.getByRole('button', {name: 'Block…'}).getByTestId('do-block');
+    this.blockButton = page.getByRole('button', {name: 'Block…', exact: true});
     this.openConversationButton = this.modal
-      .getByRole('button', {name: 'Open conversation'})
-      .getByTestId('go-conversation');
+      .getByRole('button', {name: 'Open conversation', exact: true});
     this.guestChip = page.getByTestId('status-guest');
     this.connectWarning = page.getByText(/Get certainty about .*’s identity before connecting/);
     this.participantFullname = this.modal.getByTestId('status-label');
@@ -52,10 +51,7 @@ export class UserProfileModal {
 
     this.userEmailLabel = this.modal.getByRole('paragraph').getByText('Email');
 
-    const domainContainer = page.locator('.enriched-fields__entry').filter({
-      has: page.locator('[data-uie-name="item-enriched-key"]', {hasText: 'Domain'}),
-    });
-    this.domainLabel = domainContainer.getByRole('paragraph').getByText('Domain');
+    this.domainLabel = this.modal.getByRole('paragraph').getByText('Domain');
     this.cancelButton = this.modal.getByRole('button', {name: 'Cancel', exact: true});
     this.modalCloseButton = this.modal.getByTestId('do-close');
     this.openProfileButton = this.modal.getByTestId('go-profile');
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
@@ -35,6 +35,7 @@ export class UserProfileModal {
   readonly domainLabel: Locator;
   readonly cancelButton: Locator;
   readonly modalCloseButton: Locator;
+  readonly openProfileButton: Locator;
 
   constructor(page: Page) {
     this.modal = page.getByTestId('modal-user-profile');
@@ -62,6 +63,7 @@ export class UserProfileModal {
     this.domainLabel = domainContainer.getByRole('paragraph').getByText('Domain');
     this.cancelButton = this.modal.getByRole('button', {name: 'Cancel', exact: true});
     this.modalCloseButton = this.modal.getByTestId('do-close');
+    this.openProfileButton = this.modal.getByTestId('go-profile');
   }
 
   async isVisible() {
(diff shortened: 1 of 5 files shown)


apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+
+      const [userAPage, userBPage, userCPage, userDPage] = await Promise.all([
+        createPage(withLogin(userA), withConnectionRequest(userC)),
+        createPage(withLogin(userB), withConnectedUser(userA), withConnectionRequest(userD)),
+        createPage(withLogin(userC)),
+        createPage(withLogin(userD)),
+      ]);
+
+      const [userAPageManager, userBPageManager, userCPageManager, userDPageManager] = await Promise.all([
+        PageManager.from(userAPage),
```
markbrockhoff: In this case the error from Sonar is not a false positive, you're awaiting an array of non promises within `Promise.all`, you already awaited all pages above so the creation of the PageManagers is no longer async.

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+        PageManager.from(userBPage),
+        PageManager.from(userCPage),
+        PageManager.from(userDPage),
+      ]);
+
+      await Promise.all(
+        [userBPage, userCPage, userDPage].map(page =>
+          page.context().grantPermissions(['clipboard-write', 'clipboard-read']),
+        ),
+      );
```
markbrockhoff: This should not be needed, all browsers created within playwright already have permission to write and read to the clipboard, it's granted globally within the `playwright.config.ts`

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
```diff
@@ -24,12 +24,44 @@ export class UserProfileModal {
...
   readonly startConversationButton: Locator;
   readonly unblockButton: Locator;
+  readonly blockButton: Locator;
+  readonly openConversationButton: Locator;
+  readonly guestChip: Locator;
+  readonly connectWarning: Locator;
+  readonly participantFullname: Locator;
+  readonly participantUsername: Locator;
+  readonly userEmailLabel: Locator;
+  readonly userEmailEntry: Locator;
```
markbrockhoff: Is it really necessary to have individual locators for the email label and entry? E.g. is there a specific condition to assert something only on the label? Otherwise we could maybe only use the label to build locators off of it instead.

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
```diff
@@ -24,12 +24,44 @@ export class UserProfileModal {
...
+    this.guestChip = page.getByTestId('status-guest');
+    this.connectWarning = page.getByText(/Get certainty about .*’s identity before connecting/);
+    this.participantFullname = this.modal.getByTestId('status-label');
+    this.participantUsername = this.modal.getByTestId('status-username');
+
+    const emailContainer = page.locator('.enriched-fields__entry').filter({
+      has: page.locator('[data-uie-name="item-enriched-key"]', {hasText: 'Email'}),
+    });
+    this.userEmailLabel = emailContainer.getByRole('paragraph').getByText('Email');
+    this.userEmailEntry = emailContainer.getByTestId('item-enriched-value');
```
markbrockhoff: Those are pretty complicated locators for a simple email text, do you think it might me more reasonable to fix the implementation to be accessible instead of having to use such workarounds to make it testable?
Looking at the used classes the component under test is defined in `apps/webapp/src/script/components/panel/EnrichedFields.tsx`, adding an aria-labelledby attribute there could allow the locator to just be `this.userEmailEntry = this.modal.getByRole('paragraph', {name: 'Email'});`

Something like this inside EnrichedFields.tsx could do the trick:
```tsx
      {fields?.length >= 1 &&
        fields.map(({type, value}) => (
          <div key={type} className="enriched-fields__entry">
            <p
              id={`${enrichedFieldsId}-${type}`}
              role="label"
              className="enriched-fields__entry__key"
              data-uie-name="item-enriched-key"
            >
              {type}
            </p>
            <p
              aria-labelledby={`${enrichedFieldsId}-${type}`}
              className="enriched-fields__entry__value"
              data-uie-name="item-enriched-value"
              data-uie-value={value}
            >
              {value}
            </p>
          </div>
        ))}
```

> With `const enrichedFieldsId = useId();`

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+      await userDPages.conversationList().pendingConnectionRequest.click();
+      await userDPages.connectRequest().connectButton.click();
+
+      await test.step('User B copies and sends profile deeplink to User A and User A verifies information', async () => {
+        const copiedProfileLinkUserB = await copyProfileLink(userBPage, userBPageManager);
+        await userBPages.conversationList().openConversation(userA.fullName, {protocol: 'mls'});
+        await userBPages.conversation().sendMessage('UserB profile link: ' + copiedProfileLinkUserB);
+
+        await userAPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+        await userAPages.conversation().getMessage({sender: userB}).click();
```
markbrockhoff: Better only click the link than the whole message ;)

```suggestion
        await userAPages.conversation().getMessage({sender: userB}).getByRole('link').click();
```

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    if (options.showCancelButton) {
+      await expect(profileModal.cancelButton).toBeVisible();
+    }
+  });
+}
+
+async function copyProfileLink(page: Page, pageManager: PageManager): Promise<string> {
+  const {pages, components} = pageManager.webapp;
```
markbrockhoff: 🎨 Instead of passing both the page and the pageManager you could just create the PageManager from the given page. This ensures there will never be a mixup where, e.g. userB page could be used with the userCPageManager acidentially.

```suggestion
async function copyProfileLink(page: Page): Promise<string> {
  const {pages, components} = PageManager.from(page).webapp;
```

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+        });
+      });
+    },
+  );
+
+  test(
+    'Opening profile deep links as a personal account',
+    {tag: ['@TC-591', '@regression']},
+    async ({createTeam, createUser, createPage}) => {
+      userA = await createUser();
```
markbrockhoff: The setup for this test looks very similar to the setup of the one above, do you think it could make sense to extract it into a shared `beforeEach`?

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+      });
+
+      await test.step('User D copies and sends profile deeplink to User B, who sends this link to User A and User A verifies information', async () => {
+        const copiedProfileLinkUserD = await copyProfileLink(userDPage, userDPageManager);
+        await userDPages.conversationList().openConversation(userB.fullName, {protocol: 'mls'});
+        await userDPages.conversation().sendMessage('UserD profile link: ' + copiedProfileLinkUserD);
+        await expect(userDPages.conversation().getMessage({sender: userD})).toBeVisible();
+
+        await userBPages.conversationList().openConversation(userD.fullName, {protocol: 'mls'});
+        const profileLinkUserD = userBPages.conversation().getMessage({sender: userD}).last();
```
markbrockhoff: User D doesn't send any other messages right? Then there should be no need for the `.last()`


```suggestion
        const profileLinkUserD = userBPages.conversation().getMessage({sender: userD});
```

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
markbrockhoff: (Also applies to all other cases where a link in a message shall be clicked)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: true, deleted it

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
JacquelineLehner: not necessary, I reduced it to only the emailLabel 👍

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
JacquelineLehner: ✅ 

thank you :)

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: I already looked into extracting this into a `beforeEach`, but because the setup varies slightly between the two tests (in the first test userA is the team owner, while in the second it's userB), it gets a bit tricky to manage with the Playwright fixtures. Instead I decided to parameterize the test since I noticed that the actual test steps are identical. Lmk what you think of this approach 🤔 😄

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+    if (options.showOpenProfileButton) {
+      await expect(profileModal.openProfileButton).toBeVisible();
+    }
+  });
+}
+
+async function copyProfileLink(page: Page): Promise<string> {
+  const {pages, components} = PageManager.from(page).webapp;
+
+  await components.conversationSidebar().preferencesButton.click();
```
zhannach: ConversationSidebar has method for this:
`
await components.conversationSidebar().clickPreferencesButton()`

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+  await pages.settings().copyProfileLinkButton.click();
+  await components.conversationSidebar().allConverationsButton.click();
```
zhannach: `await components.conversationSidebar().clickAllConversationsButton()`

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  return copiedLink;
+}
+
+test.describe('Deep Links', () => {
+  const testCases = [
+    {
+      title: 'Opening profile deep links as a team member',
+      tag: ['@TC-590', '@regression'],
+      setupUsers: async (createTeam: any, createUser: any, createPage: any) => {
```
zhannach: Can you substitute "any" for "types" here?

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
```diff
@@ -0,0 +1,270 @@
...
+      const userBPageManager = PageManager.from(userBPage);
+      const userCPageManager = PageManager.from(userCPage);
+      const userDPageManager = PageManager.from(userDPage);
+
+      const {pages: userAPages, modals: userAModals} = userAPageManager.webapp;
+      const {pages: userBPages} = userBPageManager.webapp;
+      const {pages: userCPages} = userCPageManager.webapp;
+      const {pages: userDPages} = userDPageManager.webapp;
+
+      await userCPages.conversationList().pendingConnectionRequest.click();
```
zhannach: In most cases, we already have methods for all this click action inside PageObject

`await userCPages.conversationList().openPendingConnectionRequest()`

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
```diff
@@ -24,12 +24,41 @@ export class UserProfileModal {
...
+  readonly cancelButton: Locator;
+  readonly modalCloseButton: Locator;
+  readonly openProfileButton: Locator;
 
   constructor(page: Page) {
     this.modal = page.getByTestId('modal-user-profile');
     this.connectButton = page.getByTestId('modal-user-profile').getByTestId('do-send-request');
     this.startConversationButton = page.getByTestId('modal-user-profile').getByTestId('start-conversation');
     this.unblockButton = page.getByTestId('modal-user-profile').getByTestId('do-unblock');
+    this.blockButton = page.getByRole('button', {name: 'Block…'}).getByTestId('do-block');
```
zhannach: Is the double-chaining necessary for this element?
We generally prefer using accessibility selectors (like getByRole) over data-testid where possible. However, if we are going to use getByTestId('do-block'), is there a reason we’re chaining it with the role? Usually, the test ID should be unique enough on its own

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: For some reason I thought we shouldn't use these click-functions. Changed it 👍

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: replaced it ✅

apps/webapp/test/e2e_tests/pageManager/webapp/modals/userProfile.modal.ts
JacquelineLehner: I got rid of the `getByTestId` and `page.getByRole('button', {name: 'Block…', exact: true})` did the trick 😊 👍

apps/webapp/test/e2e_tests/specs/DeepLinks/deepLinks.spec.ts
JacquelineLehner: ✅

markbrockhoff (approved):

screendriver (approved):

PR #21035, opened 2026-04-17
feat: search for apps (WPB-20363)

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20363" title="WPB-20363" target="_blank">TaskWPB-20363</a>  [Web] [Integration] - Searching for Apps specifically
  </td></table>
  <br />
 

## Summary

- What did I change and why?

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
```diff
@@ -123,6 +124,19 @@ const AddParticipants: FC<AddParticipantsProps> = ({
     return connectedUsers;
   }, [connectedUsers, isServicesRoom, isTeam, isTeamOnly, teamMembers, teamUsers]);
 
+  const apps = useMemo(() => {
+    const normalizedQuery = searchInput.trim().toLowerCase();
+    return contacts
+      .filter(contact => contact.type === UserType.APP)
+      .map(contact => integrationRepository.mapServiceFromUser(contact))
+      .filter(contact => compareTransliteration(contact.name(), normalizedQuery))
+      .sort((serviceA, serviceB) => sortByPriority(serviceA.name(), serviceB.name(), normalizedQuery));
```
e-maad, line 133: Can you please normalize the code?
Do we need 2 filters ? Is the array not already sorted?

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
```diff
@@ -190,7 +197,7 @@ const AddParticipants: FC<AddParticipantsProps> = ({
   const onSearchInput = async (value: string) => {
     setSearchInput(value);
 
-    if (isAddServiceState) {
+    if (isAddServiceState && CONVERSATION_PROTOCOL.PROTEUS) {
```
e-maad: Do you know `CONVERSATION_PROTOCOL.PROTEUS` will always be true?
Maybe you forgot the comparison.

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
```diff
@@ -273,7 +280,7 @@ const AddParticipants: FC<AddParticipantsProps> = ({
 
           {isAddServiceState && (
             <>
-              {!!services.length && (
+              {!!servicesList.length && (
```
e-maad, line 283: Please use: https://github.com/sindresorhus/is for better type guarding and comparisons.

apps/webapp/src/script/repositories/integration/IntegrationMapper.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  it('Should return an empty mapper', () => {
+    const user = new User();
+    const previewAsset = new AssetRemoteData({assetKey: 'previewKey', assetDomain: 'previewDomain'});
+    const mediumAsset = new AssetRemoteData({assetKey: 'mediumKey', assetDomain: 'mediumDomain'});
+    user.id = "testid";
+    user.description = "testdescription";
+    user.category = "testcategory";
+    user.previewPictureResource(previewAsset);
+    user.mediumPictureResource(mediumAsset);
+    user.type = UserType.APP;
```
e-maad: Do we have a factory that can generate test user object?
If not can you please add a factory function to generate it?

apps/webapp/src/script/repositories/integration/IntegrationMapper.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    const result = IntegrationMapper.mapServiceFromUser(user);
+
+    expect(result.id).toEqual("testid");
+    expect(result.description).toEqual("testdescription");
```
e-maad: Once you have an object, you don't need hardcoded string for assertions.

apps/webapp/src/script/repositories/integration/IntegrationMapper.ts
```diff
@@ -89,6 +92,20 @@ export const IntegrationMapper = {
...
+      serviceEntity.id = user.id;
+      serviceEntity.name = user.name;
+      serviceEntity.description = user.description;
+      serviceEntity.category = user.category;
+      serviceEntity.previewPictureResource = user.previewPictureResource;
+      serviceEntity.mediumPictureResource = user.mediumPictureResource;
+    }
+
+    return serviceEntity;
+  },
```
e-maad, line 107:
```suggestion
    const serviceEntity = new ServiceEntity();
    return {...serviceEntity, ...user};
  },
```

apps/webapp/src/script/repositories/integration/IntegrationRepository.ts
```diff
@@ -101,6 +101,10 @@ export class IntegrationRepository {
     return this.getServiceById(providerId, serviceId, entity.qualifiedId.domain);
   }
 
+  mapServiceFromUser(user: User): ServiceEntity {
+    return IntegrationMapper.mapServiceFromUser(user);
+  }
```
e-maad, line 106: Why do we need a function that just calls another function?

apps/webapp/src/script/repositories/integration/ServiceEntity.ts
```diff
@@ -43,17 +44,27 @@ export class ServiceEntity {
...
+    } = serviceData;
 
     this.id = id;
     this.description = description;
     this.name = ko.observable(name);
     this.providerId = providerId;
     this.providerName = ko.observable(' ');
     this.summary = summary;
+    this.category = category;
     this.tags = tags;
```
e-maad: Maybe you can do something like:

```
this.id = serviceData.id ?? '';
this.description = serviceData.description ?? '';
```

e-maad (changes requested):

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
panuschek: Good catch. That got overlooked in the hurry.

apps/webapp/src/script/repositories/integration/IntegrationRepository.ts
panuschek: I didn't want to move IntegrationMapper up in scope. Right now it's only used in IntegrationRepository and addParticipants already has a reference to the repository so I thought this is cleaner even though it's just another function call. I'm open to suggestions though on how to make it prettier!

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
panuschek: I know this doesn't look nice but I haven't found another way to do it honestly.
The filter and sort logic from line 132 & 133 is directly taken from the logic in searchForServices() in the IntegrationRepository.

I do need the other filter before that though since new Apps are also just Users and I have to filter out which users are actually apps before mapping them to ServiceEntity.

Also open for suggestions on how to make this better but given the time this is the nicest I could come up with.

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
panuschek: Can you elaborate on how I should use `is` here?
servicesList is always of type ServiceEntity[] and I just want to check that length is not 0

apps/webapp/src/script/repositories/integration/IntegrationMapper.ts
panuschek: This breaks the code on the calling side.
I could suggest

```
mapServiceFromUser: (user: User): ServiceEntity => {
  const serviceEntity = new ServiceEntity();

  if (user.type === UserType.APP) {
    const { 
      id, name, description, category, 
      previewPictureResource, mediumPictureResource 
    } = user;

    Object.assign(serviceEntity, { 
      id, name, description, category, 
      previewPictureResource, mediumPictureResource 
    });
  }

  return serviceEntity;
};
```
instead?

apps/webapp/src/script/repositories/integration/IntegrationRepository.ts
e-maad: Make sense.

apps/webapp/src/script/repositories/integration/IntegrationMapper.ts
e-maad: A bit cleaner:
```
mapServiceFromUser: (user: User) => {
    const serviceEntity = new ServiceEntity();
    if (user.type === UserType.APP) { 
        return {...serviceEntity, ...user};
     }

    return serviceEntity;
  },
```

apps/webapp/src/script/page/RightSidebar/addParticipants/addParticipants.tsx
e-maad:
```suggestion
              { is.nonEmptyArray(servicesList) && (
```

apps/webapp/src/script/repositories/integration/IntegrationMapper.ts
panuschek: I think I will keep it without the spreading operation since this
a) spreads all fields of the user into the new object (which I don't need) and
b) strips ServiceEntity of all its methods (I'm using getters in my next PR and am not quite sure they would survive the spread)

We can discuss this part again after the deadline is met I'd say :)

PR #21039, opened 2026-04-17
fix:  Messages that contain assets do not make the asset content available to the search engine(WPB-22078)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22078" title="WPB-22078" target="_blank">BugWPB-22078</a>  [Web] Messages that contain assets do not make the asset content available to the search engine
  </td></table>
  <br />
 

## Summary

- User should be able to search all text messages with and without files.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/repositories/conversation/ConversationService.ts
+++ b/apps/webapp/src/script/repositories/conversation/ConversationService.ts
@@ -48,11 +48,13 @@ import {MLSServiceEvents} from '@wireapp/core/lib/messagingProtocols/mls';
 import {container} from 'tsyringe';
 
 import type {Conversation as ConversationEntity} from 'Repositories/entity/Conversation';
+import {ClientEvent} from 'Repositories/event/Client';
 import type {EventService} from 'Repositories/event/EventService';
 import {search as fullTextSearch} from 'Repositories/search/FullTextSearch';
 import {StorageService} from 'Repositories/storage';
 import {ConversationRecord} from 'Repositories/storage/record/ConversationRecord';
 import {StorageSchemata} from 'Repositories/storage/StorageSchemata';
+import {getLogger} from 'Util/logger';
 
 import {MLSCapableConversation} from './ConversationSelectors';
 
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/repositories/conversation/ConversationService.ts
```diff
@@ -48,23 +48,62 @@ import {MLSServiceEvents} from '@wireapp/core/lib/messagingProtocols/mls';
...
 import {StorageSchemata} from 'Repositories/storage/StorageSchemata';
+import {getLogger} from 'Util/logger';
 
 import {MLSCapableConversation} from './ConversationSelectors';
 
 import {MessageCategory} from '../../message/MessageCategory';
 import {APIClient} from '../../service/APIClientSingleton';
 import {Core} from '../../service/CoreSingleton';
 
+const logger = getLogger('AccountForm');
```
zskhan: This logger name is duplicate from this file
apps/webapp/src/script/auth/component/AccountForm.tsx

arjita-mitra: oops! good catch :-)

apps/webapp/src/script/repositories/conversation/ConversationService.ts
```diff
@@ -48,23 +48,62 @@ import {MLSServiceEvents} from '@wireapp/core/lib/messagingProtocols/mls';
...
+        if (item?.text) {
+          return [item.text.content || item.text.message || ''];
+        }
+        if (item?.button) {
+          return [item.button.text || ''];
+        }
+        return [];
+      })
+      .filter((text: string) => text.length > 0)
+      .join(' ');
```
zskhan, line 98: It can return falsy value even when item.text.message exist but item.text.content is empty string.
Maybe a little bit better version.

```suggestion
    return items
  .flatMap(item => [
    item?.text?.content ?? item?.text?.message,
    item?.button?.text,
  ])
  .filter((v): v is string => typeof v === 'string' && v.length > 0)
  .join(' ');
```

zskhan: Misunderstood it

zskhan (approved):

PR #21048, opened 2026-04-17
fix: Add monotonic timestamp handling and adjust WebGL pipeline logging [WPB-24373]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24373" title="WPB-24373" target="_blank">StoryWPB-24373</a>  [Web] Firefox MediaPipe segmentation crash due to timestamp mismatch (norm_rect stream)
  </td></table>
  <br />
 

## Summary

 - fix MediaPipe segmentation failures caused by non-monotonic timestamps  in Firefox
 - temporary deactivate webgl segmentation worker for Firefox
---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/pipelines/mainWebGlPipeline.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/pipelines/mainWebGlPipeline.ts
@@ -69,6 +69,7 @@ export class MainWebGlPipeline implements BackgroundEffectsRenderingPipeline {
   private segmenter: SegmenterLike | null = null;
   private segmenterFactory: SegmenterFactory = MediaPipeSegmenterFactory;
   private segmentationModelByTier: SegmentationModelByTier = {};
+  private lastSegmentationTimestampMs = -1;
   private currentModelPath: string | null = null;
   private maskPostProcessor: MaskPostProcessor = new NoopMaskPostProcessor();
   private qualityController: QualityController | null = null;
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/repositories/media/backgroundEffects/pipelines/mainWebGlPipeline.ts
```diff
@@ -69,6 +69,7 @@ export class MainWebGlPipeline implements BackgroundEffectsRenderingPipeline {
   private segmenter: SegmenterLike | null = null;
   private segmenterFactory: SegmenterFactory = MediaPipeSegmenterFactory;
   private segmentationModelByTier: SegmentationModelByTier = {};
+  private lastSegmentationTimestampMs = -1;
```
arjita-mitra: Member 'lastSegmentationTimestampMs' is never reassigned; mark it as `readonly`. we could do that right?

apps/webapp/src/script/repositories/media/backgroundEffects/pipelines/mainWebGlPipeline.ts
```diff
@@ -203,7 +204,16 @@ export class MainWebGlPipeline implements BackgroundEffectsRenderingPipeline {
...
+          this.logger.warn('Frame due to non-monotonic timestamp', {
+            timestampMs,
+            lastTimestampMs: this.lastSegmentationTimestampMs,
+            rawTimestamp: timestamp,
+          });
+        }
+
         const includeClassMask = this.config.debugMode === 'classOverlay' || this.config.debugMode === 'classOnly';
         const result = await this.segmenter.segment(frame, timestampMs, {includeClassMask});
         segmentationMs = performance.now() - segStart;
```
zskhan, line 217: I think once the segmentation call is successful `lastSegmentationTimestampMs` should be updated to the latest timestamp.
`this.lastSegmentationTimestampMs = timestampMs;`

EnricoSchw: Really good find, but we no longer need `this.lastSegmentationTimestampMs`. It was supposed to fix a race condition that we don't have. It was part of my initial approach. I will remove `lastSegmentationTimestampMs` again.

apps/webapp/src/script/repositories/media/backgroundEffects/pipelines/mainWebGlPipeline.ts
EnricoSchw: Good catch: I will remove. See also my comment on Zafar's comment.

arjita-mitra (approved):

PR #21062, opened 2026-04-20
fix: hide filter tabs when feature flag is disabled [WPB-18861]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18861" title="WPB-18861" target="_blank">TaskWPB-18861</a>  [Web] add filter for unread messages, mentions and replies
  </td></table>
  <br />
 

## Summary

- What did I change and why?
Hide tabs from the left side bar when the feature flag of advance filters is disabled.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):


arjita-mitra (dismissed):

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/Conversations.tsx
```diff
@@ -330,6 +332,14 @@ export const Conversations = ({
     };
   }, [changeTab, conversationLabelRepository]);
 
+  const isAdvancedFiltersEnabled = Config.getConfig().FEATURE.ENABLE_ADVANCED_FILTERS;
+
+  useEffect(() => {
+    if (!isAdvancedFiltersEnabled) {
+      resetDisabledFeatureTabs();
+    }
+  }, [isAdvancedFiltersEnabled, resetDisabledFeatureTabs]);
```
screendriver, line 341: Just as a side note: `Config.getConfig()` is a **mutable** external singleton, not reactive React state, so reading it here and evaluating it in `useEffect()` is fragile. If `ENABLE_ADVANCED_FILTERS` changes without causing this component to rerender, the effect never runs. Even when it does re-render.

When this is fine for us, we can leave it as it is.

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.ts
```diff
@@ -45,6 +45,30 @@ export enum SidebarTabs {
...
+  SidebarTabs.MENTIONS,
+  SidebarTabs.REPLIES,
+  SidebarTabs.DRAFTS,
+  SidebarTabs.PINGS,
+];
+
+/**
+ * Tabs that should always be visible until user toggles it off.
+ */
+export const DEFAULT_TABS = [
```
screendriver: Why is this Array not readonly?

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
```diff
@@ -26,20 +26,7 @@ describe('useSidebarStore', () => {
...
-        SidebarTabs.CHANNELS,
-        SidebarTabs.DIRECTS,
-        SidebarTabs.UNREAD,
-        SidebarTabs.MENTIONS,
-        SidebarTabs.REPLIES,
-        SidebarTabs.DRAFTS,
-        SidebarTabs.PINGS,
-        SidebarTabs.ARCHIVES,
-      ],
+      visibleTabs: [...DEFAULT_TABS],
```
screendriver: When `DEFAULT_TABS` never gets mutated (which is hopefully the case) we don't need a shallow clone. So the spreading is not needed.

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
```diff
@@ -85,4 +72,53 @@ describe('useSidebarStore', () => {
...
+  it('falls back to RECENT when current tab is a filter tab during reset', () => {
+    useSidebarStore.setState({
+      currentTab: SidebarTabs.UNREAD,
+      visibleTabs: [SidebarTabs.RECENT, SidebarTabs.FOLDER, SidebarTabs.UNREAD],
+    });
+
+    useSidebarStore.getState().resetDisabledFeatureTabs();
+
+    expect(useSidebarStore.getState().currentTab).toBe(SidebarTabs.RECENT);
+    expect(useSidebarStore.getState().visibleTabs).toEqual([...DEFAULT_TABS]);
```
screendriver:
```suggestion
    expect(useSidebarStore.getState().visibleTabs).toEqual(DEFAULT_TABS);
```

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
```diff
@@ -85,4 +72,53 @@ describe('useSidebarStore', () => {
...
+  });
+
+  it('does not include filter tabs by default', () => {
+    const {visibleTabs} = useSidebarStore.getState();
+    const filterTabs = [...FILTER_TABS];
```
screendriver: We don't need the copy here at all, right?

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
```diff
@@ -85,4 +72,53 @@ describe('useSidebarStore', () => {
...
+  });
+
+  it('restores hidden default tabs during reset', () => {
+    useSidebarStore.setState({
+      visibleTabs: [SidebarTabs.RECENT, SidebarTabs.FOLDER],
+    });
+
+    useSidebarStore.getState().resetDisabledFeatureTabs();
+
+    expect(useSidebarStore.getState().visibleTabs).toEqual([...DEFAULT_TABS]);
```
screendriver:
```suggestion
    expect(useSidebarStore.getState().visibleTabs).toEqual(DEFAULT_TABS);
```

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/Conversations.tsx
zskhan: Exactly.
It will immediately hide it on next re-render but it would be not because of the toggle.
As soon as this feature is tested and ready to be release we are going to remove this so I think it is fine for now.

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
zskhan: Hopefully it is not mutated anywhere but to avoid any accidental mutation it will cause trouble thats why I did the shallow clone.

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.ts
zskhan: Because the store state was not readonly before.
Fixed that on both places.

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
screendriver: > Hopefully it is not mutated anywhere

➡️ https://github.com/wireapp/wire-webapp/pull/21062#discussion_r3108808263

apps/webapp/src/script/page/LeftSidebar/panels/Conversations/useSidebarStore.test.ts
```diff
@@ -85,4 +72,53 @@ describe('useSidebarStore', () => {
...
+
+    expect(useSidebarStore.getState().currentTab).toBe(SidebarTabs.RECENT);
+    expect(useSidebarStore.getState().visibleTabs).toEqual(DEFAULT_TABS);
+  });
+
+  it('does not include filter tabs by default', () => {
+    const {visibleTabs} = useSidebarStore.getState();
+    const filterTabs = FILTER_TABS;
+
+    filterTabs.forEach(tab => {
```
screendriver, line 112: This loop never executes when `filterTabs` is an empty Array. Which means the test gets green because of no single assertion is executed.

Even when this is not possible theoretically, the test here when looked at in isolation does and should not know about it. So I would like to see an assertion here that the Array has at least one element in it.

zskhan: https://github.com/wireapp/wire-webapp/pull/21062/commits/3f39ee5bbe36c7467fe76d6769551fd58b7e2704

screendriver (approved):

PR #21070, opened 2026-04-20
test: write the calling regression tests - the first part [WPB-19943]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19943" title="WPB-19943" target="_blank">TaskWPB-19943</a>  [Web/QA] Write the calling regression tests in Playwright
  </td></table>
  <br />
 

## Summary

This PR introduces the first phase of the Calling regression E2E test suite. To ensure a manageable review process, the full suite has been split into two logical parts.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
@@ -28,7 +28,7 @@ export const FullScreenCallPage = (page: Page) => {
   const raiseHandButton = component.getByTestId('do-toggle-hand-raise');
   const toggleParticipantsBtn = component.getByTestId('do-toggle-call-participants-list');
   const selfVideoThumbnail = page.getByTestId('self-video-thumbnail-wrapper');
-  const callParticipants = component.getByTestId('list-call-ui-participants').getByRole('listitem');
+  const participantListItems = component.getByTestId('list-call-ui-participants').getByRole('listitem');
   const gridTiles = component.getByTestId('item-grid');
 
   /* Press the react button and click the given emoji within the opened toolbar */
@@ -58,16 +58,17 @@ export const FullScreenCallPage = (page: Page) => {
     await toggleParticipantsBtn.click();
   };
 
-  const getCallParticipant = (userName: string) => {
-    const user = callParticipants.filter({hasText: userName});
+  // Sidebar list of participants in the call
+  const getSidebarParticipant = (userName: string) => {
+    const participant = participantListItems.filter({hasText: userName});
 
-    return Object.assign(user, {
-      muteIcon: user.getByTestId('status-audio-off'),
-      guestIcon: user.getByTestId('status-guest'),
+    return Object.assign(participant, {
+      muteIcon: participant.getByTestId('status-audio-off'),
+      guestIcon: participant.getByTestId('status-guest'),
       // User is an active speaker
-      speakIcon: user.getByTestId('status-active-speaking'),
-      videoIcon: user.getByTestId('status-video'),
-      screenShareIcon: user.getByTestId('status-screenshare'),
+      speakIcon: participant.getByTestId('status-active-speaking'),
+      videoIcon: participant.getByTestId('status-video'),
+      screenShareIcon: participant.getByTestId('status-screenshare'),
     });
   };
 
@@ -88,7 +89,7 @@ export const FullScreenCallPage = (page: Page) => {
     toggleHandRaise,
     selfVideoThumbnail,
     toggleParticipantsList,
-    getCallParticipant,
+    getSidebarParticipant,
     getGridTile,
   };
 };
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
@@ -25,6 +25,8 @@ export const FullScreenCallPage = (page: Page) => {
   const component = page.locator('.video-calling-wrapper');
 
   const reactButton = component.getByTitle('Reactions');
+  const raiseHandButton = component.getByTestId('do-toggle-hand-raise');
+  const selfVideoThumbnail = page.getByTestId('self-video-thumbnail-wrapper');
 
   /* Press the react button and click the given emoji within the opened toolbar */
   const sendReaction = async (emoji: '👍') => {
@@ -45,8 +47,14 @@ export const FullScreenCallPage = (page: Page) => {
     });
   };
 
+  const toggleHandRaise = async () => {
+    await raiseHandButton.click();
+  };
+
   return {
     sendReaction,
     getReaction,
+    toggleHandRaise,
+    selfVideoThumbnail,
   };
 };
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
@@ -26,7 +26,10 @@ export const FullScreenCallPage = (page: Page) => {
 
   const reactButton = component.getByTitle('Reactions');
   const raiseHandButton = component.getByTestId('do-toggle-hand-raise');
+  const toggleParticipantsBtn = component.getByTestId('do-toggle-call-participants-list');
   const selfVideoThumbnail = page.getByTestId('self-video-thumbnail-wrapper');
+  const callParticipants = component.getByTestId('list-call-ui-participants').getByRole('listitem');
+  const gridTiles = component.getByTestId('item-grid');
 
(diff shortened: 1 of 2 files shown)


apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -65,6 +71,43 @@ test.describe('Calling', () => {
...
+    await expect(userBPages.calling().callCell).toBeVisible();
+    await userBPages.calling().clickAcceptCallButton();
+
+    await expect(userBPages.calling().goFullScreen).toBeVisible();
+
+    const userACall = await userAPages.calling().maximizeCell();
+    await expect(userACall.selfVideoThumbnail).toBeVisible();
+
+    // Wait for the call to stabilize before testing the toast and thumbnail indicators
+    await userAPage.waitForTimeout(2000);
```
markbrockhoff, line 98: If this is really necessary I think it should be treated as a bug. Because if the user can see the raise hand button and click it it should work. Otherwise it should be disabled until it is working.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+    await expect(userBPages.calling().callCell).toBeVisible();
+
+    await userBPages.calling().clickAcceptCallButton();
+    // Confirm the calls grid is visible
+    await expect(userBPages.calling().goFullScreen).toBeVisible();
+
+    // User C joins the ongoing call
+    await userCPage.waitForTimeout(5000);
+    await expect(userCPages.conversationList().getConversationLocator(groupName).joinCallButton).toBeVisible();
+    await userCPages.conversationList().getConversationLocator(groupName).joinCallButton.click();
```
markbrockhoff:
```suggestion
    await userCPages.conversationList().getConversationLocator(groupName).joinCallButton.click({timeout: 30_000});
```

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+
+    await userAPages.conversationList().openConversation(groupName);
+    await userAPages.conversation().clickCallButton();
+
+    await expect(userAPages.calling().callCell).toBeVisible();
+    await expect(userBPages.calling().callCell).toBeVisible();
+
+    await userBPages.calling().clickAcceptCallButton();
+    // Confirm the calls grid is visible
+    await expect(userBPages.calling().goFullScreen).toBeVisible();
```
markbrockhoff, line 247: I think it might be good to also assert the number of tiles before C joins, what do you think?

```suggestion
    await expect(userBPages.calling().goFullScreen).toBeVisible();
    await expect(userBPages.calling().gridTiles).toHaveCount(2);
```

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+    await userBPages.calling().clickAcceptCallButton();
+    await expect(userBPages.calling().goFullScreen).toBeVisible();
+
+    // User B leaves the group call
+    await userBPages.calling().clickLeaveCallButton();
+    await expect(userBPages.calling().callCell).toBeHidden();
+
+    // User B re-joins the ongoing call from the conversation list
+    const joinButton = userBPages.conversationList().getConversationLocator(groupName).joinCallButton;
+    await expect(joinButton).toBeVisible({timeout: 5000});
```
markbrockhoff: Did you intentionally lower the timeout for this assertion from the default of 10s?

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+        await expect(userBPages.calling().goFullScreen).toBeVisible();
+      });
+
+      await test.step('User A joins then leaves the call from their second device', async () => {
+        await userADevice2Pages.conversationList().openConversation(groupName);
+        await userADevice2Pages.conversation().clickCallButton();
+        await expect(userADevice2Pages.calling().callCell).toBeVisible();
+
+        await userAPage2.waitForTimeout(30_000);
+        await expect(userADevice2Pages.calling().gridTiles).toHaveCount(3);
```
markbrockhoff: Is this 30s wait placed here intentionally or just because the default timeout of `toHaveCount` wasn't sufficient? (If so please just define the timeout on `toHaveCount`)

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+      });
+
+      await test.step('Verify User A’s first device remains in the call throughout', async () => {
+        await expect(userAPages.calling().goFullScreen).toBeVisible();
+
+        await userAPages.calling().clickLeaveCallButton();
+        await expect(userAPages.calling().callCell).not.toBeAttached();
+        // Verify the call is still "joinable" (User B is still there)
+        await expect(userAPages.conversationList().getConversationLocator(groupName).joinCallButton).toBeVisible({
+          timeout: 1000,
```
markbrockhoff: 1s is a really tight timeout, what's the reason for it? I'm a bit worried that it could introduce flake if the join call button takes a moment to appear

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+
+  test(
+    'I want to see warning when calling a big group',
+    {tag: ['@TC-2842', '@regression']},
+    async ({createPage, createUser}) => {
+      // Create Users and add to team
+      const extraMembers = await Promise.all(Array.from({length: 4}, () => createUser()));
+
+      for (const member of extraMembers) {
+        team.addTeamMember(member);
```
markbrockhoff: `addTeamMember()` returns a promise ;)

```suggestion
        await team.addTeamMember(member);
```

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -164,4 +225,334 @@ test.describe('Calling', () => {
...
+          },
+        },
+      ];
+
+      for (const participant of participants) {
+        for (const feature of features) {
+          await test.step(`${participant.label}: Verify ${feature.name}`, async () => {
+            await feature.action(participant.pages);
+            await feature.verify(participant.name);
+            await feature.action(participant.pages);
```
markbrockhoff: Beautiful, but I think you don't even need the second call to `.action()` since the toggles don't interfere with each other. (Doesn't hurt though)

apps/webapp/test/e2e_tests/pageManager/webapp/pages/fullScreenCall.page.ts
```diff
@@ -45,8 +50,45 @@ export const FullScreenCallPage = (page: Page) => {
...
 
+  const toggleHandRaise = async () => {
+    await raiseHandButton.click();
+  };
+
+  const toggleParticipantsList = async () => {
+    await toggleParticipantsBtn.click();
+  };
+
+  const getCallParticipant = (userName: string) => {
```
markbrockhoff: Could we maybe name this a bit different or add a JSDoc comment to it explaining that this is for the participant list while `getGridTile()` is for the tiles? At least for me it wasn't clear immediately that this one is for the participants list.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
zhannach: Without this, the first notification ('You have raised your hand up') fails to appear. However, all subsequent notifications display as expected.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
zhannach: I recall why I didn't check it - in a group call with 2 participants, there is only 1 grid-item (the other person) and 1 self-video-thumbnail

markbrockhoff: Ah ok, I see, yeah asserting for one tile would be confusing then

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
markbrockhoff: Yeah let's see, if a bug is created to investigate this please add its ticket number as a comment ;)

markbrockhoff (approved):

screendriver (approved):

PR #21073, opened 2026-04-20
fix: replace close button with x and update performance panel styles [WPB-24452]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24452" title="WPB-24452" target="_blank">BugWPB-24452</a>  Close button for the performance pop-up should be replaced with 'X'.
  </td></table>
  <br />
 

## Summary

- What did I change and why?
Replace close button with x
Update performance panel UI and make it compatible with dark mode.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
+++ b/apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
@@ -19,6 +19,8 @@
 
 import {ReactNode, useCallback, useEffect, useMemo, useState} from 'react';
 
+import {Maybe} from 'true-myth';
+
 import {Button, ButtonVariant, CloseIcon, Option, Select} from '@wireapp/react-ui-kit';
 
 import {
@@ -39,7 +41,7 @@ import {
 import {QualityMode} from 'Repositories/media/backgroundEffects';
 import {CapabilityInfo} from 'Repositories/media/backgroundEffects/backgroundEffectsWorkerTypes';
 import type {BackgroundEffectsHandler} from 'Repositories/media/backgroundEffectsHandler';
-import {useBackgroundEffectsStore} from 'Repositories/media/useBackgroundEffectsStore';
+import {RenderMetrics, useBackgroundEffectsStore} from 'Repositories/media/useBackgroundEffectsStore';
 import {t} from 'Util/localizerUtil';
 
 type PerformancePanelProps = {
@@ -60,9 +62,45 @@ const formatValue = (value?: string | number | null): string => {
   return value === null || value === undefined || value === '' ? '-' : String(value);
 };
 
+const getMetricRows = (renderMetrics: RenderMetrics) => {
+  return Maybe.of(renderMetrics)
+    .map(metrics => [
+      {label: 'Quality', value: formatValue(metrics.tier)},
+      {label: 'Total', value: formatMs(metrics.avgTotalMs)},
+      {label: 'Segmentation', value: formatMs(metrics.avgSegmentationMs)},
+      {label: 'GPU', value: formatMs(metrics.avgGpuMs)},
+      {label: 'Budget', value: formatMs(metrics.budget)},
+      {label: 'ML delegate type', value: formatValue(metrics.ml)},
+      {label: 'Utilization', value: formatPercent(metrics.utilShare)},
+      {label: 'ML', value: formatPercent(metrics.mlShare)},
+      {label: 'WebGL', value: formatPercent(metrics.webglShare)},
+      {
+        label: 'Delegate',
+        value: formatValue(metrics.segmentationDelegate),
+      },
+      {label: 'Dropped', value: formatValue(metrics.droppedFrames)},
+    ])
+    .unwrapOr([]);
+};
+
+const getCapabilityRows = (capabilityInfo: CapabilityInfo | null | undefined) => {
+  return Maybe.of(capabilityInfo)
+    .map(info => [
+      {label: 'WebGL2', value: info.webgl2 ? '✔' : '✖'},
+      {label: 'Worker', value: info.worker ? '✔' : '✖'},
+      {label: 'OffscreenCanvas', value: info.offscreenCanvas ? '✔' : '✖'},
+      {label: 'VideoFrameCallback', value: info.requestVideoFrameCallback ? '✔' : '✖'},
+    ])
+    .unwrapOr([]);
+};
+
 type MetricRowProps = {
   label: string;
   value: ReactNode;
+};
+
+type MetricsDisplayProps = {
+  readonly capabilityInfo: CapabilityInfo;
 };
 
 const MetricRow = ({label, value}: MetricRowProps) => (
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -106,67 +174,33 @@ export const VideoBackgroundPerformancePanel = ({backgroundEffectsHandler}: Perf
...
+    } else {
+      handleOpenPanel();
+    }
+  }, [handleClosePanel, handleOpenPanel, isPanelOpen]);
+
   const handleQualityChange = useCallback(
-    (event: ChangeEvent<HTMLSelectElement>) => {
-      const nextQuality = event.target.value as QualityMode;
+    (quality: Option) => {
+      if (!quality) {
```
screendriver: Not `undefined`? Not `null`? We should always try to express the intention explicitly

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -68,15 +74,77 @@ const MetricRow = ({label, value}: MetricRowProps) => (
...
+    () =>
+      QUALITY_OPTIONS.map(option => ({
+        label: option,
+        value: option,
+      })),
+    [],
+  );
+
+  const selectedOption = useMemo(
+    () => qualitySelectOptions.find(option => option.value === selectedQuality) ?? null,
```
screendriver, line 153: A `Maybe` would fit better here. You could also use [find](https://true-myth.js.org/api/maybe/functions/find.html) to make it even more safe.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -68,15 +74,77 @@ const MetricRow = ({label, value}: MetricRowProps) => (
...
 const POLLING_INTERVAL = 500;
 
-export const VideoBackgroundPerformancePanel = ({backgroundEffectsHandler}: PerformancePanelProps) => {
-  const isFeatureEnabled = useBackgroundEffectsStore(state => state.isFeatureEnabled);
+const MetricsDisplay = ({capabilityInfo}: {capabilityInfo: CapabilityInfo | null}) => {
   const renderMetrics = useBackgroundEffectsStore(state => state.metrics);
   const model = useBackgroundEffectsStore(state => state.model);
 
+  const metricRows = useMemo(() => {
+    if (!renderMetrics) {
```
screendriver: Same as in https://github.com/wireapp/wire-webapp/pull/21073/changes#r3115315326

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -68,15 +74,77 @@ const MetricRow = ({label, value}: MetricRowProps) => (
 
 const POLLING_INTERVAL = 500;
 
-export const VideoBackgroundPerformancePanel = ({backgroundEffectsHandler}: PerformancePanelProps) => {
-  const isFeatureEnabled = useBackgroundEffectsStore(state => state.isFeatureEnabled);
+const MetricsDisplay = ({capabilityInfo}: {capabilityInfo: CapabilityInfo | null}) => {
```
screendriver:
```suggestion
type MetricsDisplayProperties = {
  readonly capabilityInfo: Maybe<CapabilityInfo>;
}

const MetricsDisplay: FunctionComponent< MetricsDisplayProperties> = (properties) => {
```

e-maad: @screendriver, any reason to use `FunctionComponent`?
Why not: 
`const MetricsDisplay = (properties: MetricsDisplayProperties) => {`

I see this approach much more cleaner.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -68,15 +74,77 @@ const MetricRow = ({label, value}: MetricRowProps) => (
...
+      {label: 'Total', value: formatMs(renderMetrics.avgTotalMs)},
+      {label: 'Segmentation', value: formatMs(renderMetrics.avgSegmentationMs)},
+      {label: 'GPU', value: formatMs(renderMetrics.avgGpuMs)},
+      {label: 'Budget', value: formatMs(renderMetrics.budget)},
+      {label: 'ML delegate type', value: formatValue(renderMetrics.ml)},
+      {label: 'Utilization', value: formatPercent(renderMetrics.utilShare)},
+      {label: 'ML', value: formatPercent(renderMetrics.mlShare)},
+      {label: 'WebGL', value: formatPercent(renderMetrics.webglShare)},
+      {label: 'Delegate', value: formatValue(renderMetrics.segmentationDelegate)},
+      {label: 'Dropped', value: formatValue(renderMetrics.droppedFrames)},
```
e-maad: Can we move it outside of component?

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -68,15 +74,77 @@ const MetricRow = ({label, value}: MetricRowProps) => (
...
+  const capabilityRows = useMemo(() => {
+    if (!capabilityInfo) {
+      return [];
+    }
+
+    return [
+      {label: 'WebGL2', value: capabilityInfo.webgl2 ? '✔' : '✖'},
+      {label: 'Worker', value: capabilityInfo.worker ? '✔' : '✖'},
+      {label: 'OffscreenCanvas', value: capabilityInfo.offscreenCanvas ? '✔' : '✖'},
+      {label: 'VideoFrameCallback', value: capabilityInfo.requestVideoFrameCallback ? '✔' : '✖'},
```
e-maad: Same as: https://github.com/wireapp/wire-webapp/pull/21073/changes#r3115787536

e-maad (commented):
Added few suggestions 🚀

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
zskhan: So this selectedOption is used in the React Select and if use `Maybe` here then the value={selectedOption} will not be happy.
Maybe will return `Just<Option> | Nothing` whereas value is looking for `Option | null`

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
screendriver: Because you are missing an explicit return type here. Now it could be any kind of function, disconnected from React itself without any compile time safety.

thisisamir98 (approved):

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
zskhan: But TypeScript infers it as JSX.Element so it’s still type-safe and tied to React.
Why explicit type is needed here when there is no other benefit of it?

PR #21074, opened 2026-04-20
fix: add e2e test for new appLock pass flow [WPB-23186]

e-maad:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Added e2e tests for appLock new flow added here: https://github.com/wireapp/wire-webapp/pull/21025
Screenshot 2026-04-21 at 10 32 58

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
@@ -149,7 +149,7 @@ test.describe('AppLock', () => {
 
         await page.reload();
         await modals.appLock().clickForgotPassphrase();
-        await modals.appLock().clickWipeDB();
+        await modals.appLock().clickProceedToLogout();
         await modals.appLock().checkClearData();
         await modals.appLock().clickReset();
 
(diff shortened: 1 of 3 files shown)


apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
```diff
@@ -172,11 +172,37 @@ test.describe('AppLock', () => {
         await page.reload();
         await modals.appLock().clickForgotPassphrase();
         await modals.appLock().clickWipeDB();
```
markbrockhoff: Does this util maybe need to be renamed? Because we're now calling `clickWipeDB()` but then assert the DB to not be wiped ^^'

apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
```diff
@@ -172,11 +172,37 @@ test.describe('AppLock', () => {
...
-        // The modal should show an error for the invalid password and not wipe indexDB
-        await expect(modals.appLock().errorMessage).toContainText('Wrong password');
+        // The database should not be wiped when the checkbox was not checked
+        await expect.poll(() => page.evaluate(() => indexedDB.databases())).not.toHaveLength(0);
+      },
+    );
+
+    test(
+      'Web: I should be able to login and recover my existing session after logout without wiping data',
+      {tag: ['@TC-2764', '@regression']},
```
markbrockhoff: This Test ID doesn't exist within Testiny, did you already inform @iskvortsov about the new behavior so the test cases can be updated accordingly?

apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
```diff
@@ -172,11 +172,37 @@ test.describe('AppLock', () => {
...
+        await modals.appLock().clickWipeDB();
+        // Intentionally do not check the clear data checkbox
+        await modals.appLock().clickReset();
+
+        // Log back in with the same user and verify the existing session is intact
+        await loginUser(memberA, pageManager);
+
+        // App lock setup modal should appear again (passcode was cleared on logout, session data is preserved)
+        await expect(modals.appLock().appLockModal).toBeVisible();
+        await expect(modals.appLock().lockPasscodeInput).toBeVisible();
```
markbrockhoff, line 186: Isn't this test identical to TC-2763 except that the user logs in again after the reset? Do you think it could make sense to just keep them as one test or is there a specific reason to test the re-login individually?

apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
e-maad: The plan was to ask @iskvortsov for the review and create this test ID during the process. Let's see if we want this test or not on the first place.

apps/webapp/test/e2e_tests/specs/AppLock/AppLock.spec.ts
e-maad: @iskvortsov, I updated @TC-2763, can you please update it in Testiny?

markbrockhoff (approved):

thisisamir98 (approved):

PR #21081, opened 2026-04-21
feat: app details view (WPB-21440)

panuschek:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21440" title="WPB-21440" target="_blank">StoryWPB-21440</a>  [Web] [Integration] Toggle Apps in Conversation Flow
  </td></table>
  <br />
 

## Summary

- What did I change and why?

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
+++ b/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
@@ -21,6 +21,8 @@ import {forwardRef, useEffect, useMemo, useState} from 'react';
 
 import {CONVERSATION_ACCESS, CONVERSATION_CELLS_STATE} from '@wireapp/api-client/lib/conversation';
 import {RECEIPT_MODE} from '@wireapp/api-client/lib/conversation/data/';
+import {CONVERSATION_PROTOCOL} from '@wireapp/api-client/lib/team';
+import {UserType} from '@wireapp/api-client/lib/user';
 
 import {TabIndex} from '@wireapp/react-ui-kit';
 
@@ -179,7 +181,7 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
 
     const userParticipants = useMemo(() => {
       const filteredUsers: User[] = participatingUserEts.flatMap(user => {
-        const isUser = !isServiceEntity(user);
+        const isUser = user.type === UserType.REGULAR;
         return isUser ? [user] : [];
       });
 
@@ -194,10 +196,25 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
     const exceedsMaxUserCount = usersCount > CONFIG.MAX_USERS_VISIBLE;
     const allUsersCount = exceedsMaxUserCount ? usersCount : 0;
 
-    const serviceParticipants: ServiceEntity[] = participatingUserEts.flatMap(service => {
-      const isService = isServiceEntity(service);
-      return isService ? [service] : [];
-    });
+    const serviceParticipants: ServiceEntity[] = useMemo(() => {
+      const services = new Array<ServiceEntity>();
+
+      if (activeConversation.protocol === CONVERSATION_PROTOCOL.PROTEUS) {
+        participatingUserEts.forEach(service => {
+          if (isServiceEntity(service)) {
+            services.push(service);
+          }
+        });
+      }
+
+      participatingUserEts.forEach(user => {
+        if (user.type === UserType.APP) {
+          services.push(integrationRepository.mapServiceFromUser(user));
+        }
+      });
+
+      return services;
+    }, [activeConversation.protocol, integrationRepository, participatingUserEts]);
 
     const toggleMute = () => actionsViewModel.toggleMuteConversation(activeConversation);
 
@@ -209,7 +226,11 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
     const showUser = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
     const showService = async (entity: ServiceEntity) => {
-      const serviceEntity = await integrationRepository.getServiceFromUser(entity);
+      const serviceEntity = entity.isService
+        ? await integrationRepository.getServiceFromUser(entity)
+        : integrationRepository.mapServiceFromUser(
+            participatingUserEts.find(participant => participant.id === entity.id),
+          );
 
       if (serviceEntity) {
         togglePanel(PanelState.GROUP_PARTICIPANT_SERVICE, serviceEntity);
--- a/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
+++ b/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
@@ -222,13 +222,16 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
     const showUser = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
     const showService = async (entity: ServiceEntity) => {
-      const serviceEntity = entity.isService
-        ? await integrationRepository.getServiceFromUser(entity)
-        : integrationRepository.mapServiceFromUser(
-            participatingUserEts.find(participant => participant.id === entity.id),
-          );
+      if (entity.isService) {
+        const serviceEntity = await integrationRepository.getServiceFromUser(entity);
+        togglePanel(PanelState.GROUP_PARTICIPANT_SERVICE, serviceEntity);
+        return;
+      }
 
-      if (serviceEntity) {
+      const user = participatingUserEts.find(participant => participant.id === entity.id);
+
+      if (user) {
+        const serviceEntity = integrationRepository.mapServiceFromUser(user);
         togglePanel(PanelState.GROUP_PARTICIPANT_SERVICE, serviceEntity);
       }
     };
--- a/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
+++ b/apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
@@ -200,18 +200,14 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
       const services = new Array<ServiceEntity>();
 
       if (activeConversation.protocol === CONVERSATION_PROTOCOL.PROTEUS) {
-        participatingUserEts.forEach(service => {
-          if (isServiceEntity(service)) {
-            services.push(service);
-          }
-        });
+        services.push(...participatingUserEts.filter(service => isServiceEntity(service)));
+      } else {
+        services.push(
+          ...participatingUserEts
+            .filter(user => user.type === UserType.APP)
+            .map(user => integrationRepository.mapServiceFromUser(user)),
+        );
       }
-
-      participatingUserEts.forEach(user => {
-        if (user.type === UserType.APP) {
-          services.push(integrationRepository.mapServiceFromUser(user));
-        }
-      });
 
       return services;
     }, [activeConversation.protocol, integrationRepository, participatingUserEts]);
--- a/apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
+++ b/apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
@@ -29,6 +29,7 @@ import {Conversation} from 'Repositories/entity/Conversation';
 import {User} from 'Repositories/entity/User';
 import {IntegrationRepository} from 'Repositories/integration/IntegrationRepository';
 import {ServiceEntity} from 'Repositories/integration/ServiceEntity';
+import {TeamRepository} from 'Repositories/team/TeamRepository';
 import {generatePermissionHelpers} from 'Repositories/user/UserPermission';
 import {useKoSubscribableChildren} from 'Util/componentUtil';
 import {handleKeyDown, KEY} from 'Util/keyboardUtil';
@@ -42,6 +43,7 @@ interface GroupParticipantServiceProps {
   actionsViewModel: ActionsViewModel;
   integrationRepository: IntegrationRepository;
   conversationRepository: ConversationRepository;
+  teamRepository: TeamRepository;
   enableRemove: boolean;
   goToRoot: () => void;
   onBack: () => void;
@@ -56,6 +58,7 @@ const GroupParticipantService: FC<GroupParticipantServiceProps> = ({
   actionsViewModel,
   integrationRepository,
   conversationRepository,
+  teamRepository,
   enableRemove,
   goToRoot,
   onBack,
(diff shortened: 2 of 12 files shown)


apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -194,10 +196,25 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
...
-    });
+    const serviceParticipants: ServiceEntity[] = useMemo(() => {
+      const services = new Array<ServiceEntity>();
+
+      if (activeConversation.protocol === CONVERSATION_PROTOCOL.PROTEUS) {
+        participatingUserEts.forEach(service => {
+          if (isServiceEntity(service)) {
+            services.push(service);
+          }
+        });
```
e-maad: Can't you do something like:
```suggestion
         services.push(...participatingUserEts.filter(service => isServiceEntity(service));
```

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -194,10 +196,25 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
...
+      }
+
+      participatingUserEts.forEach(user => {
+        if (user.type === UserType.APP) {
+          services.push(integrationRepository.mapServiceFromUser(user));
+        }
+      });
```
e-maad: Same as: https://github.com/wireapp/wire-webapp/pull/21081/changes#r3118673848

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -209,7 +226,11 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
     const showUser = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
     const showService = async (entity: ServiceEntity) => {
-      const serviceEntity = await integrationRepository.getServiceFromUser(entity);
+      const serviceEntity = entity.isService
```
e-maad: Please be explicit. Are you expecting entity.isService to be null or undefined?
You can use `is.nullOrUndefined`.

apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
```diff
@@ -73,7 +76,12 @@ const GroupParticipantService: FC<GroupParticipantServiceProps> = ({
 
   const {canChatWithServices} = generatePermissionHelpers(teamRole);
 
-  const serviceUser = participatingUserEts.find(user => user.serviceId === serviceEntity.id);
+  const byServiceId = (id: string) => (user: User) => user.serviceId === id;
+  const byUserId = (id: string) => (user: User) => user.id === id;
+  const serviceUser = participatingUserEts.find(user => {
+    const predicate = serviceEntity.isService ? byServiceId(serviceEntity.id) : byUserId(serviceEntity.id);
+    return predicate(user);
+  });
```
e-maad, line 86: Why to create predicate in every iteration?

```suggestion
const predicate = serviceEntity.isService
  ? byServiceId(serviceEntity.id)
  : byUserId(serviceEntity.id);

const serviceUser = participatingUserEts.find(predicate);
```

apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
```diff
@@ -99,6 +107,12 @@ const GroupParticipantService: FC<GroupParticipantServiceProps> = ({
     integrationRepository.addProviderNameToParticipant(serviceEntity);
   }, [integrationRepository, serviceEntity]);
 
+  useEffect(() => {
+    if (serviceUser?.teamId) {
```
e-maad: Please be explicit, https://github.com/wireapp/wire-webapp/pull/21081/changes#r3118682954

apps/webapp/src/script/repositories/integration/ServiceEntity.ts
```diff
@@ -33,6 +33,7 @@ export interface ServiceData {
   provider?: string;
   summary?: string;
   category?: string;
+  author?: string;
```
e-maad, line 36: I would suggest you to use `Maybe<string>` from `true-myth`.

apps/webapp/src/script/repositories/integration/ServiceEntity.ts
```diff
@@ -50,6 +51,7 @@ export class ServiceEntity {
   summary: string;
   tags: string[];
   category?: string;
+  author?: ko.Observable<string>;
```
e-maad, line 54: Same for https://github.com/wireapp/wire-webapp/pull/21081/changes#r3118733753

apps/webapp/src/script/view_model/ActionsViewModel.ts
```diff
@@ -370,7 +370,16 @@ export class ActionsViewModel {
...
       throw new Error();
     }
-    const conversationEntity = await this.integrationRepository.get1To1ConversationWithService(serviceEntity);
+
+    let conversationEntity;
+    if (serviceEntity.isService) {
+      conversationEntity = await this.integrationRepository.get1To1ConversationWithService(serviceEntity);
+    } else {
+      if (!serviceEntity.qualifiedId) {
+        throw new Error();
```
e-maad: Why are you just throwing an error? What is the intention?
Are we catching it up somewhere? Any logs?

apps/webapp/src/script/view_model/ActionsViewModel.ts
```diff
@@ -370,7 +370,16 @@ export class ActionsViewModel {
     if (!serviceEntity) {
       throw new Error();
     }
-    const conversationEntity = await this.integrationRepository.get1To1ConversationWithService(serviceEntity);
+
+    let conversationEntity;
+    if (serviceEntity.isService) {
+      conversationEntity = await this.integrationRepository.get1To1ConversationWithService(serviceEntity);
+    } else {
+      if (!serviceEntity.qualifiedId) {
```
e-maad: Same as: https://github.com/wireapp/wire-webapp/pull/21081/changes#r3118722734

apps/webapp/src/script/view_model/ActionsViewModel.ts
```diff
@@ -370,7 +370,16 @@ export class ActionsViewModel {
     if (!serviceEntity) {
       throw new Error();
     }
-    const conversationEntity = await this.integrationRepository.get1To1ConversationWithService(serviceEntity);
+
+    let conversationEntity;
```
e-maad: What do you think about early return?

```suggestion
if (serviceEntity.isService)  { // check explicitly here for null / undefined
  return this.integrationRepository.get1To1ConversationWithService(serviceEntity);
}

if (!serviceEntity.qualifiedId) throw new Error(); // check explicitly here for null / undefined

return this.getOrCreate1to1Conversation({ qualifiedId: serviceEntity.qualifiedId });
```

e-maad (changes requested):
Added few comments

apps/webapp/src/script/view_model/ActionsViewModel.ts
panuschek: TBH I looked at the code that's already there and copied the error handling which was: Just throwing a new Error().
Seemed to work until now so I didn't change it.

e-maad: I would suggest to spend some time and see if we can improve the existing approach.

panuschek: I would agree with that if it wasn't 3 days before a release where this feature is expected to be in.

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
panuschek: I expect it to be true or false because isService is a boolean value.
If serviceEntity is an app and not a service I need a full User entity to map all of its fields correctly.

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -194,10 +196,25 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
...
+
+      if (activeConversation.protocol === CONVERSATION_PROTOCOL.PROTEUS) {
+        participatingUserEts.forEach(service => {
+          if (isServiceEntity(service)) {
+            services.push(service);
+          }
+        });
+      }
+
+      participatingUserEts.forEach(user => {
```
markbrockhoff: Could an entity match "isServiceEntity" **and** have type `UserType.APP`? Because currently if the protocol was proteus this could lead to an entity being added to the serviceParticipants twice. (Not sure though if it's a realistic case but the conditions don't clearly exclude each other)

If duplicates should be avoided using a reducer might be a safer solution, e.g.:
```ts
const serviceParticipants: ServiceEntity[] = useMemo(() => {
  return participatingUserEts.reduce<ServiceEntity[]>((acc, entity) => {
    if (activeConversation.protocol === CONVERSATION_PROTOCOL.PROTEUS && isServiceEntity(entity)) {
      acc.push(entity);
    } else if (entity.type === UserType.APP) {
      acc.push(integrationRepository.mapServiceFromUser(entity));
    }
    return acc;
  }, []);
}, [activeConversation.protocol, integrationRepository, participatingUserEts]);
```

The `else if` ensures no duplicates can occur and the reduce only iterates the list of userEts only once instead of twice (at the cost of one enum equality check) + it allocates less memory than using e.g. `.filter()` which would allocate two arrays just to have them garbage collected right away.

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -194,10 +196,25 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
...
+        if (user.type === UserType.APP) {
+          services.push(integrationRepository.mapServiceFromUser(user));
+        }
+      });
+
+      return services;
+    }, [activeConversation.protocol, integrationRepository, participatingUserEts]);
```
markbrockhoff, line 213: I don't think the `integrationRepository` is reactive right? So watching it should be useless I think

```suggestion
    }, [activeConversation.protocol, participatingUserEts]);
```

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
markbrockhoff: I think `isService` is of type `boolean` so IMHO checking for something known to not be `null` or `undefined` at build time is a waste of resources. Otherwise we wouldn't trust our type system and that would be a much bigger issue ^^'
Only suggestion I could see is to explicitly check for the value to be `entity.isService === true` to avoid the expression weakening should the type of it change some time in the future (unexpected for something called "is*" but who knows)

But TBH instead of doing that everywhere in the codebase I would probably rather set up a linter rule like e.g. https://typescript-eslint.io/rules/strict-boolean-expressions/ to avoid this without having to think of it manually.

apps/webapp/src/script/page/RightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -209,7 +226,11 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
     const showUser = (userEntity: User) => togglePanel(PanelState.GROUP_PARTICIPANT_USER, userEntity);
 
     const showService = async (entity: ServiceEntity) => {
-      const serviceEntity = await integrationRepository.getServiceFromUser(entity);
+      const serviceEntity = entity.isService
+        ? await integrationRepository.getServiceFromUser(entity)
+        : integrationRepository.mapServiceFromUser(
+            participatingUserEts.find(participant => participant.id === entity.id),
```
markbrockhoff: 🔧 Please note: `.find()` could always return `undefined`. The reason TS isn't catching this is because it's not configured to be this strict. But still I'd recommend to check for it in this case just to be safe :)

E.g. something like
```ts
const showService = async (entity: ServiceEntity) => {
  if (entity.isService) {
    const serviceEntity = await integrationRepository.getServiceFromUser(entity);
    togglePanel(PanelState.GROUP_PARTICIPANT_SERVICE, serviceEntity);
    return;
  }

  const user = participatingUserEts.find(participant => participant.id === entity.id);
  if (user) {
    const serviceEntity = integrationRepository.mapServiceFromUser(user);
    togglePanel(PanelState.GROUP_PARTICIPANT_SERVICE, serviceEntity);
  }
};
```

apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
markbrockhoff: Why even create the predicate in the first place? I assume for readability? But TBH I think the tradeoff of creating the predicate using curried functions only used in one place isn't worth it.

E.g. this is pretty readable I think (could also be a ternary, I don't have a strong preference there)

```suggestion
  const serviceUser = participatingUserEts.find(user => {
    if (serviceEntity.isService) {
      return user.serviceId === serviceEntity.id;
    } else {
      return user.id === serviceEntity.id;
    }
  });
```

apps/webapp/src/script/page/RightSidebar/groupParticipantService/groupParticipantService.tsx
```diff
@@ -99,6 +107,12 @@ const GroupParticipantService: FC<GroupParticipantServiceProps> = ({
...
+      teamRepository.getTeamNameById(serviceUser.teamId).then(name => serviceEntity.author(name));
```
markbrockhoff, line 114: Are you intentionally mutating the author of the serviceEntity which was passed as prop? This is typically seen as an anti pattern since it mutates the given object as a side effect. Looking at all those useEffect hooks in general maybe it could be better to create a clone from the serviceEntity with the edited fields and pass that to ServiceDetails?

> Sadly it was already done like this in this component so I understand that you're aligning with the existing style. But I at least wanted to leave this comment to explain that this IMHO isn't a good way of doing things and why

apps/webapp/src/script/repositories/integration/ServiceEntity.ts
markbrockhoff: Why though? This class is a data object so making the property required IMHO wouldn't give much value and most likely require quiet a lot of changes all over the place. Only changing the type for this one property and keeping the others optional would also be inconsistent. I think this change is out of scope for this PR and if done should be done as stand alone refactoring.

apps/webapp/src/script/view_model/ActionsViewModel.ts
markbrockhoff: Yeah I think catching an handling the error is out of scope here especially since there is already a nameless error in this function. But maybe adding a little name to it won't hurt e.g.:

```suggestion
        throw new Error("Can't create 1on1 conversation for an entity without qualifiedId");
```

e-maad: I would also suggest to have a `log` where we are catching it.

markbrockhoff: Why though? That would only swallow the error, if it happens it should be picked up by some reporting and not just logged and effectively be ignored.

e-maad: > Why though? That would only swallow the error, if it happens it should be picked up by some reporting and not just logged and effectively be ignored.

These logs will be reported to datadog where we can monitor.

apps/webapp/src/script/repositories/integration/ServiceEntity.ts
panuschek: A topic for future discussion

apps/webapp/src/script/view_model/ActionsViewModel.ts
panuschek: I decided to go with the "throw error with message" approach here since it's the closest behaviour to the original code. Let's definitely discuss error handling in general though asap.

PR #21096, opened 2026-04-22
ci: add realistic timeouts to Playwright workflows [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Because sometimes we have something like this and need to timeout at some point:

2026-04-22 13 29 33 github com e98abe0df643

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -63,6 +63,7 @@ jobs:
   test:
     name: Run E2E Tests
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -63,6 +63,7 @@ jobs:
   test:
     name: Run E2E Tests
     runs-on: ubuntu-24.04
+    timeout-minutes: 45
```
arjita-mitra, line 66: could you explain how did you get this  realistic timeouts data?

screendriver: Investigated the intermittent GitHub Actions hangs in the Playwright workflows (for the last 24 hours). The problem is the `playwright install --with-deps` step which is usually fast but can occasionally stall for much longer. So adding a 30-minute timeout to that install step and a 45-minute timeout to the surrounding Playwright jobs seemed okayish. The 45-minute job timeout is there as a safety cap for the full job, while the 30-minute limit fails the actual flaky step earlier.

arjita-mitra: sounds good!

e-maad (approved):

PR #21109, opened 2026-04-23
feat: add capability polling and move it behind feature flag [WPB-24520]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24520" title="WPB-24520" target="_blank">BugWPB-24520</a>  Issue with System Capabilities Not Updating in Panel
  </td></table>
  <br />
 

## Summary

- What did I change and why?
Moved capability polling and Quality polling behind the feature flag.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
+++ b/apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
@@ -23,6 +23,7 @@ import {Maybe} from 'true-myth';
 
 import {Button, ButtonVariant, CloseIcon, Option, Select} from '@wireapp/react-ui-kit';
 
+import {areCapabilityInfosEqual} from 'Components/calling/VideoControls/videoBackgroundPerformancePanel/capabilityInformationValidator';
 import {
   buttonBaseStyles,
   buttonNeutralStyles,
@@ -92,27 +93,6 @@ const getCapabilityRows = (capabilityInfo: CapabilityInfo | null | undefined) =>
       {label: 'VideoFrameCallback', value: info.requestVideoFrameCallback ? '✔' : '✖'},
     ])
     .unwrapOr([]);
-};
-
-const areCapabilityInfosEqual = (
-  prev: CapabilityInfo | null | undefined,
-  current: CapabilityInfo | null | undefined,
-): boolean => {
-  if (prev == null || current == null) {
-    return prev === current;
-  }
-
-  return Maybe.of(prev)
-    .andThen(prevValue =>
-      Maybe.of(current).map(
-        currentValue =>
-          prevValue.webgl2 === currentValue.webgl2 &&
-          prevValue.worker === currentValue.worker &&
-          prevValue.offscreenCanvas === currentValue.offscreenCanvas &&
-          prevValue.requestVideoFrameCallback === currentValue.requestVideoFrameCallback,
-      ),
-    )
-    .unwrapOr(false);
 };
 
 type MetricRowProps = {
@@ -206,7 +186,7 @@ export const VideoBackgroundPerformancePanel = ({backgroundEffectsHandler}: Perf
 
     const syncCapabilities = () => {
       const current = backgroundEffectsHandler.getCapabilityInfo();
-      setCapabilityInfo(prev => (areCapabilityInfosEqual(prev, current) ? prev : current));
+      setCapabilityInfo(prev => (areCapabilityInfosEqual(Maybe.of(prev), Maybe.of(current)) ? prev : current));
     };
 
     syncCapabilities();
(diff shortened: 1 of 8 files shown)


apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -94,6 +94,27 @@ const getCapabilityRows = (capabilityInfo: CapabilityInfo | null | undefined) =>
...
 
+const areCapabilityInfosEqual = (
+  prev: CapabilityInfo | null | undefined,
+  current: CapabilityInfo | null | undefined,
+): boolean => {
+  if (prev === current) {
+    return true;
+  }
+
+  return Maybe.of(prev)
```
screendriver: We are still mixing `Maybe` with `null | undefined` here instead of choosing one representation to keeping it consistent.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -133,40 +154,64 @@ const MetricsDisplay = ({capabilityInfo}: MetricsDisplayProps) => {
...
       setSelectedQuality(prev => (prev !== current ? current : prev));
     }, POLLING_INTERVAL);
 
     return () => clearInterval(interval);
-  }, [backgroundEffectsHandler]);
+  }, [backgroundEffectsHandler, isFeatureEnabled]);
+
+  // Capability polling (controller updates these after pipeline start)
+  useEffect(() => {
+    if (!isFeatureEnabled) {
```
screendriver: Shouldn't there be at least one single test that should fail now?

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -133,40 +154,64 @@ const MetricsDisplay = ({capabilityInfo}: MetricsDisplayProps) => {
...
+      setCapabilityInfo(null);
+      return undefined;
```
screendriver: Why the explicit return statement here? A `useEffect()` usually does not (and cannot) return anything.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -133,40 +154,64 @@ const MetricsDisplay = ({capabilityInfo}: MetricsDisplayProps) => {
...
+    }
+
+    const syncCapabilities = () => {
+      const current = backgroundEffectsHandler.getCapabilityInfo();
+      setCapabilityInfo(prev => (areCapabilityInfosEqual(prev, current) ? prev : current));
+    };
+
+    syncCapabilities();
+
+    const interval = setInterval(syncCapabilities, POLLING_INTERVAL);
```
screendriver, line 194: The new capability polling runs whenever the feature flag is enabled, even if the panel is closed

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
```diff
@@ -94,6 +94,27 @@ const getCapabilityRows = (capabilityInfo: CapabilityInfo | null | undefined) =>
     .unwrapOr([]);
 };
 
+const areCapabilityInfosEqual = (
+  prev: CapabilityInfo | null | undefined,
+  current: CapabilityInfo | null | undefined,
+): boolean => {
+  if (prev === current) {
```
screendriver: `areCapabilityInfosEqual` can miss updates if `CapabilityInfo` is mutated in place, because it short-circuits on reference equality. If the controller ever updates fields on that object instead of replacing it, `prev === current` will always be `true` and React will keep the same object in state, and the panel will never re-render.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
zskhan: Good question. I mentioned this to you in 1:1 but would be good for all.
This PerformancePanel ist just for internal testing purpose and it will not be client facing.
Putting efforts on adding test to this component would be a bit waste.

apps/webapp/src/script/components/calling/VideoControls/videoBackgroundPerformancePanel/videoBackgroundPerformancePanel.tsx
zskhan: Purpose here is early return.
I can also use if/else here but there are useEffects where we have cleanup methods like we have for `syncCapabilities`. So I just wanted to be consistent. Also early return in useEffect is also not wrong.
What do you suggest?

screendriver: Early return in `useEffect()` is fine but I would still prefer `return;` over `return undefined;` here. `useEffect` has two meaningful shapes for readers:

1. no cleanup / exit early
2. return a cleanup function

`return undefined;` is technically valid, but it adds noise and makes the reader pause on the return value even though there is no cleanup

e-maad (approved):

PR #21113, opened 2026-04-23
fix(core): skip stale MLS pending proposal commits [WPB-24984]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Guard pending proposal commits behind local conversation existence and prune stale rehydrated pending proposal tasks when the local MLS group is missing. This prevents deterministic "Couldn't find conversation" retries and startup log storms while preserving normal proposal handling for valid groups.

### DIFF
--- a/libraries/core/src/messagingProtocols/mls/mlsService/mlsService.ts
+++ b/libraries/core/src/messagingProtocols/mls/mlsService/mlsService.ts
@@ -1043,7 +1043,7 @@ export class MLSService extends TypedEventEmitter<Events> {
 
     const doesConversationExist = await this.conversationExists(groupId);
     if (!doesConversationExist) {
-      this.logger.warn('Skipping pending proposals commit because local MLS conversation is missing', {
+      this.logger.info('Skipping pending proposals commit because local MLS conversation is missing', {
         groupId,
         shouldRetry,
       });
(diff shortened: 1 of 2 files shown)


libraries/core/src/messagingProtocols/mls/mlsService/mlsService.ts
```diff
@@ -1077,11 +1088,20 @@ export class MLSService extends TypedEventEmitter<Events> {
     try {
       const pendingProposals = await this.coreDatabase.getAll('pendingProposals');
       if (pendingProposals.length > 0) {
-        pendingProposals.forEach(({groupId, firingDate}) =>
-          TaskScheduler.addTask({
-            task: () => this.commitPendingProposals(groupId),
-            firingDate,
-            key: this.createPendingProposalsTaskKey(groupId),
+        await Promise.all(
```
arjita-mitra, line 1091: This fixes the root cause of the repeated error loop so stale pending proposal jobs are no longer executed or rehydrated when the local MLS conversation is gone.

But It does not appear to solve why stale pendingProposals entries get created or left behind in the database in the first place. There's possibly some lifecycle inconsistency that is the real cause. wdyt?

arjita-mitra (approved):

libraries/core/src/messagingProtocols/mls/mlsService/mlsService.ts
thisisamir98: @arjita-mitra issue is when MLS conversations are created we add a task in our internal db to commit pending proposals and when conversation is deleted for whatever reason the task is executed without checking if the conversation is still there or no and this is exactly what has been addressed in this PR. we have no way of knowing when conversations are deleted from core crypto so checking their existence later when trying to commit pending proposals is the only solid option to handle this.

PR #21117, opened 2026-04-23
test: write the calling regression tests 2 part [WPB-19943]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19943" title="WPB-19943" target="_blank">TaskWPB-19943</a>  [Web/QA] Write the calling regression tests in Playwright
  </td></table>
  <br />
 

## Summary

This PR introduces the second phase of the Calling regression E2E test suite. To ensure a manageable review process, the full suite has been split into two logical parts.

---

## Accessibility (required)

---

## Notes for reviewers

- Trade-offs:
- Follow-ups (linked issues):

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
@@ -32,6 +32,12 @@ import {
 import {isPlayingAudio} from 'test/e2e_tests/utils/audio.util';
 import {createGroup, sendConnectionRequest} from 'test/e2e_tests/utils/userActions';
 
+async function joinCall(userPages: PageManager['webapp']['pages']) {
+  await expect(userPages.calling().callCell).toBeVisible();
+  await userPages.calling().clickAcceptCallButton();
+  await expect(userPages.calling().goFullScreen).toBeVisible();
+}
+
 test.describe('Calling', () => {
   let userA: User;
   let userB: User;
@@ -94,10 +100,7 @@ test.describe('Calling', () => {
     await userAPages.conversationList().getConversation(groupName).open();
     await userAPages.conversation().clickCallButton();
 
-    await expect(userBPages.calling().callCell).toBeVisible();
-    await userBPages.calling().clickAcceptCallButton();
-
-    await expect(userBPages.calling().goFullScreen).toBeVisible();
+    await joinCall(userBPages);
 
     const userACall = await userAPages.calling().maximizeCell();
     await expect(userACall.selfVideoThumbnail).toBeVisible();
@@ -259,11 +262,7 @@ test.describe('Calling', () => {
       await userAPages.conversation().clickCallButton();
 
       await expect(userAPages.calling().callCell).toBeVisible();
-      await expect(userBPages.calling().callCell).toBeVisible();
-
-      await userBPages.calling().clickAcceptCallButton();
-      // Confirm the calls grid is visible
-      await expect(userBPages.calling().goFullScreen).toBeVisible();
+      await joinCall(userBPages);
 
       // // User C joins the ongoing call
       await userCPage.waitForTimeout(5000);
@@ -293,10 +292,7 @@ test.describe('Calling', () => {
     await userAPages.conversation().clickCallButton();
 
     await expect(userAPages.calling().callCell).toBeVisible();
-    await expect(userBPages.calling().callCell).toBeVisible();
-
-    await userBPages.calling().clickAcceptCallButton();
-    await expect(userBPages.calling().goFullScreen).toBeVisible();
+    await joinCall(userBPages);
 
     await userAPages.calling().maximizeCell();
     await expect(userAPages.calling().selfVideoThumbnail).toBeVisible();
@@ -316,10 +312,9 @@ test.describe('Calling', () => {
     await userAPages.conversation().clickCallButton();
 
     await expect(userAPages.calling().callCell).toBeVisible();
-    await expect(userBPages.calling().callCell).toBeVisible();
 
     // User B joins the call
-    await userBPages.calling().clickAcceptCallButton();
+    await joinCall(userBPages);
     await expect(userBPages.calling().goFullScreen).toBeVisible();
 
     // User B leaves the group call
@@ -373,10 +368,8 @@ test.describe('Calling', () => {
         await userAPages.conversation().clickCallButton();
 
         await expect(userAPages.calling().callCell).toBeVisible();
-        await expect(userBPages.calling().callCell).toBeVisible();
 
-        await userBPages.calling().clickAcceptCallButton();
-        await expect(userBPages.calling().goFullScreen).toBeVisible();
+        await joinCall(userBPages);
       });
 
       await test.step('User A joins then leaves the call from their second device', async () => {
@@ -424,9 +417,7 @@ test.describe('Calling', () => {
 
       // Ensure all invited members join the call
       for (const member of [userBPages, userCPages]) {
-        await expect(member.calling().callCell).toBeVisible();
-        await member.calling().clickAcceptCallButton();
-        await expect(member.calling().goFullScreen).toBeVisible();
+        await joinCall(member);
       }
 
       // User A removes User B from the group
@@ -525,9 +516,7 @@ test.describe('Calling', () => {
 
       await test.step('Join call with all participants', async () => {
         for (const member of [userBPages, guestPages]) {
-          await expect(member.calling().callCell).toBeVisible();
-          await member.calling().clickAcceptCallButton();
-          await expect(member.calling().goFullScreen).toBeVisible();
+          await joinCall(member);
         }
       });
 
@@ -690,8 +679,7 @@ test.describe('Calling', () => {
       await test.step('Action: All members accept the incoming call', async () => {
         await Promise.all(
           memberPages.map(async memberPage => {
-            await memberPage.calling().acceptCallButton.click({timeout: 30_000});
-            await expect(memberPage.calling().goFullScreen).toBeVisible();
+            await joinCall(memberPage);
           }),
         );
       });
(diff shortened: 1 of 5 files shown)


apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -226,33 +238,47 @@ test.describe('Calling', () => {
...
-    await userCPage.waitForTimeout(5000);
-    await userCPages.conversationList().getConversationLocator(groupName).joinCallButton.click();
+      // Confirm that user C joined the call
+      await expect(userCPages.calling().goFullScreen).toBeVisible();
+      await expect(userCPages.calling().gridTiles).toHaveCount(3);
 
-    // Confirm that user C joined the call
-    await expect(userCPages.calling().goFullScreen).toBeVisible();
-    await expect(userCPages.calling().gridTiles).toHaveCount(3);
+      if (verifyScreenShare) {
```
markbrockhoff, line 271: Reading the test title it sound's like the screen should have already been shared before the user joins the conversation, but in this implementation the screen is shared after he already joined.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+        await expect(callScreen.getGridTile(localUser)).toBeHidden();
+        await callScreen.goToPreviousPage();
+        await expect(callScreen.getGridTile(localUser)).toBeVisible();
+      },
+    },
+    {
+      description: 'I want to see video tiles ordered alphabetically by user names',
+      tag: '@TC-2927',
+      verify: async (callScreen: ReturnType<PageManager['webapp']['pages']['fullScreenCall']>, localUser: string) => {
+        const displayedNames = await callScreen.gridTiles.getByTestId('call-participant-name').allInnerTexts();
```
markbrockhoff, line 635: 🔧  This assertion can't be retried automatically by Playwright since it only gets the inner texts once. So please make sure e.g. the number of gridTiles matches the number of users you expect before getting the texts to make sure it can fail.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+          const memberPages = await Promise.all(
+            groupMembers.map(async member => {
+              if (![userB, userC].includes(member)) {
+                await team.addTeamMember(member);
+              }
+
+              const page = await createPage(withLogin(member));
+              return PageManager.from(page).webapp.pages;
+            }),
+          );
```
markbrockhoff, line 659: 🎨 Nesting the if statement for adding the members to the team inside the map maybe it's easier to read and understand if the addition to the team was done as a standalone section:


```suggestion
          const extraMembers = await Promise.all(Array.from({length: 7}, () => createUser()));
          await Promise.all(extraMembers.map(member => team.addTeamMember(member)));
          
          const groupMembers = [userB, userC, ...extraMembers];
          const memberPages = await Promise.all(
            groupMembers.map(async member => {
              const page = await createPage(withLogin(member));
              return PageManager.from(page).webapp.pages;
            }),
          );
```

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+
+  test(
+    'I want to accept a group video call as a personal account guest',
+    {tag: ['@TC-2852', '@regression']},
+    async ({createPage}) => {
+      const [userAPage, guestPage] = await Promise.all([createPage(withLogin(userA)), createPage()]);
+      const {pages: ownerPages, modals: ownerModals} = PageManager.from(userAPage).webapp;
+      const guestPages = PageManager.from(guestPage).webapp.pages;
+
+      const createdLink = await test.step('Owner: Create group and generate guest invitation link', async () => {
```
markbrockhoff, line 574: That's really cool, didn't know `test.step()` proxied its return value. I should start using this too 🤔

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+
+          return {groupMembers, memberPages};
+        });
+
+      await test.step('Action: User A initiates the group call', async () => {
+        await createGroup(userAPages, groupName, groupMembers);
+        await userAPages.conversationList().openConversation(groupName);
+        await userAPages.conversation().clickCallButton();
+        // Warning modal about large group call
+        await userAPage.getByTestId('modal-without-title').getByRole('button', {name: 'Call', exact: true}).click();
```
markbrockhoff: Guess we don't have a POM for this type of modal yet. Could you please create one instead of hardcoding it here? Me might need it multiple times.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+      });
+    });
+  });
+
+  [
+    {
+      description: 'As a moderator I want to mute another participant',
+      tag: '@TC-2928',
+      verify: async (userPage: Page, callScreen: ReturnType<PageManager['webapp']['pages']['fullScreenCall']>) => {
+        await userPage.getByRole('button', {name: 'Mute', exact: true}).click();
```
markbrockhoff: Doesn't this cause a strict mode violation? I could imagine multiple buttons with the text "Mute" on the page, e.g. for the user himself and the users in the participant list. Maybe the exact is enough to prevent the violation but the locator isn't really deterministic like this. Do you think it would be possible to narrow down the part of the page where the mute button is supposed to be in? Just to keep maintenance effort low for the future :)

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+      description: 'I want to unmute myself after I got muted',
+      tag: '@TC-2931',
+      verify: async (userCallPage: Page) => {
+        const userCallPages = PageManager.from(userCallPage).webapp.pages;
+        await userCallPages.calling().fullScreenMuteButton.click();
+        await expect(userCallPages.calling().getGridTile(userB.fullName).muteIcon).not.toBeVisible();
+      },
+    },
+    {
+      description: 'I want to get notified when I got muted',
```
markbrockhoff, line 762: @iskvortsov some of the test cases in this suite are almost identical. Nothing to change for this PR, but maybe we could double check if those could be merged instead of having two tests with just a different assertion at the end :)

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+      });
+
+      await test.step('Moderator mutes User B', async () => {
+        const userACall = await userAPages.calling().maximizeCell();
+        await userACall.toggleParticipantsList();
+
+        await expect(userACall.getSidebarParticipant(userB.fullName).muteIcon).not.toBeVisible();
+
+        await userACall.getSidebarParticipant(userB.fullName).menuButton.click();
+        await userAPage.getByRole('button', {name: 'Mute', exact: true}).click();
```
markbrockhoff: Maybe `toggleParticipantsList()` / `getSidebarParticipant()` should return an enhanced Locator similar to e.g. the conversationItem context menu. This way the muteIcon and button could be scoped instead of checking the whole page potentially causing conflicts in the future. What do you think?

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+      await userAPages.conversationList().openConversation(groupName);
+      await userAPages.conversation().clickCallButton();
+      await expect(userAPages.calling().callCell).toBeVisible();
+
+      await expect(
+        userAPages
+          .conversation()
+          .systemMessages.filter({hasText: 'Your call was ended because no other participant joined.'}),
+      ).toBeVisible({
+        timeout: 302_000,
```
markbrockhoff: Ok, this might sound hypocritical from me but in this case I'd suggest to use `await page.waitForTimeout()` instead of increasing the delay of the assertion. The reason here is simply that we know the time it will take before the assertion can succeed. If the timeout is increased playwright will run and check the assertion constantly for 5m. With a dedicated await it knows that there's nothing to do and it will also show up like this in e.g. the traces.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
```diff
@@ -550,4 +566,353 @@ test.describe('Calling', () => {
...
+  );
+
+  test(
+    'I want to see a group call timing out after 30s if I`m the last one left in the call',
+    {tag: ['@TC-2937', '@regression']},
+    async ({createPage}) => {
+      const [userAPages, userBPages, userCPages] = await Promise.all([
+        PageManager.from(createPage(withLogin(userA), withConnectedUser(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userB))).then(pm => pm.webapp.pages),
+        PageManager.from(createPage(withLogin(userC))).then(pm => pm.webapp.pages),
```
markbrockhoff, line 843: Do we really need 3 users for this test or would a group with only 2 participants also be sufficient? I'm just thinking of saving some resources :)

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
zhannach: Valid concern, but here is no strict mode violation.
This menu is a separate element outside the main tag injected into the body only after user clicks the menuButton.
The owner has a different icon button to mute themselves. I can switch to a testId to make the locator more deterministic.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
zhannach: I looked into it, but unfortunately, the menu is implemented as a separate element outside the main tag, making it difficult to link to other elements. I would really appreciate any hints or suggestions on how to handle this

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
zhannach: Two users should be enough to check the behavior. I stuck to the description in Testiny.

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
markbrockhoff: TBH I would probably just go into the component (`CallingParticipantList.tsx`) and change the div to an `aside` element, maybe also adding ids and `aria-labelledby` attributes for the participant lists. Then you could scope the locators for the participants the same way as in e.g. [guestOptions.page.ts](https://github.com/wireapp/wire-webapp/blob/97ec70c812a4f9372ab72a8797d2c91c44618cc0/apps/webapp/test/e2e_tests/pageManager/webapp/pages/guestOptions.page.ts#L25). Plus it would improve a11y :)

apps/webapp/test/e2e_tests/specs/Calling/calling.spec.ts
markbrockhoff: I had the same issue with the context menu in the conversation list, but the menu is accessible, you can target it by using `getByRole('menu')` as an example you can take a look at the [openContextMenu function](https://github.com/wireapp/wire-webapp/blob/97ec70c812a4f9372ab72a8797d2c91c44618cc0/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationList.page.ts#L94) :)

markbrockhoff (dismissed):

markbrockhoff (approved):

zskhan (approved):

PR #21161, opened 2026-04-27
ci: use manual testrun import instead of testiny "automation" [WPB-24713]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24713" title="WPB-24713" target="_blank">TaskWPB-24713</a>  [Web/QA] Report nightly e2e test runs to Testiny as manual test runs
  </td></table>
  <br />
 

Since the way Testiny handles "automated" test runs is way more limited than the manual import of test runs, we decided to for now import the test results as if they were manual ones instead of using the automations feature.

Sadly using the Testiny CLI directly wasn't possible since it didn't have all features needed and would e.g. interpret flaky tests as failed so I decided to take full control using a custom ts script and the Testiny API directly.

### DIFF
--- a/apps/webapp/test/e2e_tests/scripts/uploadTestReport.ts
+++ b/apps/webapp/test/e2e_tests/scripts/uploadTestReport.ts
@@ -153,6 +153,7 @@ async function main() {
   const report: JSONReport = JSON.parse(fs.readFileSync(reportAbsPath, 'utf-8'));
   const testRun = await createTestRun(`
     <!--markdown-->
+    
     Build URL: [${args.buildURL}](${args.buildURL})
   `);
 
--- a/apps/webapp/test/e2e_tests/scripts/uploadTestReport.ts
+++ b/apps/webapp/test/e2e_tests/scripts/uploadTestReport.ts
@@ -30,6 +30,11 @@ const {values: args} = parseArgs({
       alias: 'name',
       description: 'Name of the test run',
     },
+    testPlanId: {
+      type: 'string',
+      alias: 'testPlan',
+      description: 'ID of the test plan this run should be associated with',
+    },
     description: {
       type: 'string',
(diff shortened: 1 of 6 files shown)


apps/webapp/test/e2e_tests/scripts/uploadTestReport.ts
```diff
@@ -0,0 +1,184 @@
...
+  ids: {testcase_id: number; testrun_id: number};
+  mapped: {assigned_to: 'ANY'; result_status: 'NOTRUN' | 'PASSED' | 'FAILED' | 'BLOCKED' | 'SKIPPED'};
+};
+
+function transformReportToTestinyMappings(report: JSONReport, runId: number) {
+  // Mapping from playwrights status to the result expected by Testiny
+  const resultMap: Record<JSONReportTest['status'], TestinyTestCaseMapping['mapped']['result_status']> = {
+    expected: 'PASSED',
+    unexpected: 'FAILED',
+    flaky: 'PASSED',
```
JacquelineLehner, line 118: Why does `flaky` count as `passed`? Just curious

markbrockhoff: Because Testiny doesn't have a state for flaky tests. So since a flaky test is one which e.g. failed due to rate limiting on a first try but then passed on a retry we're considering them as passed for this report. But we might add this information somewhere for visibility in the future as it is relevant for us devs to improve the stability of our tests :)

JacquelineLehner: Got it, thank you! :)

markbrockhoff: Feel free to ask any time ;)

zskhan (approved):

PR #21166, opened 2026-04-28
test: write missing History Backup test [WPB-24920]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24920" title="WPB-24920" target="_blank">TaskWPB-24920</a>  [Web][QA] Write missing History Backup test
  </td></table>
  <br />
 

Summary

This PR adds a new automated regression test to History Backup test suites to verify that a user remains restricted from sending messages in a group they have left, even after performing a History Backup and Restore.

Changes

Added test case TC-10549 to the regression suite.

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/HistoryBackup/historyBackup.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/HistoryBackup/historyBackup.spec.ts
@@ -27,6 +27,8 @@ import {RequestResetPasswordPage} from '../../pageManager/webapp/pages/requestRe
 test.describe('History Backup', () => {
   let userA: User;
   let userB: User;
+  const conversationName = 'Test group';
+
   test.beforeEach(async ({createTeam, createUser}) => {
     userB = await createUser();
     const team = await createTeam('Test Team', {users: [userB]});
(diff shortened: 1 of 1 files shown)


apps/webapp/test/e2e_tests/specs/HistoryBackup/historyBackup.spec.ts
```diff
@@ -378,4 +376,55 @@ test.describe('History Backup', () => {
...
+        await userAComponents.conversationSidebar().clickPreferencesButton();
+        return await createAndSaveBackup(testInfo, userAPageManager);
+      });
+
+      await test.step('User A restores backup', async () => {
+        await logOutUser(userAPageManager, true);
+        await loginUser(userA, userAPageManager);
+
+        await userAPages.historyInfo().clickConfirmButton();
+        await userAComponents.conversationSidebar().clickPreferencesButton();
```
markbrockhoff: Could it make sense to add the LOGIN_TIMEOUT to this action to prevent flake? It should be pretty ok since assets are already cached in this browser session but in case some backend endpoints take a bit longer it might still be a good idea.

> I'd suggest to just use the button locator directly instead of the `clickPreferencesButton()` util since it already supports setting a custom timeout ;)

markbrockhoff (dismissed):

JacquelineLehner (dismissed):

markbrockhoff (approved):

JacquelineLehner (approved):

zskhan (approved):

PR #21172, opened 2026-04-29
test: verify link preview and media playback [WPB-24824]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24824" title="WPB-24824" target="_blank">TaskWPB-24824</a>  [Web][QA] Write test for Regression -> Link Preview
  </td></table>
  <br />
 

This PR introduces automated test coverage for Link Previews within group conversations (TC-1264). 
The test verifies that when a user sends a link from supported providers (YouTube, SoundCloud, Spotify, and Vimeo), the application correctly renders a rich preview iframe and the media player is functional

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/LinkPreview/linkPreview.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/LinkPreview/linkPreview.spec.ts
@@ -1,0 +1,93 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {FrameLocator} from 'playwright/test';
+import {PageManager} from 'test/e2e_tests/pageManager';
+import {test, withLogin, expect, withConnectedUser} from 'test/e2e_tests/test.fixtures';
+
+test.describe('Link Preview', () => {
+  test(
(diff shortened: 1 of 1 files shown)


apps/webapp/test/e2e_tests/specs/LinkPreview/linkPreview.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            await expect(frame.getByRole('application', {name: 'Pause', exact: true})).toBeVisible();
+          },
+        },
+        {
+          url: 'https://www.youtube.com/watch?v=BMFsJiAcELY',
+          iframeSelector: 'iframe.youtube',
+          validate: async (frame: FrameLocator) => {
+            await frame.getByRole('button', {name: 'Play video', exact: true}).click();
+            // Video player might take some time to load, so here we aren't checking for the pause button
+            await expect(frame.getByLabel('Time elapsed 0:00')).toBeVisible();
```
markbrockhoff: Please don't use a label with time in it as locator 😅
At least don't include the exact time since this will break as soon as the video jumps from 0:00 to 0:01 making timing here very tight.
Also please add a counter assertion before clicking the play button otherwise the label might already exist initially causing this test to always pass even if playing didn't work

apps/webapp/test/e2e_tests/specs/LinkPreview/linkPreview.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+          iframeSelector: 'iframe.vimeo',
+          validate: async (frame: FrameLocator) => {
+            await frame.getByRole('button', {name: 'Play', exact: true}).click();
+            await expect(frame.getByRole('button', {name: 'Pause', exact: true})).toBeVisible();
+          },
+        },
+      ];
+
+      for (const link of linkConfigs) {
+        await userAPages.conversation().sendTypedMessage(link.url);
```
markbrockhoff: Why do you send the message typed and don't just use `sendMessage(link.url)`? Is there a specific reason?

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/LinkPreview/linkPreview.spec.ts
zhannach: Great catch!

markbrockhoff (approved):

JacquelineLehner (approved):

screendriver (approved):

PR #21174, opened 2026-04-29
fix(mls): ensure conversations are established using remote epoch state [WPB-24984]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Refactor MLS conversation establishment to rely on core-crypto + backend epoch checks instead of passing a local epoch through call sites. This handles locally unestablished groups more safely (wipe + rehydrate/join), introduces safe epoch retrieval via `getSafeEpoch`, and adds targeted logging around init, decrypt failures, external commits, and pending proposal commits to improve recovery/debugging.


e-maad (dismissed):

apps/webapp/src/script/mls/MLSMigration/migrationFinaliser/joinConversationsAfterMigrationFinalisation/joinConversationsAfterMigrationFinalisation.test.ts
```diff
@@ -115,6 +115,12 @@ const createConversation = (
   return conversation;
 };
 
+const mockSafeEpoch = (core: Core) => {
+  (core.service!.mls! as any).getSafeEpoch = jest
```
screendriver: - never use `any` type assertions
- never use definite assignment assertions `!`

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+    }
+
+    const remoteConversation = await this.conversationService.getConversationById(conversationId);
+    const remoteEpoch = remoteConversation.epoch;
+
+    // establish the conversation if remote epoch is 0
+    if (remoteEpoch === 0) {
+      this.logger.info('Establishing conversation as remote epoch is 0', {remoteEpoch});
+      await this.establishMlsGroupConversation({conversationId, groupId, epoch: remoteEpoch, core});
       return;
```
screendriver: This is a missing/invalid remote epoch, now silently no-ops. The removed recovery code threw when the remote epoch was unavailable. The new implementation does nothing if `remoteConversation.epoch` is `undefined`, `null` or otherwise falsy but not exactly `0`. In the message-send path, `ensureConversationExists` can return successfully and then sending continues.

apps/webapp/src/script/mls/MLSConversations.test.ts
```diff
@@ -64,6 +69,10 @@ describe('MLSConversations', () => {
       const conversationRepository = await testFactory.exposeConversationActors();
       const repositoryCore = (conversationRepository as any).core as Core;
       jest.spyOn(repositoryCore.service!.conversation, 'mlsGroupExistsLocally').mockResolvedValue(false);
+      jest
+        .spyOn((conversationRepository as any).conversationService, 'getConversationById')
```
screendriver: No `as any` please

apps/webapp/src/script/mls/MLSMigration/migrationInitialiser/joinUnestablishedMixedConversations/joinUnestablishedMixedConversations.test.ts
```diff
@@ -60,8 +60,13 @@ describe('joinUnestablishedMixedConversations', () => {
       return Promise.resolve(!groupId.includes('unestablished'));
     });
 
+    (repositoryCore.service!.mls! as any).getSafeEpoch = jest.fn().mockResolvedValue({isOk: true, error: null});
```
screendriver: Never use `as any` please

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
-        }
-        return this.recoverFromLocalUnestablishedMLSConversations({
-          conversationId,
-          groupId,
-          epoch: coreCryptoEpochNumber,
-          core,
-        });
-      }
+    const conversationExistsOnCoreCrypto = await this.conversationService.mlsGroupExistsLocally(groupId);
+    const coreCryptoEpochNumberResult = await core.service?.mls?.getSafeEpoch(groupId);
```
screendriver: Can we avoid optional chaining here and introduce an explicit type guard for the MLS service before calling `getSafeEpoch`? Right now `core.service?.mls?.getSafeEpoch(groupId)` can return `undefined`, but line 2234 treats the result as present.

typfel (dismissed):

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+    const coreCryptoEpochNumber = coreCryptoEpochNumberResult.isOk ? coreCryptoEpochNumberResult.value : undefined;
```
screendriver: Can we return or throw immediately on `coreCryptoEpochNumberResult.isErr` instead of logging and continuing? A failed epoch read means we do not have a valid `number` to reason about. Continuing mixes the error case into the normal control flow.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+    this.logger.info('Ensuring conversation exists', {
+      conversationId,
+      groupId,
+      coreCryptoEpochNumber,
+      conversationExistsOnCoreCrypto,
+    });
+
+    if (coreCryptoEpochNumberResult.isErr) {
+      this.logger.warn(
+        'conversation existed on core crypto but there was an error when retrieving its epoch number',
```
screendriver: This log message is misleading because it says the conversation existed on core-crypto, but this branch only checks `getSafeEpoch` failure. It can run even when `conversationExistsOnCoreCrypto` is false.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+        coreCryptoEpochNumberResult.error,
+      );
+    }
+
+    if (conversationExistsOnCoreCrypto && coreCryptoEpochNumber > 0) {
```
screendriver: Can we avoid comparing a possibly `undefined` value here? `coreCryptoEpochNumber` is `undefined` when the result is `Err`, so this condition depends on a value that is not cleanly narrowed.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
-      await this.establishMlsGroupConversation({conversationId, groupId, epoch, core});
+    if (conversationExistsOnCoreCrypto && coreCryptoEpochNumber === 0) {
+      this.logger.info('Conversation already exists locally but epoch is 0, wiping it');
+
+      // The conversation was created locally but the initial commit was never accepted by the backend
+      // Wipe the conversation locally and join by external commit
+      await core.service?.conversation?.wipeMLSConversation(groupId);
+    }
+
+    const remoteConversation = await this.conversationService.getConversationById(conversationId);
```
screendriver: Can we add an explicit type guard for `remoteConversation.epoch` right after fetching the remote conversation? This method’s behavior depends entirely on epoch being a valid number, but the current code accepts any shape and lets invalid values fall through.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+    const remoteEpoch = remoteConversation.epoch;
+
+    // establish the conversation if remote epoch is 0
+    if (remoteEpoch === 0) {
```
screendriver: Can we make the epoch decision exhaustive instead of relying on partial conditions? Or in other words: this handles `0` but there is no corresponding explicit branch for invalid epoch values.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2223,50 +2223,57 @@ export class ConversationRepository {
...
+      this.logger.info('Establishing conversation as remote epoch is 0', {remoteEpoch});
+      await this.establishMlsGroupConversation({conversationId, groupId, epoch: remoteEpoch, core});
       return;
     }
 
     // join by external commit
-    this.logger.info('Joining conversation by external commit', {conversationId, epoch});
-    if (epoch && epoch > 0) {
+    if (remoteEpoch && remoteEpoch > 0) {
```
screendriver: Can we avoid the truthy/falsy check here?

`if (remoteEpoch && remoteEpoch > 0)` conflates “missing”, “zero”, and “invalid” with normal control flow. Or in other words: `undefined` or `null` will skip joining and the method still resolves.

apps/webapp/src/script/mls/MLSMigration/migrationFinaliser/joinConversationsAfterMigrationFinalisation/joinConversationsAfterMigrationFinalisation.test.ts
```diff
@@ -144,6 +151,10 @@ describe('joinConversationsAfterMigrationFinalisation', () => {
 
     const onSuccess = jest.fn();
     const conversationRepository = await testFactory.exposeConversationActors();
+    jest
+      .spyOn((conversationRepository as any).conversationService, 'getConversationById')
```
e-maad: Same as: https://github.com/wireapp/wire-webapp/pull/21174/changes#r3167603248

apps/webapp/src/script/mls/MLSMigration/migrationFinaliser/joinConversationsAfterMigrationFinalisation/joinConversationsAfterMigrationFinalisation.test.ts
```diff
@@ -115,6 +116,12 @@ const createConversation = (
   return conversation;
 };
 
+const mockSafeEpoch = (core: Core) => {
+  (core.service?.mls as unknown as {getSafeEpoch: jest.Mock}).getSafeEpoch = jest
+    .fn()
+    .mockResolvedValue(task.fromResult(result.ok(1)));
```
screendriver:
```suggestion
    .mockResolvedValue(task.resolve(1));
```

apps/webapp/src/script/mls/MLSConversations.test.ts
```diff
@@ -47,6 +48,12 @@ function createMLSConversations(nbConversations: number, type?: CONVERSATION_TYP
   return Array.from(new Array(nbConversations)).map(() => createMLSConversation(type));
 }
 
+function mockSafeEpoch(core: {service?: Core['service']}) {
+  (core.service?.mls as unknown as {getSafeEpoch: jest.Mock}).getSafeEpoch = jest
+    .fn()
+    .mockResolvedValue(task.fromResult(result.ok(1)));
```
screendriver:
```suggestion
    .mockResolvedValue(task.resolve(1));
```

screendriver (dismissed):

arjita-mitra (dismissed):

zskhan (dismissed):

apps/webapp/src/script/mls/MLSConversations.ts
```diff
@@ -169,19 +179,231 @@ export async function initialiseSelfAndTeamConversations(
...
+  conversationId: QualifiedId,
+  dependencies: {
+    core: Account;
+    conversationService: ConversationService;
+    userState: UserState;
+    conversationState: ConversationState;
+  },
+): Promise<void> {
+  const {core, conversationService, userState, conversationState} = dependencies;
+  const mlsService = getMLSService(core);
```
screendriver: Why not just


```suggestion
  const mlsService = Maybe.of(core.service?.mls);
  
  if (mlsService.isNothing) {
    logger.error('MLS service is not available!');
    return;
  }
```

thisisamir98: Bitte schön @screendriver  https://github.com/wireapp/wire-webapp/pull/21174/commits/74ac887da62397a4e703e8b643aa094215792b59

e-maad (approved):

PR #21190, opened 2026-04-30
feat: update avs to version 10.3.19 [WPB-25162]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25162" title="WPB-25162" target="_blank">TaskWPB-25162</a>  [Web] - Update avs to 10.3.16
  </td></table>
  <br />
 

Updated AVS to v10.3.19.
It also has interface update for quality_info inside wCall.setNetworkQualityHandler which is integrated in WebApp.
Added few tests to verify the json string parsing.

### DIFF
--- a/apps/webapp/src/script/repositories/calling/CallingRepository.ts
+++ b/apps/webapp/src/script/repositories/calling/CallingRepository.ts
@@ -59,6 +59,7 @@ import {useCallAlertState} from 'Components/calling/useCallAlertState';
 import {PrimaryModal} from 'Components/Modals/PrimaryModal';
 import {CALL_QUALITY_FEEDBACK_KEY} from 'Components/Modals/QualityFeedbackModal/constants';
 import {RatingListLabel} from 'Components/Modals/QualityFeedbackModal/typings';
+import {NetworkQualityInfo, NetworkQualityInfoSchema} from 'Repositories/calling/calling.schema';
 import {isMLSConversation, MLSConversation} from 'Repositories/conversation/ConversationSelectors';
 import {ConversationState} from 'Repositories/conversation/ConversationState';
 import {ConversationVerificationState} from 'Repositories/conversation/ConversationVerificationState';
@@ -148,24 +149,6 @@ type SubconversationData = {
   epoch: number;
   secretKey: string;
   members: SubconversationEpochInfoMember[];
-};
-
-type NetworkQualityInfo = {
-  quality: QUALITY;
-  rtt: number;
-  loss: {
-    tx: number;
-    rx: number;
-  };
-  jitter: {
-    audio: {tx: number; rx: number};
-    video: {tx: number; rx: number};
-  };
-  connection: {
-    candidate: 'Relay' | 'Host' | 'Srflx' | 'Prflx' | 'Unknown';
-    protocol: 'UDP' | 'TCP' | 'Unknown';
-  };
-  peer: 'Server' | 'User' | 'Unknown';
 };
 
 export class CallingRepository {
@@ -642,7 +625,8 @@ export class CallingRepository {
     let qualityInfo: NetworkQualityInfo;
 
     try {
-      qualityInfo = JSON.parse(qualityInfoJson) as NetworkQualityInfo;
+      const parsedQualityInfo = JSON.parse(qualityInfoJson);
+      qualityInfo = NetworkQualityInfoSchema.parse(parsedQualityInfo);
     } catch (error) {
       this.logger.warn(`Invalid network quality info: ${qualityInfoJson}`, error);
       return;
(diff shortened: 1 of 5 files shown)


apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -619,8 +637,19 @@ export class CallingRepository {
     conversationId: SerializedConversationId,
     userId: string,
     clientId: string,
-    quality: number,
+    qualityInfoJson: string,
   ) => {
+    let qualityInfo: NetworkQualityInfo;
+
+    try {
+      qualityInfo = JSON.parse(qualityInfoJson) as NetworkQualityInfo;
```
screendriver: Could we avoid the `as NetworkQualityInfo` type assertion here?

`JSON.parse()` gives us runtime data (as `any`) so TypeScript cannot guarantee that it actually matches `NetworkQualityInfo`. A type assertion only tells TypeScript to trust it.

Since this is a boundary I think this would be a good place for runtime validation, for example with `Zod` (which we [already have](https://github.com/wireapp/wire-webapp/blob/da47945caa08a586ade51af022b0c1227a4fe5f3/package.json#L13))

apps/webapp/src/script/repositories/calling/calling.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import {z} from 'zod';
+
+import {QUALITY} from '@wireapp/avs';
+
+const CandidateEnum = z.enum(['Relay', 'Host', 'Srflx', 'Prflx', 'Unknown']);
+const ProtocolEnum = z.enum(['UDP', 'TCP', 'Unknown']);
+const PeerEnum = z.enum(['Server', 'User', 'Unknown']);
+const QualityEnum = z.nativeEnum(QUALITY);
```
screendriver: Could we replace this `z.enum([...])` with `z.union([z.literal(...), ...])`? I think a string-literal union better represents the domain here and keeps the schema explicit without enum-style helpers we don’t currently use. (also `enum` does not exist in standard JavaScript which therefore can always be avoided when using string literals. TypeScript `enum` also introduces unnecessary runtime and interop complexity.)

For example

```ts
const fooSchema = z.union([
  z.literal("bar"),
  z.literal("baz"),
  z.literal("hello"),
]);
```

Also consider renaming the schemas with a schema postfix like `candidateSchema` for clarity.

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -619,8 +620,20 @@ export class CallingRepository {
...
-    quality: number,
+    qualityInfoJson: string,
   ) => {
+    let qualityInfo: NetworkQualityInfo;
+
+    try {
+      const parsedQualityInfo = JSON.parse(qualityInfoJson);
+      qualityInfo = NetworkQualityInfoSchema.parse(parsedQualityInfo);
+    } catch (error) {
+      this.logger.warn(`Invalid network quality info: ${qualityInfoJson}`, error);
```
screendriver: - it logs raw `qualityInfoJson` which may be large/untrusted and can pollute logs
- String interpolation loses structure
- now we log the `zod` parsing error. Which could be fine, or not. Depending on what exactly we want to have here?

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -624,8 +620,20 @@ export class CallingRepository {
...
+      return;
+    }
+
+    const {quality} = qualityInfo;
```
screendriver, line 641: Could we simplify this by validating only the `quality` field and not over-schema the rest of the payload? In `updateCallQuality` we currently parse the full object but only read `quality` afterward. A minimal schema like `{ quality: ... }` seems sufficient here. Also `Zod` can strip unknown/unused properties, so we can keep validation focused on what we actually use without carrying a larger schema.

apps/webapp/src/script/repositories/calling/CallingRepository.ts
zskhan: There is no need of qualityInfoJson at all.
What I did now is just distinguished between zod error and json parsing error.

apps/webapp/src/script/repositories/calling/CallingRepository.ts
zskhan: That is something we are going to use in the next PRs.
We need to show more quality standards so for now I just made other things optional but I will update the schema as soon as we implement the new features on top of NetworkQualityInfoSchema

screendriver (approved):

PR #21220, opened 2026-05-04
fix: change bg effects implementation [WPB-24370]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24370" title="WPB-24370" target="_blank">StoryWPB-24370</a>  [Web] In Firefox BackgroundEffects worker pipeline fails due to OffscreenCanvas and  captureStream incompatibility
  </td></table>
  <br />
 

# Summery

With the current implementation, we've reached our limits. The implementation was so complex that we could only consider large parts of it as a black box. Unfortunately, it wasn't stable enough for that, so we decided to simplify it significantly.

* This enabled us to create a version that works across all browsers.
* We have greatly simplified the complexity of the source code in this version.
* The new version uses among other things WritableStream, resulting in a significant performance improvement. This prevents the executing task from blocking:

    

### Before:

Bildschirmfoto 2026-05-04 um 14 48 30

### After:

Bildschirmfoto 2026-05-04 um 16 42 32


EnricoSchw:
@zskhan @screendriver 

Please pay special attention to the new dependency. It has some advantages but comes with a few significant behavioral changes.

https://www.npmjs.com/package/worker-timers

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -0,0 +1,310 @@
...
+  function attachCanvasEvents() {
+    canvas.addEventListener('webglcontextlost', onContextLost, {once: true});
+    canvas.addEventListener('webglcontextrestored', onContextRestored, {once: true});
+  }
+  attachCanvasEvents();
+
+  let segmenter = await createSegmenter(canvas);
+  let currentModelPath = segmenterOptions.modelPath;
+
+  function restartSegmenter() {
```
screendriver: I think we have a race here. `restartSegmenter()` can be triggered from multiple places and because it's fire-and-forget we can end up applying an older restart after a newer one. Can we serialize restarts so only the latest restart wins? I'm also concerned to not use `async` and `await` here.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -0,0 +1,310 @@
...
+    logger.log(`[virtual-background] webglcontextlost (${!!webGLRenderer})`);
+    event.preventDefault();
+    webGLRenderer?.close();
+    webGLRenderer = null;
+  }
+
+  function onContextRestored() {
+    logger.log(`[virtual-background] webglcontextrestored (${!!webGLRenderer})`);
+    if (!webGLRenderer) {
+      setTimeout(() => {
```
screendriver: I think this `setTimeout()` restart needs coordination. If context restore flaps, we can queue multiple delayed restarts and race segmenter swaps. Can we debounce/cancel the pending timer and guard execution with the same restart generation id used by `restartSegmenter()`? Also, please use our wall clock abstraction here instead of raw setTimeout so this path stays deterministic in tests.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -0,0 +1,310 @@
...
+          lastStatsTime = now;
+          totalMsSum = 0;
+          segmentationMsSum = 0;
+          gpuMsSum = 0;
+          filterMsSum = 0;
+          frames = 0;
+        }
+
+        // Restart segmenter to avoid memory leaks.
+        if (segmenterOptions.restartEvery && totalFrames % segmenterOptions.restartEvery === 0) {
```
screendriver: This periodic restart can overlap with the context-restore restart path. Might be safer to funnel both paths through the same serialized restart mechanism so we never run two restarts at once.

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  public setBackgroundSource(source: BackgroundSource): void {
+    if (source.type !== 'image') {
+      return;
+    }
+
+    const {media, url} = source;
+
+    if (media instanceof HTMLImageElement) {
+      createImageBitmap(media)
```
screendriver: This async decode is not ordered against subsequent `setBackgroundSource()` calls. If the user switches backgrounds quickly, an older decode can finish later and overwrite the latest choice. Can we guard this with a request/version check before applying? I'm also concerned about to not use `async` and `await` here.

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -0,0 +1,343 @@
...
+      const {options: workerOptions, transferables} = getWorkerOptions(resolved);
+
+      transferables.push(offscreen, readable);
+
+      this.worker.postMessage(
+        {name: 'runSegmenter', canvas: offscreen, readable, options: workerOptions},
+        transferables,
+      );
+
+      this.worker.addEventListener('message', ({data}) => {
```
screendriver: Looks like we add a new worker message listener on every `start()`. Do we want to clean this up on stop/terminate (or register once) to avoid listener buildup and duplicated metrics callbacks over time?

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -0,0 +1,343 @@
...
+    };
+  }
+
+  /**
+   * In the old version, ending the effect was done centrally in the calling repository. Now we've tied ending the
+   * effect directly to the video track. This method is therefore no longer necessary. However, I'm leaving this
+   * method in to resolve any potential About errors. Sometimes the effect is started twice, in which case one needs
+   * to be stopped before starting again.
+   */
+  public stop(): void {}
```
screendriver: `stop()` being a no-op worries me a bit since callers still use stop semantics in cleanup/error paths. Should this do real teardown or should we remove/rename it so the contract is explicit?

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -0,0 +1,343 @@
...
+      ...this.options,
+      blurStrength: value,
+      // Only raise bgBlur when in blur mode; virtual/passthrough use bgBlur=0.
+      bgBlur: this.options.mode === 'blur' ? blurStrengthToBgBlur(value) : this.options.bgBlur,
+    };
+    this.pushOptionsUpdate();
+  }
+
+  public setBackgroundSource(source: BackgroundSource): void {
+    if (source.type !== 'image') {
```
screendriver: BackgroundSource now includes video/stream, but this setter ignores non-image sources. Is that intentional for this pull request? If yes, can we make that explicit in types or logs to avoid silent no-op behavior for callers?

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/renderer.ts
```diff
@@ -0,0 +1,679 @@
...
+          canvas.width = videoFrame.codedWidth;
+          canvas.height = videoFrame.codedHeight;
+          ctx?.drawImage(videoFrame, 0, 0);
+          videoFrame.close();
+        },
+        close() {
+          logger.error('[virtual-background] video background close');
+        },
+      });
+      media.pipeTo(writer).catch((err: unknown) => {
```
screendriver: Should we cancel/abort the previous `pipeTo()` when background source changes or renderer closes? Otherwise old video streams may keep pumping frames after source switch. I'm also concerned to not use `async` and `await` here.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -0,0 +1,310 @@
...
+  buildMetrics,
+  createMetricsWindow,
+  pushMetricsSample,
+} from 'Repositories/media/backgroundEffects/helper/metrics';
+
+import {VideoFilter} from './filter';
+import {WorkerProcessVideoTrackOptions} from './options';
+import {WebGLRenderer} from './renderer';
+
+export let segmenterOptions = {} as WorkerProcessVideoTrackOptions;
```
screendriver: This module-global mutable options object makes concurrent instances risky. If two processing sessions ever run in parallel, options can bleed between them. Can we scope options per run/instance?

screendriver (changes requested):
This refactor changes core pipeline behavior significantly. I'd feel safer if we add targeted tests for restart serialization + background-switch ordering (stale bitmap prevention), since those are the highest-risk race paths.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
EnricoSchw: I will implement a queue so that we only ever have a single restart. In general, we avoid segmenter restarts, so this shouldn't happen very often.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
EnricoSchw: I'll take a look at your timer. If it's worker safe, I'll use it.

Otherwise, I will introduce a "debounce" state; this will prevent multiple segment restarts at this point. This is a good idea.

I will not implement a restart id, parallel processing is already covered by the serialization of the restart mechanism. I don't see the idea behind it.  A restart shouldn't actually ever be necessary, this is merely a safeguard for the case thats never should happen.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
EnricoSchw: This option we not use. It is for situations where you are constantly changing the settings. I will remove this case.

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
EnricoSchw: Thx good finding!

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
EnricoSchw: Good point, I will check if we even reset the worker on start.

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
EnricoSchw: I wil remove!

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
EnricoSchw: Oh yes... theoretically, we could use videos as a background. But I'd better remove that.

EnricoSchw:
@screendriver Could you please review and approve this for us?

zskhan (approved):

PR #21217, opened 2026-05-04
Add ADR-0001 and distinguish it from ADR guide [WPB-22420]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-22420" title="WPB-22420" target="_blank">TaskWPB-22420</a>  [Web] General maintenance ticket for PR merges
  </td></table>
  <br />
 

Adds `docs/decision-records/0001-use-lightweight-architecture-decision-records.md` as a formal architecture decision record for this repository. Unlike `docs/ADRs/README.md`, which serves as a practical writing guide and template, ADR-0001 documents an explicit, accepted decision together with its rationale and consequences. It establishes lightweight architecture decision records as the standard approach and defines the core sections each ADR should contain.

PR #21221, opened 2026-05-04
Add automated release process ADR with branch/workflow diagrams and long-term-support hotfix flow [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

  ## Purpose

Establish a release process that is safe, predictable, and fully executable from GitHub without local release operations. The process ensures that quality assurance controls promotion from beta to production, production tags represent actual successful deployments, hotfixes can be propagated quickly without branch drift, and long-term-support customers receive stable, low-frequency releases with a dedicated hotfix path.

## What this pull request contains

 - New ADR: `docs/decision-records/0002-automated-quality-assurance-gated-release-process.md`
 - ADR status is **Proposed** to enable team discussion before implementation.

## Review request

Please read the **entire ADR** carefully, including:

- Context and alternatives
- Full decision text
- Both Mermaid diagrams
- Consequences

## Discussion focus

Please comment on:

- Branch and release model clarity (`dev`, `master`, `release/...`, `lts/...`)
- Hotfix propagation model and traceability
- Quality assurance approval model
- Edge deployment gating
- Long-term-support cadence and hotfix handling

## Process

This is the discussion round.
After comments are resolved, I will freeze the ADR text and request final approvals.


docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+  hotfixCommit[Hotfix commit merged to master] -->|Cherry-pick -x via workflow| releaseBranch
+  hotfixCommit -->|Cherry-pick -x via workflow| developmentBranch
+  hotfixCommit -->|Cherry-pick -x via workflow| longTermSupportBranch
+  longTermSupportBranch -->|Deploy long-term-support hotfix and tag| longTermSupportHotfixTag
+```
+
+### Alternatives
+
+- Keep release triggers primarily tag-driven for production deployments.
+- Keep automatic Edge deployments without additional approval gates.
```
markbrockhoff: I'd like to question the need for approval gates on Edge. I get the idea of it but who is supposed to approve the deployment and when? If Edge is deployed from dev directly there would be one request for deployment per commit. I feel like this would either result in the deployments being ignored or just approved without previous checking at some point, simply due to the overload of approval requests. With Beta and Prod there's a clear responsibility for QA to do the approval but what about Edge?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+The release process will use ISO-date identifiers in UTC with same-day sequence numbers: `YYYY-MM-DD.N`.
+Sequence numbers are determined by workflow automation from existing tags for the same UTC date, and workflows must fail on sequence conflicts.
+
+The process is:
+
+- A release branch is cut from `master` as `release/YYYY-MM-DD.N`.
+- `Release Beta` is triggered manually through GitHub Actions using `release_identifier`.
+- Quality assurance validates the deployment in beta.
+- `Release Production` is triggered manually through GitHub Actions using the same `release_identifier`.
```
markbrockhoff: What is meant by "is triggered manually through Github Actions" for Beta and Prod? Because I'd question at least the deployment to beta being a manual extra step. Why not deploy to beta from the release branch right away? (And after deployment, create a beta tag for the deployed commit of course)
Then the continuation of the deployment of this beta artifact to prod could be guarded by the already mentioned prod env approval.

Example for a workflow like this:
1. Branch `release/2026-05-05.1` is created from `master`
2. The "Release" workflow automatically kicks off and deploys this branch to the Beta env
3. After deployment the tag `2026-05-05-beta.1` is added to the commit
4. The "Release" workflow runs E2E Tests agains the deployed beta version reporting to Testiny
5. QA finds a bug in the beta release and declines approval for the prod deployment
6. A hotfix for the bug is created on `master`
7. The hotfix PR is manually cherry picked to the branch `release/2026-05-05.1` (ideally via a PR to allow adding multiple hotfixes at once + reviews)
8. The "Release" workflow once again kicks in and deploys the new version of the branch to Beta, creating the tag `2026-05-05-beta.2` on the commit on the release branch after deployment
9. E2E tests are triggered for beta once again and QA runs manual tests
10. No bugs were found and QA approves the deployment to prod
11. After successful deployment the tag `2026-05-05-prod.1` is added to the last commit on the release branch

> Note: My idea would be to use the "-prod" tag on the commit to prevent the "Release" workflow from running again for the release branch after it has been shipped to prod

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- The production workflow promotes the matching beta artifact and does not rebuild.
+- After successful production deployment, the workflow creates production tag `YYYY-MM-DD-production.N`.
+
+We will keep continuous deployment to Edge from `dev`, but deployment execution will be gated by GitHub Environment approval.
+
+Hotfix propagation will be automated and GitHub-only:
+
+- Hotfixes are merged to `master`.
+- A dedicated workflow applies `git cherry-pick -x` to the active `release/YYYY-MM-DD.N` branch and to `dev`.
+- The workflow creates pull requests for both targets to keep traceability and reviewability.
```
markbrockhoff: If I'm not mistaken the remaining envs after deployment would be Dev, Beta & Prod right? Please correct me if I'm wrong but that would get rid of the current master env (which I find reasonable), but that to me raises the question if the split between the dev and master branch should stay and if so for which reasons.

I feel like we could safe on quiet some complexity by switching to a trunk based workflow where `master` would be our new `dev`. Then every commit on master could go to the edge env and beta & prod are reserved for the release branches. Cherry picking to one of the release branches would be a manual effort through a PR giving us traceability and reviews on hotfixes. But we would no longer have to also update dev from master reducing extra steps where things could go wrong and cognitive load.

I'd ask the simple question of who is the env / stage for and currently I only see value in 3 of them:
- Edge -> Immediate response for Devs / Internals who like living on the edge
- Beta -> Mainly for QA to verify everything works before going to prod
- Prod -> Customers use this, everything should be battle tested once it arrives here. Bugs happening here need to be tackled in the next release

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+For on-premise distribution, we will maintain a separate long-term-support line with a lower scheduled cadence, using explicit long-term-support tags created from approved production tags.
+
+Long-term-support hotfixes will use dedicated maintenance branches:
+
+- Each long-term-support release creates `lts/YYYY-MM-DD.N` from the corresponding approved production state.
```
markbrockhoff: Could we please define clearly when a LTS release is created? E.g. is every 4th release a LTS release? Or are they created on demand e.g. one customer is using an older version and a vulnerability needs to be patched for this version without causing breaking changes.
Because I don't see the real benefit in having both a "normal" version and a "LTS" version for the same release, furthermore having "LTS" versions could prevent people from updating early sticking to the latest LTS version until the next one comes thinking it would give them more security.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- Keep release triggers primarily tag-driven for production deployments.
+- Keep automatic Edge deployments without additional approval gates.
+- Handle hotfix propagation manually on local machines.
+- Use sequence-only release names such as `YYYY.N`.
+
+## Decision
+
+We will adopt a fully automated, GitHub-driven release process with explicit beta-to-production promotion and post-deployment production tagging.
+
+The release process will use ISO-date identifiers in UTC with same-day sequence numbers: `YYYY-MM-DD.N`.
```
markbrockhoff: I assume beta tags are then created based on the identifer of the release branch they are created from and not the date they are created on right? I feel like it would be confusing to e.g. have a branch `release/2026-05-05.1` with a commit with tag `2026-05-06-beta.2` on it simply due to a hotfix being added to it the day later.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- The workflow creates pull requests for both targets to keep traceability and reviewability.
+
+For on-premise distribution, we will maintain a separate long-term-support line with a lower scheduled cadence, using explicit long-term-support tags created from approved production tags.
+
+Long-term-support hotfixes will use dedicated maintenance branches:
+
+- Each long-term-support release creates `lts/YYYY-MM-DD.N` from the corresponding approved production state.
+- Hotfixes are merged to `master` first and then propagated via workflow to `dev` and the affected `lts/YYYY-MM-DD.N` branch with `git cherry-pick -x`.
+- Long-term-support hotfix deployments are triggered from the long-term-support branch, not from `dev` or active cloud release branches.
+- A new long-term-support tag is created only after successful long-term-support hotfix deployment using numeric sequence increments, for example `YYYY-MM-DD-lts.0`, `YYYY-MM-DD-lts.1`, `YYYY-MM-DD-lts.2`.
```
markbrockhoff: Where are lts branches supposed to be deployed? If QA is supposed to review the release of a lts hotfix it would need to be deployed somewhere, if there is more than 1 LTS version actively supported this would require 1 env per LTS release to stay active

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
thisisamir98: I'm also wondering what is the value of keeping a `dev` branch if we would be creating release branches from master? we could just have a trunk based strategy as @markbrockhoff suggested and cut release branches from master.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
thisisamir98: @markbrockhoff they are created on demand.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
thisisamir98: @markbrockhoff I believe they are shipped to customers who have on-premise wire installed.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: I added this on purpose as a discussion point. My reasoning is: `Edge` currently serves internal dogfooding and when we deploy broken changes there, the feedback volume is immediate and loud. An approval gate could reduce that risk by adding a pause before rollout. That said, I also see the downside clearly: if `Edge` deploys on every dev commit, approvals can become pure click-through noise and lose value quickly. Unlike production there is no clear owner role (like quality assurance) to reliably review each deployment request.

I am personally in favor of continuous deployment, so I am not attached to keeping an `Edge` approval gate. I included it deliberately so the team can make an explicit tradeoff decision: maximum flow (continuous Edge) vs. additional guardrail (manual approval).

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: Exactly right. Beta tags should be anchored to the release identifier, not the calendar day of deployment.

So for `release/2026-05-05.1` tags should stay in that family:

- `2026-05-05-beta.1`
- `2026-05-05-beta.2`
- ...

and final:
- `2026-05-05-production.1`

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: Your understanding is correct and I agree that `beta` should not require a manual click.

A better model is:

1. Creating or updating `release/YYYY-MM-DD.N` automatically triggers the `Release` workflow
2. Workflow deploys to beta automatically
3. After successful beta deploy, create `YYYY-MM-DD-beta.X` on the deployed commit (`X` increments per beta attempt for that release identifier)
4. Run E2E against beta and report to Testiny
5. Production deployment stays gated by `wire-webapp-prod` environment approval
6. After successful production deployment, create `YYYY-MM-DD-production.N` on the promoted commit

For hotfixes:

- merge hotfix to `master`
- propagate to `release/YYYY-MM-DD.N` via pull request/cherry-pick workflow
- release workflow re-runs automatically
- creates next beta tag (`beta.2`, `beta.3`, ...)

I also agree with your note: once a commit on the release branch has a -prod tag, we should prevent unnecessary repeat release execution for that exact commit (idempotency guard). Important implementation detail: do this with explicit workflow conditions/concurrency, not by relying only on tag events, to avoid accidental trigger loops.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: I think this is a very good challenge. If we remove the `master` environment and keep only Edge, Beta, and Prod, then the branch model should be questioned as well.

A trunk-based setup (~~master~~ `main` as the integration branch) could reduce complexity significantly:

- no `dev` ➡️  `master` promotion step
- fewer synchronization mistakes
- simpler mental model and fewer operational handoffs

In that model:

- `main` (trunk) deploys continuously to Edge
- `release/YYYY-MM-DD.N` is cut from `main` for stabilization
- Beta and Prod are only for release branches
- hotfixes are merged to `main` and cherry-picked to active release branches via pull request/workflow for traceability and review

This removes the extra "also update `dev`" step and lowers cognitive load.

Main tradeoff:

- trunk increases the need for strong discipline on feature flags and merge quality, because `main` is always close to dogfooding.
- if we are comfortable with that, trunk-based is likely the cleaner long-term direction.

So I support discussing trunk-based as default, with Edge/Beta/Prod as the three meaningful environments **BUT** I don't know which environments/stages our quality assurance department is using. So maybe something for @iskvortsov to answer?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: > [@markbrockhoff](https://github.com/markbrockhoff) I believe they are shipped to customers who have on-premise wire installed.

Exactly. The are not deployed by us at all. The integration team is "taking" the branch and deploy it to on-premise customers.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: Well you know my opinion now, what does the rest of the team think? :)

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: Yeah that's the semi continuous flow I was imagining 👍 
We should also be careful of release triggers for older releases e.g. "LTS" ones we should not implicitly deploy an older version to beta just because a hotfix was added for this LTS version.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: I'd like to point out that with our current setup every commit to dev is already deployed to edge right? So there wouldn't be a real change to the dogfooding process. The main question would be: Is anyone currently relying on the master env which would become obsolete with this transition

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: Then could it maybe make sense to not introduce the term "LTS" at all (for the reasons I already mentioned)?
How about we allow every release branch to become a "LTS" release by simply adding new commits to it after the a prod release already happened from it. E.g. `release/06-05-2026.1` already has a release with tag `06-05-2026.1-prod.1` on it and we have to add a "patch" for it. Then we could just create a tag like `06-05-2026.1-prod.2` on it and release this artifact. (Of course doing beta testing etc beforehand :))
This would remove one layer from the kind of releases we have, there wouldn't be "normal" vs "LTS" there would just be "releases". What do you think?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: Ok, if it's a deliberate decision to proceed with "LTS" releases like this, but I think we should note this down as part of the ADR for documentation purposes :)

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
zskhan: I would like to do continuous Edge deployment from dev branch but with an option to rollback to some stable state.
If we need to wait for the approval then we are not utilizing the internal dog fooding potential to find bugs but at the same time users should not be blocked from using edge due to broken deployment.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+Sequence numbers are determined by workflow automation from existing tags for the same UTC date, and workflows must fail on sequence conflicts.
+
+The process is:
+
+- A release branch is cut from `master` as `release/YYYY-MM-DD.N`.
+- `Release Beta` is triggered manually through GitHub Actions using `release_identifier`.
+- Quality assurance validates the deployment in beta.
+- `Release Production` is triggered manually through GitHub Actions using the same `release_identifier`.
+- Production deployment is protected by a GitHub Environment approval gate for quality assurance reviewers.
+- Only quality assurance members may approve the production environment and trigger the production release workflow.
```
zskhan: If QA members can deploy on Production, who will be responsible for observability like watching grafana logs, etc?
Also what about roll back strategy here? Rollback should also be github action based so QA can also rollback the deployment in case of issues?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: In continuous deployment the system usually treats the latest commit on the deploy branch (`main`) as the desired state. If you manually roll production back to version N-1 but  `main` still contains version N, any later merge or rerun will push production away from the rollback target.

Usually you revert the bad pull request on  `main`. This creates a new commit that represents the known-good state. Or freeze production deployments after rollback. But this requires manual approval afterwards and increases complexity.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: I agree we should split approval authority from operational ownership. My intent is not "QA owns production operations.". QA should own the go/no-go quality gate while engineering owns production rollout safety.

So the model should be:

- Quality assurance: approves that the candidate passed verification.
- Engineering release owner: executes/oversees production deployment and post-deploy observability checks (Grafana, logs, alerts).
- Rollback: must be GitHub Action-based with clear permissions and runbook.

For rollback I suggest we explicitly add:

- a dedicated manual workflow: `Rollback Production`
- input: production tag to redeploy (or latest known good tag)
- restricted permissions (engineering release owners)
- mandatory reason field + Wire notification (+ incident ticket reference?)

So yes, let's make rollback first-class in the ADR and define operational ownership explicitly

markbrockhoff: Good point, but what shall be the source of truth for what's currently running on production? E.g. do we remove the "-prod" tag again or something like that? Because if we just "re-deploy" the previous prod version git would no longer be the source of truth and it's no longer visible at one glance what is running on prod.
We don't have infrastructure as code at the moment right? Because then we could do the rollback in form of updating e.g. the yaml files of k8s. I think it could make sense to decouple deployment and delivery so our release branches just "deliver" the production release artifact while what is running in prod is defined somewhere else :)

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: > In continuous deployment the system usually treats the latest commit on the deploy branch (`main`) as the desired state. If you manually roll production back to version N-1 but  `main` still contains version N, any later merge or rerun will push production away from the rollback target.
> 
> Usually you revert the bad pull request on  `main`. This creates a new commit that represents the known-good state. Or freeze production deployments after rollback. But this requires manual approval afterwards and increases complexity.

That's also what I'm familiar with, IHMO people should be aware that a "dev" stage is meant to be broken. (Of course not intentionally but this is what it's for, to catch issues before they go to the next stage)
I don't think the additional approvals for a rollback should be a big issue, if we keep them as pure isolated rollback PRs.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- Keep automatic Edge deployments without additional approval gates.
+- Handle hotfix propagation manually on local machines.
+- Use sequence-only release names such as `YYYY.N`.
+
+## Decision
+
+We will adopt a fully automated, GitHub-driven release process with explicit beta-to-production promotion and post-deployment production tagging.
+
+The release process will use ISO-date identifiers in UTC with same-day sequence numbers: `YYYY-MM-DD.N`.
+Sequence numbers are determined by workflow automation from existing tags for the same UTC date, and workflows must fail on sequence conflicts.
```
arjita-mitra: workflows must fail on sequence conflicts—what exactly triggers this and how should teams respond when it happens?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- Production deployment is protected by a GitHub Environment approval gate for quality assurance reviewers.
+- Only quality assurance members may approve the production environment and trigger the production release workflow.
+- The production workflow promotes the matching beta artifact and does not rebuild.
+- After successful production deployment, the workflow creates production tag `YYYY-MM-DD-production.N`.
+
+We will keep continuous deployment to Edge from `dev`, but deployment execution will be gated by GitHub Environment approval.
+
+Hotfix propagation will be automated and GitHub-only:
+
+- Hotfixes are merged to `master`.
```
arjita-mitra: Cherry-Pick Conflicts: How should the workflow handle cherry-pick conflicts? Should it fail, create an issue, or open a manual PR?

Hotfix Traceability: How will the team track which hotfixes are in which release/LTS branches? Should there be a dashboard or report?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+The release process will use ISO-date identifiers in UTC with same-day sequence numbers: `YYYY-MM-DD.N`.
+Sequence numbers are determined by workflow automation from existing tags for the same UTC date, and workflows must fail on sequence conflicts.
+
+The process is:
+
+- A release branch is cut from `master` as `release/YYYY-MM-DD.N`.
+- `Release Beta` is triggered manually through GitHub Actions using `release_identifier`.
+- Quality assurance validates the deployment in beta.
+- `Release Production` is triggered manually through GitHub Actions using the same `release_identifier`.
+- Production deployment is protected by a GitHub Environment approval gate for quality assurance reviewers.
```
arjita-mitra: What metrics/alerts should trigger if the release process stalls or fails?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- Only quality assurance members may approve the production environment and trigger the production release workflow.
+- The production workflow promotes the matching beta artifact and does not rebuild.
+- After successful production deployment, the workflow creates production tag `YYYY-MM-DD-production.N`.
```
arjita-mitra: Should old release branches and tags be archived or deleted after a certain period?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+Proposed
+
+## Context
+
+The current release process is mostly driven by branch and tag pushes, with production deployment triggered by pre-existing production tags. That creates two main problems:
+
+- Production tags are created before production deployment, which makes the tag represent intent rather than a verified deployment result.
+- The process does not provide a clean and auditable promotion flow from beta validation to production deployment.
+
```
e-maad: We can add more context here, for example about beta environment and Intentions around beta deployment. Also for on-prem releases to customers.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
e-maad: I am in a favor of keeping continuous edge deployment. For other companies it may not make much sense but for us, we are the beta users for our own product.

It is known that the `Edge` environment is there to be broken for good reasons.
All of the `Edge` users have prod app installed and they are fine if the app breaks down completely.

We don't need to rollback any change if we kept it hidden behind feature flag.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+  productionTag -->|Create long-term-support branch and snapshot| longTermSupportBranch
+  longTermSupportBranch -->|Create long-term-support tag| longTermSupportTag
+
+  hotfixCommit[Hotfix commit merged to master] -->|Cherry-pick -x via workflow| releaseBranch
+  hotfixCommit -->|Cherry-pick -x via workflow| developmentBranch
+  hotfixCommit -->|Cherry-pick -x via workflow| longTermSupportBranch
+  longTermSupportBranch -->|Deploy long-term-support hotfix and tag| longTermSupportHotfixTag
+```
+
+### Alternatives
```
e-maad: Should we document tradeoffs/ pros & cons for each alternative?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+## Decision
+
+We will adopt a fully automated, GitHub-driven release process with explicit beta-to-production promotion and post-deployment production tagging.
+
+The release process will use ISO-date identifiers in UTC with same-day sequence numbers: `YYYY-MM-DD.N`.
+Sequence numbers are determined by workflow automation from existing tags for the same UTC date, and workflows must fail on sequence conflicts.
+
+The process is:
+
+- A release branch is cut from `master` as `release/YYYY-MM-DD.N`.
```
e-maad: What about dev branch? When are we going to merge from dev -> master?
Should we add it in the flow?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+- `Release Beta` is triggered manually through GitHub Actions using `release_identifier`.
+- Quality assurance validates the deployment in beta.
+- `Release Production` is triggered manually through GitHub Actions using the same `release_identifier`.
+- Production deployment is protected by a GitHub Environment approval gate for quality assurance reviewers.
+- Only quality assurance members may approve the production environment and trigger the production release workflow.
+- The production workflow promotes the matching beta artifact and does not rebuild.
+- After successful production deployment, the workflow creates production tag `YYYY-MM-DD-production.N`.
+
+We will keep continuous deployment to Edge from `dev`, but deployment execution will be gated by GitHub Environment approval.
```
e-maad: What do you mean by `GitHub Environment` approval?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
e-maad: If I am not mistaken the word LTS is used for the a release targeted to a specific customer. We may have two same day release for different customers and later they both diverged due different requirements.

e-maad: We should come up with a better name for these releases.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
e-maad: Yes, as far as I remember we had 20 users within the company who uses master (beta) environment. The has the same purpose as `Edge` but only targeting 'to be released' changes.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+  releaseBetaWorkflow --> betaDeployment --> qualityAssuranceValidation --> releaseProductionWorkflow
+  releaseProductionWorkflow --> productionApproval --> productionDeployment --> createProductionTag
+  createProductionTag --> createLongTermSupportRelease --> createLongTermSupportTag
+  createLongTermSupportTag --> applyLongTermSupportHotfix --> longTermSupportValidation
+  longTermSupportValidation --> releaseLongTermSupportHotfix --> longTermSupportDeployment --> createLongTermSupportHotfixTag
+
+  developmentPush --> edgeApproval --> edgeDeployment
+```
+
+## Consequences
```
e-maad: This word is commonly used for negative impacts.
Maybe we can use **Outcome**, **Result** or **Beneficial impact**.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+  edgeApproval[GitHub Environment approval<br/>wire-webapp-edge]
+  edgeDeployment[Deploy to wire-webapp-edge]
+
+  releaseBetaWorkflow --> betaDeployment --> qualityAssuranceValidation --> releaseProductionWorkflow
+  releaseProductionWorkflow --> productionApproval --> productionDeployment --> createProductionTag
+  createProductionTag --> createLongTermSupportRelease --> createLongTermSupportTag
+  createLongTermSupportTag --> applyLongTermSupportHotfix --> longTermSupportValidation
+  longTermSupportValidation --> releaseLongTermSupportHotfix --> longTermSupportDeployment --> createLongTermSupportHotfixTag
+
+  developmentPush --> edgeApproval --> edgeDeployment
```
e-maad: We should keep the LTS flow separate from regular prod releases.
For LTS it takes an approved production state but then diverge and get hotfixes cherry-picked. Not all the prod releases become LTS.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: Hmm, I see... I think I have an idea what you mean. Maybe this point is something to discuss somewhere else ^^

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: IMHO exactly this is the reason hotfix propagation can't be fully automated. Backporting from master to dev maybe but I belive for hotfixing releases it's probably best to manually cherry pick and merge them to the release branch via PR.
If we decide and manage to get rid of the "dev" branch entirely this automation would probably not be needed. (See discussion hier up)

Maybe we could only allow rebased fast forward merges to the release branches to try and keep traceability for the hotfixes pulled from master?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: There's a dedicated release captain right? So I'd guess it would be the responsibility of this person to have an eye on the release process to pass and make sure it completes and doesn't stall.
For failures maybe a message into a group chat could support, but I don't think we can do anything about stalling. Although I'd hope the release has high enough priority so it doesn't get forgotten ^^'

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: I'm also a bit worried about cluttering the repo with too many open branches, but I'm not sure the cleanup is worth the effort. What we could do is simply delete the branch after the release tag was added to the last commit. Branches in git are effectively also just "markers" for a specific commit, so once the tag's on there the branch could be restored at any time. But I feel like we'd loose some sense of easy traceability by doing this... :/

Also the deletion of tags won't be possible since I think we should not remove part of the release history

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
markbrockhoff: I assume [this](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/review-deployments) is meant. Basically in Github Actions you can define multiple environments and require an approval by a certain group of people before an action is allowed to use it.

E.g. the release Action would request access to the "prod" environment, so in order to use it it would request access from the people who can approve it. This request would stay around for up to 30 days. The reviewers (e.g. QA) can then just click the approve button for the release in the UI and it will be deployed to prod or decline it, causing the release to stop.

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
```diff
@@ -0,0 +1,121 @@
...
+
+The process is:
+
+- A release branch is cut from `master` as `release/YYYY-MM-DD.N`.
+- `Release Beta` is triggered manually through GitHub Actions using `release_identifier`.
+- Quality assurance validates the deployment in beta.
+- `Release Production` is triggered manually through GitHub Actions using the same `release_identifier`.
+- Production deployment is protected by a GitHub Environment approval gate for quality assurance reviewers.
+- Only quality assurance members may approve the production environment and trigger the production release workflow.
+- The production workflow promotes the matching beta artifact and does not rebuild.
```
iskvortsov: What about feature flags? Or are they already removed by the time we cut beta?

docs/decision-records/0002-automated-quality-assurance-gated-release-process.md
screendriver: In https://github.com/wireapp/wire-webapp/pull/21217 we agreed to use [lightweight-architecture-decision-records](https://github.com/peter-evans/lightweight-architecture-decision-records), which names it exactly like this.

screendriver:
First of all: big thank you for all the discussions, comments, and challenges on this ADR. The feedback changed the proposal quite a bit and made the process clearer.

I updated the ADR based on the discussion.

The proposal now moves from "automated release process with `dev` and `master`" to a trunk-based release process:

- `master` will be renamed to `main`
- `dev` will be retired as a long-lived branch
- `main` becomes the single trunk branch
- every merge to `main` deploys continuously to Edge
- release branches are cut from `main` as `release/YYYY-MM-DD.N`
- creating or updating a release branch automatically deploys to Beta
- Beta tags are anchored to the release identifier, not the deployment date
- end-to-end tests run against Beta and report to Testiny
- Production remains protected by GitHub Environment approval
- Production promotes the Beta-tested artifact instead of rebuilding from source
- Production tags are created only after successful Production deployment
- hotfixes are merged to `main` first, then cherry-picked through reviewed pull requests into active release or maintenance branches

I also renamed the ADR file to:

`0002-trunk-based-automated-release-process.md`

The previous name was too narrow because the ADR is no longer only about the quality assurance gate. It now also covers the branch model, `main`, retiring `dev`, automatic Beta deployment, Edge behavior, rollback, feature flags, and on-premises maintenance releases.

I replaced the "long-term-support release" wording with "on-premises maintenance release". The intent is not that every Production release becomes a long-term-support line. A maintenance line is created only when a customer-managed deployment needs a stable branch for later patches. The webapp team produces the branch, artifact, and tags; customer deployment stays outside this workflow.

I also clarified ownership:

- quality assurance owns the go/no-go quality gate
- the engineering release captain owns Production rollout, observability, and incident response
- rollback is a first-class GitHub Actions workflow owned by engineering release owners

The main tradeoff is that `main` must stay healthy enough for Edge at all times. That means stronger pull request discipline and feature flag usage. I think that tradeoff is acceptable because Edge is the dogfooding environment, while Beta and Production remain tied to release branches and explicit approval.

docs/decision-records/0002-trunk-based-automated-release-process.md
```diff
@@ -0,0 +1,161 @@
...
+- The rollback workflow deploys a previous known-good production tag or artifact.
+- Rollback is owned by engineering release owners, not quality assurance.
+- A rollback requires a reason, a Wire notification, and, when applicable, an incident reference.
+- Rollback does not delete, move, or rewrite release tags. Runtime state must be visible through GitHub deployment metadata, deployment logs, and notifications.
+
+Feature flags are required for trunk-based development:
+
+- Incomplete or risky work must be hidden behind feature flags before it is merged to `main`.
+- Release branches validate the intended feature flag state for that release.
+- Edge may expose trunk changes earlier than Beta and Production by design.
```
arjita-mitra, line 157: After successful feature branch testing do we remove the feature flag before merging to main for edge deployment?

docs/decision-records/0002-trunk-based-automated-release-process.md
```diff
@@ -0,0 +1,161 @@
...
+- On-premises maintenance branches are created only when a customer-managed release line needs maintenance after the original production release.
+
+The release identifier uses the release branch name: `YYYY-MM-DD.N`. The date part is anchored to the release branch identifier, not to the later deployment date. For example, updates to `release/2026-05-05.1` always create tags in the `2026-05-05` family, even if a hotfix is added on a later day.
+
+The cloud release process is:
+
+- Every merge to `main` deploys continuously to Edge without an approval gate.
+- Creating or updating `release/YYYY-MM-DD.N` automatically starts the release workflow.
+- The release workflow deploys the release branch to Beta.
+- After a successful beta deployment, the workflow creates a beta tag such as `YYYY-MM-DD-beta.1`, `YYYY-MM-DD-beta.2`, and so on.
```
markbrockhoff: I'd suggest to keep the "N" component of the release identifier in the tag name e.g. `YYYY-MM-DD.N-beta.1` instead of `YYYY-MM-DD-beta.1` simply to keep the release identifier consistent. This should also help with sorting of tag names if viewed in a list

docs/decision-records/0002-trunk-based-automated-release-process.md
```diff
@@ -0,0 +1,161 @@
...
+- Beta tag numbers are derived from existing beta tags for the release identifier. If concurrent workflows try to create the same tag for different commits, the later workflow must fail and be rerun after fetching the latest tags.
```
markbrockhoff, line 78: Do we even want to allow concurrency for the release workflow(s)? IMHO there should be max 1 worker running for the tasks of one stage e.g. beta at a time. This would rule out a lot of potential causes for issues without much impact on our workflow.

docs/decision-records/0002-trunk-based-automated-release-process.md
markbrockhoff: Same for the `-production` tag

docs/decision-records/0002-trunk-based-automated-release-process.md
iskvortsov: Yeah, please add clarity about feature flag removal.

screendriver: Feature flags are not automatically removed before merging to `main`.
A feature may be merged to `main` behind a disabled flag, enabled on Edge for dogfooding and later enabled for Beta/Production when it is part of the release scope.
Feature flag removal is a separate cleanup step after the feature is fully rolled out and no rollback-by-flag is needed anymore.

Reason: requiring flag removal before `main` would weaken trunk-based development. The whole point is that incomplete or risky work can land safely behind a flag.

(I will update the document regarding this)

docs/decision-records/0002-trunk-based-automated-release-process.md
screendriver: I agree. Current `YYYY-MM-DD-beta.X` is ambiguous once you have `release/YYYY-MM-DD.1` and `release/YYYY-MM-DD.2`

I will change the ADR to:

- `release/2026-05-05.1`
- `2026-05-05.1-beta.1`
- `2026-05-05.1-beta.2`
- `2026-05-05.1-production.1`

docs/decision-records/0002-trunk-based-automated-release-process.md
screendriver: I agree with the concern. We should not rely on "fail on tag conflict" as the main protection. I will update the ADR.

markbrockhoff (approved):

iskvortsov (approved):

panuschek (approved):

thisisamir98 (approved):

e-maad (approved):

arjita-mitra (approved):

JacquelineLehner (approved):

PR #21231, opened 2026-05-05
ci: only run crit flow e2e tests in PRs & move reporting to nightly runs [WPB-24893]

markbrockhoff:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24893" title="WPB-24893" target="_blank">TaskWPB-24893</a>  [Web][QA] Use E2E tests as monitoring instead of PR criteria
  </td></table>
  <br />
 

This PR attempts to unblock PRs by only running the crit flow E2E tests as part of them. Instead the nightly E2E test run will run all tests and report the results to Testiny as well as a chat group in case of failure.
Only when attempting to merge to master all E2E tests will be run to prevent regressions in releases.

The report published into the wire group will look like this:
image

The same results are also added to the description of the run in Testiny so e.g. information about the flaky tests is preserved.

### DIFF
--- a/.github/workflows/precommit-crit-flows.yml
+++ b/.github/workflows/precommit-crit-flows.yml
@@ -117,7 +117,7 @@ jobs:
     with:
       env_name: ${{ needs.build.outputs.env_name }}
       backendUrl: https://staging-nginz-https.zinfra.io/
-      grep: '@crit-flow-web'
+      grep: ${{ case(github.base_ref == "master", "", "@crit-flow-web") }} # Run all E2E tests when attempting to merge to master
     secrets:
       OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
       WEBTEAM_AWS_ACCESS_KEY_ID: ${{ secrets.WEBTEAM_AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/precommit-crit-flows.yml
+++ b/.github/workflows/precommit-crit-flows.yml
@@ -117,7 +117,8 @@ jobs:
     with:
       env_name: ${{ needs.build.outputs.env_name }}
       backendUrl: https://staging-nginz-https.zinfra.io/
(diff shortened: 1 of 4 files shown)


.github/workflows/precommit-crit-flows.yml
```diff
@@ -109,50 +109,15 @@ jobs:
...
     name: Deploy to precommit and run E2E
     if: ${{ success() && github.repository == 'wireapp/wire-webapp' }}
-    needs: [build, check_edits_to_e2e_tests]
+    needs: [build]
     uses: ./.github/workflows/precommit-slot-run.yml
     with:
       env_name: ${{ needs.build.outputs.env_name }}
       backendUrl: https://staging-nginz-https.zinfra.io/
-      grep: ${{ needs.check_edits_to_e2e_tests.outputs.tests_to_run }}
+      grep: ${{ case(github.base_ref == "master", "", "@crit-flow-web") }} # Run all E2E tests when attempting to merge to master
```
zskhan: I just read few things about the github workflow case() method.
Then I found that `github.base_ref` is only available for Pull Request and will not be available for merge queue.
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context
https://docs.github.com/en/webhooks/webhook-events-and-payloads#merge_group

Can you please verify my findings or if you already have some insights about it?

markbrockhoff: Good catch, thanks 👍 
You're right, using `github.base_ref` will be an empty string in the merge queue which would cause the merge queue to only run the crit flows even against master.
Pretty annoying that github doesn't support it for merge queue runs as well...
I updated the condition to also check the context var `github.event.merge_group.base_ref` which is the equivalent when running in the merge queue. Sadly it's not consistent so it contains the full ref instead of just the short name like it does for PRs, but still easy enough to read :/

zskhan (approved):

screendriver (approved):

PR #21236, opened 2026-05-05
fix: handle websocket connection on bg tabs [WPB-24284]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24284" title="WPB-24284" target="_blank">BugWPB-24284</a>  [Web] Tab in background doesn't receive messages or calls after a while
  </td></table>
  <br />
 

## WebSocket Sleep/Wake Reliability Hardening

This change improves WebSocket resilience after sleep/wake and unstable network transitions by hardening both transport and app-level reconnect behavior.

### Impact

- Reduces zombie open socket scenarios after sleep/wake.
- Lowers risk of duplicate active socket instances.
- Improves reconnection observability and debugging clarity.

### DIFF
--- a/apps/webapp/src/script/repositories/event/EventRepository.test.ts
+++ b/apps/webapp/src/script/repositories/event/EventRepository.test.ts
@@ -21,6 +21,7 @@ import {BackendEvent, CONVERSATION_EVENT, USER_EVENT} from '@wireapp/api-client/
 import {ConnectionState} from '@wireapp/core';
 import {WebAppEvents} from '@wireapp/webapp-events';
 import {amplify} from 'amplify';
+import {Runtime} from '@wireapp/commons';
 
 import {ClientConversationEvent} from 'Repositories/conversation/EventBuilder';
 import {Warnings} from '../../view_model/WarningsContainer';
@@ -305,19 +306,15 @@ describe('EventRepository', () => {
     });
 
     it('should setup visibilitychange listener in browser', async () => {
-      const originalIsElectron = Reflect.get(window, 'isElectron');
-      Reflect.set(window, 'isElectron', undefined);
+      jest.spyOn(Runtime, 'isElectron').mockReturnValue(false);
 
       await eventRepository.connectWebSocket(mockAccount, false, jest.fn());
 
       expect(document.addEventListener).toHaveBeenCalledWith('visibilitychange', expect.any(Function));
-
-      Reflect.set(window, 'isElectron', originalIsElectron);
     });
 
     it('should reconnect when visible and websocket is open but unhealthy', async () => {
-      const originalIsElectron = Reflect.get(window, 'isElectron');
-      Reflect.set(window, 'isElectron', undefined);
+      jest.spyOn(Runtime, 'isElectron').mockReturnValue(false);
       jest.spyOn(eventRepository, 'notificationHandlingState').mockReturnValue(NOTIFICATION_HANDLING_STATE.WEB_SOCKET);
       mockAccount.isWebsocketHealthy.mockResolvedValueOnce(false);
 
@@ -337,14 +334,11 @@ describe('EventRepository', () => {
       await Promise.resolve();
 
       expect(mockAccount.isWebsocketHealthy).toHaveBeenCalled();
-      expect(mockAccount.listen).toHaveBeenCalledTimes(2);
-
-      Reflect.set(window, 'isElectron', originalIsElectron);
+      expect(mockAccount.listen).toHaveBeenCalledTimes(1);
     });
 
     it('should not reconnect when visible and websocket is open and healthy', async () => {
-      const originalIsElectron = Reflect.get(window, 'isElectron');
-      Reflect.set(window, 'isElectron', undefined);
+      jest.spyOn(Runtime, 'isElectron').mockReturnValue(false);
       jest.spyOn(eventRepository, 'notificationHandlingState').mockReturnValue(NOTIFICATION_HANDLING_STATE.WEB_SOCKET);
       mockAccount.isWebsocketHealthy.mockResolvedValueOnce(true);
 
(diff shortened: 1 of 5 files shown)


apps/webapp/src/script/repositories/event/EventRepository.test.ts
```diff
@@ -304,14 +305,68 @@ describe('EventRepository', () => {
     });
 
     it('should setup visibilitychange listener in browser', async () => {
-      const originalIsElectron = (window as any).isElectron;
-      (window as any).isElectron = undefined;
+      const originalIsElectron = Reflect.get(window, 'isElectron');
```
screendriver: I recommend replacing `Reflect` with an explicit property override helper using `Object.defineProperty`. This keeps intent explicit and avoids leaked global mutation.

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -198,19 +205,34 @@ export class ReconnectingWebsocket {
...
+    this.logger.debug('Sending ping');
     this.hasUnansweredPing = true;
     this.send(PingMessage.PING);
   };
 
   public connect(): void {
+    this.logger.info('Initializing WebSocket connection');
     this.resetLongRunningRetrySequence();
+
+    if (this.socket && this.socket.readyState !== WEBSOCKET_STATE.CLOSED) {
```
screendriver: 1. `this.socket !== undefined` or `this.socket !== null`?
2. `close(...)` old socket before replacing `this.socket` is correct but stale async events from old socket can still run shared handlers after reassignment at libraries/api-client/src/tcp/reconnectingWebsocket.ts:236. Please detach old handlers before `close(...) `

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -248,6 +249,38 @@ export class EventRepository {
       void connect();
     };
 
+    const verifyConnectionHealthAndReconnect = async (reason: 'Focus' | 'Visibility' | 'Heartbeat') => {
```
screendriver: This function is doing state checks, health probing and reconnection orchestration in one function. Can we split this into pure explicit functions to keep behavior easier to test and reason about?

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -248,6 +249,38 @@ export class EventRepository {
...
+      }
+
+      healthCheckInProgress = true;
+      try {
+        const isHealthy = await account.isWebsocketHealthy();
+        if (!isHealthy) {
+          this.logger.warn(`${reason}: WebSocket appears unhealthy, reconnecting...`);
+          void connect();
+        }
+      } catch {
```
screendriver: Swallows the error details. When this is on purpose, ignore this comment.

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -297,11 +324,7 @@ export class EventRepository {
 
     // Heartbeat to check connection state every 30 seconds
     const heartbeatInterval = window.setInterval(() => {
-      const currentState = this.notificationHandlingState();
-      if (currentState === NOTIFICATION_HANDLING_STATE.CLOSED && navigator.onLine) {
-        this.logger.info('Heartbeat: Connection is closed and app is online, attempting reconnection...');
-        handleOnline();
-      }
+      void verifyConnectionHealthAndReconnect('Heartbeat');
```
screendriver, line 345: Heartbeat now always probes when online (including non closed states). When this is on purpose, ignore this comment.

apps/webapp/src/script/repositories/event/EventRepository.test.ts
```diff
@@ -304,14 +305,68 @@ describe('EventRepository', () => {
...
+      Reflect.set(window, 'isElectron', undefined);
 
       await eventRepository.connectWebSocket(mockAccount, false, jest.fn());
 
       expect(document.addEventListener).toHaveBeenCalledWith('visibilitychange', expect.any(Function));
 
-      (window as any).isElectron = originalIsElectron;
+      Reflect.set(window, 'isElectron', originalIsElectron);
```
screendriver: Restoration is not protected by a `finally` clause, so the test has a side effect to all other tests

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -198,19 +205,34 @@ export class ReconnectingWebsocket {
...
+      );
+      try {
+        this.socket.close(CloseEventCode.NORMAL_CLOSURE, 'Reinitializing WebSocket connection');
+      } catch (error) {
+        this.logger.warn('Failed to close existing WebSocket instance before reconnecting', error);
+      }
+    }
+
+    this.stopPinging();
     this.socket = this.getReconnectingWebsocket();
```
screendriver: New socket assignment happens immediately after closing old one. This increases stale-event risk.

apps/webapp/src/script/repositories/event/EventRepository.test.ts
```diff
@@ -304,14 +305,68 @@ describe('EventRepository', () => {
...
+      jest.spyOn(eventRepository, 'notificationHandlingState').mockReturnValue(NOTIFICATION_HANDLING_STATE.WEB_SOCKET);
+      mockAccount.isWebsocketHealthy.mockResolvedValueOnce(false);
+
+      await eventRepository.connectWebSocket(mockAccount, false, jest.fn());
+
+      const visibilityHandler = (document.addEventListener as jest.Mock).mock.calls.find(
+        ([eventName]) => eventName === 'visibilitychange',
+      )?.[1] as (() => void) | undefined;
+
+      Object.defineProperty(document, 'visibilityState', {
```
screendriver, line 327: `document.visibilityState` override should be restored after test execution, otherwise it has side effects to all other tests

apps/webapp/src/script/repositories/event/EventRepository.ts
e-maad: Yes, we will swallow the error, log it and then initialize a new connection.

apps/webapp/src/script/repositories/event/EventRepository.ts
e-maad: Added a comment for better understanding.

screendriver (approved):

PR #21240, opened 2026-05-05
fix: add null checks for call in getGrid and updateVideoGrid functions [WPB-25267]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25267" title="WPB-25267" target="_blank">BugWPB-25267</a>  [Web] Call Ending with error
  </td></table>
  <br />
 

Add null checks for call in getGrid and updateVideoGrid.
Previously it was throwing exception when the call is undefined.

### DIFF
--- a/apps/webapp/src/script/repositories/calling/videoGridHandler.ts
+++ b/apps/webapp/src/script/repositories/calling/videoGridHandler.ts
@@ -50,7 +50,7 @@ export function getGrid(call: Call | undefined) {
 }
 
 export function updateVideoGrid(call: Call | undefined, setGrid: Dispatch<SetStateAction<Grid>>) {
-  if (!call) {
+  if (call === undefined) {
     return;
   }
 
--- a/apps/webapp/src/script/repositories/calling/videoGridHandler.ts
+++ b/apps/webapp/src/script/repositories/calling/videoGridHandler.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {useEffect, useState} from 'react';
+import {Dispatch, SetStateAction, useEffect, useState} from 'react';
 
 import {useKoSubscribableChildren} from 'Util/componentUtil';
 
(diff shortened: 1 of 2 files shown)


thisisamir98 (dismissed):

apps/webapp/src/script/repositories/calling/videoGridHandler.ts
```diff
@@ -42,15 +49,21 @@ export function getGrid(call: Call) {
   };
 }
 
+export function updateVideoGrid(call: Call | undefined, setGrid: Dispatch<SetStateAction<Grid>>) {
+  if (!call) {
```
screendriver: Please avoid truthy/falsy checks whenever possible 🙏

```suggestion
  if (call === undefined) {
```

(Or use a type guard from `@sindresorhus/is`, which would be equal in this case)

apps/webapp/src/script/repositories/calling/videoGridHandler.ts
```diff
@@ -42,15 +49,21 @@ export function getGrid(call: Call) {
...
   const [grid, setGrid] = useState<Grid>(() => getGrid(call));
   const {participants, currentPage, pages} = useKoSubscribableChildren(call, ['participants', 'currentPage', 'pages']);
 
   useEffect(() => {
-    const updateGrid = () => {
-      call.updatePages();
-      setGrid(getGrid(call));
-    };
+    const updateGrid = () => updateVideoGrid(call, setGrid);
     updateGrid();
```
screendriver, line 67: Do we need this function definition at all? One line below it is immediately called. So instead we just could call `updateVideoGrid()` and don’t need another more function definition 


```suggestion
    updateVideoGrid(call, setGrid);
```

thisisamir98 (approved):

e-maad (approved):

apps/webapp/src/script/repositories/calling/videoGridHandler.ts
zskhan: The function `updateGrid` reference is needed for subscriptions below, not just the immediate call.

screendriver (approved):

PR #21257, opened 2026-05-07
feat: enable high quality blur option and update related handlers [WPB-25492]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25492" title="WPB-25492" target="_blank">TaskWPB-25492</a>  [Web] - Make High quality blur checkbox work
  </td></table>
  <br />
 

Enable the checkbox to switch between models.

By default selfie_multiclass model is used with checkbox enabled.

On disabling this checkbox it will use selfie_segmentation model.

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
@@ -96,24 +96,29 @@ export async function runSegmenter(
   attachCanvasEvents();
 
   let segmenter = await createSegmenter(canvas);
+  let currentModelPath = segmenterOptions.modelPath;
+  let restartPending = false;
 
   function restartSegmenter() {
+    restartPending = true;
+    const targetModelPath = segmenterOptions.modelPath;
     createSegmenter(canvas)
       .then(newSegmenter => {
         const oldSegmenter = segmenter;
         segmenter = newSegmenter;
+        currentModelPath = targetModelPath;
+        restartPending = false;
         oldSegmenter.close();
       })
       .catch((e: unknown) => {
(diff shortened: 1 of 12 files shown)


apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -96,25 +96,30 @@ export async function runSegmenter(
   attachCanvasEvents();
 
   let segmenter = await createSegmenter(canvas);
+  let currentModelPath = segmenterOptions.modelPath;
+  let restartPending = false;
 
   function restartSegmenter() {
```
EnricoSchw, line 117: Restarting the Segmenter results in a runtime error:

Image

We should only restart the Segmenter in a few specific cases.

I think the segmenter operates frame by frame. So, we could change the model between frame transitions.
I think MediaPipe provides a `setOptions` method for this purpose. It should work as follows:

```ts
await segmenter.setOptions({
  baseOptions: {
    modelAssetPath: "/assets/mediapipe-models/selfie_multiclass_256x256.tflite",
  },
});
```

EnricoSchw (changes requested):

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
zskhan: [e14f75](https://github.com/wireapp/wire-webapp/pull/21257/changes/e14f7594a8e69eb0adfc6aa649c5c8ed9f463b0b)

EnricoSchw (approved):
LGTM

EnricoSchw (approved):
LGTM

PR #21260, opened 2026-05-08
Inject fire-and-forget invoker into application services [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

Extend the webapp composition root so fire-and-forget invocation is created via dependency injection. The intent is to establish a single, explicit construction point for runtime services and make fire-and-forget behavior injectable/testable before exposing it through React context and migrating call sites.

### DIFF
--- a/apps/webapp/src/script/main/createApplicationServices.test.ts
+++ b/apps/webapp/src/script/main/createApplicationServices.test.ts
@@ -17,21 +17,33 @@
  *
  */
 
+import {FireAndForgetInvoker} from '@wireapp/core';
+
 import {createDeterministicWallClock} from '../clock/deterministicWallClock';
 import {createApplicationServices} from './createApplicationServices';
 
 describe('createApplicationServices', () => {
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/main/createApplicationServices.test.ts
```diff
@@ -17,21 +17,33 @@
  *
  */
 
+import {FireAndForgetInvoker} from '@wireapp/core';
+
 import {createDeterministicWallClock} from '../clock/deterministicWallClock';
 import {createApplicationServices} from './createApplicationServices';
 
 describe('createApplicationServices', () => {
   it('creates wall clock through injected dependency', () => {
```
zskhan: We can also rename this test as it is not only injecting Wall Clock

screendriver: That's right: edcfddffb0c7ee9729f0c2a56ecfceb7f4edc6a1

zskhan (approved):

PR #21263, opened 2026-05-08
ci: enable type checking for e2e_tests folder [WPB-22420]

markbrockhoff:
<table>
<td>
  </td></table>
  <br />
 

I noticed that the e2e_tests folder was excluded from type checking on CI. Since we want to catch silly mistakes early I re-enabled it and fixed the few issues which were found.

> Note: All other `.test.*` files are still excluded from type checking, but changing that is out of scope for this PR

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/modals/createConversation.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/modals/createConversation.ts
@@ -25,7 +25,7 @@ export const CreateConversationModal = (page: Page) => {
 
     await nextButton.click();
 
-    if (options.members) {
+    if (options?.members !== undefined) {
       await addMembers(options.members);
     }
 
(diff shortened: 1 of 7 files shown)


apps/webapp/test/e2e_tests/pageManager/webapp/modals/createConversation.ts
```diff
@@ -25,7 +25,7 @@ export const CreateConversationModal = (page: Page) => {
 
     await nextButton.click();
 
-    if (options.members) {
+    if (options?.members) {
```
screendriver:
```suggestion
    if (options?.members !== undefined) {
```

apps/webapp/test/e2e_tests/pageManager/webapp/modals/createConversation.ts
```diff
@@ -40,7 +40,7 @@ export const CreateConversationModal = (page: Page) => {
     await nextButton.click();
     await nextButton.click();
 
-    if (options.members) {
+    if (options?.members) {
```
screendriver:
```suggestion
    if (options?.members !== undefined) {
```

zskhan (dismissed):

screendriver (approved):

PR #21265, opened 2026-05-08
fix(core): defer MLS pending-proposals rehydration until LIVE [WPB-24348]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24348" title="WPB-24348" target="_blank">SpikeWPB-24348</a>  [Web] Investigate and reduce MLS decryption failures for MlsErrorOther (top decryption error type)
  </td></table>
  <br />
 

Stop calling initialisePendingProposalsTasks() from initClient. Persisted TaskScheduler tasks often have firing dates in the past, so they run on the next tick while the client is still draining the notification catch-up. That can advance the local MLS epoch ahead of queued application traffic and surface Wrong Epoch / MessageEpochTooOld during decryption.

Rehydrate those timers only when we transition to ConnectionState.LIVE: before resumeProposalProcessing in handleSynchronizationNotification (async marker path) and in the legacy notification stream processor (same ordering).

Add regression tests that drive createLegacyNotificationStreamProcessor directly: one checks that rehydration runs once and before LIVE is emitted; the other checks it is skipped when MLS is disabled.


libraries/core/src/account.ts
```diff
@@ -1067,6 +1074,15 @@ export class Account extends TypedEventEmitter<Events> {
...
           this.logger.info(`Resuming message sending. ${getQueueLength()} messages to be sent`);
+          // Rehydrate persisted MLS pending-proposals timers now, on the LIVE transition.
+          // Doing this earlier (e.g. in initClient) races with the notification catch-up:
+          // persisted firingDates from the previous session are often already in the past,
+          // so the underlying TaskScheduler timers fire "on next tick" and advance the local
+          // MLS epoch while we are still replaying old application messages from the stream,
+          // producing "Wrong Epoch" / "MessageEpochTooOld" decryption failures.
+          if (this.hasMLSDevice && this.service?.mls) {
+            await this.service.mls.initialisePendingProposalsTasks();
+          }
```
typfel: I don't think attempts to commit pending proposals during the catch up phase lead advance the epoch and lead to  "Wrong Epoch" errors unless it tries to re-join as some kind of error recovery. Trying to commit pending proposals while catching up is likely to fail with 'mis-stale-message' though.

This explanation is duplicated 3 times and very likely to get out of sync in the future.

Other than that I think this is good change, can't comment on technical details but idea of starting to commit proposals after we are live is good.

typfel (changes requested):

libraries/core/src/account.ts
```diff
@@ -937,6 +935,24 @@ export class Account extends TypedEventEmitter<Events> {
...
+   * **Why defer:** rehydrating during {@link Account.initClient} or while notifications are still
+   * replaying is unsafe. Persisted `firingDate` values are often in the past, so {@link TaskScheduler}
+   * fires auto-commits on the next tick while the backlog is still being applied. That usually surfaces
+   * as failing or conflicting commits (for example `mls-stale-message` / delivery rejection), not as a
+   * dependable way to “catch up”. “Wrong epoch” skew during replay is not uniquely attributable to
+   * proposal commits (recovery / rejoin paths matter too); this method exists to keep **all**
+   * rehydrated pending-proposal side effects aligned with a session that has finished ordered replay.
+   */
+  private readonly rehydrateMlsPendingProposalsTasksOnLiveTransition = async (): Promise<void> => {
+    if (this.hasMLSDevice && this.service?.mls) {
```
screendriver:
```suggestion
    if (this.hasMLSDevice && this.service?.mls !== undefined) {
```

libraries/core/src/account.test.ts
```diff
@@ -523,4 +523,95 @@ describe('Account', () => {
...
+    const setup = async (mlsService: MlsServiceStub) => {
+      const {account, apiClient} = await createAccount();
+
+      if (!account.service) {
+        throw new Error('Service is not set');
+      }
+
+      // Override the (heavy) real MLS service with a minimal stub. We only care about
+      // whether/when initialisePendingProposalsTasks is invoked, not what it does.
+      Reflect.set(account.service, 'mls', mlsService);
```
screendriver: Please don't use `Reflect` here. It bypasses type guarantees and can hide shape mismatches.

libraries/core/src/account.ts
```diff
@@ -937,6 +935,24 @@ export class Account extends TypedEventEmitter<Events> {
...
+      await this.service.mls.initialisePendingProposalsTasks();
```
screendriver: What is the failure policy if `initialisePendingProposalsTasks()` rejects here? A rejection can block the hand-off path

libraries/core/src/account.ts
```diff
@@ -957,6 +973,7 @@ export class Account extends TypedEventEmitter<Events> {
      * if the marker ID matches the current marker ID.
      */
     if (markerId === currentMarkerId) {
+      await this.rehydrateMlsPendingProposalsTasksOnLiveTransition();
```
screendriver, line 1000: Can this path run multiple times across reconnect/LIVE transitions? If yes, please confirm rehydration is idempotent (or guard it) to avoid duplicate pending-proposal timers.

libraries/core/src/account.ts
```diff
@@ -1067,6 +1084,7 @@ export class Account extends TypedEventEmitter<Events> {
       void this.notificationProcessingQueue
         .push(async () => {
           this.logger.info(`Resuming message sending. ${getQueueLength()} messages to be sent`);
+          await this.rehydrateMlsPendingProposalsTasksOnLiveTransition();
```
screendriver, line 1111: Same idempotency concern as line 976

screendriver (changes requested):

libraries/core/src/account.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21265/commits/b93e17ce663e9ca4979d72ec3ad677ee6f001953

libraries/core/src/account.ts
thisisamir98: errors are absorbed inside initialisePendingProposalsTasks, so this line should not reject.
see: https://github.com/wireapp/wire-webapp/blob/dev/libraries/core/src/messagingProtocols/mls/mlsService/mlsService.ts#L1133-L1162

libraries/core/src/account.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21265/commits/ef497ee92f2109f482b6e6880a93aeb8c5b2cd38

libraries/core/src/account.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21265/commits/ef497ee92f2109f482b6e6880a93aeb8c5b2cd38

libraries/core/src/account.ts
```diff
@@ -166,6 +167,13 @@ export class Account extends TypedEventEmitter<Events> {
   private coreCallbacks?: CoreCallbacks;
   private connectionState: ConnectionState = ConnectionState.CLOSED;
 
+  /**
+   * {@link once}-wrapped {@link MLSService.initialisePendingProposalsTasks}; recreated in
+   * {@link resetContext} (login / logout) so each logged-in session runs it at most once after MLS
+   * is available on the first catch-up → LIVE hand-off.
+   */
+  private initialisePendingProposalsTasksOnce!: () => Promise<void>;
```
screendriver: [!](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) ❓

libraries/core/src/account.ts
```diff
@@ -232,6 +240,13 @@ export class Account extends TypedEventEmitter<Events> {
     });
 
     this.logger = LogFactory.getLogger('@wireapp/core/Account');
+    this.initialisePendingProposalsTasksOnce = this.createInitialisePendingProposalsTasksOnce();
+  }
+
+  private createInitialisePendingProposalsTasksOnce(): () => Promise<void> {
+    return once(async () => {
+      await this.service!.mls!.initialisePendingProposalsTasks();
```
screendriver: [!](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) ❓

libraries/core/src/account.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21265/commits/77cad161692c0af201b7b94388344ac5255cd997

libraries/core/src/account.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21265/commits/a231d92f12597a7c2d85c076ee3ab35ae314025e

libraries/core/src/account.ts
```diff
@@ -232,6 +240,18 @@ export class Account extends TypedEventEmitter<Events> {
     });
 
     this.logger = LogFactory.getLogger('@wireapp/core/Account');
+    this.initialisePendingProposalsTasksOnce = this.createInitialisePendingProposalsTasksOnce();
+  }
+
+  private createInitialisePendingProposalsTasksOnce(): () => Promise<void> {
+    return once(async () => {
+      if (!this.service?.mls) {
```
screendriver:
```suggestion
      if (this.service?.mls === undefined) {
```

screendriver (approved):

PR #21288, opened 2026-05-11
feat: implement Canvas2D renderer [WPB-25236]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25236" title="WPB-25236" target="_blank">StoryWPB-25236</a>  [Web] Add 2d Canvas Render Pipeline
  </td></table>
  <br />
 

Implement Canvas2D renderer for virtual background processing and fallback to CPU when WebGL is unavailable.
Also in case of Canvas2D, use selfie_segmenter model so that it works smoothly but the user will still have the option to change the model using "Enable high quality blur" checkbox.

### DIFF
--- a/apps/webapp/src/script/auth/util/test/TestUtil.tsx
+++ b/apps/webapp/src/script/auth/util/test/TestUtil.tsx
@@ -206,7 +206,7 @@ export const buildMediaDevicesHandler = () => {
 export const buildCallingRepository = () => {
   const controller: BackgroundEffectsController = {
     getQuality: jest.fn(),
-    getCapabilityInfo: jest.fn(),
+    getCapabilityInfo: jest.fn(() => ({webgl2: undefined})),
   } as unknown as BackgroundEffectsController;
   const backgroundEffectsHandler = new BackgroundEffectsHandler(controller);
 
--- a/apps/webapp/src/script/auth/util/test/TestUtil.tsx
+++ b/apps/webapp/src/script/auth/util/test/TestUtil.tsx
@@ -206,7 +206,7 @@ export const buildMediaDevicesHandler = () => {
 export const buildCallingRepository = () => {
   const controller: BackgroundEffectsController = {
     getQuality: jest.fn(),
-    getCapabilityInfo: jest.fn(() => ({webgl2: undefined})),
+    getCapabilityInfo: jest.fn(() => ({webgl2: false})),
   } as unknown as BackgroundEffectsController;
   const backgroundEffectsHandler = new BackgroundEffectsHandler(controller);
 
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
@@ -29,7 +29,7 @@ import {
 } from 'Repositories/media/backgroundEffects/helper/metrics';
 import {createRestartQueue} from 'Repositories/media/backgroundEffects/helper/restartQueue';
 
-import {Canvas2DRenderer, createCanvas2DRenderer} from './canvas2DRenderer';
+import {Canvas2dRenderer, createCanvas2DRenderer} from './canvas2dRenderer';
 import {VideoFilter} from './filter';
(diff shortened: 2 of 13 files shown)


apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
```diff
@@ -102,6 +104,14 @@ export class BackgroundEffectsHandler {
     backgroundEffectsStore.getState().setPreferredEffect(this.readPreferredBackgroundEffectFromStore());
     backgroundEffectsStore.getState().setLastVirtualBackgroundId(this.readLastVirtualBackgroundIdFromStore());
 
+    // On CPU-only devices (no WebGL2) default to the lighter model
+    if (
+      !is.nullOrUndefined(controller.getCapabilityInfo) &&
+      is.nullOrUndefined(controller.getCapabilityInfo().webgl2)
```
screendriver: This condition looks wrong for the CPU-only case. `webgl2` is a boolean, so devices without WebGL2 should report `false`, not `undefined`. With the current check, this branch never runs for the real fallback case and `isHighQualityBlurEnabled` stays at its default `true`

I think this should check `controller.getCapabilityInfo().webgl2 === false` instead.

apps/webapp/src/script/auth/util/test/TestUtil.tsx
```diff
@@ -197,7 +197,7 @@ export const buildMediaDevicesHandler = () => {
 export const buildCallingRepository = () => {
   const controller: BackgroundEffectsController = {
     getQuality: jest.fn(),
-    getCapabilityInfo: jest.fn(),
+    getCapabilityInfo: jest.fn(() => ({webgl2: undefined})),
```
screendriver: This test stub is masking the bug below by returning `{ webgl2: undefined }`. In production `getCapabilityInfo().webgl2` is typed as a boolean, so the CPU-only path is `false` not `undefined`.

Can we change the stub to `{ webgl2: false }` and add/assert the actual fallback behaviour? Otherwise the tests validate a state that the real code should never produce.

apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
```diff
@@ -153,8 +163,13 @@ export class BackgroundEffectsHandler {
     }
 
     try {
+      const modelPath = backgroundEffectsStore.getState().isHighQualityBlurEnabled
```
screendriver, line 163: There is still a gap between capability-based model selection and the runtime fallback in `runSegmenter()`. We choose the model here once at startup but `runSegmenter()` can later fall back from WebGL to Canvas2D/CPU if `new WebGLRenderer(canvas)` throws. In that path we keep the already-selected model, so a device that advertises WebGL2 but fails actual renderer init will still run the wrong one.

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -27,34 +28,36 @@ import {
...
 } from 'Repositories/media/backgroundEffects/helper/metrics';
 
+import {Canvas2DRenderer, createCanvas2DRenderer} from './canvas2DRenderer';
 import {VideoFilter} from './filter';
 import {WorkerProcessVideoTrackOptions} from './options';
 import {WebGLRenderer} from './renderer';
 
 export let segmenterOptions = {} as WorkerProcessVideoTrackOptions;
 
+let useGPU = true;
```
screendriver: `useGPU` becoming module-level **mutable** state worries me. `runSegmenter()` mutates it per pipeline instance but `createSegmenter()` and the frame processing path both read it later. If two `runSegmenter()` calls overlap, one instance can flip the other between GPU and CPU behaviour and we have some kind of race condition here.

I recommend making this per-run state inside `runSegmenter()` and passing it explicitly into `createSegmenter()` instead of storing it globally.

screendriver (changes requested):

screendriver (dismissed):

e-maad (approved):

PR #21292, opened 2026-05-12
Make boolean checks explicit without changing how the app behaves [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

This change makes our code stricter and easier to reason about by removing implicit truthy/falsy checks


apps/server/routes/root.ts
```diff
@@ -37,7 +37,7 @@ async function addGeoIP(req: Request) {
 
     const lookup = await maxmind.open<CountryResponse>(countryDatabasePath);
     const result = lookup.get(ipAddress);
-    if (result) {
+    if (result !== undefined && result !== null) {
```
zskhan, line 40: Can we use @sindresorhus/is here?
```suggestion
    if (!is.nullOrUndefined(result)) {
```

PR #21293, opened 2026-05-12
test: implement read receipt e2e tests [WPB-19965]

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19965" title="WPB-19965" target="_blank">TaskWPB-19965</a>  [Web/QA] Write the Read Receipts regression tests in Playwright
  </td></table>
  <br />
 

Add new test suite for read receipts.

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
@@ -64,7 +64,7 @@ test.describe('Read Receipts', () => {
     await expect(messageWithImage).toBeVisible();
 
     const messageWithReadReceipt = userBPages.conversation().getMessage({sender: userB});
-    await expect(await userAPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
+    await expect(await userBPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
   });
 
   test('I want to see read receipts for assets: audio', {tag: ['@TC-3553', '@regression']}, async ({createPage}) => {
@@ -97,7 +97,7 @@ test.describe('Read Receipts', () => {
     await expect(messageWithAudio).toBeVisible();
 
     const messageWithReadReceipt = userBPages.conversation().getMessage({sender: userB});
-    await expect(await userAPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
+    await expect(await userBPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
   });
 
   test('I want to see read receipts for assets: video', {tag: ['@TC-3554', '@regression']}, async ({createPage}) => {
@@ -130,7 +130,7 @@ test.describe('Read Receipts', () => {
     await expect(messageWithVideo).toBeVisible();
 
     const messageWithReadReceipt = userBPages.conversation().getMessage({sender: userB});
-    await expect(await userAPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
+    await expect(await userBPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
   });
 
   test('I want to see read receipts for assets: file', {tag: ['@TC-3555', '@regression']}, async ({createPage}) => {
@@ -163,7 +163,7 @@ test.describe('Read Receipts', () => {
     await expect(messageWithFile).toBeVisible();
 
     const messageWithReadReceipt = userBPages.conversation().getMessage({sender: userB});
-    await expect(await userAPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
+    await expect(await userBPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
   });
 
   test(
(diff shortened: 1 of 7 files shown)


apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await userAComponents.conversationSidebar().preferencesButton.click();
+      await userAPages.account().readReceiptsCheckbox.click();
+      await userAComponents.conversationSidebar().allConversationsButton.click();
+
+      await userBComponents.conversationSidebar().preferencesButton.click();
+      await userBPages.account().readReceiptsCheckbox.click();
+      await userBComponents.conversationSidebar().allConversationsButton.click();
+
+      const conversationName = 'Group Conversation';
+      await createGroup(userAPages, conversationName, [userB]);
```
markbrockhoff: 🔧 If userB doesn't use the created group don't add him to it. This avoids issues with e.g. wire automatically switching to it upon creation causing potential race condition and flake ;)
```suggestion
      await createGroup(userAPages, conversationName, []);
```

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await userAPages.conversationList().getConversation(conversationName).open();
+
+      await userBPages.conversationList().getConversation(userA.fullName, {protocol: 'mls'}).open();
+
+      switch (tag) {
+        case '@TC-3552':
```
markbrockhoff: 🔧 Using the test ids as plain strings to alter the test here is a bit brittle, if e.g. a test id changes or one is removed. Maybe it would be safer to add e.g. a `sendMessage(page: Page, api: API)` callback to the test cases and just call this function here. This way the differentiating action between these tests would be colocated with the title and ID.

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await expect(receivedMessage).toBeVisible();
+
+      const messageWithReadReceipt = userBPages.conversation().getMessage({sender: userB});
+      await expect(await userBPages.conversation().getMessageReadReceipt(messageWithReadReceipt)).toBeVisible();
+    });
+  }
+
+  // TODO: [WPB-25618] read receipt is not visible on location share
+  // TODO: unskip this test when Bug Ticket [WPB-25618] is resolved
+  test.skip(
```
markbrockhoff: Tipp: You can still include this test in the parameterized one, even if it is skipped. You can just skip the one test based on a condition by calling `test.skip()` inside it. E.g.:

```ts
test("Some test", () => {
  test.skip(tag === "@TC-1234", "This test case is broken due to WPB-12345");
});
```

Docs: https://playwright.dev/docs/api/class-test#test-skip

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await userBComponents.conversationSidebar().preferencesButton.click();
+      await userBPages.account().readReceiptsCheckbox.click();
+      await userBComponents.conversationSidebar().allConversationsButton.click();
+
+      // User A turns on read receipts on second device
+      await userAComponents2.conversationSidebar().preferencesButton.click();
+      await userAPages2.account().readReceiptsCheckbox.click();
+      await userAComponents2.conversationSidebar().allConversationsButton.click();
+
+      // User A sees the modal about read receipts being enabled
```
markbrockhoff:
```suggestion
      // User A sees the modal about a new device being added
```

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+      await userAComponents.conversationSidebar().preferencesButton.click();
+      await expect(userAModals.newDevice().modal).toBeVisible();
+      await userAModals.newDevice().actionButton.click();
+
+      // to see the enabled read receipt modal, User A needs to switch to another settings tab
```
markbrockhoff: Maybe it would be good to open the settings once right after the second device was added to get the first modal out of the way, then go back to the conversations. Also maybe add an assertion for the little green "bubble" on the settings icon indicating that settings changed. Otherwise you might open the settings to early, before the change triggering the modal was synced.

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
JacquelineLehner: changed the flow ✅

apps/webapp/test/e2e_tests/specs/ReadReceipts/readReceipts.spec.ts
JacquelineLehner: changed it ✅

markbrockhoff (dismissed):

arjita-mitra (dismissed):

markbrockhoff (approved):

screendriver (approved):

PR #21297, opened 2026-05-13
Remove unused method from MLS service [WPB-24348]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24348" title="WPB-24348" target="_blank">SpikeWPB-24348</a>  [Web] Investigate and reduce MLS decryption failures for MlsErrorOther (top decryption error type)
  </td></table>
  <br />
 

This method is not used anywhere in the codebase at all.


screendriver (approved):

typfel:
@bardiawire is something else doing this then because you need to do key material renewals?

PR #21304, opened 2026-05-13
feat: update de-DE locale links [WPB-23924]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23924" title="WPB-23924" target="_blank">BugWPB-23924</a>  [Web] Terms of use and privacy policy have wrong link in settings
  </td></table>
  <br />
 

This PR strengthens locale handling and improves external legal URL management, with a particular focus on German users.

### **Key updates**

- Introduced a strongly typed SupportedLocale type to replace generic string locale usage, improving type safety and consistency across locale-related logic.
- Updated locale configuration utilities and selection functions to use the new locale type throughout the codebase.
- Added dedicated German legal document URLs (PRIVACY_POLICY_URL_DE and TERMS_OF_USE_URL_DE).
- Enhanced external route handling so users with the de-DE locale automatically receive German-specific privacy policy and terms of use links.
- Simplified external URL usage by consolidating terms-of-use handling into a single termsOfUse property that dynamically resolves the correct localized URL.

### DIFF
--- a/apps/webapp/src/script/auth/supportedLocales.ts
+++ b/apps/webapp/src/script/auth/supportedLocales.ts
@@ -17,44 +17,7 @@
  *
  */
 
-export type SupportedLocale =
-  | 'ar-SA'
-  | 'cs-CZ'
-  | 'da-DK'
-  | 'de-DE'
-  | 'el-GR'
-  | 'en-US'
-  | 'es-ES'
-  | 'et-EE'
-  | 'fa-IR'
-  | 'fi-FI'
-  | 'fr-FR'
-  | 'hi-IN'
-  | 'hr-HR'
-  | 'hu-HU'
-  | 'id-ID'
-  | 'it-IT'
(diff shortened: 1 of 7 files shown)


apps/webapp/src/script/auth/supportedLocales.ts
```diff
@@ -17,7 +17,44 @@
...
+  | 'sk-SK'
+  | 'sl-SI'
+  | 'sr-SP'
+  | 'sv-SE'
+  | 'tr-TR'
+  | 'uk-UA'
+  | 'zh-CN'
+  | 'zh-TW';
+
+export const Locales: readonly SupportedLocale[] = [
```
screendriver: Can’t we just use `as const` here and export the inferred type? This would prevent us from declaring everything twice

screendriver (approved):

zskhan (approved):

PR #21305, opened 2026-05-13
fix: disable retry for 503 status code [WPB-25507]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25507" title="WPB-25507" target="_blank">BugWPB-25507</a>  Webapp Login issue on custom backend after 5.30 update
  </td></table>
  <br />
 

This pull request refines the HTTP client’s incremental retry backoff behavior by preventing retries for requests that fail with a 503 Service Unavailable response. Previously, all 5xx responses were treated as retryable; the updated logic now excludes 503 errors from retry attempts. Associated tests were also updated to reflect the new behavior.

### Retry backoff policy updates

- Updated shouldRetryWithIncrementalBackoff in incrementalRetryBackoff.ts so that all 5xx responses except 503 Service Unavailable are considered retryable.
- Improved readability by replacing hardcoded status values with constants from http-status-codes.
- Added test coverage to verify that 503 responses are not retried and updated existing retry expectations accordingly.

### HTTP client and runner test changes

- Updated HttpClient tests to confirm that 503 responses do not trigger retries.
- Replaced previous 503-based retry scenarios with 500-level errors that remain retryable.
- Updated IncrementalRetryBackoffRunner tests with additional parameterized cases to validate the new non-retry behavior for 503 responses.

### Cleanup and maintainability improvements

- Replaced the hardcoded 429 status code with StatusCodes.TOO_MANY_REQUESTS to improve clarity and consistency across the codebase.

### DIFF
--- a/libraries/api-client/src/http/incrementalRetryBackoff.ts
+++ b/libraries/api-client/src/http/incrementalRetryBackoff.ts
@@ -34,9 +34,10 @@ export type IncrementalRetryBackoffPolicy = {
 
 const initialRetryDelayInMilliseconds = 100;
 const maximumRetryDelayInMilliseconds = 10 * 60 * 1000;
-const nonStandardRetryableStatusCode = StatusCodes.METHOD_FAILURE;
-const serverErrorStatusCodeRangeStart = StatusCodes.INTERNAL_SERVER_ERROR;
-const serverErrorStatusCodeRangeEnd = StatusCodes.NETWORK_AUTHENTICATION_REQUIRED;
+const nonStandardRetryableStatusCode = 420;
+const serverErrorStatusCodeRangeStart = 500;
+const serverErrorStatusCodeRangeEnd = 599;
+const serviceUnavailableStatusCode = 503;
 
 export function createIncrementalRetryBackoffPolicy(): IncrementalRetryBackoffPolicy {
   return {
@@ -72,7 +73,7 @@ export function createIncrementalRetryBackoffPolicy(): IncrementalRetryBackoffPo
           return (
             statusCodeValue >= serverErrorStatusCodeRangeStart &&
             statusCodeValue <= serverErrorStatusCodeRangeEnd &&
-            statusCodeValue !== StatusCodes.SERVICE_UNAVAILABLE
+            statusCodeValue !== serviceUnavailableStatusCode
           );
         })
         .unwrapOr(false);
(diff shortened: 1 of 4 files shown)


zskhan (dismissed):
LGTM 🚀

libraries/api-client/src/http/incrementalRetryBackoff.ts
```diff
@@ -34,7 +35,6 @@ export type IncrementalRetryBackoffPolicy = {
 const initialRetryDelayInMilliseconds = 100;
 const maximumRetryDelayInMilliseconds = 10 * 60 * 1000;
 const nonStandardRetryableStatusCode = 420;
-const tooManyRequestsStatusCode = 429;
```
screendriver, line 37: When we remove this line we should be consistent for all other status codes as well and not just this single one.

screendriver (changes requested):

libraries/api-client/src/http/incrementalRetryBackoff.ts
e-maad: Basically I intentionally kept others. None of the others specifying a status code.
I also thought about:
```
const serverErrorStatusCodeRangeStart = StatusCode.<something>;
const serverErrorStatusCodeRangeEnd = StatusCode.<something>;
```
But it is more readable when we have numbers for the range.

For `nonStandardRetryableStatusCode`, `StatusCode.METHOD_FAILURE` is deprecated.

e-maad: If you still suggest the change, I will do it.

arjita-mitra (dismissed):

libraries/api-client/src/http/incrementalRetryBackoff.ts
screendriver: > None of the others specifying a status code.

Line 37, 38 and 39 does. So yes I would change it everywhere herein this module to be consistent. **Or** don't use `http-status-codes` at all to be consistent as it was before.

e-maad: Resolved: 878c1e8b92940946179f08bfdb60b5bd6b8dcb98

screendriver (dismissed):

screendriver (approved):

PR #21334, opened 2026-05-18
Type-check root bin/ scripts and align ts-node release commands [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

This change adds real TypeScript coverage for the repo-root `bin/` scripts and makes the runtime `ts-node` commands use the same config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,9 @@ jobs:
       - name: Build libraries
         run: yarn nx run-many -t build --projects=tag:type:lib
 
+      - name: Type check workspace tools
+        run: yarn nx run workspace-tools:type-check
+
       - name: Lint file names
         run: yarn lint:fileNames
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,8 +59,8 @@ jobs:
       - name: Build libraries
         run: yarn nx run-many -t build --projects=tag:type:lib
 
-      - name: Type check workspace tools
-        run: yarn nx run workspace-tools:type-check
+      - name: Type check
+        run: yarn type-check
 
       - name: Lint file names
         run: yarn lint:fileNames
--- a/bin/release_tag.ts
+++ b/bin/release_tag.ts
@@ -18,7 +18,7 @@
  */
 
 import {execSync} from 'child_process';
-import * as logdown from 'logdown';
+import logdown = require('logdown');
 import * as path from 'path';
 import * as readline from 'readline';
 
--- a/bin/release_tag.ts
+++ b/bin/release_tag.ts
@@ -18,7 +18,7 @@
  */
 
 import {execSync} from 'child_process';
-import logdown = require('logdown');
+import logdown from 'logdown';
 import * as path from 'path';
 import * as readline from 'readline';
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -59,6 +59,9 @@ jobs:
       - name: Build libraries
         run: yarn nx run-many -t build --projects=tag:type:lib
 
+      - name: Type check workspace tools
```
markbrockhoff: Wait a minute, why only run type-check for the workspace tools? Shouldn't there already be a script which runs type-check in general for the repo? Shouldn't that already include the workspace tools if configured correctly in nx?

bin/release_tag.ts
```diff
@@ -18,7 +18,7 @@
  */
 
 import {execSync} from 'child_process';
-import * as logdown from 'logdown';
+import logdown = require('logdown');
```
markbrockhoff: Why is it necessary to switch to require here? logdown's export is a function so I think `import logdown from 'logdown'` should work right?

.github/workflows/ci.yml
screendriver: There is already a repo-wide type-check script: `yarn type-check` runs `nx run-many -t type-check --all`. In principle, yes, `workspace-tools:type-check` should be included there once it exists.

I used the targeted Nx command in our pipeline because the repo-wide aggregate currently fails on unrelated existing `core-lib:type-check` issues, so switching the pipeline to `yarn type-check` in this pull request would broaden the scope beyond bin/ tooling.

If we want full repo-wide type-check in our pipeline I think that should be a follow-up once the existing library type-check failures are resolved.

But: I can also switch to it right now? I just wanted to have the pull request focused with the minimal amount of changes. What do you think?

bin/release_tag.ts
screendriver: You're right that `import logdown from 'logdown'` should also work here.

I switched to `import logdown = require('logdown')` because logdown [publishes](https://github.com/caiogondim/logdown.js/blob/aa08713ec09b0e486e1836db79a4b583e098204b/index.d.ts#L38) `export = logdown` and that form matches the package's declared module shape exactly.

That said I'm fine switching it to `import logdown from 'logdown'` if we prefer the ES import style consistently **or** replace `logdown` (which has the last update 6 years ago) with something more modern. What do you think?

.github/workflows/ci.yml
markbrockhoff: How big of a change are we talking about? If it's just adding the build of core-lib as a dependency of type-check I'd be fine with it, but if it becomes bigger I'm all on your side to do it as individual PR :)

bin/release_tag.ts
markbrockhoff: Yeah replacing logdown would be welcome in general, but out of scope for this PR. I personally would prefer the ES import here simply for consistency. :)

.github/workflows/ci.yml
screendriver: Changed in c6305c5a698f24f645d285f49f594665c315eefc

bin/release_tag.ts
screendriver: Changed in 25c0bf5dcef3902b6d15f4c96692a2b5b16178d0

zskhan (approved):

PR #21351, opened 2026-05-19
Add accessibility in video background settings component [WPB-24448]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24448" title="WPB-24448" target="_blank">StoryWPB-24448</a>  Accessibility feature for the Blur Background
  </td></table>
  <br />
 

Add accessibility in video background settings component also improve focus management and few styling glitches

### DIFF
--- a/apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
@@ -59,7 +59,7 @@ import {useToggleState} from 'src/script/hooks/useToggleState';
 import {CallViewTab} from 'src/script/view_model/CallingViewModel';
 import {useKoSubscribableChildren} from 'Util/componentUtil';
 import {isDetachedCallingFeatureEnabled} from 'Util/isDetachedCallingFeatureEnabled';
-import {handleKeyDown, KEY} from 'Util/keyboardUtil';
+import {handleKeyDown, isTabKey, KEY} from 'Util/keyboardUtil';
 import {t} from 'Util/localizerUtil';
 import {preventFocusOutside} from 'Util/util';
 
@@ -254,10 +254,14 @@ const FullscreenVideoCall = ({
         return;
       }
 
+      if (!isTabKey(event)) {
+        return;
+      }
+
       event.preventDefault();
       event.stopPropagation();
 
-      preventFocusOutside(event, 'video-calling', targetDocument);
+      preventFocusOutside(event, 'video-calling-wrapper', targetDocument);
     };
 
     targetDocument.addEventListener('keydown', onKeyDown);
@@ -302,6 +306,7 @@ const FullscreenVideoCall = ({
 
   return (
     <div
+      id="video-calling-wrapper"
       data-uie-name="fullscreen-video-call"
       className={cx('video-calling-wrapper', {
         'app--small-offset': hasOffset && isMiniMode,
--- a/apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
+++ b/apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {ChangeEvent, CSSProperties, ReactNode} from 'react';
+import {ChangeEvent, CSSProperties, ReactNode, useEffect, useId, useRef} from 'react';
 
 import {BlurHighIcon, BlurLowIcon, Checkbox, CheckboxLabel, CircleIcon} from '@wireapp/react-ui-kit';
 
@@ -31,6 +31,7 @@ import {
   backgroundSettingsScrollableContentStyles,
   backgroundSettingsTitleStyles,
   backgroundSettingsWrapperStyles,
+  closeButtonStyles,
   sectionLabelStyles,
   tileButtonStyles,
   tileGridStyles,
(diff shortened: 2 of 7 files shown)


e-maad (dismissed):

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -60,10 +61,26 @@ const isEffectSelected = (selected: BackgroundEffectSelection, candidate: Backgr
   return true;
 };
 
+const getBackgroundEffectLabel = (effect: BackgroundEffectSelection, backgrounds: BuiltinBackground[]): string => {
```
screendriver: `BackgroundEffectSelection` looks like a discriminated union. Could we make this mapping exhaustive? (pattern matching for example)

That would fail at compile time when a new effect type is added, rather than silently falling back.

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -60,10 +61,26 @@ const isEffectSelected = (selected: BackgroundEffectSelection, candidate: Backgr
...
+  switch (effect.type) {
+    case 'none':
+      return t('videoCallBackgroundNoEffect');
+    case 'blur':
+      return effect.level === 'low' ? t('videoCallBackgroundBlurLow') : t('videoCallBackgroundBlurHigh');
+    case 'virtual': {
+      const background = backgrounds.find(({id}) => id === effect.backgroundId);
+      return background ? t(background.labelKey) : t('videoCallBackgroundVirtual');
+    }
+    default:
```
screendriver: Can we avoid this `default` branch and use an exhaustive check instead? The fallback may hide missing cases if the union grows.

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -81,7 +99,9 @@ const BackgroundTile = ({
       type="button"
       css={tileButtonStyles}
       data-selected={selected}
-      aria-pressed={selected}
+      role="radio"
```
screendriver, line 106: Since this element is `role="radio"` can we align the parent container to `role="radiogroup"` semantics?

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -128,12 +177,15 @@ export const VideoBackgroundSettings = ({
 
         {/* Blur section */}
         <div>
-          <div css={sectionLabelStyles}>{t('videoCallBackgroundBlurSectionLabel')}</div>
-          <div css={tileGridStyles}>
+          <h3 id={blurSectionId} css={sectionLabelStyles}>
+            {t('videoCallBackgroundBlurSectionLabel')}
+          </h3>
+          <div css={tileGridStyles} role="group" aria-labelledby={blurSectionId}>
```
screendriver: This group contains radio options. Should this be `role="radiogroup"`?

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -171,19 +224,26 @@ export const VideoBackgroundSettings = ({
...
-                onSelectEffect={onSelectEffect}
-                previewStyle={{
-                  backgroundImage: `url(${background.imageUrl}), ${background.previewGradient}`,
-                }}
-              />
-            ))}
+          <h3 id={virtualSectionId} css={sectionLabelStyles}>
+            {t('videoCallBackgroundVirtualSectionLabel')}
+          </h3>
+          <div css={tileGridStyles} role="group" aria-labelledby={virtualSectionId}>
```
screendriver: Same question as above: because children are `role="radio"` I think `role="radiogroup"` would be the semantically correct container role here.

apps/webapp/src/script/components/calling/VideoControls/VideoBackgroundSettings/VideoBackgroundSettings.tsx
```diff
@@ -99,25 +119,54 @@ export const VideoBackgroundSettings = ({
   onEnableHighQualityBlur,
   onClose,
 }: VideoBackgroundSettingsProps) => {
+  const titleId = useId();
+  const blurSectionId = useId();
+  const virtualSectionId = useId();
+  const closeButtonRef = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
```
screendriver, line 131: Can we add a test for this autofocus behavior?

apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -254,10 +254,14 @@ const FullscreenVideoCall = ({
         return;
       }
 
+      if (!isTabKey(event)) {
```
screendriver, line 257: We now intercept only `Tab` but there is a test missing for this behavior change.

apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -254,10 +254,14 @@ const FullscreenVideoCall = ({
...
+        return;
+      }
+
       event.preventDefault();
       event.stopPropagation();
 
-      preventFocusOutside(event, 'video-calling', targetDocument);
+      preventFocusOutside(event, 'video-calling-wrapper', targetDocument);
```
screendriver, line 264: Changed from `video-calling` to `video-calling-wrapper`. Was this intentional?

screendriver (changes requested):

apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
zskhan: Yes this is intentional as background settings panel is outside of `video-calling`. So assigned the id `video-calling-wrapper` to the parent which contains both.

screendriver (dismissed):

screendriver (approved):

PR #21353, opened 2026-05-19
fix(core): skip conference MLS decrypt while leaving a call [WPB-24984]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

After leaving a call, queued conference subconversation message-add events can fail with core-crypto "Couldn't find conversation" once the local MLS group has been wiped. 
That is expected and not actionable.

Add isMlsConversationNotFoundError() alongside the other MLS error guards and use it when logging handleMLSMessageAdd failures for subconversation events.
Those cases now emit a dedicated message so Datadog triage can separate them from genuine MLS message-add failures.

### DIFF
--- a/libraries/core/src/conversation/conversationService/conversationService.ts
+++ b/libraries/core/src/conversation/conversationService/conversationService.ts
@@ -58,6 +58,7 @@ import {
 
 import {MessageTimer, MessageSendingState, RemoveUsersParams} from '../../conversation/';
 import {MLSService, MLSServiceEvents} from '../../messagingProtocols/mls';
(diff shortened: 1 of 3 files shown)


arjita-mitra (dismissed):

typfel (commented):
I'm concerned we are adding extra complexity & state for essentially avoiding a warning log. Wouldn't it be enough to not log a warning if conversation is missing when it's targeting a subconversation.

thisisamir98:
@typfel thanks for your comment, I pushed a new approach to only customize the log message. https://github.com/wireapp/wire-webapp/pull/21353/changes/d1ac80848e5fd19f9b27cf0383463b25b2964b08

typfel (approved):

zskhan (approved):

PR #21358, opened 2026-05-20
feat: make conversation filter single select(WPB-25541)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25541" title="WPB-25541" target="_blank">TaskWPB-25541</a>  Conversation filter → single-select (All Files)
  </td></table>
  <br />
 

- make conversation filter single select
- checkbox styling updates

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
@@ -30,7 +30,7 @@ export type PopoverFilterConfig = {
   selectedIds: string[];
   onSelectionChange: (ids: string[]) => void;
   disabled?: boolean;
-  singleSelect?: boolean;
+  singleSelect: boolean;
 };
 
 export type ToggleFilterConfig = {
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
@@ -30,6 +30,7 @@ export type PopoverFilterConfig = {
   selectedIds: string[];
   onSelectionChange: (ids: string[]) => void;
   disabled?: boolean;
+  singleSelect?: boolean;
 };
 
 export type ToggleFilterConfig = {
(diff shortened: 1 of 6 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
```diff
@@ -30,6 +30,7 @@ export type PopoverFilterConfig = {
   selectedIds: string[];
   onSelectionChange: (ids: string[]) => void;
   disabled?: boolean;
+  singleSelect?: boolean;
```
screendriver: Can we make this property required? Required property reduces the "attack" surface of having three states: `undefined`, `true` and `false` and makes everything more explicit.

So the consumer just have to decide between two instead of three states.

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/common/FilterPopover/FilterPopover.tsx
```diff
@@ -166,17 +168,19 @@ export const FilterPopover = ({
...
             ) : (
               filteredItems.map(item => {
                 const isSelected = selectedIds.includes(item.id);
+                const isItemDisabled = singleSelect && selectedIds.length > 0 && !isSelected;
                 return (
                   <li key={item.id} css={itemRowHoverStyles} role="option" aria-selected={isSelected}>
                     <Checkbox
                       wrapperCSS={checkboxWrapperStyles}
                       checked={isSelected}
+                      disabled={isItemDisabled}
```
zskhan, line 183: Just asking from the UX perspective, it will disable other checkboxes as soon as 1 item is selected. Is this intentional?
I mean we can make it in other way that on selection of other checkbox it will uncheck the previous one.

arjita-mitra: no thats the expected design.  it will disable other checkboxes as soon as 1 item is selected in conversation filter.

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsFiltersBar/filterConfig.ts
arjita-mitra: fixed a7ef27982036fc49fa1d0bd6f91d3234d52d4704

screendriver (approved):

PR #21366, opened 2026-05-21
test: write the Proteus Verification regression tests [WPB-19949]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19949" title="WPB-19949" target="_blank">TaskWPB-19949</a>  [Web/QA] Write the Proteus Verification regression tests in Playwright
  </td></table>
  <br />
 

## Summary

This PR introduces automated test coverage for Proteus Device Verification. The tests validate the behavior of device trust states, cross-device verification workflows, and conversation security degradation triggers.


JacquelineLehner (dismissed):

apps/webapp/test/e2e_tests/specs/ProteusVerification/proteusVerification.spec.ts
```diff
@@ -0,0 +1,407 @@
...
+    await createPage(withLogin(userA));
+
+    const {pages, modals, components} = PageManager.from(
+      await createPage(withLogin(userA, {confirmNewHistory: true})),
+    ).webapp;
+
+    await components.conversationSidebar().clickPreferencesButton();
+    await pages.settings().devicesButton.click();
+
+    const firstDevice = pages.devices().activeDevices.first();
```
markbrockhoff, line 73: If there is only one device it shouldn't be necessary to call `.first()` on the locator. If there was a second device shown we would want the test to fail since there was an unexpected initial state. So please just leave it out unless you really expect the locator to match multiple devices.

```suggestion
    const firstDevice = pages.devices().activeDevices;
```

markbrockhoff: In fact I think you can remove all `activeDevices.first()` since you never have more than 2 devices per user

apps/webapp/test/e2e_tests/specs/ProteusVerification/proteusVerification.spec.ts
```diff
@@ -0,0 +1,407 @@
...
+      await test.step('Action: User A sends a message into the conversation with User B', async () => {
+        await userAPages.conversation().sendMessage('Message from User A');
+
+        await expect(userAModals.confirm().modalTitle).toContainText(`${userB.fullName} started using a new device`);
+        await userAModals.confirm().clickAction();
+      });
+
+      await test.step('Action: User B sends a message to User A from the second device', async () => {
+        const userB2Pages = PageManager.from(userB2Device).webapp.pages;
+        await userB2Pages.conversation().sendMessage('Message from User B');
```
markbrockhoff, line 297: 🔧  Just to be save we should maybe not rely on wire opening the conversation automatically :)

```suggestion
        const userB2Pages = PageManager.from(userB2Device).webapp.pages;
        await userB2Pages.conversationList().getConversation(userA.fullName).open();
        await userB2Pages.conversation().sendMessage('Message from User B');
```

markbrockhoff (dismissed):

apps/webapp/test/e2e_tests/specs/ProteusVerification/proteusVerification.spec.ts
zhannach: True, it works without it! I only added .first() for readability since activeDevices is plural, and clicking a plural locator feels a bit counterintuitive. If you feel strongly about removing it though, I'm fine to take it out

markbrockhoff: I fully understand that, my only concern is that this test would still pass or maybe remove the wrong device if a second showed up as. (E.g. at some point maybe the current device was shown here as well)
I'd also be fine with using `.first()` if there was an assertion for activeDevices to have count 1 in front of it. This way the naming stays consistent and easy to follow but we get the safety that the test would fail in case of a second device. What do you think? :)

markbrockhoff (dismissed):

markbrockhoff (approved):

zskhan (approved):

PR #21385, opened 2026-05-26
feat: Integrate created by filter(WPB-25534)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25534" title="WPB-25534" target="_blank">TaskWPB-25534</a>  Client side integration of filter created by
  </td></table>
  <br />
 

Wires the Created by filter through to Pydio Cells so selected participants actually constrain search results. Selecting one user sends Operation: Must; selecting multiple sends Operation: Should (OR),

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/useDriveEnabledParticipantFilterItems/useDriveEnabledParticipantFilterItems.tsx
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/useDriveEnabledParticipantFilterItems/useDriveEnabledParticipantFilterItems.tsx
@@ -1,0 +1,68 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {stringifyQualifiedId} from '@wireapp/core/lib/util/qualifiedIdUtil';
+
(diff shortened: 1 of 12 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/common/useDriveEnabledParticipantFilterItems/useDriveEnabledParticipantFilterItems.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {matchQualifiedIds} from 'Util/qualifiedId';
+
+export const getCreatedByFilterId = (user: User): string => stringifyQualifiedId(user.qualifiedId);
+
+const sortConversationsByRecentFirst = (conversations: Conversation[]): Conversation[] =>
+  conversations.toSorted(
+    (conversationA, conversationB) => conversationB.last_event_timestamp() - conversationA.last_event_timestamp(),
+  );
+
+export const getDriveEnabledParticipants = (conversationRepository: ConversationRepository): User[] => {
```
thisisamir98, line 36: `getDriveEnabledParticipants` this function/helper might be a good candidate to live on ConversationState.ts

arjita-mitra: Thanks for your review. Here's why I would like to defer making this change,

### Cost

1. Touches 6 files (state class, state test, hook, hook test, 2 wrapper hooks, 2 components) to move a 10-line helper that has one caller.
2. The hook param changes from conversationRepository to conversationState. Both wrapper hooks and both components need to be updated to pass it through. No behavior change.
3. The cells-enabled check (cellsState === READY || PENDING) ends up in two places: still on ConversationRepository.getAllCellEnabledGroupConversations, and again in the new state method.
4. The current test is a plain function test. Moving it means setting up knockout observables to test the same thing.

### Benefit

1. Nothing else uses this helper today. It's a view for the Created-by filter, not shared state.
2. The current location is fine. The name says what it does.
3. Yes, it would sit nicely next to groupConversations and connectedUsers on the state class, but those are used by lots of things. This isn't.

thisisamir98: @arjita-mitra something to consider for future in case it's needed somewhere else then!

thisisamir98 (approved):

PR #21390, opened 2026-05-27
fix(core): defer MLS epoch recovery until notification replay finishes [WPB-21916]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Problem
After login, the client replays a large backlog of legacy notifications while local MLS state may already be ahead, behind, or incomplete. When decrypting an MLS message-add failed with WrongEpoch, the recovery orchestrator ran an external-commit rejoin immediately. That could happen in the middle of replay, before later notifications had a chance to apply missing commits.

This led to unnecessary MLSConversationRecovered system messages and made replay noisier. Pausing the rejoin queue did not help, because orchestrator recovery bypassed that queue entirely.

Approach
- Defer epoch-mismatch recovery for inbound handleMessageAdd while the connection is not LIVE. The failed message returns null and replay continues.
- After catch-up finishes, runDeferredEpochRecovery() reconciles any conversations that still need external commit (once per conversation).


libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -753,6 +769,37 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+    this.logger.info('Running deferred MLS epoch mismatch recoveries after notification replay', {
+      count: this.deferredEpochRecoveries.size,
+    });
+
+    const entries = [...this.deferredEpochRecoveries.values()];
+    this.deferredEpochRecoveries.clear();
+
+    for (const {conversationId, subconvId, trigger} of entries) {
+      try {
+        await this.recoverMLSGroupFromEpochMismatch(conversationId, subconvId, trigger, {force: true});
```
zskhan: If `entries` are already cleared and after that if for some reason `recoverMLSGroupFromEpochMismatch` fails then there will be no way to retry the entry.

What do you think?

zskhan (dismissed):

libraries/core/src/account.ts
```diff
@@ -1017,7 +1011,7 @@ export class Account extends TypedEventEmitter<Events> {
 
       const firstEventPayload = notification.data.event.payload[0];
       const notificationTime = firstEventPayload ? this.getNotificationEventTime(firstEventPayload) : null;
-      if (this.connectionState !== ConnectionState.LIVE && notificationTime !== null && notificationTime.length > 0) {
+      if (!this.connectionStateTracker.isLive() && notificationTime !== null && notificationTime.length > 0) {
```
e-maad:
```suggestion
      if (!this.connectionStateTracker.isLive() && is.nonEmptyString(notificationTime)) {
```

libraries/core/src/account.ts
```diff
@@ -1109,6 +1103,7 @@ export class Account extends TypedEventEmitter<Events> {
         .push(async () => {
           this.logger.info(`Resuming message sending. ${getQueueLength()} messages to be sent`);
           await this.rehydrateMlsPendingProposalsTasksOnLiveTransition();
+          await this.service!.conversation.runDeferredEpochRecovery();
```
e-maad: Do we need non-null assertion here?

libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -1084,11 +1140,44 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+    throw new MlsEpochRecoveryDeferredError(conversationId, subconvId);
+  }
+
   private async recoverMLSGroupFromEpochMismatch(
     conversationId: QualifiedId,
     subconversationId?: SUBCONVERSATION_ID,
     trigger?: MlsEpochRecoveryTrigger,
+    options: {force: boolean} = {force: false},
   ) {
+    if (options.force === false && this.shouldDeferEpochRecovery(trigger)) {
```
e-maad: If `option.force` is boolean (and controlled by us) the please rename to `isForced`
```suggestion
    if (!options.force && this.shouldDeferEpochRecovery(trigger)) {
```

libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -1084,11 +1140,44 @@ export class ConversationService extends TypedEventEmitter<Events> {
     }
   }
 
+  private getDeferredEpochRecoveryKey(conversationId: QualifiedId, subconvId?: SUBCONVERSATION_ID): string {
+    return `${conversationId.id}@${conversationId.domain}:${subconvId ?? 'none'}`;
```
e-maad: Is this `none` can be provided by any enum?

libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -753,6 +769,37 @@ export class ConversationService extends TypedEventEmitter<Events> {
     return this.mlsService.wipeConversation(groupId);
   };
 
+  /**
+   * Runs epoch-mismatch recovery for conversations that deferred external commit during
+   * notification replay. Called when the connection transitions to LIVE.
+   */
+  public async runDeferredEpochRecovery(): Promise<void> {
+    if (this.deferredEpochRecoveries.size === 0) {
```
e-maad: I would suggest to use `is.emptySet(deferredEpochRecoveries)`

libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -1084,11 +1140,44 @@ export class ConversationService extends TypedEventEmitter<Events> {
...
+  }
+
   private async recoverMLSGroupFromEpochMismatch(
     conversationId: QualifiedId,
     subconversationId?: SUBCONVERSATION_ID,
     trigger?: MlsEpochRecoveryTrigger,
+    options: {force: boolean} = {force: false},
   ) {
+    if (options.force === false && this.shouldDeferEpochRecovery(trigger)) {
+      this.deferEpochRecovery(conversationId, subconversationId, trigger);
```
e-maad, line 1172: You are throwing an error from `deferEpochRecovery`, are we handling it somewhere?

libraries/core/src/conversation/conversationService/conversationService.ts
```diff
@@ -102,6 +115,9 @@ export class ConversationService extends TypedEventEmitter<Events> {
     private readonly subconversationService: SubconversationService,
     private readonly isMLSConversationRecoveryEnabled: () => Promise<boolean>,
     private readonly _mlsService?: MLSService,
+    private readonly connectionStateTracker: ConnectionStateTracker = createConnectionStateTracker(
```
e-maad: Why are we maintaining two state-managers for connection?
I can see another one in`libraries/core/src/account.ts`. These states will not be synced, is this intentional?

libraries/core/src/conversation/conversationService/conversationService.ts
thisisamir98: - We're not left with zero recovery paths if deferred recovery fails.
- later MLS operation while LIVE — send, proposal, another message-add — hits WrongEpoch again and goes through MlsRecoveryOrchestrator (one attempt per operation, not infinite).
handleConversationsEpochMismatch() — full scan on missed notifications / legacy path.
New inbound traffic after LIVE — no longer deferred; recovery runs inline.
So it’s not “retry forever on the same deferred entry,” but “eventual recovery via the next real trigger.” Plus this is how it used to be, this PR only defers the operation.

libraries/core/src/conversation/conversationService/conversationService.ts
thisisamir98: it's just here and it's a in memory key, using hyphen instead.

libraries/core/src/conversation/conversationService/conversationService.ts
thisisamir98: recovery orchestrator is going to catch it

zskhan (approved):

PR #21397, opened 2026-05-28
feat: add refresh CRL list debug option [WPB-24944]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24944" title="WPB-24944" target="_blank">TaskWPB-24944</a>  [Web] Add a debug option to receive information about revoked certificates
  </td></table>
  <br />
 

### Description 
Added a debug option for refreshing the CRL list

Screenshot 2026-05-28 at 16 24 01
Screenshot 2026-05-28 at 16 22 33

### DIFF
--- a/libraries/core/src/messagingProtocols/mls/e2eIdentityService/e2eiServiceInternal.ts
+++ b/libraries/core/src/messagingProtocols/mls/e2eIdentityService/e2eiServiceInternal.ts
@@ -17,6 +17,7 @@
  *
  */
 
+import is from '@sindresorhus/is';
 import {Decoder} from 'bazinga64';
 
 import {APIClient} from '@wireapp/api-client';
@@ -224,7 +225,7 @@ export class E2EIServiceInternal {
     });
     this.logger.debug('oidc data', oidcData);
 
-    if (oidcData.data.validated !== 'valid') {
+    if (is.nullOrUndefined(oidcData.data.validated)) {
       throw new Error('Error while trying to continue OAuth flow. OIDC challenge not validated');
     }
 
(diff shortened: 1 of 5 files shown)


libraries/core/src/messagingProtocols/mls/e2eIdentityService/e2eiServiceInternal.ts
```diff
@@ -224,7 +225,7 @@ export class E2EIServiceInternal {
     });
     this.logger.debug('oidc data', oidcData);
 
-    if (oidcData.data.validated !== 'valid') {
+    if (is.nullOrUndefined(oidcData.data.validated)) {
```
zskhan, line 228: Previously it was 
`idcData.data.validated !== 'valid'`
Which means `idcData.data.validated` can have other string values.
But with your new check it will not throw error on other values. Only throw if it is null/undefined.

zskhan (approved):

libraries/core/src/messagingProtocols/mls/e2eIdentityService/e2eiServiceInternal.ts
e-maad: Basically, `oidcData.data.validated` returns date string which should never be null or undefined.
After [this change](https://github.com/wireapp/wire-webapp/commit/e4e932bc8cefe17eced7c49d4ef9690a11e6dc4b#diff-2d0ce2ce7ab7bdae95842951c6835a5c107e5478d8a90aac34b987ca0226ed49R227) we introduced a bug and now I am reverting it.

PR #21403, opened 2026-05-28
fix: narrow outdated notification check to duplicate-risk events [WPB-21916]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

The global lastEventDate outdated filter was applied to every notification stream event. When two unrelated conversations share the same millisecond timestamp, a later OTR message can be silently dropped after an earlier one advances lastEventDate.

Scope the check to event types that can be handled locally and replayed on the stream (member-join, member-leave, create, rename, protocol-update, message-timer-update, receipt-mode-update, add-permission-update). OTR and MLS content events are no longer filtered this way.

Introduce isOutdatedNotificationStreamEvent in core and reuse it from NotificationService and EventValidator.


libraries/core/src/notification/outdatedNotificationStreamEventTypes.ts
```diff
@@ -0,0 +1,60 @@
...
+  CONVERSATION_EVENT.RENAME,
+  CONVERSATION_EVENT.PROTOCOL_UPDATE,
+  CONVERSATION_EVENT.MESSAGE_TIMER_UPDATE,
+  CONVERSATION_EVENT.RECEIPT_MODE_UPDATE,
+  CONVERSATION_EVENT.ADD_PERMISSION_UPDATE,
+]);
+
+export function isOutdatedNotificationStreamEvent(
+  event: {time?: string; type: string},
+  source: string,
```
zskhan: Shouldn't it be `NotificationSource`?

zskhan (dismissed):

libraries/core/src/notification/outdatedNotificationStreamEventTypes.ts
```diff
@@ -0,0 +1,60 @@
...
+  lastEventDate?: Date,
+): boolean {
+  if (source !== NotificationSource.NOTIFICATION_STREAM) {
+    return false;
+  }
+
+  if (event.time === undefined || event.time.length === 0 || lastEventDate === undefined) {
```
e-maad: Could be more readable with `is.nullOrUndefined` and `is.nonEmptyString`.

e-maad (dismissed):
Add one suggestion

apps/webapp/src/script/repositories/event/EventValidator.ts
```diff
@@ -18,25 +18,20 @@
...
 
 import {CONVERSATION_EVENT, USER_EVENT} from '@wireapp/api-client/lib/event/';
+import {isOutdatedNotificationStreamEvent} from '@wireapp/core/lib/notification';
 
 import {EventSource} from './EventSource';
 import {EventValidation} from './EventValidation';
 
 export function validateEvent(
-  event: {time: string; type: CONVERSATION_EVENT | USER_EVENT},
+  event: {time?: string; type: CONVERSATION_EVENT | USER_EVENT},
```
arjita-mitra: The type signature changed from time: string (required) to time?: string (optional), BUT the code is called with potentially missing time fields while the check logic depends on it. is not this making it more unsafe?

apps/webapp/src/script/repositories/event/EventValidator.ts
```diff
@@ -18,25 +18,20 @@
...
-  const shouldCheckEventDate = !!eventTime && isFromNotificationStream && lastEventDate;
-
-  if (shouldCheckEventDate) {
-    /** This check prevents duplicated "You joined" system messages. */
-    const isOutdated = new Date(lastEventDate).getTime() >= new Date(eventTime).getTime();
-    if (isOutdated) {
-      return EventValidation.OUTDATED_TIMESTAMP;
-    }
+  if (
+    isOutdatedNotificationStreamEvent(event, source, lastEventDate !== undefined ? new Date(lastEventDate) : undefined)
```
arjita-mitra: Problems:

- Ternary operator for simple null coercion is hard to parse
- The lastEventDate parameter is a string, but being converted to Date inside the ternary
- Multi-line if with long function call reduces redability

Better approach

const lastEventDateObj = lastEventDate ? new Date(lastEventDate) : undefined;
if (isOutdatedNotificationStreamEvent(event, source, lastEventDateObj))

libraries/core/src/notification/outdatedNotificationStreamEventTypes.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21403/commits/f8fd896d2b33f16ba47eb258e07d7e32ed6da289

libraries/core/src/notification/outdatedNotificationStreamEventTypes.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21403/commits/ed3d61bd1c7b481df406dd70f374afcbb223b192

apps/webapp/src/script/repositories/event/EventValidator.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21403/commits/f6e2a7e30d0080a284a7b59b29f72dd947f599c2

apps/webapp/src/script/repositories/event/EventValidator.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21403/commits/24c3189dcdd8b5773d96f9b1b74e0280f44d544d

zskhan (approved):

arjita-mitra (approved):

PR #21407, opened 2026-05-29
test: add critical flow test for on-prem login redirect [WPB-25319]

JacquelineLehner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25319" title="WPB-25319" target="_blank">TaskWPB-25319</a>  [Web][QA] Write Critical Flow -> On Prem Login Redirect
  </td></table>
  <br />
 

This PR introduces a critical flow E2E test (TC-8757) to verify the on-premise login redirect behavior. It ensures that users entering an email associated with a claimed domain on the SSO page are correctly prompted and redirected to the correct backend.

Refs: [WPB-25319]

[WPB-25319]: https://wearezeta.atlassian.net/browse/WPB-25319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/webapp/test/e2e_tests/backend/brigRepository.e2e.ts
```diff
@@ -251,4 +251,31 @@ export class BrigRepositoryE2E {
...
+      },
+      {
+        headers: {
+          Authorization: `Basic ${BASIC_AUTH}`,
+        },
+      },
+    );
+  }
+
+  public async deleteDomain(domain: string) {
```
markbrockhoff: 🎨  Maybe call it `deleteDomainClaim()` or something in the sense of "unclaim" domain just so it's clear that it's not a domain which is being deleted here 🙈

apps/webapp/test/e2e_tests/backend/brigRepository.e2e.ts
```diff
@@ -251,4 +251,31 @@ export class BrigRepositoryE2E {
...
     );
   }
+
+  public async claimDomain(domain: string) {
+    await this.axiosInstance.put(
+      `i/domain-registration/${domain}`,
+      {
+        backend: {
+          config_url: 'https://nginz-https.anta.wire.link/deeplink.json',
+          webapp_url: 'https://webapp.anta.wire.link',
```
markbrockhoff: Could we maybe pass these config options as parameters? It feels a bit wrong to have them hardcoded and hidden in here. Maybe we could even re-use the env vars which will be added for the federation tests here, since they also point to Anta?

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+    const page = await createPage(context);
+    const pageManager = PageManager.from(page);
+    const {pages} = pageManager.webapp;
+
+    await test.step('Open login page and enter email with claimed domain', async () => {
+      await pageManager.openSSOPage();
+      await pages.singleSignOn().enterEmailOnSSOPage(email);
+      await pages.singleSignOn().ssoSignInButton.click();
```
markbrockhoff: 🔧  I don't think you need to click the sign in button yourself. When testing it it automatically shows the dialog after entering it. It might still work if the button is clicked before the dialog opens automatically, but let's try to avoid undefined timing specific behavior :)

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const ON_PREM_WEBAPP_URL = 'https://webapp.anta.wire.link';
+
+test('On Prem Login Redirect', {tag: ['@TC-8757', '@regression']}, async ({context, createPage, api}) => {
+  const domain = faker.internet.domainName();
+  const email = `redirect-user@${domain}`;
+
+  try {
+    await test.step('Claim domain and configure on-prem redirect', async () => {
+      await api.brig.claimDomain(domain);
```
markbrockhoff, line 35: The claiming of the domain is specified as precondition for the test, so no need to do it inside it. I'd suggest to put it in a beforeEach hook outside the test, this way you also don't need the manual try-finally block as you can just do the deletion inside a afterEach hook then :)

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      const urlPattern = new RegExp(ON_PREM_WEBAPP_URL.replace(/\./g, '\\.'));
+
+      await Promise.all([
+        page.waitForURL(urlPattern, {
+          timeout: 20_000,
+        }),
+        pages.customBackend().connectButton.click(),
+      ]);
+
+      await expect(page).toHaveURL(urlPattern);
```
markbrockhoff: Why so complicated? In the end all you're doing is clicking a button and waiting for the URL to change, there's no need to start listening for the URL change before the button is clicked. Also the Pattern for matching the URL can be simplified by using `String.raw` to escape it automatically (no need to maintain the escape regex ourself 😉)


```suggestion
      await page.getByRole('button', {name: 'Connect'}).click();
      await expect(page).toHaveURL(new RegExp(String.raw`${ON_PREM_WEBAPP_URL}`), {timeout: 20_000});
```

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    await test.step('Open login page and enter email with claimed domain', async () => {
+      await pageManager.openSSOPage();
+      await pages.singleSignOn().enterEmailOnSSOPage(email);
+      await pages.singleSignOn().ssoSignInButton.click();
+    });
+
+    await test.step('Verify connect-to-organization backend dialog is shown', async () => {
+      await expect(pages.customBackend().title).toBeVisible();
+      await expect(pages.customBackend().redirectWarningText).toBeVisible();
+      await expect(pages.customBackend().adminInfoText).toBeVisible();
```
markbrockhoff: 🔧  Let's try not to be overly specific here. Testing the exact texts shown on the page isn't a good practice as they easily change which would break the test for no reason.
Looking into the test description I don't think it's necessary to assert the redirectWarningText and adminInfoText. I'd say it should be sufficient to just check the page contains the headline and optionally the domain of the backend the user will be redirected to:


```suggestion
      await expect(page.getByText("Connect to your organization's backend?")).toBeVisible();
      await expect(page.getByText(ON_PREM_WEBAPP_URL)).toBeVisible();
```

This way the test is less likely to break for unrelated reasons which means less maintenance effort ;)

Also notice ho I inlined the locator for the title. This locator will most likely be only used once in here as it is very specific. I think we don't need the dedicated POM for now. Also having the title in the test itself makes it easier to read :)

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/backend/brigRepository.e2e.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
JacquelineLehner: agreed-  I removed the button click ✅

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
JacquelineLehner: I changed the locators and removed the now redundant dedicated POM ✅

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
JacquelineLehner: I didn't know that, thank you! Changed it ✅

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/backend/brigRepository.e2e.ts
JacquelineLehner: ✅

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
```diff
@@ -23,46 +23,43 @@ import {faker} from '@faker-js/faker';
 
 const ON_PREM_WEBAPP_URL = 'https://webapp.anta.wire.link';
 
-test('On Prem Login Redirect', {tag: ['@TC-8757', '@regression']}, async ({context, createPage, api}) => {
-  const domain = faker.internet.domainName();
-  const email = `redirect-user@${domain}`;
+test.describe('On Prem Login Redirect Flow', () => {
```
markbrockhoff, line 26: :art: Nothing wrong with this, just for you to know in case you don't: You can also use beforeEach etc outside of describe. It will then simply effect all tests in the file :)

markbrockhoff (approved):

apps/webapp/test/e2e_tests/specs/CriticalFlow/onPremLoginRedirect-TC-8757.spec.ts
JacquelineLehner: Good to know, thank you :)

zskhan (approved):

PR #21416, opened 2026-06-02
fix(api-client): force WebSocket reconnect after sleep when socket stays OPEN [WPB-21916]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

After laptop sleep, the TCP connection can die while the browser still reports readyState as OPEN. The sleep handler detected wake correctly, but reconnect was gated on isDisconnected (socket === CLOSED), so the common zombie-OPEN case was skipped and the client could stay silent with no messages and no offline banner.

Always call reconnect on wake, regardless of socket state. Add Datadog-friendly logs for wake detection, reconnect initiation, and successful reopen (including idle time and unansweredPing context). Add regression tests to prevent reintroducing the isDisconnected gate.

### DIFF
--- a/libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
+++ b/libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
@@ -50,6 +50,7 @@ describe('ReconnectingWebsocket back from sleep', () => {
   it('forces reconnect when waking with an OPEN zombie socket', () => {
     const RWS = new ReconnectingWebsocket(async () => 'ws://localhost');
     const reconnect = jest.fn();
+    const rescheduleSpy = jest.spyOn(RWS as never as {reschedulePinging: () => void}, 'reschedulePinging');
 
     RWS['socket'] = {
       readyState: WEBSOCKET_STATE.OPEN,
@@ -60,6 +61,7 @@ describe('ReconnectingWebsocket back from sleep', () => {
     sleepWakeCallback!();
 
     expect(reconnect).toHaveBeenCalledTimes(1);
+    expect(rescheduleSpy).toHaveBeenCalledTimes(1);
     expect(RWS['sleepReconnectPending']).toBe(true);
   });
 
--- a/libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
+++ b/libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
@@ -1,0 +1,96 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+jest.mock('../utils/backFromSleepHandler/backFromSleepHandler');
+
+import {onBackFromSleep} from '../utils/backFromSleepHandler/backFromSleepHandler';
+
+import {ReconnectingWebsocket, WEBSOCKET_STATE} from './reconnectingWebsocket';
+
+const mockedOnBackFromSleep = jest.mocked(onBackFromSleep);
+
+describe('ReconnectingWebsocket back from sleep', () => {
+  let sleepWakeCallback: (() => void) | undefined;
+
+  beforeEach(() => {
+    sleepWakeCallback = undefined;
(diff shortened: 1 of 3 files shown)


libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  it('forces reconnect when waking with an OPEN zombie socket', () => {
+    const RWS = new ReconnectingWebsocket(async () => 'ws://localhost');
+    const reconnect = jest.fn();
+
+    RWS['socket'] = {
+      readyState: WEBSOCKET_STATE.OPEN,
+      reconnect,
+    } as never;
+
+    expect(sleepWakeCallback).toBeDefined();
```
e-maad, line 60: I don't understand why we have to assert if the `sleepWakeCallback` is defined.

libraries/api-client/src/tcp/reconnectingWebsocket.backFromSleep.test.ts
```diff
@@ -0,0 +1,96 @@
...
+jest.mock('../utils/backFromSleepHandler/backFromSleepHandler');
+
+import {onBackFromSleep} from '../utils/backFromSleepHandler/backFromSleepHandler';
+
+import {ReconnectingWebsocket, WEBSOCKET_STATE} from './reconnectingWebsocket';
+
+const mockedOnBackFromSleep = jest.mocked(onBackFromSleep);
+
+describe('ReconnectingWebsocket back from sleep', () => {
+  let sleepWakeCallback: (() => void) | undefined;
```
e-maad, line 29: Instead of `undefined` if we pass an empty function we don't have to use `Non-Null Assertion !`

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -162,6 +155,28 @@ export class ReconnectingWebsocket {
...
   };
 
+  private handleBackFromSleep(): void {
+    if (!this.socket) {
+      this.logger.info('Back from sleep detected, skipping reconnect — WebSocket instance does not exist');
+      return;
+    }
+
+    const state = this.getState();
+    const timeSinceLastMessage = this.lastMessageTimestamp > 0 ? Date.now() - this.lastMessageTimestamp : undefined;
```
e-maad, line 166: Why `undefined` instead of `0`?

zskhan (approved):

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -162,56 +156,117 @@ export class ReconnectingWebsocket {
...
+        ? `last message ${timeSinceLastMessage}ms ago`
+        : 'no messages received on this connection yet';
+
+    this.logger.info(
+      `Back from sleep detected, WebSocket state: ${WEBSOCKET_STATE[state]} (${state}), ${idleDescription}, unansweredPing: ${this.hasUnansweredPing}, forcing reconnect`,
+    );
+
+    // Force reconnect even if the browser keeps the socket in OPEN state after sleep.
+    this.sleepReconnectPending = true;
+    this.reschedulePinging();
```
arjita-mitra, line 178: System sleeps → Browser keeps socket "OPEN" → System wakes → 
handleBackFromSleep() calls reschedulePinging() but isPingingEnabled is false →
Socket stays "zombie OPEN" → Connection appears healthy but is dead

Fix: Add explicit logging and ensure pinging is restored:

```
// Ensure pinging is active before rescheduling
  if (!this.isPingingEnabled) {
    this.logger.warn('Back from sleep detected but pinging is disabled; connection may become unresponsive');
  }
  
  const state = this.getState();
  // ... rest of logic
  
  this.sleepReconnectPending = true;
  this.reschedulePinging();
  this.socket.reconnect();
```

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -162,56 +156,117 @@ export class ReconnectingWebsocket {
...
+
+    if (state === WEBSOCKET_STATE.CONNECTING || state === WEBSOCKET_STATE.CLOSING) {
+      this.logger.debug(`Ping tick — socket is transitioning (state: ${WEBSOCKET_STATE[state]}), skipping ping`);
+      return;
+    }
+
+    if (state === WEBSOCKET_STATE.CLOSED) {
+      this.logger.info('Ping tick — socket CLOSED, forcing reconnect');
+      this.hasUnansweredPing = false;
+      this.socket.reconnect();
```
arjita-mitra, line 257: Better to let the reconnecting-websocket library handle reconnection timing? No guarantee that reconnect() actually establishes a connection before the next ping tick (20ms later).Could trigger rapid re-reconnect loops.

libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -162,56 +156,117 @@ export class ReconnectingWebsocket {
...
       this.onClose(event);
     }
   };
 
-  private startPinging(): void {
+  /**
+   * Starts the ping interval if pinging is enabled and not already running.
+   * The interval is kept alive across reconnects and is only cleared on disconnect/cleanup.
+   */
+  private ensurePinging(): void {
```
arjita-mitra, line 207: ensurePinging and reschedulePinging can create a possible race condition. No locking mechanism prevents both from creating intervals simultaneously

arjita-mitra (changes requested):

PR #21417, opened 2026-06-02
feat: keep the focus trapped inside the wrapper when background sidebar is closed [WPB-24448]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24448" title="WPB-24448" target="_blank">StoryWPB-24448</a>  Accessibility feature for the Blur Background
  </td></table>
  <br />
 

On closing the background settings sidebar via keyboard, focus should be trapped inside the on going call wrapper instead of going to the previously opened conversation

### DIFF
--- a/apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
+++ b/apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {ChangeEvent, useEffect, useState} from 'react';
+import {ChangeEvent, useEffect, useRef, useState} from 'react';
 
 import {DefaultConversationRoleName} from '@wireapp/api-client/lib/conversation/';
 import cx from 'classnames';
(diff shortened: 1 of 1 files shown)


apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -198,6 +198,15 @@ const FullscreenVideoCall = ({
 
   const [isParticipantsListOpen, toggleParticipantsList] = useToggleState(false);
   const [isBackgroundSidebarOpen, setIsBackgroundSidebarOpen] = useState(false);
+  const wrapperRef = useRef<HTMLDivElement>(null);
```
e-maad, line 201: Instead of using `useRef` to persist the old state, can't we just implement a `backgroudSidebarHandler`, like this:
```

const backgroudSidebarToggleHandler = (newValue: boolean) {
 const wasBackgroundSidebarOpen = isBackgroundSidebarOpen;
// implementation
}
```
WDYT?

zskhan: Thanks for the idea.

e-maad (approved):

PR #21440, opened 2026-06-03
test: write the federation regression tests [WPB-19988]

zhannach:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-19988" title="WPB-19988" target="_blank">TaskWPB-19988</a>  [Web/QA] Write the Federation regression tests in Playwright
  </td></table>
  <br />
 

This PR introduces new Federation regression test suite.

These tests validate key federation capabilities, including asset sharing, cross-backend communication, backup imports, and real-time audio/video calls between different backend environments.

Changes:
 - New Regression Coverage: Added 6 new automated test cases.
 - Refactored API-related logic to dynamically support requests across different backend environments.
 - Extended the CreateTeam fixture and PageManager to accept and initialize distinct backend configurations.
 - Refactored conversation participant selectors to eliminate duplication and improve maintainability.

### DIFF
--- a/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
+++ b/apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
@@ -123,16 +123,17 @@ export class ConversationDetailsPage {
     return userLocator.isVisible();
   }
 
-  async openParticipantDetails(fullName: string) {
-    await this.getLocatorByUser(fullName).click();
-  }
+  getParticipant(fullName: string) {
+    const baseLocator = this.getLocatorByUser(fullName);
 
-  getUserRoleIcon(fullName: string) {
-    return this.getLocatorByUser(fullName).getByTestId(/^status-(external|guest|admin)$/);
-  }
-
-  getUserAvailabilityIcon(fullName: string) {
-    return this.getLocatorByUser(fullName).getByTestId('status-availability-icon');
+    return Object.assign(baseLocator, {
+      roleIcon: baseLocator.getByTestId(/^status-(external|guest|admin)$/),
+      availabilityIcon: baseLocator.getByTestId('status-availability-icon'),
+      federatedIcon: baseLocator.getByTestId('status-federated-user'),
+      openDetails: async () => {
+        await baseLocator.click();
+      },
+    });
   }
 
   getLocatorByUser(fullName: string) {
--- a/apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
@@ -5,7 +5,7 @@ import {PageManager} from 'test/e2e_tests/pageManager';
 import {test, withLogin, expect, createTeam} from 'test/e2e_tests/test.fixtures';
 import {getAudioFilePath, getTextFilePath, getVideoFilePath, shareAssetHelper} from 'test/e2e_tests/utils/asset.util';
 import {getImageFilePath} from 'test/e2e_tests/utils/sendImage.util';
-import {createGroup, sendConnectionRequest} from 'test/e2e_tests/utils/userActions';
+import {createAndSaveBackup, createGroup, sendConnectionRequest} from 'test/e2e_tests/utils/userActions';
 
 test.describe('Federation', () => {
   const federationBaseUrl = process.env.FEDERATION_WEBAPP_URL!;
@@ -16,6 +16,7 @@ test.describe('Federation', () => {
 
   let normalUser: User;
   let federatedUser: User;
+  const groupName = 'Federated group';
 
   test.beforeEach(async ({api}) => {
     normalUser = (await createTeam(api, 'Normal Team', {features: {conferenceCalling: true}})).owner;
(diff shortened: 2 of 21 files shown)


apps/webapp/test/e2e_tests/pageManager/webapp/pages/conversationDetails.page.ts
```diff
@@ -123,16 +123,17 @@ export class ConversationDetailsPage {
...
-    await this.getLocatorByUser(fullName).click();
-  }
-
-  getUserRoleIcon(fullName: string) {
-    return this.getLocatorByUser(fullName).getByTestId(/^status-(external|guest|admin)$/);
-  }
-
-  getUserAvailabilityIcon(fullName: string) {
-    return this.getLocatorByUser(fullName).getByTestId('status-availability-icon');
+  getParticipant(fullName: string) {
```
markbrockhoff, line 126: Lovely ❤️

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
```diff
@@ -0,0 +1,564 @@
...
+      await test.step('Normal user can send and delete Text message in the federated group', async () => {
+        await normalUserPages.conversation().sendMessage('Hello from staging user');
+
+        const messageNormalUser = normalUserPages.conversation().getMessage({sender: normalUser});
+        await expect(messageNormalUser).toBeVisible();
+        await expect(federatedUserPages.conversation().getMessage({sender: normalUser})).toBeVisible();
+
+        await normalUserPages.conversation().deleteMessage(messageNormalUser, 'Everyone');
+        await expect(messageNormalUser).not.toBeVisible();
+        await expect(federatedUserPage.getByTestId('element-message-delete')).toBeVisible();
```
markbrockhoff, line 255: Why use the "element-message-deleted" test id? How about checking the message is there and then expecting it to be gone after deletion? :)


```suggestion
        const messageNormalUser = normalUserPages.conversation().getMessage({sender: normalUser});
        const messageFederatedUser = federatedUserPages.conversation().getMessage({sender: normalUser});

        await expect(messageNormalUser).toBeVisible();
        await expect(messageFederatedUser).toBeVisible();

        await normalUserPages.conversation().deleteMessage(messageNormalUser, 'Everyone');

        await expect(messageNormalUser).not.toBeVisible();
        await expect(messageFederatedUser).not.toBeVisible();
```

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
```diff
@@ -0,0 +1,564 @@
...
+          .getMessage({sender: federatedUser})
+          .filter({has: normalUserPage.getByRole('img')});
+        await expect(messageWithImage).toBeVisible();
+      });
+
+      let backupName: string;
+      await test.step('Create and save backup for the normal user', async () => {
+        await normalUserComponents.conversationSidebar().clickPreferencesButton();
+        backupName = await createAndSaveBackup(testInfo, normalUserPageManager);
+      });
```
markbrockhoff: 🎨 I learned that this was possible from you in one of your previous PRs ;)


```suggestion
      const backupName = await test.step('Create and save backup for the normal user', async () => {
        await normalUserComponents.conversationSidebar().clickPreferencesButton();
        return await createAndSaveBackup(testInfo, normalUserPageManager);
      });
```

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
```diff
@@ -0,0 +1,564 @@
...
+      await expect(
+        federatedUserPages.conversationList().getConversation(normalUser.fullName, {protocol: 'mls'}),
+      ).toBeVisible();
+
+      await test.step('Normal user creates a group and adds federated user to it', async () => {
+        await createGroup(normalUserPages, groupName, []);
+        await normalUserPages.conversationList().getConversation(groupName).open();
+        await normalUserPages.conversation().toggleGroupInformation();
+        await normalUserPages.conversationDetails().clickAddPeopleButton();
+        await normalUserPages.conversationDetails().addUsersToConversation([federatedUser.fullName]);
```
markbrockhoff: Is there a specific reason why you're adding the federatedUser like this instead of just passing him to the users array?

```suggestion
        await createGroup(normalUserPages, groupName, [federatedUser]);
```

markbrockhoff (changes requested):

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
zhannach: Yes, the requirements were to create an empty group and then invite the federated user to it. However, since there are no specific checks during this step, I'd like to see if we can align on simplifying it.

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
zhannach: It was implemented according to the requirements:
```
User A deletes previous text message for all users 
User B sees deleted message 
```

apps/webapp/test/e2e_tests/specs/Federation/federation.spec.ts
markbrockhoff: Yeah because if there are no intermediate assertions passing the user to the util directly would do the same thing from a test perspective, just with less code.

markbrockhoff (approved):

zskhan (approved):

PR #21446, opened 2026-06-05
feat: load-more pagination for conversation drive search view(WPB-25540)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25540" title="WPB-25540" target="_blank">TaskWPB-25540</a>  Pagination for search results
  </td></table>
  <br />
 

Conversation Drive now switches pagination model based on the search view state

- Browse mode (search view closed): Page number and rows per page based pagination unchanged

- Search mode (search view open, regardless of typed query or filters):

  Load More button with 30 initial rows + 20 per click, matching the global All Files view's progression.

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/loadMorePagination/loadMorePagination.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/loadMorePagination/loadMorePagination.ts
@@ -17,24 +17,26 @@
  *
  */
 
+import {Maybe} from 'true-myth';
+
 import {CellPagination} from '../cellPagination/cellPagination';
 
 export const LOAD_MORE_INITIAL_SIZE = 30;
 export const LOAD_MORE_INCREMENT = 20;
 
-export const getLoadMoreOffset = (pagination: CellPagination | null): number | null => {
+export const getLoadMoreOffset = (pagination: CellPagination | null): Maybe<number> => {
   const nextOffset = pagination?.nextOffset;
 
   if (nextOffset === undefined) {
-    return null;
+    return Maybe.nothing();
   }
 
   const currentPage = pagination?.currentPage;
   const totalPages = pagination?.totalPages;
 
   if (currentPage === undefined || totalPages === undefined) {
-    return nextOffset;
+    return Maybe.just(nextOffset);
   }
 
-  return currentPage < totalPages ? nextOffset : null;
+  return currentPage < totalPages ? Maybe.just(nextOffset) : Maybe.nothing();
 };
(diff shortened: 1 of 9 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
```diff
@@ -200,6 +233,26 @@ export const useConversationSearchFiles = ({
...
+    }
+    hasFiredInitialFetchRef.current = true;
+    fireAndForgetInvoker.fireAndForget(async (): Promise<void> => {
+      await searchNodes({
+        query: searchQuery.trim().length > 0 ? searchQuery : FETCH_ALL_QUERY,
+        filters,
+      });
+    });
+  }, [enabled, searchNodes, searchQuery, filters, fireAndForgetInvoker]);
+
```
thisisamir98, line 255: it might be possible to fire this this on the onClick event that triggers the search view open, this way we might be able to reduce the complexity by avoiding useEffect

apps/webapp/src/script/components/Conversation/ConversationCells/common/loadMorePagination/loadMorePagination.ts
```diff
@@ -0,0 +1,40 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CellPagination} from '../cellPagination/cellPagination';
+
+export const LOAD_MORE_INITIAL_SIZE = 30;
+export const LOAD_MORE_INCREMENT = 20;
+
+export const getLoadMoreOffset = (pagination: CellPagination | null): number | null => {
```
thisisamir98: a true-myth "Maybe" could be used for the return type of this function so the consumer is always forced to handle the emptiness.

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: enabled here is a derived value (isCellsStateReady && isSearchViewOpen), so it can flip from sources other than the search input click (cells state becoming ready after open, keyboard shortcut). Moving the fetch to onClick would miss those.

apps/webapp/src/script/components/Conversation/ConversationCells/common/loadMorePagination/loadMorePagination.ts
arjita-mitra: Addressed f15bcbaa7a8d5b0bceca28c0580babc522ed513f

thisisamir98 (approved):

PR #21452, opened 2026-06-08
Handle missing conversation domains in qualified IDs [WPB-26233]

### LINKED ISSUE #21409: Regression in 2026.05.26.08.40.42
# App stuck on "Fetching your connections and conversations" — `matchQualifiedIds` throws on legacy conversations with an undefined `domain`

## Environment

- webapp version: `2026.05.26.08.40.42`
- Observed in wire-desktop (macOS, Mac App Store build `com.wearezeta.zclient.mac` v3.40.5442) loading `https://app.wire.com`
- Also reproducible in a browser against `app.wire.com` for any account whose local IndexedDB contains pre-federation conversation records
- Backend: production (`wire.com`)

## Summary

On startup, the app hangs forever on the "Fetching your connections and conversations" spinner. The cause is an uncaught `TypeError` during initial conversation load:

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')
    at matchQualifiedIds
    at <anonymous>
    at Array.find (<anonymous>)
    at ConversationRepository.filterDeletedConnectionRequests
    at ConversationRepository.filterLoadedConversations
    at async ConversationRepository.loadRemoteConversations
    at async <init>.initApp
```

(Function names from the minified bundle: `matchQualifiedIds` = `(0,F.a)`, `ConversationRepository` = `we`.)

The rejection is never caught, so `initApp` aborts and the conversation list never renders.

## Root cause

`matchQualifiedIds` guards a `null` domain but not an `undefined` domain, then reads `domain.length`:

```js
// util/QualifiedId.ts  (minified `i`, exported as F.a)
function matchQualifiedIds(e, t) {
  if (e === undefined || t === undefined) return false;
  const sameId = e.id === t.id;
  const sameDomain =
    e.domain === null || t.domain === null ||
    e.domain.length === 0 || t.domain.length === 0 ||   // <-- throws if domain is undefined
    e.domain === t.domain;
  return sameId && sameDomain;
}
```

When either argument has `domain === undefined` (not `null`), the `=== null` checks are false and execution falls through to `e.domain.length` / `t.domain.length`, which throws.

The undefined domain originates in the caller:

```js
// conversation/ConversationRepository.ts (minified)
filterDeletedConnectionRequests(conversations, /* found */ t, connections, /* deleted */ i = []) {
  for (const conversation of conversations) {
    const { type, qualified_id, id, domain } = conversation;
    type !== CONVERSATION_TYPE.CONNECT ||
      connections.find(c => matchQualifiedIds(c.conversationId, qualified_id || { id, domain }))
      ? /* keep */ : /* drop */;
  }
  ...
}
```

For a CONNECT-type conversation that has no `qualified_id`, it constructs `{ id, domain }` from the conversation. Legacy records persisted by pre-federation clients have **no `domain` field**, so the constructed id is `{ id, domain: undefined }`, which `matchQualifiedIds` then throws on.

## Steps to reproduce

1. Use an account whose local IndexedDB `conversations` store contains a **CONNECT-type (type 3)** conversation record with **no `domain` property** and no `qualified_id` (i.e. data written by a pre-federation webapp version). In our case the offending record was a pending connection request created in 2015.
2. Open webapp `2026.05.26` and log in.
3. During init, `loadRemoteConversations` → `filterLoadedConversations` → `filterDeletedConnectionRequests` builds `{ id, domain: undefined }` for that conversation and passes it to `matchQualifiedIds`.
4. `matchQualifiedIds` reads `undefined.length` and throws. The unhandled rejection aborts `initApp`.

**Actual:** app stays on the "Fetching your connections and conversations" spinner indefinitely.
**Expected:** legacy conversations without a `domain` are handled gracefully and the conversation list loads.

## Notes

- A fresh login in a clean browser does not reproduce, because the backend now returns conversations with a `domain`. The bug only surfaces against existing local data that predates the `domain` field.
- In one affected account, 143 of 857 stored conversations had no `domain`; in another, 827 of 827. Only CONNECT-type records reach the throwing path, but the missing-`domain` data is broader.

## Suggested fix

Treat `undefined` domain the same as `null` in `matchQualifiedIds`, for example:

```js
const sameDomain =
  e.domain == null || t.domain == null ||   // == null covers null AND undefined
  e.domain.length === 0 || t.domain.length === 0 ||
  e.domain === t.domain;
```

Optionally, also normalize the constructed id in `filterDeletedConnectionRequests` so a missing `domain` defaults to the self/local backend domain rather than `undefined`.

## Workaround (data side)

Setting `domain` to the user's backend domain on the affected local conversation records clears the crash and preserves history. This is only a per-client mitigation. The matcher guard above fixes it for everyone.

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26233" title="WPB-26233" target="_blank">BugWPB-26233</a>  [Web] App stuck on "Fetching your connections and conversations" — matchQualifiedIds throws on legacy conversations with an undefined domain
  </td></table>
  <br />
 

Accept legacy qualified IDs with omitted or undefined domains in matchQualifiedIds to prevent startup failures when older IndexedDB conversation records are loaded.

Add regression tests for omitted and undefined domains and preserve existing matching behavior for null, empty, and fully qualified domains.

Fixes #21409

### DIFF
--- a/apps/webapp/src/script/util/qualifiedId.ts
+++ b/apps/webapp/src/script/util/qualifiedId.ts
@@ -17,10 +17,16 @@
  *
  */
 
+import is from '@sindresorhus/is';
+
 interface QualifiedEntity {
   [index: string]: any;
-  domain: string | null;
+  domain?: string | null;
   id: string;
+}
+
+function isUnqualifiedDomain(domain: string | null | undefined): boolean {
+  return is.nullOrUndefined(domain) || is.emptyString(domain);
 }
 
 /**
@@ -31,17 +37,14 @@ interface QualifiedEntity {
  * @param entity2 - The second entity to compare
  * @return boolean - do the entities match
  */
-export function matchQualifiedIds(entity1?: QualifiedEntity, entity2?: QualifiedEntity) {
-  if (entity1 === undefined || entity2 === undefined) {
+export function matchQualifiedIds(entity1?: QualifiedEntity, entity2?: QualifiedEntity): boolean {
+  if (is.undefined(entity1) || is.undefined(entity2)) {
     return false;
   }
 
   const idsMatch = entity1.id === entity2.id;
   const domainsMatch =
-    entity1.domain === null ||
-    entity2.domain === null ||
-    entity1.domain.length === 0 ||
-    entity2.domain.length === 0 ||
-    entity1.domain === entity2.domain;
+    isUnqualifiedDomain(entity1.domain) || isUnqualifiedDomain(entity2.domain) || entity1.domain === entity2.domain;
+
   return idsMatch && domainsMatch;
 }
--- a/apps/webapp/src/script/util/qualifiedId.test.ts
+++ b/apps/webapp/src/script/util/qualifiedId.test.ts
@@ -27,6 +27,14 @@ describe('QualifiedId util', () => {
         {domain: '', id: '1', property: 1},
       ],
       [
+        {id: '1', stuff: 'extra'},
+        {domain: 'wire.com', id: '1', property: 1},
+      ],
+      [
+        {domain: undefined, id: '1', stuff: 'extra'},
+        {domain: 'wire.com', id: '1', property: 1},
+      ],
+      [
         {domain: 'wire.com', id: '1', other: 12},
         {domain: 'wire.com', id: '1'},
       ],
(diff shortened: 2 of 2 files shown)


apps/webapp/src/script/util/qualifiedId.ts
```diff
@@ -17,12 +17,18 @@
...
+
 interface QualifiedEntity {
   [index: string]: any;
-  domain: string | null;
+  domain?: string | null;
   id: string;
 }
 
+function isUnqualifiedDomain(domain: string | null | undefined): boolean {
+  return domain === undefined || domain === null || is.emptyString(domain);
```
zskhan: return is.nullOrUndefined(domain) || is.emptyString(domain);

zskhan (dismissed):

apps/webapp/src/script/util/qualifiedId.ts
screendriver: Changed

zskhan (approved):

PR #21472, opened 2026-06-09
feat: integrate team feature flag for background effects [WPB-26162]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26162" title="WPB-26162" target="_blank">TaskWPB-26162</a>  [Web] Add Team-Based Feature Flag for Background Effect Rendering
  </td></table>
  <br />
 

Moved the whole Background Effects feature behind Team based feature flag.
If it is enabled for the Team the user will see it even if the Debug Menu feature flag is disabled.


apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
```diff
@@ -209,6 +219,12 @@ export class BackgroundEffectsHandler {
   }
 
   public readFeatureEnabledStateFromStore(): boolean {
+    const isEnabledByTeam = this.teamState.isBackgroundEffectsEnabled();
+
+    if (isEnabledByTeam) {
```
screendriver, line 226: Can we make the precedence between the team feature flag and the local debug/storage flag explicit here?

With the current logic, `backgroundEffects: DISABLED` from the backend does not actually disable the feature if the user still has `video-background-effects-feature-enabled=true` in local storage.

If this pull request is meant to put background effects behind the team feature flag, I would expect an explicit backend `DISABLED` state to win over the local debug flag once `teamFeatures` has been loaded.

If the local debug flag is intentionally allowed to override a disabled team flag, then I think we should make that explicit in naming/comment and add a test for `backgroundEffects: DISABLED` + local storage `true`.

screendriver (changes requested):

apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
zskhan: This is intentional that the local Debug Menu flag is allowed to override the team feature flag, including when the backend returns `backgroundEffects: DISABLED`.

The reasoning is that the Debug Menu is currently used for development/testing purposes, and we wanted to preserve the ability to force-enable the feature locally regardless of the team configuration.

The Debug Menu flag is planned to be removed completely in the future, at which point the backend/team feature flag will become the single source of truth.

screendriver (dismissed):

e-maad (dismissed):
LGTM 🚀

screendriver (approved):

PR #21477, opened 2026-06-10
Force WebSocket reconnect after wake from sleep [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Problem

After a laptop goes to sleep and wakes up again, the app can stop receiving new messages.

The likely failure mode is a zombie WebSocket connection:

- the browser still reports the WebSocket as `OPEN`
- the physical TCP/backend connection is stale or gone
- the app still believes it is connected
- no reconnect is forced
- new backend events may no longer reach the active app session

This is especially problematic because the app can look healthy while silently missing new messages.

## Root cause

The existing wake-from-sleep detection already intended to reconnect the WebSocket after sleep.

However, the reconnect callback was gated behind an `isDisconnected` check that only returned `true` when the socket was already `CLOSED`.

That skipped the more dangerous case:

`wake from sleep -> browser still reports OPEN -> reconnect callback is not called -> app keeps using a stale connection`

After wake-up, we should not trust `OPEN` as proof that the backend connection is still alive.

## Change

This changes the WebSocket wake-from-sleep recovery behavior behind the startup feature toggle:

`reliable-websocket-connection`

When the toggle is enabled and the app detects that the machine came back from sleep, the existing WebSocket wrapper forces an in-place reconnect whenever a WebSocket instance exists, regardless of the current `readyState`.

The important rule is:

`wake from sleep -> do not trust OPEN -> reconnect`

Without the startup toggle, the existing gated behavior is preserved.

## Additional hardening

The sleep detector lifecycle is now explicit.

Previously, cleanup could stop the sleep detector while the same `ReconnectingWebsocket` instance could later be reused. That could leave wake-from-sleep detection disabled after a disconnect/reconnect cycle.

The handler is now restarted when needed and protected against duplicate registration.

## Out of scope

This pull request intentionally does not:

- replace `reconnecting-websocket`
- migrate to `partysocket`
- introduce an XState-based connection state machine
- rewrite `EventRepository.connectWebSocket()`


libraries/api-client/src/tcp/reconnectingWebsocket.ts
```diff
@@ -66,8 +92,8 @@ export class ReconnectingWebsocket {
   };
 
   private readonly logger: logdown.Logger;
-  private socket?: RWS;
-  private pingerId?: NodeJS.Timeout;
+  private socket?: ReconnectingWebsocketWrapper;
+  private pingerId?: ReturnType<typeof globalThis.setInterval>;
```
thisisamir98: maybe we could use IntervalIdentifier type we have above?

arjita-mitra (dismissed):

e-maad (dismissed):
LGTM 👍

libraries/api-client/src/tcp/reconnectingWebsocket.ts
screendriver: Good catch. Yes, this should use the existing `IntervalIdentifier` type for consistency. I'll update it in 92cd5acf417c258c9a9a53f5fda6827d52c2c243

zskhan (approved):

thisisamir98 (approved):

PR #21478, opened 2026-06-10
Move Playwright critical flows out of pull requests [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

Run the Playwright critical flow as a scheduled and release-gated confidence check instead of a required check on every pull request.

The previous setup deployed pull request builds to shared precommit environments and ran critical-flow E2E tests before merge. That made normal pull request feedback slower and added contention around a limited number of precommit slots.

The workflow now runs twice per day and can still be triggered manually when needed. Staging and production releases call the same precommit slot workflow as a release gate, so critical flows still block releases without blocking ordinary pull requests.

Failed critical-flow test runs now notify Deployoholics using the existing deployment failure notification channel. This makes failed scheduled or release-gated runs visible without relying on pull request comments.

### DIFF
--- a/.github/workflows/precommit-crit-flows.yml
+++ b/.github/workflows/precommit-crit-flows.yml
@@ -13,7 +13,7 @@ on:
       timezone: Europe/Berlin
 
 concurrency:
-  group: critical-flow-${{ github.event_name }}-${{ github.ref }}
+  group: precommit-crit-flows-${{ github.event_name }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/precommit-crit-flows.yml
+++ b/.github/workflows/precommit-crit-flows.yml
@@ -1,48 +1,37 @@
+---
 name: precommit-crit-flows
 
 on:
-  pull_request:
-    branches: [master, dev, release/*]
-    types: [opened, synchronize, reopened, ready_for_review]
-  merge_group:
-    branches: [master, dev, release/*]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: Why this critical flow run was started
+        required: false
+        type: string
+  schedule:
+    - cron: '17 8,20 * * *'
+      timezone: Europe/Berlin
 
-# Cancel superseded runs on the same PR; slot-level concurrency guards the shared EB environments
 concurrency:
-  group: precommit-pr-${{ github.event.pull_request.number || github.run_id }}
+  group: critical-flow-${{ github.event_name }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
   build:
-    if: ${{ github.event_name == 'merge_group' || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository) }}
+    if: ${{ github.repository == 'wireapp/wire-webapp' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/precommit-crit-flows.yml
```diff
@@ -1,48 +1,37 @@
...
+        required: false
+        type: string
+  schedule:
+    - cron: '17 8,20 * * *'
+      timezone: Europe/Berlin
 
-# Cancel superseded runs on the same PR; slot-level concurrency guards the shared EB environments
 concurrency:
-  group: precommit-pr-${{ github.event.pull_request.number || github.run_id }}
+  group: critical-flow-${{ github.event_name }}-${{ github.ref }}
```
markbrockhoff: Do we really want to base concurrency off of the github ref? We're using 3 pre defined envs to deploy to, so this could allow for more concurrent runs on different commits. Maybe only take the event name so max two runs can happen at the same time?

.github/workflows/precommit-crit-flows.yml
```diff
@@ -1,48 +1,37 @@
...
-  merge_group:
-    branches: [master, dev, release/*]
+  workflow_dispatch:
+    inputs:
+      reason:
+        description: Why this critical flow run was started
+        required: false
+        type: string
+  schedule:
+    - cron: '17 8,20 * * *'
```
markbrockhoff, line 12: Does it even make sense to run this as part of its own workflow? We already have the e2e-test-nightly.yml workflow which does pretty much the same but runs against dev instead of deploying the state first. Maybe we could get rid of some duplication by just changing the times it runs throughout a day? What do you think?

.github/workflows/precommit-crit-flows.yml
screendriver: Good point. The `ref` part does not match the new slot model anymore.

The workflow now maps scheduled runs to `wire-webapp-precommit-1` and manual runs to `wire-webapp-precommit-2`, independent of the ref. Keeping `github.ref` in the caller concurrency group can allow multiple runs for different refs even though they target the same fixed slot.

I’ll change this to `precommit-crit-flows-${{ github.event_name }}`. That keeps scheduled and manual runs independent, but prevents multiple runs of the same event type from competing for the same precommit environment. The reusable `precommit-slot-run.yml` still keeps the slot-level `precommit-slot-${{ inputs.env_name }}` guard.

Changed in 225d325ecde6822ddffc4d9aff2ed496f0df6d02

.github/workflows/precommit-crit-flows.yml
screendriver: Good point. I would keep them separate for now (!).

`e2e-tests-nightly.yml` checks the already deployed dev environment. This workflow builds the current state, deploys that exact artifact to a precommit environment and then runs the critical flow.

That distinction matters for the release gate, because we need to test the release artifact before staging/production deployment.

There is some overlap for the scheduled health check, but keeping it aligned with the release-gate path is intentional. We can reduce duplication later by extracting shared reporting/notification logic but I would avoid mixing that into this pull request.

arjita-mitra (approved):

PR #21483, opened 2026-06-10
Remove Corepack from Yarn bootstrap [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Remove [Corepack](https://github.com/nodejs/corepack) from the Yarn bootstrap path.

The repository already commits the intended Yarn release under `.yarn/releases` and points to it through `yarnPath` in `.yarnrc.yml`. This pull request adds a repository-local `bin/yarn` wrapper and uses that instead of `corepack install` / `corepack enable`.

## Why

Corepack is only needed to bootstrap a package-manager binary. In this repository, the Yarn binary is already part of the repository through `.yarn/releases` and is referenced from `yarnPath` in `.yarnrc.yml`.

Corepack is also no longer a safe long-term assumption for Node.js itself. [The official Corepack documentation](https://github.com/nodejs/corepack#default-installs) says Corepack is distributed with Node.js from `14.19.0` up to, but not including, `25.0.0`. That means a future Node.js 25 upgrade would break any workflow or Docker image that still assumes `corepack` is available by default.

Using the committed Yarn release directly makes CI, Docker, and local setup more explicit. It keeps the existing Yarn setup unchanged while removing an unnecessary bootstrap dependency.


zskhan:
There is one Sonar comment rest it looks good to me 🚀

screendriver:
> There is one Sonar comment rest it looks good to me 🚀

Changed in fb0d2de613be8a712d18560d3d2609c9975c9d3b

zskhan (approved):

PR #21510, opened 2026-06-11
feat: fix grace period for CTR generation [WPB-24526]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24526" title="WPB-24526" target="_blank">BugWPB-24526</a>  [Web] Wrong time shown on Web when grace period is active
  </td></table>
  <br />
 

This pull request improves E2E identity enrollment by making grace period handling more accurate.

- Added getRemainingGracePeriodDelay to calculate the actual remaining enrollment grace period using the user's identity details, activation time, and team settings.
- Updated the enrollment flow to use this remaining time for timers and snooze confirmation modals, ensuring users see the correct grace period instead of the original full duration.
- Added tests for getRemainingGracePeriodDelay, including parameterized cases covering different grace period lengths to verify expected behavior.

### DIFF
--- a/apps/webapp/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.test.ts
+++ b/apps/webapp/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.test.ts
@@ -19,10 +19,34 @@
 
 import {TimeInMillis} from '@wireapp/commons/lib/util/TimeUtil';
 import {CredentialType} from '@wireapp/core/lib/messagingProtocols/mls';
+import {createDeterministicWallClock} from 'src/script/clock/deterministicWallClock';
 
 import {getEnrollmentTimer, getRemainingGracePeriodDelay, messageRetentionTime} from './EnrollmentTimer';
 
-import {MLSStatuses} from '../E2EIdentityVerification';
+import {MLSStatuses, WireIdentity} from '../E2EIdentityVerification';
+
+const generateWireIdentity = (
+  credentialType: CredentialType = CredentialType.X509,
+  status: MLSStatuses = MLSStatuses.NOT_ACTIVATED,
+): WireIdentity => ({
+  x509Identity: {
+    free: jest.fn(),
+    certificate: '',
+    displayName: 'John Doe',
+    domain: 'domain',
+    handle: 'johndoe',
+    notAfter: BigInt(0),
+    notBefore: BigInt(0),
+    serialNumber: '',
+    [Symbol.dispose]: () => {},
+  },
+  thumbprint: '',
+  credentialType,
+  status,
+  clientId: 'client-id',
+  deviceId: 'client-id',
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.ts
```diff
@@ -102,6 +102,15 @@ export function getEnrollmentTimer(
   return {isSnoozable: nextTick > 0, firingDate: Date.now() + nextTick};
 }
 
+export function getRemainingGracePeriodDelay(
+  identity: WireIdentity | undefined,
+  e2eiActivatedAt: number,
+  teamGracePeriodDuration: number,
+): number {
+  const {end} = getGracePeriod(identity, e2eiActivatedAt, teamGracePeriodDuration);
+  return Math.max(0, end - Date.now());
```
screendriver: Why no abstracted and deterministic [wall clock](https://github.com/wireapp/wire-webapp/blob/8499c865c3fa4c27b0f20747d2f0204df90caec7/apps/webapp/src/script/clock/wallClock.ts#L20) instead of a side effect which is not testable?

screendriver: Plus: this does not work for the new test case where `identity` is `undefined`.

`getGracePeriod()` only treats `CredentialType.Basic` as first enrollment. For `identity === undefined` or `status === NOT_ACTIVATED`, it falls through to the X509 path, reads `identity?.x509Identity?.notAfter`, and ends up with `NaN`. Then `Math.max(0, NaN)` also returns `NaN`.

This should probably align with `startTimers()`, which already treats `!identity`, `NOT_ACTIVATED`, and `Basic` as first E2EI activation.

apps/webapp/src/script/E2EIdentity/EnrollmentTimer/EnrollmentTimer.test.ts
```diff
@@ -99,4 +99,20 @@ describe('e2ei delays', () => {
...
+
+  it.each([
+    TimeInMillis.HOUR,
+    TimeInMillis.HOUR * 6,
+    TimeInMillis.HOUR * 12,
+    TimeInMillis.HOUR * 24,
+    TimeInMillis.WEEK,
+  ])('should keep full remaining grace period for first enrollment: %i ms', grace => {
+    const remainingDelay = getRemainingGracePeriodDelay(
+      undefined,
```
screendriver, line 146: This test covers the "full grace period" case but not the actual bug scenario where the grace period has already been active for some time.

Could we add at least one elapsed-time case? For example:

- activation happened 2 days ago
- team grace period is 7 days
- expected remaining delay is 5 days

That would protect the behavior this pull request is supposed to fix.

screendriver (approved):

PR #21525, opened 2026-06-12
[WPB-26249] Migrate from anta to QA federation env for federation tests

lizaveta-korshuk:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26249" title="WPB-26249" target="_blank">TaskWPB-26249</a>  [Web/QA] Migrate from anta to QA federation env for federation tests
  </td></table>
  <br />


markbrockhoff (changes requested):
I triggered a CI run for your branch: https://github.com/wireapp/wire-webapp/actions/runs/27529455602/job/81363978580

There are some failures due to call instability, but those are not affected by your changes. However inside `onPremLoginRedirect-TC-8757.spec.ts` it looks like the url of the federation environment was hardcoded which is causing it to fail now. Could you please also fix that? Ideally by making the function `claimDomain` on the brig api client accept the domains for the backend to federate as parameter. This way it will be more transparent where they are actually used. :)

apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
```diff
@@ -85,7 +85,8 @@ test.describe('account settings', () => {
     password: process.env.SSO_CLAIMED_USER_PASSWORD,
   });
 
-  test(
+  // Todo: This test is skipped because of disabled Okta account [WPB-26999]
```
markbrockhoff: Are you sure this needs to be skipped? Since the credentials have been updated to the claimed user ones it's working just fine for me ^^'
Maybe see #21820

lizaveta-korshuk: @zhannach please take a look

zhannach: @markbrockhoff I'll unskip it once the mentioned PR is merged.

markbrockhoff: Ok, but then it's time to rebase now as the mentioned PR was just merged @zhannach ;)

markbrockhoff (approved):

screendriver (approved):

PR #21529, opened 2026-06-13
feat: revert bg blur radius to 30 [WPB-25998]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25998" title="WPB-25998" target="_blank">StoryWPB-25998</a>  [Web] : Blur Effect Strength Reduced After Deployment on May 29
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/options.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/options.ts
@@ -104,7 +104,7 @@ export const defaultOpts = {
   smoothstepMax: 0.9,
   restartEvery: 0,
   bgBlur: 0,
-  bgBlurRadius: 5,
+  bgBlurRadius: 30,
 
   // Filter options.
   enableFilters: false,


apps/webapp/src/script/repositories/media/backgroundEffects/pipe/options.ts
```diff
@@ -104,7 +104,7 @@ export const defaultOpts = {
   smoothstepMax: 0.9,
   restartEvery: 0,
   bgBlur: 0,
-  bgBlurRadius: 5,
+  bgBlurRadius: 30,
```
screendriver, line 107: Out of curiosity: shouldn't some test be failing right now?

EnricoSchw (approved):

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/options.ts
zskhan: This is default config for fine tuning of the library. No tests here around it for now.

arjita-mitra (approved):

PR #21535, opened 2026-06-15
Migrate translations to typed translate function [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

- Replace deprecated `t()` usage with the typed `translate()` function
- Inject `translate` explicitly through app roots, context, repositories, entities, helpers, and tests
- Remove the generated/manual translation declaration file and related type generation path
- Remove localizer module mocks from tests in favor of explicit translation helpers

All critical flow tests [are also green](https://github.com/wireapp/wire-webapp/actions/runs/27748874590) for this change ✅


apps/webapp/src/script/util/test/createConversationForTest.ts
```diff
@@ -0,0 +1,32 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CONVERSATION_PROTOCOL} from '@wireapp/api-client/lib/team';
+
+import {Conversation} from 'Repositories/entity/Conversation';
+import type {Translate} from 'Util/localizerUtil';
+
+export function createConversationForTest(
```
markbrockhoff: Why create a wrapper function for the constructor with exactly the same parameters?

screendriver: Good catch. This wrapper does not add any behavior and was introduced during the migration churn. I’ll remove it and inline the Conversation constructor calls in tests.

Changed in 42cfd8d1befa080a652ac35777a4b9bcdbe30a61

arjita-mitra (approved):

PR #21557, opened 2026-06-16
fix: ignore stale default drive fetches(WPB-20817)

arjita-mitra:
<table>
<td>
  </td></table>
  <br />
 

## Summary
- Add request-version guarding to `useGetAllCellsNodes`
- Prevent stale default-drive fetch responses from writing nodes, pagination, status, or errors after a newer request/disabled state
- Add regression tests for disabled mid-flight and older-response-overwrites-newer cases


zskhan (approved):
LGTM

PR #21558, opened 2026-06-16
Integrate meetings UI with backend [WPB-25678]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25678" title="WPB-25678" target="_blank">TaskWPB-25678</a>  [Web] Create / edit meetings
  </td></table>
  <br />
 

Wire meetings list and schedule modal to the backend API so users see real meetings and can create them, using a repository layer and hook-based fetch/refresh instead of mock data.


zskhan (dismissed):

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -62,12 +71,34 @@ export const MeetingList = () => {
...
+  const {
+    today: meetingsToday,
+    tomorrow: meetingsTomorrow,
+    past: meetingsPast,
+  } = useMemo(() => partitionMeetingsByDay(meetings, wallClock), [meetings, wallClock]);
+
+  const groupedMeetingsTomorrow = groupByStartHour(meetingsTomorrow);
+  const groupedMeetingsPast = groupByStartHour(meetingsPast);
+
+  const hasMeetingsToday = meetingsToday.length > 0;
```
e-maad: Good to have:

```suggestion
  const hasMeetingsToday = is.nonEmptyArray(meetingsToday);
```

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -62,12 +71,34 @@ export const MeetingList = () => {
...
+  const hasMeetingsTomorrow = meetingsTomorrow.length > 0;
+  const hasMeetingsPast = meetingsPast.length > 0;
+
+  if (isLoading && meetings.length === 0) {
```
e-maad: Good to have:

```suggestion
  if (isLoading && is.nonEmptyArray(meetings)) {
```

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -62,12 +71,34 @@ export const MeetingList = () => {
...
+      <div css={emptyListContainerStyles} data-uie-name="meetings-list-loading">
+        <Loading data-uie-name="status-loading" />
+      </div>
+    );
+  }
 
-  const hasMeetingsToday = MEETINGS_TODAY.length > 0;
-  const hasMeetingsTomorrow = MEETINGS_TOMORROW.length > 0;
-  const hasMeetingsPast = MEETINGS_PAST.length > 0;
+  if (errorKey !== undefined && meetings.length === 0) {
```
e-maad: Good to have:
```suggestion
  if (!is.notNullOrUndefined(errorKey) && is.nonEmptyArray(meetings)) {
```

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/ScheduleMeetingModal.tsx
```diff
@@ -74,14 +76,16 @@ export const ScheduleMeetingModal = () => {
     reset();
   };
 
-  const handleSubmit = () => {
+  const handleSubmit = async () => {
```
e-maad: Calling a promise from `click event handler`, don't you think we can utilize fire and forget invoker for better logging/monitoring?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.test.ts
```diff
@@ -0,0 +1,106 @@
...
+import {User} from 'Repositories/entity/User';
+
+import {ScheduleMeetingService} from './scheduleMeetingService';
+import type {MeetingsListRefresher, ScheduleMeetingNotifier} from './scheduleMeetingService.types';
+import type {ScheduleMeetingFormState} from './scheduleMeetingTypes';
+import {MeetingsRepository} from 'Repositories/meetings';
+
+const formState: ScheduleMeetingFormState = {
+  title: 'Weekly sync',
+  start: new Date('2026-06-15T10:00:00.000Z'),
```
e-maad, line 28: Can we use wallClock to avoid sideEffects?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -0,0 +1,50 @@
...
+import type {ScheduleMeetingFormState} from './scheduleMeetingTypes';
+
+export class ScheduleMeetingService {
+  constructor(
+    private readonly meetingsRepository: MeetingsRepository,
+    private readonly meetingsListRefresher: MeetingsListRefresher,
+    private readonly notifier: ScheduleMeetingNotifier,
+  ) {}
+
+  async scheduleMeeting(formState: ScheduleMeetingFormState): Promise<boolean> {
```
e-maad: Can we change the name in a way that it matches with return type?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/useScheduleMeetingSubmit.ts
```diff
@@ -0,0 +1,58 @@
...
+import {ScheduleMeetingService} from './scheduleMeetingService';
+import type {ScheduleMeetingFormState} from './scheduleMeetingTypes';
+
+export const useScheduleMeetingSubmit = (onMeetingScheduled?: () => Promise<void>) => {
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const rootContext = useContext(RootContext);
+  const meetingsRepository = rootContext?.mainViewModel.content.repositories.meetings;
+
+  const scheduleMeetingService = useMemo(() => {
+    if (!meetingsRepository) {
```
e-maad: Is `meetingsRepository` null/undefined? Can we be more specific?

e-maad (commented):
Added few comments.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.test.ts
thisisamir98: this is a test file, wallclock is already used in the module

e-maad: I remember there is a fakeWallClock for tests.

e-maad (commented):
LGTM 🚀

zskhan (approved):

screendriver (approved):
This looks much better to me now, thanks for adjusting it 👍 

The service now returns semantic results instead of knowing about modals or translated copy. That is exactly the boundary that we were talking about.

I also like that recurrence is now mapped as a semantic value and translated where it is rendered in React. Same for the meeting list item.

Small nit: the error translation key mapping is still UI-related, but since it only contains keys and is consumed from the React hook I think this is fine for this pull request.

So from the separation-of-concerns perspective we discussed, this is now in a much better shape.

PR #21561, opened 2026-06-16
fix: show admin promotion message for self user [WPB-26186]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26186" title="WPB-26186" target="_blank">BugWPB-26186</a>  [Web] System message is not visible for newly promoted admin
  </td></table>
  <br />
 

### Description

Show admin promotion message for self user.

image


libraries/react-ui-kit/src/DataDisplay/Icon/ConversationSettingsIcon.tsx
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {SVGIcon, SVGIconProps} from './SVGIcon';
+
+export const ConversationSettingsIcon = (props: SVGIconProps) => (
+  <SVGIcon realWidth={14} realHeight={14} {...props}>
```
arjita-mitra, line 23: aria-hidden="true", SR users shouldn't hear about icons but they should be able to hear the promotion message

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4203,6 +4194,44 @@ export class ConversationRepository {
...
   }
 
+  /**
+   * A member's conversation role was updated.
+   *
+   * @param conversationId Qualified id of the conversation the role update happened in
+   * @param eventData Data of the 'conversation.member-update' event
+   * @param from Id of the user who triggered the role update
+   */
+  private async onConversationRoleUpdate(
```
arjita-mitra: could you refactor this method following clean code SRP Principle?

you can break it down to-

findConversation() - finds the conversation
extractUserId() - extracts user ID from event data
updateConversationRoles() - updates roles
isPromotionToAdmin() - checks promotion logic
notifyAdminPromotion() - sends notification

this will help with readability, reduce cognitive load and we can also get rid of nested early return conditions.

Individual methods can be tested as well.

arjita-mitra (changes requested):

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
e-maad: > findConversation() - finds the conversation

Only adding `getConversationFromStateById`, other wrapper functions will be useless.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4238,6 +4229,44 @@ export class ConversationRepository {
...
```
zskhan:
```suggestion
  private async onConversationMemberRoleUpdated(
```

zskhan (approved):

PR #21565, opened 2026-06-17
Reduce Jest output noise in continuous integration [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

Make the GitHub Actions test output easier to read by using quieter Jest reporting in continuous integration.

CI should show the important signal: failing tests, GitHub annotations, the test summary, and the coverage summary. It should not print unnecessary per-suite or full coverage-table output when the run is successful.

This keeps local Jest output unchanged for developer feedback, but switches CI to the built-in GitHub Actions and summary reporters. Coverage reporting also uses text-summary in CI so the logs stay useful without becoming noisy.

The intent is to make CI failures easier to scan and reduce log noise while keeping coverage thresholds, coverage artifacts, and actionable failure output.

### DIFF
--- a/apps/server/jest.config.js
+++ b/apps/server/jest.config.js
@@ -17,6 +17,8 @@
  *
  */
 
+const isContinuousIntegrationEnvironment = process.env.CI === 'true';
+
 module.exports = {
   preset: '../../jest.preset.js',
   collectCoverageFrom: [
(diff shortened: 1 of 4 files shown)


e-maad (approved):
LGTM 🚀

PR #21569, opened 2026-06-17
feat: external links should work on detached window - [WPB-26355]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26355" title="WPB-26355" target="_blank">BugWPB-26355</a>  [Desktop] Support Article link for camera is not clickable in maximized window on Mac
  </td></table>
  <br />
 

Clicking on the support article was not doing anything on Detached Call mode.
image

### DIFF
--- a/apps/webapp/src/script/repositories/calling/CallingRepository.ts
+++ b/apps/webapp/src/script/repositories/calling/CallingRepository.ts
@@ -153,6 +153,19 @@ type SubconversationData = {
   members: SubconversationEpochInfoMember[];
 };
 
+export const setupDetachedWindowExternalLinksClick = (detachedWindow: Window, openerWindow: Window): (() => void) => {
+  const handleClick = (event: MouseEvent) => {
+    const anchor = (event.target as HTMLElement).closest('a');
+    if (anchor?.target === '_blank' && anchor.href) {
+      event.preventDefault();
+      openerWindow.open(anchor.href);
+    }
+  };
+
+  detachedWindow.document.addEventListener('click', handleClick, true);
+  return () => detachedWindow.document.removeEventListener('click', handleClick, true);
+};
+
 export class CallingRepository {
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1498,6 +1511,8 @@ export class CallingRepository {
     detachedWindow.addEventListener('pagehide', this.closeDetachedWindow);
     window.addEventListener('pagehide', this.onPageHide);
 
+    setupDetachedWindowExternalLinksClick(detachedWindow, window);
```
arjita-mitra, line 1514: shouldn't we use globalThis here?

zskhan: Not only at this place but in overall repository it should be replaced.

screendriver: > shouldn't we use globalThis here?

I would prefer not to use `globalThis` here. To me it has the same downside as `window`: it still hides a dependency on global browser state.

The helper is intentionally written with explicit `Window` dependencies, which makes it easier to test and keeps the browser access visible. The current `window` usage at the call site is the boundary to the browser environment.

I agree that we should clean this up more consistently across the repository, ideally behind a small adapter/service, but I would keep that as a separate refactoring rather than replacing this one usage with `globalThis`.

arjita-mitra (approved):

screendriver (approved):
Approved from my side.

One small non-blocking thought: `setupDetachedWindowExternalLinksClick()` returns a cleanup function, but the production call currently does not store or call it. Since the detached window/document should go away anyway, I would not block this pull request on it. But longer term it might be cleaner to either wire the cleanup into `closeDetachedWindow()` or remove the returned cleanup if we do not need it in production.

PR #21572, opened 2026-06-17
feat: show loading spinner forever for AV preferences when any of the permission is denied - [WPB-26451]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26451" title="WPB-26451" target="_blank">TaskWPB-26451</a>  [web] - AV Preferences keeps loading when camera permissions are denied
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/hooks/useInitializeMediaDevices.test.ts
+++ b/apps/webapp/src/script/hooks/useInitializeMediaDevices.test.ts
@@ -35,10 +35,6 @@ const createStreamHandler = (requestMediaStreamAccess = jest.fn().mockResolvedVa
   }) as unknown as MediaStreamHandler;
 
 describe('useInitializeMediaDevices', () => {
-  beforeEach(() => {
-    jest.clearAllMocks();
-  });
-
   it('requests media access, stops received tracks and initializes media devices', async () => {
     const stop = jest.fn();
     const stream = {
--- a/apps/webapp/src/script/hooks/useInitializeMediaDevices.test.ts
+++ b/apps/webapp/src/script/hooks/useInitializeMediaDevices.test.ts
@@ -1,0 +1,95 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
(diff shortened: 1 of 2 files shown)


screendriver (dismissed):

apps/webapp/src/script/hooks/useInitializeMediaDevices.ts
```diff
@@ -36,10 +36,10 @@ export const useInitializeMediaDevices = (devicesHandler: MediaDevicesHandler, s
         stream.getTracks().forEach(track => track.stop());
       }
       await devicesHandler?.initializeMediaDevices(true, true);
-      setAreMediaDevicesInitialized(true);
     } catch (error: unknown) {
       logger.warn(`Initialization of media devices failed:`, error);
-      setAreMediaDevicesInitialized(false);
```
screendriver, line 42: No tests are failing after this change?

zskhan: Tests added

apps/webapp/src/script/hooks/useInitializeMediaDevices.test.ts
```diff
@@ -0,0 +1,95 @@
...
+const createStreamHandler = (requestMediaStreamAccess = jest.fn().mockResolvedValue(undefined)) =>
+  ({
+    requestMediaStreamAccess,
+  }) as unknown as MediaStreamHandler;
+
+describe('useInitializeMediaDevices', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
```
screendriver: I do not think this is needed. All mocks are created per test, either directly or through the factory functions, so there is no shared mock state to clear between tests. Removing this keeps the test setup simpler and avoids suggesting that the tests depend on global mock cleanup.


```suggestion
```

screendriver (changes requested):

screendriver (approved):

PR #21577, opened 2026-06-18
fix: clear filter and no search query should reloads the unfiltered search-owned list(WPB-20817)

arjita-mitra:
<table>
<td>
  </td></table>
  <br />
 

**##Summary**
when filters go from active to none with no query, the hook now reloads the unfiltered search-owned list if search mode still owns the view, instead of calling browse onClear.

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
@@ -190,11 +190,13 @@ describe('useConversationSearchFiles', () => {
     const cellsRepository = {searchNodes: jest.fn().mockReturnValue(search.promise)} as FakeCellsRepository;
     const fireAndForgetInvoker = createExecutingFireAndForgetInvokerForTest();
 
-    useCellsStore.getState().setNodes({
-      conversationId: CONV_ID,
-      nodes: [staleFolderNode],
+    act(() => {
+      useCellsStore.getState().setNodes({
+        conversationId: CONV_ID,
+        nodes: [staleFolderNode],
+      });
+      useCellsStore.getState().setStatus('success');
     });
-    useCellsStore.getState().setStatus('success');
 
     renderSearchHook({cellsRepository, fireAndForgetInvoker});
 
@@ -279,9 +281,49 @@ describe('useConversationSearchFiles', () => {
     );
   });
 
-  it('calls onClear when the search input is emptied', async () => {
+  it('reloads the unfiltered search list when all filters are cleared with no query', async () => {
     const onClear = jest.fn();
-    const {result, fireAndForgetInvoker} = renderSearchHook({onClear});
+    const cellsRepository = createFakeCellsRepository({
+      Nodes: [createRestNode('all-files.pdf')],
+    });
+    const fireAndForgetInvoker = createExecutingFireAndForgetInvokerForTest();
+    const {rerender} = renderHook(
+      ({filters}: {filters: ConversationDriveFiltersState}) =>
+        useConversationSearchFiles({
+          cellsRepository: cellsRepository as unknown as CellsRepository,
+          userRepository: createFakeUserRepository() as unknown as UserRepository,
+          conversationQualifiedId: QUALIFIED_ID,
+          enabled: true,
+          fireAndForgetInvoker,
+          filters,
+          onClear,
+        }),
+      {
+        initialProps: {
+          filters: {
+            ...emptyFilters,
+            selectedFileTypeIds: ['pictures'],
+          },
+        },
+      },
+    );
+
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+
+    act(() => rerender({filters: emptyFilters}));
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+
+    expect(cellsRepository.searchNodes).toHaveBeenLastCalledWith(expect.objectContaining({query: '', recursive: false}));
+    expect(useCellsStore.getState().getNodes({conversationId: CONV_ID})[0]?.name).toBe('all-files.pdf');
+    expect(onClear).not.toHaveBeenCalled();
+  });
+
+  it('reloads the unfiltered search list when the search input is emptied while search is open', async () => {
+    const onClear = jest.fn();
+    const cellsRepository = createFakeCellsRepository({
+      Nodes: [createRestNode('all-files.pdf')],
+    });
+    const {result, fireAndForgetInvoker} = renderSearchHook({cellsRepository, onClear});
     await act(() => fireAndForgetInvoker.waitUntilAllSettled());
 
     act(() => result.current.handleSearch('test'));
@@ -290,7 +332,9 @@ describe('useConversationSearchFiles', () => {
     act(() => result.current.handleSearch(''));
     await act(() => fireAndForgetInvoker.waitUntilAllSettled());
 
-    expect(onClear).toHaveBeenCalled();
+    expect(cellsRepository.searchNodes).toHaveBeenLastCalledWith(expect.objectContaining({query: '', recursive: false}));
+    expect(useCellsStore.getState().getNodes({conversationId: CONV_ID})[0]?.name).toBe('all-files.pdf');
+    expect(onClear).not.toHaveBeenCalled();
   });
 
   it('clears search-owned rows before handing control back to browse mode', async () => {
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
```diff
@@ -310,10 +311,16 @@ export const useConversationSearchFiles = ({
   // restore the default unfiltered file list.
   useEffect(() => {
     if (hadActiveSearchParamsRef.current && !hasActiveParams && !searchValue) {
```
screendriver: Could we make this condition explicit and avoid truthy/falsy checks?

For me, truthy/falsy checks are always a problem because they hide the actual condition. Here `!searchValue` means "empty string", but the hook already treats whitespace-only input as empty in `handleSearch` / `normalizeSearchQuery`.

So this condition currently behaves differently for `''` and `' '`.

Could we make this explicit, for example by checking the normalized value?

```typescript
const hasNoSearchQuery = normalizeSearchQuery(searchValue).length === 0;

if (hadActiveSearchParamsRef.current && hasActiveParams === false && hasNoSearchQuery) {
  // ...
}
```

and/or use [@sindresorhus/is](https://www.npmjs.com/package/@sindresorhus/is) to make the type guard more visible and explicit?

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
```diff
@@ -251,7 +251,8 @@ export const useConversationSearchFiles = ({
     setSearchQuery('');
     shouldPerformSearch.current = false;
 
-    const shouldRefreshSearchResults = preserveFilters && hasActiveParams;
+    const shouldRefreshSearchResults =
+      preserveFilters && (enabledRef.current || (allowSearchWhenDisabledRef.current && hasActiveParams));
```
screendriver: This ownership condition is getting hard to reason about.

We now have this check here:

> preserveFilters && (enabledRef.current || (allowSearchWhenDisabledRef.current && hasActiveParams))

and a similar-but-not-identical check below in the effect:

> enabledRef.current || allowSearchWhenDisabledRef.current

The difference may be intentional because one path runs before filters are cleared and the other runs after filters are cleared, but that is not obvious from the code.

Could we extract this into a named predicate?

In addition please try to avoid truthy/falsy checks and construct checks explicitly to make the intention cristal clear.

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
```diff
@@ -279,9 +281,49 @@ describe('useConversationSearchFiles', () => {
...
+        },
+      },
+    );
+
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+
+    act(() => rerender({filters: emptyFilters}));
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+
+    expect(cellsRepository.searchNodes).toHaveBeenLastCalledWith(expect.objectContaining({query: '', recursive: false}));
```
screendriver: `toHaveBeenLastCalledWith()` makes this test weaker than it looks.

The important behavior is not only that the last call reloads the unfiltered list. We also want to know that clearing the filters does not trigger extra duplicate reloads.

Could we capture the call count before `rerender()`, then assert that exactly one additional search happened and that this new call contains `{query: '', recursive: false}`?

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
```diff
@@ -290,21 +332,25 @@ describe('useConversationSearchFiles', () => {
     act(() => result.current.handleSearch(''));
     await act(() => fireAndForgetInvoker.waitUntilAllSettled());
 
-    expect(onClear).toHaveBeenCalled();
+    expect(cellsRepository.searchNodes).toHaveBeenLastCalledWith(expect.objectContaining({query: '', recursive: false}));
```
screendriver: Same here: `toHaveBeenLastCalledWith()` would still pass if clearing the input triggered an extra unexpected search before the final expected call.

Could we assert the call count before clearing the input, then verify that exactly one additional call was made with `{query: '', recursive: false}`?

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: Using normalizeSearchQuery(searchValue).length === 0 makes contract explicit and avoids duplicating the whitespace rule in another place. @sindresorhus/is already makes sense inside normalizeSearchQuery is already there.

Resolved 3858306b825b9edd0b96d4fd2f3a8c803fae01dd

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: Addressed 8186d687d20c745ba421cd8acd8f19e61099db8a

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
arjita-mitra: Addressed 2e5cb39e85d11daf2bd7d10c3ad6d9b6dba34845

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.test.ts
arjita-mitra: Addressed 2e5cb39e85d11daf2bd7d10c3ad6d9b6dba34845

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
```diff
@@ -88,12 +88,38 @@ export const useConversationSearchFiles = ({
...
 
-  const isCurrentSearchRequest = useCallback((requestVersion: number): boolean => {
-    return (
-      (enabledRef.current || allowSearchWhenDisabledRef.current) && !requestVersionGate.current.isStale(requestVersion)
-    );
+  const canSearchOwnResults = useCallback((): boolean => {
+    return enabledRef.current === true || allowSearchWhenDisabledRef.current === true;
   }, []);
 
+  const isCurrentSearchRequest = useCallback(
```
zskhan: Maybe a better name like `isValidSearchRequest`

screendriver: One additional thought: I would also be careful with `useCallback()`. It is not automatically a performance improvement by itself. Unless we need the stable function identity for a concrete reason, or we have **measured** that it improves rendering behavior or reduces CPU work, I would prefer the simpler version without memoization.

Small memoization can easily become a micro-optimization that adds complexity without a proven benefit.

(this also applies to all other `useCallback()` here

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
```diff
@@ -88,12 +88,38 @@ export const useConversationSearchFiles = ({
...
+  );
+
+  const shouldRefreshSearchResultsAfterClearingInput = useCallback(
+    ({
+      preserveFilters,
+      hasActiveParamsBeforeClear,
+    }: {
+      preserveFilters: boolean;
+      hasActiveParamsBeforeClear: boolean;
+    }): boolean => {
```
screendriver: What do you think about extracting this into its own type alias instead of inlining it? Would help regarding readability

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: i have not added for performance reason. the design of wire cells is such a way we face frequent referential integrity issue and the app keeps looping. therefore i would like to keep it.

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: Addressed e6690d4ed5fc4a6d7b866d7736dd6580ab88fcf2

apps/webapp/src/script/components/Conversation/ConversationCells/useConversationSearch/useConversationSearchFiles.ts
arjita-mitra: > Maybe a better name like `isValidSearchRequest`

Addressed e6690d4ed5fc4a6d7b866d7736dd6580ab88fcf2

screendriver (approved):

PR #21581, opened 2026-06-18
Prevent open redirect through join parameters [WPB-25894]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25894" title="WPB-25894" target="_blank">securityWPB-25894</a>  [Webapp] Open redirect in /join handler
  </td></table>
  <br />
 

## Summary

Fixes a security issue in the `/join` handler.

The handler previously interpolated `key` and `code` query values directly into the redirect `Location` header. Because these values were not safely encoded, an attacker could use encoded query separators such as `&` or `#` to escape the intended `join_key` / `join_code` values.

This could inject additional query parameters such as `destination_url` or replace the expected `#/join-conversation` fragment.

## Security impact

This is an open redirect issue.

In the worst case, an attacker could craft a link that starts on the trusted `app.wire.com` domain, but eventually redirects the user to an attacker-controlled website.

That attacker-controlled website could imitate Wire or an SSO/login flow and trick the user into entering credentials, recovery information, or other sensitive data.

This does not imply direct server compromise or authentication bypass by itself, but it is still security-relevant because the trusted Wire domain can be abused as part of a phishing or social-engineering flow.

## Intent

The client-side redirect route should not be treated as the security boundary.

The `/join` server handler must only emit a safe redirect target. User-controlled values must stay inside the intended `join_key` and `join_code` query parameters and must not be able to change the final redirect target.

### DIFF
--- a/apps/server/routes/redirectRoutes.ts
+++ b/apps/server/routes/redirectRoutes.ts
@@ -33,7 +33,7 @@ type JoinRedirectQuery = {
   readonly code?: unknown;
 };
 
-export function createJoinRedirectLocation(query: JoinRedirectQuery): Maybe<string> {
+export function createJoinConversationRedirectUrl(query: JoinRedirectQuery): Maybe<string> {
   const {key, code} = query;
 
   if (!is.nonEmptyString(key) || !is.nonEmptyString(code)) {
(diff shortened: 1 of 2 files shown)


apps/server/routes/redirectRoutes.ts
```diff
@@ -17,15 +17,53 @@
...
 import * as BrowserUtil from '../util/browserUtil';
 
 const router = express.Router();
 
+type JoinRedirectQuery = {
+  readonly key?: unknown;
+  readonly code?: unknown;
+};
+
+export function createJoinRedirectLocation(query: JoinRedirectQuery): Maybe<string> {
```
zskhan:
```suggestion
export function createJoinConversationRedirectUrl(query: JoinRedirectQuery): Maybe<string> {
```

screendriver: Done 1e727dd66965dacd44bf9f39f8f5838cde6310f8

zskhan (approved):

PR #21585, opened 2026-06-18
People search update on query change [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

The Connect people search stopped reacting to new input after the search effect was changed to depend only on the debounced callback.

The debounced callback still reads the current search query, so changing the query must trigger the effect explicitly. Without that, the people tab can keep showing the initial unfiltered user list even after the user types a search term.

Include the search query in the effect dependencies and add regression coverage that rerenders the people search with a new query and verifies that only matching users remain visible.

### DIFF
--- a/apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.tsx
+++ b/apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.tsx
@@ -49,7 +49,7 @@ export type SearchResultsData = {contacts: User[]; others: User[]};
 const TOP_PEOPLE_LIMIT = 6;
 const SEARCH_DEBOUNCE_MILLISECONDS = 300;
 
-interface PeopleTabProps {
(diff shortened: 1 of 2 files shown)


e-maad (dismissed):

apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.tsx
```diff
@@ -214,7 +214,7 @@ export const PeopleTab = ({
 
   useEffect(() => {
     void debouncedSearch();
```
zskhan, line 216: fireAndForget?

screendriver: Yes. But not now. I did not change that, so I kept that on purpose (small and focused pull requests)

zskhan (approved):

PR #21587, opened 2026-06-18
Use fire-and-forget invoker for people search [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

As a follow-up to https://github.com/wireapp/wire-webapp/pull/21585#discussion_r3436812571

### DIFF
--- a/apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.test.tsx
+++ b/apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.test.tsx
@@ -20,7 +20,13 @@
 import {render, screen, waitFor} from '@testing-library/react';
 import ko from 'knockout';
 
-import {withTheme} from 'src/script/auth/util/test/TestUtil';
+import {StyledApp, THEME_ID} from '@wireapp/react-ui-kit';
+
+import {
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/page/LeftSidebar/panels/StartUI/PeopleTab.test.tsx
```diff
@@ -90,6 +96,13 @@ function createTeamState(): TeamState {
   return teamState;
 }
 
+function withThemeAndRootContext(
```
zskhan: Can we make it reusable by moving it into TestUtil.tsx just like `withTheme`?

screendriver: Good point, thanks. I moved the wrapper into `TestUtil.tsx` next to `withTheme()` and updated `PeopleTab.test.tsx` to use the shared helper.

I kept the helper accepting an explicit `rootProviderWrapper`, because this test needs a custom `RootContext` value instead of the default context used by `withTheme`.

Done in 6d36c89475b4001a77ba463dba01ebf904ef983d

zskhan (approved):

PR #21603, opened 2026-06-19
Fix WebSocket reconnect after sleep/wake token refresh race [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Problem

WebSocket recovery can still fail after a MacBook sleep/wake cycle.

Even after forcing WebSocket reconnects on wake and replacing terminal `CLOSED` wrappers, the app can still get stuck in the reconnect state.

A typical failure pattern after wake looks like this:

- the network changes while the laptop wakes up
- HTTP requests fail with `ERR_NETWORK_CHANGED`
- some requests return `401` or `409`
- the `/await` WebSocket reconnects while access-token refresh is also happening
- the app stays in a reconnecting state instead of recovering to `LIVE`

## Root cause

`WebSocketClient.refreshAccessToken()` only used a boolean guard for concurrent refreshes.

When a refresh was already in progress, another reconnect attempt could return immediately instead of waiting for the same refresh to finish.

That meant WebSocket reconnect could continue and build a WebSocket URL with a stale or expired token.

Refresh failures were also swallowed after emitting the existing error events, so reconnect could continue even though the token refresh did not succeed.

## Change

This makes WebSocket reconnect wait for access-token refresh correctly.

The WebSocket client now:

- shares an in-flight access-token refresh promise between concurrent reconnect attempts
- waits for the refresh to complete before building the WebSocket URL
- rejects reconnect when token refresh fails
- preserves the existing invalid-token and generic error events
- avoids building a WebSocket URL with a known stale token

## Why

After sleep/wake, the browser can produce several network and authentication failures at once.

WebSocket reconnect must not race ahead of access-token recovery. Otherwise the client can reconnect with invalid credentials and remain stuck in the reconnecting state.

### DIFF
--- a/libraries/api-client/src/tcp/webSocketClient.ts
+++ b/libraries/api-client/src/tcp/webSocketClient.ts
@@ -68,7 +68,7 @@ export type WebSocketClientOptions = {
 
 export class WebSocketClient extends EventEmitter {
   private clientId?: string;
-  private isRefreshingAccessToken: boolean;
+  private accessTokenRefreshPromise: Maybe<Promise<void>> = Maybe.nothing<Promise<void>>();
   private readonly baseUrl: string;
   private readonly logger: logdown.Logger;
   private readonly socket: ReconnectingWebsocket;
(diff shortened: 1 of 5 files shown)


libraries/api-client/src/tcp/webSocketClient.ts
```diff
@@ -193,11 +201,24 @@ export class WebSocketClient extends EventEmitter {
   }
 
   private async refreshAccessToken(): Promise<void> {
-    if (this.isRefreshingAccessToken) {
-      return;
+    if (this.accessTokenRefreshPromise !== undefined) {
```
zskhan, line 204:
```suggestion
    if (!is.undefined(this.accessTokenRefreshPromise)) {
```

zskhan (approved):

PR #21612, opened 2026-06-22
Keep WebSocket reconnect alive during token refresh failures [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Problem

WebSocket recovery can still get stuck after a MacBook sleep/wake cycle.

The observed failure pattern is:

* the Mac wakes up and the network changes
* requests temporarily fail with `ERR_NETWORK_CHANGED`
* some requests return `401`
* the `/await` WebSocket reconnects during the same recovery window
* access-token refresh can fail transiently
* the app remains stuck in a reconnecting state instead of recovering

## Root cause

The WebSocket URL is produced by an async reconnect callback.

After the previous fix https://github.com/wireapp/wire-webapp/pull/21603, that callback correctly waited for access-token refresh and rejected when refresh failed. That avoided reconnecting with a stale token.

However, the underlying reconnecting WebSocket library is not robust when the async URL provider rejects during a reconnect attempt. A transient token-refresh failure after wake can therefore stop reconnect progress instead of allowing the reconnect loop to continue.

So the problem is not that we should reconnect with an invalid token.

The problem is that a temporary token-refresh failure must not permanently break the reconnect loop.

## Change

This makes the reconnect URL provider resilient to transient token-refresh failures.

The WebSocket client now keeps reconnect recovery alive while waiting for a valid access token:

* concurrent reconnect attempts still share the same in-flight token refresh
* transient token-refresh failures are retried instead of escaping from the URL provider
* the WebSocket URL is only built after a valid token is available
* fatal invalid-session handling is preserved
* raw access tokens are not logged

### DIFF
--- a/libraries/api-client/src/tcp/webSocketClient.test.ts
+++ b/libraries/api-client/src/tcp/webSocketClient.test.ts
@@ -20,6 +20,7 @@
 /* eslint-disable dot-notation */
 
 import {WebSocketClient} from './webSocketClient';
+import {ReconnectingWebsocketWallClock} from './reconnectingWebsocket';
 import {noop} from 'noop-esm';
 
 import {InvalidTokenError} from '../auth/authenticationError';
@@ -34,6 +35,8 @@ const accessTokenPayload = {
   token_type: 'Bearer',
   user: 'aaf9a833-ef30-4c22-86a0-9adc8a15b3b4',
 };
+
+const reconnectPendingMarker = 'reconnect-pending';
 
 const fakeHttpClient: any = {
   accessTokenStore: {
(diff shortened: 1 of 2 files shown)


thisisamir98 (approved):

arjita-mitra (approved):

arjita-mitra:
some requests return 401
the /await WebSocket reconnects during the same recovery window
access-token refresh can fail transiently
the app remains stuck in a reconnecting state instead of recovering

I have faced this same issue several times in the past. Good catch @screendriver i think this will improve the re-connection.

PR #21620, opened 2026-06-22
feat: De-duplicate user ids before owner lookup[WPB-26426]

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26426" title="WPB-26426" target="_blank">TaskWPB-26426</a>  De-duplicate user ids before owner lookup
  </td></table>
  <br />
 

### ##Summary
getUsersFromNodes passes every node owner id into userRepository.getUsersById, including duplicates.

This is avoidable work and may add delay after the Cells API response, because UI loading remains active until user lookup finishes.

Proposed change: build a unique list of valid qualified user ids before calling getUsersById.

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
@@ -24,6 +24,26 @@ import {UserRepository} from 'Repositories/user/userRepository';
 
 import {getUserQualifiedIdFromNode} from '../common/getUserQualifiedIdFromNode/getUserQualifiedIdFromNode';
 
+const getQualifiedIdKey = ({domain, id}: QualifiedId): string => `${domain}/${id}`;
+
+const getUniqueQualifiedIds = (userIds: QualifiedId[]): QualifiedId[] => {
+  const uniqueUserIds: QualifiedId[] = [];
+  const uniqueUserKeys = new Set<string>();
+
+  userIds.forEach(userId => {
+    const key = getQualifiedIdKey(userId);
+
+    if (uniqueUserKeys.has(key)) {
+      return;
+    }
+
+    uniqueUserKeys.add(key);
+    uniqueUserIds.push(userId);
+  });
+
+  return uniqueUserIds;
+};
+
 export const getUsersFromNodes = async ({
   nodes,
   userRepository,
@@ -35,7 +55,7 @@ export const getUsersFromNodes = async ({
     return [];
   }
 
-  return userRepository.getUsersById(
-    nodes.map(node => getUserQualifiedIdFromNode(node)).filter(Boolean) as QualifiedId[],
-  );
+  const userIds = nodes.map(node => getUserQualifiedIdFromNode(node)).filter(Boolean) as QualifiedId[];
+
+  return userRepository.getUsersById(getUniqueQualifiedIds(userIds));
 };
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
```diff
@@ -35,7 +55,7 @@ export const getUsersFromNodes = async ({
     return [];
   }
 
-  return userRepository.getUsersById(
-    nodes.map(node => getUserQualifiedIdFromNode(node)).filter(Boolean) as QualifiedId[],
-  );
+  const userIds = nodes.map(node => getUserQualifiedIdFromNode(node)).filter(Boolean) as QualifiedId[];
```
e-maad: I don't think we need `as QualifiedId[]`

e-maad (dismissed):
Added a comment.

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
```diff
@@ -24,6 +24,26 @@ import {UserRepository} from 'Repositories/user/userRepository';
 
 import {getUserQualifiedIdFromNode} from '../common/getUserQualifiedIdFromNode/getUserQualifiedIdFromNode';
 
+const getQualifiedIdKey = ({domain, id}: QualifiedId): string => `${domain}/${id}`;
+
+const getUniqueQualifiedIds = (userIds: QualifiedId[]): QualifiedId[] => {
+  const uniqueUserIds: QualifiedId[] = [];
```
zskhan: When we already have a set then do we really need another array to keep the unique IDs? We can have Map instead of Set.

ikotarac: I feel like Map and Set would be similar complexity. Is there a requirement that if there is duplicate qualified key we take the first userId with that key?

we could have an one liner using function from lodash:
```ts
const getUniqueQualifiedIds = (userIds: QualifiedId[]): QualifiedId[] => uniqBy(userIds, getQualifiedIdKey);
```

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
ikotarac: also no need for arrow function:

```ts
const userIds = nodes.map(getUserQualifiedIdFromNode).filter(Boolean);
```

ikotarac:
image

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
arjita-mitra: @e-maad Addressed efe8ef0a56cc83e422d84854cad7995a502a6e15

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/getUsersFromNodes.ts
arjita-mitra: @ikotarac possibly it doesn't matter.

arjita-mitra: @zskhan addressed efe8ef0a56cc83e422d84854cad7995a502a6e15

arjita-mitra: I have made the changes because it has improved the redability.

e-maad (approved):

PR #21628, opened 2026-06-23
Exclude meeting conversations from the sidebar list [WPB-25678]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25678" title="WPB-25678" target="_blank">TaskWPB-25678</a>  [Web] Create / edit meetings
  </td></table>
  <br />
 

Meeting conversations remain in state for messaging but are filtered out of sorted/visible conversations so they do not appear alongside regular chats.

### DIFF
--- a/apps/webapp/src/script/repositories/conversation/ConversationSelectors.ts
+++ b/apps/webapp/src/script/repositories/conversation/ConversationSelectors.ts
@@ -143,7 +143,7 @@ export const isProteus1to1ConversationWithUser = (userId: QualifiedId) =>
 export const isMLS1to1ConversationWithUser = (userId: QualifiedId) =>
   is1to1ConversationWithUser(userId, CONVERSATION_PROTOCOL.MLS);
 
-export const isMeetingConversation = (conversation: Conversation): boolean => {
+export const isConversationForScheduledMeeting = (conversation: Conversation): boolean => {
   return conversation.groupConversationType() === GROUP_CONVERSATION_TYPE.MEETING;
 };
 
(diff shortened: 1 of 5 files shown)


libraries/api-client/src/conversation/conversation.ts
```diff
@@ -57,6 +57,7 @@ export enum CONVERSATION_ACCESS {
 export enum GROUP_CONVERSATION_TYPE {
```
zskhan, line 57: Maybe we can rename it to `CONVERSATION_TYPE`

zskhan (dismissed):

libraries/api-client/src/conversation/conversation.ts
thisisamir98: this is pre-existing and same name comes from the backend

apps/webapp/src/script/repositories/conversation/ConversationSelectors.ts
```diff
@@ -139,6 +143,10 @@ export const isProteus1to1ConversationWithUser = (userId: QualifiedId) =>
 export const isMLS1to1ConversationWithUser = (userId: QualifiedId) =>
   is1to1ConversationWithUser(userId, CONVERSATION_PROTOCOL.MLS);
 
+export const isMeetingConversation = (conversation: Conversation): boolean => {
```
e-maad: The name is really confusing, I know where this is coming from but is it possible to rename it?
Maybe `isConversationForScheduledMeeting`.

e-maad (dismissed):
LGTM 🚀 
Just one concern.

apps/webapp/src/script/repositories/conversation/ConversationSelectors.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21628/commits/7fc0f64c834d77c0ddae8fbfc1f8851a94d7061d

zskhan (approved):

PR #21629, opened 2026-06-23
allow hosts to edit scheduled meetings from the calendar list [WPB-25503]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25503" title="WPB-25503" target="_blank">TaskWPB-25503</a>  [Web] Edit a meeting
  </td></table>
  <br />
 

- Wire up the existing schedule modal in edit mode so meeting hosts can update title, participants, times, and recurrence from the context menu, with list refresh on confirm and no success toast.
- Replace nullable dates and editingMeetingId with Maybe across schedule/edit form state, validation, mappers, and tests.


apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -57,3 +64,44 @@ export async function tryScheduleMeeting(
...
+
+  const {meetingsRepository, fetchMeetings} = dependencies;
+  let didUpdateMeeting = false;
+
+  try {
+    await meetingsRepository.updateMeeting(meetingId, mapping.payload);
+    didUpdateMeeting = true;
+
+    if (mapping.removedEmails.length > 0) {
+      await meetingsRepository.removeMeetingInvitation(meetingId, mapping.removedEmails);
```
zskhan: If any of these `removeMeetingInvitation` or `addMeetingInvitation` it will also return `{status: 'updateFailed'};`
This can be misleading and will not be able to show specific error for the above 2 scenarios.

apps/webapp/src/script/components/Meeting/utils/canEditMeeting.test.ts
```diff
@@ -0,0 +1,72 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {Meeting} from 'Components/Meeting/MeetingList/MeetingList';
+import {User} from 'Repositories/entity/User';
+import {translateForTest} from 'Util/test/translateForTest';
+
+import {canEditMeeting} from './canEditMeeting';
+
```
zskhan, line 25: For a bit more readability.
const FUTURE_MEETING = Date.parse('2026-06-15T13:00:00.000Z');
const ONGOING_MEETING = Date.parse('2026-06-15T14:30:00.000Z');
const PAST_MEETING = Date.parse('2026-06-15T16:00:00.000Z');

apps/webapp/src/script/components/Meeting/mapMeetingToScheduleFormState.ts
```diff
@@ -0,0 +1,52 @@
...
+const createPlaceholderUserForEmail = (email: string): User => {
+  const user = new User(`email:${email}`, 'local', translate);
+  user.name(email);
+  user.email(email);
+  return user;
+};
+
+const resolveInvitedEmailsToUsers = (invitedEmails: string[], availableUsers: User[]): User[] =>
+  invitedEmails.map(email => {
+    const normalizedEmail = email.toLowerCase();
```
zskhan:
```suggestion
    const normalizedEmail = email.trim().toLowerCase();
```

apps/webapp/src/script/components/Meeting/mapMeetingToScheduleFormState.ts
```diff
@@ -0,0 +1,52 @@
...
+    const matchedUser = availableUsers.find(
+      user => is.nonEmptyString(user.email()) && user.email()!.toLowerCase() === normalizedEmail,
+    );
```
zskhan:
```suggestion
    const matchedUser = availableUsers.find(user => {
  const userEmail = user.email();
  return is.nonEmptyString(userEmail) && userEmail.toLowerCase() === normalizedEmail;
});
```

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/useScheduleMeetingSubmit.ts
```diff
@@ -19,41 +19,76 @@
...
+  editingMeetingId: Maybe<QualifiedId>,
+  formState: ScheduleMeetingFormState,
+  originalInvitedEmails: string[],
+  dependencies: TryScheduleMeetingDependencies,
+): Promise<MeetingSubmitResult> => {
+  if (mode === 'edit' && editingMeetingId.isJust) {
+    return tryUpdateMeeting(editingMeetingId.value, formState, originalInvitedEmails, dependencies);
+  }
+
+  return tryScheduleMeeting(formState, dependencies);
```
screendriver: I think this should fail closed instead of falling back to creating a meeting.

If `mode === 'edit'` but `editingMeetingId` is missing, this currently creates a new scheduled meeting. That feels risky because an inconsistent edit state could create duplicates.

Can we make the branches explicit?

- create mode -> create meeting
- edit mode + meeting id -> update meeting
- edit mode + missing meeting id -> return an update error / impossible state error

apps/webapp/src/script/components/Meeting/mapScheduleFormToUpdateMeeting.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const recurrence = mapRecurrenceOptionToMeetingRecurrence(formState.recurrence);
+  const {addedEmails, removedEmails} = computeInvitationDiff(originalInvitedEmails, invitedEmails.value);
+
+  return result.ok({
+    payload: {
+      title: formState.title.trim(),
+      start_time: start.toISOString(),
+      end_time: end.toISOString(),
+      ...(recurrence !== undefined && {recurrence}),
```
screendriver: Can we add a test for changing a recurring meeting back to "Does not repeat"?

My concern is this update payload only includes `recurrence` when there is a recurrence value. If the existing meeting is recurring and the user changes it to "Does not repeat", we need to be sure the backend interprets the missing `recurrence` field as "clear recurrence" and not as "leave recurrence unchanged".

If the API needs an explicit value to clear recurrence, this would keep the meeting recurring.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingAction/MeetingAction.tsx
```diff
@@ -19,61 +19,42 @@
...
-        {
-          css: contextMenuDangerItemStyles,
-          icon: () => <TrashIcon css={contextMenuDangerItemIconStyles} />,
-          label: translate('meetings.action.deleteMeetingForAll'),
-        },
-      ],
+      entries: getMeetingActionEntries({
+        meeting,
+        selfUser,
+        nowMs,
```
screendriver, line 56: This `nowMs` comes from render time, but the menu can be opened later.

Example: the list item renders before the meeting starts. Then the meeting starts, but the list does not rerender. If the user opens the menu now, `canEditMeeting()` can still receive the old timestamp and show Edit.

Can we compute the current time when opening the menu, or re-check edit eligibility inside `onEdit` before opening the modal?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -42,18 +48,60 @@ export async function tryScheduleMeeting(
...
+  const {payload, addedEmails, removedEmails} = mappingResult.value;
+  const {meetingsRepository, fetchMeetings} = dependencies;
+  let didUpdateMeeting = false;
+
+  try {
+    await meetingsRepository.updateMeeting(meetingId, payload);
+    didUpdateMeeting = true;
+
+    if (removedEmails.length > 0) {
+      await meetingsRepository.removeMeetingInvitation(meetingId, removedEmails);
```
screendriver: This flow can partially update the meeting.

`updateMeeting()` already succeeded before this line. If removing or adding invitations fails afterwards, we return `updateFailed`, but the meeting title/time/recurrence may already have changed.

That is not necessarily wrong, but the user-facing error becomes misleading. Can we either make this partial-success behavior explicit, or return a more specific error for invitation update failures?

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
thisisamir98: [84568bb1a9c5a54d0c8d44cc1594a0409fac1c34](https://github.com/wireapp/wire-webapp/pull/21629/commits/84568bb1a9c5a54d0c8d44cc1594a0409fac1c34)

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21629/commits/84568bb1a9c5a54d0c8d44cc1594a0409fac1c34

apps/webapp/src/script/components/Meeting/utils/canEditMeeting.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21629/commits/9428681d56f1d04a6877c1f74b2252b75c3e8c3e

apps/webapp/src/script/components/Meeting/mapMeetingToScheduleFormState.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21629/commits/3677516ce07b6a8a2ebfebc44142e08c9c941af8

apps/webapp/src/script/components/Meeting/mapMeetingToScheduleFormState.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21629/commits/723f4ff557f25ca49c2b4a27128112c2b2019b5d

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingAction/MeetingAction.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21629/commits/3b84166b2f15c691e3ede2f5c47b729dae760c4b

apps/webapp/src/script/components/Meeting/mapScheduleFormToUpdateMeeting.ts
thisisamir98: The backend models update recurrence as [Maybe (Maybe Recurrence)](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Meeting.hs#L116-L135) ([wire-server#5065](https://github.com/wireapp/wire-server/pull/5065)):

field omitted → leave unchanged
recurrence: null → clear
recurrence: { ... } → set
So omitting the field on "Does not repeat" would have kept the meeting recurring.

Fixed in [7fd29fe](https://github.com/wireapp/wire-webapp/pull/21629/commits/7fd29fe12fd51cdf9ca867d46eee3673d8466495): we now send recurrence: recurrence ?? null on update, with a test asserting doesNotRepeat maps to recurrence: null.

Maybe from true-myth doesn't help much here, axios needs plain JSON (null, not a Maybe), and the API has three states while Maybe only has two. We'd still convert to null at the wire boundary.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/useScheduleMeetingSubmit.ts
thisisamir98: Fixed in [bae912ce](https://github.com/wireapp/wire-webapp/pull/21629/commits/bae912ce44): performMeetingSubmit now has explicit branches:

create mode → tryScheduleMeeting
edit mode + meeting id → tryUpdateMeeting
edit mode + missing id → editMeetingIdMissing error (no API call)
Added tests for all three paths in scheduleMeetingService.test.ts.

screendriver (approved):

PR #21635, opened 2026-06-24
Skip critical flow status job without run-e2e label [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Summary

Fixes the `precommit-crit-flows` workflow for pull requests without the `run-e2e` label which was introduced in https://github.com/wireapp/wire-webapp/pull/21567

The workflow was intended to run critical flow E2E tests on pull requests only when the `run-e2e` label is set.

## Problem

The `build` and E2E execution path already respect the label guard.

However, the `run_e2e_tests` status job used `always()` without the same pull request label condition. This meant the job still ran for unlabeled pull requests after the upstream E2E job was skipped.

As a result, unlabeled pull requests could fail with `needs.deploy_and_run_e2e.result` being `skipped`.

## Intent

Keep unlabeled pull requests quiet.

They should not run critical flow E2E tests, and they should not fail because the workflow intentionally skipped the upstream jobs.

### DIFF
--- a/.github/workflows/precommit-crit-flows.yml
+++ b/.github/workflows/precommit-crit-flows.yml
@@ -94,7 +94,7 @@ jobs:
 
   run_e2e_tests:
     name: Playwright Critical Flow
-    if: ${{ always() && github.repository == 'wireapp/wire-webapp' }}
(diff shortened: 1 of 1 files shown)


screendriver:
Was already fixed in https://github.com/wireapp/wire-webapp/pull/21631

PR #21640, opened 2026-06-24
Reject past start and end times when scheduling meetings [WPB-25503]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25503" title="WPB-25503" target="_blank">TaskWPB-25503</a>  [Web] Edit a meeting
  </td></table>
  <br />
 

Return structured form errors on submit, validate inline in the modal, and prevent selecting past dates in the date pickers.


apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
```diff
@@ -67,4 +67,28 @@ describe('requireScheduleMeetingTimes', () => {
...
+      start: maybe.just(pastStart),
+    });
+
+    expect(result.isErr).toBe(true);
+    expect(unwrapErr(result)).toBe(scheduleFormErrors.startInPast);
+  });
+
+  it('returns endInPast when end is not in the future', () => {
+    const futureStart = new Date(Date.now() + 3_600_000);
+    const pastEnd = new Date(Date.now() - 60_000);
```
e-maad: Can we use deteministicWallClock here?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
```diff
@@ -27,17 +27,17 @@ import type {ScheduleMeetingFormState} from './scheduleMeetingTypes';
 
 const baseFormState = (): ScheduleMeetingFormState => ({
   title: 'Weekly sync',
-  start: maybe.just(new Date('2026-06-15T10:00:00.000Z')),
-  end: maybe.just(new Date('2026-06-15T11:00:00.000Z')),
+  start: maybe.just(new Date('2026-07-15T10:00:00.000Z')),
```
e-maad: Why we changed the date here? Is it something to be in the future? What about the time after `2026-07-15`, would the tests still pass?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.ts
```diff
@@ -30,5 +30,16 @@ export const requireScheduleMeetingTimes = (
...
 
-  return result.ok({start: formState.start.value, end: formState.end.value});
+  const start = formState.start.value;
+  const end = formState.end.value;
+
+  if (start.getTime() <= Date.now()) {
+    return result.err(scheduleFormErrors.startInPast);
+  }
+
+  if (end.getTime() <= Date.now()) {
```
screendriver: Could we avoid using `Date.now()` here and pass the current time in explicitly?

We already have a WallClock abstraction with `currentTimestampInMilliseconds/currentDate` and a deterministic implementation for tests. This function now has a hidden dependency on real time, which makes the mapping logic harder to test and creates time-based behavior outside the existing clock abstraction.

I would prefer passing a WallClock dependency from the caller.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/ScheduleMeetingForm.tsx
```diff
@@ -119,6 +124,20 @@ export const ScheduleMeetingForm = ({
     [translate],
   );
 
+  const todayValue = useMemo(() => today(getLocalTimeZone()), []);
```
screendriver: This also bypasses the WallClock abstraction.

Can we derive `todayValue` from `wallClock.currentDate` instead of `today(getLocalTimeZone())`? Otherwise the date picker and the validation path can use different notions of "now", especially in tests or when using a deterministic clock.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
```diff
@@ -27,17 +27,17 @@ import type {ScheduleMeetingFormState} from './scheduleMeetingTypes';
...
+  end: maybe.just(new Date('2026-07-15T11:00:00.000Z')),
```
screendriver: This test now depends on an absolute future date. It will start failing after `2026-07-15` because `requireScheduleMeetingTimes` now rejects past start/end values.

Please use the deterministic wall clock / injected timestamp instead of hard-coded "future" dates. Same applies to the schedule service and mapScheduleForm tests that were changed to July 2026.

screendriver (changes requested):

screendriver (changes requested):
The behavior looks generally correct, but please do not use `Date.now()` / `today()` directly in this validation and mapping path. We already have WallClock and DeterministicWallClock. Please pass the current timestamp/current date in explicitly and update the tests to use the deterministic clock instead of absolute future dates.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/ScheduleMeetingForm.tsx
```diff
@@ -65,6 +67,9 @@ const toDateTimePickerValue = (value: Maybe<Date>): Date | null => value.unwrapO
 const fromDateTimePickerValue = (value: Date | null): Maybe<Date> =>
   value === null ? maybe.nothing() : maybe.just(value);
 
+const firstNonEmptyError = (...errorMessages: Array<string | undefined>): string | undefined =>
```
e-maad, line 69: To keep things simple, I would avoid using `...` (rest parameter) because it should only be used when we don't have a way to determine how many arguments we are going to pass. In our case the usage is controlled and can be passed in an array directly.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/ScheduleMeetingForm.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

thisisamir98:
@screendriver thanks for the review, refactored to use wall clock in https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingValidation.test.ts
```diff
@@ -22,34 +22,79 @@ import {maybe} from 'true-myth';
...
 
 describe('scheduleMeetingValidation', () => {
-  const start = maybe.just(new Date('2026-06-15T10:00:00'));
-  const end = maybe.just(new Date('2026-06-15T11:00:00'));
+  const fixedNow = new Date('2026-06-23T14:30:00.000Z');
+  const futureStart = maybe.just(new Date('2026-06-23T16:00:00.000Z'));
+  const futureEnd = maybe.just(new Date('2026-06-23T17:00:00.000Z'));
+
+  beforeEach(() => {
+    jest.useFakeTimers();
```
e-maad, line 30: Is there something from `deterministicWallClock` that we can use instead of `jest.useFakeTimers()`?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/ScheduleMeetingForm.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingValidation.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21640/commits/89941193c196f9f4061a293c1b534009939800b0

e-maad (approved):

PR #21642, opened 2026-06-24
feat: drive sorting part 1(WPB-25179)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25179" title="WPB-25179" target="_blank">TaskWPB-25179</a>  [Web] Sorting of all columns
  </td></table>
  <br />
 

- sorting icons for 3 states asc, desc, no sorting
- For this feature, sorting has three visual/data states overall: unsorted, ascending, and descending. But “unsorted” is only the initial/reset state when entering a view, not a state users cycle back to by repeatedly clicking an active column.
- feat:  Click a header → icon cycles asc/desc, hover reveals the glyph on inactive columns
- feat: make sorting headers accessible


apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTableColumns/CellsTableColumns.tsx
```diff
@@ -36,12 +39,22 @@ const columnHelper = createColumnHelper<CellNode>();
...
   translate: RootContextValue['translate'];
+  getDirectionFor: (field: CellsSortField) => CellsSortDirection | undefined;
+  onToggleSort: (field: CellsSortField) => void;
 }) => [
   columnHelper.accessor('name', {
-    header: translate('cells.tableRow.name'),
+    header: () => (
+      <CellsTableSortableHeader
+        label={translate('cells.tableRow.name')}
+        direction={getDirectionFor('name_ci')}
```
e-maad: What is name 'name_ci', a better name for understanding?

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTable.tsx
```diff
@@ -58,18 +66,22 @@ export const CellsTable = ({nodes, cellsRepository}: CellsTableProps) => {
...
-                    colSpan={header.colSpan}
-                    style={{
-                      width: header.id == 'name' ? undefined : header.getSize(),
-                    }}
-                  >
-                    {header.isPlaceholder ? null : flexRender(header.column.columnDef.header, header.getContext())}
-                  </th>
-                ))}
+                {headerGroup.headers.map(header => {
+                  const sortField = SORTABLE_COLUMN_FIELD[header.column.id];
```
e-maad: Please don't add TS code in the middle of the template. If you still want to add, then move it to a new component.

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTable.tsx
```diff
@@ -81,7 +93,9 @@ export const CellsTable = ({nodes, cellsRepository}: CellsTableProps) => {
                     <td
                       key={cell.id}
                       css={cell.column.id === 'id' ? tableActionsCellStyles : tableCellStyles}
-                      data-cell={cell.column.id === 'id' ? undefined : cell.column.columnDef.header}
+                      data-cell={
+                        typeof cell.column.columnDef.header === 'string' ? cell.column.columnDef.header : undefined
```
e-maad: Why we need a typeof check here? Do we really need it?

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSortIcon/CellsSortIcon.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+export const CellsSortIcon = ({direction}: CellsSortIconProps) => {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="14"
+      height="14"
+      viewBox="0 0 14 14"
+      fill="none"
```
e-maad: I think we don't have to specify `fill="none"`.

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSortIcon/CellsSortIcon.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import {sortIconStyles} from './CellsSortIcon.styles';
+
+export type CellsSortDirection = 'asc' | 'desc';
+
+interface CellsSortIconProps {
+  direction?: CellsSortDirection;
+}
+
+export const CellsSortIcon = ({direction}: CellsSortIconProps) => {
```
e-maad, line 39: I would suggest to have 3 different icons (maybe outside of cells / globally) for each direction. It will be really helpful to use it in other places.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -0,0 +1,82 @@
...
+      if (current?.field !== field) {
+        return {field, direction: DEFAULT_DIRECTION[field]};
+      }
+      // Re-selecting the active column flips the direction. There is no click path
+      // back to the unsorted state — that only returns when the view is left/re-entered.
+      return {field, direction: current.direction === 'asc' ? 'desc' : 'asc'};
+    });
+  }, []);
+
+  const resetSort = useCallback(() => setSort(null), []);
```
e-maad: Do we need to use `useCallback` here?

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTableColumns/CellsTableColumns.tsx
arjita-mitra: its the backend API filed name for sorting. in next PR it becomes just name,  direction={getDirectionFor('name')} so not changing this here.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const getDirectionFor = useCallback(
+    (field: CellsSortField): CellsSortDirection | undefined => (sort?.field === field ? sort.direction : undefined),
+    [sort],
+  );
+
+  return {sort, toggleSort, resetSort, getDirectionFor};
```
screendriver, line 81: I might be missing the intended scope of "part 1", but this currently looks like UI-only sorting state. The hook returns `sort`, but the pull request does not seem to use it to reorder `nodes` or pass it into the Cells search request.

That feels a bit dangerous because the table can claim a sorted state via the icon / `aria-sort`, while the data is still loaded in the existing/default order.

Could we either wire this into the actual data loading/sorting path in this pull request or keep the headers non-interactive until sorting is really implemented?

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTable.tsx
screendriver: This changes more than just avoiding a type issue. For sortable columns, `columnDef.header` is now a function, so `data-cell` becomes `undefined`.

That looks like it breaks the responsive table labels because the mobile layout uses `td[data-cell]::before` with `attr(data-cell)`. It may also break tests/page objects that locate cells via `td[data-cell="Name"]`.

Could we keep an explicit cell label separate from the rendered header? For example via column meta or a small column-label map, instead of deriving `data-cell` from `columnDef.header`.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const toggleSort = useCallback((field: CellsSortField) => {
+    setSort(current => {
+      // Selecting a different column starts at that column's default direction.
+      if (current?.field !== field) {
+        return {field, direction: DEFAULT_DIRECTION[field]};
+      }
+      // Re-selecting the active column flips the direction. There is no click path
+      // back to the unsorted state — that only returns when the view is left/re-entered.
+      return {field, direction: current.direction === 'asc' ? 'desc' : 'asc'};
```
screendriver, line 70: The pull request description mentions three states: ascending, descending, and no sorting. But the click behavior here only cycles between ascending and descending once a column is active.

Which behavior do we want? If "no sorting" is part of the UX, I would expect the cycle to be `none → default direction → opposite direction → none`. Otherwise I would remove “no sorting” from the PR description to avoid confusion.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -0,0 +1,82 @@
+/*
```
screendriver, line 1: Why are there no unit tests for this module?

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: i will be adding the feature flag as well so it won't be usable for users. since no-one using this feature currently i think its okay to leave the implementation not fully wired.

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTable.tsx
arjita-mitra: Resolved 90d192cca511ad157d7cf273a49dac4b44313be9

apps/webapp/src/script/components/CellsGlobalView/CellsTable/CellsTable.tsx
arjita-mitra: Addressed 6f63b611397e5e90a67b75cd23496b3688206936

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: not sure i removed it for now will test it after integration PR

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSortIcon/CellsSortIcon.tsx
arjita-mitra: keeping it changed the icon slightly i feel

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSortIcon/CellsSortIcon.tsx
arjita-mitra: Addressed ac596a15adfd65ab7ae37509f54033bb908a901b

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: Good point.

For this feature, sorting has three visual/data states overall: unsorted, ascending, and descending. But “unsorted” is only the initial/reset state when entering a view, not a state users cycle back to by repeatedly clicking an active column. Edited PR description as well.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: Added some 13e6e26aeef26504a9304212fbc5fea289f8d36d

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: Added bcdd78de093fba232f232a564c0e50c2773ba6fc

screendriver (approved):
Looks good to me now. Thanks for addressing the `data-cell`/mobile label issue and for making the current scope explicit.

I’m approving this as a feature-gated UI foundation for sorting. The important follow-up is still to wire the sort state into the actual data loading/sorting path in the next pull request.

PR #21643, opened 2026-06-24
Wire only meeting participant invites on schedule modal [WPB-25056]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25056" title="WPB-25056" target="_blank">TaskWPB-25056</a>  [Web] Schedule a meeting: participants (Wire users)
  </td></table>
  <br />
 

Use contact search with remote lookup and showAllProvidedUsers, drop edit prefill email placeholders, baseline invitation diffs on resolved emails only, and surface named ParticipantMissingEmailError messages via i18n.


apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -94,66 +95,92 @@ const resultToTask = <T, E>(run: () => Promise<Result<T, E>>): Task<T, E> =>
...
     if (createResult.isErr) {
       return result.err(meetingSubmitErrors.createFailed);
     }
 
+    if (formState.selectedUsers.length > 0) {
+      try {
+        const conversation = await conversationRepository.getConversationById(
+          createResult.value.qualified_conversation,
+        );
+        await conversationRepository.addUsers(conversation, formState.selectedUsers);
```
screendriver: Can we confirm that `conversationRepository.addUsers()` actually rejects for the participant-add failures we want to show as `addParticipantsFailed` here? From the repository implementation it looks like backend errors are caught and handled internally for several labels instead of being rethrown. In that case this `try/catch` would not run and the schedule submit flow could report success even though participants were not added. Maybe this flow needs an explicit result/failed-users signal instead of relying on thrown errors?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -94,66 +95,92 @@ const resultToTask = <T, E>(run: () => Promise<Result<T, E>>): Task<T, E> =>
...
+    if (usersToAdd.length === 0 && userIdsToRemove.length === 0) {
+      await fetchMeetings();
+      return result.ok(undefined);
+    }
+
+    if (qualifiedConversation.isNothing) {
+      return result.err(meetingSubmitErrors.updateFailed);
+    }
+
+    const conversation = await conversationRepository.getConversationById(qualifiedConversation.value);
```
screendriver: Should this be inside a `try/catch` as well? If `getConversationById()` fails here `tryUpdateMeeting()` rejects before returning a `MeetingSubmitErrors` result. `useScheduleMeetingSubmit()` only displays the schedule-meeting error modal when `performMeetingSubmit()` resolves with `result.isErr`, so this path could fail without the translated update error. Since the meeting metadata may already have been updated at this point, we probably also want the same `fetchMeetings()` recovery behavior as in the add/remove branches.

apps/webapp/src/script/components/Meeting/useEditMeeting.ts
```diff
@@ -21,25 +21,23 @@ import {useCallback} from 'react';
...
 
   const editMeeting = useCallback(
-    (meeting: Meeting) => {
-      const userState = container.resolve(UserState);
-      const teamState = container.resolve(TeamState);
-      const availableUsers = getScheduleMeetingParticipantPool(userState, teamState);
-      const formState = mapMeetingToScheduleFormState(meeting, availableUsers);
+    async (meeting: Meeting) => {
+      const conversationRepository = container.resolve(ConversationRepository);
+      const conversation = await conversationRepository.getConversationById(meeting.qualified_conversation);
```
screendriver: Since opening the edit modal now depends on this async fetch, could we add an error path here? If the conversation cannot be loaded, `openEdit()` is never called and the caller discards the promise with `void editMeeting(meeting)` (fire-and-forget invoker?), so clicking _"Edit meeting"_ can silently do nothing. Even a small catch with logging and a user-facing error would make this safer.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingAction/MeetingAction.tsx
```diff
@@ -57,7 +57,7 @@ export const MeetingAction = ({meeting}: MeetingActionProps) => {
         translate,
         onEdit: () => {
           if (canEditMeeting(meeting, selfUser, wallClock.currentTimestampInMilliseconds)) {
-            editMeeting(meeting);
+            void editMeeting(meeting);
```
screendriver: It would be better to use our fire-and-forget invoker here

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export type MeetingStoreState = {
+  meetings: Meeting[];
+  isLoading: boolean;
+  errorKey: MeetingsListErrorKey | undefined;
+  loadMeetings: () => Promise<void>;
+  scheduleMeeting: (formState: ScheduleMeetingFormState) => Promise<Result<MeetingSubmitSuccess, MeetingSubmitErrors>>;
+  updateMeeting: (params: UpdateMeetingParams) => Promise<Result<MeetingSubmitSuccess, MeetingSubmitErrors>>;
+  loadMeetingForEdit: (meeting: Meeting) => Promise<Result<EditMeetingData, MeetingSubmitErrors>>;
```
screendriver: Small consistency nit: now that this flow already uses true-myth `Task`, should we avoid exposing `Promise<Result<...>>` here?

A `Promise<Result<Success, Error>>` is basically the raw async-result shape that `Task<Success, Error>` models, but returning `Task` would keep the abstraction consistent and avoid switching between `Task` and `Promise<Result>` in the same flow.

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+export type MeetingStore = StoreApi<MeetingStoreState>;
+
+type MeetingStoreInitialState = Partial<Pick<MeetingStoreState, 'meetings' | 'isLoading' | 'errorKey'>>;
+
+const refreshMeetingsAfterSubmit = async (
+  loadMeetings: () => Promise<void>,
+  getErrorKey: () => MeetingsListErrorKey | undefined,
+): Promise<Result<void, MeetingSubmitErrors>> => {
```
screendriver: Same here: now that this flow already uses true-myth `Task`, should we avoid exposing `Promise<Result<...>>` here?

A `Promise<Result<Success, Error>>` is basically the raw async-result shape that `Task<Success, Error>` models, but returning `Task` would keep the abstraction consistent and avoid switching between `Task` and `Promise<Result>` in the same flow.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -2649,6 +2738,20 @@ export class ConversationRepository {
...
+   * Add users to a conversation with explicit error handling.
+   *
+   * Returns a `Task` so failures are captured in the error channel instead of surfacing
+   * as an uncaught exception. Prefer to compose async work with `Task`/`Result` and use this
+   * method to handle errors explicitly.
+   */
+  safeAddUsers(conversation: Conversation, userEntities: Pick<User, 'qualifiedId'>[]): Task<void, unknown> {
+    return task.tryOrElse(
+      error => error,
+      () => this.addUsers(conversation, userEntities),
```
screendriver: I think this still does not fully solve the original concern https://github.com/wireapp/wire-webapp/pull/21643#discussion_r3492716730. `safeAddUsers()` wraps `addUsers()`, but `addUsers()` still catches backend errors internally and does not rethrow them. So this `Task` can still resolve successfully even when adding users failed.

For this meeting flow, I think we need an explicit result here, for example `{failedToAdd}` or a dedicated `Task<AddUsersResult, AddUsersError>` instead of wrapping the existing `addUsers()` method.

thisisamir98: Good point, thanks. I refactored `safeAddUsers` so it no longer wraps `addUsers()`. It now calls `addUsersToMLSConversation` directly and returns `Task<{failedToAdd}, ...>` — same pattern as `establishMeetingConversation`. Partial failures are passed through to the meeting submit flow on update as well. Added tests for this.

screendriver: Thanks, this looks much better now 👍 The original `safeAddUsers()` concern is addressed from my perspective.

One small thing I still see: if `updateMeeting()` succeeds but the later participant sync fails, we now return an error before refreshing the meetings list. That means the metadata may already be updated, but the UI may not recover/refresh, which was part of the original concern around the `getConversationById()` failure path.

Can we make sure `loadMeetings()` still runs after a post-update sync failure, similar to the previous recovery behavior?

thisisamir98: Good catch, thanks 👍

I moved the refresh decision to the submit hook. Now `loadMeetings()` runs before we show the error when metadata was already saved but participant sync failed (both create and update).

I also mapped the post-update sync failures to `addParticipantsFailed` / `removeParticipantsFailed` so we can tell them apart from a real API update failure. Added tests for this.

screendriver (approved):

PR #21644, opened 2026-06-24
Open schedule modal from Create meeting and rename action button [WPB-25057]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Wire the primary meetings header button to schedule a meeting, keep Meet Now in the dropdown only, and rename MeetNowMultiActionButton to MeetingMultiActionButton.


apps/webapp/src/script/components/Meeting/MeetingMultiActionButton/MeetingMultiActionButton.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    expect(handleMeetNow).not.toHaveBeenCalled();
+  });
+
+  it('shows only Meet Now in the dropdown menu', () => {
+    const {props, triggerContextMenu} = createTestProps();
+
+    render(withThemeAndRootContext(<MeetingMultiActionButton {...props} />, rootProviderWrapper));
+
+    const buttons = screen.getAllByRole('button');
+    fireEvent.click(buttons[1]);
```
screendriver: Could we avoid selecting the dropdown button by array index here?

`screen.getAllByRole('button')[1]` makes the test depend on render order. It would be more robust if the icon-only dropdown button had an accessible name, for example an `aria-label`, and the test selected it with `screen.getByRole('button', {name: ...})`.

That would improve both the test and the accessibility of the actual button.

apps/webapp/src/script/components/Meeting/MeetingMultiActionButton/MeetingMultiActionButton.tsx
```diff
@@ -24,20 +24,28 @@ import {ButtonGroup, ButtonVariant, CallIcon, TriangleIcon} from '@wireapp/react
...
   const [invertIcon, setInvertIcon] = useState(false);
-  const {handleMeetNow, handleScheduleMeeting} = useMeetingActions();
+  const {handleMeetNow, handleScheduleMeeting} = useMeetingActionsHook();
 
   const handleMeetingOptionButton = (event: MouseEvent<HTMLElement>) => {
     setInvertIcon(val => !val);
-    showContextMenu({
+    triggerContextMenu({
       event,
       anchor: event.target as HTMLElement,
```
screendriver: Small thing while touching this: should this use `event.currentTarget` instead of `event.target`?

`target` can be the nested SVG/icon depending on where the user clicks. `currentTarget` should be the button that owns the click handler, which seems like the more stable anchor for the context menu.

screendriver (changes requested):

screendriver (approved):

zskhan (approved):

PR #21646, opened 2026-06-24
fix: preload MediaPipe WASM resources to reduce BGE initialization latency [WPB-26552]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26552" title="WPB-26552" target="_blank">BugWPB-26552</a>  [Calling] Background Effects: Camera switch causes endless loading and second camera stream is not displayed
  </td></table>
  <br />
 

## Summary

Add prefetch links for MediaPipe WASM resources to warm up the browser cache before background effects are initialized.
This reduces initialization latency and helps prevent BGE startup failures caused by delayed loading of libs an bg images.


apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
```diff
@@ -358,6 +358,30 @@ export class BackgroundEffectsHandler {
...
+      await loadBackgroundSource(DEFAULT_BUILTIN_BACKGROUND_ID);
+
+      const {preferredEffect} = backgroundEffectsStore.getState();
+      if (preferredEffect.type === 'virtual') {
+        await loadBackgroundSource(preferredEffect.backgroundId);
+      }
+    }
+  }
+
+  private prefetch(url: string) {
```
zskhan, line 379: Can we add a check to find existing link element? To avoid duplication of same tag.

```
const existingPrefetchLink = document.head.querySelector<HTMLLinkElement>(
    `link[rel="prefetch"][href="${url}"]`,
  );

if (existingPrefetchLink !== null) {
    return;
  }
```

apps/webapp/src/script/main/app.ts
```diff
@@ -547,6 +548,12 @@ export class App {
       // We load all the users the self user is connected with
       await userRepository.loadUsers(selfUser, connections, conversations, teamMembers);
 
+      try {
```
zskhan: Can we make this non-blocking(use fireAndForget)?
Otherwise slow network will delay app ready state.

zskhan (changes requested):

apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
EnricoSchw: Good point, will add

zskhan (dismissed):

zskhan (approved):

zskhan (approved):

PR #21598, opened 2026-06-19
Add release metadata helpers for ADR 0002 [WPB-26469]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26469" title="WPB-26469" target="_blank">TaskWPB-26469</a>  [Web] Use a trunk-based automated release process
  </td></table>
  <br />
 

## Summary

Adds pure release metadata helpers for ADR 0002.

This includes:

- validation for `release/YYYY-MM-DD.N` branch names
- extraction of the release identifier
- computation of `YYYY-MM-DD.N-beta.M` tags
- computation of `YYYY-MM-DD.N-production`
- detection of existing Production tags
- detection of whether a Production tag points to the current release commit

## Notes

This pull request intentionally does not change release behavior.

No GitHub Actions workflows are changed.

## Tracking

Part of https://github.com/wireapp/wire-webapp/issues/21597

Issue #21597, opened 2026-06-19
Implement ADR 0002: trunk-based automated release process


screendriver:
## Context

ADR 0002 defines the target release process for `wire-webapp`:

- `main` becomes the trunk branch.
- Every merge to `main` deploys to Edge.
- Release branches use `release/YYYY-MM-DD.N`.
- Release branch updates deploy to Beta.
- Successful Beta deployments create `YYYY-MM-DD.N-beta.M` tags.
- Production waits for GitHub Environment approval.
- Production promotes the Beta-tested artifact and does not rebuild from source.
- Successful Production deployments create `YYYY-MM-DD.N-production` tags.
- Production tags are immutable release history.
- Rollback is handled by a dedicated GitHub Actions workflow.
- On-premises maintenance releases are handled separately from the normal cloud release flow.

ADR:

`docs/decision-records/0002-trunk-based-automated-release-process.md`

## Goal

Implement ADR 0002 in small, reviewable, low-risk steps.

The final release process should be GitHub-driven, auditable, and safe:

- build once
- deploy to Beta
- validate on Beta
- approve Production through GitHub Environment approval
- promote the same artifact to Production
- create the Production tag only after successful Production deployment
- keep rollback explicit and auditable

## Non-goals for the first phase

The first phase should not try to do everything at once.

Out of scope initially:

- Deleting `dev`.
- Deleting or renaming `master`.
- Removing all old release triggers immediately.
- Full on-premises maintenance release automation.
- Perfect durable artifact storage for rollback.
- Full feature flag governance automation.
- Large workflow rewrites without intermediate compatibility.

## Progress checklist

- [x] Inspect the current release implementation.
- [x] Add release metadata helpers.
- [x] Prepare reusable E2E / Testiny support.
- [x] Add Edge deployment from trunk.
- [x] Add release branch to Beta workflow.
- [x] Add Production approval and artifact promotion.
- [x] Add release branch creation workflow.
- [x] Add first-class Production rollback workflow.
- [x] Cut over branch model from `dev` / `master` to `main`.
- [x] Retire old tag-driven release path.
- [x] Add maintenance release automation separately.

## Phase 1: Inspect the current release implementation

Create an implementation inventory before changing code.

The inventory should cover:

- existing workflows for build, deploy, release, E2E, rollback, Docker, Helm, and notifications
- existing scripts for release tagging and changelog generation
- current branch assumptions
- current tag assumptions
- current deployment targets and environment names
- Testiny integration
[...]

PR #21664, opened 2026-06-26
Add Production promotion to release-cloud workflow [WPB-26469]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26469" title="WPB-26469" target="_blank">TaskWPB-26469</a>  [Web] Use a trunk-based automated release process
  </td></table>
  <br />
 

## Summary

Extends the release-cloud workflow with optional Production promotion for ADR 0002.

The workflow can now promote the Beta-tested artifact to Production after successful Beta deployment and E2E validation.

## Details

This extends `.github/workflows/release-cloud.yml`.

The workflow now:

- keeps release-cloud manual-only during the transition
- adds an explicit `promote_to_production` input
- runs Production preflight after successful Beta E2E
- verifies that the Beta tag points to the selected release commit
- computes the `YYYY-MM-DD.N-production` tag
- skips Production deployment when promotion was not requested
- skips Production deployment when the matching Production tag already points to the selected release commit
- fails when the Production tag exists for a different commit
- waits for `wire-webapp-prod` GitHub Environment approval
- deploys the same artifact that was deployed to Beta
- creates the Production tag only after successful Production deployment

## Notes

This pull request intentionally does not rebuild from source for Production.

Production promotion is opt-in during the transition so Beta-only workflow runs do not request Production approval.

This pull request also does not change:

- the old tag-driven release path
- rollback behavior
- maintenance release behavior
- Docker publishing
- Helm publishing
- `wire-builds` updates
- branch renaming

The Production approval gate is enforced through the `wire-webapp-prod` GitHub Environment settings.

## Tracking

Part of https://github.com/wireapp/wire-webapp/issues/21597.

Follow-up to:

- https://github.com/wireapp/wire-webapp/pull/21598
- https://github.com/wireapp/wire-webapp/pull/21634
- https://github.com/wireapp/wire-webapp/pull/21641
- https://github.com/wireapp/wire-webapp/pull/21648

### DIFF
--- a/.github/workflows/release-cloud.yml
+++ b/.github/workflows/release-cloud.yml
@@ -370,8 +370,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [build_artifact, production_preflight]
     if: ${{ needs.production_preflight.outputs.should_deploy == 'true' }}
-    permissions:
-      contents: write
+    permissions: {}
     # Production approval is enforced by the wire-webapp-prod GitHub Environment settings.
     environment: wire-webapp-prod
     concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-cloud.yml
```diff
@@ -238,10 +244,197 @@ jobs:
...
+            echo "- Approval gate: ${PRODUCTION_ENVIRONMENT_NAME} GitHub Environment settings"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+  deploy_to_production:
+    name: Deploy to Production
+    runs-on: ubuntu-24.04
+    needs: [build_artifact, production_preflight]
+    if: ${{ needs.production_preflight.outputs.should_deploy == 'true' }}
+    permissions:
+      contents: write
```
zskhan: Is write permission in this job for tag creation only?
We already have `create_beta_tag`, can we re-use it and also remove the write permission from this job

screendriver: Yes, the write permission is only needed for the Production tag creation.

The production deployment job should not need `contents: write`. I will split this into a dedicated `create_production_tag` job after `deploy_to_production`, mirroring the `create_beta_tag` pattern, and reduce `deploy_to_production` back to `permissions: {}`.

I would not reuse `create_beta_tag` directly because the tag name and preflight semantics are different, but the permission boundary should be the same ➡️ a3c180b4886af736f6a9e8ae3879e101b966148e

zskhan (approved):

PR #21670, opened 2026-06-26
fix(webapp): map workspace libraries to source for IDE type resolution [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

Workspace @wireapp/* packages expose types from compiled lib/ output that is not present on a fresh clone until Nx builds dependencies. Webpack and CI still work because they build libs first, but the TypeScript language server and ESLint import/no-unresolved read package.json types paths and fail with unresolved module errors.

Add tsconfig paths so the editor and tsc resolve these imports from libraries/*/src directly, matching the existing @wireapp/core pattern in tsconfig.eslint.json. This does not change webpack bundling, which still resolves through node_modules and compiled lib/.


screendriver (dismissed):
Only small non-blocking thought: we now duplicate the same workspace package mappings in multiple tsconfigs, so we should keep an eye on drift when new workspace packages are added.

screendriver (dismissed):

screendriver (approved):

PR #21671, opened 2026-06-26
feat(meetings): allow hosts to delete meetings for everyone [WPB-25513]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25513" title="WPB-25513" target="_blank">TaskWPB-25513</a>  [Web] Delete a meeting
  </td></table>
  <br />
 

Wire the existing delete meeting API through the repository layer and show a confirmation modal before removing a meeting. Only meeting hosts see the action, and the list refreshes after a successful deletion.


apps/webapp/src/script/components/Meeting/deleteMeetingService.ts
```diff
@@ -0,0 +1,49 @@
...
+  dependencies,
+}: TryDeleteMeetingParams): Promise<Result<void, DeleteMeetingErrorCode>> {
+  const {meetingsRepository, fetchMeetings} = dependencies;
+  const deleteResult = await meetingsRepository.deleteMeeting(meetingId);
+
+  if (deleteResult.isErr) {
+    return result.err(deleteMeetingErrors.deleteFailed);
+  }
+
+  await fetchMeetings();
```
screendriver, line 47: `tryDeleteMeeting` returns a `Result`, but `fetchMeetings()` can still reject here. That means a successful delete followed by a failed refresh escapes the `Result` flow and the modal cannot show the mapped error state.

Can we make this behavior explicit? Either make the refresh best-effort, add a separate refresh error, or catch the refresh failure here and return a `Result`. A test for a rejecting `fetchMeetings` would also help.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingListItem.tsx
```diff
@@ -113,7 +114,7 @@ const MeetingListItemComponent = ({nowMs, ...meeting}: MeetingListItemProps) =>
       </div>
       <div css={rightStyles}>
         <MeetingStatus start_date={start_date} end_date={end_date} attending={attending} nowMs={timestamp} />
-        <MeetingAction meeting={meeting} />
+        <MeetingAction meeting={meeting} onMeetingDeleted={onMeetingDeleted ?? (async () => {})} />
```
screendriver, line 117: Can we avoid the silent no-op fallback here?

After deleting a meeting, refreshing the list is part of the expected success flow. If `onMeetingDeleted` is not wired, the delete action can still run but the UI silently skips the refresh. I would prefer making this callback required through the component chain, or not exposing the delete action when the callback is missing.

Additionally we would have [noop-esm](https://github.com/screendriver/noop-esm) in place for situations like that.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingAction/getMeetingActionEntries.tsx
```diff
@@ -63,16 +72,6 @@ export const getMeetingActionEntries = ({
...
-      css: contextMenuDangerItemStyles,
-      icon: () => <CloseIcon css={contextMenuDangerItemIconStyles} />,
-      label: translate('meetings.action.deleteMeetingForMe'),
-    },
-    {
-      css: contextMenuDangerItemStyles,
-      icon: () => <TrashIcon css={contextMenuDangerItemIconStyles} />,
-      label: translate('meetings.action.deleteMeetingForAll'),
-    },
+    ...(canEditMeeting(meeting, selfUser, nowMs) ? [editEntry, deleteForAllEntry] : []),
```
screendriver, line 75: Is delete-for-everyone intentionally using exactly the same eligibility rule as edit?

If yes, I would still prefer a more neutral predicate name, for example `canManageUpcomingMeeting`, or a dedicated `canDeleteMeetingForAll`. Reusing `canEditMeeting` for deletion makes the delete rule look accidental and couples future edit/delete behavior.

apps/webapp/src/script/components/Meeting/utils/isMeetingHost.ts
```diff
@@ -0,0 +1,25 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {Meeting} from 'Components/Meeting/MeetingList/MeetingList';
+import type {User} from 'Repositories/entity/User';
+import {matchQualifiedIds} from 'Util/qualifiedId';
+
+export const isMeetingHost = (meeting: Meeting, selfUser: User): boolean =>
```
screendriver, line 24: Nit: can we avoid the immediate-return arrow function and use a block body with an explicit `return` here?

```ts
export function isMeetingHost(meeting: Meeting, selfUser: User): boolean {
  return matchQualifiedIds(meeting.qualified_creator, selfUser.qualifiedId);
};

File: apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingAction/getMeetingActionEntries.test.tsx [hunk 5]
```diff
@@ -42,55 +42,88 @@ const createSelfUser = (id = 'host-id') => {
 
 const translate = (key: string) => key;
 
-const getEditEntryLabel = (entries: ReturnType<typeof getMeetingActionEntries>) =>
-  entries.find(entry => entry.label === 'meetings.action.editMeeting');
+const getEntryByLabel = (entries: ReturnType<typeof getMeetingActionEntries>, label: string) =>
```
screendriver, line 45: Nit: same here. Can we use an explicit `return` statement for this helper?

```ts
function getEntryByLabel(entries: ReturnType<typeof getMeetingActionEntries>, label: string) {
  return entries.find(entry => entry.label === label);
};

File: apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItemGroup.tsx [hunk 6]
```diff
@@ -77,7 +79,12 @@ const MeetingListItemGroupComponent = ({
         <div>
           {nonEmptyGroups.flatMap(([, items]) =>
             items.map(item => (
```
screendriver, line 81: Nit: this nested `flatMap` / `map` callback now renders multi-line JSX. Can we use block bodies with explicit `return` statements here? It is more verbose, but easier to read and debug.

screendriver (changes requested):

PR #21678, opened 2026-06-27
Fix WebSocket reconnect watchdog timing and guest link E2E flow [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Summary

This fixes the failing precommit Playwright checks by addressing two separate issues:

* defer the WebSocket CONNECTING watchdog until `onReconnect()` has resolved
* keep reconnect-in-place behavior for active sockets so we do not create duplicate `/await` sessions
* only start the watchdog when the current socket is still CONNECTING
* update the password-protected guest link E2E flow to click the existing-user Join button after login before expecting the password modal

## Why

`onReconnect()` can wait for access-token refresh before returning the WebSocket URL. Starting the CONNECTING watchdog before that means the watchdog can measure token-refresh time instead of actual WebSocket connection time.

The fix keeps the existing reconnect-in-place behavior intact while making the watchdog timing more precise.

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
@@ -1,4 +1,4 @@
-import {Page} from 'playwright/test';
+import {Locator, Page} from 'playwright/test';
 import {getUser, User} from 'test/e2e_tests/data/user';
 import {PageManager} from 'test/e2e_tests/pageManager';
 import {test, expect, withLogin, Team, LOGIN_TIMEOUT, withGuestUser} from 'test/e2e_tests/test.fixtures';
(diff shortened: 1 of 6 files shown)


screendriver:
The pull request related Playwright failures are fixed now:

- AccountSettings / TC-60 passed on retry
- Guestroom / TC-8140 passed
- Guestroom / TC-3477 passed

The remaining Precommit E2E failures are unrelated to the files changed in this pull request:

- Calling -> I want to see the full call participant list
- Federation -> I want to import my backup of conversations with users from different BE and see the conversation contents
- Federation -> I want to start a conference call in a federated group with Users from 2 different backends

I do not think we should continue changing this pull request to chase unrelated full-suite failures.

apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
```diff
@@ -13,13 +13,120 @@ import {createGroup, sendConnectionRequest} from 'test/e2e_tests/utils/userActio
...
+    }
+
+    await page.waitForTimeout(guestroomEnterpriseLoginPollInterval);
+  }
+
+  throw new Error(`Guestroom enterprise login did not reach the sidebar within ${guestroomEnterpriseLoginTimeout}ms`);
+}
+
+function waitForFirstVisibleLocator(locators: Locator[], timeout: number): Promise<Locator | undefined> {
+  return Promise.race(
```
zskhan: Promise.any is more appropriate here because we want to wait until all of them are undefined or at least one is fulfilled.
What do you think?

apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
```diff
@@ -13,13 +13,120 @@ import {createGroup, sendConnectionRequest} from 'test/e2e_tests/utils/userActio
...
+  const currentUrl = new URL(page.url());
+
+  currentUrl.protocol = testEnvironmentUrl.protocol;
+  currentUrl.hostname = testEnvironmentUrl.hostname;
+  currentUrl.port = testEnvironmentUrl.port;
+
+  await page.goto(currentUrl.toString());
+}
+
+function isLocatorVisible(locator: Locator): Promise<boolean> {
```
zskhan: Duplicate methods from accountSettings.spec.ts?
Can we extract them to a util file?

zskhan (changes requested):

apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
screendriver: Yes, fair point.

`isLocatorVisible` and `waitForVisible` are generic enough to move out of the spec files. I’ll extract those into a small E2E utility and keep the more scenario-specific post-login flows local to the specs for now: 46a504168aacd554a268032123dceecfae55974d

apps/webapp/test/e2e_tests/specs/Guestroom/guestroom.spec.ts
screendriver: Good point. I agree for `waitForFirstVisibleLocator`.

`Promise.race` is not ideal here because one locator can time out first and resolve to `undefined`, even if another locator becomes visible afterwards. `Promise.any` expresses the intent better: return the first locator that actually becomes visible, and return `undefined` only if none of them becomes visible.

I’ll update this: 0a66ae80c943aacd6dc786cb68db60d6b4be0b4f

zskhan (approved):

PR #21693, opened 2026-06-29
Revert "fix: optimize renderer for WebGL through ANGLE on Direct3D [WPB-25918]"

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25918" title="WPB-25918" target="_blank">StoryWPB-25918</a>  [Web] Add more Direct3D Support for Chromium on Windows for Background Effect Rendering
  </td></table>
  <br />
 

Reverts wireapp/wire-webapp#21633

This is reverted for now so that it will not be part of next release.
This feature will go in the release.


e-maad (approved):

screendriver (approved):

screendriver (approved):
Would be great to have a description why it should be reverted

EnricoSchw (approved):

PR #21696, opened 2026-06-30
Remove WebSocket connectivity diagnostics banner action [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Problem

We temporarily added a WebSocket connectivity diagnostics action to the reconnect banner while investigating sleep/wake recovery issues.

The action was hidden behind a startup feature flag and was useful for dogfooding/debugging, but it is still debug UI in a user-facing warning banner.

The current WebSocket recovery implementation works well enough now, so we should remove this temporary diagnostics feature again.

## Why

The reconnect banner should stay focused on the user-facing connection state.

The diagnostics action was intentionally temporary and should not remain in the product now that the WebSocket recovery path is stable enough.


e-maad (approved):
LGTM 🚀

PR #21703, opened 2026-06-30
feat: enhance detached window handling for external links and modals [WPB-26355]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26355" title="WPB-26355" target="_blank">BugWPB-26355</a>  [Desktop] Support Article link for camera is not clickable in maximized window on Mac
  </td></table>
  <br />
 

### Before
- In Electron, links with target="_blank" opened from the detached about:blank window were not handled correctly and could fail to open.
- When user interacts with video button on main window while detached calling was active, No Camera Access modal is crashed and will not be displayed.

### After
- _blank links are delegated to the opener window using openerWindow.open(...), ensuring they follow the same handling as links opened from the main application window.
- Modals are rendered in whichever window currently has focus (main window or detached window)

### DIFF
--- a/apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
+++ b/apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
@@ -56,20 +56,20 @@ import {CallingViewMode, CallState, MuteState} from './CallState';
 import {CALL_MESSAGE_TYPE} from './enum/CallMessageType';
 import {LEAVE_CALL_REASON} from './enum/LeaveCallReason';
 import {Participant} from './Participant';
-import type {ActiveWindowState} from 'Hooks/useActiveWindow';
 import {useActiveWindowState} from 'Hooks/useActiveWindow';
 
 import {buildMediaDevicesHandler, createConversation, createSelfParticipant} from '../../auth/util/test/testUtil';
 import {Core} from '../../service/coreSingleton';
 import {Warnings} from '../../view_model/WarningsContainer';
 import {z} from 'zod';
-import {translateForTest, translateWithPrefixForTest} from 'Util/test/translateForTest';
+import {translateForTest} from 'Util/test/translateForTest';
 import {MessageRepository} from 'Repositories/conversation/MessageRepository';
 import {MediaStreamHandler} from 'Repositories/media/MediaStreamHandler';
 import {MediaDevicesHandler} from 'Repositories/media/MediaDevicesHandler';
 import {BackgroundEffectsHandler} from 'Repositories/media/backgroundEffectsHandler';
 import {APIClient} from '../../service/apiClientSingleton';
 import {ConversationState} from 'Repositories/conversation/ConversationState';
+import {Translate} from 'Util/localizerUtil';
 
 type AudioFlowStat = {
   bytesReceived?: number;
@@ -126,6 +126,9 @@ function createCallingRepositoryForTest({
   };
 }
 
+const translateWithPrefixForTest: Translate = (translationKey, _substitutions, _dangerousSubstitutions, _skipEscape) =>
+  `translated:${translationKey}`;
+
 describe('CallingRepository', () => {
   const testFactory = new TestFactory();
   let callingRepository: CallingRepository;
@@ -149,7 +152,7 @@ describe('CallingRepository', () => {
     callingRepository['callState'].calls([]);
     callingRepository['conversationState'].conversations([]);
     callingRepository.destroy();
-    jest.clearAllMocks();
+    jest.restoreAllMocks();
   });
 
   afterAll(() => {
@@ -178,8 +181,8 @@ describe('CallingRepository', () => {
 
       callingRepository['conversationState'].conversations.push(conversation);
       callingRepository['callState'].calls([call]);
-      spyOn(callingRepository, 'muteCall').and.callThrough();
-      spyOn(wCall, 'recvMsg').and.callThrough();
+      jest.spyOn(callingRepository, 'muteCall');
+      jest.spyOn(wCall, 'recvMsg');
 
       const event: CallingEvent = {
         content: {
@@ -224,8 +227,8 @@ describe('CallingRepository', () => {
 
       callingRepository['conversationState'].conversations.push(conversation);
       callingRepository['callState'].calls([call]);
-      spyOn(callingRepository, 'muteCall').and.callThrough();
-      spyOn(wCall, 'recvMsg').and.callThrough();
+      jest.spyOn(callingRepository, 'muteCall');
+      jest.spyOn(wCall, 'recvMsg');
 
       const event: CallingEvent = {
         content: {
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
```diff
@@ -406,6 +407,77 @@ describe('CallingRepository', () => {
...
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        translate,
+      );
+      const showModal = jest.spyOn(PrimaryModal, 'show').mockImplementation(() => undefined as never);
+      const activeWindowSpy = jest
+        .spyOn(useActiveWindowState, 'getState')
+        .mockReturnValue({activeWindow: window} as any);
```
screendriver: Please try to avoid `any` at all cost 🙏

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
```diff
@@ -406,6 +407,77 @@ describe('CallingRepository', () => {
...
+    it('renders in the detached window when the detached window is focused', () => {
+      const translate = jest.fn((translationKey: string) => `translated:${translationKey}`);
+      const isolatedCallingRepository = new CallingRepository(
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
```
screendriver: Same here: please no `any`

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -163,7 +168,6 @@ export const setupDetachedWindowExternalLinksClick = (detachedWindow: Window, op
   };
 
   detachedWindow.document.addEventListener('click', handleClick, true);
-  return () => detachedWindow.document.removeEventListener('click', handleClick, true);
```
screendriver, line 168: Could we keep the cleanup return here?

This function registers an event listener on another window's document. Even if the current caller does not use the cleanup yet, I think the function should still make the lifecycle explicit and keep the previous cleanup test. For the non-desktop branch we could return a noop cleanup function.

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
screendriver: Could we restore `PrimaryModal.show` as well?

`jest.clearAllMocks()` only clears calls; it does not restore the original implementation. Since this is in the large `CallingRepository` suite, leaving `PrimaryModal.show` mocked can leak into later tests. Either `showModal.mockRestore()` or `jest.restoreAllMocks()` in the relevant afterEach would make the test isolation clearer.

screendriver (changes requested):

apps/webapp/src/script/repositories/calling/CallingRepository.ts
zskhan: This is only used for the detached window. Once the detached window is closed and the call returns to the main window, the entire React context is unmounted, so the listener is cleaned up as part of the window teardown. There's no scenario where this listener outlives the detached window, so we don't need an explicit cleanup function here.

screendriver: That makes sense for the current detached-window lifecycle.

My concern is less about an immediate leak in the current flow and more about keeping the lifecycle ownership explicit. This function registers an event listener on a `document`, so I would still expect it to return the matching cleanup function, independent of the caller currently relying on the detached window teardown.

That keeps the helper easier to reason about, preserves the previous cleanup behavior/test and avoids future changes accidentally depending on this exact window lifetime. For the non-desktop branch we could simply return a noop cleanup function.

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
screendriver: Conversation was resolved without changing anything? 🤔

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
```diff
@@ -406,6 +408,79 @@ describe('CallingRepository', () => {
...
+    it('renders in the main window when the main window is focused', () => {
+      const translate = jest.fn((translationKey: string) => `translated:${translationKey}`);
+      const isolatedCallingRepository = new CallingRepository(
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
+        {} as any,
```
screendriver: Same here: please no `any`

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
screendriver: Resolved without changing it? I think the `PrimaryModal.show` mock should still be restored explicitly. The suite-level `afterEach` currently uses `jest.clearAllMocks()`, which clears calls but does not restore mocked implementations.

Could we either use `jest.restoreAllMocks()` in the relevant `afterEach`, or make sure every `PrimaryModal.show` spy is restored reliably?

apps/webapp/src/script/util/test/translateForTest.ts
```diff
@@ -22,3 +22,5 @@ import type {Translate} from 'Util/localizerUtil';
 export function translateForTest(translationKey: Parameters<Translate>[0]): string {
   return translationKey;
 }
+
+export const translateWithPrefixForTest: Translate = translationKey => `translated:${translationKey}`;
```
screendriver: This helper looks a bit misleading to me. It is typed as `Translate`, but it only uses the translation key and silently ignores substitutions, dangerous substitutions and `skipEscape`.

That is probably fine for these assertions, but then I would rather not expose it as a full `Translate`. Maybe we can keep the helper intentionally narrow, similar to `translateForTest`, or implement the full signature if we want to use it as a real test translation function.

apps/webapp/src/script/repositories/calling/CallingRepository.test.ts
```diff
@@ -406,6 +483,61 @@ describe('CallingRepository', () => {
     });
   });
 
+  describe('showNoCameraModal', () => {
+    it('renders in the main window when the main window is focused', () => {
+      const translate = jest.fn(translateWithPrefixForTest);
+      const {repository: isolatedCallingRepository} = createCallingRepositoryForTest({translate});
+      const showModal = jest.spyOn(PrimaryModal, 'show').mockImplementation(() => undefined as never);
+      const activeWindowSpy = jest
+        .spyOn(useActiveWindowState, 'getState')
```
screendriver: Minor: this test verifies the current implementation by mocking `useActiveWindowState.getState()` directly. Would it be possible to test the behavior through the active-window/focus mechanism instead? Otherwise the test is very tightly coupled to the Zustand store access inside `CallingRepository`.

apps/webapp/src/script/util/test/translateForTest.ts
zskhan: Moved this helper into the CallingRepository.test.ts where it is getting used.

thisisamir98 (approved):

screendriver (approved):

PR #21707, opened 2026-07-01
feat: drive sorting integration(WPB-25179)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25179" title="WPB-25179" target="_blank">TaskWPB-25179</a>  [Web] Sorting of all columns
  </td></table>
  <br />
 

### Sorting of all columns

**Sort persistence —**

Sort selected in browse view → carry into search view of same conversation? Vice versa?

Sort selected in Conversation A's browse → carry to Conversation B?

Sort selected in Shared Drive → carry to global Drive (and back)?

based on our last discussions with @Olga Anna Skoczylas we reset in all 3

**Rule of thumb:** user-applied sort always wins. Default is mtime/desc whenever content is being actively searched/filtered. Default is BE natural (folders-first) when the view is in a passive/browse state (browse mode, or search open but nothing entered).

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
@@ -92,7 +92,8 @@ export const ConversationCells = memo(
     const isCellsStateReady = cellsState === CONVERSATION_CELLS_STATE.READY;
     const isCellsStatePending = cellsState === CONVERSATION_CELLS_STATE.PENDING;
 
-    const {sort, setSort, getDirectionFor, toggleSort} = useCellsSorting();
+    const sortScopeKey = `${conversationId}:${isSearchViewOpen ? 'search' : 'browse'}`;
+    const {sort, setSort, getDirectionFor, toggleSort} = useCellsSorting(sortScopeKey);
 
     const {refresh, setOffset} = useGetAllCellsNodes({
       cellsRepository,
@@ -148,12 +149,6 @@ export const ConversationCells = memo(
       }
       wasSearchViewOpen.current = isSearchViewOpen;
     }, [clearAll, clearAllFilters, clearSearch, conversationId, isSearchViewOpen]);
-
-    // Sort is per-view: switching conversation or toggling between search and browse
-    // returns to the default (unsorted) order.
-    useEffect(() => {
-      setSort(null);
-    }, [conversationId, isSearchViewOpen, setSort]);
 
     // Navigating into a folder or the recycle bin happens via the URL hash without
     // remounting; reset the sort on those transitions too.
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
@@ -92,6 +92,8 @@ export const ConversationCells = memo(
     const isCellsStateReady = cellsState === CONVERSATION_CELLS_STATE.READY;
     const isCellsStatePending = cellsState === CONVERSATION_CELLS_STATE.PENDING;
 
+    const {sort, getDirectionFor, toggleSort, resetSort} = useCellsSorting();
+
     const {refresh, setOffset} = useGetAllCellsNodes({
       cellsRepository,
       conversationQualifiedId,
@@ -100,6 +102,7 @@ export const ConversationCells = memo(
       enabled: isCellsStateReady && !isSearchViewOpen,
       fireAndForgetInvoker,
       userRepository,
+      sort,
     });
 
     const {filters, filterState, clearAllFilters} = useConversationDriveFilters({
@@ -107,8 +110,6 @@ export const ConversationCells = memo(
       conversationRepository,
       translate,
     });
-
-    const {getDirectionFor, toggleSort} = useCellsSorting();
 
     const {
       searchValue,
@@ -125,6 +126,7 @@ export const ConversationCells = memo(
       userRepository,
       filters: filterState,
       onClear: refresh,
+      sort,
     });
 
     // Search view open ⇒ load-more UI + search-hook data; closed ⇒ page-nav UI + browse-hook data.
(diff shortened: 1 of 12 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
```diff
@@ -147,6 +149,20 @@ export const ConversationCells = memo(
       wasSearchViewOpen.current = isSearchViewOpen;
     }, [clearAll, clearAllFilters, clearSearch, conversationId, isSearchViewOpen]);
 
+    // Sort is per-view: switching conversation or toggling between search and browse
+    // returns to the default (unsorted) order.
+    useEffect(() => {
+      setSort(null);
```
screendriver: Could we avoid resetting the sort in an effect after the data hooks already received the previous `sort` value?

On a conversation switch or search/browse transition, the hooks above can already schedule a request with the old sort before this `setSort(null)` runs. Then the reset triggers another request with the default sort. I think the first request in a new sorting scope should already use the default sort.

Maybe the effective sort should be derived before passing it into `useGetAllCellsNodes` / `useConversationSearchFiles`, or the sorting state should be keyed by scope.

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/useGetAllCellsNodes.ts
```diff
@@ -121,7 +126,13 @@ export const useGetAllCellsNodes = ({
     }
     // cellsRepository and userRepository are not dependencies because they're singletons
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [domain, id, isCurrentFetchRequest, offset, pageSize, setError, setNodes, setPagination, setStatus]);
+  }, [domain, id, isCurrentFetchRequest, offset, pageSize, sort, setError, setNodes, setPagination, setStatus]);
+
+  // A new sort criterion re-pages from the top; the fetch effect below then refetches
+  // because fetchNodes depends on both offset and sort.
+  useEffect(() => {
+    setOffset(0);
```
screendriver: I think this can issue an unnecessary request when the user changes sorting while being on a later page.

`fetchNodes` depends on `sort`, so the fetch effect can run once with the new sort but the old `offset`. After that, this effect resets the offset to `0`, which triggers another request. Could we make the sort transition reset pagination before fetching, or add a test for changing sort while `offset > 0`?

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -31,15 +31,15 @@ export interface CellsSort {
...
+  name: 'asc',
   mtime: 'desc',
   size: 'asc',
 };
 
 // Maps a table column id (the accessor key) to its backend sort field. Columns absent
 // from this map are not sortable and receive no aria-sort.
 export const SORTABLE_COLUMN_FIELD: Record<string, CellsSortField> = {
-  name: 'name_ci',
+  name: 'name',
```
screendriver, line 42: Can we verify that `name` is the backend sort field we want here?

The previous value was `name_ci`, which reads like an intentional case-insensitive name sort. Since this value is passed through to the backend as `SortField`, changing it to `name` could silently change the user-facing order if the backend distinguishes between `name` and `name_ci`.

If the backend contract changed, I think it would be good to make that explicit in the type or with a short comment.

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
```diff
@@ -76,7 +76,5 @@ export const useCellsSorting = () => {
     [sort],
   );
 
-  const resetSort = () => setSort(null);
-
-  return {sort, toggleSort, resetSort, getDirectionFor};
+  return {sort, setSort, toggleSort, getDirectionFor};
```
screendriver, line 105: I would prefer keeping a domain-specific `resetSort` callback instead of exposing raw `setSort` from `useCellsSorting`. Consumers only reset it to `null`, so exposing the setter leaks the hook’s internal representation.

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: it didn't change. name_ci, is a case-insensitive name sort so i was using it but ios and android client confirmed there is some issue with name_ci and it doesn't work on the backend so i changed it to name.

apps/webapp/src/script/components/Conversation/ConversationCells/ConversationCells.tsx
arjita-mitra: that could be a good improvement. made it scope based here 8f30b9ef6969c30afa2453da3866242e63414cb0

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/useGetAllCellsNodes.ts
arjita-mitra: good catch addressed here c41fc8211babad115492983d3eed41658f63d367

apps/webapp/src/script/components/Conversation/ConversationCells/common/useCellsSorting/useCellsSorting.ts
arjita-mitra: I removed `resetSort` earlier because the callback was not memoized, and `ConversationCells` used it inside an effect. That made the effect rerun on unrelated renders and could resubscribe/retrigger work unnecessarily.

But I agree with the API point: exposing `setSort(null)` leaks the hook’s internal representation. I can bring it back as a memorised callback to keep it stable.

screendriver (approved):

PR #21716, opened 2026-07-02
Deduplicate group conversations in search results [WPB-24651]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24651" title="WPB-24651" target="_blank">BugWPB-24651</a>  Duplicate group conversation appears in search results when guest name matches part of the group name
  </td></table>
  <br />
 

## Summary

This fixes a conversation search issue where the same group conversation could appear twice when:

- the group conversation name matched the search query
- and one of the group participants / guests also matched the same query

The search result building now keeps conversation-name matches as the primary result and excludes those conversations from the participant-name result list.

Participant-name search still works for groups where only the participant name matches.

### DIFF
--- a/apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.test.tsx
+++ b/apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.test.tsx
@@ -18,6 +18,7 @@
  */
 
 import {CONVERSATION_TYPE} from '@wireapp/api-client/lib/conversation';
+import {createFactory} from '@enormora/objectory';
 
 import {Conversation} from 'Repositories/entity/Conversation';
 import {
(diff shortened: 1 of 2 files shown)


thisisamir98 (dismissed):

apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.tsx
```diff
@@ -97,14 +97,27 @@ export function getTabConversations({
...
+    const filteredConversations = conversations.filter(conversationSearchFilter);
+    const filteredConversationIds = new Set(
+      filteredConversations.map(conversation => {
+        const {domain, id} = conversation.qualifiedId;
+        return `${domain ?? ''}/${id}`;
+      }),
+    );
     const filteredGroupConversations = groupConversations.filter(group => {
+      const {domain, id} = group.qualifiedId;
+      const groupConversationId = `${domain ?? ''}/${id}`;
```
arjita-mitra: we could create getQualifiedConversationId and reuse it. 

```
const getQualifiedConversationId = ({qualifiedId}: {qualifiedId: {domain?: string | null; id: string}}) =>
  `${qualifiedId.domain ?? ''}/${qualifiedId.id}`;
```

apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.test.tsx
```diff
@@ -284,9 +284,60 @@ describe('getTabConversations', () => {
...
+      groupConversations: [florianSupportConversation],
+      directConversations: [],
+      favoriteConversations: [],
+      archivedConversations: [],
+      conversationsFilter: 'Florian',
+      channelAndGroupConversations: [florianSupportConversation],
+      channelConversations: [],
+      isChannelsEnabled: false,
+      draftConversations: [],
+      searchInputPlaceholders,
```
e-maad: Can we create a function (data object factory) to generate this object?

apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.tsx
```diff
@@ -97,14 +97,27 @@ export function getTabConversations({
     }
 
     const filterWord = replaceAccents(conversationsFilter.toLowerCase());
+    const filteredConversations = conversations.filter(conversationSearchFilter);
+    const filteredConversationIds = new Set(
+      filteredConversations.map(conversation => {
+        const {domain, id} = conversation.qualifiedId;
+        return `${domain ?? ''}/${id}`;
```
e-maad: Do we need `/` when the domain is not there?

```
return domain ? `${domain}/${id}` : id;
```

screendriver: Good point. No, the `/` itself is not meaningful when the domain is missing.

I would avoid introducing a local string format here and use the existing qualified-id comparison instead. That keeps the deduplication logic aligned with the rest of the codebase.

apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.tsx
screendriver: Agree with the intention. I will address this together with the other qualified-id comment.

Instead of adding a new local stringifier, I’ll use the existing qualified-id comparison helper so we don't have to define another string representation here.

apps/webapp/src/script/page/leftSidebar/panels/conversations/helpers.test.tsx
screendriver: I will extract a small test input builder so the tests focus on the scenario-specific data instead of repeating the full `getTabConversations` input object.

zskhan (approved):

PR #21732, opened 2026-07-03
Parallelize continuous integration checks for faster pull request feedback [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

This pull request restructures the continuous integration workflow so independent validation checks can run in parallel instead of being executed sequentially inside one large `test` job.

The goal is not to reduce coverage or skip validation. Pull requests should still fail for the same reasons as before, but developers should get feedback earlier and see more clearly which part of CI failed. This improves CI feedback and observability. It does not try to fully solve the remaining workspace build-output coupling in lint; making lint independent from generated `lib` output should be handled separately.

During this refactoring we also uncovered a hidden workspace/Nx assumption: some lint targets are not fully self-contained and depend on generated library output from a previous build step. In the old sequential CI job this was hidden because libraries were built before linting in the same workspace.

With lint running in an isolated job, this coupling became visible. For now, the lint job explicitly builds workspace libraries before running lint to preserve the existing behavior. Making lint fully independent from generated `lib` output should be handled in a separate cleanup.

---

2026-07-03 15 10 00 github com 30cc6b75d87b


zskhan (approved):
LGTM. We need to update the branch protection rules once this merged so that instead of test(from ci.yml) it will look for new rules(from continous-integration.yml).

PR #21746, opened 2026-07-06
Replace single-instance cookie with local storage [WPB-26744]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26744" title="WPB-26744" target="_blank">BugWPB-26744</a>  Strange configuration of the app_opened cookie
  </td></table>
  <br />
 

The web app used the `app_opened` cookie to coordinate whether another browser tab was already running the application. This value is purely client-side state. It is only used for single-instance detection and does not need to be sent to the server with every request.

Using a cookie for this caused the browser to report `app_opened` as a non-secure cookie. Adding cookie flags would only treat the symptom. The better fix is to stop using a cookie for local browser coordination entirely.

This change replaces the cookie-based implementation with browser storage and introduces a small abstraction for storage access.

The abstraction is intentional: it keeps direct access to browser APIs out of the hook, makes the single-instance logic easier to test, and gives us one controlled place for reading, writing, removing, and validating stored values. It also makes malformed storage data explicit instead of letting invalid JSON crash the hook.

As a result, the `app_opened` marker is no longer an HTTP cookie, is no longer sent with requests, and no longer appears as an insecure cookie in browser tooling. The existing behavior stays the same: the app still registers the current instance, detects another running instance, and clears the marker when the current instance is closed.


screendriver:
All end-to-end tests are green. GitHub has issues with test reports artifacts: `Directory does not exist`

apps/webapp/src/script/browser/storage/createStringKeyValueStorageFromWebStorage.ts
```diff
@@ -0,0 +1,69 @@
...
+import {Maybe, result} from 'true-myth';
+
+import {StringKeyValueStorage} from 'src/script/storage/StringKeyValueStorage';
+
+export function createStringKeyValueStorageFromWebStorage(webStorage: Maybe<Storage>): StringKeyValueStorage {
+  return {
+    getItem: key => {
+      return webStorage.andThen(storage => {
+        const storedValue = result.tryOrElse(
+          error => {
```
zskhan, line 29: Can we add error logging here?
Otherwise the consumer will be responsible for logging.

zskhan (approved):

apps/webapp/src/script/browser/storage/createStringKeyValueStorageFromWebStorage.ts
screendriver: This is intentional.

I would avoid logging here because this helper is a low-level leaf around `Storage`. It does not know whether a failed read/write is critical, expected, recoverable, or irrelevant for the caller.

Logging from here would make this leaf decide operational severity without enough context and could create noisy or duplicated logs. The owning/root flow should decide whether this failure matters and log it with the relevant product/user context.

If we want this to be observable, I would rather expose the failure to the caller, for example via a `Result`/diagnostic, and log at the boundary instead of injecting logging into this leaf.

PR #21747, opened 2026-07-06
ci: make Prettier lint failures obvious in CI [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary
- Split the lint CI job into separate ESLint, Prettier, and Stylelint steps so format failures show up as their own failed check instead of a bare exit code 1 buried in ESLint output
- Add `bin/lint-format-check.sh` to wrap `nx format:check` with clear log output and GitHub error annotations
- Remove the vague `lint:other` script and inline `lint:format` + `stylelint` into `lint` and `lint:affected`

## Test plan
- [ ] CI lint job shows three distinct steps on a PR
- [ ] A deliberately unformatted file fails the "Check formatting (Prettier)" step with file paths and fix commands in the log
- [ ] `yarn lint:format` locally prints the same clear failure output


bin/lint-format-check.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
screendriver, line 1: One small non-blocking thought: this script currently treats every output line from `nx format:check` as a file path. That is probably fine for the expected Prettier output, but if Nx prints a warning, config error, daemon message, or any other non-file line, we would create misleading GitHub annotations.

Maybe we should either filter this to existing files or only create per-file annotations when the line looks like a path. The raw output can still be printed unchanged.

screendriver (changes requested):

bin/lint-format-check.sh
thisisamir98: Good point, thanks. I changed it so we only create annotations when the line is an existing file. The raw nx output in the log stays unchanged.

e-maad (approved):

PR #21767, opened 2026-07-07
chore: WPB-18035 Rename translation keys to not use keyword Services but to use Apps instead.

ikotarac:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-18035" title="WPB-18035" target="_blank">TaskWPB-18035</a>  [Web] [Integrations] Rename Services to Apps
  </td></table>
  <br />
 

Rename translation keys to not use keyword Services but to use Apps instead.

The strategy to test and deploy this change is as follows:
- new translation keys will upload to Crowdin after merge to dev
- use Translation Memory (Crowdin feature) to link previous translations to new keys
- pull from Crowdin by manually dispatching github workflow 'Pull translations from Crowdin'
- test other languages in branch created by github action & merge


apps/webapp/src/i18n/en-US.json
```diff
@@ -221,21 +221,29 @@
   "acme.settingsChanged.headline.alt": "End-to-end identity certificate",
   "acme.settingsChanged.headline.main": "Team settings changed",
   "acme.settingsChanged.paragraph": "As of today, your team uses end-to-end identity to make Wire's usage more secure and practicable. The device verification takes place automatically using a certificate and replaces the previous manual process. This way, you communicate with the highest security standard. <br/> <br/> Enter your identity provider's credentials in the next step to automatically get a verification certificate for this device. <br/> <br/> <a href=\"{url}\" target=\"_blank\">Learn more about end-to-end identity</a>",
+  "addParticipantsAppConfirmButton": "Add app",
```
zskhan: There are few translation keys which are added/updated but they are not in use.

ikotarac: I just renamed service -> app, but didn't try to do the cleanup of the keys that are not in use.

I can try to do that for these, but maybe it's better to do a full search. For example, use Claude to find other strings that are orphaned and delete everything in follow-up PR.

apps/webapp/src/script/repositories/conversation/ConversationStateHandler.ts
```diff
@@ -82,13 +82,9 @@ export class ConversationStateHandler extends AbstractConversationEventHandler {
             const {featureName, ...featureInfo} = featureFromStateChange(prevAccessState, accessState);
 
             if (featureInfo.isAvailable) {
-              messageString = this.translate(
-                `modalConversationOptionsAllow${featureName as 'Guest' | 'Service'}Message`,
-              );
+              messageString = this.translate(`modalConversationOptionsAllow${featureName as 'Guest' | 'App'}Message`);
```
screendriver: Changing the type assertion from `Service` to `App` does not change the runtime value. `featureFromStateChange()` still returns "Service", so this constructs `modalConversationOptionsAllowServiceMessage`, but that translation key is removed by this pull request.

The same problem exists for the disable lookup on line 87 and the toggle lookup on line 96.

Could we map the backend/domain feature name to the UI translation key explicitly instead of hiding the mismatch with a type assertion? I would also add focused tests for failed allow, disable, and toggle operations so that the generated translation keys are verified.

apps/webapp/src/i18n/en-US.json
screendriver: I agree that a general cleanup of all historical orphaned translation keys can happen separately. However, the keys being touched by this pull request should not be renamed if they are already unused.

At least these three appear to be unused:

* `addParticipantsAppConfirmButton` on line 224
* `appsRoomToggleName` on line 246
* `searchAppPlaceholder` on line 1889

Renaming them creates unnecessary Crowdin work and preserves dead configuration. Could we remove these three as part of this change instead of introducing renamed replacements?

screendriver (changes requested):
The described rollout seems to make `main` depend on a manual follow-up:

1. Merge the application changes.
2. Upload the new keys to Crowdin.
3. Manually pull the translations.
4. Test and merge another branch.

Does this mean `main` temporarily contains usages of the new keys without the corresponding translations for the supported languages? Since `main` should remain independently releasable, I would prefer a two-phase migration: first add and synchronize the new keys, then switch the usages and remove the old keys once all translations are available.

The description also refers to merging into `dev`, although this pull request targets `main`. Could the actual rollout be clarified?

ikotarac:
> The described rollout seems to make `main` depend on a manual follow-up:
> 
> 1. Merge the application changes.
> 2. Upload the new keys to Crowdin.
> 3. Manually pull the translations.
> 4. Test and merge another branch.
> 
> Does this mean `main` temporarily contains usages of the new keys without the corresponding translations for the supported languages? Since `main` should remain independently releasable, I would prefer a two-phase migration: first add and synchronize the new keys, then switch the usages and remove the old keys once all translations are available.
> 
> The description also refers to merging into `dev`, although this pull request targets `main`. Could the actual rollout be clarified?

I think this is what changed with removal of *dev* branch, although there was still similar risk before. My plan was to merge this, wait for github action to automatically upload keys, wait for Astrid to translate the keys, wait for another scheduled github workflow to download the translations in follow-up PR, and only then we would have dev in a state that it could be merged into main.

I agree that two phase is safer. I could do it by creating a new branch only with addition of keys to en-US.json, without removal of the old keys, then merging it and waiting for translations, and only when translations are merged in, doing to clean-up follow up.

Alternatively, which is less changes to the work that has already been done, I could manually upload keys using a script, and then pull the translations into this PR. However, there is a risk that some currently used keys would get deleted through manual push, so I guess I have to supply a version of en-US.json file containing both old and new keys in any case.

ikotarac:
> > The described rollout seems to make `main` depend on a manual follow-up:
> > 
> > 1. Merge the application changes.
> > 2. Upload the new keys to Crowdin.
> > 3. Manually pull the translations.
> > 4. Test and merge another branch.
> > 
> > Does this mean `main` temporarily contains usages of the new keys without the corresponding translations for the supported languages? Since `main` should remain independently releasable, I would prefer a two-phase migration: first add and synchronize the new keys, then switch the usages and remove the old keys once all translations are available.
> > The description also refers to merging into `dev`, although this pull request targets `main`. Could the actual rollout be clarified?
> 
> I think this is what changed with removal of _dev_ branch, although there was still similar risk before. My plan was to merge this, wait for github action to automatically upload keys, wait for Astrid to translate the keys, wait for another scheduled github workflow to download the translations in follow-up PR, and only then we would have dev in a state that it could be merged into main.
> 
> I agree that two phase is safer. I could do it by creating a new branch only with addition of keys to en-US.json, without removal of the old keys, then merging it and waiting for translations, and only when translations are merged in, doing to clean-up follow up.
> 
> Alternatively, which is less changes to the work that has already been done, I could manually upload keys using a script, and then pull the translations into this PR. However, there is a risk that some currently used keys would get deleted through manual push, so I guess I have to supply a version of en-US.json file containing both old and new keys in any case.

I have added a new PR which contains only addition of new keys, as part of 2 phase rename. However, I think we need to discuss how do we handle translations in general from now on. If main should always be releasable, that means we can't rely on pushing and pulling translations from Crowdin on push to main.

Does this mean that every time we need to add new translation keys, we need to make a PR that only adds new keys, merge, then wait for translations to be finished, and then merge main into our working branch before creating a PR? This doesn't sound realistic. Alternatively, we could manually push and pull translations as part of work on the feature, and only when translations are present then we open a PR, but that means that features containing any localized strings will take 1-2 days more to finish. I think we need to discuss about this process @screendriver

apps/webapp/src/script/repositories/conversation/ConversationStateHandler.ts
ikotarac: this is a very good catch. I am currently fixing this

ikotarac: I changed the implementation here to use map instead of string interpolation. However, behavior is slightly different now. Previous typing allowed **featureName** to be undefined and modal would be possibly opened with broken message string. This possibility is only based on typing and I am not saying that this was happening in runtime.

New implementation performs check `featureName !== undefined` before resolving anything, so if there was a possibility of showing modal with broken message, now there is a possibility of not showing modal at all in that case. I think this would be nice to inspect and refactor, but I think it it outside of scope of this PR.

zskhan (approved):

screendriver (approved):

PR #21771, opened 2026-07-08
Validate meeting API responses with Zod in api-client [WPB-25056]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25056" title="WPB-25056" target="_blank">TaskWPB-25056</a>  [Web] Schedule a meeting: participants (Wire users)
  </td></table>
  <br />
 

## Summary
- Add Zod schemas for meeting and meetings list API responses in `api-client`.
- Parse backend payloads in `MeetingsAPI` and throw `ZodError` when responses are invalid.
- Export the schemas and cover validation behavior with unit tests.

## Test plan
- [x] `yarn nx run api-client-lib:test --testFile=src/meetings/meeting.schema.test.ts`
- [x] `yarn nx run api-client-lib:test --testFile=src/meetings/meetingsApi.test.ts`
- [x] `yarn type-check`
- [x] `BASE_REF=dev yarn lint:affected`


arjita-mitra (dismissed):

libraries/api-client/src/meetings/meetingSchema.ts
```diff
@@ -0,0 +1,59 @@
...
+    qualified_conversation: qualifiedIdSchema,
+    qualified_creator: qualifiedIdSchema,
+    qualified_id: qualifiedIdSchema,
+    recurrence: meetingRecurrenceSchema.optional(),
+    start_time: z.string(),
+    title: z.string().min(1).max(256),
+    trial: z.boolean(),
+    updated_at: z.string(),
+  })
+  .passthrough();
```
screendriver: Do we want to preserve unknown backend fields on purpose here?

libraries/api-client/src/meetings/meetingSchema.ts
```diff
@@ -0,0 +1,59 @@
...
+const meetingRecurrenceSchema = z.object({
+  frequency: z.nativeEnum(MeetingRecurrenceFrequency),
+  interval: z.number().int().positive().optional(),
+  until: z.string().optional(),
+});
+
+export const meetingSchema = z
+  .object({
+    created_at: z.string(),
+    end_time: z.string(),
```
screendriver: Since this schema is now our runtime boundary for meeting API responses, should we validate the date fields as `datetimes` instead of plain strings?

Right now `created_at`, `updated_at`, `start_time`, `end_time`, and `recurrence.until` accept any string. That means values like `"foo"` still pass validation. You could use `z.string().datetime()` for ISO date fields (unless the backend intentionally allows non-ISO values)

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it('accepts optional recurrence and unknown backend fields', () => {
+    const result = meetingSchema.safeParse({
+      ...validMeeting,
+      recurrence: {frequency: MeetingRecurrenceFrequency.WEEKLY, interval: 2},
+      invited_emails: ['guest@example.com'],
+    });
+
+    expect(result.success).toBe(true);
```
screendriver, line 47: Maybe we should make the test assert the parsed data, not only `result.success` so this behavior is locked down explicitly:

```suggestion
    expect(result.success).toBe(true);
    expect(result.data.invited_emails).toEqual(['guest@example.com']);
```

screendriver (changes requested):

libraries/api-client/src/meetings/meetingSchema.ts
thisisamir98: this is for the sake of forward compatibility so the app would not break in case backend adds a new property in future versions

libraries/api-client/src/meetings/meetingSchema.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21771/commits/725a4d8eb89d587de3729647fcf9e5e03eaa3927

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21771/commits/725a4d8eb89d587de3729647fcf9e5e03eaa3927

libraries/api-client/src/meetings/meetingSchema.ts
screendriver: I think there is an important distinction here:

Default `z.object(...)` behavior in Zod 3 is not strict. Unknown properties do not make parsing fail, but they are stripped from the parsed result.

That is different from `.passthrough()`, which preserves unknown backend fields and exposes them to the rest of the application.

And `.strict()` would be different again, because it would reject unknown fields.

So for forward compatibility, I think the default strip behavior might already be enough: the app would not break when the backend adds a new property, but we also would not leak unvalidated backend data past this API boundary.

I would only use `.passthrough()` here if we intentionally want consumers of `meetingSchema` to access unknown backend fields.

thisisamir98: makes sense https://github.com/wireapp/wire-webapp/pull/21771/commits/788291dce43ce32dde33a75096fb3b80d5b8d969

screendriver (approved):

e-maad (approved):

PR #21776, opened 2026-07-08
fix: update the HistoryImport and HistoryExport modal re-render issue [WPB-22420]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Add tests to ensure stable modal behavior

Remove eslint-disable-next-line which was introduced in [this PR](https://github.com/wireapp/wire-webapp/pull/21730)


thisisamir98 (approved):
Looks good to me.

The stable modal ids make sense for stopping duplicate password modals, and putting the effect deps back is the right move now that each show call has a fixed id. The tests cover the main cases I was worried about — rerender should not open another modal or start another import with the same file.

I did not find any critical behavior problems in this change.

PR #21777, opened 2026-07-08
feat(meetings): meeting series occurrence data layer and virtualized list [WPB-25057]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary

- Introduce `MeetingSeries` as the Zustand source model (fetch once, no backend pagination) and pure selectors that expand recurrence into `MeetingInstance` rows grouped by day.
- Wire the meeting list to a 14-day occurrence window with edit flow anchored on the series (`mapSeriesToScheduleFormState`, `canEditMeeting` on series start date).
- Virtualize day-group sections with TanStack Virtual, expand the window by 14 days on scroll near the bottom, and use per-day height estimates (75px rows) for stable first-paint alignment.

## Architecture

```
Backend → mapApiMeetingToSeries → Zustand meetingSeries[]
→ getVisibleTimeWindow(dayCount) → getMeetingInstances → groupMeetingInstancesByDay
→ MeetingList (virtualized day groups) → MeetingListItem (MeetingInstance rows)
→ edit whole series via mapSeriesToScheduleFormState(series anchor)
```

## Commits

1. Store `MeetingSeries` as source model
2. Instance selectors (expand, group, visible window)
3. List UI wired to selectors (14-day window, no hour grouping)
4. Edit via `MeetingInstance`/`MeetingSeries` anchor
5. TanStack Virtual day groups + scroll-based load-more

## Test plan

- [x] Open Meetings tab — list loads without crash, day sections align on first paint
- [x] Recurring meetings with past series anchor show upcoming occurrences
- [x] Scroll to bottom — additional days load when more occurrences exist
- [x] Edit a meeting — form pre-fills from series; save updates the whole series
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/MeetingList/meetingList.test.tsx`
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/selectors/getMeetingInstances.test.ts`
- [x] `yarn type-check` and `BASE_REF=dev yarn lint:affected`


apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -48,42 +58,84 @@ export interface Meeting {
...
 
-  const {today, tomorrow} = getTodayTomorrowLabels();
-  const headerForToday = `${translate('meetings.list.today')} (${today})`;
-  const headerForTomorrow = `${translate('meetings.list.tomorrow')} (${tomorrow})`;
+  const now = useMemo(() => new Date(nowMs), [nowMs]);
+
+  const meetingInstancesByDay = useMemo(() => {
+    const {from, to} = getVisibleTimeWindow(now, {dayCount: visibleDayCount});
+    const meetingInstances = getMeetingInstances(meetingSeries, from, to).filter(
+      meetingInstance => meetingInstance.end.getTime() >= nowMs,
```
screendriver: Can we avoid coupling recurrence expansion to the 1-second clock tick?

`nowMs` changes every second, so this memo re-runs every second and expands/groups all meeting instances again. Since `getMeetingInstancesInRange` walks from the series anchor to the visible window, a long-running daily recurring series can become unnecessarily expensive.

I would split this into two concerns:
- the visible window / occurrence expansion, based on start-of-day + visibleDayCount
- the current timestamp, used only for status/action rendering

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
```diff
@@ -48,42 +58,84 @@ export interface Meeting {
...
+
+export const MeetingList = ({
+  meetingSeries,
+  isLoading,
+  hasLoadError,
+  scrollElementRef,
+  useMeetingDayGroupVirtualizer: useMeetingDayGroupVirtualizerDependency = useMeetingDayGroupVirtualizer,
+}: MeetingListProps) => {
   const {translate, wallClock} = useApplicationContext();
   const [nowMs, setNowMs] = useState(() => wallClock.currentTimestampInMilliseconds);
```
screendriver:
```suggestion
  const [nowMilliseconds, setNowMilliseconds] = useState(() => wallClock.currentTimestampInMilliseconds);
```

apps/webapp/src/script/components/Meeting/utils/canEditMeeting.ts
```diff
@@ -17,15 +17,16 @@
...
+export const isMeetingHost = (meetingSeries: MeetingSeries, selfUser: User): boolean =>
+  matchQualifiedIds(meetingSeries.qualified_creator, selfUser.qualifiedId);
 
-export const canEditMeeting = (meeting: Meeting, selfUser: User, nowMs: number): boolean => {
-  const hasNotStarted = nowMs < new Date(meeting.start_date).getTime();
+export const canEditMeeting = (meetingInstance: MeetingInstance, selfUser: User, nowMs: number): boolean => {
+  const instanceHasNotStarted = nowMs < meetingInstance.start.getTime();
 
-  return isMeetingHost(meeting, selfUser) && hasNotStarted;
+  return isMeetingHost(meetingInstance.meetingSeries, selfUser) && instanceHasNotStarted;
```
screendriver, line 37: Can we make the edit invariant explicit here?

Right now editing is allowed based on the selected `MeetingInstance`, but the edit form is populated from the original `MeetingSeries` anchor. For a recurring series whose anchor is already in the past, this means the edit action can be visible for a future occurrence, while the modal opens with past start/end dates and validation rejects them.

I think we should choose one clear rule:
- either editing is only allowed while the series anchor has not started yet
- or the form should use the selected instance start/end while keeping the series id for the update

apps/webapp/src/script/components/Meeting/mapApiMeetingToSeries.ts
```diff
@@ -19,16 +19,23 @@
...
-  title: apiMeeting.title,
-  recurrence: mapMeetingRecurrenceToOption(apiMeeting.recurrence),
-  qualified_id: apiMeeting.qualified_id,
-  qualified_creator: apiMeeting.qualified_creator,
-});
+export const mapApiMeetingToSeries = (apiMeeting: ApiMeeting): MeetingSeries => {
+  const durationMs = Date.parse(apiMeeting.end_time) - Date.parse(apiMeeting.start_time);
+  const recurrenceUntil = apiMeeting.recurrence?.until;
+
+  return {
```
screendriver: Can we validate the API dates at this boundary instead of storing a potentially invalid duration?

`Date.parse` can produce `NaN`, and then the store contains a `MeetingSeries` that looks valid but later creates invalid instances or silently disappears from the list.

This mapper is the boundary between the backend model and our UI/domain model. I would prefer an explicit `Result`/validation step here, or at least a guard that logs and drops invalid meetings. We should also enforce `end_time > start_time`. And we should use `date-fns` whenever possible for date related topics.

apps/webapp/src/script/components/Meeting/selectors/getMeetingInstancesInRange.ts
```diff
@@ -0,0 +1,120 @@
...
+const daysPerBiweeklyPeriod = 14;
+
+const createMeetingInstance = (meetingSeries: MeetingSeries, start: Date): MeetingInstance => ({
+  meetingSeries,
+  start,
+  end: new Date(start.getTime() + meetingSeries.duration_ms),
+});
+
+const isMeetingInstanceInRange = (meetingInstance: MeetingInstance, from: Date, to: Date): boolean =>
+  meetingInstance.end.getTime() >= from.getTime() &&
```
screendriver: Is the intended semantics "starts inside the window" or "overlaps the window"?

The `end >= from` check suggests overlapping meetings should be included, but `start >= from` excludes every instance that started before the window. If we care about ongoing long-running meetings, the predicate should probably be based on overlap, for example `end > from && start < to`.

If starts-inside-window is intentional, I would remove the redundant `end >= from` check or rename this to make the contract explicit.

apps/webapp/src/script/components/Meeting/selectors/getMeetingInstancesInRange.ts
```diff
@@ -0,0 +1,120 @@
...
+    case 'doesNotRepeat':
+      return start;
+    case 'daily':
+      return addDays(start, 1);
+    case 'weekly':
+      return addDays(start, daysPerWeek);
+    case 'everyTwoWeeks':
+      return addDays(start, daysPerBiweeklyPeriod);
+    case 'monthly':
+      return addMonths(start, 1);
```
screendriver, line 52: Can we add tests that define the expected monthly recurrence behavior?

The implementation relies on repeated `addMonths(current, 1)`, but monthly recurrence semantics can be tricky around end-of-month and DST boundaries. I do not know what the backend contract is here, so I think this selector should document it with tests before we depend on it in the list.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
screendriver: Can we remove the old `Meeting` interface from this component?

`MeetingList` now receives `MeetingSeries[]` and renders `MeetingInstance` rows. Keeping the old API-shaped `Meeting` type in the UI component makes the boundary less clear and looks like leftover code.

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21777/commits/149bbfea2180052f3fc70be3fa77d35a39a7bcb2

apps/webapp/src/script/components/Meeting/selectors/getMeetingInstancesInRange.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21777/commits/fde6d5738b452570670b0a2238cebf9d3f0b1604

apps/webapp/src/script/components/Meeting/mapApiMeetingToSeries.ts
thisisamir98: https://github.com/wireapp/wire-webapp/pull/21777/commits/ce4b51d2b0ffad2ce947df6080bb34f38f6892a5

apps/webapp/src/script/components/Meeting/utils/canEditMeeting.ts
thisisamir98: Went with **option B**: edit is per-instance, and the form now prefills from the selected occurrence's start/end while the series `qualified_id` is still used for the update.

See 98cb9f93f6 (https://github.com/wireapp/wire-webapp/commit/98cb9f93f6dea33e5d65313c39ebcc8fbf6f65fc):
- Replaced `mapSeriesToScheduleFormState` with `mapMeetingInstanceToScheduleFormState`
- Documented the invariant on `canEditMeeting` with a link to the mapper
- Added tests for recurring series with a past anchor

This keeps the edit action visible for upcoming occurrences without opening the modal with past anchor dates that fail validation.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
thisisamir98: Done — renamed `nowMs` / `setNowMs` to `nowMilliseconds` / `setNowMilliseconds` across the meetings feature (list, status, edit eligibility, and related tests).

See 162c8ebaca: https://github.com/wireapp/wire-webapp/commit/162c8ebaca

apps/webapp/src/script/components/Meeting/MeetingList/MeetingList.tsx
thisisamir98: Split as suggested:

- **`expandedMeetingInstancesByDay`** — recurrence expansion + day grouping from `visibleDayStart` (start of calendar day) and `visibleDayCount`; only recomputes when the day changes or the window widens
- **`meetingInstancesByDay`** — lightweight filter that drops ended instances on the 1s `nowMilliseconds` tick

`nowMilliseconds` is still passed down for status/actions and section headers. Load-more now uses `visibleDayStart` as well.

See 59d209f236: https://github.com/wireapp/wire-webapp/commit/59d209f236

apps/webapp/src/script/components/Meeting/utils/canEditMeeting.ts
thisisamir98: Follow-up on option B — we hit a regression.

**Bug:** Editing a recurring meeting from a *later* future instance prefilled the form with that row's start/end. The update still targets the series `qualified_id`, so saving sent that later date as the new anchor. All earlier occurrences disappeared, even though the modal showed the correct recurrence and the user may not have changed the time at all.

**Fix:** Keep edit eligibility per list row, but prefill recurring series from the **upcoming anchor** — the first instance on or after now — via `getUpcomingMeetingInstanceStart`. Non-repeating meetings still use the series anchor.

Example: weekly series anchored Jun 1, user edits the Jun 29 row on Jun 10 → form now opens with Jun 15 (upcoming), not Jun 29, so saving no longer jumps the anchor forward and wipes Jun 15/Jun 22.

See d0cb5bae47.

screendriver (approved):

PR #21787, opened 2026-07-09
fix(webapp): keep video blob URLs alive during playback [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary

- Fixes inline video playback in non-cells conversations failing shortly after play starts with repeated `GET blob:... net::ERR_FILE_NOT_FOUND` errors.
- `VideoAsset` now revokes blob URLs only on unmount or when replacing a preview/source URL, instead of on every re-render.
- Memoizes `getAssetUrl` in `useAssetTransfer` and adds regression tests for playback re-renders, upload progress re-renders, and unmount cleanup.

## Root cause

Legacy `VideoAsset` loads video files via `URL.createObjectURL()` and stores the resulting blob URL in component state.

A `useEffect` cleanup disposed both preview and playback blob URLs whenever the effect re-ran. The effect depended on `videoPreview`, `videoSrc`, and `getAssetUrl`, so any re-render could trigger cleanup and revoke the active playback URL.

During playback, `onTimeUpdate` updates the remaining-time display on every tick, which re-renders the component. Upload progress updates cause the same behavior while a file is sending. Once the blob URL was revoked, the `<video>` element could still play briefly from buffered data, then fail with `ERR_FILE_NOT_FOUND` while trying to read more of the file.

Larger or longer videos (for example screen recordings) fail more visibly because they need continued reads from the blob URL. Smaller clips can appear unaffected if enough data was buffered before revocation.

This only affects the legacy asset path used when cells is disabled. Cells conversations use presigned HTTP URLs via `MultipartAssets`.

## Fix

- Split blob URL lifecycle into separate effects: one for loading/replacing the preview URL, and one for unmount cleanup via refs.
- Dispose previous URLs only when replacing them or when the component unmounts.
- Wrap `getAssetUrl` (and related callbacks) in `useCallback` to avoid unnecessary effect re-runs.
- Dispose unplayable video URLs when playback validation fails.

## Test plan

- [x] `yarn nx run webapp:test --testFile=src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx`
- [x] `yarn nx run webapp:test --testFile=src/script/components/MessagesList/Message/ContentMessage/asset/common/useAssetTransfer/useAssetTransfer.test.ts`
- [x] Upload a large `.mp4` in a non-cells conversation, play it inline, and confirm playback continues without blob URL errors in the console
- [x] Confirm preview and playback blob URLs are still cleaned up when leaving the conversation


apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+import {
+  createRootContextValueForTest,
+  createRootProviderWrapperForTest,
+} from 'src/script/page/testSupport/rootContextTestSupport';
+import {translateForTest} from 'Util/test/translateForTest';
+
+import {VideoAsset} from './VideoAsset';
+import {AssetUrl, useAssetTransfer} from '../common/useAssetTransfer/useAssetTransfer';
+
+jest.mock('../common/useAssetTransfer/useAssetTransfer');
```
thisisamir98: @screendriver 
in order to avoid this I need to do a larger refactor

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.tsx
```diff
@@ -140,11 +166,15 @@ const VideoAsset = ({
             // ToDo: This needs to be revisited
             // amplify.publish(WebAppEvents.ANALYTICS.EVENT, EventName.MESSAGES.VIDEO.PLAY_FAILED);
             // amplify.publish(WebAppEvents.ANALYTICS.EVENT, EventName.MESSAGES.VIDEO.UNPLAYABLE_ERROR);
+            assetUrl.dispose();
             setVideoPlaybackError(true);
             return;
           }
 
-          setVideoSrc(assetUrl);
+          setVideoSrc(previous => {
```
screendriver: Could we guard this asynchronous path against the component being unmounted?

The unmount cleanup only knows about URLs that have already been committed to state and copied into the refs. If the user leaves the conversation after `getAssetUrl()` creates the object URL but before this state update commits, React ignores the update and the URL is never revoked.

A mounted/request token should be checked after the asynchronous operations. When the request has been cancelled, `assetUrl.dispose()` should be called instead of updating the component. A regression test with a deferred `getAssetUrl()` promise and an unmount before resolution would cover this case.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.tsx
```diff
@@ -77,16 +77,42 @@ const VideoAsset = ({
...
 
+    let cancelled = false;
+
+    void getAssetUrl(assetPreviewResource).then(url => {
+      if (cancelled) {
+        url.dispose();
+        return;
+      }
+
+      setVideoPreview(previous => {
```
screendriver: Could we let effects own the lifetime of these URLs instead of disposing them inside state updater functions?

State updaters should remain pure. Revoking an object URL here can happen while React is still calculating the next state, potentially before or without that state being committed. The current early-return path also leaves the existing preview allocated when `assetPreviewResource` becomes `undefined` or file receiving is disabled.

I think the lifecycle becomes simpler with separate cleanup effects:

```typescript
useEffect(() => {
  return () => videoPreview?.dispose();
}, [videoPreview]);

useEffect(() => {
  return () => videoSrc?.dispose();
}, [videoSrc]);
```

Then we can use `setVideoPreview(url)` and `setVideoSrc(assetUrl)` directly. This cleans up the previous value after replacement and the current value on unmount, without refs or side effects inside state updates. We should also clear `videoPreview` when its resource is no longer available.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+      expect(mockGetAssetUrl).toHaveBeenCalledTimes(1);
+    });
+
+    fireEvent.click(screen.getByTestId('do-play-media'));
+
+    await waitFor(() => {
+      expect(mockGetAssetUrl).toHaveBeenCalledTimes(2);
+    });
+
+    for (const uploadProgress of [25, 50, 75]) {
```
screendriver, line 189: Before triggering the upload-progress re-renders, could we wait until the playback URL is actually assigned to the `<video>` element as the previous test does?

Waiting for `mockGetAssetUrl` to have been called twice only proves that loading started. `isVideoPlayable()` still resolves asynchronously through the queued metadata event, so these re-renders may happen while `videoSrc` is still `undefined`. In that case, this test passes even if re-rendering would revoke an active playback URL.

Something equivalent to this should happen before the loop:

```ts
const videoElement = document.querySelector(
  '[data-uie-name="video-asset"] video',
) as HTMLVideoElement;

await waitFor(() => {
  expect(videoElement.getAttribute('src')).toBe('blob:mock-video-src');
});
```

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
screendriver: I do not think avoiding the module mock requires a larger refactor. `VideoAssetProps` already exposes an `assetRepository` dependency; it is currently just not destructured or passed to `useAssetTransfer` 🙂

screendriver (changes requested):

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.tsx
thisisamir98: Addressed in https://github.com/wireapp/wire-webapp/commit/07d38039df — blob URL lifecycle is now owned by dedicated cleanup effects instead of state updater side effects.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.tsx
thisisamir98: Addressed in https://github.com/wireapp/wire-webapp/commit/6d9f305ebd — the async play path now checks a mounted ref after getAssetUrl() and isVideoPlayable() resolve, and includes a regression test with a deferred load + unmount.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
thisisamir98: Addressed in https://github.com/wireapp/wire-webapp/commit/5e2d28bcd7 — VideoAsset now destructures assetRepository and passes it to useAssetTransfer, and the tests inject a mock repository instead of mocking the module.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
thisisamir98: Addressed in https://github.com/wireapp/wire-webapp/commit/aadfb7144d — the upload-progress test now waits for the playback URL to be assigned to the video element before triggering progress updates.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/VideoAsset/VideoAsset.test.tsx
thisisamir98: Addressed in https://github.com/wireapp/wire-webapp/commit/5e2d28bcd7 — agreed, no larger refactor needed. VideoAsset now passes the existing assetRepository prop through to useAssetTransfer.

apps/webapp/src/script/components/MessagesList/Message/ContentMessage/asset/common/useAssetTransfer/useAssetTransfer.test.ts
```diff
@@ -90,6 +90,17 @@ describe('useAssetTransfer', () => {
...
       expect(window.URL.revokeObjectURL).toHaveBeenCalledWith('assetUrl');
     });
+
+    it('should keep a stable getAssetUrl reference between re-renders', () => {
+      const {result, rerender} = renderHook(() => useAssetTransfer(message, assetRepository));
+
+      const initialGetAssetUrl = result.current.getAssetUrl;
+
+      rerender();
+      rerender();
```
e-maad: Do we need it twice? Is it intentional?

thisisamir98: @e-maad yes, this is to try to reproduce the bug so we make sure it does not come back.

screendriver: What is missing is the opposite dependency path. One rerender should already reproduce the original problem. 

Calling `rerender()` twice does not add further coverage. What may be more valuable is testing the opposite path as well.

That would verify both guarantees:

* the callback remains stable while its dependency is unchanged;
* the callback updates when its dependency changes.

thisisamir98: Good point, thanks. I removed the extra rerender and added a test that checks getAssetUrl gets a new reference when assetRepository changes. That should cover both sides of the useCallback contract now.

screendriver (approved):

PR #21797, opened 2026-07-10
fix: never exposed original video when background effect enabled [WPB-26402]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26402" title="WPB-26402" target="_blank">BugWPB-26402</a>  [Calling] Never display real Background when BGE is enabled for any participante.
  </td></table>
  <br />
 

## Summary

This PR refactors the background effect application flow to ensure that the original camera stream is never temporarily exposed while background effects are enabled.

Previously, several code paths assigned the raw camera stream to the self participant before the background effect pipeline had finished initializing. This could briefly reveal the user's real background during camera changes or stream refreshes.

When background effects are enabled, camera changes no longer temporarily expose the original camera stream. The participant now switches directly to the processed stream once it is available while keeping the previous processed stream alive during the transition.


apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -377,74 +389,103 @@ export class CallingRepository {
...
 
-    if (sharesScreen) {
-      this.logger.error('The application allows to apply background effects in case of screen shares');
+    if (selfParticipant.sharesScreen()) {
+      this.logger.warn('Background effects are not applied while screen sharing');
       return;
     }
 
-    // let's check if background should be disabled, then let's do it and go back to the original video
+    const hasVideo = stream.getVideoTracks().length > 0;
```
zskhan:
```suggestion
    const hasVideo = stream.getVideoTracks().some(
  track => track.readyState === 'live',
);
```

zskhan (changes requested):

zskhan (approved):

PR #21799, opened 2026-07-10
feat: allow app to use Proteus when MLS is not available [WPB-26776]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26776" title="WPB-26776" target="_blank">BugWPB-26776</a>  [Web] Webapp needs MLS removal key to be set even if MLS is disabled on the Backend
  </td></table>
  <br />
 

### Description

Allow the app to use Proteus when MLS is not supported by the backend.


thisisamir98 (approved):
Looks good to me. Making MLS optional in ConversationService matches how account already skips MLSService when the backend does not support it, and Proteus paths stay untouched. I did not spot any critical behavior issues.

PR #21802, opened 2026-07-10
Improve WebSocket lifecycle observability [WPB-21916]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

Add correlated, privacy-safe lifecycle logs across WebSocket transport, authentication, notification processing, and reconnect orchestration. Include wake-suspension diagnostics and focused regression coverage without changing connection behavior.

### DIFF
--- a/libraries/api-client/src/tcp/webSocketClient.ts
+++ b/libraries/api-client/src/tcp/webSocketClient.ts
@@ -24,13 +24,14 @@ import {Maybe} from 'true-myth';
 
 import {EventEmitter} from 'events';
 
-import {LogFactory} from '@wireapp/commons';
+import {LogFactory, StringUtil} from '@wireapp/commons';
 
 import {AcknowledgeType} from './acknowledgeEvent.types';
 import {
   LongRunningRetryDetails,
   ReconnectingWebsocket,
   ReconnectingWebsocketWallClock,
+  WebSocketReconnectContext,
   WEBSOCKET_STATE,
 } from './reconnectingWebsocket';
 
@@ -66,7 +67,7 @@ export interface WebSocketClient {
   on(event: TOPIC.ON_STATE_CHANGE, listener: (state: WEBSOCKET_STATE) => void): this;
 }
 
-export type OnConnect = (abortHandler: AbortController) => void;
+export type OnConnect = (abortHandler: AbortController, reconnectContext: WebSocketReconnectContext) => void;
 
 export type WebSocketClientOptions = {
   readonly wallClock: ReconnectingWebsocketWallClock;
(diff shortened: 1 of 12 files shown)


libraries/commons/src/util/StringUtil.ts
```diff
@@ -88,16 +91,82 @@ function safeJsonStringify(obj: any): string {
...
+
+function sanitizeErrorName(errorName: string): string {
+  const isSafeErrorName =
+    errorName.length > 0 && errorName.length <= maximumSafeErrorNameLength && /^[A-Za-z0-9_.:-]+$/u.test(errorName);
+
+  return isSafeErrorName ? errorName : 'UnknownError';
+}
+
+export function getSafeErrorDetails(error: unknown): SafeErrorDetails {
+  if (is.error(error)) {
```
e-maad, line 162: Would ts-pattern be helpful here?

libraries/commons/src/util/StringUtil.ts
```diff
@@ -88,16 +91,82 @@ function safeJsonStringify(obj: any): string {
...
+  return redactSensitiveString(message.replace(/[\r\n]+/gu, ' ')).slice(0, maximumSafeLogMessageLength);
+}
+
+export function formatSafeLogValue(value: string): string {
+  return JSON.stringify(sanitizeLogMessage(value));
+}
+
+function sanitizeErrorName(errorName: string): string {
+  const isSafeErrorName =
+    errorName.length > 0 && errorName.length <= maximumSafeErrorNameLength && /^[A-Za-z0-9_.:-]+$/u.test(errorName);
```
e-maad: Can we use regex to define max and min length?
Or should we use is.nonEmptyString over length checks?

libraries/commons/src/util/StringUtil.ts
```diff
@@ -88,16 +91,82 @@ function safeJsonStringify(obj: any): string {
...
+function sanitizeUrlForLog(urlValue: string): string {
+  try {
+    const url = new URL(urlValue);
+    return url.search.length > 0 ? `${url.origin}${url.pathname}?${redactedValue}` : `${url.origin}${url.pathname}`;
+  } catch {
+    return urlValue;
+  }
+}
+
+function redactSensitiveString(value: string): string {
```
e-maad, line 131: Can we define regex separately?

libraries/api-client/src/tcp/webSocketClient.ts
```diff
@@ -189,11 +193,11 @@ export class WebSocketClient extends EventEmitter {
     this.socket.setOnMessage(this.onMessage);
     this.socket.setOnError(this.onError);
     this.socket.setOnLongRunningRetry(this.onLongRunningRetry);
-    this.socket.setOnOpen(() => {
+    this.socket.setOnOpen((_event, reconnectContext) => {
       this.onOpen();
       if (onConnect !== undefined) {
         this.abortHandler = new AbortController();
-        void onConnect(this.abortHandler);
+        void onConnect(this.abortHandler, reconnectContext);
```
e-maad, line 200: Should we use fireAndForget invoker?

e-maad (dismissed):
Added some suggestions

libraries/commons/src/util/StringUtil.ts
screendriver: I do not think `ts-pattern` would improve this particular function. It is a small, linear narrowing from unknown to Error, string or fallback.

Using `ts-pattern` here would also introduce a new dependency into `@wireapp/commons` for a case that is already explicit with type guards.

But generally speaking: `ts-pattern` is often a good idea 👍

libraries/commons/src/util/StringUtil.ts
screendriver: I prefer keeping the maximum length explicit because it is a semantic logging limit rather than part of the allowed-character definition.

Good point about the minimum length: the `+` in the pattern already rejects an empty string, so the explicit `length > 0` check is redundant. I will remove it. `is.nonEmptyString()` would be redundant for the same reason ➡️ 0162f73e6d653cc0a762ec09b5c35e44bdbe1bfc

libraries/commons/src/util/StringUtil.ts
screendriver: Yes, agreed. These patterns describe different sensitive-data categories, so naming them separately will make the sanitization pipeline easier to read and review ➡️ 0162f73e6d653cc0a762ec09b5c35e44bdbe1bfc

libraries/api-client/src/tcp/webSocketClient.ts
screendriver: I would keep the existing `void` invocation here.

The fire-and-forget invoker currently belongs to `@wireapp/core`, while this code is in `@wireapp/api-client`. Importing it would reverse the intended dependency direction.

This pull request also deliberately preserves the existing callback behaviour and only adds the reconnect context. Introducing tracked error handling would be a separate behavioural change. A generic invoker could be moved to commons or injected in a separate change if we want to standardize this.

thisisamir98 (approved):
LGTM

PR #21812, opened 2026-07-13
Fix accumulated window title event subscriptions [WPB-27018]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27018" title="WPB-27018" target="_blank">BugWPB-27018</a>  [Web] Window title event subscriptions accumulate and flood application logs
  </td></table>
  <br />
 

## Problem

`WindowTitleUpdater` subscribes to `WebAppEvents.EVENT.NOTIFICATION_HANDLING_STATE` from a React effect, but the subscription is never removed.

The effect reruns whenever its callback dependencies change. Each rerender therefore adds another listener while the previous listeners remain active. Those listeners also retain stale React state.

After the application has been running for some time, a single notification handling state change can invoke thousands of callbacks. This floods the logs with repeated messages such as:

```text
@wireapp/webapp/WindowTitlesViewModel Set window title update state to 'true'
```

This is not only a logging issue. Every stale callback can also attempt to update the component state.

The title update callback additionally enables title updates on every execution, which can overwrite the state derived from the current notification handling state.

## Solution

This change separates notification state handling from document title updates.

The notification subscription now has an explicit cleanup function that unsubscribes the same handler when the effect reruns or the component unmounts. This ensures that only one active listener exists.

The title update effect no longer enables title updates itself. Whether title updates are allowed is controlled exclusively by the current notification handling state.

The debug message also reports the new state being applied rather than the previous React state value.

## Result

Notification state transitions invoke a single handler, stale listeners no longer accumulate, and `WindowTitlesViewModel` no longer floods the logs with duplicate messages.

### DIFF
--- a/apps/webapp/src/script/page/components/windowTitleUpdater.test.tsx
+++ b/apps/webapp/src/script/page/components/windowTitleUpdater.test.tsx
@@ -17,16 +17,177 @@
  *
  */
 
-import {render} from '@testing-library/react';
+import {act, render} from '@testing-library/react';
+import {amplify} from 'amplify';
 
+import {WebAppEvents} from '@wireapp/webapp-events';
+
+import {NOTIFICATION_HANDLING_STATE} from 'Repositories/event/NotificationHandlingState';
+import {useKoSubscribableChildren} from 'Util/componentUtil';
+import {getLogger} from 'Util/logger';
 import {translateForTest} from 'Util/test/translateForTest';
 
+import {ContentState, useAppState} from '../useAppState';
 import {WindowTitleUpdater} from './windowTitleUpdater';
 
+jest.mock('amplify');
+jest.mock('tsyringe', () => {
+  const actualTsyringeModule = jest.requireActual('tsyringe');
+
+  return {
+    ...actualTsyringeModule,
+    container: {
+      resolve: jest.fn(),
+    },
+  };
+});
+jest.mock('Util/componentUtil');
+jest.mock('Util/logger', () => {
+  return {
+    getLogger: jest.fn(() => {
+      return {debug: jest.fn()};
+    }),
+  };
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/page/components/windowTitleUpdater.test.tsx
```diff
@@ -17,16 +17,177 @@
...
+  };
+});
+jest.mock('Util/componentUtil');
+jest.mock('Util/logger', () => {
+  return {
+    getLogger: jest.fn(() => {
+      return {debug: jest.fn()};
+    }),
+  };
+});
```
screendriver, line 51: `jest.mock()` is on purpose because otherwise I would have to rewrite the whole underlying "system" here

thisisamir98 (approved):
.

thisisamir98 (changes requested):
https://github.com/wireapp/wire-webapp/pull/21812/changes#diff-8a0b6250d27069da0b5e2b1c14514c1f60330c5917279bc7ab1da3683dc5673cR33-R51

using jest apis to mock entire libraries and files does not sound the best

screendriver:
> [#21812 (changes)](https://github.com/wireapp/wire-webapp/pull/21812/changes#diff-8a0b6250d27069da0b5e2b1c14514c1f60330c5917279bc7ab1da3683dc5673cR33-R51)
> 
> using jest apis to mock entire libraries and files does not sound the best

Did you read my comment? https://github.com/wireapp/wire-webapp/pull/21812/changes#r3568498274 `jest.mock()` is on purpose because otherwise I would have to rewrite the whole underlying "system" here

thisisamir98 (approved):

PR #21820, opened 2026-07-13
Fix TC-60 by removing condition from test flow [WPB-22420]

markbrockhoff:
<table>
<td>
  </td></table>
  <br />
 

This is a partial revert of #21790 which introduced a condition only removing a device from the list of devices in some cases. Tests depending on this user rely on the fact that it's re-used and always contains to many devices. It's a trade off, but much more reliable than using conditions within tests to tailor to different scenarios. The modified test already failed as part of the CI of #21790

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
@@ -107,13 +107,10 @@ test.describe('account settings', () => {
 
       await test.step('Remove an existing device and confirm new history', async () => {
         // Since this test re-uses the same user over and over again we need to always remove one of the previously registered devices
-        const removeDeviceButton = page.getByRole('button', {name: 'Remove device'}).first();
-        const isRemoveDeviceButtonVisible = await removeDeviceButton.isVisible({timeout: 3_000}).catch(() => false);
-        if (isRemoveDeviceButtonVisible) {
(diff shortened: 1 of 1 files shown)


apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
```diff
@@ -107,13 +107,10 @@ test.describe('account settings', () => {
 
       await test.step('Remove an existing device and confirm new history', async () => {
         // Since this test re-uses the same user over and over again we need to always remove one of the previously registered devices
-        const removeDeviceButton = page.getByRole('button', {name: 'Remove device'}).first();
```
screendriver, line 110: Is https://github.com/wireapp/wire-webapp/pull/21821 going in the same direction?

markbrockhoff: Yes, just with more overhead and complicated then / catch structures IMHO
Wasn't aware of this one, thanks for the heads up :)

iskvortsov (approved):
Thank you, I've closed my PR: https://github.com/wireapp/wire-webapp/pull/21821 

Let's merge yours instead as I don't feel like we'll benefit a lot from checking if delete device screen is shown.

zskhan (approved):

e-maad (approved):

PR #21821, opened 2026-07-13
chore(e2e): Fixed TC-60 [WPB-26999]

iskvortsov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26999" title="WPB-26999" target="_blank">TaskWPB-26999</a>  [Plaform] Create and configure SCIM management tool instance to be used for automated (e2e) and manual testing
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-26999

Fixed the checking for visibility that returned wrong results in the old implementation.

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
@@ -108,9 +108,9 @@ test.describe('account settings', () => {
       await test.step('Remove an existing device and confirm new history', async () => {
         // Since this test re-uses the same user over and over again we need to always remove one of the previously registered devices
         const removeDeviceButton = page.getByRole('button', {name: 'Remove device'}).first();
-        const isRemoveDeviceButtonVisible = await removeDeviceButton.isVisible({timeout: 3_000}).catch(() => false);
+        const isRemoveDeviceButtonVisible = await removeDeviceButton.waitFor({state: 'visible', timeout: 5_000}).then(() => true).catch(() => false);
         if (isRemoveDeviceButtonVisible) {
-          await removeDeviceButton.click({timeout: LOGIN_TIMEOUT});
+          await removeDeviceButton.click();
         }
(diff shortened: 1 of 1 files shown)


apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
```diff
@@ -108,9 +108,9 @@ test.describe('account settings', () => {
       await test.step('Remove an existing device and confirm new history', async () => {
         // Since this test re-uses the same user over and over again we need to always remove one of the previously registered devices
         const removeDeviceButton = page.getByRole('button', {name: 'Remove device'}).first();
-        const isRemoveDeviceButtonVisible = await removeDeviceButton.isVisible({timeout: 3_000}).catch(() => false);
+        const isRemoveDeviceButtonVisible = await removeDeviceButton.waitFor({state: 'visible', timeout: 5_000}).then(() => true).catch(() => false);
```
screendriver, line 111: Combining `await` and `.then()`/`.catch()` is generally not a good idea because it mixes two different approaches. Can't we just use plain `try/catch` with `await` here?

screendriver (changes requested):

apps/webapp/test/e2e_tests/specs/AccountSettings/accountSettings.spec.ts
screendriver, line 111: Is https://github.com/wireapp/wire-webapp/pull/21820 going in the same direction?

iskvortsov: Yep, definitely. Thanks for the suggestion,

iskvortsov: Yeah, it does. Tbh, I'd go with Mark's solution as I don't think it's a big problem that we don't have a processing for a case when delete device screen is not shown.

PR #21822, opened 2026-07-13
Join meeting calls from the Meetings tab [WPB-25057]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary

There are two kinds of meetings: **scheduled** meetings and **Meet Now** (instant) meetings. Both use the same meeting conversation and call-join path.

- Add the shared meeting call-join path (`joinMeetingCall`, `useJoinMeetingCall`, `CallingRepository` meeting/conference support).
- Wire the scheduled meeting **Join** action end to end from the Meetings list.
- Meet Now: the button handler is still a stub. A follow-up will create an instant meeting and then call the same `joinMeetingCall` path.
- Handle `conversation.create-meeting` websocket events and persist the embedded `conversation` from meeting create/update API responses so meeting call UI uses `conversation.display_name()` instead of client-side title syncing.
- Show the call island on the Meetings tab via `MeetingCallingView`, with improved join/attending state and group-call UI for meeting conversations.

## Test plan

- [x] Schedule a meeting and join from the Meetings list while the call is active; verify audio/video connects.
- [x] Confirm the call island appears on the Meetings tab with the correct meeting title.
- [x] Verify **Attending** state updates promptly after joining.
- [x] Edit a meeting title and confirm the call island/conversation name updates once the backend rename lands (depends on wire-server PR).
- [x] Run `yarn nx run webapp:test --testFile=src/script/components/Meeting/joinMeetingCall.test.ts`
- [x] Run `yarn nx run webapp:test --testFile=src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.test.ts`
- [x] Run `yarn nx run api-client-lib:test --testFile=src/meetings/meetingSchema.test.ts`


apps/webapp/src/script/components/calling/CallingCell/CallingCell.tsx
```diff
@@ -378,18 +384,18 @@ export const CallingCell = ({
           )}
 
           <CallingHeader
-            isGroup={isGroupOrChannel}
+            isGroup={isGroupCall}
```
e-maad: Maybe it's time to rename `isGroup`

apps/webapp/src/script/components/Meeting/MeetingCallingView/meetingCallingView.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const {properties: propertiesRepository} = mainViewModel.content.repositories;
+  const callState = container.resolve(CallState);
+  const teamState = container.resolve(TeamState);
+  const {activeCalls} = useKoSubscribableChildren(callState, ['activeCalls']);
+  const {classifiedDomains} = useKoSubscribableChildren(teamState, ['classifiedDomains']);
+  const {callingRepository} = callingViewModel;
+
+  const meetingCalls = activeCalls.filter(call => call.conversation.isMeeting());
+
+  if (meetingCalls.length === 0) {
```
e-maad, line 41: `is.nonEmptyArray`

apps/webapp/src/script/components/Meeting/MeetingCallingView/MeetingCallingView.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const meetingCalls = activeCalls.filter(call => call.conversation.isMeeting());
+
+  if (meetingCalls.length === 0) {
+    return null;
+  }
+
+  return (
+    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-view">
+      {meetingCalls.map(call => {
+        const {conversation} = call;
```
e-maad: I would suggest to not use `const` within the template.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+const qualifiedConversation = {domain: 'example.com', id: 'meeting-conversation-id'};
+
+const createTestProps = (): MeetingStatusProps & {joinMeeting: jest.Mock} => {
+  const joinMeeting = jest.fn();
+
+  return {
+    qualifiedConversation,
+    start_date: '2026-07-13T09:00:00.000Z',
+    end_date: '2026-07-13T10:00:00.000Z',
+    nowMilliseconds: new Date('2026-07-13T09:30:00.000Z').getTime(),
```
e-maad: Please make sure that the days don't have any sideeffect in future.
I would also suggest to use `optimisticWallClock`.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -28,28 +30,36 @@ import {
...
 }
 
 const MeetingStatusComponent = ({
+  qualifiedConversation,
   start_date,
   end_date,
   attending,
-  nowMilliseconds: providedNowMilliseconds,
+  nowMilliseconds,
+  useJoinMeetingCallHook = useJoinMeetingCall,
```
e-maad: Passing a hook as a prop is not the best approach. I see that you are doing it for tests only but can you look for other options?

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -28,28 +30,36 @@ import {
...
 import {getMeetingStatusAt, MeetingStatuses} from 'Components/Meeting/utils/meetingStatusUtil';
 import {useApplicationContext} from 'src/script/page/rootProvider';
 
+export type UseJoinMeetingCallResult = ReturnType<typeof useJoinMeetingCall>;
+
 export interface MeetingStatusProps {
+  qualifiedConversation: QualifiedId;
   start_date: string;
   end_date: string;
   attending?: boolean;
```
e-maad: I see that this prop only used for tests.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingListItem.tsx
```diff
@@ -110,9 +112,9 @@ const MeetingListItemComponent = ({
       <div css={rightStyles}>
         <MeetingParticipants qualifiedConversation={meetingSeries.qualified_conversation} isOngoing={isOngoing} />
         <MeetingStatus
+          qualifiedConversation={meetingSeries.qualified_conversation}
           start_date={startDateIso}
           end_date={endDateIso}
-          attending={attending}
```
e-maad, line 115: You removed the only usage but it is there is tests.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -60,6 +62,17 @@ const mapSyncErrorToSubmitError = (error: MeetingConversationSyncError): Meeting
...
 };
 
+const saveMeetingConversationFromResponse = (
+  conversationRepository: MeetingStoreDeps['conversationRepository'],
+  conversation: MeetingWithConversation['conversation'] | undefined,
+  onFailure: MeetingSubmitErrors,
+): Task<void, MeetingSubmitErrors> =>
+  conversation
+    ? conversationRepository
+        .saveMeetingConversationFromBackend(conversation as unknown as BackendConversation)
```
e-maad: How about using zod here for runtime validation?

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -105,21 +124,27 @@ export const updateMeeting = (
...
-        isCreate: false,
-      }).mapRejected(mapSyncErrorToSubmitError);
-    });
+    .andThen(updatedMeeting =>
+      saveMeetingConversationFromResponse(
+        deps.conversationRepository,
+        updatedMeeting.conversation,
+        meetingSubmitErrors.updateFailed,
+      ).andThen(() => {
+        if (usersToAdd.length === 0 && userIdsToRemove.length === 0) {
```
e-maad: Usage of `isNonEmptyArray`?

apps/webapp/src/script/repositories/entity/Conversation.ts
```diff
@@ -621,6 +621,29 @@ export class Conversation {
         }
       }
 
+      if (this.isMeeting()) {
+        if (this.name()) {
```
e-maad: Is not `null`, `undefined` or `emptyString`?

apps/webapp/src/script/repositories/entity/Conversation.ts
```diff
@@ -621,6 +621,29 @@ export class Conversation {
...
+          return this.name();
+        }
+
+        const hasUserEntities = !!this.participating_user_ets().length;
```
e-maad, line 630: Usage of `is.nonEmptyArray`?

libraries/api-client/src/meetings/meetingSchema.ts
```diff
@@ -52,8 +62,66 @@ export const meetingSchema = z.object({
...
+const conversationMemberSchema = z.object({
+  id: z.string(),
+  conversation_role: z.string().optional(),
+  hidden: z.boolean().optional(),
+  hidden_ref: z.string().nullable().optional(),
+  otr_archived: z.boolean().optional(),
+  otr_archived_ref: z.string().nullable().optional(),
+  otr_muted_ref: z.string().nullable().optional(),
+  otr_muted_status: z.number().nullable().optional(),
+  service: z.null().optional(),
```
e-maad: Why we need `service` to be always null?

e-maad (dismissed):
Well done!
I see that you used `useMemo` for most of the components, make sure we explicitely need them to prevent rerenders.

Added few suggestions.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.test.tsx
thisisamir98: done in de01f562024db3add99136ae506d55ebd0ac6122

apps/webapp/src/script/components/Meeting/MeetingCallingView/MeetingCallingView.tsx
thisisamir98: 9fe8e7d2c1cd49cae186a10dfded0ce7cc1a7023

apps/webapp/src/script/components/calling/CallingCell/CallingCell.tsx
thisisamir98: 5027bbd6dab089dfd3f4cff54423dc3ce77f35db

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
thisisamir98: it is also passed to the original [useJoinMeetingCall](https://github.com/wireapp/wire-webapp/pull/21822/changes#diff-f8a159cb7770d8c44afc2ff14e2aded92c56cdb48029bab4d9a6cf52d868fb04R37)

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingListItem.tsx
thisisamir98: Good catch — removed the unused `attending` prop from `MeetingStatus` and the `MeetingSeries` type. Participating state now comes solely from `isCallActive`.

https://github.com/wireapp/wire-webapp/commit/29dbdae6ef5b44100bd61e88ad78c96d581eb345

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
thisisamir98: 28eec081fd7ebcca4adb802468dadf245c8d0900

apps/webapp/src/script/repositories/entity/Conversation.ts
thisisamir98: fd5430588a96ce04d98c3ab81a8105774bb839bf

apps/webapp/src/script/repositories/entity/Conversation.ts
thisisamir98: e9e42af83c68e9f54854684388abdf82a4864a6a

libraries/api-client/src/meetings/meetingSchema.ts
thisisamir98: Dropped `service` from the schema — we don't use it for meeting conversations right now, so Zod will strip it if the backend sends it: https://github.com/wireapp/wire-webapp/commit/efdeedbe9d

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
thisisamir98: Zod validation already happens upstream in the API client.

apps/webapp/src/script/components/Meeting/useJoinMeetingCall.ts
```diff
@@ -0,0 +1,149 @@
...
+    setIsJoining(true);
+
+    guardCall(async () => {
+      const result = await joinMeetingCall(deps, qualifiedConversationId);
+
+      if (result.isErr && result.error === joinMeetingCallErrors.conversationNotFound) {
+        showConversationNotFoundModal();
+      }
+
+      setIsJoining(false);
```
zskhan, line 164: When there is no internet the guardCall will never execute the callback so the `isJoining` will be stuck to `true`

e-maad (dismissed):

apps/webapp/src/script/components/Meeting/useJoinMeetingCall.ts
thisisamir98: thanks, addressed in: https://github.com/wireapp/wire-webapp/pull/21822/commits/0eb52dea9b8d1357228d58d77f100bb925abd28e

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -1331,6 +1332,20 @@ export class ConversationRepository {
...
+  /**
+   * Persists a meeting conversation returned by the meetings API or websocket.
+   */
+  saveMeetingConversationFromBackend(conversationData: BackendConversation): Task<void, unknown> {
+    return task.tryOrElse(
+      error => error,
+      async () => {
+        const [conversationEntity] = this.mapConversations([conversationData]);
+        await this.updateParticipatingUserEntities(conversationEntity);
+        await this.saveConversation(conversationEntity);
```
screendriver: The new method maps a fresh entity and passes it into `saveConversation()`. Its merge path discards every function-valued property, but Knockout observables such as `name`, `type`, and `groupConversationType` are functions. Updating an already-loaded meeting therefore does not update its observable-backed data, including the title shown in the call UI.

libraries/api-client/src/meetings/meetingSchema.ts
```diff
@@ -52,8 +62,65 @@ export const meetingSchema = z.object({
...
+  qualified_id: qualifiedIdSchema,
+  creator: z.string(),
+  type: z.number(),
+  access: z.array(z.nativeEnum(CONVERSATION_ACCESS)).optional(),
+  access_role: z
+    .union([z.nativeEnum(CONVERSATION_LEGACY_ACCESS_ROLE), z.array(z.nativeEnum(CONVERSATION_ACCESS_ROLE))])
+    .optional(),
+  access_role_v2: z.array(z.nativeEnum(CONVERSATION_ACCESS_ROLE)).optional(),
+  cells_state: z.nativeEnum(CONVERSATION_CELLS_STATE).optional(),
+  group_conv_type: z.nativeEnum(GROUP_CONVERSATION_TYPE).optional(),
```
screendriver: The meeting schema strips canonical conversation fields such as `team`; the new test explicitly asserts that behavior. `ConversationMapper`, however, consumes `team` to establish `teamId`. The schema also accepts any group type and makes important MLS properties optional, despite the result later being cast to `BackendConversation`.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -3799,6 +3814,7 @@ export class ConversationRepository {
   ) {
     switch (eventJson.type) {
       case CONVERSATION_EVENT.CREATE:
+      case CONVERSATION_EVENT.CREATE_MEETING:
```
screendriver, line 3826: `conversation.create-meeting` is routed through the ordinary creation path. That path injects a creation message and decides between group and one-to-one using `isGroupOrChannel()`. Meetings return `false`, so a meeting containing other participants can produce a one-to-one creation event. The new test uses `others: []`, so it does not exercise this branch.

Also pattern matching with `ts-pattern` is not used here and could help potentially

apps/webapp/src/script/components/calling/CallingCell/CallingCell.tsx
```diff
@@ -378,18 +384,18 @@ export const CallingCell = ({
           )}
 
           <CallingHeader
-            isGroup={isGroupOrChannel}
+            isGroupOrChannel={isGroupOrChannel}
```
screendriver: The component introduces `isGroupCall = isGroupOrChannel || isMeeting`, but still passes only `isGroupOrChannel` to the header. Meetings consequently get one-to-one avatar and focus behavior, while alerts and controls classify them as group calls.

apps/webapp/src/script/components/Meeting/useJoinMeetingCall.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    guardCall(async () => {
+      setIsJoining(true);
+
+      const result = await joinMeetingCall(deps, qualifiedConversationId);
+
+      setIsJoining(false);
+
+      if (result.isErr && result.error === joinMeetingCallErrors.conversationNotFound) {
```
screendriver: `joinMeetingCall` returns both `conversationNotFound` and `joinFailed`, but the hook handles only the first. A rejected answer() or startAudio() produces no feedback.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -34,22 +34,25 @@ import {useApplicationContext} from 'src/script/page/rootProvider';
...
+  isJoinDisabled,
+  isCallActive,
 }: MeetingStatusProps) => {
   const {translate} = useApplicationContext();
-  const nowMilliseconds = providedNowMilliseconds ?? Date.now();
 
   const meetingStatus = useMemo(
-    () => getMeetingStatusAt(nowMilliseconds, start_date, end_date, attending),
-    [nowMilliseconds, start_date, end_date, attending],
+    () => getMeetingStatusAt(nowMilliseconds, start_date, end_date, isCallActive),
```
screendriver: The parent already owns `Date` objects, converts them to ISO strings, and derives `meetingStatus`. The child then parses those strings and derives the same status again. Participation is also mixed into temporal status, meaning a user in a call can still be classified as `UPCOMING` or `PAST` outside the scheduled interval.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingListItem.tsx
```diff
@@ -83,8 +85,8 @@ const MeetingListItemComponent = ({
   }, [end, start, nowMilliseconds, translate]);
 
   const meetingStatus = useMemo(
```
screendriver: Could we use `date-fns` for these date comparisons instead of manually comparing timestamps?

We should still obtain the current time from the injected WallClock, but helpers such as `isBefore`, `isAfter`, `isEqual`, or a small function built from them would make the interval semantics more explicit and easier to read.

The same applies to the meeting-status calculation below. Ideally, the date-fns-based calculation should live in one pure function and be reused here and by `MeetingStatus`, rather than implementing the same temporal rules in multiple components.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
```diff
@@ -34,22 +34,25 @@ import {useApplicationContext} from 'src/script/page/rootProvider';
...
+    [nowMilliseconds, start_date, end_date, isCallActive],
```
screendriver: Could this reuse a date-fns-based meeting interval calculation instead of parsing the ISO strings and implementing the comparisons separately?

The current time should continue to come from the WallClock, but date-fns would make the date semantics clearer and prevent the parent and child from maintaining two versions of the same calculation.

screendriver (changes requested):

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: Good catch, thanks. You were right.

`saveConversation()` was skipping Knockout observables because they are functions, so fields like `name` and `groupConversationType` never reached `updateProperties()` on the merge path.

I fixed this in 759073f3aa by adding `ConversationMapper.getUpdatableProperties()`, which unwraps writable observables before merging. I also added tests for the merge path and for `saveMeetingConversationFromBackend` with an already loaded meeting conversation.

One extra detail: once observables were unwrapped, merging `connection` could trigger its subscriber and add an empty user id. I kept the existing participant preservation logic and extended it to skip `participating_user_ets` and `connection` when the incoming conversation has no participants.

libraries/api-client/src/meetings/meetingSchema.ts
thisisamir98: Thanks for catching this. You were right.

I moved the embedded conversation validation into a shared `conversationSchema` that keeps the fields `ConversationMapper` needs, including `team`. The meeting-specific schema now requires `group_conv_type: meeting`, `protocol: mls`, and the MLS fields `group_id` and `epoch`.

I also removed the `BackendConversation` cast. `saveMeetingConversationFromBackend` now takes the validated meeting conversation type directly.

Two small follow-ups on the mapper side:
- only set `teamId` when the backend sends a non-empty team value
- keep the existing `teamId` on merge when an update payload does not include team context

The test that was stripping `team` now checks the opposite: canonical fields stay, unknown meeting-only fields still get removed.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: Good catch, thanks.

You were right about the creation message path. `addCreationMessage` only checked `isGroupOrChannel()`, so meetings could get a 1:1 creation event when other participants were present. I fixed this by treating meetings the same as groups there, and I updated the cleanup logic when loading message history. I also added a test with non-empty `others` to cover that branch.

On ts-pattern: I kept the switch for now. In this case it is only two event types sharing one handler, and the rest of `reactToConversationEvent` is still a large switch. Using ts-pattern only for the create branch would not really make the code clearer, and converting the whole dispatcher felt out of scope for this PR. Happy to revisit if we refactor that method more broadly later.

apps/webapp/src/script/components/calling/CallingCell/CallingCell.tsx
thisisamir98: Good catch, thanks. I renamed the header prop to `isGroupCall` and pass the combined `isGroupOrChannel || isMeeting` flag now, so meetings get the group avatar and the same focus behavior as alerts and controls. Fixed in d97723bf05.

apps/webapp/src/script/components/Meeting/useJoinMeetingCall.ts
thisisamir98: Good catch, thanks. I updated the hook so joinFailed now shows the call not established modal, with the same copy as the no internet guard. I also added tests for the joinFailed paths in joinMeetingCall and for the result handling in useJoinMeetingCall.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
thisisamir98: Good point, thanks for flagging this.

I split the schedule status from call participation. The list item now calculates the temporal status once from the Date objects and passes it to MeetingStatus, so we no longer parse ISO strings twice or derive the same status in both places.

Participation is handled separately now: if the user is in an active call, we show Participating regardless of the scheduled window. The temporal status is only used for the Join button and the time label.

Fixed in ad9566f830: https://github.com/wireapp/wire-webapp/commit/ad9566f830

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingListItem.tsx
thisisamir98: Good point, thanks. I addressed this in [ad9566f8](https://github.com/wireapp/wire-webapp/commit/ad9566f830f760cef63720ab824df97b8bbbbe15).

getMeetingTemporalStatusAt in meetingStatusUtil.ts now uses isAfter and isBefore from date-fns. MeetingListItem gets the current time from wallClock, computes the temporal status once, and passes it to MeetingStatus as a prop. That way the schedule logic lives in one place and MeetingStatus no longer parses dates on its own.

I also added tests in meetingStatusUtil.test.ts for the boundary cases.

apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/MeetingStatus/MeetingStatus.tsx
thisisamir98: Addressed in [ad9566f8](https://github.com/wireapp/wire-webapp/commit/ad9566f830f760cef63720ab824df97b8bbbbe15).

e-maad (dismissed):

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
```diff
@@ -102,6 +103,22 @@ export class ConversationMapper {
     });
   }
 
+  static getUpdatableProperties(conversationEntity: Conversation): Partial<Record<keyof Conversation, unknown>> {
+    const conversationData: Partial<Record<keyof Conversation, unknown>> = {};
+
+    for (const key in conversationEntity) {
+      const value = conversationEntity[key as keyof Conversation];
+
+      if (ko.isWriteableObservable(value)) {
```
screendriver, line 137: This now unwraps every writable observable from the newly created `Conversation`, not only the backend-owned fields we intend to update.

That includes local state initialized with defaults, such as `messages_unordered`, `incomingMessages`, timestamps, archive and verification state, and `readOnlyState`. When this object is merged into an already-loaded conversation, we could clear loaded messages or reset unrelated local state while only trying to update the meeting title.

Could we instead define an explicit set of backend-updatable properties, or update the existing entity directly from the validated backend payload?

Please also add a regression test where the existing conversation has loaded messages and non-default local state, then verify those values remain unchanged after updating the meeting title.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -1331,6 +1335,21 @@ export class ConversationRepository {
...
 
+  /**
+   * Persists a meeting conversation returned by the meetings API or websocket.
+   */
+  saveMeetingConversationFromBackend(conversationData: ValidatedMeetingConversation): Task<void, unknown> {
+    return task.tryOrElse(
+      error => error,
+      async () => {
+        // Validated meeting payloads are backend conversations with a narrowed meeting shape.
+        const [conversationEntity] = this.mapConversations([conversationData as BackendConversation]);
```
screendriver: In [your reply to the earlier schema comment](https://github.com/wireapp/wire-webapp/pull/21822#discussion_r3589224639), you mentioned that the `BackendConversation` type assertion had been removed, but it is still present here.

Since `conversationData` has already been validated and has the `ValidatedMeetingConversation` type, could we make `mapConversations` or the underlying mapper accept that type without an assertion?

The type assertion currently hides whichever structural difference prevents the validated type from being accepted, so TypeScript cannot verify that the schema and mapper remain compatible.

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
thisisamir98: Good catch. I changed `getUpdatableProperties` to use an explicit allowlist of backend-owned fields, so we still merge observable values like `name`, but we no longer touch client-only state such as loaded messages, archive timestamps, or verification/read-only flags.

I also added a regression test where an existing meeting conversation has loaded messages and non-default local state, then we update the title from the backend and check that local state stays the same.

Fixed in a7e451716f: https://github.com/wireapp/wire-webapp/commit/a7e451716fff42a48f23107f5cdb07ab3206cda6

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
thisisamir98: Good catch, and you were right. The cast was still there, and my earlier reply was wrong.

I fixed this in cb18ab9914: https://github.com/wireapp/wire-webapp/commit/cb18ab991452868b07e07f7306d598f13ffeaa42

The validated meeting conversation schema now matches the backend member types. Self members get null defaults for missing nullable fields, and other members turn null service values into undefined and default status to regular member. With that in place, `ValidatedMeetingConversation` is assignable to `BackendConversation`, so `saveMeetingConversationFromBackend` can call `mapConversations` without a cast. I also updated the `mapConversations` signature to accept the validated type directly.

There are tests for the normalization and for assignability to the backend conversation type.

apps/webapp/src/script/components/Meeting/useMeetingActions.ts
```diff
@@ -25,7 +25,7 @@ export const useMeetingActions = () => {
   const openCreate = useScheduleMeetingModal(state => state.openCreate);
 
   const handleMeetNow = () => {
-    // add calling functionality here
+    // Meet Now: create instant meeting, then call joinMeetingCall — see joinMeetingCall.ts
```
screendriver: `handleMeetNow()` still contains no implementation. Both the empty-state button and the meeting actions menu invoke this function, so users can select Meet now and nothing happens. Given that the Jira ticket is explicitly `[Web] "Meet now"` flow, changing the placeholder comment is not sufficient.

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
```diff
@@ -85,6 +86,32 @@ export type ConversationDatabaseData = ConversationRecord &
...
+  'cipherSuite',
+  'conversationModerator',
+  'creator',
+  'domain',
+  'epoch',
+  'globalMessageTimer',
+  'groupConversationType',
+  'groupId',
+  'initialProtocol',
+  'isGuest',
```
screendriver: `isGuest` is included in the new backend-updatable allowlist. However, it is not actually supplied by `conversationSchema` and the mapper only assigns it when `conversationData.is_guest` exists. A newly mapped API response therefore retains the constructor default of `false`; `getUpdatableProperties()` unwraps that value and overwrites an existing conversation whose `isGuest()` was `true`.

I would remove `isGuest` from the allowlist or derive updates from the validated payload while preserving whether a property was actually present. A regression test should start with `existing.isGuest(true)`, update the meeting title and verify it remains `true`.

apps/webapp/src/script/components/Meeting/useMeetingActions.ts
thisisamir98: I will be implementing meet now flow in an upcoming PR with the same ticket number to avoid the PR from grow very large

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
thisisamir98: Good catch, thanks. I removed `isGuest` from the backend updatable allowlist since it is client-derived and not part of `conversationSchema`. I also added regression tests to make sure guest status stays `true` when updating a meeting title.

Fixed in c9413e8081: https://github.com/wireapp/wire-webapp/commit/c9413e8081

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
```diff
@@ -102,6 +128,22 @@ export class ConversationMapper {
...
   }
 
+  static getUpdatableProperties(conversationEntity: Conversation): Partial<Record<keyof Conversation, unknown>> {
+    const conversationData: Partial<Record<keyof Conversation, unknown>> = {};
+
+    for (const key of BACKEND_UPDATABLE_CONVERSATION_PROPERTY_KEYS) {
+      const value = conversationEntity[key];
+
+      if (ko.isWriteableObservable(value)) {
+        conversationData[key] = value();
```
screendriver, line 138: Removing `isGuest` fixes that specific regression, but the underlying field-presence problem remains.

The meeting conversation schema allows fields such as `name` and `add_permission` to be omitted. `createConversationEntity()` converts those omissions into concrete defaults: `name` becomes an empty string and `add_permission` becomes `ADD_PERMISSION.ADMINS`.

Because `name` and `conversationModerator` are included in `BACKEND_UPDATABLE_CONVERSATION_PROPERTY_KEYS`, this line then merges those synthesized defaults into an existing conversation. A schema-valid response without `name` could therefore clear an existing meeting title.

Could we derive the update object directly from the validated backend payload so that omitted properties remain omitted, rather than deriving it from the normalized `Conversation` entity?

Please also add a regression test where an existing meeting has a title and non-default permission, the update response omits those fields and both existing values remain unchanged.

apps/webapp/src/script/components/Meeting/useMeetingActions.ts
screendriver: But then the pull request title and description is wrong. Could you please update it?

apps/webapp/src/script/repositories/conversation/ConversationMapper.ts
thisisamir98: Good catch. Meeting merges now build the update object from the validated backend payload, so omitted fields stay omitted instead of being filled with entity defaults.

I also added a regression test for an existing meeting that keeps its title and non-default add permission when those fields are missing from the update response.

https://github.com/wireapp/wire-webapp/commit/ed1bbce6c185abaae5002de72ea1fbfc2d1c6e6b

apps/webapp/src/script/components/Meeting/useMeetingActions.ts
thisisamir98: Updated. There are two meeting kinds: scheduled meetings and Meet Now (instant). This PR adds the shared call-join path and wires Join for scheduled meetings. Meet Now still only has a stub handler; a follow-up will create an instant meeting and then reuse the same `joinMeetingCall` path.

I updated the PR title and description to say that, and clarified the stub comment here:

https://github.com/wireapp/wire-webapp/commit/ce7cf339131019701ee588b8b6e144b011d35954

zskhan (approved):

PR #21824, opened 2026-07-13
Replace the custom promise queue with p-queue [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

## Summary

This replaces the internal `@wireapp/promise-queue` workspace library with [`p-queue`](https://github.com/sindresorhus/p-queue).

The existing implementation duplicated functionality already provided by a well-maintained library, including serial execution, concurrency control, pausing, resuming, task timeouts, and queue-state tracking.

During the migration, this change also uncovered and fixed a serious bug in the MLS conversation rejoin queue.

## Changes

All existing consumers now use the `p-queue` API directly:

* `push()` was replaced with `add()`
* `resume()` was replaced with `start()`
* queued task count now uses `size`
* running task state now uses `pending`
* initially paused queues now use `autoStart: false`

Every queue explicitly retains a concurrency of one and the existing 60-second task timeout.

Queues that need flushing use a small adapter based on `AbortController`. This is necessary because `PQueue.clear()` removes waiting tasks without settling the promises returned by `add()`.

The obsolete `@wireapp/promise-queue` workspace library, its tests, configuration, and remaining references have been removed.

## MLS rejoin bug fix

The previous conversation rejoin queue used a `Set` to prevent duplicate rejoin tasks for the same MLS group.

However, the group ID was removed from that set only after a successful rejoin. When the queued task rejected or timed out, the group ID remained registered indefinitely.

As a result, every later rejoin attempt for that conversation was silently ignored. An out-of-sync MLS conversation could therefore lose its ability to recover after one failed rejoin attempt.

The cleanup now happens in a `finally` block, ensuring the group ID is removed whether the task succeeds, rejects, or times out. Duplicate attempts are still suppressed while a rejoin is queued or running, but later recovery attempts remain possible after it settles.

Focused tests cover successful execution, rejection propagation, duplicate suppression, and retrying after a failed rejoin.

## Why

Queue scheduling is infrastructure code that does not provide product-specific value.

Using `p-queue` removes a custom concurrency implementation, reduces the maintenance and testing surface, and makes the intent at each call site more explicit.

The remaining custom code only covers Wire-specific flush semantics that `p-queue` deliberately does not provide.

The migration also made the lifecycle of MLS rejoin deduplication explicit and fixed a serious recovery issue in the old implementation.


apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -73,7 +73,7 @@ export class EventRepository {
   /** event processors are classes that are able to react and process an incoming event */
   private eventProcessors: EventProcessor[] = [];
 
-  private eventQueue: PromiseQueue = new PromiseQueue();
+  private eventQueue: PromiseQueue = new PromiseQueue({autoStart: true, concurrency: 1, timeout: 60_000});
```
e-maad: Can we move these option to const and used it everywhere?

e-maad (dismissed):
LGTM 🚀 
One small suggestion.

apps/webapp/src/script/repositories/event/EventRepository.ts
screendriver: Done in 8fa3a9764ed6510380441cf1cbf42f87a27d8135

zskhan (approved):

PR #21830, opened 2026-07-13
feat: add acc string for active and non-active conv [WPB-21212]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21212" title="WPB-21212" target="_blank">TaskWPB-21212</a>  [Web] Conversation list - Missing value for selected conversations
  </td></table>
  <br />
 

### Description

Add accessibility string for active and non-active conversations.


zskhan (dismissed):

zskhan (approved):

apps/webapp/src/script/components/conversationListCell/conversationListCell.tsx
ikotarac, line 116: I would put this in a function outside of the component

```tsx
const resolveConversationLabel = ({ unreadMessagesCount, isConversationMuted }: {...}) => {
  const accessibilityParts = [displayName];

  if (unreadMessagesCount > 0) {
    accessibilityParts.push(`${unreadMessagesCount} ${translate('conversationLabelUnread')}`);
  }

  if (isMutedConversation) {
    accessibilityParts.push(translate('accessibility.conversationMuted'));
  }

  accessibilityParts.push(
    translate(isActive ? 'accessibility.conversationSelected' : 'accessibility.conversationNotSelected'),
  );

  const accessibilityConversationLabel = `${accessibilityParts.join(', ')}. ${translate(
    'accessibility.openConversation',
    {
      name: displayName,
    },
  )}`;

}
```

or at least the same function inside of the component, just so that we know that all these ifs are part of label calculation. It is easier then to read the component later. For example, if you're not interested in label, you know you can just quickly skip this function, so it becomes easier to read.

ikotarac (commented):
I see it was merged in the meantime, nvm just a comment then :)

PR #21831, opened 2026-07-13
feat: add accessibility string for profile icon [WPB-21213]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21213" title="WPB-21213" target="_blank">TaskWPB-21213</a>  [Web] Conversation list - No text alternative for status indicator icon and missing output of available actions
  </td></table>
  <br />
 

### Description

Add accessibility string for profile link


apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -44,6 +44,7 @@ import {UserState} from 'Repositories/user/userState';
 import {MLSStatuses, WireIdentity} from 'src/script/e2eIdentity/e2eIdentityVerification';
 import {RootContextValue, useApplicationContext} from 'src/script/page/rootProvider';
 import {useKoSubscribableChildren} from 'Util/componentUtil';
+import {translate} from 'Util/localizerUtil';
```
screendriver: Please do not import the global `translate` function here. We inject `translate` through the application context. Pass it explicitly into `getUserVerificationBadgeLabel` as `RootContextValue['translate']` from the caller. This keeps the dependency explicit and makes the function independently testable.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,52 @@ export const UserVerificationBadges = ({
     user,
   });
 
-  let status: MLSStatuses | undefined = undefined;
+  let MLSStatus: MLSStatuses | undefined = undefined;
```
screendriver: Please use camelCase for this variable and the returned property: `mlsStatus`, not `MLSStatus`. The value is derived entirely from `mlsStatuses`, so it can also be expressed as a `const` instead of initializing a mutable `let` to `undefined` and assigning it conditionally.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,52 @@ export const UserVerificationBadges = ({
...
+};
+
+export const getUserVerificationBadgeLabel = ({
+  MLSStatus,
+  isProteusVerified,
+}: {
+  MLSStatus?: MLSStatuses;
+  isProteusVerified?: boolean;
+}): string | undefined => {
+  switch (MLSStatus) {
```
screendriver: Please use `ts-pattern` (pattern matching) for this state mapping and handle the input exhaustively. That makes every `MLSStatuses` value an explicit decision and ensures that newly added statuses cause a compile-time error instead of silently falling through.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,52 @@ export const UserVerificationBadges = ({
...
+    case MLSStatuses.VALID:
+      return translate('E2EI.userDevicesVerified');
+    case MLSStatuses.EXPIRED:
+    case MLSStatuses.EXPIRES_SOON:
+      return translate('E2EI.certificateExpired');
+    case MLSStatuses.REVOKED:
+      return translate('E2EI.certificateRevoked');
+    default:
```
screendriver: This `default` branch has no effect. It only breaks out of the switch and continues with the code below. More importantly, it hides missing `MLSStatuses` cases. Please remove it; an exhaustive `ts-pattern` match would make this branch unnecessary.

apps/webapp/src/script/page/leftSidebar/userDetails/userDetails.tsx
```diff
@@ -84,14 +90,27 @@ const UserDetailsComponent = ({user, isTeam = false, groupId, isSideBarOpen = fa
     });
   };
 
+  const avatarAriaLabel = [
+    userName,
+    userHandle,
+    isTeam ? translate(availabilityTranslationKeys[availability]) : undefined,
+    getUserVerificationBadgeLabel({MLSStatus, isProteusVerified}),
+  ]
+    .filter(Boolean)
```
screendriver: Please avoid `.filter(Boolean)` here. It is implicit about which values are removed and does not narrow the resulting type correctly. Use an explicit predicate, for example `(label): label is string => label !== undefined && label.length > 0`.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,52 @@ export const UserVerificationBadges = ({
...
+}: {
+  MLSStatus?: MLSStatuses;
+  isProteusVerified?: boolean;
+}): string | undefined => {
+  switch (MLSStatus) {
+    case MLSStatuses.VALID:
+      return translate('E2EI.userDevicesVerified');
+    case MLSStatuses.EXPIRED:
+    case MLSStatuses.EXPIRES_SOON:
+      return translate('E2EI.certificateExpired');
```
screendriver: `MLSStatuses.EXPIRES_SOON` must not return `E2EI.certificateExpired`. This announces that the certificate has already expired. Please handle it separately with `E2EI.certificateExpiresSoon`, matching the existing visual badge.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,52 @@ export const UserVerificationBadges = ({
...
+  let MLSStatus: MLSStatuses | undefined = undefined;
   if (mlsStatuses && mlsStatuses.length > 0 && mlsStatuses.every(status => status === MLSStatuses.VALID)) {
-    status = MLSStatuses.VALID;
+    MLSStatus = MLSStatuses.VALID;
+  }
+
+  return {MLSStatus, isProteusVerified};
+};
+
+export const getUserVerificationBadgeLabel = ({
```
screendriver: This function can only return one verification label. When MLS and Proteus verification are both active, `VerificationBadges` renders both badges, but the early MLS return means the Proteus status is omitted from the accessible name. Please compose all active verification labels instead of returning only the first one.

apps/webapp/src/script/page/leftSidebar/userDetails/userDetails.tsx
```diff
@@ -69,9 +73,11 @@ const UserDetailsComponent = ({user, isTeam = false, groupId, isSideBarOpen = fa
   const {
     name: userName,
     username: userHandle,
+    availability,
     isOnLegalHold,
     hasPendingLegalHold,
-  } = useKoSubscribableChildren(user, ['hasPendingLegalHold', 'isOnLegalHold', 'name', 'username']);
+  } = useKoSubscribableChildren(user, ['availability', 'hasPendingLegalHold', 'isOnLegalHold', 'name', 'username']);
+  const {MLSStatus, isProteusVerified} = useUserVerificationStatus({user, groupId, isSelfUser: isTeam});
```
screendriver: This calls `useUserVerificationStatus` a second time. `UserVerificationBadges` also calls the same hook, so the team-user case performs the identity lookup and registers the device-status listener twice. Please calculate the verification state once and reuse it for both the accessible label and the rendered badges.

apps/webapp/src/script/page/leftSidebar/userDetails/userDetails.tsx
```diff
@@ -37,18 +38,21 @@ import {AvailabilityContextMenu} from '../../../ui/availabilityContextMenu';
...
   children,
   isTeam = false,
+  ariaLabel,
   showAvailabilityContextMenu,
 }: AvailabilityStateButtonWrapperProps) => {
   return isTeam ? (
     <button
       onClick={event => showAvailabilityContextMenu(event.nativeEvent)}
       className="button-reset-default user-details-avatar"
+      aria-label={ariaLabel}
```
screendriver, line 56: This button opens the availability context menu, but that relationship is not exposed to assistive technology. Please add `aria-haspopup="menu"`. The other button that opens the same menu should expose the same semantics.

screendriver (changes requested):
Please add tests for the newly constructed accessible name. At minimum, cover the availability status, combined MLS and Proteus verification, and the `EXPIRES_SOON` translation. The current expired/expiring-soon mistake would have been caught by such a test.

apps/webapp/src/script/page/leftSidebar/userDetails/userDetails.tsx
```diff
@@ -69,9 +75,11 @@ const UserDetailsComponent = ({user, isTeam = false, groupId, isSideBarOpen = fa
   const {
     name: userName,
     username: userHandle,
+    availability,
     isOnLegalHold,
     hasPendingLegalHold,
-  } = useKoSubscribableChildren(user, ['hasPendingLegalHold', 'isOnLegalHold', 'name', 'username']);
+  } = useKoSubscribableChildren(user, ['availability', 'hasPendingLegalHold', 'isOnLegalHold', 'name', 'username']);
+  const verificationStatus = useUserVerificationStatus({user, groupId, isSelfUser: isTeam});
```
screendriver, line 82: This still causes `useUserVerificationStatus` to run twice. `UserDetailsComponent` calls it here, but the `UserVerificationBadges` rendered below calls the same hook again internally. The team-user path therefore still performs the identity lookup and registers the device-status listener twice.

Please calculate the verification status once and pass the resulting `mlsStatus` and `isProteusVerified` values to the rendered badges.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.test.tsx
```diff
@@ -92,3 +92,69 @@ describe('VerificationBadges', () => {
     expect(E2EIdentityStatus.getAttribute('data-uie-value')).toEqual(MLSStatuses.EXPIRES_SOON);
   });
 });
+
+describe('getUserVerificationBadgeLabel', () => {
```
screendriver, line 96: These tests cover `getUserVerificationBadgeLabel` in isolation, but they do not test the newly constructed accessible name assigned to the button in `UserDetails`.

Please add component tests that render `UserDetails` and verify the button’s accessible name. At minimum, cover the availability status, combined MLS and Proteus verification, and the `EXPIRES_SOON` translation. Otherwise, the integration between the status values and the actual accessible name remains untested.

apps/webapp/src/script/page/leftSidebar/userDetails/userDetails.tsx
```diff
@@ -84,14 +92,27 @@ const UserDetailsComponent = ({user, isTeam = false, groupId, isSideBarOpen = fa
...
+    .filter((label): label is string => is.nonEmptyString(label))
+    .join(', ');
+
   return (
     <div css={styles.wrapper(isSideBarOpen)}>
-      <AvailabilityStateButtonWrapper isTeam={isTeam} showAvailabilityContextMenu={showAvailabilityContextMenu}>
+      <AvailabilityStateButtonWrapper
+        isTeam={isTeam}
+        ariaLabel={avatarAriaLabel}
+        showAvailabilityContextMenu={showAvailabilityContextMenu}
```
screendriver, line 109: The other button that invokes `showAvailabilityContextMenu` still does not expose `aria-haspopup="menu"`. Please add `aria-haspopup="menu"` to both buttons that open the availability menu.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,58 @@ export const UserVerificationBadges = ({
...
+  return {mlsStatus, isProteusVerified};
+};
+
+export const getUserVerificationBadgeLabel = (
+  translate: RootContextValue['translate'],
+  {mlsStatus, isProteusVerified}: {mlsStatus?: MLSStatuses; isProteusVerified?: boolean},
+): string | undefined => {
+  const labels: string[] = [];
+
+  match(mlsStatus)
```
screendriver, line 133: `getUserVerificationBadgeLabel` has only one production caller, which passes the result of `useUserVerificationStatus`. That hook can only return `MLSStatuses.VALID` or `undefined` as `mlsStatus`.

The `EXPIRED`, `EXPIRES_SOON`, `REVOKED`, and `NOT_ACTIVATED` branches are therefore unreachable in production; they are exercised only by tests that manually pass values the real caller cannot provide.

Please limit this function and its tests to the states its actual caller can provide.

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,58 @@ export const UserVerificationBadges = ({
     user,
   });
 
-  let status: MLSStatuses | undefined = undefined;
-  if (mlsStatuses && mlsStatuses.length > 0 && mlsStatuses.every(status => status === MLSStatuses.VALID)) {
-    status = MLSStatuses.VALID;
+  const mlsStatus =
+    mlsStatuses && mlsStatuses.length > 0 && mlsStatuses.every(status => status === MLSStatuses.VALID)
+      ? MLSStatuses.VALID
+      : undefined;
```
screendriver, line 122: Can we avoid `undefined`?

apps/webapp/src/script/components/badge/components/verificationBadges/verificationBadges.tsx
```diff
@@ -115,12 +116,58 @@ export const UserVerificationBadges = ({
...
+
+  return {mlsStatus, isProteusVerified};
+};
+
+export const getUserVerificationBadgeLabel = (
+  translate: RootContextValue['translate'],
+  {mlsStatus, isProteusVerified}: {mlsStatus?: MLSStatuses; isProteusVerified?: boolean},
+): string | undefined => {
```
screendriver, line 130: Can we avoid `undefined` and use `Maybe` instead?

screendriver (changes requested):

PR #21842, opened 2026-07-14
fix(core): throttle and deduplicate focus-triggered backend sync tasks [WPB-23080]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-23080" title="WPB-23080" target="_blank">BugWPB-23080</a>  [Web] Performance - webapp makes at least 5 calls to the backend on any click I do
  </td></table>
  <br />
 

## Summary

- Returning focus to the webapp triggers several REST calls (`api-version`, `feature-configs`, team data, MLS key package count) via `RecurringTaskScheduler` tasks with `addTaskOnWindowFocusEvent`. This is intentional, but every refocus issued the full request burst with no cooldown.
- Throttle focus-triggered runs to at most once every 15 minutes, while keeping existing 24-hour scheduled syncs unchanged.
- Track focus listeners per task key so re-registration and cancellation replace the previous handler instead of accumulating duplicate `window` focus listeners.

## Test plan

- [x] `yarn nx run core-lib:test --testFile=src/util/recurringTaskScheduler/recurringTaskScheduler.test.ts`
- [x] Alt-tab away from the webapp and back within 15 minutes — no focus-triggered REST burst in Network tab
- [x] Alt-tab away and back after 15+ minutes — focus-triggered sync runs as before
- [x] Verify scheduled daily syncs still run (no change to timer-based behavior)

### DIFF
--- a/libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
+++ b/libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
@@ -19,6 +19,7 @@
 
 import {TimeUtil} from '@wireapp/commons';
 
+import {type FireAndForgetInvoker} from '../../taskExecution/fireAndForgetInvoker/fireAndForgetInvoker';
 import {LowPrecisionTaskScheduler} from '../lowPrecisionTaskScheduler';
 import {TaskScheduler} from '../taskScheduler';
 
@@ -42,7 +43,10 @@ export class RecurringTaskScheduler {
   private readonly lastExecutedAt = new Map<string, number>();
   private readonly windowFocusTaskKeys = new Set<string>();
 
-  constructor(private readonly storage: RecurringTaskSchedulerStorage) {}
+  constructor(
+    private readonly storage: RecurringTaskSchedulerStorage,
+    private readonly fireAndForgetInvoker: FireAndForgetInvoker,
+  ) {}
 
   public readonly registerTask = async ({
     every,
@@ -95,7 +99,7 @@ export class RecurringTaskScheduler {
           return;
         }
 
-        void executeTask();
+        this.fireAndForgetInvoker.fireAndForget(executeTask);
       };
 
       this.focusListeners.set(key, focusHandler);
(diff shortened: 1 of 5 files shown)


libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
```diff
@@ -68,14 +86,27 @@ export class RecurringTaskScheduler {
...
 
-    if (addTaskOnWindowFocusEvent === true && typeof window !== 'undefined') {
-      window.addEventListener('focus', taskConfig.task);
+      const focusHandler = () => {
+        const lastRun = this.lastExecutedAt.get(key);
+        if (lastRun !== undefined && Date.now() - lastRun < WINDOW_FOCUS_MIN_INTERVAL_MS) {
+          return;
+        }
+
+        void executeTask();
```
zskhan: FireAndForget?

libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
```diff
@@ -35,7 +35,13 @@ export interface TaskParams {
   addTaskOnWindowFocusEvent?: boolean;
 }
 
+const WINDOW_FOCUS_MIN_INTERVAL_MS = 15 * TimeUtil.TimeInMillis.MINUTE;
+
 export class RecurringTaskScheduler {
+  private readonly focusListeners = new Map<string, () => void>();
+  private readonly lastExecutedAt = new Map<string, number>();
```
zskhan, line 43: Along with this can we also `currently executing tasks` to avoid race condition as this `lastExecutedAt` is updated in the finally block.

libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
thisisamir98: Good call — switched to `fireAndForgetInvoker.fireAndForget(executeTask)` and injected `FireAndForgetInvoker` into `RecurringTaskScheduler` (wired from `Account`).

Fixed in 2bfdf6cfd2.

libraries/core/src/util/recurringTaskScheduler/recurringTaskScheduler.ts
thisisamir98: Added `executingTaskKeys` tracking — focus handlers and `executeTask` now bail out if a task is already running, so rapid focus events cannot bypass the throttle before `lastExecutedAt` is set in `finally`.

Fixed in 70fc2f3203.

zskhan (dismissed):

zskhan (approved):

PR #21855, opened 2026-07-15
Add collapsible conversation list [WPB-22420]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary
- Add a desktop-only collapsible conversation list that shrinks to a 64px avatar rail, toggled by clicking a `col-resize` divider between the list and messages panel.
- In collapsed mode, hide the title and show compact search/plus icon buttons; clicking search expands the list and focuses the input.
- Persist collapse state in the sidebar store, auto-expand on non-conversation tabs, and gate the feature behind the `conversation-list-collapse` startup flag (`?enabled-features=conversation-list-collapse`).

https://github.com/user-attachments/assets/1a61eade-b946-44ae-9388-dd634a0109cf

## Test plan
- [x] Open app with `?enabled-features=conversation-list-collapse` on desktop (>1000px width)
- [x] Click the divider between conversation list and messages to collapse/expand
- [x] Verify collapsed rail shows search, plus, and conversation avatars only
- [x] Click search icon in collapsed mode and confirm list expands with input focused
- [x] Switch to Preferences/Connect and confirm list auto-expands
- [x] Confirm messages panel uses freed horizontal space when collapsed
- [x] Confirm feature is disabled without the startup flag
- [x] Run `yarn nx run webapp:test --testFile=src/script/featureToggles/startupFeatureToggles.test.ts`
- [x] Run `yarn nx run webapp:test --testFile=src/script/page/leftSidebar/panels/conversations/useSidebarStore.test.ts`


apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
```diff
@@ -110,7 +117,20 @@ const ListWrapper = memo(
...
+            {showListCollapseHandle && onToggleListCollapsed && (
+              <button
+                type="button"
+                className="conversations-list-handle"
+                css={conversationsListHandleStyles}
+                onClick={onToggleListCollapsed}
+                aria-label={translate(
+                  isListCollapsed ? 'accessibility.conversationListExpand' : 'accessibility.conversationListCollapse',
+                )}
+                data-uie-name="toggle-conversation-list"
```
zskhan: aria-expanded={!isListCollapsed}

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
```diff
@@ -185,13 +195,19 @@ export const Conversations = ({
...
+  const isConversationListCollapsed =
+    isConversationListCollapseEnabled && conversationListStatus === ConversationListStatus.CLOSED;
+  const isConversationTab = isConversationListTab(currentTab);
+  const canCollapseConversationList =
+    isConversationListCollapseEnabled && isConversationTab && !isScreenLessThanMdBreakpoint;
 
   useEffect(() => {
     if (isScreenLessThanMdBreakpoint) {
       setSidebarStatus(SidebarStatus.CLOSED);
+      setConversationListStatus(ConversationListStatus.OPEN);
```
screendriver: Could we avoid overwriting the persisted collapse preference here?

`conversationListStatus` is persisted in `useSidebarStore.ts:168`, but this sets it permanently to `OPEN` when the viewport becomes smaller. The same happens when changing to a non-conversation tab at `conversations.tsx:363`.

The viewport and current tab are temporary conditions. I would persist the user's preference and derive the effective presentation instead:

```typescript
const isConversationListCollapsed =
  isConversationListCollapseEnabled &&
  conversationListStatus === ConversationListStatus.COLLAPSED &&
  isConversationListTab(currentTab) &&
  !isScreenLessThanMdBreakpoint;
```

That way, temporarily visiting Preferences or resizing the application does not silently discard the user's choice.

apps/webapp/src/style/foundation/framework.less
```diff
@@ -46,6 +46,14 @@
       border-right: none;
     }
 
+    &--conversation-list-collapsed {
+      min-width: auto;
+
+      .team-mode & {
+        width: auto;
```
screendriver: This `width: auto` appears to be overridden by the `.team-mode &` rule at line 58.

apps/webapp/src/style/list/conversation-list-cell.less
```diff
@@ -183,3 +183,49 @@ body.theme-dark {
...
 /* states */
+
+.conversations[data-list-collapsed='true'] {
+  .conversation-list-heading,
+  .conversation-list-no-results {
+    display: none;
+  }
+
+  .conversation-list-cell-center,
+  .conversation-list-cell-right .conversation-list-cell-context-menu,
```
screendriver, line 194: This hides the conversation options button with `display: none`

apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
```diff
@@ -25,6 +25,7 @@ import {throttle} from 'underscore';
 import {FadingScrollbar} from 'Components/fadingScrollbar';
 import * as Icon from 'Components/icon';
 import {useConnectionQuality} from 'src/script/hooks/useConnectionQuality';
+import {conversationsListHandleStyles} from 'src/script/page/leftSidebar/panels/conversations/conversations.styles';
```
screendriver: `ListWrapper` is a generic panel abstraction, but it now imports styles directly from the conversations feature. It also receives conversation-specific collapse properties at lines `65-67` and renders the collapse control at line `122`.

Could we keep this behavior inside the conversations feature instead?

apps/webapp/src/script/util/useConversationListCollapseFeatureFlag.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {conversationListCollapseFeatureToggleName} from 'src/script/featureToggles/startupFeatureToggleNames';
+import {useApplicationContext} from 'src/script/page/rootProvider';
+
+export const useConversationListCollapseFeatureFlag = () => {
```
screendriver: Do we need a custom hook for this?

It only reads `isFeatureToggleEnabled` from the application context, calls it with a constant, and wraps the boolean in an object. It does not combine multiple sources or hide meaningful behavior, but it adds another React dependency that tests must provide or mock.

I would resolve the feature flag once at the composition boundary and pass the boolean explicitly. This would also avoid reading the same context independently in `LeftSidebar` and `Conversations`

apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
```diff
@@ -110,7 +117,20 @@ const ListWrapper = memo(
       <>
         {sidebar}
         {children !== null ? (
-          <div id={id} className={`left-list-${id} ${id}`} css={style}>
+          <div id={id} className={`left-list-${id} ${id}`} css={style} data-list-collapsed={isListCollapsed}>
```
screendriver: Now that this element represents a named collapsible region, could we use a semantic `<section>` and connect it to its heading with `aria-labelledby`? Using semantic HTML is always a good idea.

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.styles.tsx
```diff
@@ -22,3 +22,16 @@ import {CSSObject} from '@emotion/react';
...
+  position: 'absolute',
+  zIndex: 10,
+  top: 0,
+  right: -6,
+  bottom: 0,
+  width: 12,
+  padding: 0,
+  border: 'none',
+  background: 'transparent',
+  cursor: 'col-resize',
```
screendriver: `col-resize` suggests that this divider can be dragged to resize the panel, but clicking it performs a binary collapse/expand action. Could we use a pointer cursor instead?

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationsList.tsx
```diff
@@ -253,6 +255,7 @@ export const ConversationsList = ({
               return (
                 <div
                   key={virtualItem.key}
+                  className="conversation-list-heading"
```
screendriver: This element is treated as a list heading, but it is still rendered as a generic `<div>`. The surrounding structure also places `<div>` wrappers directly inside the `<ul>` at lines `239` and `271`, so the resulting list markup is not semantic HTML.

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationHeader/conversationHeader.tsx
```diff
@@ -93,18 +100,25 @@ export const ConversationHeaderComponent = ({
     handleEnterDown(event, () => onSearchEnterClick(event));
   };
 
+  const focusSearchInput = useCallback(() => {
```
screendriver: Is the referential identity of `focusSearchInput` important here?

The function is inexpensive and `useCallback` appears to exist primarily because it is included in the effect dependency list.

The same question applies to the new callbacks at `conversations.tsx:285` and `conversations.tsx:294`

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.test.ts
```diff
@@ -57,6 +66,17 @@ describe('useSidebarStore', () => {
     expect(useSidebarStore.getState().status).toBe(SidebarStatus.CLOSED);
   });
 
+  it('sets conversation list status', () => {
```
screendriver: This test verifies that calling a `Zustand` setter assigns the provided value. That is `Zustand's` behavior rather than an application invariant.

The test at line `75` similarly checks only one included and one excluded array value.

Could we instead cover the behavior introduced by this pull request?

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.ts
```diff
@@ -81,9 +81,35 @@ export const SidebarStatus = {
...
 export type SidebarStatus = (typeof SidebarStatus)[keyof typeof SidebarStatus];
 
+export const ConversationListStatus = {
+  OPEN: 'OPEN',
+  CLOSED: 'CLOSED',
+} as const;
+
+export type ConversationListStatus = (typeof ConversationListStatus)[keyof typeof ConversationListStatus];
+
+export const CONVERSATION_LIST_TABS: readonly SidebarTabs[] = [
```
screendriver, line 91: This list duplicates the `showSearchInput` tab list in `conversations.tsx:173–185`

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.ts
```diff
@@ -81,9 +81,35 @@ export const SidebarStatus = {
 
 export type SidebarStatus = (typeof SidebarStatus)[keyof typeof SidebarStatus];
 
+export const ConversationListStatus = {
```
screendriver, line 84: Could we name these states `EXPANDED` and `COLLAPSED` rather than `OPEN` and `CLOSED`? Matching the domain terminology would make it clearer which state is being handled.

screendriver (changes requested):

apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
thisisamir98: Thanks for the suggestion! I added `aria-expanded` here, and also `aria-controls` since the panel already has an id. You can see the change in [b360b29](https://github.com/wireapp/wire-webapp/commit/b360b29f2d).

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
thisisamir98: Good catch, thanks! I removed the places that were resetting the stored preference and now derive the collapsed state from the preference plus the current tab and viewport. The list still looks expanded on Preferences and small screens, but your choice is kept when you come back. Fixed in [b5d0a69](https://github.com/wireapp/wire-webapp/commit/b5d0a6984a).

apps/webapp/src/style/foundation/framework.less
thisisamir98: You're right, the later team-mode rule was winning. I moved the collapsed width override inside that block so it comes after the calc rule and actually applies. Fixed in [b4fc267](https://github.com/wireapp/wire-webapp/commit/b4fc26721d).

apps/webapp/src/style/list/conversation-list-cell.less
thisisamir98: Yes, that is intentional. Collapsed mode is meant to stay minimal and show only avatars in the rail. Users can still open conversation options with right click, or expand the list again if they need the full controls.

apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
thisisamir98: Good point. I moved the collapse handle into a conversations-only component and kept `ListWrapper` generic with a `panelOverlay` slot instead. Fixed in [df9f607](https://github.com/wireapp/wire-webapp/commit/df9f6070f0).

apps/webapp/src/script/util/useConversationListCollapseFeatureFlag.ts
thisisamir98: Agreed. I removed the hook and now read the flag once in `LeftSidebar`, then pass `isConversationListCollapseEnabled` down to `Conversations` as a prop. Fixed in [6191859](https://github.com/wireapp/wire-webapp/commit/619185951a).

apps/webapp/src/script/page/leftSidebar/panels/listWrapper.tsx
thisisamir98: Good idea. I changed the panel root to a `section` and linked it to the heading with `aria-labelledby`. In collapsed mode the heading stays in the DOM as visually hidden text so the region keeps its name. Fixed in [6b088e8](https://github.com/wireapp/wire-webapp/commit/6b088e85ec).

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.styles.tsx
thisisamir98: Good catch. I changed it to a pointer since the handle only toggles on click and does not support drag resize. Fixed in [90f494f](https://github.com/wireapp/wire-webapp/commit/90f494f3b5).

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationsList.tsx
thisisamir98: Good point. I changed the search headings to `h3` elements and removed the `div` wrappers inside the `ul`, so virtualized rows are now direct `li` items. Fixed in [0d407aa](https://github.com/wireapp/wire-webapp/commit/0d407aabb4).

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationHeader/conversationHeader.tsx
thisisamir98: Good point, thanks for raising this.

I updated it in 521d878749:
- `focusSearchInput` is now a small helper, so we do not need `useCallback` there
- `toggleConversationList` is a plain function now, since it is only used as a click handler
- `expandConversationList` still uses `useCallback`, because `ConversationHeader` passes it into an effect dependency list

https://github.com/wireapp/wire-webapp/commit/521d8787493fedcc0c3bb294e5cee5136acefe8f

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.test.ts
thisisamir98: Good point, thanks.

I removed the setter smoke tests and added coverage for the behavior this PR introduces in 9a69bd5bea:
- collapse preference is persisted in the sidebar store
- the list stays visually expanded on non-conversation tabs even when the stored preference is closed
- collapse is gated by tab, viewport, and the feature flag

I also extracted small helpers for the derived collapse state so we can test that logic directly.

https://github.com/wireapp/wire-webapp/commit/9a69bd5bea8daabb5d490b7aa64397efc8e00d3e

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.ts
thisisamir98: Good catch, thanks.

I removed the duplicate list and now use `isConversationListTab(currentTab)` for `showSearchInput`, so search visibility and collapse logic share the same source of truth.

605951009c

https://github.com/wireapp/wire-webapp/commit/605951009c77cfdc0f4aa08d7ea1117d93ae0198

apps/webapp/src/script/page/leftSidebar/panels/conversations/useSidebarStore.ts
thisisamir98: Good idea, thanks.

I renamed the values to `EXPANDED` and `COLLAPSED` so they match the collapse terminology used elsewhere in this feature. No migration needed since this is still experimental.

f02e8db77f

https://github.com/wireapp/wire-webapp/commit/f02e8db77f6d4a841768966b9790e96387cb5a33

screendriver (approved):

PR #21858, opened 2026-07-15
ci: suppress ESLint warnings in GitHub Actions logs

thisisamir98:
## Summary
- Add optional `${ESLINT_QUIET:-}` placeholder to all Nx ESLint lint targets so CI can pass `--quiet` without changing local behavior.
- Set `ESLINT_QUIET: --quiet` on the CI lint job and library publish lint step so GitHub Actions logs only show ESLint errors.

## Test plan
- [x] Verified locally: `ESLINT_QUIET=--quiet yarn nx run webapp:lint` completes without printing ESLint warnings
- [x] CI lint job shows errors only when lint fails, without flooding logs with pre-existing warnings

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -252,6 +252,8 @@ jobs:
     name: lint
     permissions:
       contents: read
+    env:
+      ESLINT_QUIET: --quiet
 
     steps:
       - name: Checkout (pull_request)
--- a/.github/workflows/publish-libraries.yml
+++ b/.github/workflows/publish-libraries.yml
@@ -49,6 +49,8 @@ jobs:
         run: ./bin/yarn --immutable
 
       - name: Lint libraries
+        env:
+          ESLINT_QUIET: --quiet
         run: ./bin/yarn nx run-many -t lint --projects=tag:type:lib
 
       - name: Test libraries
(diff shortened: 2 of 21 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -252,6 +252,8 @@ jobs:
     name: lint
     permissions:
       contents: read
+    env:
+      ESLINT_QUIET: --quiet
```
screendriver, line 256: I do not think we should reintroduce `--quiet` in CI.

It was deliberately removed in #21282. The purpose of the fixed `--max-warnings` baseline is not only to prevent additional warnings from accumulating, but also to keep the existing warnings visible and actionable.

When `--quiet` and `--max-warnings` are used together, ESLint still counts warning-level violations, but it no longer reports their diagnostics. This means CI can fail because the warning budget was exceeded without showing which warning caused the failure. The developer then has to reproduce the lint run locally without `ESLINT_QUIET` just to find out what CI already knows.

That makes the CI output less actionable and reintroduces the guessing we intentionally removed. The existing warnings may produce a lot of output, but hiding them does not solve the underlying problem. It only hides the information needed to reduce them.

Please keep warnings visible in CI and remove the `ESLINT_QUIET` mechanism. If the log volume is the actual problem, we should address that separately rather than suppressing the diagnostics.

screendriver (changes requested):

PR #21860, opened 2026-07-15
build: envoy gateway support (helm) [WPB-24493]

jschaul:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-24493" title="WPB-24493" target="_blank">TaskWPB-24493</a>  add gateway support to webapp ingress (part of webapp helm chart)
  </td></table>
  <br />
 

In general we wish to switch from ingress-nginx to envoy-gateway. This has already been added as an option when the webapp is installed as part of the wire-server bundle (see https://github.com/wireapp/wire-server/pull/5150
 and related PRs). But when using this chart standalone, it also needs to be added here.

Ticket: https://wearezeta.atlassian.net/browse/WPB-24493

This PR goes beyond the minimal needed work though, and also does these things:

1. Introduces `ListenerSet`s. It's a newer object available since a few months, that makes it easier to share one gateway and EnvoyProxy for different domains, and define the ListernerSet / domains in the chart / overrides, without having to touch any infra.
2. Tests this thoroughly with three argocd-based deployments on three different domains:

## Deployments

- [x] Regression deployment: The default nginx-ingress helm chart is being deployed through argo
https://argocd.apps-staging.ops.zinfra.io/applications/argo-cd/webapp-pr-21860 using [these overrides](https://github.com/wireapp/argocd-web/blob/main/values/webapp/pr/nginx.yaml) and is available under https://webapp-pr-21860.apps-staging.wire.link
    - the only helm change is from `ingress: enabled: true` to `routing: enabled: true / mode: nginx` but I believe the standalone webapp is not being used/deployed anywhere except through argo manifests (which is updated alongside this PR)
- [x] Envoy based deployment: Using [these overrides](https://github.com/wireapp/argocd-web/blob/main/values/webapp/pr/envoy-gateway.yaml) this gets deployed with Argo and is available under https://webapp-pr-21860-envoy.apps-staging.wire.link and hinges on some extra infra in cailleach [this](https://github.com/zinfra/cailleach/pull/4612) and [this one](https://github.com/zinfra/cailleach/pull/4645)

Based on the the two modes there's a second "migration" mode which sets two extra annotations and then allows for a weighted DNS cutover without downtime:

- [x] Migration deployment online under https://webapp-pr-21860-migration.apps-staging.wire.link. See also [this confluence page](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/3095527493/How+to+cut+traffic+from+nginx-ingress+to+envoy-gateway+without+downtime+on+the+example+of+a+standalone+webapp) for more details. TL;DR: you can toggle/revert [this commit](https://github.com/wireapp/argocd-web/commit/72c75df65d2711d835ad451703ea68aea22a6f33) to switch incoming traffic for that url to either go through envoy-gateway or through ingress-nginx.

### DIFF
--- a/charts/webapp/templates/httproute.yaml
+++ b/charts/webapp/templates/httproute.yaml
@@ -1,5 +1,5 @@
-{{- $ingressMode := include "webapp.ingressMode" . -}}
-{{- if and .Values.ingress.enabled (or (eq $ingressMode "envoy") (eq $ingressMode "migration")) }}
+{{- $routingMode := include "webapp.routingMode" . -}}
+{{- if and .Values.routing.enabled (or (eq $routingMode "envoy") (eq $routingMode "migration")) }}
 {{- include "webapp.validateDomainConfig" . }}
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
@@ -10,9 +10,9 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-  {{- if or .Values.ingress.httpRouteAnnotations (eq $ingressMode "migration") }}
+  {{- if or .Values.gateway.httpRouteAnnotations (eq $routingMode "migration") }}
   annotations:
-    {{- range $key, $val := .Values.ingress.httpRouteAnnotations }}
+    {{- range $key, $val := .Values.gateway.httpRouteAnnotations }}
     {{ $key }}: {{ $val | quote }}
     {{- end }}
     {{- $weight := include "webapp.externalDnsWeight" (dict "ctx" . "kind" "envoy") | trim -}}
--- a/charts/webapp/templates/httproute.yaml
+++ b/charts/webapp/templates/httproute.yaml
@@ -10,19 +10,16 @@ metadata:
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
     release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
-  {{- if or .Values.gateway.httpRouteAnnotations (eq $routingMode "migration") (eq $routingMode "envoy") }}
+  {{- $setIdentifier := include "webapp.externalDnsSetIdentifier" (dict "ctx" . "kind" "envoy") | trim -}}
+  {{- $weight := include "webapp.externalDnsWeight" (dict "ctx" . "kind" "envoy") | trim -}}
+  {{- if or .Values.gateway.httpRouteAnnotations $setIdentifier $weight }}
   annotations:
     {{- range $key, $val := .Values.gateway.httpRouteAnnotations }}
     {{ $key }}: {{ $val | quote }}
     {{- end }}
-    {{- $setIdentifier := include "webapp.externalDnsSetIdentifier" (dict "ctx" . "kind" "envoy") | trim -}}
     {{- if $setIdentifier }}
     external-dns.alpha.kubernetes.io/set-identifier: {{ $setIdentifier | quote }}
     {{- end }}
-    {{- if not (hasKey .Values.gateway.httpRouteAnnotations "external-dns.alpha.kubernetes.io/hostname") }}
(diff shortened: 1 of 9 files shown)


siberijah (dismissed):
I think yes it's better to use this approach, because it covers all three: nginx, envoy and the weighted based dns

charts/webapp/values.yaml
```diff
@@ -18,12 +18,72 @@ appendReleaseNameToResources: false
 
 # webappDomain: webapp.example.com
 
-ingress:
+routing:
```
screendriver, line 21: This silently breaks existing chart upgrades. An existing values file with `ingress.enabled: true` now leaves `routing.enabled` at its default value of `false`, so neither the Ingress nor the Service is rendered. That turns a values rename into an outage.

Please either keep `ingress.enabled` as a deprecated alias for one release, or fail explicitly when the legacy key is supplied and explain the required migration.

charts/webapp/values.yaml
```diff
@@ -18,12 +18,72 @@ appendReleaseNameToResources: false
...
+  #     namespaces:
+  #       from: All
+  #   listeners:
+  #     - name: http
+  #       protocol: HTTP
+  #       port: 80
+  # - This chart creates a ListenerSet in the release namespace and an HTTPRoute that
+  #   attaches to it. When using a shared Gateway namespace, set `gateway.namespace`
+  #   to that namespace.
+  # - If tls.useCertManager=true and you rely on the default HTTP01 solver, the Gateway
```
screendriver: This is incomplete for the shared-Gateway setup described above.

cert-manager creates the temporary HTTPRoute in the Certificate namespace. When `gateway.namespace` differs from the release namespace, the Gateway's port 80 listener must also allow routes from that namespace, for example with `allowedRoutes.namespaces.from: All` or an appropriate namespace selector. The default only permits routes from the Gateway's own namespace, so HTTP-01 challenges would otherwise remain pending.

Please add `allowedRoutes` to the example and document that cert-manager's Gateway API integration must be enabled.

screendriver (changes requested):

charts/webapp/templates/httproute.yaml
```diff
@@ -0,0 +1,44 @@
...
+    external-dns.alpha.kubernetes.io/set-identifier: {{ $setIdentifier | quote }}
+    {{- end }}
+    {{- if $weight }}
+    external-dns.alpha.kubernetes.io/aws-weight: {{ $weight | quote }}
+    {{- end }}
+  {{- end }}
+spec:
+  parentRefs:
+    - name: {{ include "webapp.listenerSetName" . | quote }}
+      kind: ListenerSet
```
screendriver, line 31: This introduces an additional ExternalDNS prerequisite that is not currently documented.

Because the HTTPRoute references a ListenerSet rather than the Gateway directly, ExternalDNS must be started with `--gateway-listener-sets`. That option is disabled by default and requires Gateway API v1.5+ CRDs. Without it, Envoy may accept the route while ExternalDNS produces no DNS target, which also breaks the weighted migration mode.

Please document the required ExternalDNS configuration and supported version in `values.yaml`.

jschaul: Thanks, I'm aware, but now it's also documented in values.yaml.

charts/webapp/values.yaml
jschaul: I restructured the docs to take this into account.

charts/webapp/values.yaml
jschaul: As I wrote in the PR comment, I don't think this chart is currently used by anyone in standalone / with-ingress mode. Or am I wrong here? But if you insist, I add a backwards-compatibility to allow specifying ingress.enabled.

screendriver: Ah 💡
Good question. I don’t know to be honest. So just let’s try it out then I would say.

charts/webapp/templates/httproute.yaml
screendriver: 👍

screendriver (approved):

PR #21886, opened 2026-07-17
feat(webapp): add Meet Now modal and instant meeting flow [WPB-25057]

thisisamir98:
<table>
<td>
  </td></table>
  <br />
 

## Summary

- Add a Meet Now modal (title + optional participants) wired from the header dropdown and empty meetings list.
- Create ad-hoc meetings with RFC instant time defaults (start now, end +1h or midnight cap) and join the call directly with no preparation room.
- Reuse existing meeting creation, participant sync, list refresh, and join-call plumbing; inject `serviceTasks` at the meetings composition root.

## Test plan

- [x] Open Meetings → Meet now (header menu + empty list) and confirm modal opens
- [x] Submit without title → inline validation error
- [x] Submit with title, no participants → enters call directly; meeting shows as ongoing in list
- [x] Submit with participants → enters call; participants synced
- [x] Confirm Schedule Meeting flow is unchanged
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/MeetNowModal/useMeetNowSubmit.test.tsx`
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/meetingStore/createMeetingStore.test.ts`
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.test.ts`


apps/webapp/src/script/components/Meeting/MeetNowModal/useMeetNowSubmit.ts
```diff
@@ -0,0 +1,159 @@
...
+      translate,
+    }),
+    [translate],
+  );
+
+  const guardCall = useNoInternetCallGuard(callNotEstablishedCopy);
+
+  const joinDeps = useMemo<JoinMeetingCallDeps>(
+    () => ({
+      conversationState: container.resolve(ConversationState),
```
screendriver: Could we pass `ConversationState` as an explicit dependency instead of resolving it from the global container here?

This hook already obtains dependencies from the application context, the main view model, the meeting store and `tsyringe`. That makes the complete create-and-join flow difficult to test without constructing a substantial part of the application environment.

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const {result} = renderHook(() => useMeetNowSubmit(), {wrapper: createWrapper(store)});
+
+    let submitResult = false;
+    await act(async () => {
+      submitResult = await result.current.submit(formState);
+    });
+
+    expect(submitResult).toBe(true);
+    expect(meetNowMeeting).toHaveBeenCalledWith(formState);
+    expect(loadMeetings).toHaveBeenCalledTimes(1);
```
screendriver, line 164: Could we also verify that the returned `qualifiedConversation` is passed into the join operation?

This test currently proves only that the meeting is created and the list is refreshed. It would continue to pass if joining were removed, skipped or invoked with the wrong conversation ID, even though joining is the defining behaviour of Meet Now.

Please also cover the no-internet and join-failure outcomes so that the expected meaning of `submit()` is documented.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
```diff
@@ -0,0 +1,114 @@
+/*
```
screendriver, line 1: We agreed on that all directories and file names should be `camelCase`

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.ts
```diff
@@ -0,0 +1,159 @@
...
+          translate,
+        });
+      }
+
+      await task.tryOrElse(() => meetingSubmitErrors.refreshFailed, loadMeetings);
+
+      const {qualifiedConversation} = submitResult.value;
+
+      setIsSubmitting(false);
+      joinCreatedMeeting(qualifiedConversation);
```
screendriver: `submit()` returns `true` immediately after triggering `joinCreatedMeeting()`, but that does not mean the meeting was actually started or joined. The no-internet guard may skip the join completely, and `joinMeetingCall()` may still return an error.

Could we make this outcome explicit instead of returning a boolean that the caller interprets as `didStart`? For example, the result could distinguish between meeting creation and successful joining, or the joining operation could be awaited before reporting success.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+      }
+    });
+  };
+
+  return (
+    <ModalComponent
+      id="meet-now-modal"
+      wrapperCSS={{...modalWrapperStyles, ...meetNowModalWrapperStyles}}
+      isShown={isOpen}
+      onClosed={handleClose}
```
screendriver, line 105: Can the modal still be closed through the close button, backdrop or Escape while `isSubmitting` is true?

The asynchronous submission continues after `handleClose()` resets the form. If the user closes and reopens the modal before the first request completes, the completion handler from the previous submission can close the newly opened modal. It may also start a call after the user has apparently cancelled the flow.

Could we prevent closing while submission is active, or otherwise associate the completion with the submission/modal instance that initiated it?

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const {users} = useScheduleMeetingParticipants();
+  const portalContainer = getOverlayPortalContainer();
+
+  const contentViewModel = mainViewModel.content;
+  const conversationRepository = contentViewModel.repositories.conversation;
+  const searchRepository = contentViewModel.repositories.search;
+  const teamRepository = contentViewModel.repositories.team;
+
+  return (
+    <div css={meetNowFormLayoutStyles} data-uie-name="meet-now-form">
```
screendriver: Could this be implemented as a semantic `<form>` with an `onSubmit` handler instead of a `<div>` plus a click-only submit button?

The current implementation does not provide native form submission through Enter and splits validation/submission behaviour between the form and modal components. A real form would also give us better semantics without needing additional selectors.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
```diff
@@ -95,21 +92,56 @@ export const scheduleMeeting = (
...
+  return createMeetingAndSyncParticipants(mappingResult.value, formState.selectedUsers, deps).map(({failedToAdd}) => ({
+    failedToAdd,
+  }));
 };
 
+/**
+ * Creates an instant meeting and establishes the MLS conversation with selected participants.
+ */
+export const meetNowMeeting = (
+  formState: MeetNowFormState,
```
screendriver: Why does the service accept `MeetNowFormState` directly?

That type contains `participantsFilter`, which is presentation state and has no meaning to meeting creation. It also contains the raw, potentially empty title and relies on the modal having performed validation beforehand.

Could we validate and map the UI state into an explicit command such as `{title, selectedUsers}` before crossing into the service? The service should not depend on the internal state shape of a modal.

apps/webapp/src/script/components/Meeting/mapMeetNowCommandToCreateMeeting.ts
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {WallClock} from '@enormora/wall-clock/wall-clock';
+import type {CreateMeeting} from '@wireapp/api-client/lib/meetings/createMeeting';
+
+import type {MeetNowFormState} from 'Components/Meeting/MeetNowModal/meetNowTypes';
+import {getMeetNowMeetingTimes} from 'Components/Meeting/ScheduleMeetingModal/scheduleMeetingDefaults';
```
screendriver: This introduces dependencies between the internal implementations of two separate modal features. Meet Now imports scheduling defaults, participant hooks and styles from `ScheduleMeetingModal`, while `scheduleMeetingService.ts` now imports `MeetNowFormState`.

Could we extract the genuinely shared meeting creation, timing, participant and modal-style concerns into neutral modules under `Components/Meeting`? At the moment the two feature modules know about each other in both directions (also called cyclic dependencies), which makes their boundaries unclear.

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.ts
```diff
@@ -0,0 +1,159 @@
...
+
+export const useMeetNowSubmit = () => {
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const {translate} = useApplicationContext();
+  const {content, calling: callingViewModel} = useMainViewModel();
+  const {conversation: conversationRepository, calling: callingRepository} = content.repositories;
+  const meetNowMeeting = useMeetingStore(state => state.meetNowMeeting);
+  const loadMeetings = useMeetingStore(state => state.loadMeetings);
+
+  const callNotEstablishedCopy = useMemo(
```
screendriver: Could we remove this `useMemo()/useCallback()` cascade and separate the orchestration into a plain function with explicit dependencies?

The hook currently resolves dependencies from the application context, main view model, meeting store and global `tsyringe` container, while also handling translation, modal presentation, list refresh and joining. The successive memoization appears to exist mainly to stabilise other memoized values, but none of them crosses an evident memoization boundary.

A plain `startMeetNow(dependencies, input)` function returning a `Task` or `Result` would make the behaviour independently testable and leave the hook responsible only for React state.

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -106,6 +106,24 @@ describe('meetingSchema', () => {
     expect(meetingWithConversationSchema.safeParse(validMeetingWithConversation).success).toBe(true);
   });
 
+  it('accepts null add_permission on embedded meeting conversations', () => {
+    const result = meetingWithConversationSchema.safeParse({
+      ...validMeetingWithConversation,
+      conversation: {
+        ...validMeetingConversation,
```
screendriver, line 113: Is `add_permission: null` valid for all canonical conversation responses, or only for the conversation embedded in meeting create/update responses?

The added test demonstrates only the meeting payload, but this change broadens the shared `conversationSchema` used by every conversation consumer. If `null` is valid globally, could we add a contract test for the canonical conversation schema?

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStoreServiceTasks.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {
+  meetNowMeeting,
+  scheduleMeeting,
+  updateMeeting,
+} from 'Components/Meeting/ScheduleMeetingModal/scheduleMeetingService';
+
+import type {MeetingServiceDeps, MeetingStoreServiceTasks} from './meetingStoreDeps';
+
+export const createMeetingStoreServiceTasks = (deps: MeetingServiceDeps): MeetingStoreServiceTasks => ({
```
screendriver: What value does this separate factory provide beyond partially applying the same dependency object to three functions?

The store test creates a parallel fake `createServiceTasks()` implementation and therefore never exercises this production wiring. Could we either compose these functions directly at the composition root or add a focused test proving that this adapter delegates all three operations to the intended services with the supplied dependencies?

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
```diff
@@ -157,8 +155,18 @@ describe('createMeetingStore', () => {
 
     expect(result.isOk).toBe(true);
     expect(safeGetConversationById).toHaveBeenCalledWith(listMeetingInstance.meetingSeries.qualified_conversation);
-    expect(result.value.formState.start.unwrapOr(new Date(0))).toEqual(new Date('2026-06-16T10:00:00.000Z'));
-    expect(result.value.formState.end.unwrapOr(new Date(0))).toEqual(new Date('2026-06-16T11:00:00.000Z'));
+    expect(
```
screendriver: Could we assert the `Result` and `Maybe` variants **explicitly** instead of matching them to sentinel values?

The `Err: () => null` and `unwrapOr(new Date(0))` branches introduce fallback values into a test that is specifically expecting success. An explicit `assert(result.isOk)` followed by `assert(result.value.formState.start.isJust)` and normal Jest assertions would narrow the types and document the required invariants directly.

The same applies to the recurring-meeting assertions starting at line 195.

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowModal.ts
```diff
@@ -0,0 +1,99 @@
...
+  const errors: MeetNowFormErrors = {};
+
+  if (!title.trim()) {
+    errors.title = 'meetings.scheduleModal.error.titleRequired';
+  }
+
+  return errors;
+};
+
+export const hasMeetNowFormErrors = (errors: MeetNowFormErrors): boolean => Boolean(errors.title);
```
screendriver: Could we avoid converting a possibly absent string through truthiness here?

`Boolean(errors.title)` hides the actual condition we are testing. `errors.title !== undefined` is explicit and remains correct regardless of the string content.

The same applies to `!title.trim()` at line 57 and the truthiness check used to derive `titleError` in `MeetNowModal.tsx:54`.

I would also consider making the error property required as `TranslationKey | undefined`, or returning a `Result` from validation, instead of representing "no error" through an optional object property.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
screendriver: Do we need to introduce all of these new `data-uie-name` attributes?

The inputs and buttons already have semantic roles, labels and accessible names. None of the added unit tests uses these selectors. I would prefer tests to query the interface through those semantics rather than shipping implementation-specific test identifiers in production.

apps/webapp/src/script/components/Meeting/shared/service/meetingService.ts
```diff
@@ -63,29 +68,21 @@ const mapSyncErrorToSubmitError = (error: MeetingConversationSyncError): Meeting
...
-    return task.reject(mappingResult.error);
-  }
-
-  return deps.meetingsRepository
-    .createMeeting(mappingResult.value)
+const createMeetingAndSyncParticipants = (
+  createPayload: CreateMeeting,
+  selectedUsers: User[],
+  deps: MeetingServiceDeps,
+): Task<MeetNowSubmitSuccess, MeetingSubmitErrors> =>
```
screendriver: The supposedly shared `createMeetingAndSyncParticipants()` returns `MeetNowSubmitSuccessö, even though it is also used by `scheduleMeeting()`. I would rename that to something neutral such as `CreateMeetingSuccess`. The current type name shows that the abstraction is shaped around one caller.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {CSSObject} from '@emotion/react';
+
+export {
```
screendriver: The module acts as a re-export facade for another feature's styles. Besides the feature coupling, imports from `MeetNowModal.styles` misleadingly make those styles look locally owned. Extracting shared meeting modal styles would be clearer.

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/MeetNowModal/useMeetNowSubmit.ts
thisisamir98: Good point, thanks. I updated the hook to take `ConversationState` as a parameter. `MeetNowModal` now resolves it and passes it in, and the tests use a mock instead of the global container.

Fixed in https://github.com/wireapp/wire-webapp/commit/57fee6795e3b7b21508fb5df2ef15cb74869ebf3

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.test.tsx
thisisamir98: Good catch. I extended the hook tests to cover the join path and the failure cases.

The success test now checks that the returned `qualifiedConversation` is used for join. I also added cases for no internet, join failure, and conversation not found. They show that `submit()` returns `true` when creation succeeds, even if joining is blocked or fails afterward.

Fixed in https://github.com/wireapp/wire-webapp/commit/09af48e11afa38bbc4b6ba570648a5743c015039

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
thisisamir98: Thanks for the reminder. I checked all files introduced in this PR and the only ones that did not follow camelCase were in the Meet Now modal folder.

I renamed the directory to `meetNowModal` and updated the file names to `meetNowForm.tsx`, `meetNowModal.tsx`, and `meetNowModal.styles.ts`. All imports were updated as well.

Fixed in https://github.com/wireapp/wire-webapp/commit/55bd7db1d3ca157f9831e6a6e820ac4a33313213

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.ts
thisisamir98: Good point. I replaced the boolean with an explicit `MeetNowSubmitResult` that distinguishes `creationFailed`, `joined`, `joinBlocked`, and `joinFailed`. `submit()` now awaits the join step before it returns, so the result reflects the full flow.

The modal uses a `wasMeetNowMeetingCreated()` helper to decide when to close. It still closes when creation succeeds, even if joining is blocked or fails afterward.

Fixed in https://github.com/wireapp/wire-webapp/commit/a492d2461de21ec6d2313965f775544e91801a81

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.tsx
thisisamir98: Good catch. I addressed this in two ways.

Closing is now blocked while `isSubmitting` is true, so the close button, backdrop click, and Escape do nothing during an active submission. I also added a submit generation check, so a completion handler from an earlier submission cannot close a newly opened modal.

Fixed in https://github.com/wireapp/wire-webapp/commit/906e9e2593f69ab212e9560d3a3f5d78572641bc

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
thisisamir98: Good point, thanks. I updated `MeetNowForm` to use a semantic `<form>` with an `onSubmit` handler. The footer button now uses `type="submit"` and the `form` attribute, so Enter in the title field and the Start Meeting button share the same submit path.

https://github.com/wireapp/wire-webapp/commit/b6dafb4e74

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingService.ts
thisisamir98: Good point. I introduced explicit command types for both flows and map the modal form state into them before calling the service. `meetNowMeeting` and `scheduleMeeting` now take `{ title, selectedUsers }` and `{ title, start, end, recurrence, selectedUsers }` instead of the full form state, so presentation fields like `participantsFilter` stay in the UI layer.

https://github.com/wireapp/wire-webapp/commit/be166f750c

apps/webapp/src/script/components/Meeting/mapMeetNowCommandToCreateMeeting.ts
thisisamir98: Good point. I moved the shared pieces into `Components/Meeting/shared`, including defaults, form and modal styles, participants, command types, the meeting service, and submit helpers. Meet Now and Schedule Meeting now both depend on those shared modules, and they no longer import each other directly.

https://github.com/wireapp/wire-webapp/commit/1e171cc4a5

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowSubmit.ts
thisisamir98: Good point, thanks. I moved the orchestration into a plain `submitMeetNow` function with explicit dependencies, similar to `submitMeeting` in `useScheduleMeetingSubmit`. The hook now only wires React state and the no-internet guard, and I removed the `useMemo`/`useCallback` cascade.

https://github.com/wireapp/wire-webapp/commit/6e99522a3b

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: Good catch. I checked wire-server and `add_permission: null` is valid for canonical conversation responses in general, not only meeting payloads. The backend stores it as `Maybe AddPermission` and serializes `Nothing` as JSON null for all conversation objects.

I added direct contract tests on `conversationSchema` to document that behavior, including the null case and a non-null enum value.

https://github.com/wireapp/wire-webapp/commit/0f0a002068

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStoreServiceTasks.ts
thisisamir98: Fair point. The factory was mostly partial application, so I removed it and wired the three service operations directly in `Meetings.tsx`. The store still receives injected `serviceTasks` for tests, but the production wiring is now visible at the composition root.

https://github.com/wireapp/wire-webapp/commit/97722942de

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
thisisamir98: Good suggestion. I updated both edit-form tests to guard on `result.isOk`, assert `start` and `end` are `Just`, and read the values directly instead of using `match` fallbacks or `unwrapOr` sentinels.

https://github.com/wireapp/wire-webapp/commit/7441393cba

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowModal.ts
thisisamir98: Good point, thanks. I updated this in https://github.com/wireapp/wire-webapp/commit/5ae0ef7033.

`hasMeetNowFormErrors` and the title error display now check `errors.title !== undefined` instead of relying on truthiness. Error fields are required as `TranslationKey | undefined`, and `validateMeetNowForm` returns a `Result`. I kept `!title.trim()` as it is, since that validates the input value, not the presence of an error field.

I applied the same pattern to Schedule Meeting so both flows stay consistent.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowForm.tsx
thisisamir98: Fair point. None of the Meet Now `data-uie-name` attributes were used anywhere, so I removed them in https://github.com/wireapp/wire-webapp/commit/0a21375dbe.

The form and modal now rely on labels, roles, and ids instead. I also made `dataUieName` optional on `MeetingParticipantsPicker` so Meet Now does not need to pass it.

apps/webapp/src/script/components/Meeting/shared/service/meetingService.ts
thisisamir98: Good catch. I renamed `MeetNowSubmitSuccess` to `CreateMeetingSuccess` in https://github.com/wireapp/wire-webapp/commit/2dffe38108.

`createMeetingAndSyncParticipants()` now returns that neutral type. `scheduleMeeting()` still maps it down to `MeetingSubmitSuccess`, while `meetNowMeeting()` keeps the full result with `qualifiedConversation`.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.styles.ts
thisisamir98: Agreed. The shared shell styles already live in `meetingModalShell.styles.ts`, but Meet Now still re-exported them through `meetNowModal.styles.ts`.

I removed that facade in https://github.com/wireapp/wire-webapp/commit/551eea8c3f. `meetNowModal.tsx` now imports the shared styles directly, like Schedule Meeting, and `meetNowModal.styles.ts` only keeps Meet Now specific layout styles.

apps/webapp/src/script/components/Meeting/meetNowModal/submitMeetNow.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const submitResult = await meetNowMeeting(commandResult.value);
+
+  if (submitResult.isErr) {
+    if (shouldRefreshMeetingsListAfterSubmitError(submitResult.error)) {
+      await task.tryOrElse(() => meetingSubmitErrors.refreshFailed, loadMeetings);
+    }
+
+    showMeetingSubmitError(translate, submitResult.error);
+    return meetNowSubmitResults.creationFailed;
```
screendriver: `meetNowMeeting()` can reject after the backend meeting has already been created. For example, when saving or establishing its conversation fails.

Returning `creationFailed` here keeps the modal open because `wasMeetNowMeetingCreated()` interprets this as "nothing was created". Retrying can therefore create a duplicate meeting.

Could the result distinguish an actual creation-request failure from "meeting created, conversation setup failed", close the creation flow for the latter and cover that outcome with a regression test?

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
screendriver: `MeetingStoreState.scheduleMeeting()` now accepts a `ScheduleMeetingCommand`, but this test still passes the old form-state shape: `Maybe<Date>` values and `participantsFilter`.

Could we pass concrete `Date` values, remove `participantsFilter`, and assert that `scheduleMeeting` receives that exact command? The test currently no longer documents the production contract.

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
```diff
@@ -157,7 +155,14 @@ describe('createMeetingStore', () => {
 
     expect(result.isOk).toBe(true);
     expect(safeGetConversationById).toHaveBeenCalledWith(listMeetingInstance.meetingSeries.qualified_conversation);
+
+    if (!result.isOk) {
+      throw new Error('Expected loadMeetingForEdit to succeed');
+    }
+
+    expect(result.value.formState.start.isJust).toBe(true);
     expect(result.value.formState.start.unwrapOr(new Date(0))).toEqual(new Date('2026-06-16T10:00:00.000Z'));
```
screendriver: This still uses `unwrapOr(new Date(0))`, so the original sentinel-fallback concern remains. Checking `isJust` with Jest does not narrow the `Maybe`.

Could we use an explicit narrowing guard (`assert()`) and then read `.value` directly? Please apply the same change to `end` at line 166.

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
```diff
@@ -184,7 +189,14 @@ describe('createMeetingStore', () => {
     const result = await store.getState().loadMeetingForEdit(recurringMeetingInstance);
 
     expect(result.isOk).toBe(true);
+
+    if (!result.isOk) {
+      throw new Error('Expected loadMeetingForEdit to succeed');
+    }
+
+    expect(result.value.formState.start.isJust).toBe(true);
     expect(result.value.formState.start.unwrapOr(new Date(0))).toEqual(new Date('2026-06-22T10:00:00.000Z'));
```
screendriver: The recurring-meeting test still has the same unresolved sentinel fallback.

Could we narrow `start` and `end` explicitly (`assert()`) and then access their `.value` properties directly instead of using `unwrapOr(new Date(0))` here and at line 200?

apps/webapp/src/script/components/Meeting/shared/service/meetingService.ts
```diff
@@ -18,27 +18,35 @@
...
 } from 'Components/Meeting/meetingConversationSync';
-import type {MeetingStoreDeps} from 'Components/Meeting/meetingStore/meetingStoreDeps';
+import type {MeetingServiceDeps} from 'Components/Meeting/meetingStore/meetingStoreDeps';
 import {meetingSubmitErrors, type MeetingSubmitErrors} from 'Components/Meeting/MeetingSubmitErrors';
+import type {MeetNowMeetingCommand, ScheduleMeetingCommand} from 'Components/Meeting/shared/types/meetingCommandTypes';
 import type {User} from 'Repositories/entity/User';
 
-import type {ScheduleMeetingFormState, ScheduleMeetingRecurrenceOption} from './scheduleMeetingTypes';
+import type {
+  ScheduleMeetingFormState,
```
screendriver: This module is under `shared`, but it still imports `ScheduleMeetingFormState` and `ScheduleMeetingRecurrenceOption` from `ScheduleMeetingModal`.

That leaves a `shared` ➡️  `feature` dependency, which is the dependency direction the extraction was intended to remove. Could the update flow be mapped to a neutral command before entering `meetingService`, as the create flows now are?

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowModal.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const initialState = {
+  isOpen: false,
+  formState: getDefaultMeetNowFormState(),
+  errors: emptyMeetNowFormErrors(),
+};
+
+export const getMeetNowFormErrors = ({title}: MeetNowFormState): MeetNowFormErrors => ({
+  title: !title.trim() ? 'meetings.scheduleModal.error.titleRequired' : undefined,
```
screendriver: This still derives whether the title is empty through string truthiness.

Could we make the actual condition explicit as `title.trim().length === 0` (or something from `@sindresorhus/is`)? This was part of the original review comment.

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingValidation.ts
```diff
@@ -29,33 +30,46 @@ export interface ScheduleMeetingValidationInput {
...
+  const endInPast =
+    end.isJust && end.value.getTime() <= currentTimestampInMilliseconds
+      ? 'meetings.schedule.errors.endInPast'
+      : undefined;
 
-  if (!title.trim()) {
-    errors.title = 'meetings.scheduleModal.error.titleRequired';
-  }
+  return {
+    title: !title.trim() ? 'meetings.scheduleModal.error.titleRequired' : undefined,
```
screendriver: The same implicit string-truthiness check remains in Schedule Meeting.

Could we use `title.trim().length === 0` (or something from `@sindresorhus/is`) here as well so both validation paths express the condition explicitly and consistently?

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.tsx
```diff
@@ -0,0 +1,150 @@
...
+    if (hasMeetNowFormErrors(validationErrors)) {
+      return;
+    }
+
+    const submitGeneration = submitGenerationRef.current;
+
+    fireAndForgetInvoker.fireAndForget(async (): Promise<void> => {
+      const submitResult = await submit(formState);
+
+      if (submitGeneration !== submitGenerationRef.current) {
```
screendriver, line 90: Could we add a component-level regression test for this generation guard?

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/meetNowModal/submitMeetNow.ts
thisisamir98: Good catch, thanks for raising this. I split the outcomes so partial create failures no longer leave the modal open.

When the backend meeting exists but conversation setup fails, we now return `setupFailed` instead of `creationFailed`, and the modal closes. I also added `conversationSetupFailed` in the service layer so a failed local conversation save is not treated as a full create failure. Regression tests cover both partial failure paths.

Fixed in 4199723d21: https://github.com/wireapp/wire-webapp/commit/4199723d211ceedf12e7a274a41705b0eda9bd5f

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
thisisamir98: Good point, thanks. I updated the test to pass a real `ScheduleMeetingCommand` with concrete `Date` values, removed `participantsFilter`, and added an assertion that the store forwards that exact command to the service.

Fixed in 7ebb00bb2a: https://github.com/wireapp/wire-webapp/commit/7ebb00bb2aa15dba6fe7bac9d39b877f47c9ac7c

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
thisisamir98: Makes sense, thanks. I replaced the `unwrapOr(new Date(0))` checks with `assert(maybe.isJust(...))` and then read `.value` directly for both `start` and `end` in the edit form tests.

Fixed in 2474ee719d: https://github.com/wireapp/wire-webapp/commit/2474ee719d7fada2995b35fd7b8b313cde4bb2d5

apps/webapp/src/script/components/Meeting/shared/service/meetingService.ts
thisisamir98: Good call, thanks. I moved the update flow onto the same pattern as create.

`useScheduleMeetingSubmit` now maps the edit form to an `UpdateMeetingCommand` before calling the store, and `meetingService.updateMeeting` only accepts that command. Form mapping stays in the feature layer, and the shared service no longer imports from `ScheduleMeetingModal`.

Fixed in 8ca1c3e5c0: https://github.com/wireapp/wire-webapp/commit/8ca1c3e5c08ff0c852f6b0e334b56dec2b07a70c

apps/webapp/src/script/components/Meeting/ScheduleMeetingModal/scheduleMeetingValidation.ts
thisisamir98: Good point. I applied the same change in Schedule Meeting validation, so both flows now use `is.emptyString(title.trim())`.

Fixed in https://github.com/wireapp/wire-webapp/commit/2fd351e1bf

apps/webapp/src/script/components/Meeting/meetNowModal/useMeetNowModal.ts
thisisamir98: Good point, thanks. I updated the Meet Now validation to use `is.emptyString(title.trim())` instead of `!title.trim()`. This makes the empty title check explicit.

Fixed in https://github.com/wireapp/wire-webapp/commit/2fd351e1bf

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingStore.test.ts
thisisamir98: https://github.com/wireapp/wire-webapp/commit/2474ee719d

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.tsx
thisisamir98: Good point. I added a component test for the submit generation guard in `meetNowModal.test.tsx`. It submits the form, closes and reopens the modal before the queued submission runs, and checks that the stale callback does not close the new modal.

Fixed in https://github.com/wireapp/wire-webapp/commit/20dbeb03c2

apps/webapp/src/script/components/Meeting/shared/submit/shouldRefreshMeetingsListAfterSubmitError.ts
```diff
@@ -20,15 +20,21 @@
 import {meetingSubmitErrors, type MeetingSubmitErrors} from 'Components/Meeting/MeetingSubmitErrors';
 
 /**
- * Returns whether the meetings list should be refreshed after a failed submit.
- * Partial failures mean meeting metadata was already persisted and the UI should recover.
+ * Returns whether meeting metadata was already persisted before the submit failed.
+ * Partial failures should refresh the meetings list and must not invite a full retry.
```
screendriver, line 24: This is the intended semantic, but the Schedule Meeting caller does not currently enforce it.

`useScheduleMeetingSubmit()` refreshes the list for these persisted partial failures but then returns `false`. `ScheduleMeetingModal` interprets that as a failed submission and keeps the modal open. In create mode, retrying can therefore create a duplicate meeting even though the first meeting was already persisted.

The update path has a related problem: `meetingService.updateMeeting()` maps a failure while saving the returned conversation to `updateFailed`, even though the backend update has already succeeded. That result is not classified as persisted, so the list is not refreshed and the UI reports that the update itself failed.

Could the submission result explicitly distinguish:

* nothing was persisted;
* the meeting was persisted but subsequent setup failed;
* the complete operation succeeded?

The modal should close and the list should refresh for both persisted outcomes.

apps/webapp/src/script/components/Meeting/meetNowModal/submitMeetNow.ts
```diff
@@ -0,0 +1,167 @@
...
+  showCallNotEstablishedModal,
+  type NoInternetCallGuardCopy,
+} from 'Hooks/useNoInternetCallGuard/useNoInternetCallGuard';
+import {Config} from 'src/script/Config';
+import type {Translate} from 'Util/localizerUtil';
+
+import {meetNowSubmitResults, type MeetNowFormState, type MeetNowSubmitResult} from './meetNowTypes';
+
+const showMeetingSubmitError = (translate: Translate, error: MeetingSubmitErrors): void => {
+  const {titleKey, messageKey} = SCHEDULE_MEETING_ERROR_TRANSLATION_KEYS[error];
```
screendriver: Meet Now currently resolves its failures through `SCHEDULE_MEETING_ERROR_TRANSLATION_KEYS`.

This means a Meet Now creation failure can be presented as “Could not schedule meeting”. A `conversationSetupFailed` result is presented as “Could not update meeting” with a participant-related message, even when saving the conversation failed and no participants were selected.

Could we use neutral shared meeting-creation messages or introduce a Meet Now-specific error mapping so that the title and message describe both the actual operation and the actual failure?

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.test.tsx
```diff
@@ -0,0 +1,237 @@
...
+type DeferredFireAndForgetInvoker = FireAndForgetInvoker & {
+  runPending: () => Promise<void>;
+};
+
+const createDeferredFireAndForgetInvokerForTest = (): DeferredFireAndForgetInvoker => {
+  const pendingActions: Array<() => Promise<unknown>> = [];
+
+  return {
+    fireAndForget(asyncAction: () => Promise<unknown>): void {
+      pendingActions.push(asyncAction);
```
screendriver: This test double behaves differently from the production `FireAndForgetInvoker`.

The production implementation invokes `asyncAction()` immediately and tracks the returned promise. This fake queues the function itself without invoking it. The test can therefore submit and then close the modal while `isSubmitting` is still `false`, because the submission has not actually started.

Could the test use the production invoker and defer `meetNowMeeting()` or another dependency inside the submission instead?

It should verify that the close button, backdrop and Escape cannot dismiss the modal while the real submission promise is pending. If the generation guard is still needed in addition to that close prevention, the test should reproduce a production-realistic way for the generation to change while an earlier submission is running.

apps/webapp/src/script/components/Meeting/mapScheduleFormToMeetingCommand.ts
```diff
@@ -0,0 +1,76 @@
...
+  type ScheduleMeetingFormErrors,
+  type ScheduleMeetingFormState,
+} from 'Components/Meeting/ScheduleMeetingModal/scheduleMeetingTypes';
+import {validateScheduleMeetingForm} from 'Components/Meeting/ScheduleMeetingModal/scheduleMeetingValidation';
+import type {ScheduleMeetingCommand} from 'Components/Meeting/shared/types/meetingCommandTypes';
+
+const mapScheduleFormErrorToFormErrors = (error: ScheduleFormErrors): ScheduleMeetingFormErrors => {
+  switch (error) {
+    case scheduleFormErrors.missingTimes:
+      return {...emptyScheduleMeetingFormErrors(), endBeforeStart: 'meetings.scheduleModal.error.endBeforeStart'};
```
screendriver: `missingTimes` is not equivalent to `endBeforeStart`.

When either date is absent, this reports "End time must be after start time", which describes a different validation failure. The test currently codifies that misleading result.

Could `ScheduleMeetingFormErrors` represent missing times explicitly, or could the form make missing dates impossible before this mapper is called?

I would also avoid mapping the `default` branch to `endBeforeStart`. An exhaustive mapping would prevent a future error from silently producing unrelated validation copy.

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/shared/submit/shouldRefreshMeetingsListAfterSubmitError.ts
thisisamir98: Good catch, you were right about this.

Schedule Meeting was refreshing the list on partial failures, but it still treated them as a full failure for the modal. That could leave the dialog open and invite a duplicate create.

I pushed a fix in https://github.com/wireapp/wire-webapp/commit/6ab11b4e87. Schedule Meeting submit now returns three outcomes (`submitFailed`, `setupFailed`, `succeeded`), similar to Meet Now. The modal closes when the meeting was already persisted, and only stays open when nothing was saved.

For the update path, a failed local conversation save after a successful backend update now returns `conversationSetupFailed` instead of `updateFailed`, so the list refreshes and we no longer report the update itself as failed.

apps/webapp/src/script/components/Meeting/meetNowModal/submitMeetNow.ts
thisisamir98: Good point, thanks.

Meet Now was reusing the schedule/update error map, so the titles and messages did not always match what actually failed. I pushed a fix in https://github.com/wireapp/wire-webapp/commit/46832221d8.

Meet Now now uses its own error mapping (`MEET_NOW_ERROR_TRANSLATION_KEYS`), so create failures say "Could not start meeting" instead of "Could not schedule meeting". Shared setup failure strings now describe conversation setup separately from participant sync. Schedule Meeting also picks copy based on create vs edit mode, so create partial failures no longer show update wording.

apps/webapp/src/script/components/Meeting/meetNowModal/meetNowModal.test.tsx
thisisamir98: Good catch, thanks.

The queued fire-and-forget test double did not match production and let the modal close before submit had actually started. I pushed a fix in https://github.com/wireapp/wire-webapp/commit/d55811dde5.

The tests now use `createExecutingFireAndForgetInvokerForTest()` and defer `meetNowMeeting()` instead. There is also a new test that checks close, backdrop, and Escape while a real submit promise is pending. The generation guard test now closes and reopens in the same `act()` block before `isSubmitting` is visible, which matches the race the guard is meant to handle.

apps/webapp/src/script/components/Meeting/mapScheduleFormToMeetingCommand.ts
thisisamir98: Good catch, thanks.

I added an explicit `missingTimes` field to `ScheduleMeetingFormErrors`, validate missing dates in the form before mapping, and made the mapper switch exhaustive so we no longer fall back to `endBeforeStart`.

Fixed in https://github.com/wireapp/wire-webapp/commit/9afd4cdb25.

zskhan (approved):

PR #21890, opened 2026-07-17
refactor: refactor background renderer to support reusable segmentation masks [WPB-26366]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26366" title="WPB-26366" target="_blank">BugWPB-26366</a>  High CPU and GPU utilization when enabling Blur Background effect causes fan activation
  </td></table>
  <br />
 

## Summary

The goal is to make the rendering pipeline more flexible and allow future performance optimizations by reusing previously generated segmentation masks instead of running MediaPipe on every frame.

At the moment, this PR mainly introduces the required infrastructure. The quality/performance trade-offs of mask reuse are still being evaluated.

This PR refactors the background effects rendering pipeline to support rendering with either:

- newly generated segmentation masks,
- previously generated segmentation masks, or
- no segmentation (passthrough).

The renderer now exposes three dedicated rendering paths:

- `renderWithNewMasks()`
- `renderWithPreviousMask()`
- `renderPassthrough()`

This separation removes the previous conditional rendering logic and prepares the renderer for experiments where segmentation is not executed for every frame.

## Notes

Reusing segmentation masks can reduce segmentation workload, but early testing shows noticeable mask lag during movement. Further investigation is required before enabling this optimization by default.

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.test.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.test.ts
@@ -40,7 +40,9 @@ jest.mock('@enormora/wall-clock/wall-clock', () => ({
 
 jest.mock('./renderer', () => ({
   WebGLRenderer: jest.fn().mockImplementation(() => ({
-    render: jest.fn(),
+    renderWithNewMasks: jest.fn(),
+    renderWithPreviousMask: jest.fn(),
+    renderPassthrough: jest.fn(),
     close: jest.fn(),
   })),
 }));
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.test.ts
```diff
@@ -354,4 +356,229 @@ describe('segmenter tests', () => {
...
+      const options = {
+        enabled: true,
+        quality: 'auto',
+        enableFilters: false,
+        modelPath: 'model-a.tflite',
+        wasmLoaderPath: '/mock/vision_wasm_internal.js',
+        wasmBinaryPath: '/mock/vision_wasm_internal.wasm',
+      };
+
+      await runSegmenter(canvas, readable, options as any, jest.fn(), jest.fn());
```
zskhan, line 479: I will remove this as any in the next PR.

zskhan (approved):

PR #21897, opened 2026-07-17
Fix TC-548 and TC-553 e2e tests [WPB-27246]

iskvortsov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27246" title="WPB-27246" target="_blank">TaskWPB-27246</a>  [Web][QA] Fix folders e2e tests
  </td></table>
  <br />
 

https://wearezeta.atlassian.net/browse/WPB-27246

**TC-548** test is failing due to stale state condition and because test is running to fast. This condition cannot be reproduced manually and can be fixed by changing sequence of calls in precondition section of the code. 

**TC-553** was skipped before due to bug WPB-22266. But in reality it’s a e2e test implementation issue and stale state condition that could not be reproduced in real usage of the app.

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
@@ -102,7 +102,6 @@ test.describe('Folders', () => {
       await test.step('User A moves group conversation with User B into new custom folder', async () => {
         await userAPages.conversationList().getConversation(conversationName).open();
         await createCustomFolder(userAPageManager, conversationName, customFolderName);
-
       });
 
       await test.step('Group conversation with User B is in the custom folder', async () => {
--- a/apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
@@ -150,19 +150,23 @@ test.describe('Folders', () => {
       const customFolderName = 'Custom-Folder';
       const conversationName = 'Group Conversation with User A and User B';
 
-      // Preconditions: Create a custom folder with 1:1 conversation
-      await createCustomFolder(userAPageManager, userB.fullName, customFolderName);
-      await userAComponents.conversationSidebar().allConversationsButton.click();
-      await createGroup(userAPages, conversationName, [userB]);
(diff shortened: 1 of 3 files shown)


apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
```diff
@@ -146,35 +150,37 @@ test.describe('Folders', () => {
...
-      // Step 2: Group conversation with User B is in the custom folder
-      const actualTitle = userAPages.conversationList().conversationListHeaderTitle;
+      await test.step('User A moves group with User B into an existing custom folder', async () => {
+        await moveConversationToFolder(userAPageManager, conversationName, customFolderName);
+      });
 
-      await expect(actualTitle).toHaveText(customFolderName);
+      await test.step('Group conversation with User B is in the custom folder', async () => {
+        const actualTitle = userAPages.conversationList().conversationListHeaderTitle;
+        await expect(userAComponents.conversationSidebar().folderList.getByText(customFolderName)).toBeVisible();
```
screendriver: This wait happens after `moveConversationToFolder()`, although that operation already depends on the folder being fully created. `createCustomFolder()` only clicks the modal action and returns, so changing the sequence currently adds incidental delay rather than explicitly synchronizing the state. Could we wait for the folder immediately after `createCustomFolder()`; ideally inside that helper; before navigating away and creating the group?

screendriver (changes requested):

apps/webapp/test/e2e_tests/specs/Folders/folders.spec.ts
iskvortsov: Yep, that actually makes more sense.

iskvortsov: Updated the logic and added another check for custom folder to contain both conversation.

zskhan (dismissed):

zskhan (dismissed):

zskhan (dismissed):

screendriver (changes requested):
The whole Git commit history is broken now

screendriver (approved):

zskhan (approved):

PR #21898, opened 2026-07-17
fix: return OR results for multiple tags in Drive filter (WPB-27126)

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27126" title="WPB-27126" target="_blank">BugWPB-27126</a>  [Web] Selecting multiple tags returns no results instead of matching files with any selected tag
  </td></table>
  <br />
 

### What

Multi-tag search in the Cells API previously collapsed all selected tags into a single JSON-encoded term (e.g. `"sam,bug"`), which the backend treated as one literal tag value that no node had  so any filter with 2+ tags returned zero results.

This PR emits one `Metadata` entry per tag with `Operation: 'Should'` and a raw term, mirroring the iOS client (`WireMessaging/Sources/WireMessagingData/WireDrive/NodesAPI/RestAPI.swift:643`) and the existing `mimeTypes` / `creatorIds` pattern in the same file.

Tag filter construction is extracted into a small module-level helper `createTagMetadataFilters` with an explicit `LookupFilterMetaFilter[]` return type so the request literal reads by intent rather than by construction detail.

### Why

Restores the OR semantics defined in the PRD: a file matching any of the selected tags is returned.

### Behaviour change

- **Multiple tags**: previously returned 0; now returns files matching any selected tag.
- **Single tag**: the returned files are unchanged. The on-wire shape does change from a single JSON-joined term to a per-tag `Should` entry with a raw term. Verified against the fulu backend that a single-tag lookup with the new shape returns identical results to the old one.

### Demo
https://github.com/user-attachments/assets/0083f26e-43a0-4d0e-bfed-c8221ffddf94

### DIFF
--- a/libraries/api-client/src/cells/cellsApi.test.ts
+++ b/libraries/api-client/src/cells/cellsApi.test.ts
@@ -1914,9 +1914,9 @@ describe('CellsAPI', () => {
       );
     });
 
-    it('filters by tags when provided', async () => {
+    it('filters by a single tag with Should operation', async () => {
       const searchPhrase = 'test';
-      const tags = ['tag1', 'tag2'];
+      const tags = ['tag1'];
       const mockResponse: RestNodeCollection = {
         Nodes: [
           {
@@ -1938,13 +1938,41 @@ describe('CellsAPI', () => {
           Status: {
             Deleted: 'Not',
           },
-          Metadata: [{Namespace: 'usermeta-tags', Term: JSON.stringify(tags.join(','))}],
+          Metadata: [{Namespace: 'usermeta-tags', Term: 'tag1', Operation: 'Should'}],
         },
         Flags: ['WithPreSignedURLs'],
         Limit: '10',
         Offset: '0',
       });
       expect(result).toEqual(mockResponse);
+    });
+
+    it('filters by multiple tags with Should operation (OR) so any selected tag matches (WPB-27126)', async () => {
+      const searchPhrase = 'test';
+      const tags = ['tag1', 'tag2'];
+      const mockResponse: RestNodeCollection = {Nodes: []} as RestNodeCollection;
+
+      mockNodeServiceApi.lookup.mockResolvedValueOnce(createMockResponse(mockResponse));
+
+      await cellsAPI.searchNodes({phrase: searchPhrase, tags});
+
+      expect(mockNodeServiceApi.lookup).toHaveBeenCalledWith({
+        Scope: {Root: {Path: '/'}, Recursive: true},
+        Filters: {
+          Text: {SearchIn: 'BaseName', Term: searchPhrase},
+          Type: 'UNKNOWN',
+          Status: {
+            Deleted: 'Not',
+          },
+          Metadata: [
+            {Namespace: 'usermeta-tags', Term: 'tag1', Operation: 'Should'},
+            {Namespace: 'usermeta-tags', Term: 'tag2', Operation: 'Should'},
+          ],
+        },
+        Flags: ['WithPreSignedURLs'],
+        Limit: '10',
+        Offset: '0',
+      });
     });
 
     it('handles empty tags array', async () => {
@@ -2249,7 +2277,7 @@ describe('CellsAPI', () => {
           Type: 'UNKNOWN',
           Status: {Deleted: 'Not', HasPublicLink: true},
           Metadata: [
-            {Namespace: 'usermeta-tags', Term: JSON.stringify(tags.join(','))},
+            {Namespace: 'usermeta-tags', Term: 'important', Operation: 'Should'},
             {Namespace: 'mime', Term: 'image/*', Operation: 'Must'},
             {Namespace: 'usermeta-owner-uuid', Term: JSON.stringify(creatorId), Operation: 'Must'},
           ],
--- a/libraries/api-client/src/cells/cellsApi.test.ts
+++ b/libraries/api-client/src/cells/cellsApi.test.ts
@@ -1947,7 +1947,7 @@ describe('CellsAPI', () => {
       expect(result).toEqual(mockResponse);
     });
 
-    it('filters by multiple tags with Should operation (OR) so any selected tag matches (WPB-27126)', async () => {
+    it('creates one Should metadata filter per selected tag', async () => {
       const searchPhrase = 'test';
       const tags = ['tag1', 'tag2'];
       const mockResponse: RestNodeCollection = {Nodes: []} as RestNodeCollection;
--- a/libraries/api-client/src/cells/cellsApi.test.ts
+++ b/libraries/api-client/src/cells/cellsApi.test.ts
@@ -1914,7 +1914,7 @@ describe('CellsAPI', () => {
       );
     });
 
-    it('filters by a single tag with Should operation', async () => {
+    it('creates a Should metadata filter for a single tag', async () => {
       const searchPhrase = 'test';
       const tags = ['tag1'];
       const mockResponse: RestNodeCollection = {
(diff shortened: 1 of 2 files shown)


libraries/api-client/src/cells/cellsApi.ts
```diff
@@ -468,9 +468,11 @@ export class CellsAPI {
           ...(hasPublicLink !== undefined ? {HasPublicLink: hasPublicLink} : {}),
         },
         Metadata: [
-          ...(tags !== undefined && tags.length > 0
-            ? [{Namespace: USER_META_TAGS_NAMESPACE, Term: this.transformTagsToJson(tags)}]
-            : []),
+          ...(tags?.map(tag => ({
```
screendriver: Could we extract the metadata-filter construction from this request literal? This expression currently combines optional-value handling, collection mapping, API serialization, query semantics, and a type assertion. Together with the MIME type and creator filters below, the request is becoming difficult to scan.

Named helpers such as `createTagMetadataFilters`, `createMimeTypeMetadataFilters`, and `createCreatorMetadataFilters` would let this request describe what it contains rather than how every filter is constructed. An explicit helper return type should also remove the need for `as const`.

libraries/api-client/src/cells/cellsApi.test.ts
```diff
@@ -1947,6 +1947,34 @@ describe('CellsAPI', () => {
       expect(result).toEqual(mockResponse);
     });
 
+    it('filters by multiple tags with Should operation (OR) so any selected tag matches (WPB-27126)', async () => {
```
screendriver: This test does not verify that "any selected tag matches." The API is mocked with an empty response, so it only verifies how the lookup request is serialized.

Either rename this to describe what it actually tests, for example, `creates one Should metadata filter for each selected tag`, or add an integration/E2E regression test that selects multiple tags and verifies that a file matching either tag is returned.

libraries/api-client/src/cells/cellsApi.test.ts
```diff
@@ -1938,7 +1938,7 @@ describe('CellsAPI', () => {
           Status: {
             Deleted: 'Not',
           },
-          Metadata: [{Namespace: 'usermeta-tags', Term: JSON.stringify(tags.join(','))}],
+          Metadata: [{Namespace: 'usermeta-tags', Term: 'tag1', Operation: 'Should'}],
```
screendriver, line 1941: The pull request description says single-tag search is unaffected. But the actual request changes from:

```
{
  Namespace: USER_META_TAGS_NAMESPACE,
  Term: this.transformTagsToJson(tags),
}
```

to

```{
  Namespace: USER_META_TAGS_NAMESPACE,
  Term: tag,
  Operation: 'Should',
}
```

That changes both the term representation and the explicit operation.

screendriver (commented):
This pull request targets `dev`, but our development branch is now `main`. `dev` is currently 169 commits behind main. Please retarget the pull request to `main` and update the feature branch from `main` before continuing the review.

cristianoliveira:
@screendriver I appreciate the early feedback, but as of now this is a Draft PR, I'll take into account the concerns you raised

libraries/api-client/src/cells/cellsApi.ts
cristianoliveira: Good call. Extracted `createTagMetadataFilters` as a module-level helper with an explicit `LookupFilterMetaFilter[]` return type, which also drops the `as const`.

Kept the mime/creator extraction out of this PR to keep the diff scoped to WPB-27126 (both were untouched by the bugfix and I'd rather not modify unrelated code paths under a fix commit). Happy to open a follow-up cleanup PR that applies the same pattern to `createMimeTypeMetadataFilters` and `createCreatorMetadataFilters`, let me know if you'd prefer that as part of this one instead.

libraries/api-client/src/cells/cellsApi.test.ts
cristianoliveira: I changed the description, the old name was promising something the test can't actually see 

The OR part I verified against the real backend (fulu) I added a demo in the PR description

libraries/api-client/src/cells/cellsApi.test.ts
cristianoliveira: You're right, "unaffected" was too loose. What I meant is that a single-tag search still returns only the tagged files with the given tag (See the demo)

libraries/api-client/src/cells/cellsApi.ts
```diff
@@ -55,6 +56,17 @@ const USER_META_TAGS_NAMESPACE = 'usermeta-tags';
 const USER_META_OWNER_UUID_NAMESPACE = 'usermeta-owner-uuid';
 const MIME_NAMESPACE = 'mime';
 
+// Each selected tag is sent as its own metadata filter with the `Should` operation so the
+// backend applies OR semantics across tags (a node matching any selected tag is returned).
+// Matches the iOS client shape in WireMessaging RestAPI.swift.
+function createTagMetadataFilters(tags: string[] | undefined): LookupFilterMetaFilter[] {
```
screendriver: Could we keep the optional-value handling outside this function?

`createTagMetadataFilters` is responsible for transforming tags into metadata filters. Accepting `undefined` and deciding that it means an empty collection mixes input normalization with that transformation logic and weakens the function's contract.

Please let this function accept only `string[]` and let its caller pass `tags ?? []`.

libraries/api-client/src/cells/cellsApi.test.ts
```diff
@@ -1914,9 +1914,9 @@ describe('CellsAPI', () => {
       );
     });
 
-    it('filters by tags when provided', async () => {
+    it('filters by a single tag with Should operation', async () => {
```
screendriver: This test does not verify that the API actually filters by the tag. The mocked response is returned independently of the request, so the observable behaviour is only the construction of the lookup request.

Could we rename this to describe what the test verifies, for example:

`creates a Should metadata filter for a single tag`

libraries/api-client/src/cells/cellsApi.test.ts
```diff
@@ -2249,7 +2277,7 @@ describe('CellsAPI', () => {
           Type: 'UNKNOWN',
           Status: {Deleted: 'Not', HasPublicLink: true},
           Metadata: [
-            {Namespace: 'usermeta-tags', Term: JSON.stringify(tags.join(','))},
+            {Namespace: 'usermeta-tags', Term: 'important', Operation: 'Should'},
```
screendriver, line 2280: This test only verifies the serialized request shape. It does not verify that the selected tag still restricts the backend result when it is combined with MIME and creator filters.

Because the tag uses `Should` while the other metadata filters use `Must`, could we add a backend-level regression test for the combined case? It should verify that a file matching the MIME type and creator, but none of the selected tags, is not returned.

libraries/api-client/src/cells/cellsApi.ts
```diff
@@ -468,9 +480,7 @@ export class CellsAPI {
           ...(hasPublicLink !== undefined ? {HasPublicLink: hasPublicLink} : {}),
         },
         Metadata: [
-          ...(tags !== undefined && tags.length > 0
-            ? [{Namespace: USER_META_TAGS_NAMESPACE, Term: this.transformTagsToJson(tags)}]
-            : []),
+          ...createTagMetadataFilters(tags),
```
screendriver: Could we normalize the optional value before constructing the request and keep `createTagMetadataFilters` focused only on transforming tags?

For example:

```typescript
const tagMetadataFilters = createTagMetadataFilters(tags ?? []);
```

The function can then accept only `string[]`, and the request can compose the resulting filters with:

```typescript
Metadata: [
  ...tagMetadataFilters,
  // ...
],
```

screendriver (changes requested):

libraries/api-client/src/cells/cellsApi.test.ts
cristianoliveira: I created a follow up task to add this test case into cells-integration-tests https://wearezeta.atlassian.net/browse/WPB-27344

cristianoliveira:
Accidentally rebased with a stale main branch, but now git history is back on track

screendriver (approved):

PR #21912, opened 2026-07-20
feat: sort shared drive tags alphabetically in picker [WPB-21608]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-21608" title="WPB-21608" target="_blank">TaskWPB-21608</a>  [Web] Display tags by alphabetical order
  </td></table>
  <br />
 

## What changed?

Shared drive tags are now sorted alphabetically, ignoring case, in:

- the list of existing tags in the tag picker
- conversation Files
- All Files

Tag creation and update behavior is unchanged. This only affects display order.

## Jira

[WPB-21608](https://wearezeta.atlassian.net/browse/WPB-21608)

[WPB-21608]: https://wearezeta.atlassian.net/browse/WPB-21608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Demo

https://github.com/user-attachments/assets/a0a42907-de4d-41bf-b3f7-83dc9cfceafe


apps/webapp/src/script/components/Conversation/ConversationCells/CellsTable/CellsTableColumns/CellsTableRowOptions/CellsTagsModal/useTagsManagement/useTagsManagement.ts
```diff
@@ -70,7 +71,7 @@ export const useTagsManagement = ({
   const allTags = useMemo(() => {
     const allTagNames = [...tagNames, ...createdTags];
     const uniqueTagNames = Array.from(new Set(allTagNames));
-    return uniqueTagNames.map(transformTagToSelectOption);
+    return sortTagsAlphabetically(uniqueTagNames).map(transformTagToSelectOption);
```
screendriver: Does this actually keep newly created tags sorted in the picker?

This sorts `allTags`, which is used as the available options list. However, `handleCreateOption()` still appends the new option to `selectedTags`, and `selectedTags` is passed to the select as its current value.

For example, when `Alpha` and `Zulu` are selected, creating `Beta` still produces the visible selected-tag order `Alpha, Zulu, Beta`.

Please either sort `selectedTags` as well or clarify that the requirement applies only to the available options. The test should cover the intended selected-tag order.

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/transformDataToCellsNodes.ts
```diff
@@ -136,5 +137,6 @@ const getTags = (node: RestNode): string[] => {
   }
 
   const parsedTags = JSON.parse(tags);
-  return parsedTags.split(',').map((tag: string) => tag.trim());
+  const tagNames = parsedTags.split(',').map((tag: string) => tag.trim());
+  return sortTagsAlphabetically(tagNames);
```
screendriver: Why do we sort the tags while transforming the backend response into the `CellNode` model?

`node.tags` is later passed to the edit modal as `initialSelectedTags`. Those values can then be written back through `setNodeTags()` when the user saves.

This means that opening and saving the modal can change stored tag metadata from, for example, `Zulu, Alpha` to `Alpha, Zulu`, even when the user did not change the tags. That contradicts the PR description that update behaviour is unchanged and that this only affects display order.

Please preserve the backend order in the model and sort at the actual presentation boundaries instead. The same applies to `transformCellsNodes()`.

apps/webapp/src/script/components/Conversation/ConversationCells/common/sortTagsAlphabetically/sortTagsAlphabetically.ts
```diff
@@ -0,0 +1,21 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+export const sortTagsAlphabetically = (tags: string[]): string[] =>
+  tags.toSorted((tagA, tagB) => tagA.localeCompare(tagB, undefined, {sensitivity: 'base'}));
```
screendriver: `sensitivity: 'base'` ignores accents as well as letter case, while the requirement only says that sorting should ignore case. Passing `undefined` as the locale also makes the result depend on the browser/runtime locale rather than Wire's selected application locale.

Is that intended?

Please define the sorting contract explicitly. If we only want case-insensitive sorting while preserving accent differences, `sensitivity: 'accent'` would be more accurate. We should also decide whether ordering follows the Wire locale or a fixed locale and add coverage with accented and other non-ASCII tag names.

apps/webapp/src/script/components/cellsGlobalView/cellsHeader/cellsFilters/cellsFiltersMenu/useGetAllTags/useGetAllTags.ts
```diff
@@ -45,7 +47,7 @@ export const useGetAllTags = (properties: UseGetAllTagsProps): UseGetAllTagsResu
     try {
       const tags = await cellsRepository.getAllTags();
       const filteredTags = tags.Values?.map(tag => tag).filter(Boolean) ?? [];
-      setTags(filteredTags);
+      setTags(sortTagsAlphabetically(filteredTags));
```
screendriver: This path still normalizes the response using `filter(Boolean)`, so a whitespace-only tag such as `" "` survives and can appear as an apparently blank option.

The centralized tags store already removes these values using `tagName.trim() !== ''`. Could we reuse the same validation and normalization instead of maintaining two slightly different implementations?

Please also add a test containing an empty string and a whitespace-only string.

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/common/sortTagsAlphabetically/sortTagsAlphabetically.ts
cristianoliveira: Fair, I checked Android’s implementation for the linked WPB-21606 task. It uses `tags.sortedBy { it.lowercase() }`, so the effective contract is case-insensitive ordering https://github.com/wireapp/wire-android/blob/9e50c4bb2016621c2b83d7d315dec4b4730c9bab/features/cells/src/main/java/com/wire/android/feature/cells/ui/tags/AddRemoveTagsViewModel.kt#L58

I think Web should follow the same contract for consistency across clients. I’ll replace the runtime-locale collator with lowercase lexical comparison and keep coverage for accented and non-ASCII tags

apps/webapp/src/script/components/cellsGlobalView/cellsHeader/cellsFilters/cellsFiltersMenu/useGetAllTags/useGetAllTags.ts
cristianoliveira: Done

apps/webapp/src/script/components/Conversation/ConversationCells/common/sortTagsAlphabetically/sortTagsAlphabetically.ts
cristianoliveira: Changed the approach in [3643c21](https://github.com/wireapp/wire-webapp/pull/21912/commits/3643c213616cc16a2ca8a79722082da72860788d)

apps/webapp/src/script/components/Conversation/ConversationCells/useGetAllCellsNodes/transformDataToCellsNodes.ts
cristianoliveira: ✅

apps/webapp/src/script/components/Conversation/ConversationCells/CellsTable/CellsTableColumns/CellsTableRowOptions/CellsTagsModal/useTagsManagement/useTagsManagement.ts
cristianoliveira: I checked the Jira ACs and Android implementation. The requirement is to sort the list of existing tags, without changing the creation or update flow

 Android does the same thing: it sorts fetched allTags, but keeps newly added and selected tags in insertion order

 I clarified this in the PR description and added regression coverage in 0c3da6cd62

screendriver (approved):

PR #21913, opened 2026-07-20
Add enhanced background performance checkbox - [WPB-27282]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27282" title="WPB-27282" target="_blank">TaskWPB-27282</a>  Checkbox "Enhance Performance" for the better Performance from user. 
  </td></table>
  <br />
 

Add the enhance-performance setting to background effects, including UI state propagation and reduced segmentation cadence. Update video rendering behavior and add typed regression coverage for quality-tier intervals.


apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts
```diff
@@ -70,6 +72,7 @@ export const backgroundEffectsStore = createStore<BackgroundEffectsState>()(
     model: 'unknown',
     lastVirtualBackgroundId: DEFAULT_BUILTIN_BACKGROUND_ID,
     isHighQualityBlurEnabled: true,
+    isPerformanceEnhancementEnabled: false,
```
EnricoSchw: This should be `true` by default

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts
```diff
@@ -58,6 +59,7 @@ export type BackgroundEffectsState = {
   setMetrics(metrics: RenderMetrics | undefined): void;
   setModel(model: string | undefined): void;
   setIsHighQualityBlurEnabled(value: boolean): void;
```
EnricoSchw: Perhaps we should rename `isHighQualityBlurEnabled` to is `isEnhancementLLMEnabled`. After all, it essentially means allowing the multiclass model or the enhanced model.

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.test.ts
```diff
@@ -23,8 +23,9 @@ describe('backgroundEffectsStore:lastVirtualBackgroundId', () => {
     expect(typeof state.lastVirtualBackgroundId).toBe('string');
   });
 
-  it('initializes high quality blur as enabled', () => {
+  it('initializes background effects with privacy quality', () => {
     expect(backgroundEffectsStore.getState().isHighQualityBlurEnabled).toBe(true);
+    expect(backgroundEffectsStore.getState().isPerformanceEnhancementEnabled).toBe(false);
   });
```
EnricoSchw, line 32: If we do not merge the combined state of `isHighQualityBlurEnabled` and `isPerformanceEnhancementEnabled` into a single state like:

```typescript
type BackgroundEffectsQuality =
  | 'privacy'
  | 'balanced'
  | 'performance';
```

, we should at least explicitly forbid the disallowed state via a test.

```
privacy --> isHighQualityBlurEnabled == true && isPerformanceEnhancementEnabled == false
balanced -->  isHighQualityBlurEnabled == false && isPerformanceEnhancementEnabled == false
performance -->  isHighQualityBlurEnabled == false && isPerformanceEnhancementEnabled == true
```

This state does not exist:
`xx -->  isHighQualityBlurEnabled == true && isPerformanceEnhancementEnabled == true`


PS: And why "privacy"? Shouldn't it be called "best quality"?

EnricoSchw (commented):
LGTM but some questions

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts
zskhan: By default "Privacy" should be selected and combination for that is 
```isHighQualityBlurEnabled: true,```
```isPerformanceEnhancementEnabled: false,```

Comment (EnricoSchw) [reply apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts]: That's right, I misinterpreted the variable.

File: apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.test.ts (hunk 3 above)
Comment (zskhan) [reply apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.test.ts:32]: > xx -->  isHighQualityBlurEnabled == true && isPerformanceEnhancementEnabled == true

Thank you for the suggestion. I will write it down.

> And why "privacy"? Shouldn't it be called "best quality"?

I just wanted to make it same as the wording on the UI.

Comment (zskhan) [reply apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.test.ts:32]: Thank you @EnricoSchw for the suggestion.
I merged both the boolean to avoid complexity. Single boolean of `qualityTier` is more reasonable here now.

**Review by EnricoSchw — DISMISSED** [review]

File: apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts [hunk 4]
```diff
@@ -270,9 +270,10 @@ export class BackgroundEffectsHandler {
     return this.controller.getQuality();
   }
 
-  public enableSuperhighQualityTier(enable: boolean): void {
-    this.controller.setModelPath(enable ? SELFIE_MULTICLASS_MODEL_PATH : SELFIE_SEGMENTER_MODEL_PATH);
-    backgroundEffectsStore.getState().setIsHighQualityBlurEnabled(enable);
+  public setQualityTier(quality: BackgroundEffectsQuality): void {
+    this.controller.setModelPath(quality === 'privacy' ? SELFIE_MULTICLASS_MODEL_PATH : SELFIE_SEGMENTER_MODEL_PATH);
+    this.controller.setEnhancePerformance(quality === 'performance');
+    backgroundEffectsStore.getState().setQualityTier(quality);
```
thisisamir98, line 275: Could we also apply the stored `qualityTier` when the pipeline starts?

This updates the live controller, but `applyBackgroundEffect` still starts from `defaultOpts` (multiclass + `enhancePerformance: false`) and never reads the store. So if the user picks Performance/Balance and processing later restarts (new call, camera change, effect toggle), the radio can still show their choice while the worker runs Privacy without frame skipping.

Mapping the store into `modelPath` / `enhancePerformance` in `applyBackgroundEffect` (or calling `setQualityTier` right after `start`) would keep UI and pipeline aligned.

Regression worth locking: select `performance` → stop processing → `applyBackgroundEffect` again → assert segmenter model + `enhancePerformance: true`, and store still `performance`.

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts
```diff
@@ -108,9 +111,16 @@ export const backgroundEffectsStore = createStore<BackgroundEffectsState>()(
...
-        state.isHighQualityBlurEnabled = value;
+        state.qualityTier = tier;
+      }),
+
+    setEnhancedModelActive: isActive =>
+      set(state => {
+        if (!isActive && state.qualityTier === 'privacy') {
+          state.qualityTier = 'balanced';
+        }
       }),
```
thisisamir98, line 124: I would prefer not to overwrite the user's quality choice from temporary adaptive capacity here.

When adaptive quality temporarily swaps multiclass → segmenter, this turns Privacy into Balance. When the high tier comes back, `setEnhancedModelActive(true)` does nothing, so the radio can stay on Balance while multiclass runs again. That treats a temporary condition as a durable preference change.

Could we keep `qualityTier` as the user preference only, and track “enhanced model currently active” separately if the UI needs it?

Regression: `qualityTier === 'privacy'` → effective model temporarily segmenter → store still `privacy` → model recovers to multiclass → store still `privacy`.

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -310,7 +315,7 @@ export class BackgroundEffectsController {
 
     this.options.modelPath = effectiveModelPath;
 
-    backgroundEffectsStore.getState().setIsHighQualityBlurEnabled(effectiveModelPath === SELFIE_MULTICLASS_MODEL_PATH);
+    backgroundEffectsStore.getState().setEnhancedModelActive(effectiveModelPath === SELFIE_MULTICLASS_MODEL_PATH);
```
thisisamir98: Same preference concern from the caller side: this feeds the effective model into `setEnhancedModelActive`, which can rewrite the radio selection during adaptive tier changes.

If we keep preference and effective state separate, this call (or its store effect) should not mutate `qualityTier`.

apps/webapp/src/script/components/Radio/RadioGroup.tsx
```diff
@@ -67,33 +69,41 @@ const RadioGroup = <T extends string | number>({
...
-              type="radio"
-              id={currentId}
-              name={name}
-              value={value}
-              onChange={() => onChange(value)}
-              checked={isChecked}
-              data-uie-name={optionUeiName}
-            />
+          return (
+            <div key={value} css={radioOptionStyles} aria-describedby={currentId}>
```
thisisamir98: Small a11y gap: the new description is rendered below, but `aria-describedby` still points at the radio's own `id`, and the description node has no `id`.

Could we give the description an id and point `aria-describedby` at that, so screen readers get the quality help text?

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.test.ts
```diff
@@ -23,21 +23,47 @@ describe('backgroundEffectsStore:lastVirtualBackgroundId', () => {
...
-    expect(backgroundEffectsStore.getState().isHighQualityBlurEnabled).toBe(true);
+    expect(backgroundEffectsStore.getState().qualityTier).toBe('privacy');
+  });
+
+  it('never allows the enhanced model and performance enhancement to be active at the same time', () => {
+    backgroundEffectsStore.getState().setQualityTier('performance');
+
+    backgroundEffectsStore.getState().setEnhancedModelActive(true);
+
+    expect(backgroundEffectsStore.getState().qualityTier).toBe('performance');
```
thisisamir98, line 68: This test name promises that enhanced model and performance mode cannot be active together, but it only checks that `setEnhancedModelActive(true)` leaves `qualityTier` as `performance`.

That can hide the sync hole above (UI says Performance while multiclass is running). Could we assert the real invariant through the handler/controller instead, or rewrite this once preference and effective model are separated?

apps/webapp/src/i18n/en-US.json
```diff
@@ -2165,8 +2165,16 @@
...
+  "videoCallBackgroundEffectsQualityBalanced": "Balance",
+  "videoCallBackgroundEffectsQualityBalancedDescription": "Privacy meets performance",
+  "videoCallBackgroundEffectsQualityPerformance": "Device performance",
+  "videoCallBackgroundEffectsQualityPerformanceDescription": "Needs fewer resources, but may reduce background quality.",
+  "videoCallBackgroundEffectsQualityPrivacy": "Privacy",
+  "videoCallBackgroundEffectsQualityPrivacyDescription": "Improves outline and movement definition",
+  "videoCallBackgroundEffectsQualitySectionLabel": "Optimized for",
   "videoCallBackgroundEnableEnhancedQuality": "Enhanced quality",
   "videoCallBackgroundEnableEnhancedQualityHint": "Improves outline definition, but may impact the performance on your device.",
+  "videoCallBackgroundEnhancePerformance": "Enhance Performance",
```
thisisamir98: Nit: `videoCallBackgroundEnableEnhancedQuality` / `Hint` and `videoCallBackgroundEnhancePerformance` look unused after the radio rewrite. Can we drop them if nothing else references them?

thisisamir98 (changes requested):
## Summary
Nice direction overall: collapsing the two toggles into one quality choice and wiring `enhancePerformance` through the segmenter is much clearer.

I found two runtime issues around preference vs temporary/adaptive state, plus a couple of smaller follow-ups. Details are inline.

## Questions
- Should this quality choice persist across sessions (like preferred effect), or is in-memory + Privacy on load intentional?
- After an adaptive downgrade, should the UI keep showing Privacy (user preference), or reflect the temporary effective model?

apps/webapp/src/script/repositories/media/useBackgroundEffectsStore.ts
zskhan: https://github.com/wireapp/wire-webapp/pull/21913/changes/a1400bc8cab2f7006218bb8c750dd994be80a3f5

apps/webapp/src/script/repositories/media/backgroundEffectsHandler.ts
zskhan: https://github.com/wireapp/wire-webapp/pull/21913/changes/d03a20cd0b93fd62a3315bf2556485443ddc6fca

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
zskhan: User selected Quality and effected Quality are now separated in zustand store

apps/webapp/src/script/components/calling/FullscreenVideoCall.tsx
```diff
@@ -303,16 +303,16 @@ const FullscreenVideoCall = ({
   const isWebGLAvailable = detectCapabilities().webgl2;
 
   const selectedBackgroundEffect = useBackgroundEffectsStore(state => state.preferredEffect);
-  const isHighQualityBlurEnabled = useBackgroundEffectsStore(state => state.isHighQualityBlurEnabled);
+  const backgroundEffectsQuality = useBackgroundEffectsStore(state => state.effectiveQualityTier);
```
thisisamir98, line 306: Could we bind the radio to `qualityTier` (the user preference) instead of `effectiveQualityTier`?

`onChange` already writes the preference via `setBackgroundEffectsQualityTier`, but `selectedValue` currently comes from the temporary/adaptive effective tier. During privacy → segmenter degradation the radio can show Balance while preference is still Privacy, then the next restart correctly applies Privacy again. That makes the control disagree with the durable setting.

Prefer `state.qualityTier` for the selected radio. If we want to surface temporary degradation, that can be a separate hint/status rather than the editable selection.

Regression: preference is Privacy → adaptive degradation sets `effectiveQualityTier` to Balance → radio still shows Privacy selected.

thisisamir98 (changes requested):
Thanks for the follow-ups. The pipeline start path and preference vs effective store split look good, and the restart regressions help a lot.

One remaining point below: the radio is still bound to temporary effective state. After that, this looks good to go from my side.

thisisamir98 (approved):

PR #21921, opened 2026-07-21
fix: drive file type filter order(WPB-27130)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27130" title="WPB-27130" target="_blank">BugWPB-27130</a>  [Web] File types are not displayed in the defined fixed order
  </td></table>
  <br />
 

Reorder the drive file type catalog to match the PRD-defined display sequence for both Global and Shared Drive filters.

Add a regression test that asserts the catalog order so future changes do not accidentally alter the filter display order.

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/driveFilters/driveFilters.test.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/driveFilters/driveFilters.test.ts
@@ -28,6 +28,7 @@ import {
   toConversationDriveSearchParams,
   toGlobalDriveSearchParams,
 } from './driveFilters';
+import {FILE_TYPE_CATALOG} from './fileTypeCatalog';
 
 const emptyFilters: ConversationDriveFiltersState = {
   selectedTagIds: [],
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/common/driveFilters/driveFilters.test.ts
```diff
@@ -42,6 +43,22 @@ const emptyGlobalFilters: GlobalDriveFiltersState = {
   path: undefined,
 };
 
+describe('FILE_TYPE_CATALOG', () => {
+  it('lists file types in this fixed display order', () => {
+    expect(FILE_TYPE_CATALOG.map(({id}) => id)).toEqual([
+      'pdfs',
+      'documents',
```
screendriver, line 50: This only tests the order of the catalog constant, not the behavior described by [WPB-27130](https://wearezeta.atlassian.net/browse/WPB-27130).

The regression concerns the file type order exposed by both the Shared Drive and Global Drive filters. This test would still pass if either `useConversationDriveFilters` or `useGlobalDriveFilters` reordered or otherwise transformed the catalog before passing the items to `FilterPopover`.

Please add regression tests for both consumers and assert that the `fileType` filter contains the item IDs in the required display order.

screendriver (changes requested):

apps/webapp/src/script/components/Conversation/ConversationCells/common/driveFilters/driveFilters.test.ts
arjita-mitra: The file catalog is shared between global and shared drive per conversation. 

I get your point to add hook level testing so if later someone changes useConversationDriveFilters or useGlobalDriveFilters to sort/filter/map the catalog differently before passing items to FilterPopover, the catalog test would still pass while the UI behaviour regresses.

arjita-mitra: Addressed 233002923d3ab8ea66085d3b422b89a4b98dfa61

screendriver (approved):
Looks good now. The setup required for the global hook test highlights some hidden dependencies worth addressing separately in the near future.

PR #21924, opened 2026-07-21
chore(e2e): Clean-up and update of Git Actions jobs related to e2e tests [WPB-22420]

iskvortsov:
<table>
<td>
  </td></table>
  <br />
 

- Removed Beta option, added Dev option for e2e tests to run against
- Removed backendUrl as an input parameter
- Removed playwright-crit-flow-tests.yml


.github/workflows/e2e-tests.yml
```diff
@@ -100,7 +95,7 @@ jobs:
       - name: Run bootstrap smoke test
         env:
           WEBAPP_URL: ${{ inputs.webappUrl }}
-          BACKEND_URL: ${{ inputs.backendUrl }}
+          BACKEND_URL: ${{ github.event_name == 'workflow_dispatch' && 'https://staging-nginz-https.zinfra.io/' || inputs.backendUrl }}
```
screendriver: `backendUrl` is no longer defined for `workflow_dispatch`, but this fallback is only applied to the bootstrap smoke test. The actual E2E test job still uses `${{ inputs.backendUrl }}` on line 189, which will be empty for manually triggered runs.

Could we use the same resolved backend URL for both jobs so that the bootstrap test and the full E2E suite consistently run against Staging?

screendriver (changes requested):

.github/workflows/e2e-tests.yml
iskvortsov: Is there a plan to actually use another BE for these tests? It was never the case before so maybe I should just use 'https://staging-nginz-https.zinfra.io/' everywhere? We can change it later if the need arises

screendriver: I don't know to be honest.

iskvortsov: Let me remove it then.

.github/workflows/e2e-tests.yml
```diff
@@ -63,6 +55,9 @@ on:
         description: URL to the playwright report artifact
         value: ${{ jobs.e2e-report.outputs.reportUrl }}
 
+env:
+  BACKEND_URL: https://staging-nginz-https.zinfra.io/
```
screendriver, line 59: Defining `BACKEND_URL` once at workflow level fixes the inconsistency between the bootstrap test and the full E2E suite.

However, removing `backendUrl` from `workflow_call.inputs` breaks the existing invocation in `.github/workflows/precommit-slot-run.yml:194`, which still passes `backendUrl` to this workflow. That call chain is used by the `Release WebApp` E2E gate.

Could we either keep `backendUrl` as an optional reusable-workflow input with Staging as the fallback, or remove it consistently throughout the complete call chain in this PR?

iskvortsov: Removed backendUrl everywhere now.

screendriver (approved):

PR #21930, opened 2026-07-21
fix: harden cells search against stale requests [WPB-20817]

cristianoliveira:
<table>
<td>
  </td></table>
  <br />
 

## What changed

This hardens the global Cells search against stale requests

The annoying part was that debounce reduced the number of requests, but it did not decide which response was still allowed to update the shared store

So this PR adds lifecycle ownership around global Cells search:

- cancel pending debounced searches when clearing the input
- ignore stale responses with request version gating
- invalidate in-flight requests when the hook unmounts
- keep unmount cleanup lifecycle-only, so normal rerenders do not invalidate valid requests
- parallelize user/conversation enrichment after the Cells response

## Why

When typing quickly and then clearing the search input, an old request could finish after the clear request and overwrite the list with partial search results

This makes latest valid request win, and prevents unmounted hooks from writing into the shared Cells store

## DEMO

https://github.com/user-attachments/assets/1b2d90eb-10f0-4f0a-953a-e36ecc9fc5a1

### DIFF
--- a/apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
+++ b/apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
@@ -147,7 +147,7 @@ describe('useSearchCellsNodes', () => {
     useCellsStore.getState().clearAll();
   });
 
-  it('uses recency sorting for the global all-files view by default', async () => {
+  it('requests global files with recency sorting by default', async () => {
     const cellsRepository = createFakeCellsRepository();
     const {fireAndForgetInvoker} = renderSearchHook({cellsRepository});
     await act(() => fireAndForgetInvoker.waitUntilAllSettled());
@@ -162,7 +162,7 @@ describe('useSearchCellsNodes', () => {
     );
   });
 
-  it('lets the selected sort override the global default', async () => {
+  it('requests global files with the selected sort instead of the default', async () => {
     const cellsRepository = createFakeCellsRepository();
     const {fireAndForgetInvoker} = renderSearchHook({
       cellsRepository,
@@ -179,8 +179,14 @@ describe('useSearchCellsNodes', () => {
     );
   });
 
-  it('cancels a pending debounced search when clearing the search input', async () => {
-    const cellsRepository = createFakeCellsRepository();
+  it('keeps browse results when clearing a scheduled search before it runs', async () => {
+    const cellsRepository: FakeCellsRepository = {
+      searchNodes: jest
+        .fn()
+        .mockResolvedValueOnce({Nodes: [createRestNode('initial-file.pdf')]})
+        .mockResolvedValueOnce({Nodes: [createRestNode('browse-file.pdf')]})
+        .mockResolvedValueOnce({Nodes: [createRestNode('stale-file.pdf')]}),
+    };
     const controlledDebouncedSearch = createControlledDebouncedSearch();
     const {fireAndForgetInvoker, result} = renderSearchHook({
       cellsRepository,
@@ -194,29 +200,38 @@ describe('useSearchCellsNodes', () => {
       await controlledDebouncedSearch.flush();
     });
 
-    expect(cellsRepository.searchNodes).toHaveBeenCalledTimes(2);
-    expect(cellsRepository.searchNodes).not.toHaveBeenCalledWith(expect.objectContaining({query: 'stale-query'}));
+    expect(result.current.searchValue).toBe('');
+    expect(useCellsStore.getState().nodes.map(node => node.name)).toEqual(['browse-file.pdf']);
+    expect(useCellsStore.getState().status).toBe('success');
   });
 
-  it('does not update the shared files store when an in-flight search resolves after unmount', async () => {
-    const search = createDeferred<RestNodeCollection>();
+  it('does not replace files set after unmount when the pending request resolves', async () => {
+    const requestAfterUnmount = createDeferred<RestNodeCollection>();
     const cellsRepository: FakeCellsRepository = {
-      searchNodes: jest.fn().mockReturnValue(search.promise),
+      searchNodes: jest
+        .fn()
+        .mockResolvedValueOnce({Nodes: [createRestNode('current-file.pdf')]})
+        .mockReturnValueOnce(requestAfterUnmount.promise),
     };
+    const {fireAndForgetInvoker, result, unmount} = renderSearchHook({cellsRepository});
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+    const currentFiles = useCellsStore.getState().nodes;
 
-    const {unmount} = renderSearchHook({cellsRepository});
-
+    act(() => {
+      void result.current.handleReload();
+    });
     unmount();
+    useCellsStore.setState({nodes: currentFiles});
 
     await act(async () => {
-      search.resolve({Nodes: [createRestNode('unmounted-file.pdf')]});
+      requestAfterUnmount.resolve({Nodes: [createRestNode('unmounted-file.pdf')]});
       await flushPromises();
     });
 
-    expect(useCellsStore.getState().nodes).toEqual([]);
+    expect(useCellsStore.getState().nodes.map(node => node.name)).toEqual(['current-file.pdf']);
   });
 
-  it('ignores stale search responses when a newer request has already updated the files', async () => {
+  it('keeps results from the newer request when the older request resolves last', async () => {
     const staleSearch = createDeferred<RestNodeCollection>();
     const currentSearch = createDeferred<RestNodeCollection>();
     const cellsRepository: FakeCellsRepository = {
--- a/apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
+++ b/apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
@@ -295,7 +295,9 @@ describe('useSearchCellsNodes', () => {
   });
 
   it('shows an error and clears results when the current request fails for a Cells participant', async () => {
-    const cellsRepository: CellsRepositoryMock = {searchNodes: jest.fn().mockRejectedValue(new Error('request failed'))};
+    const cellsRepository: CellsRepositoryMock = {
+      searchNodes: jest.fn().mockRejectedValue(new Error('request failed')),
+    };
     const conversationRepository: ConversationRepositoryMock = {
       ...buildConversationRepositoryMock(),
       getAllCellEnabledGroupConversations: jest.fn().mockReturnValue([{}]),
@@ -310,7 +312,9 @@ describe('useSearchCellsNodes', () => {
   });
 
   it('treats a current request failure as an empty result when the user has no Cells conversations', async () => {
-    const cellsRepository: CellsRepositoryMock = {searchNodes: jest.fn().mockRejectedValue(new Error('request failed'))};
+    const cellsRepository: CellsRepositoryMock = {
+      searchNodes: jest.fn().mockRejectedValue(new Error('request failed')),
+    };
     const {fireAndForgetInvoker} = renderSearchHook({cellsRepository});
 
     await act(() => fireAndForgetInvoker.waitUntilAllSettled());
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
```diff
@@ -71,18 +77,30 @@ const DEBOUNCE_TIME = 300;
...
   const [searchQuery, setSearchQuery] = useState('');
   const [pageSize, setPageSize] = useState(PAGE_INITIAL_SIZE);
   const isInitialLoad = useRef(true);
   const shouldPerformFullReload = useRef(true);
+  const requestVersionGate = useRef(createRequestVersionGate());
 
   const searchNodes = useCallback(
     async (properties: SearchNodesProperties): Promise<void> => {
       const {query, status, limit = pageSize} = properties;
+      const requestVersion = requestVersionGate.current.next();
```
arjita-mitra, line 105: Adding console.debug('Ignoring stale request version:', requestVersion) could be helpful for future debugging.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
```diff
@@ -71,18 +77,30 @@ const DEBOUNCE_TIME = 300;
...
+    createDebouncedSearch,
+  } = properties;
   const {setNodes, setStatus, setPagination, clearAll} = useCellsStore();
 
   const [searchValue, setSearchValue] = useState('');
   const [searchQuery, setSearchQuery] = useState('');
   const [pageSize, setPageSize] = useState(PAGE_INITIAL_SIZE);
   const isInitialLoad = useRef(true);
   const shouldPerformFullReload = useRef(true);
+  const requestVersionGate = useRef(createRequestVersionGate());
```
arjita-mitra, line 100: Not part of this PR but could be valuable to add readonly interface properties, 

export interface RequestVersionGate {
  readonly next: () => number;
  readonly invalidate: () => void;
  readonly isStale: (requestId: number) => boolean;
}

then use export const createRequestVersionGate = (): Readonly<RequestVersionGate> makes the TypeScript guarantee explicit at the call site.

finally return Object.freeze({
    next(): number { for runtime protection

cristianoliveira: I agree, we can do a refactoring maybe even move to its own module

thisisamir98 (commented):
Nice hardening of the clear vs stale race. The version gate plus debounce cancel on clear/unmount looks like the right ownership model for the shared store.

A few follow-ups below on reload cancel consistency and making the regressions match the real store outcomes a bit more closely. Nothing blocking from my side once those are considered.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
```diff
@@ -161,11 +211,12 @@ export const useSearchCellsNodes = (properties: UseSearchCellsNodesProps): UseSe
     setPageSize(PAGE_INITIAL_SIZE);
     setSearchValue(value);
     fireAndForgetInvoker.fireAndForget(async (): Promise<void> => {
-      await searchNodesDebounced(value);
+      await searchNodesDebouncedRef.current(value);
     });
   };
 
   const handleClearSearch = async (): Promise<void> => {
+    searchNodesDebouncedRef.current.cancel();
```
thisisamir98, line 218: Nice that clear cancels the pending debounced search before starting the browse fetch.

Could we do the same in `handleReload`? If the user types, then hits refresh before the 300ms debounce fires, reload can finish with the current `searchQuery`, and the still-pending debounce can overwrite the shared store afterward.

Conversation search also cancels before superseding. A small regression would help: type without flushing debounce → reload completes → flush debounce → store stays on the reload result.

Also a question: in `useConversationSearchFiles`, clear calls `invalidate()` immediately because clear does not always start a new search. Here clear always goes through `searchNodes()` → `next()`. Is the intentional contract that only `searchNodes` bumps the version, or should clear (and maybe reload) also `invalidate()` at the action boundary for the same reason as conversation search?

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
```diff
@@ -198,6 +249,19 @@ export const useSearchCellsNodes = (properties: UseSearchCellsNodesProps): UseSe
...
+  // This cleanup models hook ownership ending. Do not depend on changing callbacks,
+  // otherwise ordinary rerenders would invalidate valid in-flight requests.
+  useEffect(() => {
+    const versionGate = requestVersionGate.current;
+
+    return () => {
+      // Read from the ref during cleanup so we cancel the latest debounced search,
+      // not the one captured when this mount-only effect was created.
+      searchNodesDebouncedRef.current.cancel();
+      versionGate.invalidate();
```
thisisamir98, line 261: Mount-only cleanup with the ref read is clear, thanks.

One open question: invalidate makes in-flight work a no-op after it may already have set `status` to `loading`. Since this store is shared, is a stuck `loading` until the next mount acceptable, or should cleanup reset status when invalidating?

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
```diff
@@ -143,15 +184,24 @@ export const useSearchCellsNodes = (properties: UseSearchCellsNodesProps): UseSe
...
 
-  const searchNodesDebounced = useDebouncedCallback(async (value: string): Promise<void> => {
+  const defaultSearchNodesDebounced = useDebouncedCallback(async (value: string): Promise<void> => {
     shouldPerformFullReload.current = false;
     setSearchQuery(value);
     await searchNodes({query: value, status: 'loading'});
     shouldPerformFullReload.current = true;
   }, DEBOUNCE_TIME);
+  const searchNodesDebounced =
+    createDebouncedSearch?.(async (value: string): Promise<void> => {
```
thisisamir98: Nit: the debounced callback body is duplicated between `useDebouncedCallback` and the `createDebouncedSearch` path. Could we extract a shared `runDebouncedSearch` so production and the test injection stay in sync?

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
```diff
@@ -126,4 +189,138 @@ describe('useSearchCellsNodes', () => {
...
+    };
+    const {fireAndForgetInvoker, result, unmount} = renderSearchHook({cellsRepository});
+    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
+    const currentFiles = useCellsStore.getState().nodes;
+
+    act(() => {
+      void result.current.handleReload();
+    });
+    unmount();
+    useCellsStore.setState({nodes: currentFiles});
```
thisisamir98: Could we avoid rewriting the store here?

`handleReload()` already runs `clearAll()`, so after unmount the natural state is empty nodes (and likely still `loading`). Restoring `currentFiles` hides that and only checks that the late response is ignored.

I would prefer asserting from the real post-unmount state, for example nodes stay `[]` / status stays whatever reload set, and the late response still does not write `unmounted-file.pdf`.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
```diff
@@ -126,4 +189,138 @@ describe('useSearchCellsNodes', () => {
       }),
     );
   });
+
+  it('keeps browse results when clearing a scheduled search before it runs', async () => {
```
thisisamir98, line 198: This locks the cancel-before-run path well.

Could we also cover the in-flight clear path from the ticket? Something like: start a controllable in-flight search → `handleClearSearch` → resolve browse → resolve stale last → store stays on browse files.

The version gate should already handle it, and the reload out-of-order test is close, but a clear-specific case would lock the product path more directly.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
cristianoliveira: I’d keep this as the injected logger

Same debugging value, but it keeps the hook testable and consistent with the rest of the app logging

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
cristianoliveira: Pushed the reload part

Reload cancels the pending debounce now, so a delayed query should not overwrite the reload result

For the explicit invalidate bit, I think this one is slightly different. Here clear/reload always go through `searchNodes()`, so we already get a new request version from `next()`

In conversation search clear can stop without a replacement request, that’s why the explicit `invalidate()` felt needed there

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
cristianoliveira: Hmm I’d avoid resetting the status on cleanup here

The hook is going away at that point, so I’d expect cleanup to just invalidate/cancel what it owns. If there is a real UI case where this leaves the view stuck, then yes, we should fix that path. WDYT?

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
cristianoliveira: Done, extracted `runDebouncedSearch` so both paths share the same body

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
cristianoliveira: Done, changed the test to assert the natural post-unmount state instead of rewriting the store

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.test.ts
cristianoliveira: Done, added the clear in-flight regression as well

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
thisisamir98: Agreed. The injected logger covers the debugging case without coupling the hook to `console`, and the stale-request test can assert it cleanly.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
thisisamir98: Looks good. The gate already has the `readonly` members, `Readonly<RequestVersionGate>`, and `Object.freeze` in this PR. Moving it to its own module can stay as a follow-up.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
thisisamir98: Thanks, that addresses it. Reload cancel plus the regression looks right, and I agree explicit `invalidate()` is not needed here since clear/reload always go through `searchNodes()` → `next()`.

apps/webapp/src/script/components/cellsGlobalView/useSearchCellsNodes/useSearchCellsNodes.ts
thisisamir98: Agreed. Cleanup should only cancel/invalidate what the hook owns. This view mounts and unmounts with the hook, so a remount starts a fresh search, and the updated unmount test documents the leftover `loading` state. No status reset needed unless we find a real stuck UI path later.

thisisamir98 (approved):
The follow-ups look good. Reload cancel, the in-flight clear regression, and the natural unmount assertions close the remaining race gaps. Approving.

PR #21942, opened 2026-07-22
fix: cells empty search results message(WPB-25184)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25184" title="WPB-25184" target="_blank">TaskWPB-25184</a>  update no result found message
  </td></table>
  <br />
 

## Summary
- Add shared 'CellsStateInfo' for conversation and global cells views
- Use the search no-results layout for typed search and filter-only empty results
- Add accessible 'role="status"' for search no-results messaging
- Prevent empty table wrappers from pushing empty-state messages to the bottom


cristianoliveira (commented):
Looks good, no blocker from me

Only thing I’d add is a small behavior test around the container logic, since that’s the risky part here

Search/filter with no results should show the search empty state, while normal empty drive/recycle bin should keep the regular copy

cristianoliveira (approved):
Approving since that isn't a blocker

PR #21947, opened 2026-07-22
fix: wire core-crypto db if client is stale [WPB-26879]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26879" title="WPB-26879" target="_blank">BugWPB-26879</a>  [Web][Registration] Key material of deleted clients gets used to register new clients 
  </td></table>
  <br />
 

## Summary

Fixes a bug where registering a new client after a long-expired session could reuse MLS key material from a client that had already been deleted server-side, instead of generating fresh key material.

### Root cause

`ClientService.loadClient()` validates the locally cached client against the backend and, on a 404 (client deleted remotely), wipes the local Proteus identity via `proteusService.wipe()`. However, the MLS/CoreCrypto keystore is a separate IndexedDB store scoped by **user ID** (`corecrypto-${storeEngine.storeName}`), not by client ID. This keystore was never cleared on that path, so it persisted across client deletion and got silently reused the next time a client was registered for that user — while the client ID itself was fresh, its MLS key material was not.

### Fix

Extended the existing 404 branch in `ClientService.loadClient()` (`libraries/core/src/client/clientService.ts`) to also call `wipeCoreCryptoDb(this.storeEngine)` alongside `proteusService.wipe()`. This reuses the same wipe function `Account.logout()` already calls on explicit logout/account deletion, so no new deletion logic was introduced and no new dependencies were required.


libraries/core/src/client/clientService.test.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import {wipeCoreCryptoDb} from '../messagingProtocols/proteus/proteusService/cryptoClient/coreCryptoWrapper';
+import {getUUID} from '../test/payloadHelper';
+import {createMemoryEngine} from '../test/storeHelper';
+import {ProteusService} from '../messagingProtocols/proteus';
+
+jest.mock('../messagingProtocols/proteus/proteusService/cryptoClient/coreCryptoWrapper', () => ({
+  ...jest.requireActual('../messagingProtocols/proteus/proteusService/cryptoClient/coreCryptoWrapper'),
+  wipeCoreCryptoDb: jest.fn().mockResolvedValue(undefined),
+}));
```
thisisamir98: is there a way to avoid mocking entire files?

e-maad: I just need to assert if `wipeCoreCryptoDb` was called through the test.

thisisamir98 (approved):

PR #21951, opened 2026-07-23
fix: adjust adaptive background effect quality adjustment based on runtime performance [WPB-26253]

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26253" title="WPB-26253" target="_blank">TaskWPB-26253</a>  Implement Automatic BGE Performance Scaling up Based on GPU Render Time
  </td></table>
  <br />
 

## Summary

This PR introduces adaptive quality adjustment for virtual background effects. The quality controller continuously monitors runtime performance and automatically adjusts the processing quality to maintain smooth rendering on a wide range of devices and automatically adjusts segmentation resolution based on measured processing performance.

### DIFF
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
@@ -301,7 +301,7 @@ export class BackgroundEffectsController {
       return;
     }
 
-    const isLowQualityTier = this.activeQualityTier !== 'hd' && this.activeQualityTier !== 'fhd';
+    const isLowQualityTier = !TIER_DEFINITIONS[this.activeQualityTier].isHighQuality;
 
     const effectiveModelPath =
       isLowQualityTier && this.requestedModelPath === SELFIE_MULTICLASS_MODEL_PATH
--- a/apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
+++ b/apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
@@ -69,10 +69,14 @@ export class BackgroundEffectsController {
     requestVideoFrameCallback: false,
   };
 
+  // adaptive quality mode settings:
   private qualityController: QualityController | null = null;
   private qualitySampleQueue = Promise.resolve();
-  private maxResolution: Resolution = TIER_DEFINITIONS.hd.resolution;
-  private maxQualityTier: QualityTier = 'hd';
+  private maxResolution: Resolution = TIER_DEFINITIONS.fhd.resolution;
+  private maxQualityTier: QualityTier = 'fhd';
+  // the manually set mode. if set to auto, then adaptive quality mode is activated
+  private qualityMode: QualityMode = 'auto';
+  private activeQualityTier: QualityTier = 'fhd';
   private requestedModelPath: string = SELFIE_SEGMENTER_MODEL_PATH;
   private refcount = 0;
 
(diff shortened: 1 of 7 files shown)


apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
```diff
@@ -278,11 +300,18 @@ export class BackgroundEffectsController {
     if (!this.refcount) {
       return;
     }
-    // in case of not HD we will always use more performant model
-    if (this.options.quality !== 'hd' && this.options.quality !== 'fhd' && this.options.quality !== 'auto') {
-      this.options.modelPath = SELFIE_SEGMENTER_MODEL_PATH;
-      backgroundEffectsStore.getState().setIsHighQualityBlurEnabled(false);
-    }
+
+    const isLowQualityTier = this.activeQualityTier !== 'hd' && this.activeQualityTier !== 'fhd';
```
arjita-mitra: This assumes only 2 high quality tiers. If new tiers are added the code will treat it as loa quality and downgrades silently. Better approach would be explicit tier Classification + validation

```
const HIGH_QUALITY_TIERS = new Set<QualityTier>(['hd', 'fhd']);

const isKnownTier = Object.values(QUALITY_TIERS).includes(this.activeQualityTier as QualityTier);

if (!isKnownTier) {
this.logger.error(...)
}
```

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -270,6 +281,80 @@ export async function runSegmenter(
     canvas.removeEventListener('webglcontextrestored', onContextRestored);
   }
 
+  function triggerGpuTracking(gpuStart: number, frameDeltaMs: number, segmentationMs: number, filterMs: number) {
+    const gl = webGLRenderer?.gl;
+    if (!gl) {
```
arjita-mitra, line 288: If i understood correctly when webGLRenderer becomes null after context loss this silently returns. shouldn't we add a logger.warn and call updateMetrics?

apps/webapp/src/script/repositories/media/backgroundEffects/pipe/segmenter.ts
```diff
@@ -270,6 +281,80 @@ export async function runSegmenter(
...
+        filterMsSum += data.filterMs;
+        frames++;
+        totalFrames++;
+
+        // 2. Update your metrics (delayed until GPU was ready)
+        updateMetrics(totalMs, data.segmentationMs, gpuMs);
+
+        // 3. Tidying Up
+        gl.deleteSync(sync);
+        activeGpuQueries.delete(sync);
```
arjita-mitra, line 353: activeGpuQueries.delete(sync) modifies the map during iteration. This can cause the iterator to skip entries or behave unpredictably.

EnricoSchw: Deleting the currently iterated entry from a JavaScript `Map` is safe. `Map` iterators remain valid when the current entry is removed. Do you see a case where it could make an issue?

apps/webapp/src/script/repositories/media/backgroundEffects/backgroundEffectsController.ts
EnricoSchw: I have make the high-quality tier handling explicit. I don't think runtime validation is needed here because activeQualityTier is already type-safe.

zskhan (approved):

PR #21957, opened 2026-07-23
fix: show all app participants in group details [WPB-27417]

MohamadJaara:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27417" title="WPB-27417" target="_blank">BugWPB-27417</a>  [Web] show group members regardless of protocol
  </td></table>
  <br />
 

## Goal

Keep group participant details complete and consistent when a conversation changes protocol.

## Changes

- combine legacy bot/service participants and newer app participants for every conversation protocol
- remove the protocol-specific participant-list branch
- add regression coverage for Proteus, mixed, and MLS groups

## Root cause

The group details view selected legacy service participants only for Proteus conversations and selected newer app participants for every other protocol. As a result, legacy bots were omitted from the displayed list after a conversation entered mixed or MLS state.

## Impact

Group details now show all bot and app participants consistently throughout protocol migration.

## Validation

- Conversation details regression coverage passes for Proteus, mixed, and MLS groups
- scoped lint and formatting checks pass

### DIFF
--- a/apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.test.tsx
+++ b/apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.test.tsx
@@ -25,6 +25,7 @@ import {UserType} from '@wireapp/api-client/lib/user';
 import {CellsRepository} from 'Repositories/cells/cellsRepository';
 import {ConnectionRepository} from 'Repositories/connection/connectionRepository';
 import {ConversationRepository} from 'Repositories/conversation/ConversationRepository';
+import {ConversationMapper} from 'Repositories/conversation/ConversationMapper';
 import {ConversationRoleRepository} from 'Repositories/conversation/ConversationRoleRepository';
 import {MessageRepository} from 'Repositories/conversation/MessageRepository';
 import {Conversation} from 'Repositories/entity/Conversation';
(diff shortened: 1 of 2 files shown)


arjita-mitra:
@MohamadJaara need a jira ticket in the PR title for the failing check - [Link and Lint PR with Jira Ticket Number / add-jira-description (pull_request)](https://github.com/wireapp/wire-webapp/actions/runs/29996293552/job/89170652845?pr=21957)

apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.test.tsx
```diff
@@ -120,6 +122,48 @@ const getDefaultParams = () => {
 };
 
 describe('ConversationDetails', () => {
+  it.each([CONVERSATION_PROTOCOL.PROTEUS, CONVERSATION_PROTOCOL.MIXED, CONVERSATION_PROTOCOL.MLS])(
```
arjita-mitra, line 126: The current test validates that each protocol works independently, but it doesn't test the actual bug scenario where:

1. Conversation starts as Proteus (with legacy bots visible)
2. Protocol changes to Mixed or MLS
3. Legacy bots should still remain visible (this was the bug)

it would be good to add this.

MohamadJaara: The transition itself does not matter, it is just a state the app can go through to repro the issue. What matters is that the showing of members should never be guarded by protocol.

conv protocol should dictate what message type is sent, but never the list of members

arjita-mitra: Imo adding it as part of regression test is worth it because that tells us why exactly the old code broke.

arjita-mitra (approved):

PR #21966, opened 2026-07-23
feat: Add self user drive role (editor / viewer) in shared drive(WPB-26111)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26111" title="WPB-26111" target="_blank">TaskWPB-26111</a>  [Web] Add self user drive role (editor / viewer) in shared drive
  </td></table>
  <br />
 

As a system, I want to know my role (editor / viewer) in the shared drive screen so I know what are the actions available to me.

Editor = all actions

Viewer = restricted actions

Note:

In milestone1 (this current task), this will default to editor for all users including guests.

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSelfUserDriveRole/CellsSelfUserDriveRoleContext.tsx
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSelfUserDriveRole/CellsSelfUserDriveRoleContext.tsx
@@ -1,0 +1,62 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSelfUserDriveRole/CellsSelfUserDriveRoleContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const CellsSelfUserDriveRoleContext = createContext<CellsSelfUserDriveRole>(CELLS_SELF_USER_DRIVE_ROLE.EDITOR);
+
+interface CellsSelfUserDriveRoleProviderProps {
+  children: ReactNode;
+  selfUserDriveRole: CellsSelfUserDriveRole;
+}
+
+export const CellsSelfUserDriveRoleProvider = ({children, selfUserDriveRole}: CellsSelfUserDriveRoleProviderProps) => {
+  return (
```
ikotarac, line 53: why are we not calculating role directly here:

```tsx
 const selfUserDriveRole = getSelfUserDriveRole({
      conversationTeamId: activeConversation.teamId,
      selfUserTeamId: selfUser?.teamId,
    });
```

then we don't have to export `getSelfUserDriveRole` and keep it internal to this context provider, so it's more encapsulated

arjita-mitra: I kept `getSelfUserDriveRole` exported because the role derivation is business logic, while the context is only the delivery mechanism for descendants. 

We already test the role calculation directly, and I expect to reuse the same viewer/editor logic outside the context subtree for permission checks.

If we inline the calculation into the provider setup, we improve local encapsulation but make reuse harder and risk duplicating role logic elsewhere.

cristianoliveira (approved):

PR #21970, opened 2026-07-23
fix: update concurrency group of pull translation [WPB-22420]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Update concurrency group name of pull translation to crowdin-pull-main and cancel-in-progress to true for both workflow
This ensures source uploads and translation downloads cannot run concurrently.

### DIFF
--- a/.github/workflows/push-translations-to-crowdin.yml
+++ b/.github/workflows/push-translations-to-crowdin.yml
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: crowdin-push-${{ github.ref }}
-  cancel-in-progress: true
+  group: crowdin-push-main
+  cancel-in-progress: false
 
 jobs:
   push_to_crowdin:
--- a/.github/workflows/pull-translations-from-crowdin.yml
+++ b/.github/workflows/pull-translations-from-crowdin.yml
@@ -1,8 +1,8 @@
 name: Pull translations from Crowdin
 
 on:
-  push:
-    branches: [main]
+  schedule:
+    - cron: '17 12,14,16,18 * * 1-5'
   workflow_dispatch:
 
 permissions:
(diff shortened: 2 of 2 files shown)


.github/workflows/push-translations-to-crowdin.yml
```diff
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: crowdin-push-${{ github.ref }}
-  cancel-in-progress: true
+  group: crowdin-push-main
+  cancel-in-progress: false
```
ikotarac, line 10: why are we not cancelling workflows that are in progress?

I guess if we have a new merge into main, while previous is still running, we wouldn't cancel it but just do both of them, which is fine. I think they will probably take a few seconds anyway.

however, I don't see how is this preventing push and pull from running concurently, since we have them in different concurrency groups: 'crowdin-pull-main' and 'crowdin-push-main'

.github/workflows/push-translations-to-crowdin.yml
```diff
@@ -6,8 +6,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: crowdin-push-${{ github.ref }}
```
ikotarac, line 9: I guess this didn't change a lot it's just that name won't be `crowdin-push-refs/heads/main` but `crowdin-push-main`, but I am ok with keeping those two consistent

zskhan: No it is not making a difference but there is no need of `github.ref` as this will only run on `main`

.github/workflows/push-translations-to-crowdin.yml
zskhan: It is simply to avoid cancelling an in-flight push if another commit lands on `main` a few seconds later.
[Pushing workflow](https://github.com/wireapp/wire-webapp/actions/workflows/push-translations-to-crowdin.yml) approximately takes 1.5-2 minutes

screendriver (approved):

PR #22026, opened 2026-07-28
fix: change recurrence option label of "Monthly" to "Every four weeks" [WPB-27426]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27426" title="WPB-27426" target="_blank">StoryWPB-27426</a>  [Web] Change Meeting Recurrence option label to "Every 4 weeks"
  </td></table>
  <br />
 

Old view with Monthly badge
image

Updated view with Every four weeks badge
image


apps/webapp/src/i18n/en-US.json
```diff
@@ -1331,8 +1331,8 @@
   "meetings.scheduleModal.previousMonthAriaLabel": "Previous month",
   "meetings.scheduleModal.recurrence.daily": "Daily",
   "meetings.scheduleModal.recurrence.doesNotRepeat": "Does not repeat",
+  "meetings.scheduleModal.recurrence.everyFourWeeks": "Every four weeks",
```
thisisamir98, line 1334: Could we change this to `Every 4 weeks`? The acceptance criteria require that exact copy, and it keeps the recurrence labels consistent with the numeric `Every 2 weeks` format.

zskhan: I just discussed this with Federico and we need to be consistent so "Every four weeks" is the correct string. I posted the comment in the ticket.

thisisamir98 (approved):

PR #22029, opened 2026-07-29
feat: add epoch advancing debug tool [WPB-22420]

e-maad:
<table>
<td>
  </td></table>
  <br />
 

### Description

Added debug option for epoch advancement.

https://github.com/user-attachments/assets/813a97c2-a2ee-4b55-96c2-3f5bcd02c3c8

### DIFF
--- a/apps/webapp/src/script/components/configToolbar/configToolbar.tsx
+++ b/apps/webapp/src/script/components/configToolbar/configToolbar.tsx
@@ -60,6 +60,7 @@ export function ConfigToolbar() {
   const alphabeticallySortedStartupFeatureToggleNames = startupFeatureToggleNames.toSorted();
   const [showConfig, setShowConfig] = useState(false);
   const [isResettingMLSConversation, setIsResettingMLSConversation] = useState(false);
+  const [isAdvancingEpoch, setIsAdvancingEpoch] = useState(false);
   const [isGzipEnabled, setIsGzipEnabled] = useState(window.wire?.app.debug?.isGzippingEnabled() ?? false);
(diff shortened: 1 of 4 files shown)


thisisamir98 (dismissed):

apps/webapp/src/script/components/configToolbar/configToolbar.tsx
```diff
@@ -388,6 +389,17 @@ export function ConfigToolbar() {
     }
   };
 
+  const advanceEpoch = async () => {
+    setIsAdvancingEpoch(true);
+    try {
+      await window.wire?.app?.debug?.updateActiveConversationKeyPackages();
+    } catch (error: unknown) {
+      console.error('Error advancing epoch:', error);
```
screendriver: Why don't we use our logger? And why is no ESLint rule complaining about using `console` directly? 🤔

e-maad: The eslint rule no-console in `libraries/eslint-config/eslintrc.js:75-80` isn't disabled. It's configured with allow: ['error', 'info', 'warn'], so console.error/info/warn are permitted; only console.log/debug would be flagged. All the console.error calls in `configToolbar.tsx`

e-maad: This file is dev-only tooling, gated behind showConfig and never shown in production UI flow. Should we wire it up with a named logger instance?

screendriver: > Should we wire it up with a named logger instance?

Yes, because consistency over everything. Otherwise it leads to confusions and discussions like this one 🙂

zskhan (approved):

PR #22036, opened 2026-07-29
Fix Meetings empty state copy and update banner overlap [WPB-27520]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27520" title="WPB-27520" target="_blank">BugWPB-27520</a>  UI issues in Web - Meetings 
  </td></table>
  <br />
 

## Summary

- Update the empty Meetings help text so it only mentions team members, not guests or external parties.
- Adjust the Meetings layout to fill the main column height instead of using a fixed viewport height, so the full view moves below the update banner.

## Test plan

- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/EmptyMeetingList/EmptyMeetingList.test.tsx`
- [x] `yarn nx run webapp:test --testFile=src/script/components/Meeting/meeting.styles.test.ts`
- [x] Checked Meetings in the browser with a simulated update banner: Create meeting stays below the banner with no overlap
- [x] Confirm empty Meetings state shows the updated help text on an account with no meetings


apps/webapp/src/script/components/Meeting/EmptyMeetingList/EmptyMeetingList.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  return translationKey;
+};
+
+const rootProviderWrapper = createRootProviderWrapperForTest(
+  createRootContextValueForTest({translate: translateForEmptyMeetingListTest}),
+);
+
+describe('EmptyMeetingList', () => {
+  it('does not mention guests or external parties in the empty-state help text', () => {
```
zskhan: We usually don't test the translations, and here we are testing that  `guests` or `external parties` should not be in the translated string.

thisisamir98: Removed the test — agreed that we shouldn't be asserting on translation copy here.

zskhan (approved):

PR #22051, opened 2026-07-31
push helm charts to quay.io

smatting:


.github/workflows/publish.yml
```diff
@@ -2,7 +2,7 @@ name: Publish
 
 on:
   push:
-    branches: [master, dev, release/*, new-navigation]
+    branches: [master, dev, release/*, new-navigation, smatting/push-to-quay]
```
screendriver: I think this can be removed?

.github/workflows/stefan-test.yml
```diff
@@ -0,0 +1,79 @@
+name: Stefan Test
```
screendriver: Should we rename the filename and the name of the workflow?

.github/workflows/stefan-test.yml
```diff
@@ -0,0 +1,79 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
screendriver:
```suggestion
    runs-on: ubuntu-24.04
```

screendriver (approved):

PR #22080, opened 2026-08-03
fix(drive): disable search in recycle bin [WPB-27709]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27709" title="WPB-27709" target="_blank">TaskWPB-27709</a>  [Web] Remove the option to sort, filter and search files and folders in Recycling Bin
  </td></table>
  <br />
 

Avoid showing the conversation Drive search view when navigating to the recycle bin, and hide search controls and filters while it is active. Match recycle-bin path segments exactly so folders with similar names retain Drive controls.

Jira: https://wearezeta.atlassian.net/browse/WPB-27709

### DIFF
--- a/apps/webapp/src/script/components/Conversation/ConversationCells/common/getCellsFilesPath/getCellsFilesPath.ts
+++ b/apps/webapp/src/script/components/Conversation/ConversationCells/common/getCellsFilesPath/getCellsFilesPath.ts
@@ -19,10 +19,10 @@
 
 const MIN_PARTS_LENGTH = 2;
 
-export const getCellsFilesPath = () => {
-  const hash = window.location.hash.replace('#', '');
+export const getCellsFilesPath = (hash = window.location.hash) => {
+  const normalizedHash = hash.replace('#', '');
 
-  const parts = hash.split('/files/');
+  const parts = normalizedHash.split('/files/');
   const hasFilesPath = parts.length >= MIN_PARTS_LENGTH;
 
   if (!hasFilesPath) {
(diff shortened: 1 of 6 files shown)


apps/webapp/src/script/components/Conversation/ConversationCells/common/recycleBin/recycleBin.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    window.location.hash = '';
+  });
+
+  it.each([
+    'recycle_bin',
+    'recycle_bin/deleted-file',
+    'recycle_bin/deleted-file/nested-folder',
+    'recycle_bin%2Fdeleted-file',
+  ])('recognizes a recycle bin path: %s', path => {
+    window.location.hash = `#/conversation/conversation-id/files/${path}`;
```
zskhan: We can avoid window.location mutation in test if we introduce another pure function for this validation.

zskhan (dismissed):

apps/webapp/src/script/components/Conversation/ConversationCells/common/recycleBin/recycleBin.test.ts
cristianoliveira: Makes sense, let's refactor that

apps/webapp/src/script/components/Conversation/ConversationCells/common/getCellsFilesPath/getCellsFilesPath.ts
```diff
@@ -19,10 +19,10 @@
 
 const MIN_PARTS_LENGTH = 2;
 
-export const getCellsFilesPath = () => {
-  const hash = window.location.hash.replace('#', '');
+export const getCellsFilesPath = (hash = window.location.hash) => {
```
cristianoliveira, line 22: Here @zskhan added a default parameter that allows the dependency injection of window, without the need of a bigger refactor :)

zskhan (approved):

PR #22081, opened 2026-08-03
Expose meeting fetch and route lifecycle events [WPB-26735]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26735" title="WPB-26735" target="_blank">TaskWPB-26735</a>  [Web] Support new event types for Meetings
  </td></table>
  <br />
 

## Summary
Add single-meeting fetch to the meetings repository and route `meeting.create`, `meeting.update`, and `meeting.delete` through validated WebApp events.

## Stack
First of 5 PRs for WPB-26735. Merge this before reviewing later layers.

Follow-up: #22082

### DIFF
--- a/apps/webapp/src/script/repositories/event/EventRepository.test.ts
+++ b/apps/webapp/src/script/repositories/event/EventRepository.test.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {BackendEvent, CONVERSATION_EVENT, USER_EVENT} from '@wireapp/api-client/lib/event/';
+import {BackendEvent, CONVERSATION_EVENT, MEETING_EVENT, USER_EVENT} from '@wireapp/api-client/lib/event/';
 import {ConnectionState} from '@wireapp/core';
 import {WebAppEvents} from '@wireapp/webapp-events';
 import {amplify} from 'amplify';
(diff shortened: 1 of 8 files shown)


cristianoliveira (dismissed):

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -581,18 +581,38 @@ export class EventRepository {
...
+        return;
+      }
+      case MEETING_EVENT.DELETE: {
+        if (!hasValidQualifiedId) {
+          this.logger.warn('Ignored meeting.delete event without a valid qualified_id', event);
+          return;
+        }
+        amplify.publish(WebAppEvents.MEETING.DELETED, meetingId);
+        return;
+      }
```
cristianoliveira: Nitpick
These 3 branches pretty much do the same checks, I wonder if there is a less verbose way

apps/webapp/src/script/repositories/event/EventRepository.test.ts
```diff
@@ -261,40 +261,88 @@ describe('EventRepository', () => {
...
+      async (eventType, webAppEvent) => {
+        const event = {
+          ...meetingLifecycleEventBase,
+          type: eventType,
+        };
+
+        await eventRepository['distributeEvent'](event as any, EventSource.WEBSOCKET);
+
+        expect(amplify.publish).toHaveBeenCalledWith(webAppEvent, meetingId);
+        expect(amplify.publish).not.toHaveBeenCalledWith(eventType, expect.anything());
```
e-maad, line 292: If you are expecting anything why can't you use `toHaveBeenCalle`?

apps/webapp/src/script/repositories/event/EventRepository.test.ts
```diff
@@ -261,40 +261,88 @@ describe('EventRepository', () => {
...
+        expect(amplify.publish).not.toHaveBeenCalledWith(
```
e-maad, line 293: https://github.com/wireapp/wire-webapp/pull/22081/changes#r3705460334

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -581,18 +581,38 @@ export class EventRepository {
...
-      amplify.publish(event.type, event);
-      return;
-    }
-
     const meetingId = 'qualified_id' in event ? event.qualified_id : undefined;
-    if (meetingId === undefined || typeof meetingId.id !== 'string' || typeof meetingId.domain !== 'string') {
-      this.logger.warn('Ignored meeting.delete event without a valid qualified_id', event);
-      return;
+    const hasValidQualifiedId =
+      meetingId !== undefined && typeof meetingId.id === 'string' && typeof meetingId.domain === 'string';
```
e-maad, line 598: If you want runtime validation, should we use zod?
We can set up a schema and validate it with better reusability.

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -581,18 +581,38 @@ export class EventRepository {
...
+
+    switch (event.type) {
```
e-maad: Wouldn't be nice if we use ts-pattern?

apps/webapp/src/script/repositories/meetings/meetingsApiDataSource.test.ts
```diff
@@ -41,4 +41,31 @@ describe('MeetingsApiDataSource', () => {
     expect(createMeeting).toHaveBeenCalledTimes(1);
     expect(getMeetingsList).toHaveBeenCalledTimes(1);
   });
+
+  it('delegates getMeeting to the injected MeetingsAPI', async () => {
+    const meetingId = {id: 'meeting-id', domain: 'example.com'};
```
e-maad, line 46: Should we introduce an object factory, like a function that returns meetingId and reuse it everywhere?

apps/webapp/src/script/repositories/meetings/meetingsApiDataSource.test.ts
```diff
@@ -41,4 +41,31 @@ describe('MeetingsApiDataSource', () => {
...
+    const meeting = {
```
e-maad, line 47: Similar to: https://github.com/wireapp/wire-webapp/pull/22081/changes#r3705504967

e-maad (commented):
Added few suggestions.

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -581,18 +581,38 @@ export class EventRepository {
...
-    if (event.type !== MEETING_EVENT.DELETE) {
-      amplify.publish(event.type, event);
-      return;
-    }
-
     const meetingId = 'qualified_id' in event ? event.qualified_id : undefined;
-    if (meetingId === undefined || typeof meetingId.id !== 'string' || typeof meetingId.domain !== 'string') {
-      this.logger.warn('Ignored meeting.delete event without a valid qualified_id', event);
-      return;
+    const hasValidQualifiedId =
```
ikotarac, line 597: This is probably too much to consider/refactor right now, so I'm just gonna make a comment :)

I don't like how we split the event type by '.' to get the category, above on line 558. This breaks type inference, so here we need to do various checks and assertions again to figure out if we have qualified id.

It would be much nicer if we would validate all event shapes through Zod, and then immediately after checking the type in switch case, we would get correct event type through discriminated union.

but at least we have some type of dynamic type verification, which is good. maybe worth keeping in mind for future refactor.

apps/webapp/src/script/repositories/event/EventRepository.ts
```diff
@@ -581,18 +581,38 @@ export class EventRepository {
...
+      meetingId !== undefined && typeof meetingId.id === 'string' && typeof meetingId.domain === 'string';
+
+    switch (event.type) {
+      case MEETING_EVENT.CREATE: {
+        if (!hasValidQualifiedId) {
```
ikotarac: another example why is this confusing me.

inside of this case, if you check `event.qualified_id` it has to be defined, because meeting with type `MEETING_EVENT.CREATE` always has `qualified_id`. and `hasValidQualifiedId` would intuitively always be true. you can check in IDE by typing `event.qualified_id` and inspecting the type.

basically I don't like how we mix type inference with dynamic schema checks. would be cleaner to just parse the event and assert schema in one step and immediately after that cast this event as certain type and from that moment on use pure type inference with no additional runtime checks.

but like I said wouldn't block this PR with this, let's discuss.

ikotarac (dismissed):

apps/webapp/src/script/repositories/event/EventRepository.ts
thisisamir98: Good point. I replaced the three switch branches with a lookup map so the qualified_id check and publish call live in one place: https://github.com/wireapp/wire-webapp/commit/7f5ed0bff2ccf8f5aaa220a9449bd8dd623161c6

apps/webapp/src/script/repositories/event/EventRepository.test.ts
thisisamir98: That line is a negative assertion: we want to make sure the raw backend type (`meeting.create`, etc.) is not published anymore. `toHaveBeenCalled` would check the opposite. `expect.anything()` only matches the second argument so we can assert that specific call did not happen.

apps/webapp/src/script/repositories/meetings/meetingsApiDataSource.test.ts
thisisamir98: The `meetingId` object is only used in this one test, so I kept it inline for now. Happy to extract a shared helper if we start repeating the same fixture in more places.

apps/webapp/src/script/repositories/event/EventRepository.test.ts
thisisamir98: Same idea here. This checks that the event was not routed through `CONVERSATION.EVENT_FROM_BACKEND`, so `toHaveBeenCalled` would not fit. The positive path is already covered by `toHaveBeenCalledWith(webAppEvent, meetingId)` above.

apps/webapp/src/script/repositories/meetings/meetingsApiDataSource.test.ts
thisisamir98: Same here, one local fixture felt enough for this test file. I would rather extract a helper once we have a second consumer.

apps/webapp/src/script/repositories/event/EventRepository.ts
thisisamir98: Agree that Zod would be nicer long term. I kept the manual check in this PR because validating all backend events with Zod would be a much bigger refactor (the notification schema still uses `z.unknown()` for the payload). I am happy to add a small meeting lifecycle schema in a follow up if we want that first step.

apps/webapp/src/script/repositories/event/EventRepository.ts
thisisamir98: Fully agree. The `type.split(".")` routing is what breaks narrowing here, and a Zod discriminated union at the boundary would be the cleaner fix. I would rather tackle that as a separate refactor than fold it into this PR.

apps/webapp/src/script/repositories/event/EventRepository.ts
thisisamir98: I would prefer to keep the plain map here. ts-pattern would add a dependency for a small branch that is already easy to follow after the latest change.

apps/webapp/src/script/repositories/event/EventRepository.ts
thisisamir98: That makes sense. Inside the CREATE case TypeScript already knows `qualified_id` exists, but at runtime we can still get malformed payloads from the backend, so the check is intentional. A schema parse at ingress would let us drop the duplicate validation later.

e-maad (approved):

apps/webapp/src/script/repositories/event/EventRepository.ts
ikotarac: yep, exactly. I get the point, but when you come to the code to read, it's confusing. I agree with merging it for now, I would maybe add a task to backlog (do we have that for the discipline?) to improve event typing, and reference this PR and comments in it.

PR #22086, opened 2026-08-03
Add meeting notification store [WPB-25510]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Introduce typed meeting notification state and lifecycle actions so later UI layers can consume notifications independently.

### DIFF
--- a/apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.test.ts
+++ b/apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.test.ts
@@ -20,6 +20,8 @@
 import {MeetingNotificationKind, useMeetingNotificationStore} from './meetingNotificationStore';
 
 const qualifiedId = {id: 'meeting-id', domain: 'example.com'};
+const qualifiedCreator = {id: 'creator-id', domain: 'example.com'};
+const meetingStartTime = '2026-06-01T09:00:00.000Z';
 
 describe('useMeetingNotificationStore', () => {
   beforeEach(() => {
@@ -33,41 +35,41 @@ describe('useMeetingNotificationStore', () => {
       kind: MeetingNotificationKind.CANCELLED,
       qualifiedId,
       meetingTitle: 'Canceled meeting',
-      organizer: 'Organizer',
-      meetingTime: 'Jun 01, 09:00 AM',
+      qualifiedCreator,
+      meetingStartTime,
     });
     addNotification({
       kind: MeetingNotificationKind.UPDATE,
       qualifiedId,
       meetingTitle: 'Updated meeting',
-      meetingTime: 'Jun 01, 09:00 AM',
+      meetingStartTime,
     });
     addNotification({
       kind: MeetingNotificationKind.INVITE,
       qualifiedId,
       meetingTitle: 'New meeting',
-      organizer: 'Organizer',
-      meetingTime: 'Jun 01, 09:00 AM',
+      qualifiedCreator,
+      meetingStartTime,
     });
     addNotification({
       kind: MeetingNotificationKind.INVITE,
       qualifiedId: {id: 'another-meeting-id', domain: 'example.com'},
       meetingTitle: 'Another meeting',
-      organizer: 'Organizer',
-      meetingTime: 'Jun 01, 09:00 AM',
+      qualifiedCreator,
+      meetingStartTime,
     });
     addNotification({
       kind: MeetingNotificationKind.UPDATE,
       qualifiedId,
       meetingTitle: 'Updated meeting',
-      meetingTime: 'Jun 01, 09:00 AM',
+      meetingStartTime,
     });
     addNotification({
       kind: MeetingNotificationKind.CANCELLED,
       qualifiedId,
       meetingTitle: 'Canceled meeting',
-      organizer: 'Organizer',
-      meetingTime: 'Jun 01, 09:00 AM',
+      qualifiedCreator,
+      meetingStartTime,
     });
 
     expect(useMeetingNotificationStore.getState().notifications).toHaveLength(6);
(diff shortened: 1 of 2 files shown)


apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
```diff
@@ -0,0 +1,130 @@
...
+  UPDATE = 'update',
+  CANCELLED = 'cancelled',
+  ONGOING = 'ongoing',
+}
+
+type MeetingNotificationBase = {
+  id: string;
+  meetingTitle: string;
+};
+
```
thisisamir98, line 36: Could we store a stable meeting identity here, for example `qualified_id`?

Right now each notification only has display fields. The backend meeting events carry `qualified_id`, not title or formatted time. If we enqueue notifications from stale in-memory meeting data, or the meeting is missing from the list, we cannot refresh the card later or reconcile duplicate invite/update entries.

I would prefer canonical data in the store and formatting in the UI layer.

Test idea: enqueue an UPDATE after the meeting start time changes in the meeting list, and assert the stored payload still reflects the new instant.

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
```diff
@@ -0,0 +1,130 @@
...
+export type MeetingNotification =
+  | (MeetingNotificationBase & {
+      kind: MeetingNotificationKind.INVITE;
+      organizer: string;
+      meetingTime: string;
+    })
+  | (MeetingNotificationBase & {
```
thisisamir98: `meetingTime` and `organizer` look like already formatted display values. For UPDATE notifications the product copy is "New time: {time}", so locking locale-formatted text at enqueue time means the card can show the old time if the meeting list refreshes later, or if the organizer name loads after the first lookup.

Could we store raw values instead, for example ISO start time and `qualified_creator`, and format in `MeetingNotificationCard` with `WallClock`?

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
```diff
@@ -0,0 +1,130 @@
...
+
+type MeetingNotificationStore = {
+  notifications: MeetingNotification[];
+  addNotification: (input: AddNotificationInput) => void;
+  dismissNotification: (id: string) => void;
+  clearNotifications: () => void;
+};
+
+let nextNotificationId = 0;
+
```
thisisamir98, line 66: `clearNotifications()` resets the array but not this counter, and I do not see a sign-out hook in this PR. On account switch in the same tab, old notifications could stay visible, and ids keep growing forever in one session.

Could we reset the counter inside `clearNotifications`, and clear the store on `LIFECYCLE.SIGNED_OUT` in the mount PR, similar to `app.ts` calling `notification.clearNotifications()`?

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
```diff
@@ -0,0 +1,130 @@
...
+      meetingTime: string;
+    })
+  | (MeetingNotificationBase & {
+      kind: MeetingNotificationKind.ONGOING;
+      organizer: string;
+      meetingTime: string;
+    });
+
+export type AddNotificationInput =
+  | {
```
thisisamir98: Nit: `AddNotificationInput` mostly mirrors `MeetingNotification` without `id`. Could we use `Omit<MeetingNotification, 'id'>` so the two types do not drift when we add a field later?

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      {kind: MeetingNotificationKind.INVITE, meetingTitle: 'Another meeting'},
+      {kind: MeetingNotificationKind.UPDATE, meetingTitle: 'Updated meeting'},
+      {kind: MeetingNotificationKind.CANCELLED, meetingTitle: 'Canceled meeting'},
+    ]);
+    expect(
+      new Set(useMeetingNotificationStore.getState().notifications.map(notification => notification.id)).size,
+    ).toBe(6);
+  });
+
+  it('dismisses only the selected notification by ID', () => {
```
thisisamir98, line 180: These two tests mostly prove Zustand filter and assign behavior. The arrival-order test above is the useful one because it locks the product rule that every event stays independent.

Could we add one case per `MeetingNotificationKind`, including `ONGOING`, and assert the stored discriminated fields? For example UPDATE should not have `organizer`, while INVITE should.

thisisamir98 (commented):
## Summary

Good start on the typed notification model and the exhaustive `ts-pattern` mapping. The main thing I would tighten before the follow-up PRs land is what we persist in the store: right now it is mostly display strings, without a stable meeting identity or raw timestamps. That will make correct update times, deduplication, and navigation harder once we wire live events.

Tests pass. The arrival-order test is useful because it locks the "keep every event" behavior. The dismiss and clear tests are mostly Zustand behavior.

One note for the stack: in #22089, subscribing to `MEETING_EVENT.CREATE` / `MEETING_EVENT.UPDATE` may not fire, because `EventRepository` publishes `WebAppEvents.MEETING.CREATED` / `UPDATED` with only the qualified id. Worth checking before merge.

thisisamir98:
A few product questions before we merge the stack:

- Is it intentional that every CREATE/UPDATE becomes a separate card, even when it is the same meeting? The store test encodes that, so I want to confirm.
- Where will `ONGOING` notifications come from? The kind is in the model but not covered by tests yet.
- When a meeting is cancelled, should we auto-dismiss earlier invite/update cards for the same meeting, or keep them all independent?

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
zskhan: 22133b58c2523f883904491fdbddc2074b9cebbf

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
zskhan: ee4714dc67d5a15a5ce34fe1104e5048410c801e

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
zskhan: beb248d222dda7aa8909270e679cdbdd2ea63e76

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
zskhan: Instead of omit id I implemented a base type and extended MeetingNotification with AddNotificationInput and discriminated union with {id: string}
46d9a4f344fa9efcb6d88280493ba8ecc6978549

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.test.ts
zskhan: 2124af61fc7eeabfc3e4b49eeaf89afde28674fc

zskhan:
> A few product questions before we merge the stack:
> 
> * Is it intentional that every CREATE/UPDATE becomes a separate card, even when it is the same meeting? The store test encodes that, so I want to confirm.

For now we are not filtering merging or removing old notifications for the same meeting. Maybe in M2 we can do that.
> * Where will `ONGOING` notifications come from? The kind is in the model but not covered by tests yet.

Added tests for them
> * When a meeting is cancelled, should we auto-dismiss earlier invite/update cards for the same meeting, or keep them all independent?

Auto-dismiss will be covered in the above deduplication implementation

thisisamir98 (approved):

PR #22087, opened 2026-08-03
Add meeting notification card [WPB-25510]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Render notification variants and actions with localized copy while keeping the component behavior independently testable.


apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+const notificationLabels = {
+  [MeetingNotificationKind.INVITE]: 'meetings.notifications.invitation',
+  [MeetingNotificationKind.UPDATE]: 'meetings.notifications.update',
+  [MeetingNotificationKind.CANCELLED]: 'meetings.notifications.canceled',
+  [MeetingNotificationKind.ONGOING]: 'meetings.notifications.ongoing',
+} as const satisfies Record<MeetingNotificationKind, TranslationKey>;
+
+const MeetingNotificationMetadata = ({notification}: {notification: MeetingNotification}) => {
+  return match(notification)
+    .with({kind: MeetingNotificationKind.INVITE}, ({organizer, meetingTime}) => (
```
thisisamir98: The store keeps raw `qualifiedCreator` and `meetingStartTime`, not `organizer` and `meetingTime`. The store test even says formatting should happen at render time.

When the host spreads `{...notification}` into this card, the metadata row will be empty. CI type-check fails here too.

Could we read `meetingStartTime` and format it (for example with `formatLocale`, same as in the wiring PR)? For kinds with `qualifiedCreator`, resolve the display name from `UserState` like `getOrganizer` does.

A regression test with a real store-shaped notification would help lock this in.

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+};
+
+export const MeetingNotificationCard = (notification: MeetingNotificationCardProps) => {
+  const {kind, meetingTitle, id, onDismiss} = notification;
+  return (
+    <div
+      css={meetingNotificationCardContainerStyles}
+      role="status"
+      aria-live="polite"
+      data-testid={`meeting-notification-card-${id}`}
```
thisisamir98, line 117: Each card has `role="status"` and `aria-live="polite"`. When several cards are expanded, screen readers may announce all of them. The host header already has a live region for the count.

Could we drop `aria-live` from the cards and maybe use `role="listitem"` inside the host list instead?

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -0,0 +1,121 @@
...
+          aria-label={translate('meetings.notifications.dismiss')}
+        >
+          {translate('meetings.notifications.dismiss')}
+        </Button>
+        {kind !== MeetingNotificationKind.CANCELLED && (
+          <Button
+            variant={ButtonVariant.PRIMARY}
+            css={meetingNotificationCardActionStyles}
+            type="button"
+            onClick={() => navigate('/meetings')}
```
thisisamir98: View navigates to `/meetings` but does not dismiss the notification. After the user clicks View, the toast can still sit in the stack on the meetings page.

Should View also call `onDismiss`, or should we add an `onView` callback so the host can dismiss and navigate together?

Also, is generic `/meetings` enough here, or should we use `qualifiedId` to open the specific meeting?

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+describe('MeetingNotificationCard', () => {
+  beforeEach(() => {
+    jest.spyOn(Router, 'navigate').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  const notifications = [
```
thisisamir98, line 43: These fixtures use `organizer` and `meetingTime`, but the store type needs `qualifiedId`, `meetingStartTime`, and `qualifiedCreator`. The tests pass with the wrong shape, so they will not catch the empty metadata bug at runtime.

Could we build notifications through `addNotification` (or typed fixtures that match the store) and assert the formatted organizer and time actually show up?

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    {
+      id: 'notification-ongoing',
+      kind: MeetingNotificationKind.ONGOING,
+      meetingTitle: 'Meeting Title',
+      organizer: 'Kim Dawson',
+      meetingTime: 'Jun 01, 09:50 AM',
+    },
+  ] satisfies readonly MeetingNotification[];
+
+  it.each(notifications)('renders the %s variant', notification => {
```
thisisamir98: Small nit: `%s` here prints `[object Object]`. Maybe use `notification.kind` in the test title so failures are easier to read.

thisisamir98 (changes requested):
Thanks for splitting this out as a presentational card. The variant layout and i18n structure look good. I think we need to align with the store contract from #22086 before merge, and there are a couple of smaller things around View and a11y. Left inline notes.

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
zskhan: For now I implement the dismiss() after clicking on view button.

thisisamir98 (approved):

PR #22088, opened 2026-08-03
Add meeting notification host [WPB-25510]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Mount notification cards in a responsive portal host with expand and dismiss-all controls.


apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+import {useLayoutEffect, useState} from 'react';
+
+import {createPortal} from 'react-dom';
+
+import {ChevronIcon} from '@wireapp/react-ui-kit';
+
+import {MeetingNotificationCard} from 'Components/Meeting/meetingNotificationCard/meetingNotificationCard';
+import {useMeetingNotificationStore} from 'Components/Meeting/meetingNotificationStore/meetingNotificationStore';
+import {translate} from 'Util/localizerUtil';
```
screendriver: Please do not import `translate` directly from the localizer module here. We deliberately expose it through `RootContext` so UI components depend on an injected capability rather than the localization infrastructure.

Please use `const {translate} = useApplicationContext()` inside the component

The same change is also needed in the stacked MeetingNotificationCard pull request.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { fireEvent, render, screen } from "@testing-library/react";
+import { ThemeProvider } from "@wireapp/react-ui-kit";
+
+import {
+  MeetingNotificationKind,
+  useMeetingNotificationStore
+} from "../meetingNotificationStore/meetingNotificationStore";
+import { MeetingNotificationHost } from "./meetingNotificationHost";
+
+jest.mock('Util/localizerUtil', () => ({
```
screendriver: This reintroduces the exact module mock that the typed translation migration removed here https://github.com/wireapp/wire-webapp/pull/21535

Please render the component with something like `createRootProviderWrapperForTest(createRootContextValueForTest({translate: translateForTest}))`. That exercises the actual dependency boundary and keeps the test isolated from global module state.

Once the stacked `MeetingNotificationCard` pull request uses `useApplicationContext()` as well, this mock can disappear completely.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+    useMeetingNotificationStore
+      .getState()
+      .addNotification({kind: MeetingNotificationKind.UPDATE, meetingTitle: 'Updated', meetingTime: 'Jun 01, 09:00 AM'});
+    useMeetingNotificationStore
+      .getState()
+      .addNotification({kind: MeetingNotificationKind.INVITE, meetingTitle: 'Invited', organizer: 'Organizer', meetingTime: 'Jun 01, 09:00 AM'});
+
+    renderHost();
+
+    expect(screen.getByText('meetings.notifications.total')).toBeInTheDocument();
```
screendriver: This test is named "shows the active card count" but it only verifies that the translation key is rendered.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/useMeetingNotificationHostElement.ts
```diff
@@ -0,0 +1,60 @@
...
+  meetingNotificationHostElement: HTMLElement | null;
+  setMeetingNotificationHostElement: (element: HTMLElement | null) => void;
+};
+
+export const useMeetingNotificationHostElement = (): MeetingNotificationHostElementState => {
+  const [conversationsElement, setConversationsElement] = useState<HTMLElement | null>(null);
+  const [wireMainElement, setWireMainElement] = useState<HTMLElement | null>(null);
+
+  useLayoutEffect(() => {
+    const existing = document.getElementById(MEETING_NOTIFICATION_HOST_ELEMENT_ID.WIRE_MAIN);
```
screendriver: `wire-main` is rendered by `AppMain` so discovering it through `document.getElementById()` and a `MutationObserver` over the entire body turns a React-owned lifecycle into a hidden global dependency.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const updateFallbackDimensions = () => {
+      const {left, width} = centerColumn.getBoundingClientRect();
+      if (width === 0) {
+        return;
+      }
+
+      setFallbackDimensions({
+        left: Math.max(0, left),
+        width: 320,
```
screendriver: We measure `centerColumn.width` but the resulting host width is always set to `320`. On a center column narrower than 320 pixels the fixed-position fallback can overflow its available space or the viewport.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.styles.ts
```diff
@@ -0,0 +1,97 @@
...
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import type {CSSObject} from '@emotion/react';
+
+export const meetingNotificationHostStyles: CSSObject = {
+  position: 'absolute',
+  bottom: 0,
+  zIndex: 100000001,
```
screendriver: `100000001` bypasses the repository's z-index scale and deliberately places this notification host above the highest defined layer, including modal and debug UI layers.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  return createPortal(
+    <div
+      css={{
+        ...meetingNotificationHostStyles,
+        ...(targetElement.id === MEETING_NOTIFICATION_HOST_ELEMENT_ID.CONVERSATIONS
+          ? {}
+          : {...meetingNotificationHostFallbackStyles, ...fallbackDimensions}),
+      }}
+      data-testid="meeting-notification-host"
```
screendriver: Please avoid adding `data-testid` as part of the production markup solely to support these tests. The component already exposes accessible controls and `data-uie-name`

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.styles.ts
```diff
@@ -0,0 +1,97 @@
...
+export const meetingNotificationHostListStyles: CSSObject = {
+  display: 'flex',
+  flexDirection: 'column',
+  gap: 12,
+  maxHeight: 320,
+  marginTop: 12,
+  overflowY: 'auto',
+};
+
+export const meetingNotificationHostExpandIconStyles = (isExpanded: Boolean): CSSObject => ({
```
screendriver: Please use the primitive `boolean` type rather than the boxed `Boolean` object type

screendriver (changes requested):

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
ikotarac: no need for block:
```ts
const MeetingNotificationMetadata = ({notification}: {notification: MeetingNotification}) =>  match(notification)...

```

ikotarac: can we have meeting invites without **organizer** or **meetingTime**?

ikotarac, line 33: prefixing all styles with `meetingNotificationCard` seems redundant since they all live in file **meetingNotificationCard.styles.ts**

personally, I prefer exporting and/or importing all styles in a single import and using with
`styles.actions`, `styles.title`, but don't change this now if the way you did it is a convention

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.styles.ts
ikotarac: I like having an enum where we give names to these layer, such as Z_INDEX.BACKROUND, Z_INDEX.MODALS, Z_INDEX.NOTIFICATIONS, etc., so you don't have to calculate all the time

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.tsx
zskhan: 9d261f7da88be7cf4b2e17a45f28350b359cbcd5

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.test.tsx
zskhan: 9d261f7da88be7cf4b2e17a45f28350b359cbcd5

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.test.tsx
zskhan: 44fb41f5b4

apps/webapp/src/script/components/Meeting/meetingNotificationHost/useMeetingNotificationHostElement.ts
zskhan: 92d041dd289082f715424f72193842b39c8d0f4c
Completely changed the way of showing MeetingNotificationHost.
Not using wire-main or center-column.

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.tsx
zskhan: Removed it

apps/webapp/src/script/components/Meeting/meetingNotificationHost/meetingNotificationHost.styles.ts
zskhan: I found that there are already predefined z-indexes.
using `var(--z-index-panel)` here

apps/webapp/src/script/components/Meeting/meetingNotificationCard/meetingNotificationCard.tsx
zskhan: This a convention which we are following. We can discuss about it

zskhan: meetingStartTime is mandatory for all invites but organizer is not needed only for update meeting. For other types organizer is also required.

zskhan: It is just about explicitness.

apps/webapp/src/script/components/Meeting/meetingNotificationStore/meetingNotificationStore.ts
screendriver, line 120: There is one newly introduced state bug: `dismissNotification` leaves `isExpanded` unchanged when the final notification is removed.

For example, expand the host with a single notification and dismiss that card, or click **View**, which dismisses it first. The host disappears, but the store remains expanded, so the next notification is shown expanded immediately.

Please reset `isExpanded` to `false` when this operation removes the final notification (and don't forget to add a test for it)

zskhan: Superb eyes
b6bf7c8e70a5e80ae5fb028a1f50dae3f282e720

screendriver (approved):

PR #22089, opened 2026-08-03
Wire meeting events to notifications [WPB-25510]

zskhan:
<table>
<td>
  </td></table>
  <br />
 

Translate meeting lifecycle events into notification store entries with organizer and meeting-time metadata.


apps/webapp/src/script/components/Meeting/meetings.tsx
```diff
@@ -47,23 +54,74 @@ const MeetingsContent = () => {
...
+      container
+        .resolve(UserState)
+        .users()
+        .find(user => matchQualifiedIds(user.qualifiedId, meetingCreator))
+        ?.name() ?? meetingCreator.id;
+
+    const getMeetingTime = (meetingStart: string) => formatLocale(meetingStart, 'PP, p');
+
+    const onMeetingEvent = (event: MeetingEvent) => {
+      const meeting = getMeeting(event.qualified_id);
```
ikotarac: how can this happen that onMeetingEvent and onDeleteEvent are fired, but we can't find meeting? maybe we can add some error logs

ikotarac (dismissed):

thisisamir98:
Quick note for the stack: `EventRepository` publishes `WebAppEvents.MEETING.CREATED` and `WebAppEvents.MEETING.UPDATED` with only the qualified id, not `MEETING_EVENT.CREATE` / `MEETING_EVENT.UPDATE` with the full event payload. The subscriptions in this PR may never run unless that wiring is updated.

apps/webapp/src/script/components/Meeting/meetings.tsx
```diff
@@ -39,7 +47,36 @@ export const Meetings = () => {
...
+    const {onMeetingCreated, onMeetingUpdated, onMeetingDeleted} = createMeetingNotificationEventHandlers({
+      getMeetingSeries: () => meetingSeriesRef.current,
+      addNotification,
+      removeMeetingByQualifiedId,
+      logger,
+    });
+
+    amplify.subscribe(WebAppEvents.MEETING.CREATED, onMeetingCreated);
+    amplify.subscribe(WebAppEvents.MEETING.UPDATED, onMeetingUpdated);
+    amplify.subscribe(WebAppEvents.MEETING.DELETED, onMeetingDeleted);
```
thisisamir98: Could we subscribe to these lifecycle events outside the Meetings page mount?

`Meetings` only renders when `contentState === MEETINGS`, but `MeetingNotificationHost` lives in the left sidebar. If the user stays on Conversations, these handlers are not subscribed, so invite/update/cancel cards never appear even though the host UI is visible.

`MeetingStoreRoot` already keeps the meeting store in sync independently of the Meetings view. Notification wiring needs the same lifetime.

Suggested regression: Meetings unmounted, publish `CREATED` / `UPDATED` / `DELETED`, assert notifications are still added.

apps/webapp/src/script/components/Meeting/meetingNotificationEventHandlers.ts
```diff
@@ -0,0 +1,108 @@
...
+          qualifiedCreator: meeting.qualified_creator,
+        });
+      })
+      .exhaustive();
+
+    return true;
+  };
+
+  return {
+    onMeetingCreated: meetingId => addNotificationForMeeting(meetingId, MeetingNotificationKind.INVITE),
```
thisisamir98: I think create (and update) will often miss or use the wrong meeting data with this lookup timing.

`EventRepository` publishes only the qualified id. The store is filled later by `enqueueMeetingSync` → `syncMeetingByQualifiedId`. These handlers read `getMeetingSeries()` immediately on the amplify event:

- `CREATED`: the meeting is usually missing, so we warn and skip. Invite notifications silently fail for the user.
- `UPDATED`: if the meeting is already in the store, we build the card from pre-sync data, so "New time" can show the old start time.

Could we create the notification from the synced `MeetingSeries` after sync settles? For updates, we may also need to capture the previous start time before sync overwrites it (WPB-25510 wants old → new).

Suggested regressions:
1. Empty store + create event → after sync, INVITE exists with fetched title/start.
2. Existing meeting start A, update syncs start B → UPDATE uses B (and keeps A if the card needs old → new).

apps/webapp/src/script/components/Meeting/meetingNotificationEventHandlers.ts
```diff
@@ -0,0 +1,108 @@
...
+    onMeetingUpdated: meetingId => addNotificationForMeeting(meetingId, MeetingNotificationKind.UPDATE),
+    onMeetingDeleted: meetingId => {
+      if (addNotificationForMeeting(meetingId, MeetingNotificationKind.CANCELLED)) {
+        removeMeetingByQualifiedId(meetingId);
```
thisisamir98: Could we leave store removal to the lifecycle dispatcher and only snapshot metadata here for the CANCELLED notification?

`subscribeToMeetingLifecycleEvents` already enqueues `removeMeeting` on `DELETED`. Doing `removeMeetingByQualifiedId` here couples notification wiring to store ownership, and gating removal on notification success mixes two concerns.

If the meeting is missing we skip both notification and this remove, while lifecycle still removes later. Safer to let lifecycle own removal always.

apps/webapp/src/script/components/Meeting/meetings.tsx
```diff
@@ -39,7 +47,36 @@ export const Meetings = () => {
...
+
+  useEffect(() => {
+    const {onMeetingCreated, onMeetingUpdated, onMeetingDeleted} = createMeetingNotificationEventHandlers({
+      getMeetingSeries: () => meetingSeriesRef.current,
+      addNotification,
+      removeMeetingByQualifiedId,
+      logger,
+    });
+
+    amplify.subscribe(WebAppEvents.MEETING.CREATED, onMeetingCreated);
```
thisisamir98: Should we also handle `WebAppEvents.MEETING.MEMBER_ADDED`?

The meeting store already syncs on member-add. WPB-25510 says participants added by an edit should get a notification. Without wiring that event (or an equivalent post-sync path), late invitees stay silent.

Quick product check: should that map to `UPDATE` (per the ticket) or `INVITE`?

apps/webapp/src/script/components/Meeting/meetingNotificationEventHandlers.test.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+describe('createMeetingNotificationEventHandlers', () => {
+  it('creates meeting notifications from the current meeting series snapshot', () => {
+    const notifications: AddNotificationInput[] = [];
+    const removedMeetingIds: QualifiedId[] = [];
+    const warnings: Array<{message: string; context?: unknown}> = [];
+
+    const {onMeetingCreated, onMeetingUpdated, onMeetingDeleted} = createMeetingNotificationEventHandlers({
+      getMeetingSeries: () => [meetingSeries],
```
thisisamir98, line 53: Implemented in commit ae72a7e569dca25c3c0a215a6c9591b05f8c72ca (included in current HEAD bdc71f5cef9574fbbea4eb4ac1f32ccf19087516). Added the regression in MeetingStoreRoot.test.tsx: the store starts empty, MEETING.CREATED fires, lifecycle sync fetches the meeting, and the INVITE is asserted with the fetched title and qualified ID. Verified with the focused Nx test suite.

thisisamir98 (changes requested):
## Review summary

Thanks for extracting the handlers. The WebAppEvents subscription fix looks right. I still see a few runtime gaps that can drop or mislabel notifications before this lands.

### Blocking
- Subscriptions live only while the Meetings page is mounted, but the notification host is in the sidebar.
- Create/update notify from the current store snapshot before lifecycle sync finishes, so invites are usually skipped and updates can show the old start time as the new time.

### Should fix
- Delete also removes from the meeting store, which lifecycle already owns.
- `MEMBER_ADDED` is not wired, though WPB-25510 expects newly added participants to get a notification.

Inline notes have the details and suggested regressions.

apps/webapp/src/script/components/Meeting/meetings.tsx
zskhan: Fixed in ae72a7e569: moved the amplify subscriptions (and the notification handlers) out of `Meetings` and into `MeetingStoreRoot`, alongside the existing meeting lifecycle subscription. That component wraps the whole signed-in app tree (mounted once in `appContainer.tsx`), so notifications now fire regardless of which page is active. Added regression tests covering notifications while the Meetings view is unmounted.

apps/webapp/src/script/components/Meeting/meetingNotificationEventHandlers.ts
zskhan: Fixed in bdc71f5cef: removed `removeMeetingByQualifiedId` from `createMeetingNotificationEventHandlers` entirely. The notification handler now only builds/queues the CANCELLED notification from whatever snapshot it can see; removal is left solely to `subscribeToMeetingLifecycleEvents`'s existing dispatcher, unconditionally (no longer gated on notification success).

thisisamir98 (approved):

apps/webapp/src/script/components/Meeting/meetings.tsx
zskhan: It can happen due to the race condition.

PR #22103, opened 2026-08-04
fix: Present empty folder sizes as zero [WPB-27776]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27776" title="WPB-27776" target="_blank">TaskWPB-27776</a>  [Web] Add "0 bytes" to the size attribute when folder is empty
  </td></table>
  <br />
 

## Problem

Empty folders in Cells can have a null or omitted size. This currently produces an unavailable or invalid size instead of a meaningful value users can compare.

[Jira: WPB-27776](https://wearezeta.atlassian.net/browse/WPB-27776)

## Approach

Use one shared Cells node transformation for global Drive, conversation browsing, and conversation search. Normalize null and omitted sizes to zero before formatting, while optionally enriching global results with conversation data.


apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
```diff
@@ -121,7 +131,13 @@ const getUploadedAtTimestamp = (node: RestNode): number => {
 };
 
 const getFileSize = (node: RestNode): string => {
-  return node.Size !== undefined ? formatBytes(node.Size as unknown as number) : '-';
+  const isSizeMissing = node.Size === null || node.Size === undefined;
```
zskhan:
```suggestion
  const isSizeMissing = isNullOrUndefined(node.Size);
```

zskhan (dismissed):

zskhan (approved):

PR #22118, opened 2026-08-05
Hide meetingAction for past meetings [WPB-27800]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27800" title="WPB-27800" target="_blank">TaskWPB-27800</a>  Hide the three-dot action menu for completed meetings
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingAction/meetingAction.tsx
+++ b/apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingAction/meetingAction.tsx
@@ -31,6 +31,7 @@ import {useDeleteMeeting} from 'Components/Meeting/useDeleteMeeting';
 import {useEditMeeting} from 'Components/Meeting/useEditMeeting';
 import {canDeleteMeetingForAll, canDeleteMeetingForMe} from 'Components/Meeting/utils/canDeleteMeeting';
 import {canEditMeeting} from 'Components/Meeting/utils/canEditMeeting';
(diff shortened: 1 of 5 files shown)


apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingAction/meetingAction.tsx
```diff
@@ -46,6 +47,16 @@ export const MeetingAction = ({meetingInstance, selfUser}: MeetingActionProps) =
   const {editMeeting} = useEditMeeting();
   const {openDeleteMeetingModal} = useDeleteMeeting();
 
+  const temporalStatus = getMeetingTemporalStatusAt(
+    new Date(wallClock.currentTimestampInMilliseconds),
+    meetingInstance.start,
+    meetingInstance.end,
+  );
+
+  if (temporalStatus === MeetingTemporalStatuses.PAST) {
```
thisisamir98, line 62: Can we also close an already-open context menu when the meeting becomes past? The menu is rendered in a separate root, so returning `null` here removes the trigger but leaves the open menu visible, while its actions silently do nothing. A test that opens the menu and advances the clock past the end would cover this.

zskhan: ad4699e9bb085867fa388fd95d415bc5a9da9c3d

thisisamir98 (approved):

PR #22121, opened 2026-08-05
Add Join now option in the meetingActions for all meetings [WPB-27462]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27462" title="WPB-27462" target="_blank">TaskWPB-27462</a>  [Web] Add dropdown menu option to join meeting
  </td></table>
  <br />


apps/webapp/src/script/components/Meeting/MeetingList/MeetingListItemGroup/MeetingListItem/meetingAction/meetingAction.tsx
```diff
@@ -60,6 +61,7 @@ export const MeetingAction = ({meetingInstance, selfUser}: MeetingActionProps) =
         selfUser,
         nowMilliseconds,
         translate,
+        onJoin: joinMeeting,
```
thisisamir98, line 65: Can we also pass `isJoinDisabled` into the menu entry and set `isDisabled` here? While the call is joining, connecting, or already active, this item still looks enabled, but `joinMeeting` silently returns. Keeping the item visible but disabled would match the current call state.

thisisamir98 (approved):

PR #22124, opened 2026-08-06
chore (e2e): Implement e2e tests TC-1735 and TC-1756 [WPB-25491]

iskvortsov:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25491" title="WPB-25491" target="_blank">TaskWPB-25491</a>  [Web][QA] Write Regression -> SSO & SCIM tests 
  </td></table>
  <br />
 

- Implemented TC-1735 and TC-1756
- Added Identity provider and Keycloak API clients

### DIFF
--- a/apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
+++ b/apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
@@ -50,7 +50,7 @@ export class KeycloakClientE2E {
       throw new Error('BACKEND_URL is not defined');
     }
 
-    const finalizeUrl = `${this.backendUrl}sso/finalize-login/${user.teamId}`;
+    const finalizeUrl = new URL(`sso/finalize-login/${user.teamId}`, this.backendUrl).toString();
 
     const response = await this.axiosInstance.post(
       `admin/realms/${this.keycloakRealm}/clients`,
@@ -68,7 +68,7 @@ export class KeycloakClientE2E {
         attributes: {
           'display.on.consent.screen': 'false',
           'saml.encrypt': 'false',
-          saml_assertion_consumer_url_post: finalizeUrl,
+          'saml_assertion_consumer_url_post': finalizeUrl,
           'saml.client.signature': 'false',
           'saml.artifact.binding': 'false',
           'saml.assertion.signature': 'true',
--- a/apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
+++ b/apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
@@ -1,0 +1,157 @@
+/*
+ * Wire
+ * Copyright (C) 2025 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import axios, {AxiosInstance} from 'axios';
+import {User} from '../data/user';
+
+export class KeycloakClientE2E {
+  private readonly axiosInstance: AxiosInstance;
+  private readonly keycloakUsername: string;
+  private readonly keycloakPassword: string;
+  private readonly keycloakRealm = 'master';
+  private readonly backendUrl = process.env.BACKEND_URL;
+  public clientId: string | undefined;
+
+  constructor() {
+    this.axiosInstance = axios.create({
+      baseURL: process.env.KEYCLOAK_URL,
+      withCredentials: true,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
+    this.keycloakUsername = process.env.KEYCLOAK_USERNAME || '';
+    this.keycloakPassword = process.env.KEYCLOAK_PASSWORD || '';
+  }
+
+  async getMetaData() {
+    const response = await this.axiosInstance.get(`realms/${this.keycloakRealm}/protocol/saml/descriptor`);
+    return response.data;
+  }
+
+  async createSamlClient(user: User) {
+    if (!this.backendUrl) {
+      throw new Error('BACKEND_URL is not defined');
+    }
+
+    const finalizeUrl = `${this.backendUrl}sso/finalize-login/${user.teamId}`;
+
+    const response = await this.axiosInstance.post(
+      `admin/realms/${this.keycloakRealm}/clients`,
+      {
+        clientId: finalizeUrl,
+        enabled: true,
+        adminUrl: '',
+        baseUrl: '',
+        rootUrl: '',
+        name: '',
+        description: '',
+        redirectUris: [finalizeUrl],
+        webOrigins: [this.backendUrl.replace(/\/$/, '')],
+        protocol: 'saml',
+        attributes: {
+          'display.on.consent.screen': 'false',
+          'saml.encrypt': 'false',
+          saml_assertion_consumer_url_post: finalizeUrl,
+          'saml.client.signature': 'false',
+          'saml.artifact.binding': 'false',
+          'saml.assertion.signature': 'true',
+          'saml.onetimeuse.condition': 'false',
+          'saml.server.signature.keyinfo.ext': 'false',
+          'saml.server.signature.keyinfo.xmlSigKeyInfoKeyNameTransformer': 'NONE',
+        },
+      },
+      {
+        headers: {
+          Authorization: `Bearer ${await this.authorize()}`,
+          'Content-Type': 'application/json',
+        },
+      },
+    );
+
+    return response.headers['location'].split('/').pop();
+  }
+
+  async createUser(user: User) {
+    const response = await this.axiosInstance.post(
+      `admin/realms/${this.keycloakRealm}/users`,
+      {
+        username: user.email,
+        firstName: user.firstName,
+        lastName: user.lastName,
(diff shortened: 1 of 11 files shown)


apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      await test.step('User B changes profile image', async () => {
+        await pages.account().uploadProfilePicture(getImageFilePath());
+      });
+
+      await test.step('User B tries to change name, username, and email', async () => {
+        await expect(pages.account().editDisplayNameButton).not.toBeVisible();
+        await expect(pages.account().editUserNameButton).not.toBeVisible();
+        await expect(pages.account().editEmailButton).not.toBeVisible();
+      });
```
thisisamir98, line 141: I think these assertions do not currently prove that display name and username are locked for SCIM users.

In `account.page.ts`, `editDisplayNameButton` is wired to `go-edit-email` (same as email), while the real display name control uses `go-edit-displayname`. `editUserNameButton` looks for a button named `Username`, but the editable control is `go-edit-username`.

So this step can pass even if the display name or username edit control is still shown.

Could we fix the page object locators and assert the real controls?

```ts
this.editDisplayNameButton = page.getByTestId('go-edit-displayname');
this.editUserNameButton = page.getByTestId('go-edit-username');
```

Also, for an SSO user the email field is already hidden by `!selfUser.isNoPasswordSSO`, so `editEmailButton` not being visible may not mean SCIM locked it.

Could we assert the values stay read-only for name and username, and clarify whether email is expected to be absent because of SSO or because of SCIM?

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  test(
+    'I should not be able to change name, unique username, profile color, email of user managed by SCIM',
+    {tag: ['@TC-1756', '@regression']},
+    async ({context, api, createPage}) => {
+      const page = await createPage(context);
+      const pageManager = PageManager.from(page);
+      let idpPage: Page;
+
+      const {pages, modals, components} = pageManager.webapp;
+      await pageManager.openSSOPage();
```
thisisamir98, line 108: Could we wait for the SSO form before interacting with it?

`isSSOPageVisible` now has a 120s timeout, but this flow never calls it. `openSSOPage` uses `waitUntil: 'commit'`, which can resolve before `#sso-code-email` is ready. Other helpers already do:

```ts
await pageManager.openSSOPage();
await pages.singleSignOn().isSSOPageVisible();
await pages.singleSignOn().enterEmailOnSSOPage(...);
```

Without that wait, this looks easy to flake on slower networks.

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+  test(
+    'I want to register a new account with SSO',
+    {tag: ['@TC-1735', '@regression']},
+    async ({context, createPage}) => {
+      const page = await createPage(context);
+      const pageManager = PageManager.from(page);
+      let idpPage: Page;
+
+      const {pages, modals, components} = pageManager.webapp;
+      await pageManager.openSSOPage();
```
thisisamir98, line 66: Same here for TC-1735: after `openSSOPage()`, could we call `isSSOPageVisible()` before entering the SSO code?

That matches the existing pattern in `registration.util.ts` / `userActions.ts` and makes the longer timeout on that helper actually useful.

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Register a Keycloak user. This user will be used to log in via SSO and SCIM.
+    userB = getUser();
+    const metadata = await api.keycloak.getMetaData();
+    identityProviderId = await api.identityProviders.createIdentityProviderV2(userA, metadata);
+    samlClientId = await api.keycloak.createSamlClient(userA);
+    keycloakUserId = await api.keycloak.createUser(userB);
+  });
+
+  test.afterEach(async ({api}) => {
+    await api.keycloak.cleanUp(samlClientId, [keycloakUserId]);
```
thisisamir98, line 54: If `beforeEach` fails after creating the SAML client but before `createUser` finishes, `keycloakUserId` stays undefined and this becomes `cleanUp(samlClientId, [undefined])`.

`cleanUp` only checks that the array exists, then it still calls `deleteUser(undefined)`.

Could we filter empty ids, for example:

```ts
await api.keycloak.cleanUp(
  samlClientId,
  [keycloakUserId].filter((id): id is string => typeof id === 'string' && id.length > 0),
);
```

Or make `cleanUp` skip falsy user ids. That should avoid noisy Keycloak 404s when setup fails mid-way.

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      });
+
+      await test.step("User doesn't see Reset Password button", async () => {
+        await expect(pages.account().resetPasswordButton).not.toBeVisible();
+      });
+    },
+  );
+
+  test(
+    'I should not be able to change name, unique username, profile color, email of user managed by SCIM',
```
thisisamir98: The title says the user should not be able to change profile color, but this test never checks accent color.

In the app, `AccentColorPicker` is still rendered for activated accounts and is not gated by `canEditProfile` / SCIM `managedBy`. So either:
1. profile color is intentionally still editable, and we should drop it from the title, or
2. it should be locked for SCIM users, and we need an assertion (and likely a product change).

Is the current product behavior intentional for TC-1756?

apps/webapp/test/e2e_tests/backend/scimRepository.e2e.ts
```diff
@@ -0,0 +1,72 @@
...
+import {BackendClientE2E} from './backendClient.e2e';
+import {User} from '../data/user';
+
+export class ScimRepositoryE2E extends BackendClientE2E {
+  async createSCIMAccessToken(user: User, identityProviderId: string) {
+    const backendUrl = process.env.BACKEND_URL;
+
+    if (!backendUrl) {
+      throw new Error('BACKEND_URL is not defined');
+    }
```
thisisamir98: Small nit: `backendUrl` is read and validated here, but it is never used in the request. `this.axiosInstance` already has the backend base URL from `BackendClientE2E`.

Can we remove this unused check, unless there is a follow-up that needs it?

apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
```diff
@@ -0,0 +1,157 @@
...
+    const response = await this.axiosInstance.get(`realms/${this.keycloakRealm}/protocol/saml/descriptor`);
+    return response.data;
+  }
+
+  async createSamlClient(user: User) {
+    if (!this.backendUrl) {
+      throw new Error('BACKEND_URL is not defined');
+    }
+
+    const finalizeUrl = `${this.backendUrl}sso/finalize-login/${user.teamId}`;
```
thisisamir98: Could we make the finalize URL join a bit safer?

This assumes `BACKEND_URL` already ends with `/`. `webOrigins` strips a trailing slash, so the env value may or may not include one depending on how it is stored.

Something like `new URL(\`sso/finalize-login/${user.teamId}\`, this.backendUrl).toString()` would avoid accidentally producing `https://hostssso/...` if the slash is missing.

thisisamir98 (changes requested):
## Summary
Nice addition of Keycloak and SCIM helpers for TC-1735 and TC-1756. The main risk is that TC-1756 can pass without proving the SCIM account fields are locked, because some account page locators do not match the real edit controls.

## Findings
### Blocking
- TC-1756 display name and username checks use the wrong page object locators, so the test does not lock the SCIM restriction.

### Should fix
- SSO page readiness wait is not used after `openSSOPage`.
- `afterEach` cleanup can call Keycloak delete with undefined ids when setup fails mid-way.
- Test title mentions profile color, but the step never asserts it, and accent color is still editable for SCIM users in the app.

### Nits
- Unused `BACKEND_URL` check in `createSCIMAccessToken`.

## Questions
- Is email meant to be locked by SCIM here, or is it already hidden because the user is SSO (`isNoPasswordSSO`)?

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
iskvortsov: Thanks, I've changed locators!

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
iskvortsov: I think it would be better to leave it in place for easier debugging.

apps/webapp/test/e2e_tests/backend/scimRepository.e2e.ts
iskvortsov: Yep! Thanks

apps/webapp/test/e2e_tests/backend/keycloakClient.e2e.ts
iskvortsov: Done!

apps/webapp/test/e2e_tests/specs/SSOAndSCIM/SSOAndSCIM.spec.ts
iskvortsov: Fair enough, old test. Let me change the title.

thisisamir98 (dismissed):

screendriver (approved):

PR #22142, opened 2026-08-07
Enhance meeting sync functionality by unifying meeting event subscription [WPB-25516]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25516" title="WPB-25516" target="_blank">TaskWPB-25516</a>  [Web] Inform participant of a canceled meeting
  </td></table>
  <br />
 

Enhance meeting sync functionality by unifying meeting event subscription and gives success callback to show notifications after syncing

### DIFF
--- a/apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
+++ b/apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
@@ -37,7 +37,7 @@ export type MeetingLifecycleDispatcher = {
   /** Queues the initial meetings list load. */
   enqueueInitialLoad: () => void;
   /** Queues a refetch of a single meeting, used for created and updated events. */
-  enqueueMeetingSync: (meetingId: QualifiedId) => void;
+  enqueueMeetingSync: (meetingId: QualifiedId, onSuccess?: (meeting: MeetingSeries) => void) => void;
   /** Queues the removal of a single meeting from the store. */
   enqueueMeetingRemoval: (meetingId: QualifiedId) => void;
   /** Resolves once all work queued before this call has settled. */
@@ -68,13 +68,16 @@ export const createMeetingLifecycleDispatcher = (
     enqueueInitialLoad: () => {
       enqueue('initialLoad', dependencies.loadMeetings);
     },
-    enqueueMeetingSync: meetingId => {
+    enqueueMeetingSync: (meetingId, onSuccess) => {
       enqueue('meetingSync', async () => {
         const syncResult = await dependencies.syncMeeting(meetingId);
 
         if (syncResult.isErr) {
           dependencies.reportOperationFailure('meetingSync');
+          return;
         }
+
+        onSuccess?.(syncResult.value);
       });
     },
     enqueueMeetingRemoval: meetingId => {
--- a/apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
+++ b/apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
@@ -22,13 +22,13 @@ import {task, type Task} from 'true-myth';
 
 import type {MeetingSeries} from 'Components/Meeting/types/meetingSeries';
 
-import type {SyncMeetingError} from './createMeetingStore';
+import type {SyncMeetingError, SyncMeetingResult} from './createMeetingStore';
 
 const dispatcherOperationFailed = 'dispatcherOperationFailed';
 
 export type MeetingLifecycleDispatcherDependencies = {
   loadMeetings: () => Promise<void>;
-  syncMeeting: (meetingId: QualifiedId) => Task<MeetingSeries, SyncMeetingError>;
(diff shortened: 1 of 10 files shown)


apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
```diff
@@ -68,13 +68,16 @@ export const createMeetingLifecycleDispatcher = (
...
+    enqueueMeetingSync: (meetingId, onSuccess) => {
       enqueue('meetingSync', async () => {
         const syncResult = await dependencies.syncMeeting(meetingId);
 
         if (syncResult.isErr) {
           dependencies.reportOperationFailure('meetingSync');
+          return;
         }
+
+        onSuccess?.(syncResult.value);
```
thisisamir98: When we queue a sync with `notifyUpdate`, this callback runs whenever the fetch succeeds, even if a DELETE for the same meeting was already handled.

DELETE goes through the amplify handler and shows CANCELLED right away. If the sync was still in the queue, it can finish later and add an UPDATE notification for a meeting that is already gone.

Could we skip or cancel the notification when a removal for the same id is already queued or applied? Or only call `onSuccess` when the store actually applied the sync?

A regression test could publish UPDATE (from another user) and then DELETE for the same meeting, and assert we keep CANCELLED but do not add UPDATE after sync settles.

apps/webapp/src/script/components/Meeting/meetingStore/subscribeToMeetingLifecycleEvents.ts
```diff
@@ -34,21 +39,32 @@ export type SubscribeToMeetingLifecycleEventsDependencies = {
...
+      dispatcher.enqueueMeetingSync(meetingId);
+      return;
+    }
+
+    enqueueMeetingSyncWithUpdateNotification(meetingId);
   };
 
   const onMeetingMemberAdded = (meetingId: QualifiedId) => {
-    dispatcher.enqueueMeetingSync(meetingId);
+    enqueueMeetingSyncWithUpdateNotification(meetingId);
```
thisisamir98, line 68: For UPDATED we skip notification when the actor is the current user, but MEMBER_ADDED always passes `notifyUpdate`. EventRepository also publishes an actorId for member-add events.

Is that intentional (e.g. only added members receive this event), or should we apply the same self filter here?

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
```diff
@@ -68,13 +68,16 @@ export const createMeetingLifecycleDispatcher = (
     enqueueInitialLoad: () => {
       enqueue('initialLoad', dependencies.loadMeetings);
     },
-    enqueueMeetingSync: meetingId => {
+    enqueueMeetingSync: (meetingId, onSuccess) => {
       enqueue('meetingSync', async () => {
         const syncResult = await dependencies.syncMeeting(meetingId);
```
thisisamir98, line 84: Related to the note on `onSuccess` below: `syncMeetingByQualifiedId` can skip the upsert when the mutation version changed, but it still resolves with the fetched series. So even without a DELETE race, we might notify from data the store chose not to apply.

Could the sync Task return whether the write was applied, and only then call the callback?

thisisamir98 (commented):
Nice direction overall. Notifying from the synced meeting should fix stale update titles, and the self-authored update path looks good.

I left a blocking note on the sync success callback when a DELETE arrives while sync is still queued. There are also smaller points on store/notification alignment and MEMBER_ADDED filtering.

One question: WPB-25516 is about cancellation cards, but this change is mostly update notification timing. Is that ticket intentional here, or should the PR description point at a different one?

apps/webapp/src/script/components/Meeting/meetingStore/subscribeToMeetingLifecycleEvents.ts
zskhan: I left it earlier as it is but now I am going to filter it so there will be no useless notification for creator.

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
zskhan: a5dfdb5b918913e1e276f7f47bffad23121c0691

apps/webapp/src/script/components/Meeting/meetingStore/subscribeToMeetingLifecycleEvents.ts
zskhan: c9207393e22a73ece93ab7493f8a8d20fa16cf8f

apps/webapp/src/script/components/Meeting/meetingStore/createMeetingLifecycleDispatcher.ts
zskhan: c9207393e22a73ece93ab7493f8a8d20fa16cf8f

thisisamir98 (approved):

PR #22160, opened 2026-08-11
Add e2e test for meting notification to reduce manual testing effort [WPB-22420]

zskhan:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
@@ -51,14 +51,15 @@ test('single participant meeting notification', async ({createUser, createTeam, 
 
   const participants = modal.locator('[data-uie-name="schedule-meeting-participants"]');
   await participants.locator('[data-uie-name="schedule-meeting-participants-input"]').fill(member.fullName);
-  await ownerPage.locator('[data-uie-name="item-user"]').filter({hasText: member.fullName}).click();
+  await modal.locator('[data-uie-name="item-user"]').filter({hasText: member.fullName}).click();
 
   await modal.locator('[data-uie-name="schedule-meeting-modal-submit"]').click();
+  await expect(modal).toBeHidden();
 
   const host = memberPage.locator('[data-uie-name="meeting-notification-host"]');
   await expect.poll(() => host.count(), {timeout: 30_000}).toBe(1);
   await expect(host).toBeVisible();
-  await host.getByRole('button', {name: 'Expand'}).click();
+  await host.getByTestId('meeting-notification-expand').click();
 
   const cards = host.locator('[data-uie-name^="meeting-notification-card-"]');
   await expect(cards).toHaveCount(1);
--- a/apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
+++ b/apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
@@ -1,0 +1,66 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+
+import type {Page} from '@playwright/test';
+
+import {PageManager} from 'test/e2e_tests/pageManager';
+import {expect, LOGIN_TIMEOUT, test} from 'test/e2e_tests/test.fixtures';
+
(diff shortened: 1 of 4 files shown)


apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  await pageManager.openLoginPage();
+  await pages.login().login(user);
+  await components.conversationSidebar().sidebar.waitFor({state: 'visible', timeout: LOGIN_TIMEOUT});
+  const clientUrl = new URL('/?enabled-features=meetings#/clients', page.url());
+  await page.goto(clientUrl.href, {waitUntil: 'domcontentloaded'});
+  await expect.poll(() => new URL(page.url()).searchParams.get('enabled-features')).toBe('meetings');
+  await components.conversationSidebar().sidebar.waitFor({state: 'visible', timeout: LOGIN_TIMEOUT});
+};
+
+test('single participant meeting notification', async ({createUser, createTeam, createPage}) => {
```
thisisamir98, line 36: Could we add a tag here, for example `@regression`? Our default e2e runs only pick up specs matching `@regression|@crit-flow-web`, so without a tag this will not run in the release or maintenance gates. That would limit how much manual testing this replaces.

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  await ownerPage.locator('[data-uie-name="go-meetings"]').click();
+  await ownerPage.locator('[data-uie-name="schedule-meeting"]').click();
+
+  const modal = ownerPage.locator('[data-uie-name="schedule-meeting-modal"]');
+  await modal.locator('[data-uie-name="schedule-meeting-title"]').fill('Single participant meeting');
+
+  const participants = modal.locator('[data-uie-name="schedule-meeting-participants"]');
+  await participants.locator('[data-uie-name="schedule-meeting-participants-input"]').fill(member.fullName);
+  await ownerPage.locator('[data-uie-name="item-user"]').filter({hasText: member.fullName}).click();
```
thisisamir98: This `item-user` locator is scoped to the whole page. If another user list is visible, we might click the wrong row. Could we scope it to the modal or the participants dropdown, for example `modal.locator('[data-uie-name="item-user"]')`?

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  await modal.locator('[data-uie-name="schedule-meeting-modal-submit"]').click();
```
thisisamir98, line 56: After submit we go straight to waiting on the member notification. If scheduling or participant sync fails on the owner side, we only get a 30 second timeout on the member page, which is hard to read in CI. Could we assert owner success first, for example that the modal closes or the meeting shows up in the list?

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const host = memberPage.locator('[data-uie-name="meeting-notification-host"]');
+  await expect.poll(() => host.count(), {timeout: 30_000}).toBe(1);
+  await expect(host).toBeVisible();
+  await host.getByRole('button', {name: 'Expand'}).click();
```
thisisamir98: We added `data-uie-name="meeting-notification-expand"` on the button, but the test still uses the English label. Playwright is already set up with `testIdAttribute: data-uie-name`, so I would prefer `host.getByTestId('meeting-notification-expand')` here. That also avoids breakage if the suite runs under another locale.

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const cards = host.locator('[data-uie-name^="meeting-notification-card-"]');
+  await expect(cards).toHaveCount(1);
+  await expect(cards).toContainText('Update: Single participant meeting');
```
thisisamir98, line 66: Is `Update:` the label we expect for a newly scheduled meeting? With the current lifecycle wiring, participants get an update notification after member-add, not an invitation. That matches the code today, but I want to confirm this is what QA checks manually. If invitation is the intended UX later, this assertion will need to change.

thisisamir98 (commented):
Thanks for adding coverage here. The fixture and brig helpers look consistent with channels/cells. A few things would make this test more useful in CI and easier to debug when it fails.

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
zskhan: Right now I don't want to make it part of regular test runs. Once the test case is verified by the QA then we can add two tags `@meeting|@regression`.

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
zskhan: 594d1e14224eef44c6b4e28ac30bdab05dc97bea

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
zskhan: This is wrong and `member-add` event should be the correct event.
And once correct event is there then invite notification will be displayed.

thisisamir98 (approved):

PR #22166, opened 2026-08-11
Move meetingIdKey in to the meetings/util [WPB-22420]

zskhan:
<table>
<td>
  </td></table>
  <br />


apps/webapp/src/script/components/appContainer/appContainer.tsx
```diff
@@ -29,7 +29,7 @@ import {FireAndForgetInvoker} from '@wireapp/core';
 import {StyledApp, THEME_ID} from '@wireapp/react-ui-kit';
 import {WebAppEvents} from '@wireapp/webapp-events';
 
-import {MeetingStoreRoot} from 'Components/Meeting/meetingStore/MeetingStoreRoot';
+import {MeetingStoreRoot} from 'Components/meeting/meetingStore/MeetingStoreRoot';
```
thisisamir98: Could we update this import path to match the renamed file?

The file is now `meetingStoreRoot.tsx`, but this still imports `MeetingStoreRoot`. That works on a case-insensitive Mac disk, but Linux CI cannot resolve it (type-check is already failing with `Cannot find module .../MeetingStoreRoot`).

Same pattern shows up in several other places (`MeetingStoreProvider`, `ScheduleFormErrors`, `MeetingList`, `MeetingHeader`, and similar). Could we sweep import/export paths so they match the new camelCase filenames exactly?

apps/webapp/src/script/components/meeting/meetingStore/meetingStoreRoot.tsx
```diff
@@ -24,12 +24,12 @@ import {container} from 'tsyringe';
...
-import {createMeetingNotificationEventHandlers} from 'Components/Meeting/meetingNotificationEventHandlers';
-import {useMeetingNotificationStore} from 'Components/Meeting/meetingNotificationStore/meetingNotificationStore';
-import {createMeetingStore} from 'Components/Meeting/meetingStore/createMeetingStore';
-import {MeetingStoreProvider} from 'Components/Meeting/meetingStore/MeetingStoreProvider';
-import {deleteMeetingForAll, deleteMeetingForMe} from 'Components/Meeting/shared/service/deleteMeeting';
-import {meetNowMeeting, scheduleMeeting, updateMeeting} from 'Components/Meeting/shared/service/meetingService';
+import {createMeetingNotificationEventHandlers} from 'Components/meeting/meetingNotificationEventHandlers';
+import {useMeetingNotificationStore} from 'Components/meeting/meetingNotificationStore/meetingNotificationStore';
+import {createMeetingStore} from 'Components/meeting/meetingStore/createMeetingStore';
+import {MeetingStoreProvider} from 'Components/meeting/meetingStore/MeetingStoreProvider';
```
thisisamir98: Same case mismatch here: the provider file is `meetingStoreProvider.tsx`, but the import still uses `MeetingStoreProvider`.

On Linux CI this fails module resolution. Could we point this (and the other `MeetingStoreProvider` imports) at `meetingStoreProvider`?

apps/webapp/src/script/components/meeting/meetings.tsx
```diff
@@ -21,14 +21,14 @@ import {useCallback, useEffect, useRef} from 'react';
...
-import {meetingsContentWrapperStyles} from 'Components/Meeting/MeetingCallingView/meetingCallingView.styles';
-import {MeetingHeader} from 'Components/Meeting/MeetingHeader/MeetingHeader';
-import {MeetingList} from 'Components/Meeting/MeetingList/MeetingList';
-import {useMeetingStore} from 'Components/Meeting/meetingStore/MeetingStoreProvider';
-import {MeetNowModal} from 'Components/Meeting/meetNowModal/meetNowModal';
-import {ScheduleMeetingModal} from 'Components/Meeting/ScheduleMeetingModal';
+import {contentStyles} from 'Components/meeting/meeting.styles';
+import {MeetingCallingView} from 'Components/meeting/meetingCallingView/meetingCallingView';
+import {meetingsContentWrapperStyles} from 'Components/meeting/meetingCallingView/meetingCallingView.styles';
+import {MeetingHeader} from 'Components/meeting/meetingHeader/MeetingHeader';
```
thisisamir98: These imports still use PascalCase filenames (`MeetingHeader`, `MeetingList`, `MeetingStoreProvider`) after the files were renamed to camelCase.

Could we update them to `meetingHeader`, `meetingList`, and `meetingStoreProvider`? Right now type-check fails on case-sensitive CI for the same reason.

apps/webapp/src/script/components/meeting/mapScheduleFormToMeetingCommand.ts
```diff
@@ -20,15 +20,15 @@
 import type {WallClock} from '@enormora/wall-clock/wall-clock';
 import {result, Result} from 'true-myth';
 
-import {ScheduleFormErrors, scheduleFormErrors} from 'Components/Meeting/ScheduleFormErrors';
-import {requireScheduleMeetingTimes} from 'Components/Meeting/ScheduleMeetingModal/requireScheduleMeetingTimes';
+import {ScheduleFormErrors, scheduleFormErrors} from 'Components/meeting/ScheduleFormErrors';
```
thisisamir98: Could we change this to `Components/meeting/scheduleFormErrors`?

The file was renamed to `scheduleFormErrors.ts`, so `ScheduleFormErrors` only resolves on case-insensitive filesystems. CI already reports this as a missing module.

.ls-lint.yml
```diff
@@ -16,6 +16,12 @@ ls:
     .tsx: camelCase
     .test.tsx: camelCase
     .types.ts: camelCase
+  apps/webapp/src/script/components/Meeting:
```
thisisamir98: Could we point this rule at `apps/webapp/src/script/components/meeting` (lowercase)?

The folder was renamed, but the config still uses `Meeting`. That can leave the new tree outside the camelCase rule, so leftover PascalCase files (for example `meetingList/MeetingList.styles.ts`) may not get caught.

thisisamir98 (changes requested):
## Summary
Nice cleanup overall. Extracting `toMeetingIdKey` looks solid. The camelCase rename still leaves many import paths on the old PascalCase filenames, so type-check / lint / tests fail on Linux CI even when it looks fine locally on macOS.

## Findings
### Blocking
- Import paths still use PascalCase filenames after the rename (`MeetingStoreRoot`, `MeetingStoreProvider`, `ScheduleFormErrors`, `MeetingList`, and others). Please align every import/export with the new camelCase file names.

### Should fix
- `.ls-lint.yml` still points at `components/Meeting` while the folder is now `meeting`, so the naming rule may not cover the new path (and leftover PascalCase files can slip through).

## What looks good
- Shared `toMeetingIdKey` matches the previous `domain:id` format and is wired through store versions, lifecycle dispatch, notification pending keys, and delete in-flight tracking.

thisisamir98 (dismissed):

PR #22170, opened 2026-08-11
Show invite notification instead of update to the participant when meeting is created [WPB-26695]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26695" title="WPB-26695" target="_blank">TaskWPB-26695</a>  [Web] Inform participants of a new meeting invite
  </td></table>
  <br />
 

Also add e2e tests to cover this scenario


arjita-mitra (dismissed):

thisisamir98 (commented):
Nice fix for the first-invite copy. I left a few notes on classification durability, test coverage, and one e2e assertion gap. Nothing blocking from my side, but worth a look before merge.

apps/webapp/src/script/components/meeting/meetingNotificationEventHandlers.ts
```diff
@@ -110,6 +112,12 @@ export const createMeetingNotificationEventHandlers = ({
   };
 
   return {
+    notifyMeetingChange: meeting => {
```
thisisamir98, line 115: Could we add a unit test for this path? Something like calling `notifyMeetingChange` twice for the same meeting id and checking the first notification is INVITE and the second is UPDATE. The existing handler tests still only cover `notifyUpdate`.

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -33,6 +33,50 @@ const loginWithMeetingsEnabled = (user: {email: string; password: string}) => as
...
+  await host.getByTestId('meeting-notification-expand').click();
+
+  const cards = meetingNotificationCards(page);
+  await expect(cards).toHaveCount(1);
+  await expect(cards).toContainText(`Invitation: ${title}`);
+  await expect(cards).not.toContainText(`Update: ${title}`);
+};
+
+const editMeetingToAddParticipant = async (ownerPage: Page, title: string, participantName: string) => {
+  const meeting = ownerPage.locator('[data-uie-name="meetings-list"]').getByText(title).locator('xpath=../../..');
```
thisisamir98: Nit: the `xpath=../../..` hop from the title text feels a bit fragile if the list row markup changes. If we have or can add a `data-uie-name` on the meeting row, that would make this easier to maintain.

thisisamir98 (approved):

PR #22172, opened 2026-08-11
Meeting UI fixes [WPB-27862]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27862" title="WPB-27862" target="_blank">BugWPB-27862</a>  UI Changes in Wire Meetings
  </td></table>
  <br />


apps/webapp/src/script/components/meeting/scheduleMeetingModal/scheduleMeetingForm.tsx
```diff
@@ -183,8 +184,11 @@ export const ScheduleMeetingForm = ({
   return (
     <div css={scheduleMeetingFormLayoutCss} data-uie-name="schedule-meeting-form" data-uie-mode={mode}>
       <div css={scheduleMeetingFormLeftColumnCss}>
+        {/* eslint jsx-a11y/no-autofocus : "off" */}
```
arjita-mitra: This is not the best practice to implement autofocus. Use explicit focus management via a ref + useEffect that runs when modal open state changes. This gives better control and avoids disabling an eslint accessibility rule.

Benefits
- Timing control: focus only after the modal is mounted and visible, avoiding situations where the element isn't in the DOM yet or gets blurred by other code.
- Avoids focus stealing: autoFocus can move focus unexpectedly common issue. Programmatic focus can be gated by explicit open logic so it runs only when intended.
- Better screen reader experience: focusing after the modal and role are fully set reduces the chance of confusing SR announcements and makes it easier to ensure the right element receives focus.
- Easier to integrate with focus traps and restoration.
- programmatic focus is easier to assert in unit/integration tests because it runs in a predictable effect tied to the modal open state.
- Finally no eslint-disable: keeps the jsx-a11y/no-autofocus rule in place which exists to prevent accessibility mistakes

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingAction/meetingAction.styles.ts
```diff
@@ -38,3 +38,7 @@ export const contextMenuDangerItemStyles: CSSObject = {
 export const contextMenuDangerItemIconStyles: CSSObject = {
   fill: 'var(--danger-color)',
 };
+
+export const contextMenuDisabledItemIconStyles: CSSObject = {
+  fill: 'var(--foreground-fade-48)',
```
arjita-mitra, line 43: Please confirm the color contrast of the faded color meets WCAG for users with low vision.

arjita-mitra: Also ensure disabled control exposes set for aria-disabled for assistive tech.

apps/webapp/src/script/components/meeting/meetNowModal/meetNowModal.test.tsx
```diff
@@ -222,6 +222,16 @@ describe('MeetNowModal', () => {
     jest.restoreAllMocks();
   });
```
arjita-mitra, line 224: Missing tests

- Add test for the autoFocus behavior. Missing assertion that focus lands correctly or that focus management is safe.

- No accessibility checks (axe/jest-axe) for contrast, role, aria attributes, keyboard navigation, or screen-reader behavior.

a sample snippet

```
import {axe} from 'jest-axe';

it('has no basic a11y violations when Meet Now modal is open', async () => {
  const {container} = renderMeetNowModal(); // adapt helper to return container if needed

  act(() => useMeetNowModal.getState().open());

  // wait for modal content to be visible if necessary
  await waitFor(() => expect(screen.getByRole('dialog')).toBeInTheDocument());

  const results = await axe(container);
  expect(results).toHaveNoViolations();
});
```

arjita-mitra (changes requested):

zskhan:
- Replaced modal `autoFocus` usage with explicit ref/effect focus management and added focus regression coverage.
- Added `aria-disabled` and prevented keyboard activation for disabled context-menu entries.
- I am intentionally not adding `jest-axe` at this time. We can discuss it and introduce it later

apps/webapp/src/style/common/context.less
```diff
@@ -57,7 +57,7 @@
     }
 
     &--disabled {
-      color: var(--foreground-fade-48);
+      color: var(--main-color);
```
thisisamir98: Could we keep a distinct disabled look here?

Switching disabled items to `var(--main-color)` makes them very close to enabled items (italic is basically the only cue). This also applies to every context menu in the app, not only meetings.

I would prefer a muted token that still has good contrast, and let the icon follow the item text (for example via `currentColor`) instead of painting disabled as main color.

Test idea: open a menu with disabled Join next to enabled Edit and check that disabled stays visually distinct and still cannot be activated.

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingAction/meetingAction.styles.ts
```diff
@@ -38,3 +38,7 @@ export const contextMenuDangerItemStyles: CSSObject = {
 export const contextMenuDangerItemIconStyles: CSSObject = {
   fill: 'var(--danger-color)',
 };
+
+export const contextMenuDisabledItemIconStyles: CSSObject = {
+  fill: 'var(--main-color)',
```
thisisamir98: Same point as on `context.less`: `fill: var(--main-color)` on the disabled Join icon removes the muted disabled cue.

If the item text stays on a disabled color, can the icon inherit that (for example `currentColor`) so text and icon stay in sync without looking enabled?

apps/webapp/src/script/components/meeting/meetNowModal/meetNowModal.test.tsx
```diff
@@ -222,6 +222,26 @@ describe('MeetNowModal', () => {
...
+  });
+
+  it('focuses the title when the modal opens', () => {
+    renderMeetNowModal();
+
+    act(() => {
+      useMeetNowModal.getState().open();
+    });
+
+    expect(screen.getByLabelText('meetings.scheduleModal.titleLabel')).toHaveFocus();
```
thisisamir98, line 242: Nice focus coverage. For ticket issue 3, can we also assert that Escape closes the modal after it opens?

Focus on the title is what makes Esc reach the modal `onKeyDown`, but the product check is still dismiss on Escape (and that it stays open while submit is pending).

A similar check on the schedule meeting modal would be great too.

apps/webapp/src/i18n/en-US.json
```diff
@@ -1359,9 +1359,9 @@
   "meetings.scheduleModal.participantsSelectedSummaryTwo": "{name1}, {name2}",
   "meetings.scheduleModal.previousMonthAriaLabel": "Previous month",
   "meetings.scheduleModal.recurrence.daily": "Daily",
-  "meetings.scheduleModal.recurrence.doesNotRepeat": "Does not repeat",
   "meetings.scheduleModal.recurrence.everyFourWeeks": "Every four weeks",
   "meetings.scheduleModal.recurrence.everyTwoWeeks": "Every two weeks",
+  "meetings.scheduleModal.recurrence.never": "Never",
```
thisisamir98, line 1364: Nit / question: was renaming the key to `recurrence.never` intentional?

Other locale files still have `recurrence.doesNotRepeat`, so they will fall back to English until Crowdin syncs. If we only need the English copy change, updating the value on the existing key would keep the current translations.

apps/webapp/src/script/components/meeting/shared/styles/meetingForm.styles.ts
```diff
@@ -38,6 +38,10 @@ export const scheduleMeetingRecurrenceSelectWrapperStyles = scheduleMeetingLabel
 
 export const scheduleMeetingTitleInputWrapperStyles = scheduleMeetingLabeledFieldWrapperStyles;
 
+export const scheduleMeetingTitleInputStyles: CSSObject = {
+  paddingRight: '38px',
```
thisisamir98, line 42: Nit: `paddingRight: 38px` is always applied, even when the clear button is hidden. Not a blocker. If the group name field only adds that space when the clear control is present, matching that would be a bit cleaner.

thisisamir98 (changes requested):
Thanks for the follow-up on focus management and the disabled menu a11y work. The Meet Now icon `currentColor` fix and the title focus approach look good.

I would like a small change on the disabled context-menu colors (they currently read almost like enabled), plus an Escape-to-dismiss regression for the ticket. Details inline.

apps/webapp/src/script/components/meeting/shared/styles/meetingForm.styles.ts
zskhan: Padding will not be visible until and unless there is a text in the input field

apps/webapp/src/i18n/en-US.json
zskhan: "Does not repeat" is not relevant anymore so completely renaming it

apps/webapp/src/style/common/context.less
zskhan: image
Fixed

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingAction/meetingAction.styles.ts
zskhan: 286e396fad33c0ab24c3353af85dfb78

thisisamir98 (approved):

PR #22173, opened 2026-08-11
fix(drive): preview guest PDFs as images [WPB-27965]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27965" title="WPB-27965" target="_blank">TaskWPB-27965</a>  PDF lightbox components ignore Previews[].PreSignedGET and fall back to node.PreSignedGET — preview breaks for viewers
  </td></table>
  <br />
 

### Problem

Cells viewers cannot access an original PDF URL. Cells supplies a signed JPG preview, but Webapp treated the item as a PDF and showed no preview.

### Solution

- Use the signed image preview when a viewer has no original PDF URL.
- Render that preview through the image viewer.
- Gate the behavior with the existing `viewer-permission` startup feature toggle.

ref: https://wearezeta.atlassian.net/browse/WPB-27965

### DIFF
--- a/apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
+++ b/apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
@@ -149,6 +149,45 @@ describe('FileFullscreenModal - File Version Restore', () => {
       expect(screen.getByTestId('image-view')).toBeInTheDocument();
     });
 
+    it('should render PDF viewer when its preview URL has no file extension', () => {
+      render(
+        <FileFullscreenModal
+          {...defaultProps}
+          filePreviewUrl="https://storage.example.com/download?node=123"
+          fileExtension="pdf"
+        />,
+      );
+
+      expect(screen.getByTestId('pdf-viewer')).toBeInTheDocument();
+    });
+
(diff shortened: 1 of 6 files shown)


arjita-mitra (approved):

apps/webapp/src/script/components/FileFullscreenModal/FileFullscreenModal.test.tsx
```diff
@@ -149,6 +149,45 @@ describe('FileFullscreenModal - File Version Restore', () => {
...
+          {...defaultProps}
+          fileExtension="pdf"
+          filePreviewUrl="https://storage.example.com/preview?node=123"
+          isImagePreview
+        />,
+      );
+
+      expect(screen.getByTestId('image-view')).toBeInTheDocument();
+    });
+
```
arjita-mitra, line 190: Good to add 

- should not render preview for audio files even with previewUrl available
- should not render preview for video files even with previewUrl available
- should render image viewer for image files even when previewPdfUrl is present'

cristianoliveira:
After a small refinement we decided to keep the preview of PDFs, with the permission of possible download via presigned URLs, in favor of a better UX

PR #22186, opened 2026-08-13
Update the color of "Open in a new window" icon in dark theme [WPB-27893]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27893" title="WPB-27893" target="_blank">BugWPB-27893</a>  User cannot see the "Open in a new window" icon in a call in Dark theme because it is displayed in black
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/calling/CallingCell/CallingHeader/CallingHeader.styles.ts
+++ b/apps/webapp/src/script/components/calling/CallingCell/CallingHeader/CallingHeader.styles.ts
@@ -79,10 +79,14 @@ export const detachedWindowButton: CSSObject = {
   padding: '8px 12px',
 
   '& svg': {
-    fill: 'var(--text-color)',
+    fill: 'currentColor',
   },
 
   '& svg path': {
-    fill: 'var(--text-color)',
+    fill: 'currentColor',
+  },
+
+  '&:hover, &:focus-visible, &:active': {
+    color: 'var(--accent-color)',
   },
 };


apps/webapp/src/script/components/calling/CallingCell/CallingHeader/CallingHeader.styles.ts
```diff
@@ -79,10 +79,14 @@ export const detachedWindowButton: CSSObject = {
...
+    fill: 'currentColor',
   },
 
   '& svg path': {
-    fill: 'var(--text-color)',
+    fill: 'currentColor',
+  },
+
+  '&:hover, &:focus, &:active': {
+    color: 'var(--accent-color)',
```
arjita-mitra, line 90: why not focus-visible? focus-visible shows focus only when it matters (keyboard/tab navigation) so you don’t show focus rings on mouse clicks.

Use :focus as a fallback only if you need to support old browsers that don’t implement :focus-visible.

arjita-mitra: Do we already have an outline for the callingheader?

Color change on :hover / :active is for pointer users only. Keyboard users need a visible focus indication to know where keyboard focus is that’s an accessibility requirement.

zskhan: I’ve switched this to :focus-visible and verified that keyboard focus has a visible outline.

arjita-mitra (approved):

PR #22207, opened 2026-08-14
Fix meeting update notification classification [WPB-25510]

zskhan:
<table>
<td>
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.test.ts
+++ b/apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.test.ts
@@ -43,15 +43,17 @@ describe('subscribeToMeetingLifecycleEvents', () => {
   const activeUnsubscribeCallbacks: (() => void)[] = [];
 
   const subscribe = (dispatcher: MeetingLifecycleDispatcherDouble) => {
+    const notifyMeetingChange = jest.fn();
+    const notifyUpdate = jest.fn();
     const unsubscribe = subscribeToMeetingLifecycleEvents({
       dispatcher,
       getSelfUserQualifiedId: () => selfUserId,
-      notifyMeetingChange: jest.fn(),
-      notifyUpdate: jest.fn(),
+      notifyMeetingChange,
+      notifyUpdate,
     });
     activeUnsubscribeCallbacks.push(unsubscribe);
 
-    return unsubscribe;
+    return {notifyMeetingChange, notifyUpdate, unsubscribe};
   };
 
   afterEach(() => {
@@ -70,21 +72,21 @@ describe('subscribeToMeetingLifecycleEvents', () => {
 
   it('queues a meeting sync when a meeting updated event is published', () => {
     const dispatcher = createDispatcherDouble();
-    subscribe(dispatcher);
+    const {notifyUpdate} = subscribe(dispatcher);
 
     amplify.publish(WebAppEvents.MEETING.UPDATED, meetingId, otherUserId);
 
-    expect(dispatcher.enqueueMeetingSync).toHaveBeenCalledWith(meetingId, expect.any(Function));
+    expect(dispatcher.enqueueMeetingSync).toHaveBeenCalledWith(meetingId, notifyUpdate);
     expect(dispatcher.enqueueMeetingRemoval).not.toHaveBeenCalled();
   });
 
   it('queues a meeting sync when a meeting member-added event is published', () => {
     const dispatcher = createDispatcherDouble();
-    subscribe(dispatcher);
+    const {notifyMeetingChange} = subscribe(dispatcher);
 
     amplify.publish(WebAppEvents.MEETING.MEMBER_ADDED, meetingId, otherUserId);
 
-    expect(dispatcher.enqueueMeetingSync).toHaveBeenCalledWith(meetingId, expect.any(Function));
+    expect(dispatcher.enqueueMeetingSync).toHaveBeenCalledWith(meetingId, notifyMeetingChange);
     expect(dispatcher.enqueueMeetingRemoval).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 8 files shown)


apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.test.ts
```diff
@@ -47,6 +47,7 @@ describe('subscribeToMeetingLifecycleEvents', () => {
       dispatcher,
       getSelfUserQualifiedId: () => selfUserId,
       notifyMeetingChange: jest.fn(),
+      notifyUpdate: jest.fn(),
```
thisisamir98: Could we assert which callback is passed, rather than `expect.any(Function)`?

`notifyUpdate` is now required, but neither spy is returned or checked. The update and member-added tests below would still pass if those events were wired to the opposite handlers, or if update went back to `notifyMeetingChange`. That is the behaviour this PR changes.

Could `subscribe` return both spies, then assert `toHaveBeenCalledWith(meetingId, notifyUpdate)` for `UPDATED` and `toHaveBeenCalledWith(meetingId, notifyMeetingChange)` for `MEMBER_ADDED`?

That would lock the routing so it cannot regress silently.

apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.ts
```diff
@@ -41,13 +42,14 @@ export const subscribeToMeetingLifecycleEvents = ({
...
 }: SubscribeToMeetingLifecycleEventsDependencies): (() => void) => {
   const onMeetingCreated = (meetingId: QualifiedId) => {
     dispatcher.enqueueMeetingSync(meetingId);
   };
 
-  const enqueueMeetingSyncWithNotification = (meetingId: QualifiedId) => {
-    dispatcher.enqueueMeetingSync(meetingId, notifyMeetingChange);
+  const enqueueMeetingSyncWithNotification = (meetingId: QualifiedId, onSuccess: (meeting: MeetingSeries) => void) => {
+    dispatcher.enqueueMeetingSync(meetingId, onSuccess);
   };
```
thisisamir98: Nit: this helper is now a one-line alias of `dispatcher.enqueueMeetingSync`. Could we call `enqueueMeetingSync` directly from `onMeetingUpdated` and `onMeetingMemberAdded` so the two callbacks are easier to see?

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
```diff
@@ -119,9 +119,40 @@ test.describe('Meetings CRUD', () => {
     await ownerMeetings.scheduleMeeting(MEETING_TITLE, [member.fullName]);
     await ownerMeetings.waitForMeetingInList(MEETING_TITLE);
 
+    await memberMeetings.openMeetingsTab();
+    await memberMeetings.waitForMeetingInList(MEETING_TITLE);
+    await memberMeetings.dismissNotificationContaining(`Invitation: ${MEETING_TITLE}`);
```
thisisamir98: Could we avoid dismissing the invitation before the host edits?

Dismissing first means this meeting is already in `notifiedMeetings`. On the old `notifyMeetingChange` path, the next event would also have been an Update card. The QA failure was the first update in a session where that set is empty (reload, or the meeting was already in the list from load). The `meetingStoreRoot` unit test covers that. This e2e (and the time-edit test below) does not.

Could the invitee reload, or skip the live invite, before the edit, then assert only an Update card appears?

thisisamir98 (commented):
Thanks for the focused fix. Routing `meeting.update` to `notifyUpdate` is the right way to classify existing invitees, rather than depending on in-memory "already notified this session" state.

I left a few notes on tests and a leftover wrapper. Nothing blocking from my side.

Two product questions, in case they are still open:

1. On WPB-25510, Ruchika noted that a member added again should get Invitation, not Update. `notifyMeetingChange` still uses `notifiedMeetings`, and cancel does not clear that set. In the same session, invite then remove then add again still becomes Update. Is that in scope here, or a follow-up?

2. On WPB-25511, Federico suggested that adding a participant should not notify existing members. The backend still emits `meeting.update` to current members on the metadata PUT, then `meeting.member-add` only to the new users. This PR does not change that. Is that accepted for M1?

apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.ts
zskhan: 5483cac74ae243b7448d04e30f2ff95880e40add

apps/webapp/src/script/components/meeting/meetingStore/subscribeToMeetingLifecycleEvents.test.ts
zskhan: 5483cac74ae243b7448d04e30f2ff95880e40add

apps/webapp/test/e2e_tests/specs/Meetings/meetings.spec.ts
zskhan: 5483cac74ae243b7448d04e30f2ff95880e40add

zskhan:
> Thanks for the focused fix. Routing `meeting.update` to `notifyUpdate` is the right way to classify existing invitees, rather than depending on in-memory "already notified this session" state.
> 
> I left a few notes on tests and a leftover wrapper. Nothing blocking from my side.
> 
> Two product questions, in case they are still open:
> 
> 1. On WPB-25510, Ruchika noted that a member added again should get Invitation, not Update. `notifyMeetingChange` still uses `notifiedMeetings`, and cancel does not clear that set. In the same session, invite then remove then add again still becomes Update. Is that in scope here, or a follow-up?

I fixed this issue.

> 2. On WPB-25511, Federico suggested that adding a participant should not notify existing members. The backend still emits `meeting.update` to current members on the metadata PUT, then `meeting.member-add` only to the new users. This PR does not change that. Is that accepted for M1?


This is on backend side so it is out of scope for now but we can think about it.

thisisamir98 (dismissed):

PR #22208, opened 2026-08-14
refactor: centralize Cells viewer action permissions(WPB-27987)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27987" title="WPB-27987" target="_blank">TaskWPB-27987</a>  Hide all modification action like edit, version history etc at conversation level and full screen modal including collabora
  </td></table>
  <br />
 

Introduce a Cells action permission helper so viewer restrictions are checked by action instead of passing specific action flags throughtout the component tree.

This keeps the current restricted behavior for viewer actions while making future action pecific restrictions, such as restore, easier to add in smaller follow-up changes.

Covered use cases:

1. Guest/viewer only sees Open at conversation row-menu level.
2. Viewer cannot see Download, Edit, Version history, create/new, rename, move, tags, share, delete/restore actions.
3. Fullscreen preview header download button is hidden for viewer.
4. No-preview download CTA is hidden for viewer.
5. Member/editor can still download.
6. Previews still receive the file/preview URLs, so PDF/image preview UX should stay unchanged.
7. Collabora edit entry is hidden because EDIT is restricted and fullscreen edit mode is gated.


apps/webapp/src/script/components/FileFullscreenModal/FileHeader/FileHeader.tsx
```diff
@@ -140,7 +144,7 @@ export const FileHeader = ({
             <ShowIcon width={16} height={16} />
             Viewing
           </button>
-          {!isRecycleBin && (
+          {!isRecycleBin && canPerformCellsAction(CELLS_ACTION.EDIT) && (
```
cristianoliveira, line 147: this condition and the version history one a few lines below have no test coverage, we only test the download button

It would be nice to have the negative case here: viewer with the flag enabled should not see edit or version history even for an editable file

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSelfUserDriveRole/CellsSelfUserDriveRoleContext.tsx
```diff
@@ -58,6 +61,52 @@ export const shouldRestrictCellsViewerActions = ({
...
+export const canPerformCellsAction = ({
+  action,
+  isViewerPermissionFeatureEnabled,
+  selfUserDriveRole,
+}: CanPerformCellsActionParams): boolean => {
+  if (!shouldRestrictCellsViewerActions({isViewerPermissionFeatureEnabled, selfUserDriveRole})) {
+    return true;
+  }
+
+  return !RESTRICTED_VIEWER_ACTIONS.includes(action);
```
cristianoliveira, line 108: small thing, this is allow-by-default for viewers: any new action added later is allowed until someone remembers to put it in the list

fine given the plan to add restrictions incrementally, just worth a comment here so it stays a conscious choice

cristianoliveira (dismissed):

cristianoliveira:
Left a couple of comments but no blockers

apps/webapp/src/script/components/Conversation/ConversationCells/common/CellsSelfUserDriveRole/CellsSelfUserDriveRoleContext.tsx
arjita-mitra: comment added 18b007f50708b64c056b751033b2925f57b0184f

apps/webapp/src/script/components/FileFullscreenModal/FileHeader/FileHeader.tsx
arjita-mitra: test added 62e80937b083aff2c384258132462cc53877533e

cristianoliveira (approved):

PR #22209, opened 2026-08-14
fix: Disabled row actions in global drive shouldn't be clickable(WPB-25259)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-25259" title="WPB-25259" target="_blank">TaskWPB-25259</a>  [Web] Hide / Disable download, copy, move to bin, permanent deletion, modification and sharing via public links for shared drive and global files table view each row
  </td></table>
  <br />
 

Issue found:

In Global Drive, Share via link and Download appear visually disabled for files where the user has Viewer access, but both actions remain clickable and complete successfully.

Expected:
Action should appear visually disabled and not clickable

### DIFF
--- a/apps/webapp/src/script/components/cellsGlobalView/cellsTable/cellsTableColumns/cellsTableRowOptions/cellsTableRowOptions.tsx
+++ b/apps/webapp/src/script/components/cellsGlobalView/cellsTable/cellsTableColumns/cellsTableRowOptions/cellsTableRowOptions.tsx
@@ -52,6 +52,10 @@ export const CellsTableRowOptions = (properties: CellsTableRowOptionsProps): Rea
     isViewerPermissionFeatureEnabled: isFeatureToggleEnabled(viewerPermissionFeatureToggleName),
     selfUserDriveRole: node.selfUserDriveRole,
   });
+  // DropdownMenu.Item disabled state is visual/ARIA only for native onClick handlers.
+  // Keep restricted actions without a handler so viewer access cannot activate them.
+  const restrictedActionClickHandler = <ClickHandler extends () => void>(clickHandler: ClickHandler) =>
+    shouldDisableRestrictedActions ? undefined : clickHandler;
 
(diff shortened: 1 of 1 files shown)


apps/webapp/src/script/components/cellsGlobalView/cellsTable/cellsTableColumns/cellsTableRowOptions/cellsTableRowOptions.tsx
```diff
@@ -69,20 +69,25 @@ export const CellsTableRowOptions = (properties: CellsTableRowOptionsProps): Rea
         </DropdownMenu.Item>
         <DropdownMenu.Item
           disabled={shouldDisableRestrictedActions}
-          onClick={() =>
-            showShareModal({type: node.type, uuid: node.id, cellsRepository, fireAndForgetInvoker, translate})
+          onClick={
```
arjita-mitra: Issue: DropdownMenu.Item is a Radix menu item rendered as a non-native element, disabled sets Radix disabled state, but our wrapper was still passing the handler as native onClick. Native clicks can still invoke that callback.

i tried to prevent disabled actions getting fired from DropdownMenu component thats possible using onSelect insiread of onclick but it seems more risky can cause regression in other consumer components therefore i gated the click handler call side.

cristianoliveira: Can we add this explanation as a comment?

arjita-mitra: sure 59f34b06cd7abd5fc31edbc7e78b011613c61d60

cristianoliveira (approved):

PR #22213, opened 2026-08-14
Avoid update metadata call when there is nothing changed in it [WPB-28036]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28036" title="WPB-28036" target="_blank">BugWPB-28036</a>  Previously invited participants receive another invitation notification when new participants are added to a meeting
  </td></table>
  <br />


apps/webapp/src/script/components/meeting/shared/service/meetingService.ts
```diff
@@ -166,6 +172,10 @@ export const updateMeeting = (
 ): Task<MeetingSubmitSuccess, MeetingSubmitErrors> => {
   const {usersToAdd, userIdsToRemove} = computeParticipantDiff(command.originalSelectedUsers, command.selectedUsers);
 
+  if (!hasMeetingMetadataChanged(command)) {
+    return syncParticipantsAfterMeetingUpdate(command, deps, usersToAdd, userIdsToRemove);
```
thisisamir98: Could we still heal a drifted conversation name when only participants change?

If metadata is unchanged, this returns after participant sync and never calls `syncMeetingConversationName`. A stale live name is still healed when start or end changes (see `heals a drifted conversation name on a non-title meeting update`), but adding or removing people would leave it stale.

I would prefer running the same heal on this path with `command.title` and `command.qualifiedConversation` when the conversation id is present.

Could we add a regression: conversation is named "Stale call name", form title is unchanged, add one participant. Expect no metadata `updateMeeting`, and `renameConversation` still called with the current title.

Is skipping the heal here intentional, or only meant for the fully unchanged no-op case?

apps/webapp/src/script/components/meeting/scheduleMeetingModal/useScheduleMeetingSubmit.ts
```diff
@@ -86,10 +92,17 @@ const submitMeeting = ({
     return task.reject(meetingSubmitErrors.editMeetingIdMissing);
   }
 
+  if (originalStart.isNothing || originalEnd.isNothing) {
+    return task.reject(meetingSubmitErrors.updateFailed);
```
thisisamir98: This fails closed when original start or end is missing, which is good. It avoids sending an update with invented times.

`updateFailed` reads like the backend write failed, though. Could we keep fail-closed and use a dedicated impossible-state error, similar to `editMeetingIdMissing`?

Please add a regression: edit mode with original times missing and valid form times. Expect `submitFailed` and no `updateMeeting` call.

apps/webapp/src/script/components/meeting/shared/service/meetingService.test.ts
```diff
@@ -427,12 +431,66 @@ describe('updateMeeting', () => {
...
+  ])('detects %s metadata changes independently', async (_field, change) => {
+    const {deps, updateMeetingMock} = createDeps();
+
+    const result = await updateMeeting(updateCommand(change), deps);
+
+    expect(result.isOk).toBe(true);
+    expect(updateMeetingMock).toHaveBeenCalled();
+  });
+
+  it('compares recurring edit times with the original upcoming instance times', async () => {
```
thisisamir98, line 490: Non-blocking: this assigns the same `Date` to `start` and `originalStart`, so it cannot fail if originals were snapshotted from series start instead of the upcoming instance. `openEdit` copying `formState.start` / `formState.end` is what makes that safe in production.

Could we also cover that from `openEdit`, so the snapshot itself is locked?

thisisamir98 (commented):
Skipping the metadata PUT when title, times, and recurrence are unchanged looks right to me. Backend `updateMeeting` notifies every conversation member, while `meeting.member-add` notifies only the users who were added.

A few follow-ups are on the skip path and the fail-closed original times check.

Two notes that do not sit on a changed line:
- After a skipped metadata write, `addParticipantsFailed` is still treated as persisted (`setupFailed`, modal closes, list refreshes). If add fails and nothing was written, should the form stay open?
- `mapUpdateCommandToUpdateMeeting.test.ts` `baseUpdateCommand` does not set the new `originalTitle` / `originalStart` / `originalEnd` fields. CI type-check excludes tests, so this can drift.

apps/webapp/src/script/components/meeting/shared/service/meetingService.ts
zskhan: Done. Also added a regression covering a stale conversation name with a participant-only change.

apps/webapp/src/script/components/meeting/scheduleMeetingModal/useScheduleMeetingSubmit.ts
zskhan: added originalMeetingTimesMissing

zskhan:
> * After a skipped metadata write, `addParticipantsFailed` is still treated as persisted (`setupFailed`, modal closes, list refreshes). If add fails and nothing was written, should the form stay open?

I can confirm from product what we should do in this case but not blocking this PR.

apps/webapp/src/script/components/meeting/shared/service/meetingService.test.ts
zskhan: f3a6685495cd8cbfa96f295a516dfad3243bca2e

thisisamir98 (approved):

PR #22227, opened 2026-08-17
Show callingCell of normal conversation in the meetings listing view [WPB-27940]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27940" title="WPB-27940" target="_blank">BugWPB-27940</a>  Incoming call UI is not displayed when the user is on the Meetings page
  </td></table>
  <br />
 

Also clicking on callingHeader will navigate to correct conversation UI.

### DIFF
--- a/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.test.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.test.tsx
@@ -1,0 +1,101 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ */
+
+import {render} from '@testing-library/react';
+import {container} from 'tsyringe';
+
+import {CALL_TYPE, STATE as CALL_STATE} from '@wireapp/avs';
+import {CONVERSATION_PROTOCOL} from '@wireapp/api-client/lib/team';
+
+import {Call} from 'Repositories/calling/Call';
+import {CallState} from 'Repositories/calling/CallState';
+import {Participant} from 'Repositories/calling/Participant';
+import {Conversation} from 'Repositories/entity/Conversation';
+import {User} from 'Repositories/entity/User';
+import {
+  createRootContextValueForTest,
+  createRootProviderWrapperForTest,
+} from 'src/script/page/testSupport/rootContextTestSupport';
+import type {MainViewModel} from 'src/script/view_model/MainViewModel';
+import {buildMediaDevicesHandler} from 'src/script/auth/util/test/testUtil';
+import {translateForTest} from 'Util/test/translateForTest';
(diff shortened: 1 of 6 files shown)


apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const callState = container.resolve(CallState);
+  const mainViewModel = createMainViewModel();
+  const rootProviderWrapper = createRootProviderWrapperForTest(
+    createRootContextValueForTest({
+      mainViewModel,
+      translate: translateForTest,
+    }),
+  );
+
+  it('renders incoming 1:1, group, and meeting calls', () => {
```
thisisamir98, line 85: Could we build one 1:1, one group, and one meeting conversation here?

Right now `createCall()` always creates a default group conversation, so all three cells are groups. WPB-27940 is about 1:1 and group calls on the Meetings page. If we filtered out 1:1 calls again, this test would still pass.

Maybe set `type` / `groupConversationType` explicitly for each call and assert all three `item-call` cells render with accept and decline controls.

thisisamir98 (commented):
Thanks for fixing this. The production changes look right to me: show all active calls on the Meetings page, and reset the sidebar tab when opening a conversation from a non-list tab.

I left a few notes on the tests. Nothing blocking from my side.

apps/webapp/src/script/view_model/ListViewModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  beforeEach(() => {
+    useSidebarStore.setState({currentTab: SidebarTabs.RECENT});
+    useAppState.setState({listState: ListState.CONVERSATIONS});
+  });
+
+  it.each([SidebarTabs.MEETINGS, SidebarTabs.CELLS])(
+    'returns to the conversation list when opening a conversation from tab %s',
+    currentTab => {
+      useSidebarStore.setState({currentTab});
+      useAppState.setState({listState: ListState.MEETINGS});
```
thisisamir98: For the Cells case, could we start from `ListState.CELLS` instead of `ListState.MEETINGS`? In the app, Cells uses its own list state. The assertion would still work, but it would match the real path better.

apps/webapp/src/script/view_model/ListViewModel.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  return listViewModel;
+};
+
+describe('ListViewModel', () => {
+  beforeEach(() => {
+    useSidebarStore.setState({currentTab: SidebarTabs.RECENT});
+    useAppState.setState({listState: ListState.CONVERSATIONS});
+  });
+
+  it.each([SidebarTabs.MEETINGS, SidebarTabs.CELLS])(
```
thisisamir98, line 55: Could we also cover the opposite path: when the current tab is already a conversation list tab (for example Favorites or Groups), `openConversations()` should keep that tab instead of forcing Recent?

Right now only Meetings and Cells are checked, so a regression that always sets Recent would still pass.

apps/webapp/src/script/view_model/ListViewModel.ts
```diff
@@ -320,7 +324,12 @@ export class ListViewModel {
       newState = archive ? ListState.ARCHIVE : ListState.CONVERSATIONS;
     }
     this.switchList(newState, false);
-    setCurrentTab(archive ? SidebarTabs.ARCHIVES : currentTab);
+
+    if (archive) {
+      setCurrentTab(SidebarTabs.ARCHIVES);
+    } else {
+      setCurrentTab(isConversationListTab(currentTab) ? currentTab : SidebarTabs.RECENT);
```
thisisamir98: Non-blocking: `ContentViewModel.showConversation` already resets Preferences and Connect to Recent in its `finally` block. This change now handles Meetings, Cells, and other non-list tabs here in `openConversations`.

Might be worth aligning those two places later so the rule lives in one spot. Not needed for this PR.

thisisamir98 (approved):

PR #22234, opened 2026-08-17
Feat/show loading spinner on meeting buttons [WPB-27437]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27437" title="WPB-27437" target="_blank">StoryWPB-27437</a>  [Web] Handle network/connection errors in Meetings with retry
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/meeting/scheduleMeetingModal/scheduleMeetingModal.tsx
+++ b/apps/webapp/src/script/components/meeting/scheduleMeetingModal/scheduleMeetingModal.tsx
@@ -153,6 +153,7 @@ export const ScheduleMeetingModal = () => {
             css={submitButtonStyles}
             onClick={handleSubmit}
             disabled={isSubmitting}
+            showLoading={isSubmitting}
             data-uie-name="schedule-meeting-modal-submit"
           >
             <CalendarIcon aria-hidden="true" css={submitButtonIconStyles} />
--- a/libraries/react-ui-kit/src/Inputs/Select/SelectOption/SelectOption.tsx
+++ b/libraries/react-ui-kit/src/Inputs/Select/SelectOption/SelectOption.tsx
@@ -38,7 +38,9 @@ export const SelectOption = <IsMulti extends boolean = false, Group extends Grou
         return;
       }
 
-      selectedOptionRef.current?.scrollIntoView({block: 'center'});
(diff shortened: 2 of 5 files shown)


apps/webapp/src/script/components/meeting/scheduleMeetingModal/scheduleMeetingModal.tsx
```diff
@@ -153,6 +153,7 @@ export const ScheduleMeetingModal = () => {
             css={submitButtonStyles}
             onClick={handleSubmit}
             disabled={isSubmitting}
+            showLoading={isSubmitting}
```
arjita-mitra, line 156: some test failure 

Summary of all failing tests
 FAIL  src/script/components/meeting/meetNowModal/meetNowModal.test.tsx
  ● MeetNowModal › does not dismiss the modal while submission is pending

libraries/react-ui-kit/src/Inputs/Select/SelectOption/SelectOption.tsx
```diff
@@ -29,10 +31,20 @@ export const SelectOption = <IsMulti extends boolean = false, Group extends Grou
...
   function SelectOptionComponent(props: OptionProps<Option, IsMulti, Group>) {
     const {children, data, isMulti, isSelected, options} = props;
+    const selectedOptionRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+      if (!isSelected) {
+        return;
+      }
+
+      selectedOptionRef.current?.scrollIntoView({block: 'center'});
```
arjita-mitra: when an option becomes selected it calls selectedOptionRef.current?.scrollIntoView({block: 'center'}) . In some test environments and older browsers that method can be undefined that will throw at runtime when current exists but has no scrollIntoView method.

Production code should be robust against missing DOM APIs, and relying on per test mocks is brittle`(Element.prototype.scrollIntoView = jest.fn();)`. Either the component should guard the call or there should be a single global polyfill/mocking setup for tests.

`if (selectedOptionRef.current && typeof selectedOptionRef.current.scrollIntoView === 'function') { selectedOptionRef.current.scrollIntoView({ block: 'center' }); } `

after adding this type check i don't think we will need to add per test mock too please check.

arjita-mitra (changes requested):

arjita-mitra (approved):

PR #22237, opened 2026-08-18
Show participant name when hovering over avatars also show organizer label [WPB-28046]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28046" title="WPB-28046" target="_blank">TaskWPB-28046</a>  [Web] Show participant name when hovering avatars
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
+++ b/apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
@@ -53,7 +53,7 @@ const renderParticipantAvatarTooltip = (participant: User, organizer?: User, ind
     <ThemeProvider>
       <ParticipantAvatarTooltip
         participant={participant}
-        organizer={organizer?.qualifiedId}
+        getLabel={name => (participant === organizer ? `${name} (${translateForTest('meetings.participant.organizer')})` : name)}
         index={index}
         avatarSize={AVATAR_SIZE.X_SMALL}
         avatarRingColor="black"
--- a/apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
+++ b/apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
@@ -17,7 +17,7 @@
  *
  */
 
-import {fireEvent, render, screen} from '@testing-library/react';
+import {act, fireEvent, render, screen, waitFor} from '@testing-library/react';
 import {ThemeProvider} from '@wireapp/react-ui-kit';
 
 import {AVATAR_SIZE} from 'Components/avatar';
@@ -27,6 +27,7 @@ import {
   createRootProviderWrapperForTest,
 } from 'src/script/page/testSupport/rootContextTestSupport';
 import {translateForTest} from 'Util/test/translateForTest';
+import type {Translate} from 'Util/localizerUtil';
 
 import {ParticipantAvatarTooltip} from './participantAvatarTooltip';
 
@@ -36,16 +37,24 @@ const createUser = (id: string, name: string) => {
   return user;
 };
 
-const renderParticipantAvatarTooltip = (participant: User, organizer?: User) => {
+const translateForTooltip: Translate = (key, substitutions) => {
+  if (key === 'meetings.participant.nameWithOrganizer' && substitutions) {
+    return `${substitutions.name} (${substitutions.organizer})`;
+  }
+
+  return translateForTest(key);
+};
+
+const renderParticipantAvatarTooltip = (participant: User, organizer?: User, index = 0) => {
   const rootProviderWrapper = createRootProviderWrapperForTest(
-    createRootContextValueForTest({translate: translateForTest}),
+    createRootContextValueForTest({translate: translateForTooltip}),
   );
   const result = render(
     <ThemeProvider>
       <ParticipantAvatarTooltip
         participant={participant}
         organizer={organizer?.qualifiedId}
-        index={0}
+        index={index}
         avatarSize={AVATAR_SIZE.X_SMALL}
         avatarRingColor="black"
       />
@@ -81,5 +90,50 @@ describe('ParticipantAvatarTooltip', () => {
     renderParticipantAvatarTooltip(organizer, organizer);
 
     expect(screen.getByRole('button')).toHaveAttribute('aria-label', 'Kim Organizer (meetings.participant.organizer)');
+  });
+
+  it('updates the tooltip and aria-label when the participant name changes', async () => {
+    const participant = createUser('participant', 'Alice Anderson');
+
+    renderParticipantAvatarTooltip(participant);
+    act(() => participant.name('Alex Anderson'));
+
+    await waitFor(() => {
+      expect(screen.getByRole('button')).toHaveAttribute('aria-label', 'Alex Anderson');
+    });
+  });
+
+  it('uses the unavailable-user label when the participant is unavailable', () => {
+    const participant = createUser('', '');
+
+    renderParticipantAvatarTooltip(participant);
+
+    expect(screen.getByRole('button')).toHaveAttribute('aria-label', 'unavailableUser');
+  });
+
+  it('shows the name belonging to the hovered stacked avatar', () => {
+    const firstParticipant = createUser('first', 'First Participant');
+    const secondParticipant = createUser('second', 'Second Participant');
+    const thirdParticipant = createUser('third', 'Third Participant');
+    const rootProviderWrapper = createRootProviderWrapperForTest(
+      createRootContextValueForTest({translate: translateForTooltip}),
+    );
+
+    const result = render(
+      <ThemeProvider>
+        <>
+          <ParticipantAvatarTooltip participant={firstParticipant} avatarSize={AVATAR_SIZE.X_SMALL} />
+          <ParticipantAvatarTooltip participant={secondParticipant} index={1} avatarSize={AVATAR_SIZE.X_SMALL} />
+          <ParticipantAvatarTooltip participant={thirdParticipant} index={2} avatarSize={AVATAR_SIZE.X_SMALL} />
+        </>
+      </ThemeProvider>,
+      {wrapper: rootProviderWrapper},
+    );
+    result.container.id = 'wire-app';
+
+    const tooltipWrappers = screen.getAllByRole('presentation');
+    fireEvent.mouseEnter(tooltipWrappers[1]);
+
+    expect(screen.getByText('Second Participant')).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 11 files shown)


apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  avatarRingColor,
+}: ParticipantAvatarTooltipProps) => {
+  const {translate} = useApplicationContext();
+  const isOrganizer = !isUndefined(organizer) && matchQualifiedIds(participant.qualifiedId, organizer);
+  const organizerLabel = translate('meetings.participant.organizer');
+  const label = `${participant.name()}${isOrganizer ? ` (${organizerLabel})` : ''}`;
+
+  return (
+    <Tooltip body={label}>
+      <div css={avatarRingColor ? avatarItemStyles(index, avatarRingColor) : undefined}>
```
thisisamir98: Could we keep the stack styles on the tooltip wrapper itself?

`avatarItemStyles` (negative margin and z-index) used to live on the flex child. They now sit on a div inside `Tooltip`, whose wrapper is `width: max-content` and is the actual flex item.

Overlap and stacking then no longer live on siblings in the same flex row. The stack can gap, and hovering the overlapping crescent can show the wrong name.

Something like this would keep the layout on the flex child:

```tsx
<Tooltip
  body={label}
  css={avatarRingColor ? avatarItemStyles(index, avatarRingColor) : undefined}
>
  <Avatar ... />
</Tooltip>
```

Could we add a regression: three stacked avatars, the second flex item has `margin-left: -8px`, and hovering the overlap shows the upper avatar's name?

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+  participant,
+  organizer,
+  index = 0,
+  avatarSize,
+  avatarRingColor,
+}: ParticipantAvatarTooltipProps) => {
+  const {translate} = useApplicationContext();
+  const isOrganizer = !isUndefined(organizer) && matchQualifiedIds(participant.qualifiedId, organizer);
+  const organizerLabel = translate('meetings.participant.organizer');
+  const label = `${participant.name()}${isOrganizer ? ` (${organizerLabel})` : ''}`;
```
thisisamir98: Could we build this label with `useUserName(participant)` instead of `participant.name()`?

`useUserName` subscribes to the Knockout name and maps unavailable users to `unavailableUser` ("Name not available"). This call is not reactive, so a name change will not update the tooltip or `aria-label` until a parent re-renders. An unavailable user can also get an empty or stale tooltip while the visible name already says "Name not available".

A single translation key such as `"{name} (Organizer)"` would also let translators rearrange the copy, instead of concatenating the name and the organizer word in English order.

Could we add a regression: the name updates after render, and the tooltip plus `aria-label` show the new value. An unavailable user shows "Name not available".

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <Tooltip body={label}>
+      <div css={avatarRingColor ? avatarItemStyles(index, avatarRingColor) : undefined}>
+        <Avatar
+          participant={participant}
+          aria-label={label}
+          title={undefined}
```
thisisamir98: Non-blocking: `title={undefined}` only works because `UserAvatar` and `TemporaryGuestAvatar` spread `props` after their own `title={name}`. If that order changes, the native browser tooltip and this one will both show.

A short comment (or a dedicated `hideTitle` on `Avatar`) would make that contract obvious.

apps/webapp/src/script/components/avatar/stackedAvatars/stackedAvatars.tsx
```diff
@@ -17,15 +17,19 @@
...
-import {avatarItemStyles, overflowCountStyles, wrapperStyles} from './stackedAvatars.styles';
+import {ParticipantAvatarTooltip} from './participantAvatarTooltip';
+import {overflowCountStyles, wrapperStyles} from './stackedAvatars.styles';
 
-import {Avatar, AVATAR_SIZE} from '../avatar';
+import {AVATAR_SIZE} from '../avatar';
 
 interface StackedAvatarsProps {
   participants: User[];
+  organizer?: QualifiedId;
```
thisisamir98: Could `StackedAvatars` stay a layout shell, and leave organizer matching to the meeting layer?

This component now takes `organizer` and always renders `ParticipantAvatarTooltip`, which reads `meetings.participant.organizer`. That ties a generic avatar widget to meeting copy.

The meeting layer already knows `qualifiedCreator`. It could pass a ready-made `label` (or `isOrganizer`) into a name-only tooltip.

apps/webapp/src/script/components/avatar/stackedAvatars/stackedAvatars.tsx
```diff
@@ -17,15 +17,19 @@
  *
  */
 
+import type {QualifiedId} from '@wireapp/api-client/lib/user';
+
+import {getStackedAvatarDisplay} from 'Components/avatar';
```
thisisamir98: Could we import `getStackedAvatarDisplay` from `./getStackedAvatarDisplay` again?

`Components/avatar` re-exports this module, so this barrel import is a cycle. The local path avoided that.

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/getMeetingParticipantsForDisplay.test.ts
```diff
@@ -29,23 +29,37 @@ const createUser = (id: string, name: string) => {
...
 
 describe('getMeetingParticipantsForDisplay', () => {
-  it('places the self user first when they are in the participants list', () => {
+  it('places the organizer first, followed by the self user and other participants', () => {
     const selfUser = createUser('self', 'Kim Organizer');
     const otherUser = createUser('other', 'Jaqueline Olaho');
+    const organizer = createUser('organizer', 'Organizer');
 
-    expect(getMeetingParticipantsForDisplay([otherUser, selfUser], selfUser)).toEqual([selfUser, otherUser]);
+    expect(getMeetingParticipantsForDisplay([otherUser, selfUser, organizer], selfUser, organizer.qualifiedId)).toEqual(
```
thisisamir98, line 37: Could we cover the production participant list here?

`participating_user_ets` does not include the self user. This test puts self in that list. The old case (self missing, organizer is someone else) was removed.

The implementation still injects self, but nothing locks `[organizer, self, others]` for that real input.

Could we add:
`getMeetingParticipantsForDisplay([other, organizer], self, organizer.qualifiedId)` equals `[organizer, self, other]`?

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/getMeetingParticipantsForDisplay.ts
```diff
@@ -17,13 +17,26 @@
...
 import type {User} from 'Repositories/entity/User';
 import {matchQualifiedIds} from 'Util/qualifiedId';
 
-export const getMeetingParticipantsForDisplay = (participants: User[], selfUser: User): User[] => {
+export const getMeetingParticipantsForDisplay = (
+  participants: User[],
+  selfUser: User,
+  qualifiedCreator: QualifiedId,
+): User[] => {
+  const organizer = participants.find(participant => matchQualifiedIds(participant.qualifiedId, qualifiedCreator));
```
thisisamir98: If the organizer is not in `participants` (they left, or the user entity never loaded), they are not placed first and they get no organizer label. Self is shown first instead.

Is that accepted for this partial tooltip, or should we still resolve `qualified_creator` so the organizer stays first when we have the user?

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    expect(screen.queryByText('Alice Anderson')).not.toBeInTheDocument();
+    container.removeAttribute('id');
+  });
+
+  it('adds the localized organizer label', () => {
+    const organizer = createUser('organizer', 'Kim Organizer');
+
+    renderParticipantAvatarTooltip(organizer, organizer);
+
+    expect(screen.getByRole('button')).toHaveAttribute('aria-label', 'Kim Organizer (meetings.participant.organizer)');
```
thisisamir98, line 94: Non-blocking: this only checks `aria-label`. Could we also hover and assert that the tooltip body shows the organizer suffix? The show/dismiss test above never covers that label.

thisisamir98 (commented):
Thanks for this. Hover names and the organizer label match the ticket. A few notes on stacked layout, name reactivity, and keeping meeting-specific copy out of the generic avatar shell.

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/getMeetingParticipantsForDisplay.ts
zskhan: b6b6fa5e78

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/getMeetingParticipantsForDisplay.test.ts
zskhan: b6b6fa5e78

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
zskhan: fdf64b2993

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
zskhan: fdf64b2993

apps/webapp/src/script/components/avatar/stackedAvatars/stackedAvatars.tsx
zskhan: 72b1d829d2

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.tsx
zskhan: 366d27ded6

apps/webapp/src/script/components/avatar/stackedAvatars/participantAvatarTooltip.test.tsx
zskhan: 366d27ded6

apps/webapp/src/script/components/avatar/stackedAvatars/stackedAvatars.tsx
zskhan: eeec1047dc16d760b5e2e8ab1cff997f3ac56950

thisisamir98 (approved):

PR #22242, opened 2026-08-18
Simplify meeting notification card [WPB-26695]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26695" title="WPB-26695" target="_blank">TaskWPB-26695</a>  [Web] Inform participants of a new meeting invite
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
@@ -41,7 +41,7 @@ export const MeetingCallingView = () => {
   }
 
   return (
-    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-view">
+    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-views">
       {activeCalls.map(call => (
         <CallingCell
           key={`${call.conversation.qualifiedId.id}-${call.conversation.qualifiedId.domain}`}
--- a/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
@@ -41,7 +41,7 @@ export const MeetingCallingView = () => {
   }
 
   return (
-    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-views">
+    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-view">
       {activeCalls.map(call => (
         <CallingCell
           key={`${call.conversation.qualifiedId.id}-${call.conversation.qualifiedId.domain}`}
--- a/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
@@ -61,6 +61,7 @@ const getOrganizer = (qualifiedCreator: QualifiedId) =>
     ?.name() ?? qualifiedCreator.id;
 
 const getMeetingTime = (meetingStartTime: string) => formatLocale(meetingStartTime, 'PP, p');
+const getMeetingTimeOnly = (meetingStartTime: string) => formatLocale(meetingStartTime, 'p');
 
 const MeetingNotificationOrganizerAndTimeMetadata = ({
   qualifiedCreator,
@@ -109,7 +110,7 @@ const MeetingNotificationMetadata = ({
     )
     .with({kind: MeetingNotificationKind.ONGOING}, ({qualifiedCreator, meetingStartTime}) => {
       const organizer = getOrganizer(qualifiedCreator);
-      const meetingTime = getMeetingTime(meetingStartTime);
+      const meetingTime = getMeetingTimeOnly(meetingStartTime);
 
       return (
         <>
--- a/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
@@ -18,7 +18,7 @@
  */
 
 import type {QualifiedId} from '@wireapp/api-client/lib/user';
-import {match} from 'ts-pattern';
+import {match, P} from 'ts-pattern';
 import {container} from 'tsyringe';
 
 import {Button, ButtonVariant, CalendarIcon} from '@wireapp/react-ui-kit';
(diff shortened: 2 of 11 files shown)


apps/webapp/src/script/components/meeting/meetingNotificationEventHandlers.ts
```diff
@@ -145,7 +131,25 @@ export const createMeetingNotificationEventHandlers = ({
...
+      if (hasInvalidDates) {
+        logger.warn('meeting notification received with invalid meeting dates', {
+          meetingId: meeting.qualified_id,
+          meetingStartTime: meeting.series_start_date,
+          meetingEndTime: meeting.series_end_date,
+        });
+      }
+
+      let kind = MeetingNotificationKind.INVITE;
+      if (startTimestamp <= now && endTimestamp > now) {
```
thisisamir98: Could we reuse `getMeetingTemporalStatusAt` on the current instance window instead of comparing `series_start_date` and `series_end_date` directly?

The meetings list already decides ongoing with that helper (`start` plus `duration_ms`, and the end is inclusive). This check disagrees in two cases:
1. At the exact end time, the list stays ongoing, but this becomes Invitation or Update because `endTimestamp > now` is false.
2. For a recurring series, `series_end_date` is the first occurrence end (`mapApiMeetingToSeries` copies API `end_time`). A weekly meeting that is live this week would be classified as INVITE.

Could we derive the current instance and reuse the existing status helper so the card matches the list?

Please add regressions: recurring series, first occurrence last week, now inside this week's instance, kind is ONGOING; one-shot meeting, now equal to end, same kind as the list.

apps/webapp/src/script/components/meeting/meetingNotificationEventHandlers.ts
```diff
@@ -145,7 +131,25 @@ export const createMeetingNotificationEventHandlers = ({
   return {
     notifyMeetingChange: meeting => {
       const meetingKey = toMeetingIdKey(meeting.qualified_id);
-      const kind = notifiedMeetings.has(meetingKey) ? MeetingNotificationKind.UPDATE : MeetingNotificationKind.INVITE;
+      const now = wallClock.currentTimestampInMilliseconds;
+      const startTimestamp = Date.parse(meeting.series_start_date);
+      const endTimestamp = Date.parse(meeting.series_end_date);
+      const hasInvalidDates = Number.isNaN(startTimestamp) || Number.isNaN(endTimestamp);
+
+      if (hasInvalidDates) {
```
thisisamir98, line 139: I think we should fail closed here instead of still creating an invitation.

Invalid dates are logged, then the `NaN` comparisons fall through and we still call `notifyForMeeting`. The card later formats that start time with `formatLocale`, which throws on an invalid date. `mapApiMeetingToSeries` already rejects these dates, so this path should stop after the warning.

Could we return after logging, and change the test so invalid dates produce a warning and no notification?

apps/webapp/src/script/components/meeting/meetingNotificationEventHandlers.test.ts
```diff
@@ -31,6 +32,9 @@ import {createMeetingNotificationEventHandlers} from './meetingNotificationEvent
 const meetingId: QualifiedId = {id: 'meeting-id', domain: 'example.com'};
 const creatorId: QualifiedId = {id: 'creator-id', domain: 'example.com'};
 const conversationId: QualifiedId = {id: 'conversation-id', domain: 'example.com'};
+const wallClock = createDeterministicWallClock({
```
thisisamir98, line 35: This clock is 12:00 and the fixture meeting ends at 11:00, so the INVITE then UPDATE example below is already in the past. It does not lock the upcoming invite path. A regression that inverted the window would still pass.

Could we cover `now` before start as INVITE, keep the during-window ONGOING case, and add an explicit after-end case?

apps/webapp/src/script/components/meeting/meetingNotificationEventHandlers.test.ts
```diff
@@ -167,11 +173,71 @@ describe('createMeetingNotificationEventHandlers', () => {
...
+
+  it('logs invalid meeting dates and falls back to an invitation', () => {
+    const notifications: AddNotificationInput[] = [];
+    const warnings: Array<{message: string; context?: unknown}> = [];
+    const invalidMeeting = {...meetingSeries, series_start_date: 'invalid-start'};
+    const {notifyMeetingChange} = createHandlers({notifications, warnings});
+
+    notifyMeetingChange(invalidMeeting);
+
+    expect(notifications[0]?.kind).toBe(MeetingNotificationKind.INVITE);
```
thisisamir98: This currently locks the fail-open path (invalid dates still become INVITE).

If we stop after the warning, could this expect no notification instead of `INVITE`?

apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -70,40 +92,29 @@ const MeetingNotificationMetadata = ({
...
-        </>
-      );
-    })
-    .with({kind: MeetingNotificationKind.UPDATE}, ({meetingStartTime}) =>
-      translate('meetings.notifications.newTime', {time: getMeetingTime(meetingStartTime)}),
+    .with(
+      {
+        kind: P.union(
+          MeetingNotificationKind.INVITE,
+          MeetingNotificationKind.UPDATE,
```
thisisamir98, line 99: UPDATE cards used to show `meetings.notifications.newTime`. They now use the same "By organizer" line as invites.

WPB-26695 called out the new datetime as the update content. Is dropping "New time" agreed with design?

apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
```diff
@@ -41,7 +41,7 @@ export const MeetingCallingView = () => {
   }
 
   return (
-    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-view">
+    <div css={meetingCallingViewStyles} data-uie-name="meeting-calling-views">
```
thisisamir98: Is this rename intentional? It looks unrelated to the notification card, and the empty-state test still queries `meeting-calling-view`, so it would not catch a missing wrapper.

Could we revert this unless QA needs the new name?

thisisamir98 (commented):
Thanks for this. The shared "By organizer" metadata and classifying a member-added event as ongoing when it arrives during the meeting window look like the right M1 direction.

A few things I would like us to tighten: the time check should match `getMeetingTemporalStatusAt` on the current instance, invalid dates should not still emit a card, and the main invite test currently uses a meeting that has already ended. There is also an unrelated `data-uie-name` change on the calling view.

Questions:
- If the first `notifyMeetingChange` arrives after the meeting ended, we still show Invitation with a past datetime. Is that what we want for a late Meet Now event?
- `notifyUpdate` always emits UPDATE, even during the window. Only member-added uses the ongoing heuristic. Is that intentional?

apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
zskhan: Currently there is no changelog coming from backend so removing this newTime from the update card to avoid confusion.

apps/webapp/src/script/components/meeting/meetingCallingView/meetingCallingView.tsx
zskhan: It was a typo. Thanks for identifying.

thisisamir98 (approved):
Thanks Zafar. I went through the follow-up and the review comments are covered.

`getMeetingInstanceAt` now uses the same temporal status as the meetings list, including the current occurrence of a recurring series and the exact end time. Invalid dates stop after the warning, the invite path is tested while the meeting is still upcoming, and the calling view `data-uie-name` is back. Dropping "New time" also makes sense while the backend has no changelog.

Approved.

PR #22248, opened 2026-08-19
Preserve meeting participant names and unify avatar rendering [WPB-28046]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28046" title="WPB-28046" target="_blank">TaskWPB-28046</a>  [Web] Show participant name when hovering avatars
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/meetingParticipants.test.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/meetingParticipants.test.tsx
@@ -1,0 +1,144 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {act, fireEvent, render, screen} from '@testing-library/react';
+import type {QualifiedId} from '@wireapp/api-client/lib/user';
+import {CONVERSATION_PROTOCOL} from '@wireapp/api-client/lib/team';
+import {ThemeProvider} from '@wireapp/react-ui-kit';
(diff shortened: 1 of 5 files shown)


apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/meetingParticipants.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    organizer.name(specialCharacterName);
+    const selfUser = new User('self-id', 'example.com', translate);
+    const conversation = new Conversation(
+      qualifiedConversation.id,
+      qualifiedConversation.domain,
+      CONVERSATION_PROTOCOL.PROTEUS,
+      translate,
+    );
+    conversation.participating_user_ets([organizer]);
+    userState.self(selfUser);
```
thisisamir98, line 61: Could we also cover the single-participant layout this PR removes?

This setup uses a distinct self user, so `getMeetingParticipantsForDisplay` returns the organizer and self. Names were already tooltip-only for two people. The test would still pass if we brought back the `participants.length === 1` branch that showed `<UserName />` next to the avatar.

Could we add a case where self is the organizer and there are no other participants, then assert:
- the avatar is present
- the name is not in the document before hover
- after hover, the tooltip shows the raw name with the Organizer suffix

That would lock the layout change, not only the encoding fix.

thisisamir98 (commented):
Looks good overall. `skipEscape` is the right call for these React text nodes, and always using `StackedAvatars` matches names on hover only.

One request on the new `MeetingParticipants` test (inline). Also: this PR is on WPB-22420 (general maintenance). The layout change reads like a follow-up to WPB-28046. Is the maintenance ticket intentional?

apps/webapp/src/script/components/meeting/meetingList/meetingListItemGroup/meetingListItem/meetingParticipants/meetingParticipants.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+      const rendered = render(
+        <ThemeProvider>
+          <MeetingParticipants qualifiedConversation={qualifiedConversation} qualifiedCreator={qualifiedCreator} />
+        </ThemeProvider>,
+        {wrapper: createRootProviderWrapperForTest(createRootContextValueForTest({translate}))},
+      );
+
+      const avatar = screen.getByLabelText(`${specialCharacterName} (Organizer)`);
+      expect(avatar).toBeInTheDocument();
+      const tooltipWrapper = document.querySelector('[data-testid="tooltip-wrapper"]');
```
thisisamir98: Non-blocking: this selects the first tooltip wrapper. That is the organizer today because the organizer is rendered first. If order changes, we could hover a different avatar.

Could we start from the avatar we already found, for example `avatar.closest('[data-testid="tooltip-wrapper"]')`, or use `getByRole('presentation')` like the stacked avatar tooltip tests?

thisisamir98 (approved):

PR #22251, opened 2026-08-19
fix: count departed users in message reactions [WPB-22420]

MohamadJaara:
<table>
<td>
  </td></table>
  <br />
 

## Summary

- Count inline message reactions from the complete stored reaction ID list.
- Keep the current conversation member list only for resolving tooltip names (this is also can be an issue but for now will leave it as is)
- Add a regression test for reactions from users who left the group.

## Root cause

The inline reaction pill derived its count from users matched against the current conversation member list. When a reactor left the group, that lookup filtered them out even though their reaction ID remained on the message. Message details used the complete reaction data, causing the two views to disagree.

## Steps to reproduce

1. Have four group members react to a message with the same emoji.
2. Remove one of those members from the group.
3. Check the reaction pill in the message list and then open message details.
4. The pill shows three reactions while message details shows four.

## Fix

Pass the number of IDs stored for each reaction to the emoji pill. Current conversation members are still used for tooltip names, but they no longer determine the reaction count.

## Impact

Inline reaction counts now stay consistent with message details after a reactor leaves the group.

## Validation

- The focused reaction component tests pass, including the departed-user regression case.
- Workspace type checks pass.
- Affected lint, formatting, and style checks pass.

### DIFF
--- a/apps/webapp/src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/EmojiPill.tsx
+++ b/apps/webapp/src/script/components/MessagesList/Message/ContentMessage/MessageActions/MessageReactions/EmojiPill.tsx
@@ -49,6 +49,7 @@ interface EmojiPillProps {
   isRemovedFromConversation: boolean;
   index: number;
   emojiListCount: number;
+  emojiCount: number;
   hasUserReacted: boolean;
   reactingUsers: User[];
 }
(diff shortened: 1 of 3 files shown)


arjita-mitra:
@MohamadJaara this PR changes user facing behaviour. Its a matter of product and ux decision whether or not we want to list departed user in tooltip and message details. I agree it should be consistent whatever we choose. Is paul hinterberger and Wolfgang Lattermann aware of this change? Is this the behaviour in other clients?

arjita-mitra:
@MohamadJaara this PR changes user facing behaviour. Its a matter of product and ux decision whether or not we want to list departed user in tooltip and message details. I agree it should be consistent whatever we choose. Is CM scope PM and UX designer aware of this change? Is this the behaviour in other clients?

MohamadJaara:
> @MohamadJaara this PR changes user facing behaviour. Its a matter of product and ux decision whether or not we want to list departed user in tooltip and message details. I agree it should be consistent whatever we choose. Is CM scope PM and UX designer aware of this change? Is this the behaviour in other clients?

looking at the code it is clearly a bug, the app resolves the names from the conv members (because that data is just there) while the index DB report all users that have reacted, other clients does not do this, this does not change UX it fixes a bug imho, anyway if you think otherwise please just close the PR no hard feelings

also i explicity didn't fix the same issue with the tooltip because that will require a bit more work to do that i am not comfortable doing. So I fixed the obvious visual bug where the messages list say there is 3 likes, while if the user open the message details they see 4

screendriver (approved):

PR #22271, opened 2026-08-21
Update url on navigation [WPB-27977]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27977" title="WPB-27977" target="_blank">BugWPB-27977</a>  User is redirected to All Conversations after refreshing the Wire Meetings page
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
+++ b/apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
@@ -76,10 +76,14 @@ import {
 
 import {Config} from '../../../../Config';
 import {generateConversationUrl} from '../../../../router/routeGenerator';
+import {setHistoryParam} from '../../../../router/Router';
 import {createNavigateKeyboard} from '../../../../router/routerBindings';
 import {ListViewModel} from '../../../../view_model/ListViewModel';
 import {ListWrapper} from '../listWrapper';
 import {StartUI} from '../startUi';
+
+export const shouldClearDeepLinkForTab = (tab: SidebarTabs): boolean =>
+  ![SidebarTabs.PREFERENCES, SidebarTabs.MEETINGS].includes(tab);
 
 type ConversationsProps = {
   callState?: CallState;
@@ -352,7 +356,8 @@ export const Conversations = ({
         void conversationRepository.updateArchivedConversations();
       }
 
-      if (![SidebarTabs.PREFERENCES, SidebarTabs.CELLS, SidebarTabs.MEETINGS].includes(nextTab)) {
+      if (shouldClearDeepLinkForTab(nextTab)) {
+        setHistoryParam('/');
         onExitPreferences();
       }
 
@@ -365,8 +370,7 @@ export const Conversations = ({
         if (!isMeetingsEnabled) {
           return;
         }
-        switchList(ListState.MEETINGS);
-        switchContent(ContentState.MEETINGS);
+        listViewModel.openMeetingsList();
       }
 
       clearConversationFilter();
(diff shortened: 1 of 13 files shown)


apps/webapp/src/script/router/Router.ts
```diff
@@ -90,5 +90,15 @@ export const navigate = (path: string) => {
...
+  const hashUnchanged = window.location.hash === `#${path}`;
   window.location.hash = path;
+
+  if (hashUnchanged) {
+    // Setting the hash to its current value does not fire a native `hashchange` event, so route
+    // handlers that rely on it (e.g. re-showing the most recent conversation) would otherwise
+    // never run. Force route re-evaluation async, matching the timing of a real hashchange, so
+    // callers that synchronously update content state right after calling this still take effect
+    // first.
+    setTimeout(parseRoute, 0);
```
thisisamir98: When the hash is already `/`, this schedules `parseRoute` on the next tick. `/` always opens the most recent conversation (`showMostRecentConversation` in `appMain`).

`changeTab` for Cells or Connect often calls `setHistoryParam('/')` while already on `/`. The timeout then runs after the tab switch. `showConversation` maps Cells back to Recent, so the user lands in a conversation instead of Files or Connect.

Could we skip the forced parse when the next view is not a conversation, or make `/` a no-op while Cells, Connect, Preferences, or Meetings is selected?

Please add a regression: start on Recent (`#/`), open Cells, wait one macrotask, still on Files, not a conversation.

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
```diff
@@ -352,8 +368,19 @@ export const Conversations = ({
...
+      }
+
+      clearConversationFilter();
+      setCurrentTab(nextTab);
+
+      const {listState} = useAppState.getState();
+      const isSwitchingConversationListTab = listState === ListState.CONVERSATIONS && isConversationListTab(nextTab);
+
+      if (shouldClearDeepLinkForTab(nextTab) && !isSwitchingConversationListTab) {
+        setHistoryParam('/');
```
thisisamir98, line 382: This is the call that hits the Router race. After `setCurrentTab(CELLS)` you still clear to `/`. If the hash did not change, Router re-parses `/` and the conversation handler can undo Cells.

Could we avoid forcing a `/` re-parse when switching to Cells or Connect? Setting `setHistoryParam('/')` is fine for leaving Meetings or Preferences. Re-running the `/` route should not open a conversation in that case.

apps/webapp/src/script/view_model/ContentViewModel.test.ts
```diff
@@ -0,0 +1,115 @@
...
+};
+
+describe('ContentViewModel', () => {
+  it.each([SidebarTabs.CONNECT, SidebarTabs.PREFERENCES, SidebarTabs.CELLS, SidebarTabs.MEETINGS])(
+    'keeps the %s tab selected after a conversation navigation is triggered underneath it',
+    async tab => {
+      const contentViewModel = buildContentViewModel();
+      useSidebarStore.getState().setCurrentTab(tab);
+
+      await expect(contentViewModel.showConversation({domain: 'example.com', id: 'conversation-id'})).rejects.toThrow();
```
thisisamir98: This asserts that conversation lookup throws, then that the tab is unchanged. That does not lock the real case: a conversation is shown while Cells or Connect is selected, and the tab stays.

Could we drive a successful `showConversation` and assert tab plus content, instead of a failed lookup?

thisisamir98 (changes requested):
Meetings URL restore on refresh looks good.

Clearing the hash to `/` can reopen the last conversation and steal Cells or Connect a tick later. Please fix that race before merge.

If Meetings is turned off while that tab is selected, also clear `#/meetings`. The tab-keep tests should lock a successful showConversation, not a thrown lookup.

thisisamir98 (approved):

PR #22305, opened 2026-08-27
feat: send meeting tzid and expand recurrence in the creator's zone [WPB-27613]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27613" title="WPB-27613" target="_blank">StoryWPB-27613</a>  [Web] Compute meeting recurrence from the creator's time zone
  </td></table>
  <br />
 

## Summary
- Send the device IANA `tzid` on schedule and Meet now (API v17 requires it). Persist it on the series. Omit it on update so the zone stays immutable for M1.
- Expand recurring instances with `TZDate` in the stored zone so local wall time stays stable across DST (for example 11:00 Berlin stays 11:00 after 25 Oct 2026).
- Fail closed when a list payload omits `tzid`. The mapper no longer synthesizes `Europe/Berlin`, schema/API tests lock the v17 contract, and weekly DST is covered on the list paging helpers.

Backend: https://github.com/wireapp/wire-server/pull/5391 (WPB-27553)

## Test plan
- [ ] Schedule a weekly meeting and confirm create payload includes `tzid` matching the device IANA zone
- [ ] Open the meetings list across a DST change in that zone and confirm wall-clock start time is unchanged
- [ ] Start Meet now and confirm the 201 response includes `tzid`
- [ ] Edit title only and confirm PUT does not send `tzid`, while the series still keeps the original zone
- [ ] Confirm there is no time-zone picker in schedule/edit UI
- [ ] `yarn nx run api-client-lib:test --testFile=src/meetings/meetingSchema.test.ts`
- [ ] `yarn nx run api-client-lib:test --testFile=src/meetings/meetingsApi.test.ts`
- [ ] `yarn nx run webapp:test --testFile=src/script/components/meeting/selectors/getMeetingInstancesInRange.test.ts`


zskhan (dismissed):

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -217,6 +217,43 @@ describe('meetingSchema', () => {
...
+  });
+
+  it('strips deprecated trial and keeps tzid on a V17 payload', () => {
+    const result = meetingSchema.safeParse({...validMeeting, trial: false});
+
+    expect(result.success).toBe(true);
+
+    if (!result.success) {
+      throw new Error('Expected meeting schema parse to succeed');
+    }
```
screendriver: Tests should not throw. Tests should always `assert()` when something is not expected.

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -217,6 +217,43 @@ describe('meetingSchema', () => {
...
+  });
+
+  it('preserves tzid on a valid meeting payload', () => {
+    const result = meetingSchema.safeParse(validMeeting);
+
+    expect(result.success).toBe(true);
+
+    if (!result.success) {
+      throw new Error('Expected meeting schema parse to succeed');
+    }
```
screendriver: Tests should not throw. Tests should always `assert()` when something is not expected.

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -217,6 +217,43 @@ describe('meetingSchema', () => {
     expect(meetingSchema.safeParse(validMeeting).success).toBe(true);
   });
 
+  it('requires tzid on meeting payloads', () => {
+    const {tzid: _tzid, ...meetingWithoutTzid} = validMeeting;
+
+    expect(meetingSchema.safeParse(meetingWithoutTzid).success).toBe(false);
+  });
+
+  it('rejects a V16-shaped list item that has trial but no tzid', () => {
```
screendriver: Are we intentionally dropping support for the V16 meetings API? The backend explicitly still exposes a V15/V16 meeting shape without `tzid` as far as I can see it and the WebApp can negotiate API versions below V17. With the schema now requiring `tzid` a valid V16 response will fail parsing.

If meetings in general now require API >= 17 I think that we should introduce an explicit feature/API-version gate. Otherwise we probably need version-specific V16/V17 schemas instead of asserting that a valid V16 payload must fail.

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.ts
```diff
@@ -28,20 +29,34 @@ import {getMeetingTemporalStatusAt, MeetingTemporalStatuses} from 'Components/me
 const daysPerWeek = 7;
 const daysPerBiweeklyPeriod = 14;
 const daysPerFourWeeksPeriod = 28;
+const fallbackMeetingTimeZoneId = 'Europe/Berlin';
```
screendriver: Why do we need a fallback here? `tzid` is required by `MeetingSeries` and validated as required. Falling back to `Europe/Berlin` seems to undo the behavior introduced by this pull request changes and could silently hide an invalid internal state.

It is also not a safe V16 compatibility mechanism because the backend timezone is configurable. I would make the invariant explicit and use `meetingSeries.tzid` directly.

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.ts
```diff
@@ -28,20 +29,34 @@ import {getMeetingTemporalStatusAt, MeetingTemporalStatuses} from 'Components/me
...
 
-const createMeetingInstance = (meetingSeries: MeetingSeries, start: Date): MeetingInstance => ({
-  meetingSeries,
-  start,
-  end: new Date(start.getTime() + meetingSeries.duration_ms),
-});
+const getMeetingSeriesTimeZoneId = (meetingSeries: MeetingSeries): string =>
+  meetingSeries.tzid || fallbackMeetingTimeZoneId;
```
screendriver: An empty string becomes `Berlin` even though the schema says an empty string is invalid

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.test.ts
```diff
@@ -122,6 +125,41 @@ describe('getMeetingInstancesInRange', () => {
     ]);
   });
 
+  it('keeps weekly local wall time across Europe/Berlin DST end', () => {
```
screendriver, line 128: Could we also cover the DST-start direction? For example using a different IANA timezone? At the moment both new DST tests cover Europe/Berlin when the offset increases back from UTC+2 to UTC+1.

For example a weekly `America/New_York` meeting at 11:00 could assert `2026-03-02T16:00:00Z` ➡️  `2026-03-09T15:00:00Z` across DST start. That would cover the opposite offset transition.

libraries/api-client/src/meetings/meetingSchema.ts
```diff
@@ -50,7 +50,7 @@ const meetingFieldsSchema = {
   recurrence: meetingRecurrenceSchema.optional(),
   start_time: utcTimeSchema,
   title: z.string().min(1).max(256),
-  trial: z.boolean(),
+  tzid: z.string().min(1),
```
screendriver, line 53: Should this validate an actual IANA time-zone identifier rather than just a non-empty string? Right now "banana" is accepted as a perfectly valid timezone.

libraries/api-client/src/meetings/meetingSchema.test.ts
```diff
@@ -217,6 +217,43 @@ describe('meetingSchema', () => {
...
+
+    if (!result.success) {
+      throw new Error('Expected meeting schema parse to succeed');
+    }
+
+    expect(result.data.tzid).toBe('Europe/Berlin');
+    expect(result.data).not.toHaveProperty('trial');
+  });
+
+  it('preserves tzid on a valid meeting payload', () => {
```
screendriver: The test is redundant. The previous test already checks that `tzid` is preserved

libraries/api-client/src/meetings/meetingsApi.test.ts
```diff
@@ -124,17 +140,18 @@ describe('MeetingsAPI', () => {
...
-    await client.useVersion(MINIMUM_API_VERSION, 16);
+    await client.useVersion(MINIMUM_API_VERSION, 17);
 
     jest.spyOn(client.transport.http, 'sendJSON').mockResolvedValue({data: validMeetingWithConversation} as never);
 
     const meeting = await client.api.meetings.createMeeting({
       title: validMeeting.title,
       start_time: validMeeting.start_time,
       end_time: validMeeting.end_time,
+      tzid: validMeeting.tzid,
```
screendriver, line 140: Since sending `tzid` in the V17 POST body is one of the main changes introduced by this pull request could this test also assert the actual request data? At the moment it proves response parsing but would still pass if `createMeeting()` stopped forwarding `tzid`

libraries/api-client/src/meetings/meetingsApi.test.ts
```diff
@@ -96,23 +96,39 @@ describe('MeetingsAPI', () => {
...
     jest.spyOn(client.transport.http, 'sendJSON').mockResolvedValue({data: [{title: 'invalid'}]} as never);
 
     await expect(client.api.meetings.getMeetingsList()).rejects.toBeInstanceOf(ZodError);
   });
 
+  it('rejects list responses that omit tzid even when deprecated trial is present', async () => {
+    const client = new APIClient(testConfig);
+    jest.spyOn(client.transport.http, 'sendRequest').mockResolvedValue({
+      data: {supported: [MINIMUM_API_VERSION, 17], domain: 'test.zinfra.io'},
+    } as never);
```
screendriver: Could we avoid introducing more `jest.spyOn()` here? We're mutating an already-created dependency rather than injecting the behavior the test needs and the `as never` asserts are another indication that the test is crossing the wrong boundary.

screendriver (changes requested):

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: Meetings are client-supported from API v17 only. Added `canUseMeetings` so the tab, store, and `/meetings` route stay off when the negotiated version is below 17. The schema stays v17 (`tzid` required).

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872407016
https://github.com/wireapp/wire-webapp/commit/98c415b876c9d7bbb02c38a01546b610032cd4d0

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: Replaced the throw with `assert(result.success)` so the success path is narrowed without throwing.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872276334
https://github.com/wireapp/wire-webapp/commit/a4b69fbb4f2aeefda75c7181b3cea3e691115464

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: Same here: `assert(result.success)` instead of throw. (This test was later removed as redundant.)

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872277307
https://github.com/wireapp/wire-webapp/commit/b2db0a0a4db62bd04bfec69c212aafaa7dd64ab5

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.ts
thisisamir98: Removed the Berlin fallback. Recurrence now uses `meetingSeries.tzid` directly.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872423268
https://github.com/wireapp/wire-webapp/commit/3866a1b42ca499690309cc4df0ec11ff715f3648

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.ts
thisisamir98: Empty string no longer becomes Berlin. There is no fallback left.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872427980
https://github.com/wireapp/wire-webapp/commit/3866a1b42ca499690309cc4df0ec11ff715f3648

apps/webapp/src/script/components/meeting/selectors/getMeetingInstancesInRange.test.ts
thisisamir98: Added weekly America/New_York coverage across DST start (`2026-03-02T16:00:00Z` to `2026-03-09T15:00:00Z`), including the paging helpers.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872446624
https://github.com/wireapp/wire-webapp/commit/f1c3777b25354066a3b0792fc873b1c8b4576d07

libraries/api-client/src/meetings/meetingSchema.ts
thisisamir98: Keeping `tzid` as a non-empty string. IANA validity is the backend's job. The client sends the device id as-is; an invalid value is a 400 on create.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872458301

libraries/api-client/src/meetings/meetingSchema.test.ts
thisisamir98: Removed it. The trial-stripping test already asserts `tzid` is kept.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872466242
https://github.com/wireapp/wire-webapp/commit/de8a1561963f04d25b1584ce78d79f8068857ec2

libraries/api-client/src/meetings/meetingsApi.test.ts
thisisamir98: The create test now asserts the POST body includes `tzid`.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872482043
https://github.com/wireapp/wire-webapp/commit/8b91b1824a0f1f3c3f3742ad5872bc353a16000b

libraries/api-client/src/meetings/meetingsApi.test.ts
thisisamir98: Dropped that API test. Missing `tzid` stays locked in the schema tests, without another `jest.spyOn` on the HTTP client.

https://github.com/wireapp/wire-webapp/pull/22305#discussion_r3872492960
https://github.com/wireapp/wire-webapp/commit/6aaac57297

screendriver (approved):

PR #22306, opened 2026-08-27
Fix navigation issue while archiving/unarchiving conversation [WPB-27977]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27977" title="WPB-27977" target="_blank">BugWPB-27977</a>  User is redirected to All Conversations after refreshing the Wire Meetings page
  </td></table>
  <br />

### DIFF
--- a/apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
+++ b/apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {isNonEmptyString, isNullOrUndefined} from '@sindresorhus/is';
+import {isNonEmptyString, isNullOrUndefined, isUndefined} from '@sindresorhus/is';
 import {
   ADD_PERMISSION,
   Conversation as BackendConversation,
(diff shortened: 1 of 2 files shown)


arjita-mitra:
@zskhan could you write a little details how this nextConversationEntity addition fixes the issue?

zskhan:
> @zskhan could you write a little details how this nextConversationEntity addition fixes the issue?

When archiving the only active conversation, `nextConversationEntity` is `undefined` because there is no other conversation to display so we block `CONVERSATION.SHOW` when there is no other converstaion.
Calling CONVERSATION.SHOW with `undefined` clears the URL but cannot open another conversation.

thisisamir98 (dismissed):

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4442,7 +4442,7 @@ export class ConversationRepository {
       await this.clearConversationContent(conversationEntity, conversationEntity.cleared_timestamp());
     }
 
-    if (isActiveConversation && conversationEntity.is_archived()) {
+    if (isActiveConversation && conversationEntity.is_archived() && nextConversationEntity) {
```
screendriver: Is `nextConversationEntity` a boolean?

zskhan: Fixed in 93e907d8926f62bc1623fc124ed3facd9022db32

thisisamir98 (approved):

arjita-mitra:
> > @zskhan could you write a little details how this nextConversationEntity addition fixes the issue?
> 
> When archiving the only active conversation, `nextConversationEntity` is `undefined` because there is no other conversation to display so we block `CONVERSATION.SHOW` when there is no other converstaion. Calling CONVERSATION.SHOW with `undefined` clears the URL but cannot open another conversation.

Nice. my understanding was right then from the visual difference between mine and QA reproduction steps

PR #22307, opened 2026-08-27
fix: fullscreen modal profile name(WPB-20724)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-20724" title="WPB-20724" target="_blank">BugWPB-20724</a>  [Web] Improve header of the full screen file view
  </td></table>
  <br />
 

Inside fulscreenmodal users user name was coming instead of profile name when user open a file from cells file list

### BEFORE
Wire 2026-08-27 at 9_10 AM (1)
Wire 2026-08-27 at 10_11 AM

### AFTER
Wire 2026-08-27 at 9_10 AM

### DIFF
--- a/apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
+++ b/apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
@@ -25,6 +25,7 @@ import {getSelfUserDriveRole} from 'Components/conversation/conversationCells/co
 import {Conversation} from 'Repositories/entity/Conversation';
 import {User} from 'Repositories/entity/User';
 import {CellNode, CellNodeType} from 'src/script/types/cellNode';
+import {matchQualifiedIds} from 'Util/qualifiedId';
 import {TIME_IN_MILLIS} from 'Util/timeUtil';
 import {formatBytes, getFileExtension, getName} from 'Util/util';
 
@@ -43,7 +44,6 @@ export const transformCellsNodes = ({
 }): CellNode[] => {
   return nodes.map(node => {
     const id = node.Uuid;
-    const owner = getOwner(node);
     const name = getName(node.Path);
     const sizeMb = getFileSize(node);
     const uploadedAtTimestamp = getUploadedAtTimestamp(node);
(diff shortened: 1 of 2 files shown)


cristianoliveira (dismissed):

apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
```diff
@@ -69,7 +69,8 @@ export const transformCellsNodes = ({
     });
 
     const userQualifiedId = getUserQualifiedIdFromNode(node);
-    const user = users.find(user => user.qualifiedId.id === userQualifiedId?.id) ?? null;
+    const user = users.find(user => matchQualifiedIds(user.qualifiedId, userQualifiedId ?? undefined)) ?? null;
```
screendriver: Could we simplify this a bit? The combination of `?.`, `?? undefined`, `?? null` and `||` makes the data flow quite difficult to understand at a glance.

In particular we are converting `null` from `getUserQualifiedIdFromNode()` to `undefined` for `matchQualifiedIds()` and then converting a missing result from `find()` (which returns `undefined` when nothing was found) back to `null`

I think it would be easier to read and maintain if we normalized the absence case once and kept the lookup/fallback logic more explicit. For example we could work with a `Maybe` here to model general absence

screendriver (changes requested):

apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
arjita-mitra: Addressed it 1f8c24e4a69613d9d821196f97439a6cbcee431e

apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
```diff
@@ -68,8 +69,12 @@ export const transformCellsNodes = ({
       selfUserTeamId,
     });
 
-    const userQualifiedId = getUserQualifiedIdFromNode(node);
-    const user = users.find(user => user.qualifiedId.id === userQualifiedId?.id) ?? null;
+    const userQualifiedId = Maybe.of(getUserQualifiedIdFromNode(node));
+    const matchingUser = userQualifiedId.andThen(userQualifiedId =>
+      Maybe.of(users.find(user => matchQualifiedIds(user.qualifiedId, userQualifiedId))),
```
screendriver, line 74: Using [find](https://true-myth.dev/api/maybe/functions/find.html#function-find) would be easier here 🙂

arjita-mitra: will take care of it in next PR.

screendriver (approved):

apps/webapp/src/script/components/cells/common/transformCellsNodes/transformCellsNodes.ts
arjita-mitra: Addressed in https://github.com/wireapp/wire-webapp/pull/22313

PR #22354, opened 2026-09-01
fix overlapping of CallingCell.tsx and meetingNotificationHost.tsx [WPB-27879]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27879" title="WPB-27879" target="_blank">BugWPB-27879</a>  User cannot access the minimized call view while meeting notifications are displayed - Also affecting the normal calls UI
  </td></table>
  <br />
 

https://github.com/user-attachments/assets/e999ba42-9757-4f82-9c4c-90cf58064d24

### DIFF
--- a/apps/webapp/src/script/components/calling/CallingCell/CallingCell.styles.ts
+++ b/apps/webapp/src/script/components/calling/CallingCell/CallingCell.styles.ts
@@ -19,11 +19,16 @@
 
 import {CSSObject} from '@emotion/react';
 
-export const callingContainer: CSSObject = {
+import {meetingNotificationHostCollapsedOffset} from 'Components/meeting/meetingNotificationHost/meetingNotificationHost.styles';
+
+const callingContainerBottomPadding = 20;
+
+export const callingContainer = (hasNotifications: boolean): CSSObject => ({
   position: 'relative',
   display: 'flex',
   flexDirection: 'column',
   flexShrink: '0',
-  padding: '10px 12px 20px',
+  padding: `10px 12px ${callingContainerBottomPadding + (hasNotifications ? meetingNotificationHostCollapsedOffset : 0)}px`,
+  transition: 'padding-bottom var(--animation-timing-fast) ease-in-out',
   animation: 'show-call-ui var(--animation-timing-fast) ease-in-out 0s 1',
-};
+});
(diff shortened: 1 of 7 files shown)


apps/webapp/src/script/components/meeting/meetingNotificationHost/meetingNotificationHost.tsx
```diff
@@ -43,13 +45,38 @@ export const MeetingNotificationHost = ({isStandalone}: MeetingNotificationHostP
...
+    if (notifications.length === 0) {
+      return;
+    }
+
+    const collapseWhenClickedOutside = (event: MouseEvent) => {
+      const target = event.target;
+      const isInsideNotificationHost =
+        target instanceof Node &&
+        (hostRef.current?.contains(target) ||
+          (target instanceof Element && target.closest('[data-uie-name="meeting-notification-host"]') !== null));
```
ikotarac: in team management for example, it was decided to remove these **data-uie-name** attributes from user facing app in a build step, because they are used only in unit testing.

I don't know if web has a similar interpretation of these, but if they are only meant for testing, this might be a brittle approach.

apps/webapp/src/script/components/calling/CallingCell/CallingCell.styles.ts
```diff
@@ -19,11 +19,16 @@
...
+
+const callingContainerBottomPadding = 20;
+
+export const callingContainer = (hasNotifications: boolean): CSSObject => ({
   position: 'relative',
   display: 'flex',
   flexDirection: 'column',
   flexShrink: '0',
-  padding: '10px 12px 20px',
+  padding: `10px 12px ${callingContainerBottomPadding + (hasNotifications ? meetingNotificationHostCollapsedOffset : 0)}px`,
```
ikotarac, line 31: could we, instead of relying on hardcoded sizes, put calling cell and meeting notification host in a common parent (flex container) that is positioned in the corner of the screen and that automatically handles layout of its children?

I don't like that now calling cell is aware of other siblings in the area, if we also start adding other elements, we can't make everyone aware of everyone else and basically reinvent the layout algorithm

zskhan: This is currently implemented this way because, in the future, we’re planning to move these notifications somewhere else. So I’d prefer not to introduce a common parent/layout specifically for the current positioning if it’s going to be temporary.

e-maad (approved):

PR #22355, opened 2026-09-02
feat: integrate Shared Drive direct upload module [WPB-28356]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28356" title="WPB-28356" target="_blank">TaskWPB-28356</a>  [web] create logic library for handling upload (non-ui)
  </td></table>
  <br />
 

## Problem

Shared Drive upload actions were not connected to the Cells upload lifecycle, so users could not upload files directly into the current Shared Drive folder.

## Solution

Connect the upload action to the Cells upload manager, publish files to the current folder, and scope upload state to the Shared Drive UI through a local provider.

## Demo

https://github.com/user-attachments/assets/57483a35-31e6-492f-9a0b-afbfcc4e796a

https://wearezeta.atlassian.net/browse/WPB-28356


apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
```diff
@@ -241,6 +245,24 @@ export const ConversationCells = memo(
     return (
       <CellsSelfUserDriveRoleProvider selfUserDriveRole={selfUserDriveRole}>
         <div css={wrapperStyles}>
+          <input
+            ref={uploadInput}
+            type="file"
+            multiple
+            hidden
+            onChange={async event => {
+              const files = Array.from(event.target.files ?? []);
```
arjita-mitra: its best to extract the onChange handler to a named function using useCallback.

Inline functions are recreated on every render, which can break the referential stability or be re-rendered unnecessarily.

apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
```diff
@@ -241,6 +245,24 @@ export const ConversationCells = memo(
...
+              event.target.value = '';
+              if (!files.length) {
```
arjita-mitra: we are using fireAndForgetInvoker for async calls. with this we don't need to handle promise rejections and error handling it will take care of it. Sample code

```
const handleFileUpload = useCallback(
  async (event: React.ChangeEvent<HTMLInputElement>) => {
    const files = Array.from(event.target.files ?? []);
    event.target.value = '';
    if (!files.length) {
      return;
    }
    
    fireAndForgetInvoker.fireAndForget(async () => {
      await sharedDriveUploadController.upload(
        files,
        getCellsApiPath({conversationQualifiedId, currentPath: getCellsFilesPath()}),
        handleRefresh,
      );
    });
  },
  [fireAndForgetInvoker, sharedDriveUploadController, conversationQualifiedId, handleRefresh]
);
```

arjita-mitra (changes requested):

apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
cristianoliveira: ✅

apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
cristianoliveira: ✅

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
```diff
@@ -0,0 +1,104 @@
...
+      subscriptions.set(uploadId, result.value);
+    }
+  };
+
+  const run = async (uploadId: string, command: (id: string) => Promise<Result<void, unknown>>): Promise<void> => {
+    await command(uploadId);
+    notify();
+  };
+
+  const upload = async (files: readonly File[], path: string, onRefresh: () => void): Promise<void> => {
```
arjita-mitra: this function is handling too many tasks. how do you feel about breaking it down to reduce the cognitive load.

Keep the upload as the high-level workflow: upload files, refresh on success, notify listeners

```
// Added UploadCommandResult type alias
type UploadCommandResult = Result<void, unknown>;

const startAndPublish = async (uploadId: string): Promise<UploadCommandResult> => {
    const uploadResult = await manager.start(uploadId);
    return uploadResult.isErr ? uploadResult : manager.publish(uploadId);
  };

const uploadFile = async (file: File, path: string): Promise<UploadCommandResult> => {
    const uploadId = createUploadId();
    ids.push(uploadId);

    const registered = manager.register(uploadId, createSource(file), path);
    if (registered.isErr) {
      return Result.err(registered.error);
    }

    attach(uploadId);
    return startAndPublish(uploadId);
  };

const hasSuccessfulUpload = (results: readonly UploadCommandResult[]): boolean => results.some(result => result.isOk);

  const upload = async (files: readonly File[], path: string, onRefresh: () => void): Promise<void> => {
    const results = await Promise.all(files.map(file => uploadFile(file, path)));
    if (hasSuccessfulUpload(results)) {
      onRefresh();
    }
    notify();
  };


```

arjita-mitra (approved):

PR #22358, opened 2026-09-02
Disallow TypeScript non-null and definite assignment assertions [WPB-22420]

screendriver:
<table>
<td>
  </td></table>
  <br />
 

This pull request enables TypeScript safety policies for non-null and definite assignment assertions.

It:

- enables `@typescript-eslint/no-non-null-assertion` for TypeScript and TSX production and test sources;
- adds existing `no-restricted-syntax` restrictions for `PropertyDefinition[definite=true]` and `VariableDeclarator[definite=true]`;
- migrates existing assertion usages to explicit initialization, control flow, runtime checks, and lifecycle handling;
- applies both policies to production and test TypeScript and TSX sources.

The policies reject expression-level assertions such as:

    const value = possiblyUndefined!;

and declaration-level assertions such as:

    let value!: string;

    class Example {
      private dependency!: Dependency;
    }

Both assertion forms bypass TypeScript analysis and can hide invalid assumptions about initialization, external values, collection lookups, browser APIs, and object lifecycles.

### DIFF
--- a/apps/webapp/src/script/auth/module/action/authAction.ts
+++ b/apps/webapp/src/script/auth/module/action/authAction.ts
@@ -17,7 +17,7 @@
  *
  */
 
-import {isString} from '@sindresorhus/is';
+import {isString, isUndefined} from '@sindresorhus/is';
 import type {DomainData} from '@wireapp/api-client/lib/account/domainData';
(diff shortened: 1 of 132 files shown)


screendriver:
The failing End-to-End tests are unrelated to these changes and fail because of infrastructure issues.

zskhan (approved):

PR #22363, opened 2026-09-02
fix: recover after key-package exhaustion [WPB-27396]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27396" title="WPB-27396" target="_blank">BugWPB-27396</a>  [Web] Recover MLS conversations after key-package exhaustion
  </td></table>
  <br />
 

## Summary

Fixes **WPB-27396** by recovering MLS conversations that may have missed Welcome messages while the client had no available key packages.

## Changes

- Detect when the backend MLS key-package count reaches zero.
- Refill key packages and persist an account-scoped recovery marker in Core DB.
- Refresh the complete conversation list to discover conversations created during key-package exhaustion.
- Audit active MLS and mixed-protocol conversations in batches of 10.
- Rejoin locally missing MLS groups using the existing external-commit flow.
- Track pending conversation IDs so interrupted recovery resumes with the remaining conversations.
- Retry recovery after startup, reconnect, window focus, or returning to the foreground.
- Pause recovery while the application is inactive.
- Continue processing after individual conversation failures and retain recovery state for another attempt.
- Clear recovery state only after the entire audit completes successfully.

## Tests

Added coverage for:

- Zero key-package detection and successful refill.
- Normal low-count refills that should not trigger recovery.
- Upload and persistence failures.
- Recovery across application restarts.
- Pending-conversation progress tracking.
- MLS and mixed-protocol conversation recovery.
- Skipping Proteus, past-member, and already-established conversations.
- Pausing in the background and continuing after individual failures.


apps/webapp/src/script/mls/MLSConversations.ts
```diff
@@ -38,9 +39,132 @@ import {Conversation} from 'Repositories/entity/Conversation';
...
+}: {
+  conversations: Conversation[];
+  conversationRepository: ConversationRepository;
+  core: Account;
+  isActive: () => boolean;
+  mlsService?: MLSService;
+  batchSize?: number;
+}): Promise<MLSConversationRecoveryResult> {
+  const conversationService = core.service?.conversation;
+  if (!conversationService) {
```
zskhan, line 78:
```suggestion
  if (conversationService === undefined) {
```

apps/webapp/src/script/main/app.ts
```diff
@@ -808,6 +818,92 @@ export class App {
...
+    conversationRepository,
+    eventRepository,
+    fireAndForgetInvoker,
+  }: {
+    conversationRepository: ConversationRepository;
+    eventRepository: EventRepository;
+    fireAndForgetInvoker: FireAndForgetInvoker;
+  }): void {
+    const mlsService = this.core.service?.mls;
+    if (!mlsService) {
```
zskhan, line 831:
```suggestion
    if (mlsService === undefined) {
```

apps/webapp/src/script/mls/MLSConversations.ts
```diff
@@ -38,9 +39,132 @@ import {Conversation} from 'Repositories/entity/Conversation';
...
+          conversationId: conversation.qualifiedId,
+          error: recoveryResult.error,
+        });
+        continue;
+      }
+
+      recoveredConversationCount++;
+
+      // Remove successfully recovered conversation from pending list
+      if (mlsService) {
```
zskhan, line 147:
```suggestion
      if (mlsService !== undefined) {
```

zskhan (dismissed):

zskhan (approved):

PR #22367, opened 2026-09-03
feat: show precise Shared Drive upload progress [WPB-28302]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28302" title="WPB-28302" target="_blank">TaskWPB-28302</a>   [web] Add direct upload progress popup for Shared Drive (ui only)
  </td></table>
  <br />
 

## Problem
Indeterminate upload status does not show whether a large Shared Drive upload is advancing or stalled.

## Solution
Use lifecycle-owned Cells progress to render accessible determinate progress while preserving an indeterminate fallback when no progress has been reported. Clamp invalid values and prevent progress from moving backward within an upload attempt.

https://wearezeta.atlassian.net/browse/WPB-28302


cristianoliveira:
Closing cuz we'll cover this in a later step

PR #22368, opened 2026-09-03
feat: Adds Shared Drive drag-and-drop support for uploading a single file into the current folder(WPB-28291)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28291" title="WPB-28291" target="_blank">TaskWPB-28291</a>  [web] Add drag and drop support for single file upload in root
  </td></table>
  <br />
 

## Summary

1. Added Shared Drive-specific drop validation and upload dispatch.
2. Wired drag/drop into the Shared Drive tab without using the conversation attachment composer.
3. Added the Figma-style drag overlay with accent-aware border/background and theme-readable text.
4. Added useSharedDriveFileDrop to keep rejection handling and modal feedback out of conversationCells.tsx.
5. Blocks drops when upload permission is missing or the user is in the recycle bin.
6. Rejects empty, multiple, unsupported, and oversized drops with existing upload error UI.
7. Added deterministic unit tests for accepted root drop, nested folder target, oversized/rejected input, multiple file rejection, viewer restriction, and recycle-bin guard.


apps/webapp/src/script/components/conversation/conversationCells/useSharedDriveFileDrop.ts
```diff
@@ -0,0 +1,83 @@
...
+  onRefresh,
+  sharedDriveUploadController,
+  translate,
+  uploadPath,
+}: UseSharedDriveFileDropParameters): ((files: readonly File[]) => void) =>
+  useCallback(
+    (files: readonly File[]): void =>
+      handleSharedDriveDroppedFiles(files, {
+        fireAndForgetInvoker,
+        isAcceptedFile: () => true,
```
cristianoliveira: `isAcceptedFile` is always `true`, so the new drop path never rejects an unsupported file. With `FEATURE_ALLOWED_FILE_UPLOAD_EXTENSIONS=.pdf`, dropping `malware.exe` reaches the Cells uploader. Could we use `isAllowedFile(file.name, file.type)` here and cover the restrictive-config case?

cristianoliveira (changes requested):

apps/webapp/src/script/components/conversation/conversationCells/useSharedDriveFileDrop.ts
arjita-mitra: addressed 797562df6733c1ae602ce1a518aa70ad6783a7e1

cristianoliveira (dismissed):

cristianoliveira (approved):

PR #22401, opened 2026-09-07
feat: Add drag and drop into a specific folder with direct upload mode(WPB-28304)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28304" title="WPB-28304" target="_blank">TaskWPB-28304</a>  [web] Add drag and drop into a specific folder 
  </td></table>
  <br />
 

## Summary

- Add direct Cells upload support through the API client and `CellsRepository`
- Wire Shared Drive uploads to direct mode, including success, failure, cancel, and retry handling
- Add drag-and-drop upload support for the Shared Drive tab with file type, size, and recycle bin validation
- Support dropping files directly onto folder rows and route uploads to the selected folder path
- Add viewer permission check
- Add upload status popup wiring and focused coverage for direct uploads, drop validation, folder drop targets, and upload status behavior


apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
```diff
@@ -39,63 +40,205 @@ export type SharedDriveUploadController = {
...
-    }
+  const setState = (uploadId: string, state: UploadState): void => {
+    statesByUploadId.set(uploadId, state);
+    notify();
   };
 
-  const run = async (uploadId: string, command: (id: string) => Promise<Result<void, unknown>>): Promise<void> => {
-    await command(uploadId);
-    notify();
+  const createDraftUploadStrategy = ({manager}: DraftUploadDependencies): UploadStrategy => {
```
cristianoliveira: I'd make this a proper deterministic function and avoid depending on the closure. That makes things easy to follow and test

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
```diff
@@ -39,63 +40,205 @@ export type SharedDriveUploadController = {
...
+      snapshot: uploadId => manager.snapshot(uploadId).unwrapOr(undefined),
+      cancel: uploadId => command(uploadId, manager.cancel),
+      retryUpload: uploadId => command(uploadId, manager.retryUpload),
+      retryPublish: uploadId => command(uploadId, manager.retryPublish),
+      discard: uploadId => command(uploadId, manager.discard),
+      retryDiscard: uploadId => command(uploadId, manager.retryDiscard),
+    };
+  };
+
+  const createDirectUploadStrategy = ({
```
cristianoliveira: Same here

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.test.ts
```diff
@@ -17,43 +17,26 @@
...
-  return manager;
 }
 
-function createController(manager: ReturnType<typeof createManager>) {
+function createController(cellsRepository = createCellsRepository()) {
   const controller = createSharedDriveUploadController({
-    manager: manager as unknown as CellsUploadManager,
+    mode: 'direct',
+    cellsRepository,
+    createAbortController: () => new AbortController(),
```
cristianoliveira: Are we coupling the controller with the manager? Why?

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.test.ts
```diff
@@ -62,13 +45,51 @@ function createController(manager: ReturnType<typeof createManager>) {
...
+    discard: jest.fn().mockResolvedValue(Result.ok(undefined)),
+    retryDiscard: jest.fn().mockResolvedValue(Result.ok(undefined)),
+    release: jest.fn().mockReturnValue(Result.ok(undefined)),
+  };
+}
+
+function createDraftController(manager = createDraftManager()) {
+  const controller = createSharedDriveUploadController({
+    mode: 'draft',
+    manager,
```
cristianoliveira: For each on the `mode` you need different dependencies, that's a abstraction leak. You shouldn't need to know what dependency to use for each mode

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
```diff
@@ -39,63 +40,205 @@ export type SharedDriveUploadController = {
...
+      retryDiscard: async () => undefined,
+    };
   };
 
-  const registerFile = (file: File, path: string, conversationQualifiedId: string) => {
+  // Draft mode is not wired into production yet; it stays here as the migration path for staged Cells uploads.
+  const uploadStrategy =
+    mode === 'draft'
+      ? createDraftUploadStrategy(dependencies as DraftUploadDependencies)
+      : createDirectUploadStrategy(dependencies as DirectUploadDependencies);
```
cristianoliveira: Here we can see the drift in the dependency injection

apps/webapp/src/script/components/conversation/conversationCells/cellsTable/cellsTable.tsx
```diff
@@ -120,9 +143,78 @@ export const CellsTable = ({
   const rows = table.getRowModel().rows;
   const tableWrapperStyles = rows.length > 0 ? [wrapperStyles, wrapperWithRowsStyles] : wrapperStyles;
 
+  useEffect(() => {
+    return () => onFolderDropTargetChange?.(null);
+  }, [onFolderDropTargetChange]);
+
+  const setActiveFolderDropTarget = (node: CellNode | null): void => {
```
cristianoliveira, line 157: I think we have two drag states here. `SharedDriveDropzone` resets on window `drop`, `dragend` and `blur`, but this folder state only resets on the row events. If the drag ends outside, this overlay can stay active. Could we have one reset path for both and cover that with the components together?

libraries/api-client/src/cells/cellsApi.ts
```diff
@@ -207,6 +207,48 @@ export class CellsAPI {
     return result.data;
   }
 
+  async uploadNode({
```
cristianoliveira, line 210: I think we’re missing a test at this layer. The controller mocks `CellsRepository.uploadNode`, and the repository test mocks this method, so nothing run this new code. Could we add coverage here for `createCheck`, auto rename, abort signal and the storage upload?

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
```diff
@@ -39,63 +40,205 @@ export type SharedDriveUploadController = {
...
+          source,
+          error: {kind: 'uploadFailed', cause: error},
+        });
+        return false;
+      } finally {
+        abortControllersByUploadId.delete(uploadId);
+      }
+    };
+
+    const retryUpload = async (uploadId: string): Promise<void> => {
```
cristianoliveira: I can’t find a caller for this. The popup only calls `cancel`, so the retry test proves the method works but users can’t reach it (and a successful retry won’t refresh the table). Are we wiring retry in this PR?

apps/webapp/src/script/components/conversation/conversationCells/cellsTable/cellsTable.tsx
```diff
@@ -120,9 +143,78 @@ export const CellsTable = ({
...
+      },
+    };
+  };
+
   return (
     <CellsFilePreviewModalProvider>
       <div css={tableWrapperStyles}>
+        {activeFolderDropTargetName !== null && (
+          <div css={folderDropOverlayStyles} role="status">
+            <UploadIcon width={24} height={24} />
```
cristianoliveira: This icon is only decorative, can we add `aria-hidden="true"`? Also the folder name below can be truncated, so it should expose the full value with `title`

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
cristianoliveira: Would we need a follow up task/PR for the retry?

cristianoliveira (commented):
Left a few comments

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
cristianoliveira: Nevermind, there is this task https://wearezeta.atlassian.net/browse/WPB-28305

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
arjita-mitra: Addressed 89219b269194929399d9a9fbb87f39a4692806dd

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.test.ts
arjita-mitra: Addressed 89219b269194929399d9a9fbb87f39a4692806dd

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.ts
arjita-mitra: Addressed 89219b269194929399d9a9fbb87f39a4692806dd

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadController.test.ts
arjita-mitra: the controller is now decoupled. The mode specific dependencies are isolated inside the strategy factories https://github.com/wireapp/wire-webapp/commit/89219b269194929399d9a9fbb87f39a4692806dd

apps/webapp/src/script/components/conversation/conversationCells/cellsTable/cellsTable.tsx
arjita-mitra: ahh did resolve the issue for root SharedDriveDropzone drop but folder drop has this potential issue. Good catch.

arjita-mitra: Resolved 9812d3dec4302c9d798248f711f159a2ed5e60f6

libraries/api-client/src/cells/cellsApi.ts
arjita-mitra: added 8a2e960d95b8db9439f34a9fc2d81ad3c121310b

apps/webapp/src/script/components/conversation/conversationCells/cellsTable/cellsTable.tsx
arjita-mitra: correct. added 71fa64e6fe2b40cd1d07d7f9a5f8f66449d94bc7

cristianoliveira (approved):

PR #22429, opened 2026-09-08
Show join action in meeting notifications [WPB-28572]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28572" title="WPB-28572" target="_blank">TaskWPB-28572</a>  [Web] Show Join button in Meet now in-app notification
  </td></table>
  <br />
 

Add the in-app Join action for ongoing meeting notifications, including loading feedback and dismissal after the call becomes active.

### DIFF
--- a/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
+++ b/apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
@@ -17,12 +17,15 @@
  *
  */
 
+import {useEffect} from 'react';
+
 import type {QualifiedId} from '@wireapp/api-client/lib/user';
 import {match, P} from 'ts-pattern';
 import {container} from 'tsyringe';
 
 import {Button, ButtonVariant, CalendarIcon} from '@wireapp/react-ui-kit';
 
+import {useJoinMeetingCall} from 'Components/meeting/useJoinMeetingCall';
 import {UserState} from 'Repositories/user/userState';
 import {useApplicationContext} from 'src/script/page/rootProvider';
 import type {Translate, TranslationKey} from 'Util/localizerUtil';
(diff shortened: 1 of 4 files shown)


apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -128,6 +171,38 @@ const MeetingNotificationMetadata = ({
...
+        variant={ButtonVariant.PRIMARY}
+        css={meetingNotificationCardActionStyles}
+        type="button"
+        onClick={() => {
+          onDismiss();
+          navigate('/meetings');
+        }}
+        aria-label={translate('meetings.notifications.view')}
+      >
+        <CalendarIcon css={meetingNotificationViewBtnStyles} aria-hidden="true" />{' '}
```
arjita-mitra: Instead of {' '}, It's better to use className with gap if parent is flex

Another option is to use marginRight on the CalendarIcon

apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
```diff
@@ -44,6 +47,46 @@ import {type MeetingNotification, MeetingNotificationKind} from '../meetingNotif
...
+  }, [isCallActive, onCallJoined, onDismiss]);
+
+  return (
+    <Button
+      variant={ButtonVariant.PRIMARY}
+      css={meetingNotificationCardActionStyles}
+      type="button"
+      onClick={() => {
+        joinMeeting();
+      }}
```
e-maad: Is this valid in particular case?
```suggestion
      onClick={joinMeeting}
```

e-maad (dismissed):
LGTM 👍

apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
zskhan: 0b22085c3179a7bee65a057b982cfe877dde7828

apps/webapp/src/script/components/meeting/meetingNotificationCard/meetingNotificationCard.tsx
zskhan: 0b22085c3179a7bee65a057b982cfe877dde7828

e-maad (approved):

PR #22434, opened 2026-09-09
feat: Add shared drive upload status indicator to files tab(WPB-28319)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28319" title="WPB-28319" target="_blank">TaskWPB-28319</a>  [web] Shared drive tab icon status indication
  </td></table>
  <br />
 

Show upload progress state in the conversation Files tab by subscribing to the existing shared drive upload controller. Reuse the same latest-upload status resolution as the popup, render a spinner while uploads are active, and show the completion icon for both successful and failed uploads per the acceptance criteria.

## Summary

- Adds a Shared Drive upload status indicator to the conversation Files tab.
- Shows a spinner while an upload is in progress.
- Shows the completion icon after upload completion.
- Uses the same completion icon for failed uploads, matching the acceptance criteria.
- Extracts latest shared drive upload status lookup for reuse by both the popup host and the tab.
- Adds Shared Drive upload spinner/completed icons to `react-ui-kit`.

Screenshot 2026-09-09 at 13 47 41
Screenshot 2026-09-09 at 13 48 01

### DIFF
--- a/apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
+++ b/apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
@@ -1,0 +1,150 @@
+/*
+ * Wire
+ * Copyright (C) 2026 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {act, render} from '@testing-library/react';
+import {ThemeProvider} from '@wireapp/react-ui-kit';
+
+import type {UploadState} from 'Repositories/cells/upload';
+import {
+  createRootContextValueForTest,
+  createRootProviderWrapperForTest,
+} from 'src/script/page/testSupport/rootContextTestSupport';
+import {translateForTest} from 'Util/test/translateForTest';
+
+import type {SharedDriveUploadController} from '../conversationCells/sharedDriveUploadController';
+
+import {ConversationTabs} from './conversationTabs';
+
+const conversationQualifiedId = {id: 'conversation', domain: 'example.com'};
+const conversationQualifiedIdString = 'conversation@example.com';
+const uploadSource = {blob: new Blob(['data']), name: 'report.pdf', contentType: 'application/pdf', size: 4};
+const uploadingState: UploadState = {
+  kind: 'uploading',
+  identity: {uploadId: 'upload-1'},
+  source: uploadSource,
+  progress: 0,
+};
+const uploadedState: UploadState = {
+  kind: 'published',
+  identity: {uploadId: 'upload-1', resourceUuid: 'resource-1', versionId: 'version-1'},
+  source: uploadSource,
+};
+const failedState: UploadState = {
+  kind: 'uploadFailed',
+  identity: {uploadId: 'upload-1'},
+  source: uploadSource,
+  error: {kind: 'uploadFailed', cause: new Error('upload failed')},
+};
+
+type TestController = SharedDriveUploadController & {
+  snapshots: jest.MockedFunction<SharedDriveUploadController['snapshots']>;
+  subscribe: jest.MockedFunction<SharedDriveUploadController['subscribe']>;
+};
+
+const createController = (state: UploadState | null = null) => {
+  let currentState = state;
+  let notify = () => undefined;
+  const controller: TestController = {
+    snapshots: jest.fn(scope => (scope === conversationQualifiedIdString && currentState ? [currentState] : [])),
+    subscribe: jest.fn(listener => {
+      notify = listener;
+      return jest.fn();
+    }),
+    upload: jest.fn(),
+    cancel: jest.fn(),
+    retryUpload: jest.fn(),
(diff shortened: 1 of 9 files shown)


apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+};
+
+type TestController = SharedDriveUploadController & {
+  snapshots: jest.MockedFunction<SharedDriveUploadController['snapshots']>;
+  subscribe: jest.MockedFunction<SharedDriveUploadController['subscribe']>;
+};
+
+const createController = (state: UploadState | null = null) => {
+  let currentState = state;
+  let notify = () => undefined;
```
cristianoliveira: I think `notify` needs an explicit `() => void` type here. Right now TS infers `() => undefined`, so assigning `listener` below makes the PR type check fail

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.tsx
```diff
@@ -127,7 +159,36 @@ const ConversationTab = ({id, label, isActive, onClick, onKeyDown}: Conversation
       onKeyDown={onKeyDown}
       className="conversation-tabs__button"
     >
-      {label}
+      <span className="conversation-tabs__button-content">
+        {label}
+        {uploadStatusKind && <SharedDriveTabUploadStatusIcon kind={uploadStatusKind} />}
```
cristianoliveira: Could we expose this status to screen readers too? Right now both icons are `aria-hidden`, so the tab still only announces “Files” while the upload status changes. Maybe translated visually hidden text with `role="status"` would work here

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.tsx
```diff
@@ -17,27 +17,51 @@
...
   onIndexChange: (index: number) => void;
   conversationQualifiedId: QualifiedId;
+  sharedDriveUploadController: SharedDriveUploadController;
+  isUploadStatusIndicatorEnabled: boolean;
 }
 
 const FILE_PATH = 'files';
 
-export const ConversationTabs = ({activeTabIndex, onIndexChange, conversationQualifiedId}: ConversationTabsProps) => {
+const toConversationQualifiedIdString = ({id, domain}: QualifiedId): string => `${id}@${domain}`;
```
cristianoliveira: Can we use `stringifyQualifiedId` from `qualifiedIdUtil` here? This looks like the same conversion, and the existing helper also preserves the `StringifiedQualifiedId` type

libraries/react-ui-kit/src/dataDisplay/icon/sharedDriveUploadSpinnerIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {SVGIcon, SVGIconProps} from './svgIcon';
+
+const DEFAULT_COLOR = '#0667C8';
```
cristianoliveira: Could we use the existing color tokens here instead of raw hex values? Same for the completed icon. It avoids adding another source for these colors and makes the theme behavior explicit

apps/webapp/src/style/content/conversation.less
```diff
@@ -133,6 +139,22 @@
...
     }
   }
+
+  &__upload-status-icon {
+    width: 16px;
+    height: 16px;
+    flex: 0 0 auto;
+
+    &--uploading {
+      animation: conversation-tab-upload-spinner 1s linear infinite;
```
cristianoliveira, line 149: Can we disable this animation under `prefers-reduced-motion: reduce`? The popup progress animation already does this, so I think the same behavior makes sense here

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+        conversationQualifiedId={conversationQualifiedId}
+        sharedDriveUploadController={controller}
+        isUploadStatusIndicatorEnabled={isUploadStatusIndicatorEnabled}
+      />
+    </ThemeProvider>,
+    {wrapper: createRootProviderWrapperForTest(createRootContextValueForTest({translate: translateForTest}))},
+  );
+
+describe('ConversationTabs', () => {
+  beforeEach(() => {
```
cristianoliveira: Can we avoid changing `window.location` in every upload status test? These tests don’t care about routing, and the global hash state can leak between tests. Maybe we can keep the upload tests independent from the routing state

cristianoliveira (commented):
Left a few comments

apps/webapp/src/style/content/conversation.less
cristianoliveira: On a second thought, I think this might not be an issue for such a small animation

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.tsx
arjita-mitra: makes sense 301364631ef7fc74d30395e27d993c07f522e32a

libraries/react-ui-kit/src/dataDisplay/icon/sharedDriveUploadSpinnerIcon.tsx
arjita-mitra: Absolutely and theme aware 301364631ef7fc74d30395e27d993c07f522e32a

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
arjita-mitra: makes sense 301364631ef7fc74d30395e27d993c07f522e32a

cristianoliveira (approved):

PR #22435, opened 2026-09-09
Use 24-hour formatting for meeting times across all views [WPB-28510]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28510" title="WPB-28510" target="_blank">TaskWPB-28510</a>  [Web] Use 24-hour format for meeting date and time
  </td></table>
  <br />
 

Make meeting time ranges and scheduling time pickers respect the user’s regional locale.

  - Support localized 12/24-hour formats.
  - Preserve AM/PM parsing compatibility.
  - Add regression tests for regional locales and legacy time values.
  
**German**
image

**English**
image

### DIFF
--- a/libraries/react-ui-kit/src/inputs/timePickerField/timePickerUtils.ts
+++ b/libraries/react-ui-kit/src/inputs/timePickerField/timePickerUtils.ts
@@ -22,58 +22,78 @@ import type {Option} from '../select';
 export const TIME_INTERVAL_MINUTES = 15;
 export const TIME_OPTIONS_COUNT = 96;
 
+const HOURS_PER_HALF_DAY = 12;
+const MINUTES_PER_HOUR = 60;
+const TIME_LABEL_DIGITS = 2;
+const TIME_LABEL_REFERENCE_YEAR = 2026;
+
+const timeFormatterOptions: Intl.DateTimeFormatOptions = {
+  hour: 'numeric',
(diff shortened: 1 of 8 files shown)


libraries/react-ui-kit/src/inputs/timePickerField/timePickerUtils.ts
```diff
@@ -23,24 +23,21 @@ export const TIME_INTERVAL_MINUTES = 15;
 export const TIME_OPTIONS_COUNT = 96;
 
 export const parseTimeLabel = (value: string | number): {hour24: number; minutes: number} => {
```
arjita-mitra, line 38: This change can cause regression. parseTimeLabel no longer handles "AM"/"PM" inputs and can mis parse strings like "3:00 PM".

this is a shared parsing utility, and it is already consumed outside the time picker itself via combineDateAndTime in [dateTimeUtils.ts](/Users/arjitamitra/Documents/wire-webapp/libraries/react-ui-kit/src/inputs/dateTimePickerField/dateTimeUtils.ts)

Suggested fix:
Restore robust parsing that accepts both "HH:mm" and "h:mm AM/PM" forms and cover it with tests.”

arjita-mitra (changes requested):

arjita-mitra (approved):

PR #22453, opened 2026-09-11
feat: add a close/dismiss button once all progress is idle(WPB-28486)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28486" title="WPB-28486" target="_blank">TaskWPB-28486</a>  [web] add a close/dismiss button once all progress is idle
  </td></table>
  <br />
 

Added a dismiss action for the shared drive upload status UI once there is nothing actionable left.

- Shows a header Close button in the same location/style as Cancel only when all current upload statuses for the conversation are successful.
- Keeps Cancel for active uploads and does not show Close for failed uploads because Retry is still actionable.
- Adds an accessible label: Close upload status.
- Adds test coverage for successful dismissal, hiding Close while uploads are actionable, and the mixed-state case where one upload succeeded but another is still actionable.

Screenshot 2026-09-11 at 13 59 35
Screenshot 2026-09-11 at 13 50 06

### DIFF
--- a/apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
+++ b/apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
@@ -64,6 +64,8 @@ interface SharedDriveUploadStatusPopupProps {
   readonly toggleLabel: string;
   readonly cancelLabel: string;
   readonly dismissLabel?: string;
+  readonly dismissAriaLabel?: string;
+  readonly canDismiss?: boolean;
   readonly retryLabel: string;
   readonly isCancelling: boolean;
   readonly isRetrying: boolean;
@@ -116,6 +118,8 @@ export const SharedDriveUploadStatusPopup = ({
   toggleLabel,
   cancelLabel,
   dismissLabel = cancelLabel,
+  dismissAriaLabel = dismissLabel,
+  canDismiss = upload.kind === 'uploaded',
   retryLabel,
   isCancelling,
   isRetrying,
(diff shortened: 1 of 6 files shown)


apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
```diff
@@ -194,6 +198,17 @@ export const SharedDriveUploadStatusPopup = ({
               {cancelLabel}
             </button>
           )}
+          {canDismiss && (
+            <button
+              type="button"
+              css={sharedDriveUploadStatusPopupHeaderCancelStyles}
+              aria-label={dismissAriaLabel}
```
cristianoliveira, line 205: Since this button appears after an event we need to add `aria-live` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions#live_regions so it is announced

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
cristianoliveira, line 189: Adding `aria-live` here, should do the trick

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
arjita-mitra: The button itself should not get aria-live. Live regions are for announcing changed text/content in a region, not used to make a newly inserted control discoverable.

We already have `<div css={sharedDriveUploadStatusPopupTextStyles} role="status" aria-live="polite">`

so when the upload changes from "Uploading" to "Uploaded" that status/title change should be announced. After that the user can discover the new Close button by normal keyboard/screen reader navigation.

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadStatusPopup.tsx
arjita-mitra: putting aria-live on a button group can lead to noisier announcements

cristianoliveira (approved):

PR #22466, opened 2026-09-14
Render PrimaryModal translations with React [WPB-27666]

screendriver:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27666" title="WPB-27666" target="_blank">SecurityWPB-27666</a>  [Web] Text rendering
  </td></table>
  <br />
 

This pull request continues the localized rich-text rendering migration by moving translation-driven `PrimaryModal` content to React-based rendering behind the existing startup feature toggle:

`react-translation-rendering`

### Changes

This batch introduces an explicit React rendering path for translated PrimaryModal messages and migrates the existing translation-driven `htmlMessage` producers to it.

Translation-authored structure such as links, line breaks, and formatting is rendered as explicit React elements, while runtime values such as user names, backend names, URLs, counts and other application data remain ordinary React text.


apps/webapp/src/script/components/meeting/shared/submit/showMeetingPartialAddFailureModal.tsx
```diff
@@ -137,8 +290,12 @@ export const formatMeetingPartialAddFailureMessage = (
 
   const summary = translate('failedToAddParticipantsPlural', {total: allUserIds.length.toString()});
   const details = failedToAdd
-    .map(failure => formatFailureDetails(failure, users, translate))
-    .filter(detail => detail.length > 0);
+    .map(failure => {
+      return formatFailureDetails(failure, users, translate);
```
ikotarac, line 294: why is this expanded in full blocks?

apps/webapp/src/script/components/meeting/shared/submit/showMeetingPartialAddFailureModal.tsx
```diff
@@ -137,8 +290,12 @@ export const formatMeetingPartialAddFailureMessage = (
...
+    })
+    .filter(detail => {
```
ikotarac, line 296: also, why not 
```ts
.filter(isNonEmptyString)
```

ikotarac:
A few things confuse me with this approach.
- what are we getting with a fully custom implementation?
- why not make the translation to component mapping less manual?

I get it that we used non-standard syntax `[tag]...[/tag]`, and we might not want to now change all translation keys and trigger Crowdin work, but I have a few ideas how maybe this can be avoided.

Current implementation would in use look something like this:

**en-US.json**
```json
{
"federationDeleteWithDetails": "[bold]Your backend[/bold] stopped federating with [bold]{backendUrl}.[/bold][br]Contact your admin for details."
}
```

**SomeComponent.tsx**
```tsx
const backendUrlMarker = createReactTranslationMarker('federation-stop-backend-url');
const brMarker = createReactTranslationMarker('federation-stop-br');

const text = translate(
  'federationDeleteWithDetails',
  {backendUrl: backendUrlMarker.substitution},
  {br: brMarker.substitution}, // redirect [br] away from translate.ts's default '<br>' HTML substitution
);

renderReactTranslation({
  translatedText: text,
  componentReplacements: [
    {start: '<strong>', end: '</strong>', render: children => <strong>{children}</strong>},
  ],
  valueReplacements: [{marker: backendUrlMarker, runtimeText: domains[0]}],
  nodeReplacements: [{marker: brMarker, render: () => <br />}],
});
```

To me this is overly verbose and these steps can probably be wrapped into a reusable function, for example:

```tsx
renderReactTranslationDeluxe(
  'federationDeleteWithDetails',
  {
    strong: children => <strong>{children}</strong>,
    br: <br />,
    backendUrl: domains[0],
  },
  translate,
);
```

**renderReactTranslationDeluxe** would iterate over the map of substitutions, and if the value is a function, use it as component substitution with children, otherwise everything else can be treated as node substitution. And for every key in the map of substitutions this function can create a translation marker for us under the hood. I don't want to paste implementation here for brevity.

But then we essentially reinvent the API of the react-intl library and their **FormattedMessage** component:

```tsx
import {FormattedMessage} from 'react-intl';

<FormattedMessage
  id="federationDeleteWithDetails"
  values={{
    b: (chunks: React.ReactNode[]) => <strong>{chunks}</strong>,
    backendUrl: domains[0],
    br: <br />,
  }}
/>
```

The only problem is that **react-intl** expect something like this:
```json
"federationDeleteWithDetails": "<b>Your backend</b> stopped federating with <b>{backendUrl}.</b>{br}Contact your admin for details."
```

So my suggestion is: why don't we adapt our current json files, to follow this format by doing one-off transformation on load:
```ts
function bracketsToICUFormat(text: string): string => text.replace(/\[(\/?)(\w+)\]/g, (match, slash: string, name: string) => {
    const isClosingTag = slash === '/';
    const hasCounterpart = isClosingTag || text.includes(`[/${name}]`);

    if (!hasCounterpart) {
      // no [/name] anywhere in this string -> void token, not a tag -> plain ICU argument
      return `{${name}}`;
    }

    return `<${slash}${name}>`; // paired tag -> ICU/FormatJS rich-text tag
});
```


**bonus points:** we already use **react-intl** and **FormattedMessage** in auth flow, so this is at the same time closing this discrepancy.
**bonus points:** crowdin already understands `<customTags></customTags>` and renders them natively as UI elements that enhances translation experience and prevents errors with typos in `[ourOwnFormat][/ourOwnFormat]`
**bonus points:** we get a rich and well tested ecosystem that handles genders, complex plurals, date and time etc. out of the box.


Then we can gradually switch to using tags, and change gradually the old keys to all use standard tags and remove the need for this rexeg mapping step altogether.

<hr />

This is extra if anyone is still reading :D , from this point on begins the Fantasy Land 🦄  :D

What I actually really like, but is probably a too big change for now, is directly sourcing translation content from the components where it's used. This is possible with another library: **react-i18n**, and their **Trans** component, which eliminates maintaining even the **en-US.json** in a separate place. You just use it as a text component, and then you extract these templates as translation entries, and this is then plugged into Crowdin.

```tsx
import {Trans} from 'react-i18next';

<Trans i18nKey="federationDeleteWithDetails">
  <strong>Your backend</strong> stopped federating with <strong>{{backendUrl: domains[0]}}.</strong>
  <br />
  Contact your admin for details.
</Trans>
```

In build time json files are generated in this shape:

```json
"federationDeleteWithDetails": "<0>Your backend</0> stopped federating with <2>{{backendUrl}}.</2><3/>Contact your admin for details."
```

However what I really, really, really like is something like this, and I have actually built a library like this 7 years ago in React Native, so I know that trying to reinvent the translation library is very hard. Now if I have to do it again, I would probably do a wrapper around **react-i18n**.

**src/modules/federation/SomeComponent.tsx**
```tsx
import {TransDeluxe} from 'wire/translations';

<TransDeluxe>
  <strong>Your backend</strong> stopped federating with <strong>{domains[0]}.</strong>
  <br />
  Contact your admin for details.
</TransDeluxe>
```

Then in build time we can generate keys based on the filename, and we can create node slots based on standard JSX interpolations:

```json
"modules.federation.SomeComponent.0": "<0>Your backend</0> stopped federating with <2>{{0}}.</2><3/>Contact your admin for details."
```

screendriver:
I agree that `FormattedMessage` / a more standardized rich-text API is worth exploring separately.

For this work I **intentionally** want to keep the scope much narrower. The goal is not to redesign our translation infrastructure, translation format or Crowdin integration, but specifically to remove translated content from `dangerouslySetInnerHTML`.

The explicit mapping is somewhat verbose but it also keeps the change very controlled: only explicitly supported translation structure becomes React elements, while runtime values stay opaque text. We can do that without changing existing translation keys or introducing a second migration at the same time.

Once this security-related migration is finished I think evaluating whether we should consolidate on `react-intl` would make sense as a separate architectural change.

apps/webapp/src/script/components/meeting/shared/submit/showMeetingPartialAddFailureModal.tsx
screendriver: This is intentional. I generally prefer block-bodied arrow functions with an explicit `return` because it makes returning a value a deliberate decision rather than a side effect of the last expression.

Implicit returns can also cause subtle bugs when the expression itself returns something. For example, `.then(value => cache.set(key, value))` forwards the `Map` returned by `set()` into the next promise step, and `.filter(value => selected.push(value))` uses the numeric return value of `push()` as the predicate.

With a block body the control flow and return value are explicit. I generally prefer that clarity over saving a couple of lines.

ikotarac:
> I agree that `FormattedMessage` / a more standardized rich-text API is worth exploring separately.
> 
> For this work I **intentionally** want to keep the scope much narrower. The goal is not to redesign our translation infrastructure, translation format or Crowdin integration, but specifically to remove translated content from `dangerouslySetInnerHTML`.
> 

I don't think you understood what I wrote - I didn't say that we need to change infrastructure or translation format or Crowdin integration right now. My suggestion was to use the same exact translation files we have right now, and just to adapt them to what library (that we already use in the codebase) well understands, using a simple less than 10 line regex replacement function. Then we pass this transformed translations object to the provider that's connected with **FormattedMessage** (we also already do this), and this can be used alongside existing `translate(...)` functions.

If you take a look at current **en-US.json**, you can se that we use both `<learnMore></learnMore>` and `[faqLink][/faqLink]` for example. And the first one already works. So my suggestion was to use regex to transform `[faqLink][/faqLink]` and all similar into `<faqLink></faqLink>` only in runtime in this phase, and then later to clean it up also in Crowdin when we have time.

> The explicit mapping is somewhat verbose but it also keeps the change very controlled: only explicitly supported translation structure becomes React elements, while runtime values stay opaque text. We can do that without changing existing translation keys or introducing a second migration at the same time.

I believe the same guarantees are provided by **FormattedMessage** component, and we could also just use it initially only in places where rich text rendering is expected (cca 25 occurrences), not to change every single translation function.

> 
> Once this security-related migration is finished I think evaluating whether we should consolidate on `react-intl` would make sense as a separate architectural change.

Since this is merged and working now, yes I think we should consider this after :)

PR #22467, opened 2026-09-14
feat: Support multi-file uploads from shared drive file menu(WPB-28296)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28296" title="WPB-28296" target="_blank">TaskWPB-28296</a>  [web] Add multi-file direct upload from Shared Drive picker
  </td></table>
  <br />
 

- Support multi-file uploads from shared drive file menu
- Allow the shared drive file menu input to accept and validate multiple
- selected files before dispatching them as one upload batch. Keep the existing controller upload flow, which starts each selected file immediately without adding a custom concurrency queue.
- Add focused coverage for multi-file file-menu selection and batch upload startup behavior.


apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
```diff
@@ -304,7 +327,7 @@ export const ConversationCells = memo(
           onDropFiles={handleDroppedFiles}
         >
           <div css={wrapperStyles}>
-            <input ref={uploadInput} type="file" hidden onChange={handleUploadFiles} />
+            <input ref={uploadInput} type="file" hidden multiple onChange={handleUploadFiles} />
```
cristianoliveira: Could we test this using the actual input?

Right now the tests pass the array directly to the helper, so if someone removes `multiple` here all tests still pass, but browser only gives us first file

Something like:

```tsx
it('uploads all files selected from the file picker as one batch', async () => {
  const user = userEvent.setup()
  const firstFile = new File(['first'], 'first.txt', {type: 'text/plain'})
  const secondFile = new File(['second'], 'second.txt', {type: 'text/plain'})
  const sharedDriveUploadController = createSharedDriveUploadControllerMock()

  const {container} = renderConversationCells({
    sharedDriveUploadController,
  })

  const input = container.querySelector('input[type="file"]')
  expect(input).toBeInstanceOf(HTMLInputElement)

  await user.upload(input as HTMLInputElement, [firstFile, secondFile])

  const uploadAction = getUploadAction()
  await uploadAction()

  expect(sharedDriveUploadController.upload).toHaveBeenCalledWith(
    [firstFile, secondFile],
    expect.any(String),
    expect.any(Function),
    expect.any(String),
  )
})
```

`userEvent.upload` respects `multiple`, so this would catch that regression

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadInput.ts
```diff
@@ -29,6 +32,48 @@ type SharedDriveUploadInputDependencies = {
...
   readonly conversationQualifiedId: string;
   readonly onRefresh: () => void;
+  readonly onReject: (rejection: SharedDriveDropRejection) => void;
+  readonly isUploadFilesEnabled: boolean;
+  readonly isInRecycleBin: boolean;
+  readonly maxFileSize: number;
+  readonly isAcceptedFile: (file: File) => boolean;
+};
+
+const validateSharedDriveUploadInputFiles = (
```
cristianoliveira, line 42: I think we could reuse same validation from `sharedDriveDrop` here

Right now permission, recycle bin, empty, type and size checks are duplicated, so these two upload paths can drift. Maybe validator could receive explicit single/multiple file policy instead?

What do you think?

arjita-mitra: yes thats the plan. to reuse it in sharedrivedrop but drag and drop is not part of this PR so i didn't replace it. I’d probably rename validateSharedDriveUploadFiles and extract it to sharedDriveUploadValidation.ts

apps/webapp/src/script/components/conversation/conversationCells/conversationCells.tsx
arjita-mitra: addressed d75b9b69c9752e1b0e785d0e3fd059435de6fcd5

cristianoliveira (approved):

PR #22470, opened 2026-09-14
feat: support multiple Shared Drive upload feedback [WPB-28470]

cristianoliveira:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28470" title="WPB-28470" target="_blank">TaskWPB-28470</a>  [web] Multiple files upload feedback (popup mocked)
  </td></table>
  <br />
 

## Problem

Shared Drive uploads currently handle only one selected file at the input boundary, and the status popup exposes only the latest upload. Users cannot see queued, active, completed, and failed files together or understand aggregate progress when uploading multiple files.

## Solution

- Accept multiple files from the file selector and Shared Drive drop path, validating the complete batch before dispatch.
- Register every file immediately and schedule uploads through a controller-wide FIFO queue capped at three active uploads.
- Preserve input order and conversation-scoped status across tab changes and remounts.
- Render ordered multi-file status rows with distinct queued, uploading, uploaded, and failed states.
- Add aggregate header status, Cancel all, per-row cancel/retry/dismiss actions, and the approved Figma layout with one aggregate expanded progress bar.
- Keep status announcements, decorative icon accessibility, and light/dark design-token styling.

## Demo

https://github.com/user-attachments/assets/b3e1bea9-85d1-4039-953b-a2d6fb3f1b74

## References

- Jira: https://wearezeta.atlassian.net/browse/WPB-28470
- Figma: https://www.figma.com/design/grnVU2vAihHXwYgHryu2xE/Drive?node-id=22149-689815&m=dev


cristianoliveira:
Follow up PR to implement the upload queue will be in https://wearezeta.atlassian.net/browse/WPB-28627

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
```diff
@@ -113,17 +113,32 @@ describe('handleSharedDriveDroppedFiles', () => {
...
+    await uploadAction();
+    expect(dependencies.sharedDriveUploadController.upload).toHaveBeenCalledWith(
+      [firstFile, secondFile],
+      rootUploadPath,
+      dependencies.onRefresh,
+      conversationQualifiedId,
+    );
+  });
+
+  it('rejects the whole batch when any dropped file is invalid', () => {
```
arjita-mitra, line 132: does that mean uploaded files will. be rejected too or before starting to upload you are checking the validity of all files?

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
```diff
@@ -149,11 +164,14 @@ describe('handleSharedDriveDroppedFiles', () => {
 });
 
 describe('validateSharedDriveDroppedFiles', () => {
-  it('accepts exactly one valid file', () => {
-    const file = new File(['content'], 'document.txt', {type: 'text/plain'});
+  it('accepts multiple valid files', () => {
```
arjita-mitra: also worth testing manually uploading to two conversations simultaneously to ensure queue doesn't leak state.

arjita-mitra: Multi-file upload with mixed states (3 active, 2 queued, 1 failed)

1. Verify cancel all disables only when all settled
2. Tab switch during upload

1. Status should persist (subscription should handle it)
2. Simultaneous multi-file uploads in different conversations

1. Each conversation shows only its uploads
2. Queue doesn't leak between conversations

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
cristianoliveira: All files are validated before the upload starts. The validation runs over the whole dropped batch, so if one file is not accepted or is too large, the whole batch is rejected and no upload is queued. The mixed valid and invalid test covers this case

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
cristianoliveira: I double checked this against the task just to make sure, so the queue is intentionally wide and capped at three active uploads, while each upload is associated with its conversation and snapshots are filtered before rendering. 

So the queue can be shared for concurrency without status leaking between conversations. The tests cover mixed active, queued and failed states, tab switching, and switching conversations

arjita-mitra (approved):

PR #22473, opened 2026-09-15
feat: add hidden MLS migration button [WPB-27445]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27445" title="WPB-27445" target="_blank">StoryWPB-27445</a>  [Web] Add a hidden button to manually start a Proteus to MLS migration
  </td></table>
  <br />
 

## Summary

Adds manual migration of individual group conversations from Proteus to MLS.

- Matches eligibility: Any active member can migrate a conversation belonging to their own team.
- Gates migration behind the MLS migration feature configuration.
- Opens confirmation after five clicks on the conversation’s protocol.
- Transitions through Mixed protocol, establishes the MLS group, and finalises migration to MLS.
- Prevents duplicate migrations using a shared Zustand store.
- Shows success or failure in the same modal.

https://github.com/user-attachments/assets/c3d114c1-73fe-4c05-ad8e-229ccb8e85cd


apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+  selfUser: Pick<User, 'teamId' | 'qualifiedId'>;
+  repository: MigrationRepository;
+  getFeature: () => Maybe<NonNullable<FeatureMLSMigration>>;
+  onConversationUpdated: (conversation: Conversation) => void;
+}): Task<Conversation, ManualMigrationFailure> => {
+  if (!canManuallyMigrateConversation(conversation, selfUser, getFeature())) {
+    return task.reject({stage: 'eligibility', reason: 'notAllowed', cause: Maybe.nothing()});
+  }
+
+  const key = conversation.qualifiedId.id;
```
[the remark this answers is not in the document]
e-maad: You can only migrate conversations that you own, change not required.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+  teamState: Pick<TeamState, 'teamFeatures'>;
+  repository: Pick<
+    ConversationRepository,
+    'updateConversationProtocol' | 'tryEstablishingMLSGroup' | 'safeEnsureConversationExists'
+  >;
+}
+
+export const ManualMigrationProtocolDetails = ({conversation, selfUser, teamState, repository}: Props) => {
+  const {translate} = useApplicationContext();
+  const [current, setCurrent] = useState(conversation);
```
ikotarac: better names are **currentConversation** and **setCurrentConversation**

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const {teamFeatures} = useKoSubscribableChildren(teamState, ['teamFeatures']);
+  useKoSubscribableChildren(current, ['isSelfUserRemoved', 'isGroupOrChannel']);
+  const eligible = canManuallyMigrateConversation(current, selfUser, Maybe.of(teamFeatures?.mlsMigration));
+
+  useEffect(() => {
+    setCurrent(conversation);
+  }, [conversation]);
+
+  useEffect(() => {
+    mounted.current = true;
```
ikotarac: better name would be **isMountedRef.current**. when I was reading it first I thought we are storing reference to some mounted component, and not just a boolean, that's why I added 'is', and adding 'Ref' makes it a bit more clear why do we need to access .current (it's clear it's not just a boolean).

I like to extract these simple pieces of functionality as separate custom hooks, I think it makes the component much easier to read, even if you just define them in the same file above the component.

For example:

```ts
const useIsMounted = () => {
    const isMountedRef = useRef(true);
   
    useEffect(() => {
      isMountedRef.current = true;

      return () => {
        isMountedRef.current = false;
      };
    }, []);

    return isMountedRef.current;
}
```

then in the component below we have less ref allocation, less effects, and we can potentially also extract and reuse this

```tsx
...
  const isMounted = useIsMounted();
...
```

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+  repository: Pick<
+    ConversationRepository,
+    'updateConversationProtocol' | 'tryEstablishingMLSGroup' | 'safeEnsureConversationExists'
+  >;
+}
+
+export const ManualMigrationProtocolDetails = ({conversation, selfUser, teamState, repository}: Props) => {
+  const {translate} = useApplicationContext();
+  const [current, setCurrent] = useState(conversation);
+  const [dialog, setDialog] = useState<MigrationDialogState>({phase: 'closed'});
```
ikotarac: also **dialogState**

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const isPending = useManualMigrationStore(state => state.pendingConversationIds.has(conversation.qualifiedId.id));
+  const isBusy = isPending || dialog.phase === 'running';
+  const taps = useRef(0);
```
ikotarac: **tapsCountRef**

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
ikotarac: I would potentially do the same for the 'do something after X clicks' functionality:

```tsx
useOnMultipleClicks({
  count: 5,
  elementSelector: '[data-uie-name="manual-migration-protocol"]'
}, () => {
   if (!eligible || isBusy || dialog.phase !== 'closed') {
      return;
    }

    setDialog({phase: 'confirmation'});
});
```

then **useOnMultipleClicks** defines only the functionality of doing something on X clicks, reseting the counter etc, and this invocation here uses it, so it's easier to read.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+    if (useManualMigrationStore.getState().pendingConversationIds.has(current.qualifiedId.id)) {
+      return;
+    }
+
+    setDialog({phase: 'running'});
+    const outcome = await manuallyMigrateConversation({
+      conversation: current,
+      selfUser,
+      repository,
+      getFeature: () => Maybe.of(teamState.teamFeatures()?.mlsMigration),
```
ikotarac, line 133: why are we passing function here? I don't think that this configuration can change during the migration so that we would need to pull its value multiple times.

I would just pass 
```ts
featureConfiguration: Maybe.of(teamState.teamFeatures()?.mlsMigration)
```

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+  conversation.teamId === selfUser.teamId &&
+  [CONVERSATION_PROTOCOL.PROTEUS, CONVERSATION_PROTOCOL.MIXED].includes(conversation.protocol) &&
+  feature.isJust &&
+  feature.value.status === FEATURE_STATUS.ENABLED &&
+  feature.value.config.allowManualMigration === true;
+
+export type ManualMigrationFailure = {
+  stage: 'eligibility' | 'busy' | 'initialise' | 'establish' | 'finalise';
+  reason: 'requestFailed' | 'notAllowed' | 'alreadyRunning' | 'missingGroup' | 'protocolUnchanged';
+  cause: Maybe<NonNullable<unknown>>;
```
ikotarac: is this the same as writing `Maybe<{}>`?

libraries/api-client/src/team/feature/featureList.schema.ts
```diff
@@ -113,6 +113,7 @@ const mlsE2EIdConfigSchema = z.object({
 });
 
 const mlsMigrationConfigSchema = z.object({
+  allowManualMigration: z.boolean().optional(),
```
ikotarac, line 116: why are all these optional?

it looks like they are all strictly there (even lockStatus one level up seems to be required). this will then simplify stuff later
https://staging-nginz-https.zinfra.io/v17/api/swagger-ui/#/default/get_MlsMigrationConfigB

apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.tsx
```diff
@@ -417,9 +417,12 @@ const ConversationDetails = forwardRef<HTMLDivElement, ConversationDetailsProps>
             isChannelPublic={isChannelPublic}
           />
 
-          <ConversationProtocolDetails
```
ikotarac, line 420: we could've just wrapped this label with a button:

```tsx
<MigrateProtocolTrigger
    selfUser={selfUser}
    teamState={teamState}
    repository={conversationRepository}>
        <ConversationProtocolDetails
            protocol={activeConversation.protocol}
            cipherSuite={activeConversation.cipherSuite} />
</MigrateProtocolTrigger>
```

or even better, I don't know why don't we read **teamState**, **conversationRepository** and **selfUser** from Context, but do prop drilling instead?

```tsx
<MigrateProtocolTrigger>
        <ConversationProtocolDetails
            protocol={activeConversation.protocol}
            cipherSuite={activeConversation.cipherSuite} />
</MigrateProtocolTrigger>
```

and then after proteus is killed we just remove `<MigrateProtocolTrigger>` and we're back to the old stuff.

libraries/api-client/src/team/feature/featureList.schema.ts
ikotarac: I took a look in wire-server repo, this looks a bit messy, so I would just leave it as is now :D

claude said this type on server translates to something like this, but I wouldn't change this now, I was just a bit surprised that almost every field is optional 😱 

```ts
type FeatureStatus = "enabled" | "disabled";
type LockStatus = "locked" | "unlocked";

interface MlsMigrationConfig {
  /** omitted from the response entirely if unset */
  startTime?: string;
  /** omitted from the response entirely if unset */
  finaliseRegardlessAfter?: string;
  /** always present in responses; defaults to false only when omitted on write */
  allowManualMigration: boolean;
}

interface MlsMigrationConfigB {
  status: FeatureStatus;
  lockStatus: LockStatus;
  config: MlsMigrationConfig; // always present, not optional
  /** legacy field, effectively always "unlimited" today */
  ttl?: string;
}
```

apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.tsx
ikotarac: this way we also achieve that details rendering is separate than protocol migration, making each component simpler and more focused

ikotarac (commented):
I will continue reviewing tomorrow, there might be some edge cases where multiple migrations are running in parallel, and I want to check out what happens if migration is interrupted.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
e-maad: We need live feature getter, flags can change during migration.

apps/webapp/src/script/page/rightSidebar/conversationDetails/conversationDetails.tsx
e-maad: I tried the wrapper approach, but it required a render prop to pass the updated conversation and activation callback back to the parent. That adds indirection and makes the call site more complex.

I’ve kept `ConversationProtocolDetails` rendered inside `MigrateProtocolTrigger`. The details component remains presentational, while migration state and callback wiring stay together in the trigger.

Other approach would be to propagate the click event, which also comes with own complexity.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+  const titleId = useId();
+  const cancelButton = useRef<HTMLButtonElement>(null);
+  const wrapper = useRef<HTMLDivElement>(null);
+  const {teamFeatures} = useKoSubscribableChildren(teamState, ['teamFeatures']);
+  useKoSubscribableChildren(current, ['isSelfUserRemoved', 'isGroupOrChannel']);
+  const eligible = canManuallyMigrateConversation(current, selfUser, Maybe.of(teamFeatures?.mlsMigration));
+
+  useEffect(() => {
+    setCurrent(conversation);
+  }, [conversation]);
```
ikotarac: here we are setting new conversation in state whenever prop changes, but down we also set 
```ts

const outcome = await manuallyMigrateConversation({
  onConversationUpdated: updated => {
        if (mounted.current) {
          setCurrent(updated);
        }
      },
});
```

who wins? this can be a race condition, or we just unnecessarily change the conversation twice to the same conversation?

I think that we need to propagate this change outside of this component, new conversation is needed for other things, not just to display current protocol and cipher suite. By this logic, I think this internal state is redundant, I would just read the conversation passed through the props, and make sure that **manuallyMigrateConversation** does indeed change this external state and propagate it everywhere.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
```diff
@@ -0,0 +1,250 @@
...
+    setDialog({
+      phase: 'feedback',
+      message: outcome.match({
+        Ok: () => translate('manualMlsMigrationSuccess'),
+        Err: error => {
+          const cause = error.cause;
+          if (cause.isJust) {
+            const backendError = isAxiosError(cause.value) ? cause.value.response?.data : cause.value;
+            if (isBackendError(backendError)) {
+              return translate(errorHandlerStrings[backendError.label] ?? 'manualMlsMigrationFailure');
```
thisisamir98, line 149: This looks up every backend `label` in `errorHandlerStrings`. That table is shared with login and SSO, so a migration 403 with `access-denied` shows "Please verify your details and try again". `no-conversation` shows the raw `CONVERSATION_NOT_FOUND` key.

Could we special-case only `mls-migration-criteria-not-satisfied` here, and keep `manualMlsMigrationFailure` for any other label?

A regression that helps: `tryEstablishingMLSGroup` rejects with `{label: 'access-denied'}`, the status text is the migration failure string, not the login one.

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+export const canManuallyMigrateConversation = (
+  conversation: Conversation,
+  selfUser: Pick<User, 'teamId'>,
+  feature: Maybe<NonNullable<FeatureMLSMigration>>,
+): boolean =>
+  conversation.isGroupOrChannel() &&
+  !conversation.isSelfUserRemoved() &&
+  !!conversation.teamId &&
+  conversation.teamId === selfUser.teamId &&
+  [CONVERSATION_PROTOCOL.PROTEUS, CONVERSATION_PROTOCOL.MIXED].includes(conversation.protocol) &&
```
thisisamir98, line 41: The ticket asks for a conversation admin. This is true for any active member of a same-team group or channel.

Backend `PUT` protocol uses `LeaveConversation`, so a regular member can change the protocol. See [Action.hs](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Conversation/Action.hs#L200).

Is the extra admin gate still wanted on the 5-click control, or are we matching backend on purpose? If we still need admin, `conversation.isAdmin(selfUser.qualifiedId)` (or the role repo) should be part of this check.

thisisamir98 (commented):
## Summary
Hidden 5-click control that walks a same-team Proteus or Mixed group through Mixed, MLS group establish, then MLS. The Task chain and the pending store are solid. A few runtime gaps are still worth fixing before merge: the pending key drops the domain, the local conversation copy does not refresh from in-place protocol updates, and backend errors can show login copy in this modal.

## Findings
### Blocking
None.

### Should fix
- **Pending key.** Store uses `qualifiedId.id` only. Two federated conversations with the same UUID would share one lock. Key with `stringifyQualifiedId`.
  - Test: start migration on A, same id different domain on B is not treated as already running.
- **Local conversation copy.** `setCurrent(updated)` gets the same object `updateProperties` mutated, so React skips the render. `protocol` is not knockout, so another client finishing MLS can leave this row as a PROTEUS button.
  - Test: details open on PROTEUS, mutate `conversation.protocol` in place to MLS, the row is no longer a button.
- **Error copy.** `errorHandlerStrings[label]` is the login/SSO table. `access-denied` becomes "Please verify your details and try again". Map only `mls-migration-criteria-not-satisfied` here.
  - Test: establish rejects with `{label: 'access-denied'}`, status shows `manualMlsMigrationFailure`.
- **Success test.** It only checks call order. The mock returns a new Conversation, so it would still pass if we established the original Proteus entity. Assert `tryEstablishingMLSGroup` got the Mixed `groupId`, ids, and members, and have the mock mutate the same entity.

### Nits
None that are worth delaying this.

## Questions
- WPB-27445 asks for a conversation admin. This allows any active same-team member. Backend `PUT` protocol is `LeaveConversation`, so any member can change it. Are we matching backend on purpose, or should the 5-click control still require `conversation.isAdmin(...)`?
- Mixed to MLS uses `checkMigrationCriteria`, which ignores `allowManualMigration` and `startTime`. A group with one Proteus-only member will stay Mixed after a 400. Is that the intended test path?

## What looks good
Re-reading the feature flag before the MLS update, and keeping remounts blocked through the Zustand store. Optional `allowManualMigration` in the schema is the right call for older backends.

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
```diff
@@ -0,0 +1,166 @@
...
+              )
+              .andThen(() =>
+                repository
+                  .safeEnsureConversationExists({conversationId: current.qualifiedId, groupId: current.groupId})
+                  .mapRejected(cause => failure('establish', cause)),
+              );
+      return established
+        .andThen(() =>
+          canManuallyMigrateConversation(current, selfUser, getFeature())
+            ? update(current, CONVERSATION_PROTOCOL.MLS)
```
thisisamir98, line 132: This sends Mixed to MLS as soon as the group exists. Backend `checkMigrationCriteria` does not look at `allowManualMigration` or `startTime`. The final step only succeeds if every participant already supports MLS, or `finaliseRegardlessAfter` has passed.

https://github.com/wireapp/wire-server/blob/develop/libs/wire-subsystems/src/Wire/ConversationSubsystem/MLS/Migration.hs#L51-L79

A group with one Proteus-only member will succeed Proteus to Mixed, then fail Mixed to MLS with `mls-migration-criteria-not-satisfied`, and stay Mixed.

If that is the intended test path, the current error is fine. If testers should not land in Mixed, we should check participant protocols first and fail before the MLS PUT.

apps/webapp/src/script/mls/MLSMigration/manualMigration.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    conversation.type(CONVERSATION_TYPE.ONE_TO_ONE);
+    expect(canManuallyMigrateConversation(conversation, selfUser, Maybe.just(feature))).toBe(false);
+  });
+
+  it('establishes the group between Mixed and MLS updates and reports both updated entities', async () => {
+    const {deps, repository} = arrange();
+    const order: string[] = [];
+    repository.updateConversationProtocol.mockImplementation(async (_current, protocol) => {
+      order.push(protocol);
+      return createConversation(protocol);
```
thisisamir98: This only asserts call order. The mock returns a new Conversation from `createConversation(next)`, so the test still passes if we call `tryEstablishingMLSGroup` with the original Proteus entity (no `groupId`) or with the wrong users.

Production `updateConversationProtocol` mutates the same entity and fills `groupId` after Mixed.

Could we make the mock write `protocol` and `groupId` onto the same object, and assert establish received that `groupId`, `qualifiedId`, `selfUser.qualifiedId`, and `participating_user_ids()`?

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
thisisamir98: Yes, and even without a race I think the local copy does not work in production.

`refreshConversationProtocolProperties` calls `ConversationMapper.updateProperties`, which mutates the entity and returns it. `setCurrent(updated)` then gives React the same reference, so it skips the render.

`protocol` is not knockout. The parent panel only re-renders if a subscribed observable changes (`receiptMode` is one). A protocol update from another client can leave this row on PROTEUS and still take the 5-click confirm. Confirm then fails with `notAllowed` if the entity is already MLS, which is fail-closed, but the hidden control should already be gone.

I would drop `current` and read `conversation` from props. If we need a render after mutation, bump a generation in `onConversationUpdated`.

Regression: details open on PROTEUS, mutate `conversation.protocol` to MLS in place, the protocol row should stop being a button.

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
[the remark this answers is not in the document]
thisisamir98: Team ownership is already gated by `conversation.teamId === selfUser.teamId`. The pending key is a different issue.

Two federated conversations can share a UUID and live on different domains. Keying the store on `qualifiedId.id` would mark the wrong one busy, and a second click on the first one would not see the lock.

Could we use `stringifyQualifiedId(conversation.qualifiedId)` in `tryStart`, `finish`, and the `isPending` selector?

Regression: same id, different domains, start A, B should still be allowed to start.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
e-maad: Handled here: 0164ae3172c8dffddbe6f80b63a70404e86ab24f

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
e-maad: This is an intended path.

apps/webapp/src/script/components/panel/conversationProtocolDetails/manualMigrationProtocolDetails.tsx
e-maad: We don't need to keep it separate. For this endpoint BE has limited responses. If it returns access-denied then showing 'Please verify your details and try again' make sense.

apps/webapp/src/script/mls/MLSMigration/manualMigration.ts
e-maad: Fixed in recent pushes.

e-maad:
> ## Summary
> Hidden 5-click control that walks a same-team Proteus or Mixed group through Mixed, MLS group establish, then MLS. The Task chain and the pending store are solid. A few runtime gaps are still worth fixing before merge: the pending key drops the domain, the local conversation copy does not refresh from in-place protocol updates, and backend errors can show login copy in this modal.
> 
> ## Findings
> ### Blocking
> None.
> 
> ### Should fix
> * **Pending key.** Store uses `qualifiedId.id` only. Two federated conversations with the same UUID would share one lock. Key with `stringifyQualifiedId`.
>   
>   * Test: start migration on A, same id different domain on B is not treated as already running.
> * **Local conversation copy.** `setCurrent(updated)` gets the same object `updateProperties` mutated, so React skips the render. `protocol` is not knockout, so another client finishing MLS can leave this row as a PROTEUS button.
>   
>   * Test: details open on PROTEUS, mutate `conversation.protocol` in place to MLS, the row is no longer a button.
> * **Error copy.** `errorHandlerStrings[label]` is the login/SSO table. `access-denied` becomes "Please verify your details and try again". Map only `mls-migration-criteria-not-satisfied` here.
>   
>   * Test: establish rejects with `{label: 'access-denied'}`, status shows `manualMlsMigrationFailure`.
> * **Success test.** It only checks call order. The mock returns a new Conversation, so it would still pass if we established the original Proteus entity. Assert `tryEstablishingMLSGroup` got the Mixed `groupId`, ids, and members, and have the mock mutate the same entity.
> 
> ### Nits
> None that are worth delaying this.
> 
> ## Questions
> * WPB-27445 asks for a conversation admin. This allows any active same-team member. Backend `PUT` protocol is `LeaveConversation`, so any member can change it. Are we matching backend on purpose, or should the 5-click control still require `conversation.isAdmin(...)`?
> * Mixed to MLS uses `checkMigrationCriteria`, which ignores `allowManualMigration` and `startTime`. A group with one Proteus-only member will stay Mixed after a 400. Is that the intended test path?
> 
> ## What looks good
> Re-reading the feature flag before the MLS update, and keeping remounts blocked through the Zustand store. Optional `allowManualMigration` in the schema is the right call for older backends.

Thanks for the review.

- **Pending key:** Not required as there is no possibility of a user to be part of teams from different domain.
- **Conversation state:** Addressed in `0164ae3172c8dffddbe6f80b63a70404e86ab24f`. Removed the local state and callback; the component reads the shared conversation directly. Tests cover remote updates and completion after remounting.
- **Error copy:** Not required.
- **Success test:** The mock now mutates the existing entity.
- **Admin gate:** Added `conversation.isAdmin(selfUser.qualifiedId)`, including checks for role changes during migration.
- **Mixed state after failure:** This is intentional. If finalisation criteria aren’t met, the conversation stays Mixed and can be retried later.

thisisamir98 (approved):

PR #22482, opened 2026-09-15
feat: add shared drive drag-and-drop multi-file upload(WPB-28297)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28297" title="WPB-28297" target="_blank">TaskWPB-28297</a>  [web] Add drag-and-drop multi-file upload in Shared Drive
  </td></table>
  <br />
 

Summary:
- Add multi-file drag-and-drop uploads to Shared Drive for editors when shared-drive-direct-upload is enabled.
- Extract shared drive file validation
- Reuse shared 500 MB file validation across Shared Drive upload entry points.
- Preserve backend duplicate-name handling and folder destinations.
- Prevent viewers and disabled-feature users from initiating uploads.
- Disable the conversation-level dropzone while the Shared Drive tab is active to prevent overlay conflicts and blank-screen behavior.
- Add coverage for multi-file uploads, permissions, feature-flag behavior, and conversation dropzone ownership.


apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadValidation.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see http://www.gnu.org/licenses/.
+ *
+ */
+
+import {Result} from 'true-myth';
+
+export type SharedDriveUploadRejectionReason =
+  'empty' | 'multipleFiles' | 'notAccepted' | 'notAllowed' | 'recycleBin' | 'tooLarge';
```
cristianoliveira, line 23: Since we support multiple files now, can we remove `multipleFiles` from this type (and the feedback branch/test)? I don’t think validation can return it anymore, so keeping it here feels a bit misleading

cristianoliveira (approved):

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadValidation.ts
arjita-mitra: good point will remove

PR #22484, opened 2026-09-15
Reject stale incoming rings after an expired 1:1 SETUP [WPB-27813]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-27813" title="WPB-27813" target="_blank">BugWPB-27813</a>  [Desktop] [Calling] Delayed ghost incoming call rings hours later on desktop for the recipient
  </td></table>
  <br />
 

## Summary
- Windows client rang a 1:1 SETUP five hours after the caller hung up. Poll froze over lock. On wake AVS replayed the old invite, reported age 0, and we trusted `shouldRing`.
- We now remember the last remote incoming SETUP per conversation. If AVS asks to ring and that invite is older than 30 seconds, we skip incoming UI and `rejectCall`. Fresh invites still ring. No SETUP on file still rings. `shouldRing=0` is untouched, so a live conference does not disappear after lock.

https://wearezeta.atlassian.net/browse/WPB-27813

## Test plan
- [x] `yarn nx run webapp:test --testFile=src/script/repositories/calling/CallingRepository.test.ts`
- [x] `yarn nx run webapp:test --testFile=src/script/repositories/calling/incomingCallInvite.test.ts`
- [x] Place a 1:1 call to an unlocked desktop. It should still ring.
- [x] Join a conference, lock Windows, unlock. The call should still be there without a ringtone.
- [x] Repro: incoming 1:1, lock long enough that poll dies, unlock. Must not ring.


apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -1161,6 +1163,15 @@ export class CallingRepository {
 
     const isFederated = this.core.backendFeatures.isFederated && qualified_conversation && qualified_from;
     const userId = isFederated ? qualified_from : {domain: '', id: from};
+    const isFromSelf = !!this.selfUser && matchQualifiedIds(this.selfUser.qualifiedId, userId);
+    const incomingSetupReceivedAtMs = new Date(time).getTime();
+
+    if (isIncomingSetupOffer(content) && !isFromSelf && Number.isFinite(incomingSetupReceivedAtMs)) {
+      this.incomingSetupReceivedAtByConversation.set(
```
zskhan, line 1170: The map is set but never cleared.
Also the setup timestamp is stored per conversation, not per call/session.

EnricoSchw (commented):
For this specific issue, I think your solution works well and can serve as a useful safety net.

However, I don't think the PR addresses the underlying problem. AVS is essentially a state machine that is updated through events. When the timer loop stops running on Windows, AVS effectively enters a sleep state and is no longer able to process those events.

As a result, this can affect any state that AVS reports for example, whether a call is currently in progress or whether someone has called, even if that event happened quite some time ago.

So while this PR helps protect against this particular case, I think we still need to address the underlying timer loop issue to solve the problem more generally.

apps/webapp/src/script/repositories/calling/CallingRepository.ts
thisisamir98: Leaving it.

A new SETUP overwrites the timestamp. If we clear the entry after a stale reject, the next `incomingh` with `shouldRing=1` sees nothing recorded, fail-opens, and rings. Same ghost.

`incomingh` has no sessid. Conversation, user, client, video, shouldRing, type. Per-session would be cleaner if we could join it. We cannot. Last SETUP for the conversation is the invite we actually have.

thisisamir98:
@EnricoSchw There is no way to guarantee we keep that timer alive. When the machine sleeps the OS halts the process. Chromium can freeze the renderer on lock too. AVS lives in the page, so poll dies with JS.

This PR is only the delayed-ring safety net after wake. It does not keep AVS processing events during lock.

For the broader stale-state concern, keeping the 500ms loop alive is the wrong bet. We should detect resume in the app and give AVS an explicit call: we were asleep for N milliseconds, recalibrate timers, re-age or drop pending invites and anything else you report from a frozen clock. `setBackground(0)` after a poll gap is not that. New AVS API, webapp invokes it on wake.

EnricoSchw:
@thisisamir98 Exactly, that’s what I mean. We need to work on this solution together, and it needs to originate from the Web side. It’s important to first understand how AVS works and then share the limitations and constraints you’re facing with us. From there, we can work together to find a solution that works well for everyone.

zskhan (approved):

PR #22495, opened 2026-09-16
feat(a11y): improve keyboard navigation in conversation search [WPB-22420]

cristianoliveira:
<table>
<td>
  </td></table>
  <br />
 

## Problem

When searching conversations, `Tab` moved to the conversation search icon, the expected behaviour was to jump to the first item in the conversation filtered list.

## Solution

Keep `Tab` moving from the search input to the first available result. Arrow keys now move actual focus through the filtered conversations and group participant results, while `Tab` and `Shift+Tab` leave the list normally.

`Enter` jumps to the selected conversation

If the first result is not mounted, the event keeps the native browser behaviour instead of trapping focus. The focus candidates are shared between the parent and list rendering so the keyboard state follows the results that are displayed.

## Tradeoffs

The result list uses roving focus, so repeated `Tab` leaves the list instead of visiting every result. Arrow keys are used to move within the list, and boundary wrapping is intentional.

## Demo

https://github.com/user-attachments/assets/68728d52-fce2-4b74-94a8-105d1902d398


apps/webapp/src/script/components/cellsGlobalView/cellsTable/cellsFilePreviewModal/cellsFilePreviewModal.test.tsx
```diff
@@ -156,13 +156,12 @@ describe('CellsFilePreviewModal', () => {
...
   it('keeps fullscreen modification actions available for editors', async () => {
-    const {fireAndForgetInvoker} = renderModal({selfUserDriveRole: CELLS_SELF_USER_DRIVE_ROLE.EDITOR});
+    renderModal({selfUserDriveRole: CELLS_SELF_USER_DRIVE_ROLE.EDITOR});
 
     expect(await screen.findByRole('dialog')).toBeInTheDocument();
     expect(screen.getByRole('button', {name: 'Download'})).toBeInTheDocument();
     expect(screen.getByRole('button', {name: 'Editing'})).toBeInTheDocument();
     expect(screen.getByRole('button', {name: 'More options'})).toBeInTheDocument();
-    await act(() => fireAndForgetInvoker.waitUntilAllSettled());
-    expect(screen.getByTitle('Collabora editor')).toBeInTheDocument();
```
cristianoliveira: Not sure why was this failing 🤔

apps/webapp/src/script/components/cellsGlobalView/cellsTable/cellsFilePreviewModal/cellsFilePreviewModal.test.tsx
```diff
@@ -156,13 +156,12 @@ describe('CellsFilePreviewModal', () => {
...
+    expect(await screen.findByTitle('Collabora editor')).toBeInTheDocument();
   });
```
cristianoliveira, line 167: This was a flaky test, I can split from this PR if you'd like to

arjita-mitra:
The main concern here is that focus is being coordinated across multiple layers:

the search input
the result list
a virtualized/filtered list
a focus state hook
DOM lookup by selector

That makes the behavior dependent on timing and render state. In a virtualized list the first result may not be mounted yet or a search query may change while focus is pending, or the user may leave the list before the queued focus operation resolves. In those cases its possible that the current approach can briefly focus the wrong element or preserve stale focus state.

To make it more robust

1. use a single source of truth for focus candidates and focus order. This way we can avoid having Conversations calculate part of the behavior while ConversationsList calculates another part.Then both the search input and conversation cells use the same controller.

```
type FocusController = {
  focusedId: string;
  focusFirst: () => boolean;
  focusNext: (currentId: string) => boolean;
  focusPrevious: (currentId: string) => boolean;
  reset: () => void;
};
```

2. Use refs instead of document.querySelector. global document.querySelector is more vulnerable to timing issues during virtualization lookups, stale focus, searches the entire document rather than the current conversation list. instead register real DOM refs for each conversation result(Map of conversation elements ref).

3. Use useLayoutEffect for focus synchronization. Because focus changes should happen after the DOM has been committed but before the browser paints.

4. current onBlur cancellation is useful but it should be tied to the focus registry/controller rather than only the <ul> element.it should also cancelfocus when the component unmounts/the active tab changes/the search input receives focus/list becomes empty. This prevents delayed virtualization callbacks from stealing focus after the user has left the list.

arjita-mitra (dismissed):

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
```diff
@@ -460,6 +492,21 @@ export const Conversations = ({
...
+
+      if (!conversationsFilter || !firstResult || !focusFirstMountedConversation()) {
+        return;
+      }
+
+      event.preventDefault();
+      setCurrentFocus(firstResult.id);
+    },
+    [conversationsFilter, conversationsForFocus, focusFirstMountedConversation, setCurrentFocus],
+  );
```
thisisamir98, line 516: When Tab is pressed in search, we cancel any pending list focus and only move if the first result is already mounted. If it is not, native Tab continues.

Arrow keys already go through `focusConversation`, which scrolls the virtualizer and focuses the row after it mounts. Tab from search does not. `onSearch` also scrolls `conversationListRef`, the outer ListWrapper, while the virtualizer scrolls the inner `ul`. If the list was scrolled before typing, the first match can still be unmounted, and Tab lands on the next control outside the results. That is the original bug.

Could `handleSearchTab` call `focusConversation(firstResult.id)` instead of `focusFirstMountedConversation()`, and only call `preventDefault` when that returns true? Then Tab would share the scroll-and-focus path with arrows.

Please add a regression: a long list scrolled away from the top, type a query whose first match is not in the window, press Tab, the first match is focused.

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
```diff
@@ -261,6 +264,35 @@ export const Conversations = ({
...
+      }),
+    [
+      currentTab,
+      currentTabConversations,
+      conversationsFilter,
+      currentFolder,
+      groupParticipantsConversations,
+      isGroupParticipantsVisible,
+    ],
+  );
```
thisisamir98, line 285: `conversationsForFocus` is the list Tab and arrows now follow. Enter from the search field still opens `currentTabConversations[0]`. Those orders differ on Recent search because `getConversationsWithHeadings` puts 1:1s first, while `currentTabConversations` keeps the mixed repository order.

Tab can land on a 1:1 while Enter opens a group that sat earlier in the unfiltered list.

Could Enter use `conversationsForFocus[0]` or `currentFocus`, so both keys target the first displayed result?

Please add a regression: a matching group appears before the first 1:1 in `currentTabConversations`, the people heading still shows the 1:1 first. Tab focuses the 1:1. Enter from search should open that same 1:1.

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationsList.tsx
```diff
@@ -175,6 +188,112 @@ export const ConversationsList = ({
...
+      if (
+        conversationIndex === -1 ||
+        !conversationFocusCandidates.some(conversation => conversation.id === conversationId)
+      ) {
+        return false;
+      }
+
+      rowVirtualizer.scrollToIndex(conversationIndex, {align: 'auto'});
+      setPendingFocusRequest({conversationId, contextKey: focusContextKey});
+      return true;
```
thisisamir98: This returns true as soon as we call `scrollToIndex` and store a pending request. `handleKeyDown` then updates `currentFocus`, so `tabIndex={0}` moves to a row that may not be mounted yet.

If that row never mounts, or the pending request is cancelled, the list can sit with no tabbable result while DOM focus is still on the previous cell. Coming back with Tab later has nowhere to land.

Could we still prevent the default arrow scroll when we queue, but only call `setCurrentFocus` after the layout effect actually focuses the element? If pending is cancelled, restore the previous id.

Please add a regression: ArrowDown toward a row outside the virtual window, then change the query before it mounts. A mounted row should keep `tabIndex={0}`, and focus should not stay on the stale id.

thisisamir98 (commented):
## Summary
Search Tab now tries to land on the first filtered result, and arrows move real focus through that same list, including group participant rows. The fallback still only works if that first row is already mounted, so a scrolled virtualized list can miss the result and Tab goes back to the old next control. Enter from the search field also still uses a different first item than Tab.

## Findings
### Blocking
None.

### Should fix
- **Tab from search.** `handleSearchTab` cancels pending focus and only moves if the first result is already registered. Arrows use `focusConversation`, which scrolls the virtualizer. Search itself scrolls the outer ListWrapper, not the inner `ul`. A scrolled list can still send Tab outside the results.
  - Test: long list scrolled down, type a query, Tab. First match should be focused.
- **Enter vs Tab.** Enter from search opens `currentTabConversations[0]`. Tab focuses `conversationsForFocus[0]`, which is heading-reordered on Recent. They can open different conversations.
  - Test: matching group earlier in the raw list than the first 1:1 under the people heading. Tab and Enter should hit the same conversation.
- **Queued focus reported as done.** `focusConversation` returns true after queueing a scroll. `currentFocus` then points at a row that may never mount, so the list can lose its tab stop.
  - Test: ArrowDown to an off-screen row, change the query before it mounts. `tabIndex={0}` should stay on a mounted row.

### Nits
- Focus ids and candidate dedupe use `conversation.id`. Nearby search matching uses `matchQualifiedIds`. I would key the registry and the candidate list with the qualified id.
- The Conversations tests give the list a 1000px box, so TanStack mounts every row. Pulling nodes out of the DOM by hand is not the same as a real virtualizer window.

## Questions
- Tab from search now skips pending connection requests, which still sit above the list with `tabIndex={0}`. Is that what we want while searching?

## What looks good
- One candidate list for conversation name matches and group participants, so arrows follow what is on screen.
- Failed DOM focus does not trap the keyboard.
- Pending focus is dropped on blur, unmount, tab change, and result changes.
- Parent tests cover empty results, query changes that keep search focused, and unmount.

cristianoliveira:
> The main concern here is that focus is being coordinated across multiple layers...

Thanks, this makes sense. I agree that a shared focus controller and ref based registry would be more robust, especially with virtualization

For this PR I kept the change narrower and fixed the concrete focus issues around search Tab, Enter ordering, and pending focus. The broader controller refactor would touch the focus architecture across `Conversations` and `ConversationsList`, so I’d rather handle that separately instead of expanding this PR

I’ll keep this as a follow up improvement

cristianoliveira:
> Tab from search now skips pending connection requests, which still sit above the list with `tabIndex={0}`. Is that what we want while searching?

Good question. I kept connection requests outside the search focus candidates, so when searching, Tab moves to the first matching conversation instead of the connection request control

That matches the current scope of this change, but I agree the expected behavior should be made explicit if we want connection requests included in the search focus order

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
cristianoliveira: > When Tab is pressed in search, we cancel any pending list focus and only move if the first result is already mounted

Fixed in `52ad888f2a`. Search Tab now uses the same `focusConversation` path as arrow navigation, so the virtualizer can scroll to an unmounted result before focusing it. I also added regression coverage for the search focus behavior

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversations.tsx
cristianoliveira: > Enter from the search field still opens `currentTabConversations[0]`

Fixed in `52ad888f2a`. Enter now uses `conversationsForFocus[0]`, so it follows the same displayed ordering as Tab. The existing parent focus test covers the shared navigation path

apps/webapp/src/script/page/leftSidebar/panels/conversations/conversationsList.tsx
cristianoliveira: > This returns true as soon as we call `scrollToIndex` and store a pending request

Fixed in `52ad888f2a`. Queued focus now returns a pending result and keeps the current tab stop unchanged until the layout effect successfully focuses a mounted row. Added a regression covering the queued virtual focus case

cristianoliveira:
> Focus ids and candidate dedupe use `conversation.id`. Nearby search matching uses `matchQualifiedIds`

> The Conversations tests give the list a 1000px box, so TanStack mounts every row

For the remaining nits: I left the focus registry keyed by `conversation.id`, matching the candidate list and registered elements used by this flow. Changing that to qualified IDs would be a separate consistency change rather than part of the reported focus bug

I also agree that the current parent tests use a large viewport and do not fully exercise a real virtualizer window. The focused queued-result regression covers the stale tab-stop behavior, while a fully virtualized browser-level test would be better as follow-up coverage

arjita-mitra (approved):

PR #22505, opened 2026-09-17
feat: support folder selection in shared drive uploads(WPB-28308)

arjita-mitra:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28308" title="WPB-28308" target="_blank">TaskWPB-28308</a>  [web] Uploading a entire folder with progress management
  </td></table>
  <br />
 

Wire the existing Upload folder menu to a dedicated Shared Drive folder picker and route selected files through SharedDriveUploadController.

Use an explicit selection mode for file and folder inputs while preserving relative paths and keeping the conversation attachment flow unchanged.


apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
```diff
@@ -129,6 +129,35 @@ describe('handleSharedDriveDroppedFiles', () => {
     );
   });
 
+  it('ignores macOS metadata files before uploading dropped files', async () => {
```
cristianoliveira: Shouldn't it be ignore hidden files?

cristianoliveira: then we can give to it a list of hidden files to check

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadInput.test.tsx
```diff
@@ -79,4 +79,31 @@ describe('SharedDriveUploadFileInput', () => {
...
+    expect(input).toBeInstanceOf(HTMLInputElement);
+    expect(input).toHaveAttribute('directory', '');
+    expect(input).toHaveAttribute('webkitdirectory', '');
+
+    await user.upload(input as HTMLInputElement, [firstFile, secondFile]);
+
+    const uploadAction = getUploadAction(dependencies.fireAndForgetInvoker);
+    await uploadAction();
+
+    expect(dependencies.sharedDriveUploadController.upload).toHaveBeenCalledWith(
```
cristianoliveira, line 102: The test says it verifies relative paths, but the controller is mocked and the assertion only checks that the same `File[]` is forwarded. It would still pass if `webkitRelativePath` was ignored before the request is created

Could we assert the resulting `uploadNode` calls for both nested files? I’d expect paths like `uploadPath/Reports/first.txt` and `uploadPath/Reports/Archive/second.txt`

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
```diff
@@ -124,10 +129,28 @@ describe('ConversationTabs', () => {
     expect(view.getByTestId('shared-drive-tab-upload-uploading')).toHaveClass(
       'conversation-tabs__upload-status-icon--uploading',
     );
+    expect(view.getByTestId('shared-drive-tab-upload-uploading').querySelector('path')).toHaveAttribute(
```
cristianoliveira: These tests query the internal `path` element and assert its `fill` attribute. That makes the conversation tests depend on how the icon is implemented, rather than on stable behavior

Could we keep these tests focused on the status text and icon presence, and test the icon color closer to the icon component or in a visual test?

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveDrop.test.ts
arjita-mitra: addressed 197c2570f49341025c6195ec9af6988c189c2f80

apps/webapp/src/script/components/conversation/conversationCells/sharedDriveUploadInput.test.tsx
arjita-mitra: addressed b87c27f5d6d23dcd86017e534063da162719a5bb

apps/webapp/src/script/components/conversation/conversationTabs/conversationTabs.test.tsx
arjita-mitra: Removed the assertions that query the internal SVG path and inspect its fill attribute. The color can be covered by the icon component’s own test or visual regression coverage.

cristianoliveira (approved):

PR #22506, opened 2026-09-17
feat: handle mls reset system message [WPB-26398]

e-maad:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26398" title="WPB-26398" target="_blank">BugWPB-26398</a>  [Web] Session was reset system message not shown
  </td></table>
  <br />
 

## Summary
- Handle and persist `conversation.mls-reset` system messages.
- Show “Session successfully reset” to the resetting user.
- Show other users the resetting user’s name and recovery explanation.

https://github.com/user-attachments/assets/52674e48-cbe0-4dfb-be1d-a374dc651ef9


zskhan (approved):
LGTM 🚀

apps/webapp/src/script/repositories/conversation/EventMapper.ts
```diff
@@ -305,6 +306,11 @@ export class EventMapper {
         break;
       }
 
+      case CONVERSATION_EVENT.MLS_RESET: {
+        messageEntity = createMLSResetMessage(this.translate);
```
thisisamir98, line 310: I think this maps the wrong event for WPB-26398.

`conversation.mls-reset` is what the backend sends when an MLS group is reset (new group id and epoch). See [ConvReset in wire-server](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Event/Conversation.hs#L231).

The ticket is the Proteus "Reset session" button in conversation details, devices. iOS turns protobuf `ClientAction.RESET_SESSION` into a session-reset system message. Android and kalium do the same (`CryptoSessionReset` in `handleClientAction`).

On web that protobuf type never becomes a stored message. `CryptographyMapper._mapGenericMessage` hits `default` and skips `clientAction`:
https://github.com/wireapp/wire-webapp/blob/feat/mls-reset-system-msg-WPB-26398/apps/webapp/src/script/repositories/cryptography/CryptographyMapper.ts#L293-L297

So the QA path still shows nothing. iOS resets the web device, web should show "User X was unable to decrypt...", and it will not.

Can we map `ClientAction.RESET_SESSION` to this system message, like kalium? If we still want a chat line for MLS group reset, that should be a separate message with MLS copy.

Please add a regression: receive a decrypted `clientAction: RESET_SESSION` in a 1:1, persist it, and assert the other-user caption with the sender name. The current EventMapper test would still pass if that path stayed skipped.

apps/webapp/src/script/repositories/entity/message/mlsResetMessage.ts
```diff
@@ -0,0 +1,36 @@
...
+import {SystemMessage} from './systemMessage';
+
+import {SystemMessageType} from '../../../message/systemMessageType';
+
+export const createMLSResetMessage = (translate: Translate): SystemMessage => {
+  const message = new SystemMessage(translate);
+  message.system_message_type = SystemMessageType.MLS_RESET;
+  // The sender is resolved after mapping the event, so determine the caption when it is read.
+  Object.defineProperty(message, 'caption', {
+    get: () => translate(message.user().isMe ? 'mlsResetSelf' : 'mlsReset'),
```
thisisamir98, line 31: These keys reuse the Proteus session-reset wording from iOS (`content.system.session_reset.other`). That describes a device session that could not decrypt messages, not an MLS group reset (wipe old group, new group id, epoch 0).

If this message stays on `conversation.mls-reset`, I would use MLS-specific copy. For the ticket, match iOS:
- other user: `{name} was unable to decrypt some of your messages but has solved the issue. This affected all conversations you share together.`
- self: `You were unable to decrypt some of your messages but you solved the issue. This affected all conversations.`

`mlsResetSelf` currently says "Session successfully reset", which is the preferences confirmation, not the conversation system message.

apps/webapp/src/script/repositories/conversation/ConversationRepository.ts
```diff
@@ -4826,6 +4826,7 @@ export class ConversationRepository {
       });
 
       await this.saveConversationStateInDb(updatedConversation);
+      await this.addEventToConversation(updatedConversation, eventJson);
```
thisisamir98, line 4829: `addEventToConversation` only runs after wipe and group-id update succeed. Failures are swallowed by the `catch` below. `MLS_RESET` is now in `EventTypeHandling.STORE`, so EventStorageMiddleware already saved the event before this handler runs.

If wipe throws, the live conversation does not get the message, but a reload maps the stored event and can show a successful reset caption. That looks like the reset worked when this client did not apply it.

Could we add the chat message only after a successful local reset, and avoid claiming success when the handler bails? One way is to inject a client event after success instead of storing the raw backend event unconditionally.

Test: wipe throws, then reload conversation events, assert we do not show a successful reset caption.

thisisamir98 (changes requested):
## Summary
This persists `conversation.mls-reset` and renders it in the timeline, but that is not the event from WPB-26398. The ticket is Proteus session reset from conversation details, devices. iOS and Android persist a system message from protobuf `ClientAction.RESET_SESSION`. Web still skips that type, so the reported 1:1 path stays empty.

## Findings
### Blocking
- **Wrong event for the ticket.** Mapping `conversation.mls-reset` does not cover iOS "Reset session" on a device. That sends `ClientAction.RESET_SESSION`. `CryptographyMapper` still drops `clientAction`, so web will not show "User X was unable to decrypt..." after the QA steps.
  - Test: decrypt a `RESET_SESSION` client action in a 1:1, persist it, assert sender name plus the other-user caption.

### Should fix
- **Copy is Proteus session-reset language on an MLS group-reset event.** `mlsReset` matches iOS session-reset copy. MLS reset changes group id and epoch. If we keep a message for `conversation.mls-reset`, it needs its own wording. For session reset, iOS self copy is "You were unable to decrypt some of your messages...", not "Session successfully reset".
- **Stored success after a failed local reset.** `MLS_RESET` is now stored before `onMLSResetMessage` runs. Wipe failures are caught and skip `addEventToConversation`. A reload still maps the stored event and can show a successful reset caption.

## Questions
- Do we also want a distinct system message for real MLS group resets, or is this PR only meant to close the session-reset ticket?

## What looks good
- Storing, mapping, and adding to the conversation is the right shape if the event is correct.
- The UI test for sender name vs self caption is the right kind of check.
- WallClock in the new tests is good.

PR #22507, opened 2026-09-17
Add meeting link to Meet Now and Schedule Meeting flows. [WPB-28125]

zskhan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28125" title="WPB-28125" target="_blank">TaskWPB-28125</a>  [Web] Always create the meeting join link
  </td></table>
  <br />
 

Add meeting link to Meet Now and Schedule Meeting flows. If password is provided then link will be password protected.

  ## Changes

  - Add shared PasswordFields and MeetingLinkForm components.
  - Add password and confirmation fields to meeting form state and commands.
  - Validate optional passwords, mismatches, invalid passwords, and confirmation-only input.
  - Request conversation guest-link codes with the configured password after meeting creation.
  - Support password generation and copy-to-clipboard actions inside meeting forms.
  - Fix Meet Now validation logic that prevented valid meetings from being submitted.
  - Add unit coverage for password validation, form mapping, service behavior, and shared password fields.
  - Add an end-to-end test for creating a password-protected scheduled meeting link.
  - Rename MeetingGuestLinkForm to MeetingLinkForm.

Meet Now Form
image

Schedule Meeting Form
image


apps/webapp/src/script/components/meeting/shared/service/meetingService.ts
```diff
@@ -89,21 +90,31 @@ const createMeetingAndSyncParticipants = (
...
-      ),
+      )
+        .andThen(() =>
+          (deps.conversationRepository.requestMeetingConversationCode
+            ? deps.conversationRepository.requestMeetingConversationCode(
+                createdMeeting.qualified_conversation,
+                password,
+              )
+            : task.resolve(undefined)
+          ).mapRejected(() => meetingSubmitErrors.conversationSetupFailed),
```
thisisamir98: Could we always call `requestMeetingConversationCode` here?

The point of this ticket is to POST `/conversations/{id}/code` after every successful meeting create. This ternary treats a missing method as success, so a `conversationRepository` without that method still creates the meeting and never creates the link. Production has the method, but the happy-path tests only assert the call when a password is set, so the skip would stay green.

I would call it unconditionally, including when `password` is `undefined`.

A failed code request also stops participant sync, and Meet Now never starts the call. The meeting is already saved, the modal closes, and the user cannot repair the missing code from this form. Could we still add the selected users, and let Meet Now join, after a code failure? Galley will 409 `guest-links-disabled` when that team feature is off: https://github.com/wireapp/wire-server/blob/develop/libs/wire-subsystems/src/Wire/ConversationSubsystem/Query.hs#L709-L719

Please add a regression: create with no password still calls `requestMeetingConversationCode(conversationId, undefined)`. And one where the code request rejects, selected users are still added.

apps/webapp/src/script/components/PasswordFields/PasswordFields.tsx
```diff
@@ -0,0 +1,97 @@
...
+  isPasswordInputMarkInvalid,
+  passwordConfirmationValue,
+  onPasswordConfirmationChange,
+  isPasswordConfirmationMarkInvalid,
+}: PasswordFieldsProps) => (
+  <>
+    <Input
+      name="guest-link-password"
+      data-uie-name="guest-link-password"
+      required
```
thisisamir98: These inputs are always `required`. That is correct for guest-link create. Meeting passwords are optional, and Meet Now only survives empty fields because the form sets `noValidate`. Screen readers still hear required fields, and `PasswordFields.test.tsx` now locks `toBeRequired()`.

Could this take a `required` prop? Default `true` for `GuestLinkPasswordForm`, `false` for `MeetingLinkForm`.

A regression that empty password and confirmation still submit a meeting would lock that the fields are optional.

apps/webapp/src/script/components/meeting/shared/validation/meetingPasswordValidation.ts
```diff
@@ -0,0 +1,39 @@
...
+  password?: string,
+  passwordConfirmation?: string,
+): MeetingPasswordErrorKey | undefined => {
+  const hasPassword = Boolean(password?.trim());
+  const hasPasswordConfirmation = Boolean(passwordConfirmation?.trim());
+
+  return (hasPassword || hasPasswordConfirmation) &&
+    (!hasPassword || !isValidPassword(password ?? '') || password !== passwordConfirmation)
+    ? meetingPasswordErrorKey
+    : undefined;
```
thisisamir98: This uses one key for mismatch, confirmation-only, and an invalid password. The form then uses that key only to mark both fields invalid, so a mismatch shows the min-length helper from `PasswordFields`.

`setPassword` and `setPasswordConfirmation` also leave `errors.password` untouched. After a failed submit, typing a valid matching password still looks invalid until the next submit.

Could we split invalid vs mismatch, show that copy on the fields, and refresh the password error while typing? Title already does live validation this way.

libraries/react-ui-kit/src/inputs/dateTimePickerField/dateTimePickerField.styles.ts
```diff
@@ -24,7 +24,7 @@ import {Theme} from '../../identity/theme';
 export const dateTimePickerContentStyles: CSSObject = {
   display: 'flex',
   flexDirection: 'column',
-  gap: '8px',
+  gap: 0,
```
thisisamir98: Nit: this shared date picker used to have an 8px gap. Setting it to 0 changes every `DateTimePickerField`, not only the meeting form. If this is only to match the meeting modal layout, a local override would be safer.

thisisamir98 (commented):
## Summary
This adds optional password fields on Meet Now and Schedule, then POSTs `/conversations/{id}/code` after meeting create. Password stays in memory and is sent only when the host set one. Direction is right. Two runtime gaps worry me: the code request can be skipped, and if it fails we also skip adding people and starting Meet Now, even though the meeting is already saved.

## Findings
### Blocking
None.

### Should fix
- **Code request can be skipped.** `createMeetingAndSyncParticipants` only calls `requestMeetingConversationCode` if the method exists. That is the opposite of always creating the link. Production `ConversationRepository` has the method, but a repository without it still reports success. Happy-path tests also only assert the call when a password is set.
  - Test: create with no password still calls `requestMeetingConversationCode(id, undefined)`.

- **Code failure aborts the rest of setup.** If POST `/code` fails, we never sync participants, Meet Now never joins, and the modal closes because `conversationSetupFailed` counts as persisted. The host gets a saved meeting with no invitees and no link, and the error says try again.
  - Test: code request rejects, selected users are still added, Meet Now still attempts join.

- **Optional password fields are marked required.** `PasswordFields` always sets HTML `required`. That matches guest-link create. For meetings the password is optional. Meet Now only works because the form has `noValidate`. Screen readers still hear required fields, and the new test locks `toBeRequired()`.
  - Test: empty password and confirmation still submit a meeting.

- **Password errors are one bucket and they stick.** Mismatch, confirmation-only, and invalid password all share one key. The UI uses that key only as a boolean, so a mismatch shows the min-length helper. `setPassword` also does not refresh `errors.password`, so a fixed password still looks invalid until the next submit.

### Nits
- `dateTimePickerField` gap went from `8px` to `0` in react-ui-kit. That changes every date picker, not only these modals. If this is only for the meeting form, a local override would be safer.

## Questions
- If the team has `conversationGuestLinks` disabled, Galley returns 409 `guest-links-disabled` on POST `/code`. Guest rooms already hide that action. Meetings always POST. Should meeting create fail in that team, or do we still want the meeting and just no guest link?
- `requestMeetingConversationCode` drops the POST response and does not map it onto the conversation the way `requestAccessCode` does. The later meeting-link view may need GET `/code` if the `CODE_UPDATE` event has not arrived yet. Intentional for this ticket?

## What looks good
- Password stays on the command only when it is non-empty. Edit mode does not show the fields.
- Generate and Copy are `type="button"`, so they do not submit the Meet Now form.
- Code-request failure is mapped to `conversationSetupFailed`, so a saved meeting is not treated as nothing was created.

apps/webapp/src/script/components/PasswordFields/PasswordFields.tsx
zskhan: 5d3f29601b

apps/webapp/src/script/components/meeting/shared/validation/meetingPasswordValidation.ts
zskhan: e78ee53914dd097924393d7a760dd991fa1c1343

libraries/react-ui-kit/src/inputs/dateTimePickerField/dateTimePickerField.styles.ts
zskhan: 15e1d5e900c68b1a98b9a78e73bd9b37c43130d5

apps/webapp/src/script/components/meeting/shared/service/meetingService.ts
zskhan: dd23e924b89b55780d9672266fde9c20517a176b

thisisamir98 (approved):

PR #22513, opened 2026-09-17
Show an in-app reminder 10 minutes before a meeting starts [WPB-28730]

thisisamir98:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-28730" title="WPB-28730" target="_blank">TaskWPB-28730</a>  [Web] Notify participants in-app 10 minutes before a meeting starts
  </td></table>
  <br />
 

## Summary

Scheduled meetings now show the same in-app card as invites, ten minutes before start. The web client computes T-10 from the meeting list. Backend does not send a reminder event.

Meet Now is skipped because T-10 is already in the past the moment you create it. Same for a late invite after T-10. No catch-up card if you join after the reminder window.

Edit the start time and the timer moves. Cancel or delete and it drops. Recurring series only schedule the next occurrence that is still in the future.

Hosts get it too. We walk the meeting store, so creating a meeting is enough. You do not need an invite card first.

View still opens the Meetings list. We never built scroll-to-row or the prep modal, so this ticket does not invent that. Dismiss only removes the card. No sound on purpose. The OS toast is WPB-28731. The scheduler already calls `onReminder`. Native can hook that later.

I ran this in Chrome with the offset temporarily at 10 seconds, watched the host card fire, then put the offset back to 10 minutes.

## Test plan

- [ ] Schedule a meeting more than 10 minutes out, wait until T-10, confirm the Reminder card names the meeting and says it starts in 10 minutes
- [ ] View opens Meetings and dismisses the card
- [ ] Dismiss only removes the card
- [ ] Meet Now does not remind
- [ ] Invite after T-10 does not remind
- [ ] Edit start time, confirm the reminder moves
- [ ] Cancel or delete before T-10, confirm no card
- [ ] Host sees the reminder on a meeting they created


apps/webapp/src/script/components/meeting/createMeetingReminderScheduler.ts
```diff
@@ -0,0 +1,198 @@
...
+      return;
+    }
+
+    if (wallClock.currentTimestampInMilliseconds < scheduledReminder.fireAt) {
+      armTimeout(meetingKey);
+      return;
+    }
+
+    scheduled.delete(meetingKey);
+    firedOccurrenceKeys.add(toFiredOccurrenceKey(meetingKey, scheduledReminder.occurrenceStartMs));
```
zskhan, line 170: Before adding the reminder notification there should be a check of 
`wallClock.currentTimestampInMilliseconds < occurrenceStartMs` to avoid showing stale notification if for some reason the system is in sleep mode and meeting is already started

thisisamir98: Skipped the card when the timeout fires after start. A late wakeup before start still reminds. Recurring series still arm the next occurrence.

https://github.com/wireapp/wire-webapp/commit/b089278cc920c89fddc276d7318baac9a3c874a1

zskhan (dismissed):

zskhan (approved):

PR #22536, opened 2026-09-18
fix: media acquisition and error handling when answering calls WPB-26346

EnricoSchw:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/WPB-26346" title="WPB-26346" target="_blank">BugWPB-26346</a>  [Web] “No camera access” although the microphone appears to be the main requirement for the call
  </td></table>
  <br />
 

This PR improves media acquisition when answering calls.

Previously, audio and camera were requested together. If the combined request failed, it was not possible to reliably determine whether the microphone or camera caused the error. This could result in the wrong error being shown and allowed an incoming call to be answered even when microphone access was unavailable.

Note: Changes to the outgoing startCall flow, refresh and start camera and audio should be done some times as well!


apps/webapp/src/script/repositories/media/MediaStreamHandler.ts
```diff
@@ -17,7 +17,9 @@
  *
  */
 
+import {isString} from '@sindresorhus/is';
```
zskhan, line 20:
```suggestion
import {isObject, isString} from '@sindresorhus/is';
```
Instead of underscore we can use it from `@sindresorhus/is`

apps/webapp/src/script/repositories/calling/CallingRepository.ts
```diff
@@ -808,30 +808,116 @@ export class CallingRepository {
...
+
+      await this.acquireCallMedia(call, {audio, camera});
+
+      if (camera && call.state() !== CALL_STATE.NONE) {
+        call.getSelfParticipant().videoState(VIDEO_STATE.STARTED);
       }
 
       return true;
     } catch (error: unknown) {
+      this.logger.warn('warmupMediaStreams failed', {
```
zskhan, line 910: We have another logger couple of lines below.
We can remove one.

zskhan:
LGTM 👍 I verified it by running it locally and all the permission on fresh session including screen-sharing permission works well.

zskhan (dismissed):

zskhan (approved):

